PR #2706, opened 2023-02-20
fix: Serialization of nullable booleans

philipphofmann:
## :scroll: Description

The protocol contains a couple of nullable BOOLs that are of type NSNumber. Setting these properties to something different than 0 or 1 leads to a serialized NSNumber, which violates the protocol. With this refactoring 0 value always means false, and any nonzero value is interpreted as true when serializing such nullable BOOLs.

Came up while investigating issues for MetricKit, as the server complained for invalid boolean values.
Screenshot 2023-02-20 at 11 50 49

Unit tests.

## :pencil: Checklist


brustolin (approved):
LGTM

Issue #2713, opened 2023-02-23
watchOS crash in SentryANRTrackingIntegration


leoMehlig:
### Platform

watchOS

### Installed

Swift Package Manager

### Version

8.1.0

### Steps to Reproduce

Via Sentry, I saw a crash happening in the SentrySDK on watchOS. From looking at the code, it appears a stack trace is not available when reporting an app hang in SentryANRTrackingIntegration:71.

Here is the relevant part of the stack trace:

```
OS Version: watchOS 9.3.1 (20S664)
Report Version: 104

Crashed Thread: 8

Application Specific Information:
*** -[__NSArrayM objectAtIndexedSubscript:]: index 0 beyond bounds for empty array

Thread 8 Crashed:
0   CoreFoundation                  0x50e4690c          __exceptionPreprocess
1   libobjc.A.dylib                 0x4faba7e8          objc_exception_throw
2   CoreFoundation                  0x50e669c4          _CFThrowFormattedException
3   CoreFoundation                  0x50dd2f58          -[__NSArrayM objectAtIndexedSubscript:]
4   Structured                      0xeaf024            -[SentryANRTrackingIntegration anrDetected] (SentryANRTrackingIntegration.m:71)
5   Structured                      0xeae6bc            -[SentryANRTracker ANRDetected] (SentryANRTracker.m:111)
6   Structured                      0xeae258            -[SentryANRTracker detectANRs] (SentryANRTracker.m:98)
7   Foundation                      0x4fb6c20c          __NSThread__start__
8   libsystem_pthread.dylib         0xe92f60d4          _pthread_start
9   libsystem_pthread.dylib         0xe92f5524          thread_start
```

### Expected Result

No crash.

### Actual Result

Crash

brustolin:
Thanks @leoMehlig, we will investigate. If you have a repro that you could provide us, it would help a lot.

philipphofmann:
How to repro after merging https://github.com/getsentry/sentry-cocoa/pull/2714

1. Run `make build-for-watchos`
2. Ensure not to debug the watchOS sample otherwise, the ANRTrackingIntegration is disabled
3. Run watchOS sample
4. Click `Cause ANR` in the sample

philipphofmann:
@brustolin, the fix for avoiding the crash should be easy 

We need to ensure that `threads` has at least one element.


[...]

PR #2714, opened 2023-02-23
test: Add cause ANR for watchOS sample

philipphofmann:
Repro for https://github.com/getsentry/sentry-cocoa/issues/2713.

#skip-changelog

### DIFF
--- a/Samples/watchOS-Swift/watchOS-Swift WatchKit Extension/ContentView.swift
+++ b/Samples/watchOS-Swift/watchOS-Swift WatchKit Extension/ContentView.swift
@@ -2,6 +2,9 @@ import Sentry
 import SwiftUI
 
 struct ContentView: View {
+    
+    @StateObject var viewModel = ContentViewModel()
+    
     var addBreadcrumbAction: () -> Void = {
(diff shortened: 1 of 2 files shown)


Samples/watchOS-Swift/watchOS-Swift WatchKit Extension/ContentView.swift
```diff
@@ -56,11 +59,36 @@ struct ContentView: View {
...
+    func causeANR() {
+        
+        var i = 0
+        
+        for _ in 0...5_000_000 {
+            i += Int.random(in: 0...10)
+            i -= 1
+            
+            anrText = "\(i)"
+        }
```
brustolin, line 86: l: I would prefer a thread.sleep that is more predictable.

brustolin (approved):
LGTM

Samples/watchOS-Swift/watchOS-Swift WatchKit Extension/ContentView.swift
philipphofmann: I copied this logic from iOS-Swift. I didn't use Thread.sleep back then cause this is a more realistic use case of blocking the main thread. I would like to stick to it.

PR #2716, opened 2023-02-23
Triggering `profile-data-generator` to test profile issues

wedamija:
Just looking to trigger https://github.com/getsentry/sentry-cocoa/blob/main/.github/workflows/profile-data-generator.yml, so that we can see some profile issues in production


philipphofmann:
@wedamija, you can also add the following to profile-data-generator.yml
```
  push:
    branches:
      - danf/profile-data-trigger
```

Then only this workflow runs for your specific branch, without you having to open a PR.

wedamija:
I think we can close this and use the branch method later if we need to trigger more.

PR #2717, opened 2023-02-24
chore: Use YES and NO for objc BOOL

brustolin:
Changing `true` to `YES` and `false` to `NO` in objc files.

_#skip-changelog_


philipphofmann (approved):
Awesome, thanks @brustolin 👏

PR #1234, opened 2021-07-21
feat: Sub-millis precision for spans and events

philipphofmann:
## :scroll: Description

The SDK now sends the timestamp and startTimestamp of span and event in the precision of
NSTimestamp of timeIntervalSince1970 instead of the limited milliseconds precision
of sentry_toIso8601String.

To keep the scope of this PR small, sessions, breadcrumbs, sanitizing arrays and dicts, etc. are handled in this issue https://github.com/getsentry/sentry-cocoa/issues/1214.

## :bulb: Motivation and Context

Spans and transactions are limited to millisecond precision. Instead, we want sub-millisecond precision, because we can.

## :green_heart: How did you test it?
Sample transactions and unit tests.

## :pencil: Checklist

- [x] I reviewed the submitted code
- [x] I added tests to verify the changes
- [x] I updated the docs if needed
- [x] Review from the native team if needed
- [x] No breaking changes

PR #2724, opened 2023-02-27
feat: Time to initial display and time to full display

brustolin:
## :scroll: Description

A new Span is created to track time for initial display in automatic `UIViewController` performance tracking.

A new API was added to `SentrySDK` where users may report a full display. This is meant to be called when the view controller content is fully loaded and ready to be used.

Adding more information during view controller life cycle

Unit tests

## :pencil: Checklist


Sources/Sentry/Public/SentrySDK.h
```diff
@@ -301,6 +301,14 @@ SENTRY_NO_INIT
  */
 + (void)crash;
 
+/**
+ * Report to the current ongoing UIViewController transaction
+ * that the screen is fully loaded and ready to be used,
+ * this creates a new span with the amount of time it took
+ * from initializing the view controller until it is ready.
```
philipphofmann, line 308: `m`:
```suggestion
 * Report to the current ongoing UIViewController transaction
 * that the screen content is fully loaded and displayed.
 * This adds a span and measurement to the transaction.
```

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -462,6 +462,13 @@ NS_SWIFT_NAME(Options)
 
 #endif
 
+/**
+ * Use this in conjunction with `enableUIViewControllerTracing`.
+ * By enabling this, every UIViewController tracing transaction will wait
+ * for a call to `SentrySDK.reportFullDisplay()` until the transaction times out.
```
philipphofmann, line 468: `m`: I think we have to be more specific here. What happens when you don't call `SentrySDK.reportFullDisplay()`? Does the SDK throw the transaction away after the timeout?

Sources/Sentry/Public/SentryHub.h
```diff
@@ -249,6 +249,14 @@ SENTRY_NO_INIT
  */
 - (void)setUser:(SentryUser *_Nullable)user;
 
+/**
+ * Indicates that your screen is fully loaded and ready to be used.
```
philipphofmann, line 253: `m`: Please use the same comments as in `SentrySDK.reportFullDisplay()`.

Sources/Sentry/SentryHub.m
```diff
@@ -668,6 +670,17 @@ - (BOOL)envelopeContainsEventWithErrorOrHigher:(NSArray<SentryEnvelopeItem *> *)
     return NO;
 }
 
+- (void)reportFullDisplayed
+{
+#if SENTRY_HAS_UIKIT
+    if (_client.options.enableTimeToFullDisplay) {
+        [SentryUIViewControllerPerformanceTracker.shared reportTTFD];
```
philipphofmann, line 677: `l`: I would stick to the same naming everywhere.
```suggestion
        [SentryUIViewControllerPerformanceTracker.shared reportFullDisplayed];
```

Sources/Sentry/SentryHub.m
```diff
@@ -668,6 +670,17 @@ - (BOOL)envelopeContainsEventWithErrorOrHigher:(NSArray<SentryEnvelopeItem *> *)
     return NO;
 }
 
+- (void)reportFullDisplayed
+{
+#if SENTRY_HAS_UIKIT
```
philipphofmann, line 675: `m`: I'm unsure why CI didn't run yet for this branch. CodeCov should comment that this is uncovered by tests.

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -462,6 +462,13 @@ NS_SWIFT_NAME(Options)
 
 #endif
 
+/**
+ * Use this in conjunction with `enableUIViewControllerTracing`.
```
philipphofmann, line 466: `m`: Please mention the default value in the code docs.

Sources/Sentry/SentryOptions.m
```diff
@@ -324,6 +325,9 @@ - (BOOL)validateOptions:(NSDictionary<NSString *, id> *)options
     [self setBool:options[@"enableCaptureFailedRequests"]
             block:^(BOOL value) { self->_enableCaptureFailedRequests = value; }];
 
+    [self setBool:options[@"enableTimeToFullDisplay"]
```
philipphofmann, line 328: `m`: Missing tests for this in `SentryOptionsTests.m`

Sources/Sentry/SentrySDK.m
```diff
@@ -383,6 +383,17 @@ + (void)installIntegrations
     }
 }
 
+/**
+ * Indicates that your screen is fully loaded and ready to be used.
+ * This should be used together with automatic performance transaction,
+ * a full display span will be created if there is a view controller transaction
+ * happening.
+ */
```
philipphofmann, line 391: `l`: We don't need comments in the `.m` file, as we already have them in the `.h` file.
```suggestion
```

Sources/Sentry/include/SentrySpanOperations.h
```diff
@@ -1,5 +1,7 @@
 #import <Foundation/Foundation.h>
 
 static NSString *const SentrySpanOperationUILoad = @"ui.load";
+static NSString *const SentrySpanOperationUILoadTTID = @"ui.load.initial_display";
+static NSString *const SentrySpanOperationUILoadTTFD = @"ui.load.full_display";
```
philipphofmann, line 5: `l`: Again I would stick to the same naming.
```suggestion
static NSString *const SentrySpanOperationUILoadInitialDisplay = @"ui.load.initial_display";
static NSString *const SentrySpanOperationUILoadFullDisplay = @"ui.load.full_display";
```

Sources/Sentry/SentryTracer.m
```diff
@@ -543,12 +578,16 @@ - (void)trimEndTimestamp
 
 - (SentryTransaction *)toTransaction
 {
+    // TODO: use middleware to create appStartSpans
```
philipphofmann, line 581: `m`: There is still a todo here.

Sources/Sentry/SentryTracer.m
```diff
@@ -40,7 +41,11 @@
  */
 static const NSTimeInterval SENTRY_APP_START_MEASUREMENT_DIFFERENCE = 5.0;
 static const NSTimeInterval SENTRY_AUTO_TRANSACTION_MAX_DURATION = 500.0;
+#if SENTRY_DEVELOPMENT
+static const NSTimeInterval SENTRY_AUTO_TRANSACTION_DEADLINE = 5.0;
+#else
 static const NSTimeInterval SENTRY_AUTO_TRANSACTION_DEADLINE = 30.0;
+#endif
```
philipphofmann, line 48: `m`: Why do we need this?

philipphofmann (commented):
First pass; I don't understand the purpose of the middleware thingy. I think it would be best to discuss this over a call. There are also plenty of conflicts to resolve, so I didn't look at the tracer yet.

Sources/Sentry/SentryTracer.m
brustolin: Make sample tests while developing so much easier. Dont need to wait 30 seconds to see deadline timeout while debugging.

Samples/iOS-Swift/iOS-Swift/ViewControllers/TraceTestViewController.swift
```diff
@@ -22,13 +22,16 @@ class TraceTestViewController: UIViewController {
         }
         let session = URLSession(configuration: URLSessionConfiguration.default)
         let dataTask = session.dataTask(with: imgUrl) { (data, _, error) in
+            //Simulated delay in the download
+            Thread.sleep(forTimeInterval: 2)
```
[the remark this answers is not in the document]
brustolin: If you mean UI testing, we dont have a test for this screen. 
But  if you mean manual test, I guess we also need a way to test `reportFullyDisplayed`, and this would be the place.

Sources/Sentry/Public/SentrySDK.h
brustolin: This don't add a measurement.

Sources/Sentry/SentrySDK.m
brustolin: ctrl+c, ctrl+v 😁

Sources/Sentry/SentryTracer.m
brustolin: Yes. 
So, I was thinking about leaving this TODOs in the code as another way to track things that would be nice to have but are not related to the current PR, instead of opening issues.
Lets say we need to change appStartSpans behavior. By reading this TODO we could include this change in the same task. If we only have an open issue, we wont go back to the full board to see if there is something related to the same code being changed.

Me and @kahest talked about this, and he is onboard with the idea.

kahest: We should use a specific format to tag these refactoring opportunities. Ideally we would refer to a GH issue like this
`// TODO(ref): use middleware to create appStartSpans https://github.com/getsentry/sentry-cocoa/issues/1234`

brustolin: Nice, I like the idea.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Crash in AppHangs when no threads (#2725)
+- Time to initial display and time to full display (#2724)
```
brustolin, line 8:
```suggestion

### Features
- Time to initial display and time to full display (#2724)
```

Samples/iOS-Swift/iOS-Swift/ViewControllers/TraceTestViewController.swift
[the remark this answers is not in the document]
philipphofmann: I wonder why we need the thread sleep. We already load some data via an HTTP request.

Sources/Sentry/Public/SentrySDK.h
philipphofmann: As discussed yesterday in the daily we should add a measurement for it.

Sources/Sentry/SentryTracer.m
philipphofmann: I'm not a fan of having todos in the code. I prefer creating issues with references to the code.

Sources/Sentry/SentryTracer.m
philipphofmann: I would just change the value to 5.0 when you are debugging. This could cause quite some confusion if you are unaware of that.

brustolin: What kind of confusion?
We have debug logs to inform that the deadline expired. If one of us see something finishing early it will be clear why.

Sources/Sentry/SentryTracer.m
brustolin: These issues are forgotten in limbo.

Sources/Sentry/include/SentryTracerMiddleware.h
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ Called when the dead line timeout is triggered
+ */
+- (void)tracerDidTimeout:(SentryTracer *)tracer;
+
+/**
+ * Called when the `SentryTracer` is finished.
+ */
+- (void)tracerDidFinish:(SentryTracer *)tracer;
```
philipphofmann, line 37: `m`: Only the `SentryPerformanceTracker` uses this method. The `SentryTimeToDisplayTracker` implements all other methods except this one. Therefore, it makes sense to split this protocol up into multiple ones. I would combine `tracerDidTimeout` and `tracerDidFinish` into something like `TracerLifecycleEvents`.

Sources/Sentry/include/SentryTracerMiddleware.h
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Called when the middleware is attached to a SentryTracer
+ */
+- (void)installForTracer:(SentryTracer *)tracer;
+
+/**
+ * Called when the middleware is removed from a SentryTracer
+ */
+- (void)uninstallForTracer:(SentryTracer *)tracer;
```
philipphofmann, line 22: `m`: No class uses that. Let's remove it and add it once we need it.
```suggestion
```

Sources/Sentry/include/SentryTracerMiddleware.h
```diff
@@ -0,0 +1,41 @@
+#import "SentrySpan.h"
+#import "SentryTracer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Protocol used to intercept some of `SentryTracer` steps
+ * in order to modify its behavior and distribute concern to other classes.
+ */
+@protocol SentryTracerMiddleware <NSObject>
```
philipphofmann, line 10: `h`: I propose to rename this to `SentryTracerExtension`. It has three non optional methods `installForTracer`, `createAdditionalSpansForTrace` and `addMeasurements`.  This will allow us to also move app start measurements and the frame tracker measurements to extension classes.

As some tracer extensions could be interested in lifecycle events `SentryTracerExtension` should extend `TracerLifecycleEvents` which has only optional methods.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 - Crash in AppHangs when no threads (#2725)
 
+### Features
+- Time to initial display and time to full display (#2724)
```
philipphofmann, line 10: `l`: Even a bit shorter.
```suggestion
- Time to initial and full display (#2724)
```

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,121 @@
...
+        return @[ ttfd ];
+    }
+
+    return @[];
+}
+
+- (NSDate *)startDate
+{
+    return _startDate;
+}
```
philipphofmann, line 118: `m`: I think we don't need this.

Sources/Sentry/SentryPerformanceTracker.m
```diff
@@ -213,7 +215,12 @@ - (void)finishSpan:(SentrySpanId *)spanId withStatus:(SentrySpanStatus)status
     id<SentrySpan> spanTracker;
     @synchronized(self.spans) {
         spanTracker = self.spans[spanId];
-        [self.spans removeObjectForKey:spanId];
+        // Hold reference for tracer for longer because automatic
+        // tracer aren't  being referenced by anything else.
```
philipphofmann, line 219: `l`:
```suggestion
        // Hold reference for tracer until the tracer finishes because automatic
        // tracers aren't referenced by anything else.
```

Sources/Sentry/SentryPerformanceTracker.m
```diff
@@ -213,7 +215,12 @@ - (void)finishSpan:(SentrySpanId *)spanId withStatus:(SentrySpanStatus)status
...
+        // callback to `tracerDidFinish` will release it.
+        if (![spanTracker isKindOfClass:SentryTracer.self]) {
```
philipphofmann, line 221: `m`: Why do we need this change? What's the impact?

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -109,6 +120,55 @@ - (void)createTransaction:(UIViewController *)controller
     }
 }
 
+- (void)createTimeToDisplay:(UIViewController *)controller
+{
+    SentrySpan *vcSpan = [self viewControllerPerformanceSpan:controller];
+
+    if (![vcSpan isKindOfClass:[SentryTracer self]]) {
+        // Since TTID and TTFD are meant to the whole screen
+        // we will not track child view controllers
```
philipphofmann, line 129: 💯 for this vital comment.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -109,6 +120,55 @@ - (void)createTransaction:(UIViewController *)controller
...
+        return;
+    }
+
+    if ([vcSpan.tracer getMiddlewaresOfType:SentryTimeToDisplayTracker.self].count > 0) {
+        // Already tracking time to display.
```
philipphofmann, line 134: `l`: Please describe the situation that could lead to reaching this code path in the comment.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -109,6 +120,55 @@ - (void)createTransaction:(UIViewController *)controller
...
+    [ttdTracker registerInitialDisplay];
+}
+
+- (void)reportFullyDisplayed
+{
+    [self.currentTTDTracker registerFullDisplay];
+
+    // No need for it anymore
+    self.currentTTDTracker = nil;
+}
```
philipphofmann, line 170: `h`: If I've not mistaken, only the hub calls this method. Don't we also have to set `self.currentTTDTracker` to `nil` when the user doesn't manually call the API?

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,121 @@
...
+        [self.initialDisplaySpan finish];
+    }
+}
+
+- (void)tracerDidTimeout:(SentryTracer *)tracer
+{
+    // SentryTracer deadline timeout fired,
+    // initial display span needs to be finished.
+    if (!self.initialDisplaySpan.isFinished) {
+        [self.initialDisplaySpan finish];
```
philipphofmann, line 88: `m`:
```suggestion
        [self.initialDisplaySpan finishWithStatus:kSentrySpanStatusDeadlineExceeded];
```

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,121 @@
...
+
+- (void)installForTracer:(SentryTracer *)tracer
+{
+    self.initialDisplaySpan =
+        [tracer startChildWithOperation:SentrySpanOperationUILoadInitialDisplay
+                            description:[NSString stringWithFormat:@"%@ initial display",
+                                                  self.controllerName]];
+}
+
+- (void)registerInitialDisplay
```
philipphofmann, line 43: `m`: I'm a bit confused by the name `register`. Why not? Same applies to `registerFullDisplay`.
```suggestion
- (void)reportInitialDisplay
```

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,121 @@
...
+{
+    if (self.initialDisplaySpan.timestamp != nil) {
+        return;
+    }
+
+    self.initialDisplaySpan.timestamp = [SentryCurrentDate date];
+
+    if (self.fullDisplay != nil &&
+        [self.fullDisplay compare:self.initialDisplaySpan.timestamp] == NSOrderedAscending) {
+        self.fullDisplay = self.initialDisplaySpan.timestamp;
```
philipphofmann, line 53: `m`: I don't understand why we must set the `fullDisplay` here.

philipphofmann (commented):
Another pass, lets have a chat about my proposal for changing and renaming the protocols https://github.com/getsentry/sentry-cocoa/pull/2724#discussion_r1122883801 via a call.

Samples/iOS-Swift/iOS-Swift/ViewControllers/TraceTestViewController.swift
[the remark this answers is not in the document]
brustolin: The load was so fast, maybe cache involved, that I couldn't properly test time to full display

Sources/Sentry/SentryTimeToDisplayTracker.m
brustolin: Nope. The initialDisplaySpan is only open yet to keep SentryTracer waiting for fullDisplay. But it should not be finished with DeadlineExceeded, because its real finish timespan happened before.

Sources/Sentry/SentryTimeToDisplayTracker.m
brustolin: `reportFullyDisplayed` may be called before initial display happens.
It means the content is fully loaded, but the screen itself was not displayed yet. It will happen with initial display.

Sources/Sentry/include/SentryTimeToDisplayTracker.h
```diff
@@ -0,0 +1,43 @@
...
+ * enough to wait for a full display report if required, otherwise it finished
+ * the TTID span when a initial display is registered, allowing the tracer to
+ * finish as soon as possible.
+ * TTFD span is created when `SentryTracer` request for additional spans only if
+ * a full display was registered before.
+ */
+@interface SentryTimeToDisplayTracker : NSObject <SentryTracerMiddleware>
+SENTRY_NO_INIT
+
+@property (nonatomic, strong, readonly) NSDate *startDate;
```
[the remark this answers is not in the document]
brustolin: This is used for tests. 
Instead of having other header or having to use the `Dynamic` hack, is easier just to expose it, since the class itself is private.

Sources/Sentry/include/SentryTracerMiddleware.h
brustolin: But then we need multiples collections inside of SentryTracer, one for each type of protocol. Or, in my opinion a worst option, one non typed NSArray.

Sources/Sentry/include/SentryTracerMiddleware.h
brustolin: I like the name very much.
I don't like the idea of dividing it, IMO it does not add value and make things more complicated.
Also, I don't like the assumption that every extension will need additional spans or measurements, INP for example will only require measurements.

Sources/Sentry/SentryPerformanceTracker.m
brustolin: We need to keep a reference to SentryTracer otherwise ARC will release it.
As mentioned in the comment, tracerDidFinish will be responsable for releasing the Tracers from now on.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
brustolin: By calling loadView more than once. This is not supposed to happen but is possible.

This is just a safe guard.

Sources/Sentry/include/SentryTracerMiddleware.h
philipphofmann: We could also move this to `SentryTracerDelegate` as the `SentryPerformanceTracker` is the only delegate.

Sources/Sentry/include/SentryTracerMiddleware.h
philipphofmann: As discussed on a call, we will not create a `TracerLifecycleEvents` protocol. Instead, we do https://github.com/getsentry/sentry-cocoa/pull/2724/files#r1123098211.

We also remove the `@optional` so we don't have to check if a protocol implements a method in the tracer. We stick to the naming of `SentryTracerExtension`.

Sources/Sentry/include/SentryTracerMiddleware.h
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ Return additional spans to be added to the transaction.
+ */
+- (NSArray<id<SentrySpan>> *)createAdditionalSpansForTrace:(SentryTracer *)tracer;
+
+/**
+ Called when the dead line timeout is triggered
+ */
+- (void)tracerDidTimeout:(SentryTracer *)tracer;
```
philipphofmann, line 32: `m`: We can remove that.
```suggestion
- (void)tracerDidTimeout;
```

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
brustolin: This is not required anymore. `currentTTDTracker` is a weak reference and will take care of itself

Tests/SentryTests/SentryHubTests.swift
```diff
@@ -901,3 +930,18 @@ class SentryHubTests: XCTestCase {
         XCTAssertEqual(expected, span.sampled)
     }
 }
+
+#if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
+private class TimeToDisplayTrackerTest: SentryTimeToDisplayTracker {
```
armcknight, line 935:
```suggestion
private class TestTimeToDisplayTracker: SentryTimeToDisplayTracker {
```

Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
```diff
@@ -0,0 +1,168 @@
...
+
+    func testreportInitialDisplay_notWaitingFullDisplay() {
+        let sut = SentryTimeToDisplayTracker(for: UIViewController(), waitForFullDisplay: false)
+        let tracer = fixture.tracer
+
+        fixture.dateProvider.setDate(date: Date(timeIntervalSince1970: 7))
+        tracer.addExtension(sut)
+        XCTAssertEqual(tracer.children.count, 1)
+
+        let ttidSpan = tracer.children.first
```
armcknight, line 37: Using `XCTUnwrap` would specifically help us know if there wasn't a span like we were expecting, and avoid all the optional unwrapping/coalescing downstream:
```suggestion
        let ttidSpan = try XCTUnwrap(tracer.children.first, "Expected a TTID span")
```

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -109,6 +120,53 @@ - (void)createTransaction:(UIViewController *)controller
...
+    }
+
+    SentryTimeToDisplayTracker *ttdTracker =
+        [vcSpan.tracer getExtensionsOfType:SentryTimeToDisplayTracker.self].firstObject;
+    [ttdTracker reportInitialDisplay];
+}
+
+- (void)reportFullyDisplayed
+{
+    [self.currentTTDTracker reportFullDisplay];
```
armcknight, line 167: Please standardize the names. Either both should be `reportFullyDisplayed` or `reportFullDisplay`

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -109,6 +120,53 @@ - (void)createTransaction:(UIViewController *)controller
...
+    self.currentTTDTracker = ttdTracker;
+}
+
+- (void)finishTTIDForController:(UIViewController *)controller
+{
+    SentrySpan *vcSpan = [self viewControllerPerformanceSpan:controller];
+
+    if (![vcSpan isKindOfClass:[SentryTracer self]]) {
+        // Only tracer may have TimeToDisplayTracker
+        return;
```
armcknight, line 157: How does execution arrive at this spot? The fact that this kind of condition needs to be tested in the `if` statement is a code smell.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
armcknight: The comment is informative, but I have no idea how it is related to the class type of the span. We should not be wrapping up vital logic like that in the type system like this.

Sources/Sentry/include/SentryUIViewControllerPerformanceTracker.h
```diff
@@ -96,7 +101,10 @@ static NSString *const SENTRY_UI_PERFORMANCE_TRACKER_SPANS_IN_EXECUTION_SET
...
 - (void)viewControllerViewDidLayoutSubViews:(UIViewController *)controller
                            callbackToOrigin:(void (^)(void))callback;
-#endif
+
+- (void)reportFullyDisplayed;
+
 @end
 
 NS_ASSUME_NONNULL_END
+#endif
```
armcknight, line 110:
```suggestion
#endif // SENTRY_HAS_UIKIT
```

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,109 @@
...
+#import "SentrySwift.h"
+
+#if SENTRY_HAS_UIKIT
+
+@interface
+SentryTimeToDisplayTracker ()
+
+@property (nonatomic, strong) NSDate *startDate;
+@property (nonatomic, strong) SentrySpan *initialDisplaySpan;
+@property (nonatomic, strong) NSDate *fullDisplay;
```
armcknight, line 16: It's not obvious at callsites what this was:
```suggestion
@property (nonatomic, strong) NSDate *fullDisplayDate;
```

Sources/Sentry/include/SentryTimeToDisplayTracker.h
```diff
@@ -0,0 +1,43 @@
...
+ * finish as soon as possible.
+ * TTFD span is created when `SentryTracer` request for additional spans only if
+ * a full display was registered before.
+ */
+@interface SentryTimeToDisplayTracker : NSObject <SentryTracerExtension>
+SENTRY_NO_INIT
+
+@property (nonatomic, strong, readonly) NSDate *startDate;
+
+@property (nonatomic, readonly) BOOL waitFullDisplay;
```
armcknight, line 28: I'm going to suggest a couple changes to names like this to standardize:
```suggestion
@property (nonatomic, readonly) BOOL waitForFullDisplay;
```

Sources/Sentry/include/SentryTimeToDisplayTracker.h
```diff
@@ -0,0 +1,43 @@
...
+
+- (instancetype)initForController:(UIViewController *)controller
+               waitForFullDisplay:(BOOL)waitFullDisplay;
```
armcknight, line 31:
```suggestion
               waitForFullDisplay:(BOOL)waitForFullDisplay;
```

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,109 @@
...
+@property (nonatomic, strong) NSString *controllerName;
+@property (nonatomic) BOOL waitFullDisplay;
```
armcknight, line 18:
```suggestion
@property (nonatomic) BOOL waitForFullDisplay;
```

Sources/Sentry/include/SentryTimeToDisplayTracker.h
```diff
@@ -0,0 +1,43 @@
...
+
+- (void)reportInitialDisplay;
+
+- (void)reportFullDisplay;
+
+- (void)stopWaitingFullDisplay;
```
armcknight, line 37:
```suggestion
- (void)stopWaitingForFullDisplay;
```

Sources/Sentry/SentryTracer.m
```diff
@@ -212,8 +225,15 @@ - (instancetype)initWithTransactionContext:(SentryTransactionContext *)transacti
     return self;
 }
 
+- (nullable SentryTracer *)tracer
+{
+    return self;
+}
```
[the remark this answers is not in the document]
armcknight: Oh, after thinking looking at it again, this is because `SentryTracer` is a subclass of `SentrySpan`, right? Can you explain when `-[SentryTracer tracer]` would be called? This is a code smell.

armcknight (commented):
There's a lot of good stuff in here, nice work fleshing out the implementation. I left a bunch of suggestions you can hopefully take up easily to align some names to improve searchability for later maintainers.

I do have some questions around how types are being used. And I don't think we should move forward with the tracer extension concept. That should wait on discussion in the related issue and be saved for a later PR.

Sources/Sentry/SentryTracer.m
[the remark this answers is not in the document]
brustolin: As you mentioned, `SentryTracer` is a subclass of `SentrySpan`. 
Some places in the project contains a collection of SentrySpan that need to get the root of the transaction. All it needs to do is calling span.tracer

Sources/Sentry/SentryTracer.m
```diff
@@ -816,6 +860,74 @@ + (nullable SentryTracer *)getTracer:(id<SentrySpan>)span
     return nil;
 }
 
+#pragma mark - Extensions
```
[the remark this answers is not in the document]
brustolin: I disagree. This IMO is the simplest solution. 
If we dont use SentryTracerExtension we would need to distribute more responsibility to SentryTracer and SentryUIViewControllerPerformanceTracker.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
brustolin: A controller may be a child of another controller, therefore instead of having a SentryTracer it will have a SentrySpan.

Sources/Sentry/SentryUIViewControllerSwizzling.m
```diff
@@ -108,6 +108,7 @@ - (void)start
     }
 
     [self swizzleUIViewController];
+    SentryUIViewControllerPerformanceTracker.shared.inAppLogic = self.inAppLogic;
```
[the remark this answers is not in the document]
brustolin: Yes I agree. This also annoys me. 
I think this is old stuff that we probably may change, but not related to this PR.

Sources/Sentry/SentryTracer.m
```diff
@@ -200,6 +211,8 @@ - (instancetype)initWithTransactionContext:(SentryTransactionContext *)transacti
 #endif // SENTRY_HAS_UIKIT
 
 #if SENTRY_TARGET_PROFILING_SUPPORTED
+        // TODO(ref): Move Profiling to a SentryTracerExtension.
+        // https://github.com/getsentry/sentry-cocoa/issues/2736
```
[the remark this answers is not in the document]
armcknight: You can highlight lines of code in your diff and copy the URL from your browser address bar, then put them in a bullet point list in the description of that issue or in a separate comment. Eg https://github.com/getsentry/sentry-cocoa/pull/2724/files#diff-84fda3cf94bf21dd7ceabf3c0ced60d74c124620d5a8dfdd3b8661563a79ec49R216

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
armcknight: Thanks for the explanation, would you mind adding it to the comment here to explain the relationship between spans/tracer and parent/child VCs?

AFAIUI, we start a span for every VC in a hierarchy, then just pick the one for the top level parent to actually use? If that is correct, what happens to the rest of the spans?

The code smell here is the fact that we have to care about the receiving type at all in the implementation. If SentryTracer were not a subclass of SentrySpan, there would be no need for checks, casts, or calls to `-[tracer]`. 

The TTD span should simply be attached to the correct view controller, decided by whatever is managing TTD tracking, which would look at the VC parent/child relationships at the start, so there would only ever be one span.

Sources/Sentry/SentryUIViewControllerSwizzling.m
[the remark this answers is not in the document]
armcknight: https://github.com/getsentry/sentry-cocoa/issues/2768

Sources/Sentry/SentryTracer.m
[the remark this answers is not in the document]
armcknight: What other solutions did you consider such that you arrived at this one being the simplest? The only class that will conform to this protocol so far is the TTD tracker. What other classes do you plan to conform to it? 

Couldn't SentryTracer directly instantiate those classes, store the references as properties, and interact directly with them, without the extra machinery? Then it would be a simple composition hierarchy. There is some complexity involved with the type-erasure (`id<SentryTracerExtension>`), storing them in an array that has to be synchronized and copied, etc.

Sources/Sentry/SentryTracer.m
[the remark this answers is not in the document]
brustolin: Got it. My problem with this idea, as mentioned in [this](https://github.com/getsentry/sentry-cocoa/pull/2724#discussion_r1119863773) thread, is that those issues are never tackled. 

If we have a reminder in the code, when someone starts to work on a new feature that needs to change one of this TODOs places, they could also include this change. 

Lets say we have another feature that also requires `Finish callback` from SentryTracer. Instead of making use of the current property, or even worse, create another one, since it only accepts one callback, we would see the TODO and remember this could be done with a SentryTracerExtension.

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,109 @@
...
+    if (self.fullDisplayDate != nil &&
+        [self.fullDisplayDate compare:self.initialDisplaySpan.timestamp] == NSOrderedAscending) {
+        self.fullDisplayDate = self.initialDisplaySpan.timestamp;
+    }
+
+    if (!self.waitForFullDisplay || self.fullDisplayDate != nil) {
+        // If this class is waiting for a full display, we don't finish the TTID span
+        // because this will make the tracer wait for its children which gives the user more time to
+        // report full display, since SentryTracer have a dead line timeout, eventually it will be
+        // finished.
```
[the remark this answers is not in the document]
stefanosiano: I get it, but if the user enables the options he wants to track the full screens.
Having a span with deadline_exceeded is a way to remind him he didn't call the API.
We can make any other consideration based on user's feedback

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,109 @@
+#import "SentryTimeToDisplayTracker.h"
+#import "SentryCurrentDate.h"
+#import "SentrySpan.h"
+#import "SentrySpanContext.h"
+#import "SentrySpanId.h"
+#import "SentrySpanOperations.h"
+#import "SentrySwift.h"
+
+#if SENTRY_HAS_UIKIT
```
kahest, line 9: `l`: can be moved above imports because all @interface and @implementation is guarded by this

Sources/Sentry/SentryTimeToDisplayTracker.m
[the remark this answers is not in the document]
armcknight: I agree with @brustolin on this one. We should print a warning log and maybe set some flag on the session so that we can indicate it on their dashboard in some other way.

Sources/Sentry/SentryTracer.m
armcknight: It could just be `DEBUG` for now though, since it's only defined in the Debug build config along with `DEBUG`. If we only treated warnings as errors for nondebug builds, I'd say we could add `#warning "Using shorter auto transaction deadlines"` here as well.

I agree this would be confusing during unit testing. If you're only using it for testing with a sample app, we could instead add a way to configure it in the sample app's UI.

I wondered if it's something we could open up for config in SentryOptions but that probably has backend/product implications and I assume that'll be a nonstarter.

Sources/Sentry/SentryTracer.m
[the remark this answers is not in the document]
armcknight: Ah, sorry I missed that thread. I don't feel terribly strongly about this, I'm guilty of it myself.

Sources/Sentry/SentryTimeToDisplayTracker.m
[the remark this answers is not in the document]
brustolin: > but if the user enables the options he wants to track the full screens

I have a problem it this assertion. 
Lets say a big project with dozens of screens, but the user only want TTFD for main big screens, now they need to go at every view controller and report full display. 
Otherwise every other transaction will have trash span.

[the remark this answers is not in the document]
armcknight: > they need to go at every view controller and report full display
Isn't this how it's going to work anyways, since they must call `SentrySDK.reportFullyDiplayed` from each of those?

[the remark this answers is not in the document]
brustolin: This is what we are expecting them to do. Particularly I find this annoying. But we agreed to keep the implementation similar to Android.

Sources/Sentry/SentryTracer.m
brustolin: If we use DEBUG it will also run like this in customers projects

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
```diff
@@ -31,9 +31,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             options.attachScreenshot = true
             options.attachViewHierarchy = true
             options.environment = "test-app"
+            options.enableTimeToFullDisplay = false
```
philipphofmann, line 34: `m`: Shouldn't we enable this?

Sources/Sentry/SentryTracer.m
philipphofmann: > If we use DEBUG it will also run like this in customers projects

That's another reason for me to remove this. It confuses customers.

philipphofmann (commented):
Sorry I didn't get to give this a full review yet. I tried to enable `enableTimeToFullDisplay` in the iOS-Swift sample app and navigated to the `TraceTestViewController`, but the transaction reported to Sentry didn't contain a time to full display span.

Sources/Sentry/Public/SentrySDK.h
```diff
@@ -301,6 +301,13 @@ SENTRY_NO_INIT
  */
 + (void)crash;
 
+/**
+ * Reports to the ongoing UIViewController transaction
+ * that the screen contents are fully loaded and displayed,
+ * which will create a new span.
```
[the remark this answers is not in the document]
brustolin: I was planning to do measurement in another PR.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
brustolin: Now its me that is not understanding your explanation 😁, but it does not really matter any more because the approach changed to a simpler one.

Sources/Sentry/Public/SentrySDK.h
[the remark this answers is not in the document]
philipphofmann: Then please point this out in the PR description and create an extra issue for it so we don't forget about it and I don't have to ask the question here 😄

philipphofmann:
I think this feature is not working properly. I changed the `LoremIpsumViewController.viewDidLoad` to the following

```swift
override func viewDidLoad() {
    super.viewDidLoad()
    
    let transaction = SentrySDK.span
    
    let viewDidLoad = transaction?.startChild(operation: "manual.viewDidLoad")
    
    let dispatchQueue = DispatchQueue(label: "LoremIpsumViewController")
    dispatchQueue.async {
        let readFile = viewDidLoad?.startChild(operation: "manual.read_file")
        if let path = Bundle.main.path(forResource: "LoremIpsum", ofType: "txt") {
            if let contents = FileManager.default.contents(atPath: path) {
                readFile?.finish()
                
                let updateView = viewDidLoad?.startChild(operation: "manual.update_view")
                DispatchQueue.main.async {
                    self.textView.text = String(data: contents, encoding: .utf8)
                    SentrySDK.reportFullyDisplayed()
                    updateView?.finish()
                    viewDidLoad?.finish()
                }
            }
        }
    }
}
```
This is what I see in [Sentry](https://sentry-sdks.sentry.io/discover/sentry-cocoa:06fe92c6cdf2486cadaad6e621d670b4/?field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&homepage=true&id=18214&name=&project=5428557&query=event.type%3Atransaction&sort=-timestamp&statsPeriod=14d&topEvents=5&yAxis=count%28%29) 
Screenshot 2023-03-23 at 09 46 29

There are a couple of problems:

1. The full_display span is shorter than the initial_display span, although they both have the same end timestamp `1679560935.777921`. Their start timestamp is different: full_dispaly has `1679560935.226202` and initial_display has `1679560935.225970`
2. The start timestamps of both full and initial display are not the same as the one from the transaction / root span `1679560935.224946`.
3. The code calls `SentrySDK.reportFullyDisplayed()` one line before calling `updateView`, but the `update_view` span ends way sooner than the full_display span. I guess this could cause confusion for users. I set a breakpoint and the UIViewController is already visible when I call  `SentrySDK.reportFullyDisplayed()`.

Screenshot 2023-03-23 at 09 55 59

And another problem, now for every UIViewController for which we don't call `SentrySDK.reportFullyDisplayed()` now we get a `full_display` span that times out, lasting 5 seconds. In production, it is going to be 30 seconds. That will completely mess up the transactions. Instead, when the `full_display` span times out we could set the end timestamp to the same as the `initial_display` set the status to `deadline exceeded` and also change the description to deadline exceeded to highlight, that the user forgot to call `SentrySDK.reportFullyDisplayed()`. What do you think about that @stefanosiano? If we agree we must also update the [develop docs](https://github.com/getsentry/develop/pull/850.)

brustolin:
Thanks @philipphofmann for point this time drifting, I don't really know whats is causing this, since this is the code that creates the spans:

```objc
 self.initialDisplaySpan =
        [tracer startChildWithOperation:SentrySpanOperationUILoadInitialDisplay
                            description:[NSString stringWithFormat:@"%@ initial display",
                                                  self.controllerName]];

    if (self.waitForFullDisplay) {
        self.fullDisplaySpan =
            [tracer startChildWithOperation:SentrySpanOperationUILoadFullDisplay
                                description:[NSString stringWithFormat:@"%@ full display",
                                                      self.controllerName]];
    }
```
I have no idea how is possible for time to full span to start before initialDisplay. But anyway i have a solution for this, will commit later.

Your assumption for number 3  is wrong. Time to full display cannot happen before time to initial display, which happens after view did appear, if you called `reportFullDisplay` before `viewDidAppear` both spans will have the same timestamp. This will be explained in our docs once we make then available for cocoa as well, the same way it is for [android](https://docs.sentry.io/platforms/android/performance/instrumentation/automatic-instrumentation/#time-to-full-display).

philipphofmann (commented):
Another pass; will do a full pass on Monday.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
```diff
@@ -0,0 +1,218 @@
...
+        super.setUp()
+        CurrentDate.setCurrentDateProvider(fixture.dateProvider)
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        clearTestState()
+    }
+
+    func testreportInitialDisplay_notWaitingFullDisplay() throws {
```
philipphofmann, line 35: `l`: Please also update the other test cases.
```suggestion
    func testReportInitialDisplay_notWaitingFullDisplay() throws {
```

Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
```diff
@@ -0,0 +1,218 @@
...
+
+        fixture.dateProvider.setDate(date: Date(timeIntervalSince1970: 11))
+        sut.reportFullyDisplayed()
+
+        XCTAssertEqual(ttidSpan?.timestamp, Date(timeIntervalSince1970: 9))
+        XCTAssertTrue(ttidSpan?.isFinished ?? false)
+        XCTAssertEqual(tracer.children.count, 2)
+    }
+
+    func testreportFullDisplay_noWaitingForFullDisplay() {
```
philipphofmann, line 108: `l`:
```suggestion
    func testreportFullDisplay_notWaitingForFullDisplay() {
```

Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
```diff
@@ -0,0 +1,218 @@
...
+        let tracer = fixture.tracer
+
+        sut.start(for: tracer)
+
+        XCTAssertNotNil(sut.fullDisplaySpan)
+        XCTAssertEqual(sut.fullDisplaySpan?.startTimestamp, tracer.startTimestamp)
+        XCTAssertEqual(sut.initialDisplaySpan?.startTimestamp, tracer.startTimestamp)
+    }
+
+    func test_fullDisplay_reportedBefore_initialDisplay() {
```
philipphofmann, line 198: `l`: It would be great to stick to one naming pattern in the same test file.
```suggestion
    func testFullDisplay_reportedBefore_initialDisplay() {
```

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,127 @@
...
+
+- (void)reportFullyDisplayed
+{
+    _fullyDisplayedReported = YES;
+    if (_isReadyToDisplay) {
+        [self.fullDisplaySpan finish];
+    }
+}
+
+- (void)framesTrackerHasNewFrame
```
philipphofmann, line 104: `m`: A comment explaining why we wait until the next frame is drawn will help us to understand the code in the future.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
```diff
@@ -0,0 +1,218 @@
...
+        sut.reportReadyToDisplay()
+        fixture.displayLinkWrapper.normalFrame()
+
+        XCTAssertEqual(ttidSpan.timestamp, fixture.dateProvider.date())
+        XCTAssertTrue(ttidSpan.isFinished)
+        XCTAssertEqual(ttidSpan.spanDescription, "UIViewController initial display")
+        XCTAssertEqual(ttidSpan.operation, SentrySpanOperationUILoadInitialDisplay)
+    }
+
+    func testReportNewFrame_notReadyToDisplay() throws {
```
philipphofmann, line 56: Good edge case testing 👏

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,127 @@
...
+
+- (void)reportReadyToDisplay
+{
+    _isReadyToDisplay = YES;
+}
+
+- (void)reportFullyDisplayed
+{
+    _fullyDisplayedReported = YES;
+    if (_isReadyToDisplay) {
```
philipphofmann, line 99: `l`: It doesn't make much a difference now, but we don't need to call `[self.fullDisplaySpan finish]` below when `self.fullDisplaySpan` is `nil`.
```suggestion
    if (self.waitForFullDisplay && _isReadyToDisplay) {
```

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,127 @@
...
+        [self.fullDisplaySpan finish];
+    }
+}
+
+- (void)framesTrackerHasNewFrame
+{
+    if (_fullyDisplayedReported && self.fullDisplaySpan.isFinished == NO) {
+        [self.fullDisplaySpan finish];
```
philipphofmann, line 107: `m`: `test_fullDisplay_reportedBefore_initialDisplay` should fail if I comment out this line.

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,127 @@
...
+    if (_isReadyToDisplay && self.initialDisplaySpan.isFinished == NO) {
+        [self reportInitialDisplay];
+        [SentryFramesTracker.sharedInstance removeListener:self];
+    }
+}
+
+- (void)trimTTFDIdNecessaryForTracer:(SentryTracer *)tracer
+{
+    if (self.fullDisplaySpan.status != kSentrySpanStatusDeadlineExceeded) {
+        return;
```
philipphofmann, line 118: `m`: I miss a test case for this line of code, if I'm not mistaken.

Sources/Sentry/SentryTracer.m
```diff
@@ -166,6 +170,7 @@ - (instancetype)initWithTransactionContext:(SentryTransactionContext *)transacti
         self.wasFinishCalled = NO;
         _waitForChildren = waitForChildren;
         _measurements = [[NSMutableDictionary alloc] init];
+        _extensions = [NSMutableArray array];
```
philipphofmann, line 173: `m`:
```suggestion
```

Sources/Sentry/include/SentryTracer.h
```diff
@@ -50,6 +57,8 @@ static NSTimeInterval const SentryTracerDefaultTimeout = 3.0;
 
 @property (nonatomic, readonly) NSDictionary<NSString *, SentryMeasurementValue *> *measurements;
 
+@property (nonatomic, readonly) NSArray<id<SentryTracerExtension>> *extensions;
```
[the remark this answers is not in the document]
philipphofmann: `m`: I still see `SentryTracerExtension` here.

Sources/Sentry/include/SentryTracer.h
```diff
@@ -18,6 +20,11 @@ static NSTimeInterval const SentryTracerDefaultTimeout = 3.0;
  */
 - (nullable id<SentrySpan>)activeSpanForTracer:(SentryTracer *)tracer;
 
+/**
+ * Report that the tracer has finished.
+ */
+- (void)tracerDidFinish:(SentryTracer *)tracer;
```
philipphofmann, line 26: `m`: We already have `SentryTracer.finishCallback`. I would like to stick to one concept if possible.

Sources/Sentry/SentryTracer.m
```diff
@@ -87,6 +87,8 @@ @implementation SentryTracer {
 static NSObject *appStartMeasurementLock;
 static BOOL appStartMeasurementRead;
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshadow"
```
philipphofmann, line 91: `m`: Why do we have to add this while there are no changes in between?
```
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wshadow"
```

and
```
#pragma clang diagnostic pop
```

philipphofmann (commented):
Another pass, I still have a couple of files to go, but I will do that tomorrow. This is turning into something great 👍 😃

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,127 @@
...
+        SentryTracer *_tracer) { [self trimTTFDIdNecessaryForTracer:_tracer]; }];
+}
+
+- (void)reportInitialDisplay
+{
+    if (self.initialDisplaySpan.timestamp != nil) {
+        return;
+    }
+
+    self.initialDisplaySpan.timestamp = [SentryCurrentDate date];
```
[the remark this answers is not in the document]
brustolin: Remains of extension era. Will fix.

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,127 @@
...
+
+    if (self.fullDisplaySpan.timestamp != nil &&
```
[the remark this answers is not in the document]
brustolin: We dont need this code anymore

Sources/Sentry/include/SentryTracer.h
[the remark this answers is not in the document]
brustolin: Something left behind

Sources/Sentry/include/SentryTimeToDisplayTracker.h
```diff
@@ -0,0 +1,48 @@
...
+#    import <UIKit/UIKit.h>
+
+@class SentrySpan, SentryTracer, SentryFramesTracker;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * This is a SentryTracer extension responsible for creating
+ * TTID and TTFD spans.
+ * This extension creates the TTID span during installation and make use of
```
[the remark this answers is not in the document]
brustolin: Ill rewrite the comment.

Sources/Sentry/SentryTracer.m
brustolin: Probably some test left behind.

Sources/Sentry/include/SentryTimeToDisplayTracker.h
```diff
@@ -0,0 +1,48 @@
...
+- (instancetype)initForController:(UIViewController *)controller
+               waitForFullDisplay:(BOOL)waitForFullDisplay;
+
+- (instancetype)initForController:(UIViewController *)controller
+                     frameTracker:(SentryFramesTracker *)frametracker
+               waitForFullDisplay:(BOOL)waitForFullDisplay;
+
+- (void)startForTracer:(SentryTracer *)tracer;
+
+- (void)reportReadyToDisplay;
```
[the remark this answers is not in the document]
brustolin: Sorry, I dont like this. `TracerTimeToDisplayTracker` has no concept of the UIViewController life cycle. 
The same way we did before, the place we call this may change in the future.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
```diff
@@ -0,0 +1,218 @@
...
+    func testreportInitialDisplay_notWaitingFullDisplay() throws {
+        let sut = fixture.getSut(for: UIViewController(), waitForFullDisplay: false)
+        let tracer = fixture.tracer
+
+        fixture.dateProvider.setDate(date: Date(timeIntervalSince1970: 7))
+        sut.start(for: tracer)
+        XCTAssertEqual(tracer.children.count, 1)
+
+        let ttidSpan = try XCTUnwrap(tracer.children.first, "Expected a TTID span")
+        XCTAssertEqual(ttidSpan.startTimestamp, fixture.dateProvider.date())
```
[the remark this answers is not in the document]
brustolin: Yes we could. 
But there is no point in changing this now.

Sources/Sentry/include/SentryTracer.h
brustolin: This was a extension feature that was moved to `SentryTracerDelegate` because `SentryPerformanceTracker` needs to know when the tracer finishes but `SentryTimeToDisplay` also needs to know when SentryTracer finishes and is using `SentryTracer.finishCallback`.

This is why I prefer to implement listeners as collections, no need to refactor code unrelated to the feature later.

Now I rather leave this as it is and refactor in another PR.

Sources/Sentry/SentryFramesTracker.m
```diff
@@ -31,6 +31,7 @@
 @property (nonatomic, strong, readonly) SentryDisplayLinkWrapper *displayLinkWrapper;
 @property (nonatomic, assign) CFTimeInterval previousFrameTimestamp;
 @property (nonatomic) uint64_t previousFrameSystemTimestamp;
+@property (nonatomic, strong) NSHashTable<id<SentryFramesTrackerListener>> *listeners;
```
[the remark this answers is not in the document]
brustolin: Is almost never a good idea to create listeners as single properties.

Imo, this idea that "we change it when we need" means that we gonna need to refactor code and tests not related to the current task, making the PR bigger to review and some previous work being wasted.

Sources/Sentry/SentryTimeToDisplayTracker.m
brustolin: `testreportFullDisplay_waitingForFullDisplay` needs this line now.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
```diff
@@ -0,0 +1,229 @@
+import Foundation
+import Sentry
+import SentryTestUtils
+import XCTest
+
+#if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
+
+class SentryTimeToDisplayTrackerTest: XCTestCase {
```
philipphofmann, line 8: Those are solid tests. Great work 👏

Sources/Sentry/include/SentryTracer.h
philipphofmann: Yeah, we can do it in a follow up PR.

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -0,0 +1,107 @@
...
+#import "SentrySpanOperations.h"
+#import "SentrySwift.h"
+#import "SentryTracer.h"
+
+#if SENTRY_HAS_UIKIT
+
+@interface
+SentryTimeToDisplayTracker () <SentryFramesTrackerListener>
+
+@property (nonatomic, strong) SentrySpan *initialDisplaySpan;
```
philipphofmann, line 16: `h`: As the SentryTracer has a strong reference to this class with the `setFinishCallback`, I think we should use `weak` here to avoid a possible retain cycle. The tracer stays in memory with a strong reference in the `SentryPerformanceTracker` anyways. I don't think we need another `strong` reference here if I'm not mistaken.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -109,6 +121,39 @@ - (void)createTransaction:(UIViewController *)controller
...
 }
 
+- (void)createTimeToDisplay:(UIViewController *)controller
+{
+    SentrySpan *vcSpan = [self viewControllerPerformanceSpan:controller];
+
+    if (![vcSpan isKindOfClass:[SentryTracer self]]) {
+        // Since TTID and TTFD are meant to the whole screen
+        // we will not track child view controllers
+        return;
```
philipphofmann, line 131: `m`: If I'm not mistaken, we miss a test case for this edge case.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -109,6 +121,39 @@ - (void)createTransaction:(UIViewController *)controller
...
+    objc_setAssociatedObject(controller, &SENTRY_UI_PERFORMANCE_TRACKER_TTD_TRACKER, ttdTracker,
+        OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+    [ttdTracker startForTracer:(SentryTracer *)vcSpan];
+
+    self.currentTTDTracker = ttdTracker;
+}
+
+- (void)reportFullyDisplayed
+{
+    [self.currentTTDTracker reportFullyDisplayed];
```
philipphofmann, line 154: `m`: If I'm not mistaken, we miss a test for this piece of logic.

Sources/Sentry/SentrySDK.m
```diff
@@ -383,6 +383,11 @@ + (void)installIntegrations
     }
 }
 
++ (void)reportFullyDisplayed
+{
+    [SentrySDK.currentHub reportFullyDisplayed];
```
philipphofmann, line 388: `m`: A test for this would be nice.

philipphofmann (commented):
Almost done. It could be that we have a retain cycle, but it could also be that I'm wrong.

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -523,6 +523,22 @@ class SentrySDKTests: XCTestCase {
         XCTAssertFalse(stateAfterStop!.isSDKRunning)
     }
 #endif
+
+#if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
+    func testReportFullyDisplayed() {
```
philipphofmann, line 528: 💯

philipphofmann (approved):
LGTM, ship it 😃 🚀

PR #2545, opened 2022-12-20
ref: slice profiles per transaction

armcknight:
## :scroll: Description

#skip-changelog

This PR changes the interface between SentryTracer and SentryProfiler to always grab pertinent profiling data for each transaction and attach it as an additional envelope item (this was how the original implementation worked as well; the difference now is that it needs to work for concurrent transactions).

This will remove the span/transaction bookkeeping from SentryProfiler and move whatever is needed into SentryTracer, which will now be in full control of when to start and stop the profiler, and request data from it.

### TODO

- [x] Fix remaining broken unit tests
    - [x] Fix the concurrent transaction test
    - [x] Fix tests exercising timeout logic: need to callback to the tracer and prepare the profile payload with transaction info provided by the tracer and hold it there for when the transaction finally finishes, so the profile can be attached as an envelope item to the transaction
- [x] Write unit test for trimming out frame render timestamps outside of transaction/profile bounds, in addition to the other profile payload info that gets trimmed

## :bulb: Motivation and Context

Our current relationship of sending a separate profile envelope with many references to transactions resulted in a system that cannot know which profiles to keep or sample out with dynamic sampling. We need to be able to send associated profile data with each transaction envelope so that dynamic sampling will correctly keep and sample out profile data for transactions that it is already sampling on. This change allows SentryTracer to query SentryProfiler for the relevant profile data for each transaction it uploads.

## :green_heart: How did you test it?

Augment the concurrent transaction profile test to see if the correct samples and metrics are present in each resulting envelope. See `testConcurrentProfilingTransactions`.

## :pencil: Checklist

- [x] I reviewed the submitted code
- [x] I added tests to verify the changes
~- [ ] I updated the docs if needed~
~- [ ] Review from the native team if needed~
- [x] No breaking changes

PR #2729, opened 2023-02-28
fix: normalize timestamps relative to transaction start

armcknight:
## :scroll: Description

- **Normalizing profile payload timestamps to transaction start instead of profile start:** We formerly kept a start timestamp when the profiler began, and then reported each sample and metric reading as a time relative to that. Now we will record every timestamp using its absolute system time, and when we slice/serialize profiling data for a given transaction, calculate the relative offsets to the transaction start.

- **Using NSObjects to store samples/metrics instead of dictionaries during profiling:** Things got messy with the dictionaries we were using to store profile samples and metric readings, because we used to directly inject the computed relative timestamp into a dictionary entry called `elapsed_since_start_ns` at the time we recorded the metric/sample. Because we now need to record everything with an absolute timestamp, we would've had to use a new key to accurately describe that, then remove it during serialization after computing and storing the relative time under the old key. And, we would've had to change the type of the dictionary value as well, since we send these timestamps as strings, but to save the numeric value for computation at serialization time, I would've had to change all the dictionary subtypes in method signatures, variables etc. Instead of all the mucking about with that, I introduced some NSObjects with properties to hold the timestamps, values etc, and keep arrays of those, and only at serialization do I construct the JSON dictionaries. It's much clearer than nested dictionary subtypes, and will be easier to optimize for performance later if we want to convert to something like C++ vectors of structs.

#skip-changelog

In order for us to easily show transaction/span data in the profiling flamechart visualization, and since we now again have a 1:1 relationship between profile and transaction payloads (per https://github.com/getsentry/sentry-cocoa/pull/2545), we want to change all profiling timestamps to be relative to the start of the transaction.

See https://github.com/getsentry/team-profiling/issues/183

Added a couple simple checks to make sure any `elapsed_since_start_ns` timestamp falls within the associated transaction's start and end times.

Also added checks to make sure we always send up timestamps as strings (see https://github.com/getsentry/sentry-cocoa/pull/2729#discussion_r1119548329)

## :pencil: Checklist

~- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.~

### DIFF
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -84,10 +84,6 @@ processBacktrace(const Backtrace &backtrace,
     NSMutableDictionary<NSString *, NSNumber *> *frameIndexLookup,
     NSMutableDictionary<NSString *, NSNumber *> *stackIndexLookup)
 {
-    const auto sample = [[SentrySample alloc] init];
-    sample.absoluteTimestamp = backtrace.absoluteTimestamp;
-    sample.threadID = backtrace.threadMetadata.threadID;
-
     const auto threadID = [@(backtrace.threadMetadata.threadID) stringValue];
 
     NSString *queueAddress = nil;
@@ -137,6 +133,9 @@ processBacktrace(const Backtrace &backtrace,
         }
     }
 
+    const auto sample = [[SentrySample alloc] init];
+    sample.absoluteTimestamp = backtrace.absoluteTimestamp;
+    sample.threadID = backtrace.threadMetadata.threadID;
     if (queueAddress != nil) {
         sample.queueAddress = queueAddress;
     }
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -79,13 +79,12 @@ void
 processBacktrace(const Backtrace &backtrace,
     NSMutableDictionary<NSString *, NSMutableDictionary *> *threadMetadata,
     NSMutableDictionary<NSString *, NSDictionary *> *queueMetadata,
-    NSMutableArray<SentrySampleEntry *> *samples,
-    NSMutableArray<NSMutableArray<NSNumber *> *> *stacks,
+    NSMutableArray<SentrySample *> *samples, NSMutableArray<NSMutableArray<NSNumber *> *> *stacks,
     NSMutableArray<NSDictionary<NSString *, id> *> *frames,
     NSMutableDictionary<NSString *, NSNumber *> *frameIndexLookup,
     NSMutableDictionary<NSString *, NSNumber *> *stackIndexLookup)
 {
-    const auto sample = [[SentrySampleEntry alloc] init];
+    const auto sample = [[SentrySample alloc] init];
     sample.absoluteTimestamp = backtrace.absoluteTimestamp;
     sample.threadID = backtrace.threadMetadata.threadID;
 
@@ -259,11 +258,11 @@ processFrameRates(SentryFrameInfoTimeSeries *frameRates, SentryTransaction *tran
  * their data, with timestamps normalized relative to the provided transaction's start time. */
 NSArray<NSDictionary *> *
 serializedSamplesWithRelativeTimestamps(
-    NSArray<SentrySampleEntry *> *samples, SentryTransaction *transaction)
+    NSArray<SentrySample *> *samples, SentryTransaction *transaction)
 {
     const auto result = [NSMutableArray<NSDictionary *> array];
     [samples enumerateObjectsUsingBlock:^(
-        SentrySampleEntry *_Nonnull sample, NSUInteger idx, BOOL *_Nonnull stop) {
+        SentrySample *_Nonnull sample, NSUInteger idx, BOOL *_Nonnull stop) {
         // This shouldn't happen as we would've filtered out any such samples, but we should still
         // guard against it before calling getDurationNs as a defensive measure
         if (!orderedChronologically(transaction.startSystemTime, sample.absoluteTimestamp)) {
@@ -384,7 +383,7 @@ serializedSamplesWithRelativeTimestamps(
 
     const auto payload = [NSMutableDictionary<NSString *, id> dictionary];
 
-    NSArray<SentrySampleEntry *> *samples = _gCurrentProfiler->_profileData[@"profile"][@"samples"];
+    NSArray<SentrySample *> *samples = _gCurrentProfiler->_profileData[@"profile"][@"samples"];
 
     // We need at least two samples to be able to draw a stack frame for any given function: one
     // sample for the start of the frame and another for the end. Otherwise we would only have a
@@ -488,15 +487,15 @@ serializedSamplesWithRelativeTimestamps(
 
 #    pragma mark - Private
 
-+ (nullable NSArray<SentrySampleEntry *> *)slicedSamples:(NSArray<SentrySampleEntry *> *)samples
-                                             transaction:(SentryTransaction *)transaction
++ (nullable NSArray<SentrySample *> *)slicedSamples:(NSArray<SentrySample *> *)samples
+                                        transaction:(SentryTransaction *)transaction
 {
     if (samples.count == 0) {
         return nil;
     }
 
     const auto firstIndex = [samples indexOfObjectPassingTest:^BOOL(
-        SentrySampleEntry *_Nonnull sample, NSUInteger idx, BOOL *_Nonnull stop) {
+        SentrySample *_Nonnull sample, NSUInteger idx, BOOL *_Nonnull stop) {
         *stop = sample.absoluteTimestamp >= transaction.startSystemTime;
         return *stop;
     }];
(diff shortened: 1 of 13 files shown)


Sources/Sentry/SentryProfiler.mm
```diff
@@ -370,56 +422,30 @@ + (SentryEnvelopeItem *)createProfilingEnvelopeItemForTransaction:(SentryTransac
...
-              const auto nextSlicedMetrics = [self slicedArray:nextMetrics transaction:transaction];
-              if (!nextSlicedMetrics) {
-                  return;
-              }
-
-              slicedMetrics[metricKey] =
-                  @{ @"unit" : nextMetricsEntry[@"unit"], @"values" : nextSlicedMetrics };
-          };
-    [metrics enumerateKeysAndObjectsUsingBlock:metricSlicer];
+    const auto metrics = [_gCurrentProfiler->_metricProfiler serializeForTransaction:transaction];
```
armcknight, line 423: This implicit slicing of metrics entries now takes place in `SentryMetricProfiler::serializeValuesWithNormalizedTime(...)` (called by `-[SentryMetricProfiler serializeForTransaction:]`).

Sources/Sentry/SentryProfiler.mm
```diff
@@ -370,56 +421,30 @@ + (SentryEnvelopeItem *)createProfilingEnvelopeItemForTransaction:(SentryTransac
...
-        = processFrameRates(_gCurrentFramesTracker.currentFrames.frameRateTimestamps,
-            _gCurrentProfiler->_startTimestamp);
+        = processFrameRates(_gCurrentFramesTracker.currentFrames.frameRateTimestamps, transaction);
     if (frameRates.count > 0) {
-        const auto slicedFrameRates = [self slicedArray:frameRates transaction:transaction];
-        if (slicedFrameRates.count > 0) {
-            slicedMetrics[@"screen_frame_rates"] =
-                @{ @"unit" : @"hz", @"values" : slicedFrameRates };
-        }
+        metrics[@"screen_frame_rates"] = @{ @"unit" : @"hz", @"values" : frameRates };
```
armcknight, line 441: The slicing of frame rate entries is now done directly in processFrameRates with a couple timestamp checks.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -336,9 +380,9 @@ + (SentryEnvelopeItem *)createProfilingEnvelopeItemForTransaction:(SentryTransac
         return nil;
     }
 
-    NSMutableDictionary<NSString *, id> *payload = [_gCurrentProfiler->_profileData mutableCopy];
+    const auto payload = [NSMutableDictionary<NSString *, id> dictionary];
 
-    NSArray *samples = ((NSArray *)payload[@"profile"][@"samples"]).mutableCopy;
+    NSArray<SentrySample *> *samples = _gCurrentProfiler->_profileData[@"profile"][@"samples"];
```
armcknight, line 385: `_profileData` is a structure with nested dictionaries, and this copy operation only performed a shallow copy, and logic below wound up overwriting parts of it, and in tests of concurrent transactions, that wound up corrupting later serializations. Instead, I did away with any mutable copies and just create new structures, and manually assigned all the other separate substructures on lines 404-410 below.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -183,63 +181,109 @@
...
-        [relativeFrameRates addObject:@{
-            @"elapsed_since_start_ns" : @(relativeTimestamp),
+        if (orderedChronologically(transaction.endSystemTime, timestamp)) {
+            return;
+        }
+
+        const auto relativeTimestamp = getDurationNs(transaction.startSystemTime, timestamp);
+
+        [relativeFrameRates addObject:@ {
+            @"elapsed_since_start_ns" : [NSString stringWithFormat:@"%llu", relativeTimestamp],
```
armcknight, line 248: Interesting tidbit here, we should always be sending a string for `elapsed_since_start_ns` since it's an unsigned 64 bit integer which JSON technically doesn't support! Somehow this crept in as an `NSNumber` previously... probably isn't ever failing in practice, but it's nicer to be consistent.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -721,7 +738,6 @@ - (void)stop
     }
 
     _profiler->stopSampling();
-    _endDate = [SentryCurrentDate date];
```
Zylphrex, line 724: Can't comment on the line directly but there's a line below that says
```objc
profile[@"timestamp"] = [[SentryCurrentDate date] sentry_toIso8601String];
```
where you set the timestamp on the profile. If I'm reading it right, this is setting the timestamp on the profile to be the current time when the profile is being serialized right? So the adjustments we'll be doing on the UI will be based on this timestamp. So each sample is assumed to have occurred at the profile's timestamp + elapsed since start ns. And ideally, the profile's timestamp == transaction's start timestamps which is what I assume you're doing in this PR (I didn't read the rest)

Sources/Sentry/include/SentryTime.h
```diff
@@ -16,8 +17,20 @@ uint64_t timeIntervalToNanoseconds(double seconds);
  */
 uint64_t getAbsoluteTime(void);
 
+/**
+ * Check whether two timestamps provided as 64 bit unsigned integers are in normal
+ * chronological order, as a convenience runtime check before using @c getDurationNs.
+ * Equal timestamps are considered to be valid chronological order.
+ * @return @c true if @c b>=a, otherwise return @c false.
+ */
+bool orderedChronologically(uint64_t a, uint64_t b);
```
philipphofmann, line 26: That makes the code way easier to read than a<=b. Thanks 👏

Sources/Sentry/SentryProfiler.mm
```diff
@@ -133,21 +133,19 @@
...
-    const auto sample = [NSMutableDictionary<NSString *, id> dictionary];
-    sample[@"elapsed_since_start_ns"] =
-        [@(getDurationNs(startTimestamp, backtrace.absoluteTimestamp)) stringValue];
-    sample[@"thread_id"] = threadID;
+    const auto sample = [[SentrySample alloc] init];
+    sample.absoluteTimestamp = backtrace.absoluteTimestamp;
+    sample.threadID = backtrace.threadMetadata.threadID;
     if (queueAddress != nil) {
-        sample[@"queue_address"] = queueAddress;
+        sample.queueAddress = queueAddress;
```
philipphofmann, line 140: Way better to read 👍

Sources/Sentry/SentryProfiler.mm
philipphofmann: `l`: Maybe it makes sense to create a function for `[NSString stringWithFormat:@"%llu", timestamp]` so it's easier to change.

philipphofmann (approved):
Woa, big refactoring. Didn't see anything suspicious. The code is now easier to read. Thanks for all the comments to help with the review 🥇

Sources/Sentry/SentryProfiler.mm
armcknight: I'm pretty sure I remember this was supposed to be the time that we were sending the profile, and the timestamp for the first sample would be used for the start, but sadly that convo is long gone in slack, and isn't on GitHub. We can easily change this is that is not the case.

armcknight: @phacops can you comment on whether it would break anything if we changed this assumption, or is the assumption incorrect right now?

armcknight: in any case we can handle this in a separate PR

Sources/Sentry/SentryProfiler.mm
armcknight: Done! https://github.com/getsentry/sentry-cocoa/pull/2729/commits/dffd985bf73f25c7f00d4d726aa4ae00f9b2c257

Sources/Sentry/SentryProfiler.mm
phacops: You can change the assumption now, we never really used that field. It was meant to be something different indeed when we had support for multiple transactions, now it's best to be the reference time from which `elapsed_since_start_ns` are relative to.

Also, a maximum amount of precision is best, we support `RFC3339Nano` format, like `2023-03-01T10:00:00.123456789-07:00`.

armcknight: In that case can we remove it altogether since all times will be relative to the transaction start which is reported in the main envelope payload, or is it not possible to grab that?

I’ll have to check on RFC3339 support in iOS, I know there’s a builtin way to get three sigfigs on the seconds fraction for milliseconds, need to check about 9 for nano.

phacops: Actually, the best value for you would be exactly what you used for the start of the transaction at the same precision.

In the backend, we'll still use that field to do `timestamp - start of the transaction` and adjust relative timestamps, but in this case, it should be equal to 0 and no adjustment will take place.

armcknight: Please double check that I've captured the requirement correctly in https://github.com/getsentry/team-profiling/issues/191 @phacops @Zylphrex

PR #2731, opened 2023-02-28
ref: remove deprecated profile payload fields

armcknight:
## :scroll: Description

- Remove a few fields in the serialized profile payload, as well as all the method parameters and ivars needed to pass them around. 
- Also update the schema of the linked transaction info.

#skip-changelog

Now that we're reporting all profiling timestamps relative to transaction start, a few of the profile payload fields are redundant and can be removed. See https://github.com/getsentry/team-profiling/issues/186
## :pencil: Checklist


Sources/Sentry/SentryProfiler.mm
```diff
@@ -408,16 +405,12 @@ + (SentryEnvelopeItem *)createProfilingEnvelopeItemForTransaction:(SentryTransac
...
-    const auto profileDuration = getDurationNs(firstSampleTimestamp, getAbsoluteTime());
-
-    const auto transactionInfo = [self serializeInfoForTransaction:transaction
-                                                   profileDuration:profileDuration];
+    const auto transactionInfo = [self serializeInfoForTransaction:transaction];
     if (!transactionInfo) {
         SENTRY_LOG_WARN(@"Could not find any associated transaction for the profile.");
         return nil;
     }
-    payload[@"transactions"] = @[ transactionInfo ];
```
armcknight, line 420: This is the entirety of the linked transaction info schema change. Since profiles are now 1:1 with profiles again, we don't have to send an array. And the key name changed plural to singular.

philipphofmann (approved):
LGTM, it's always great to delete code 🥇

PR #2743, opened 2023-03-02
chore: Remove private references from public headers

brustolin:
## :scroll: Description

We made SentrySession private but we did not remove references from public headers.

Also changing a few forward declarations to #import to help with .NET hybrid SDK.

**This may be considered a "breaking change" since we're removing property from a public header, but is a property that was not meant to be used.**

## :pencil: Checklist

_#skip-changelog_

### DIFF
--- a/Sources/Sentry/include/SentryTraceContext.h
+++ b/Sources/Sentry/include/SentryTraceContext.h
@@ -12,7 +12,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class SentryScope, SentryId, SentryOptions, SentryTracer, SentryUser, SentryBaggage;
+@class SentryScope, SentryOptions, SentryTracer, SentryUser, SentryBaggage;
 
 @interface SentryTraceContext : NSObject <SentrySerializable>
 
--- a/Sources/Sentry/include/SentryTraceContext.h
+++ b/Sources/Sentry/include/SentryTraceContext.h
@@ -1,9 +1,18 @@
-#import "SentryId.h"
+#if __has_include(<Sentry/SentrySerializable.h>)
+#import <Sentry/SentrySerializable.h>
+#else
 #import "SentrySerializable.h"
+#endif
+
+#if __has_include(<Sentry/SentryId.h>)
+#import <Sentry/SentryId.h>
+#else
+#import "SentryId.h"
+#endif
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class SentryScope, SentryOptions, SentryTracer, SentryUser, SentryBaggage;
+@class SentryScope, SentryId, SentryOptions, SentryTracer, SentryUser, SentryBaggage;
 
 @interface SentryTraceContext : NSObject <SentrySerializable>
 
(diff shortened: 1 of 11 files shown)


armcknight:
@brustolin could you link to some context around how the forward class declarations were causing problems in the .NET SDK?

brustolin:
@armcknight [This is the PR](https://github.com/getsentry/sentry-dotnet/pull/2203) that started the discussion. 
Matt did a workaround with a script, but that is not the optimal solution.

brustolin:
I strongly believe that forward declaration should be only used to avoid circular reference.
If a class strongly depends on a second one, an import is the best approach to avoid the need to import a bunch of headers in the implementation file of a third class that wants to use the first one.

armcknight:
> @armcknight [This is the PR](https://github.com/getsentry/sentry-dotnet/pull/2203) that started the discussion.
> Matt did a workaround with a script, but that is not the optimal solution.

Hmm, I'm not seeing discussion about any issues around this. Does this comment in the linked issue allude to the problem? https://github.com/getsentry/sentry-dotnet/issues/2148#issuecomment-1411269898

armcknight:
> I strongly believe that forward declaration should be only used to avoid circular reference.
> If a class strongly depends on a second one, an import is the best approach to avoid the need to import a bunch of headers in the implementation file of a third class that wants to use the first one.

This is an unorthodox view, and one I disagree with based on experience. It is not fun to disentangle a chain of implicitly imported headers when you finally _do_ hit a circular reference. 

Much has been written about this, here is a good article that talks about the tradeoffs: https://qualitycoding.org/file-dependencies/

brustolin:
> I'm not seeing discussion

Oh, sorry, the discussion happened in slack. Short version, the C# tooling to do a binding to objc class don't now the body of a forward declaration if the header is not included or have no reference.

> This is an unorthodox view, and one I disagree with based on experience, It is not fun to disentangle a chain of implicitly imported headers when you finally do hit a circular reference.

By experience I would say is much better avoiding cyclic reference in classes itself, then you don't need to deal with header cyclic reference. If it is hard to "disentangle a chain of implicitly imported headers when you finally do hit a circular reference" that's a good sign that it's time to rethink the project design.

Sources/Sentry/include/SentryTraceContext.h
```diff
@@ -1,9 +1,18 @@
...
+#if __has_include(<Sentry/SentryId.h>)
+#    import <Sentry/SentryId.h>
+#else
+#    import "SentryId.h"
+#endif
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class SentryScope, SentryOptions, SentryTracer, SentryUser, SentryBaggage;
+@class SentryScope, SentryId, SentryOptions, SentryTracer, SentryUser, SentryBaggage;
```
armcknight, line 15: We wouldn't need this forward class dec since SentryId.h is always being imported one way or another

armcknight (commented):
We should set up integration tests for .NET/RN/Flutter so we can make sure we're not breaking things there. This is some weird, complicated stuff that I won't pretend to understand 😂

brustolin:
> We should set up integration tests for .NET/RN/Flutter so we can make sure we're not breaking things there. This is some weird, complicated stuff that I won't pretend to understand 😂

Yes, we have an issue for this already.

Sources/Sentry/include/SentryTraceContext.h
brustolin, line 15:
```suggestion
@class SentryScope, SentryOptions, SentryTracer, SentryUser, SentryBaggage;
```

philipphofmann (commented):
Thanks for doing this cleanup, @brustolin. I have two questions.

philipphofmann (approved):
LGTM

mattjohnsonpint:
FYI, This made `SentrySdkInfo` and `SentryTraceContext` available, but now I have the same issue with `SentryBaggage` and `SentryTraceHeader`.  I still have it with `SentrySession` as before.

Now that I know how to make type-forwarded stubs in .NET, it's not a huge deal - but I thought you'd want to know.

See `PrivateApiDefinitions.cs` in https://github.com/getsentry/sentry-dotnet/pull/2248/commits/77a69c578199f8195dfcdddf1375458b715bb47b#diff-fbdac032f98e5bb7b38c5933992d8ae94ab4e29c4472e092f7acb01cecc1b3bb

brustolin:
@mattjohnsonpint It looks like we gonna have to solve this with C# tooling. If we keep rearranging header, most likely other headers will became an issue.

PR #2749, opened 2023-03-03
ref: Code docs default for sessionTrackingInterval

philipphofmann:
Add default value for SentryOptions.sessionTrackingIntervalMillis to code docs.

#skip-changelog

### DIFF
--- a/Sources/Sentry/Public/SentryOptions.h
+++ b/Sources/Sentry/Public/SentryOptions.h
@@ -146,6 +146,7 @@ NS_SWIFT_NAME(Options)
 
 /**
  * The interval to end a session if the App goes to the background.
+ * The default is 30000 millis / 30 seconds.
  */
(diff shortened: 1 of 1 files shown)


brustolin (approved):

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -146,6 +146,7 @@ NS_SWIFT_NAME(Options)
 
 /**
  * The interval to end a session if the App goes to the background.
+ * The default is 30000 millis / 30 seconds.
```
armcknight, line 149:
```suggestion
 * The interval to end a session after the App goes to the background.
 * The default is 30000 millis / 30 seconds.
```

armcknight (approved):
One quick change that I think explains more clearly.

PR #2753, opened 2023-03-03
reorder profiler tests

armcknight:
#skip-changelog


armcknight:
Oh, this put them at the end, not the beginning, like i was trying to test.

PR #2755, opened 2023-03-03
ref: split test target into utils and profiling test targets

armcknight:
- isolate the profiling test cases from the rest of the test suite
- extract test wrapper classes into a shared SentryTestUtils lib. this pr only moves those necessary to get the profiler test target building; in the future we should move all such wrappers here, and any other code that is not a direct implementation of test cases. This required marking all such wrapper classes and their functions/properties as `public`, accounting for basically all of the diffs in those files.
- remove `SentryFileIOTrackingIntegrationTests` from `iOS-SwiftUITests`. This was causing build issues because it calls clearTestState which calls into some API in `PrivateSentrySDKOnly`, which caused some other weirdness I can't remember now. But it doesn't seem like we actually need this running in the sample app's ui tests any more. There was a comment stating: "This test is also executed under iOS-SwiftUITests, because GitHub Actions doesn't have simulators for iOS 11 and 10. That's why we need to keep it generic, without access to any private part of the SDK." This doesn't make sense–if there is no iOS 10/11 simulator for unit tests, then it's also not available for UI tests. 
- Split out DeploymentTargets.xcconfig to include into SDK.xcconfig and SentryTests.xcconfig. This is another area we should do more in, to eventually get all build settings out of the pbxproj file and into DRY xcconfigs.

#skip-changelog


SentryTestUtils/Dynamic/Dynamic.swift
```diff
@@ -283,7 +283,9 @@ extension Dynamic {
         
         var storedSize = 0
         var storedAlignment = 0
+        //swiftlint:disable force_unwrapping
         NSGetSizeAndAlignment(invocation.returnType!, &storedSize, &storedAlignment)
+        //swiftlint:enable force_unwrapping
```
armcknight, line 288: This file must have been added before we started using swiftlint. because i renamed the file, it got linted. I don't want to change the logic in this file as part of this PR.

SentryTestUtils/TestClient.swift
```diff
@@ -0,0 +1,183 @@
+import Foundation
```
armcknight, line 1: Nothing actually changed in this file besides marking things `public`. It just didn't diff well with its rename.

SentryTestUtils/TestConstants.swift
```diff
@@ -0,0 +1,28 @@
...
+     */
+    public static let realDSN: String = "https://6cc9bae94def43cab8444a99e0031c28@o447951.ingest.sentry.io/5428557"
+    
+    public static func dsnAsString(username: String) -> String {
+        return "https://\(username):password@app.getsentry.com/12345"
+    }
+    
+    public static func dsn(username: String) throws -> SentryDsn {
+        return try SentryDsn(string: self.dsnAsString(username: username))
+    }
```
armcknight, line 14: I changed this function to throw the exception up the stack instead of calling `XCTFail` in a `do/catch`, since we can't import XCTest into a static lib target. That had some downstream effects that i'll call out in comments.

SentryTestUtils/TestRandom.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
```
armcknight, line 1: Another renamed file with no changes other than `public`

SentryTestUtils/TestSentryNSProcessInfoWrapper.swift
```diff
@@ -0,0 +1,13 @@
+import Sentry
```
armcknight, line 1: Another renamed file with no changes other than `public`

SentryTestUtils/TestSentryNSTimerWrapper.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
```
armcknight, line 1: Another renamed file with no changes other than `public`

SentryTestUtils/TestSentrySystemWrapper.swift
```diff
@@ -0,0 +1,28 @@
+import Sentry
```
armcknight, line 1: Another renamed file with no changes other than `public`

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -11,7 +12,7 @@ class SentryProfilerSwiftTests: XCTestCase {
             options.dsn = SentryProfilerSwiftTests.dsnAsString
             return options
         }()
-        lazy var client: TestClient = TestClient(options: options)!
+        lazy var client: TestClient? = TestClient(options: options)
```
armcknight, line 15: No test logic changes here; I converted everything to avoid force-unwrapping to avoid crashing the test process.

Tests/SentryTests/Helper/SentryAppStateManagerTests.swift
```diff
@@ -1,9 +1,9 @@
+import SentryTestUtils
 import XCTest
 
 #if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
 class SentryAppStateManagerTests: XCTestCase {
     private static let dsnAsString = TestConstants.dsnAsString(username: "SentryOutOfMemoryTrackerTests")
-    private static let dsn = TestConstants.dsn(username: "SentryOutOfMemoryTrackerTests")
```
armcknight, line 6: There were a bunch of these that weren't actually being used. I figured it was a copypasta.

Tests/SentryTests/Integrations/Performance/IO/SentryFileIOTrackingIntegrationTests.swift
```diff
@@ -1,11 +1,8 @@
 import Foundation
 import Sentry
+import SentryTestUtils
 import XCTest
- 
-// This test is also executed under iOS-SwiftUITests, because
-// GitHub Actions doesn't have simulators for iOS 11 and 10.
-// That's why we need to keep it generic, without access
-// to any private part of the SDK.
```
armcknight, line 8: This didn't make sense to me. If we have no iOS 10/11 simulator for unit tests, how would we have one for UI tests? I removed this file from the iOS-SwiftUITest target.

Tests/SentryTests/Integrations/Performance/SentryPerformanceTracker+Testing.h
```diff
@@ -1,6 +1,7 @@
 #import "SentryPerformanceTracker.h"
 #import "SentrySpan.h"
 #import "SentrySpanId.h"
+#import "SentryTracer.h"
```
armcknight, line 4: This was implicitly imported from... somewhere? This file tries to use `SentryTracerDelegate`, and this wound up breaking after I moved some stuff around.

Tests/SentryTests/SentryNSURLRequestTests.swift
```diff
@@ -1,17 +1,20 @@
...
+    }
     
     func testRequestWithEnvelopeEndpoint() {
-        let request = try! SentryNSURLRequest(envelopeRequestWith: SentryNSURLRequestTests.dsn, andData: Data())
+        let request = try! SentryNSURLRequest(envelopeRequestWith: SentryNSURLRequestTests.dsn(), andData: Data())
         XCTAssertTrue(request.url!.absoluteString.hasSuffix("/envelope/"))
     }
     func testRequestWithStoreEndpoint() {
-        let request = try! SentryNSURLRequest(storeRequestWith: SentryNSURLRequestTests.dsn, andData: Data())
+        let request = try! SentryNSURLRequest(storeRequestWith: SentryNSURLRequestTests.dsn(), andData: Data())
```
armcknight, line 17: Another instance where TestConstants.dsn no longer calling XCTFail had a downstream effect.

Tests/SentryTests/Networking/SentryHttpTransportTests.swift
```diff
@@ -107,9 +107,13 @@ class SentryHttpTransportTests: XCTestCase {
...
     }
 
+    class func dsn() throws -> SentryDsn {
+        try TestConstants.dsn(username: "SentryHttpTransportTests")
+    }
+
     class func buildRequest(_ envelope: SentryEnvelope) -> SentryNSURLRequest {
         let envelopeData = try! SentrySerialization.data(with: envelope)
-        return try! SentryNSURLRequest(envelopeRequestWith: SentryHttpTransportTests.dsn, andData: envelopeData)
+        return try! SentryNSURLRequest(envelopeRequestWith: dsn(), andData: envelopeData)
```
armcknight, line 116: Here's the downstream effect of removing the XCTFail from the DSN construction/retrieval in TestConstants.

Tests/SentryTests/Integrations/Performance/IO/SentryFileIOTrackingIntegrationTests.swift
brustolin: UI Tests runs under SouceLabs that contains iOS 11 and 10, but im fine by removing this from ui tests.

armcknight: ahhh that’s right, totally forgot about saucelabs 😂

armcknight:
> Regarding the code, it seems ok.
> But I'm wondering what's the benefit of creating an aux lib for testing.
> Seems like more trouble than help IMO.

They are all wrapper classes compiled directly into the SentryTests target, that would've had to also be included into the new SentryProfilerTests target. IME multiple target membership leads to confusing complexity, and splitting out a separate lib alleviates that, while also avoiding duplicate compilation. We can even cache the compilation output in CI as those classes probably won't change much.

What kinds of trouble do you anticipate @brustolin?

brustolin:
> What kinds of trouble do you anticipate @brustolin?

Just more work.
First we need to remember that we have a separated lib for aux classes.
Then we need to create aux classes in the separated lib and remember to import it on the test itself.

I mean, is annoying but not a big deal. 
TBH I prefer self contained tests without all this references to shared classes/data.

PR #2761, opened 2023-03-06
randomize test execution order with scheme option

armcknight:
We shouldn't merge this, I just want to see the behavior of our tests if the order is randomized using the Xcode scheme option:
Screenshot 2023-03-06 at 9 27 38 AM

Needed to adjust the test triggers in test.yml to run the workflow whenever a scheme file changes, which we should definitely merge in a separate PR!

#skip-changelog


armcknight:
Experiment complete.

PR #2765, opened 2023-03-06
test: cancel the profiling thread at the end of the test

armcknight:
I noticed looking through the open threads when a [probably] unrelated crash occurred while running tests locally that there is still a thread hanging around from `-[SentrySamplingProfilerTests testProfiling]` where we never end the `idleThreadEntry` thread. This fixes that.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Profiling/SentrySamplingProfilerTests.mm
+++ b/Tests/SentryTests/Profiling/SentrySamplingProfilerTests.mm
@@ -54,12 +54,17 @@ using namespace sentry::profiling;
     XCTAssertGreaterThan(std::chrono::duration_cast<std::chrono::seconds>(duration).count(), 0);
     XCTAssertGreaterThan(profiler->numSamples(), static_cast<std::uint64_t>(0));
     XCTAssertGreaterThan(numIdleSamples, 0);
-    pthread_cancel(idleThread);
+
+    // end the test idle thread
+    XCTAssertEqual(pthread_cancel(idleThread), 0);
+    XCTAssertEqual(pthread_join(idleThread, nullptr), 0);
 }
 
 static void *
 idleThreadEntry(__unused void *ptr)
 {
+    pthread_testcancel();
+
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/Profiling/SentrySamplingProfilerTests.mm
```diff
@@ -54,11 +54,17 @@ - (void)testProfiling
     XCTAssertGreaterThan(std::chrono::duration_cast<std::chrono::seconds>(duration).count(), 0);
     XCTAssertGreaterThan(profiler->numSamples(), static_cast<std::uint64_t>(0));
     XCTAssertGreaterThan(numIdleSamples, 0);
+
+    // end the test idle thread
+    XCTAssertEqual(pthread_cancel(idleThread), 0);
+    XCTAssertEqual(pthread_join(idleThread, nullptr), 0);
```
indragiek, line 60: This will deadlock the test if the `pthread_cond_wait` on line 74 executes before you call `pthread_join` here - a race condition. The `pthread_cond_wait` will never return since the condition variable is never signaled, and therefore `pthread_join` will never return because the thread never exits. Two options:

1) This is unlikely to fix the issue you noted in the summary, so there's no harm in leaving an idle thread hanging (`pthread_cond_wait` means that it won't be busy-waiting while idle, so it won't be using any resources)
2) Refactor to make the `pthread_cond_t` a global and then signal the condition variable before you call `pthread_join`. I don't know how complex of a refactor this would be but it could have hidden gotchas (e.g. if tests are running in parallel, in which case having an unsychronized global won't work)

indragiek (changes requested):

Tests/SentryTests/Profiling/SentrySamplingProfilerTests.mm
armcknight: OK, i'm not sure it's worth spending any more effort on this unless we think it actually causes problems later, which it sounds like it doesn't.

PR #2771, opened 2023-03-08
fix: set the profile payload "timestamp" field to the transaction's start

armcknight:
if there is one, or fall back to previous behavior

for https://github.com/getsentry/team-profiling/issues/191

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -445,7 +445,7 @@ serializedSamplesWithRelativeTimestamps(
 
     // add the remaining basic metadata for the profile
     const auto profileID = [[SentryId alloc] init];
-    [self serializeBasicProfileInfo:payload profileID:profileID platform:transaction.platform];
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentryProfiler.mm
```diff
@@ -766,10 +766,19 @@ + (void)serializeBasicProfileInfo:(NSMutableDictionary<NSString *, id> *)profile
...
-    profile[@"timestamp"] = [[SentryCurrentDate date] sentry_toIso8601String];
+
+    const auto timestamp = transaction.startTimestamp;
+    if (UNLIKELY(timestamp == nil)) {
+        SENTRY_LOG_WARN(@"There was no start timestamp on the provided transaction. Falling back "
+                        @"to old behavior of using the current time.");
+        profile[@"timestamp"] = [[SentryCurrentDate date] sentry_toIso8601String];
+    } else {
+        profile[@"timestamp"] = [timestamp sentry_toIso8601String];
+    }
```
armcknight, line 780: Here's the real meat of the change

brustolin (approved):
LGTM

PR #2772, opened 2023-03-09
fix: profile start on app launch 

armcknight:
We noticed that frames collected for a profile connected to a transaction containing app start span were not aligned to the correct point in the span tree. This is because the transaction's start time is adjusted to the start of the app sequence. 

This PR keeps the original start time in a new property so we know where to align the frames.

For https://github.com/getsentry/team-profiling/issues/189

#skip-changelog


Sentry.xcodeproj/project.pbxproj
```diff
@@ -607,7 +607,6 @@
 		8431EFE229B27BAD00D8DC56 /* SentryNSTimerWrapperTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 849472842971C41A002603DE /* SentryNSTimerWrapperTest.swift */; };
 		8431EFE529B27BAD00D8DC56 /* SentryNSProcessInfoWrapperTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 849472822971C2CD002603DE /* SentryNSProcessInfoWrapperTests.swift */; };
 		8431EFE829B27BAD00D8DC56 /* SentrySystemWrapperTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 849472802971C107002603DE /* SentrySystemWrapperTests.swift */; };
-		8431F00529B2849A00D8DC56 /* TestProcessInfoWrapper.swift in Sources */ = {isa = PBXBuildFile; fileRef = D800B05929707A5300B8E20F /* TestProcessInfoWrapper.swift */; };
```
armcknight, line 610: Not sure how these were leftover, but this was Xcode automatically removing them.

Sources/Sentry/SentryTracer.m
```diff
@@ -649,9 +648,8 @@ - (nullable SentryAppStartMeasurement *)getAppStartMeasurement
 
     NSTimeInterval difference = [appStartEndTimestamp timeIntervalSinceDate:self.startTimestamp];
 
-    // If the difference between the end of the app start and the beginning of the current
-    // transaction is smaller than SENTRY_APP_START_MEASUREMENT_DIFFERENCE. With this we
-    // avoid messing up transactions too much.
+    // Don't attach app start measurements if too much time elapsed between the end of the app start
+    // sequence and the start of the transaction. This makes transactions too long.
```
armcknight, line 652: This comment didn't make sense so I reworded it and took a guess as to what "messing up transactions too much" meant. But, I'd like more context around _why_ we don't want to attach such spans.

Tests/SentryTests/Performance/SentryTracerTests.swift
```diff
@@ -56,8 +56,6 @@ class SentryTracerTests: XCTestCase {
             client.options.tracesSampleRate = 1
             hub = TestHub(client: client, andScope: scope)
             
-            CurrentDate.setCurrentDateProvider(currentDateProvider)
```
armcknight, line 59: This was in this init twice.

Tests/SentryTests/Performance/SentryTracerTests.swift
```diff
@@ -108,17 +106,11 @@ class SentryTracerTests: XCTestCase {
     override func setUp() {
         super.setUp()
         fixture = Fixture()
-        SentryTracer.resetAppStartMeasurementRead()
```
armcknight, line 111: Moved this to `clearTestState()`

Tests/SentryTests/Performance/SentryTracerTests.swift
```diff
@@ -108,17 +106,11 @@ class SentryTracerTests: XCTestCase {
...
     }
     
     override func tearDown() {
         super.tearDown()
         clearTestState()
-        SentryTracer.resetAppStartMeasurementRead()
-#if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
-        SentryFramesTracker.sharedInstance().resetFrames()
-        SentryFramesTracker.sharedInstance().stop()
-#endif
```
armcknight, line 121: This is all already in `clearTestState()`.

Sources/Sentry/SentryTracer+Private.h
```diff
@@ -0,0 +1,16 @@
+#import "SentryTracer.h"
+
+@interface
+SentryTracer ()
+
```
brustolin, line 5: m: The SentryTracer is already private and will never be public, I don't see a problem with these properties being part of the SentryTracer, and then we will have one less header.

Sources/Sentry/SentryTracer+Private.h
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * When an app launch is traced, after building the app start spans, the tracer's start timestamp is
+ * adjusted backwards to be the start of the first app start span. But, we still need to know the
+ * real start time of the trace for other purposes. This property provides a place to keep it before
+ * reassigning it.
+ */
+@property (strong, nonatomic) NSDate *originalStartTimestamp;
```
brustolin, line 12: l: Maybe make this a calculated property like this: 

``` Obj-c
- (NSDate*) originalStartTimestamp {
    return self.startTimeChanged ? _originalStartTimestamp  : self.startTimestamp;
}
```

so we don't have to use this if-ternary every time this property is needed, as you did in the SentryProfiler

brustolin (approved):
LGTM

With a couple of suggestions.

Sources/Sentry/SentryTracer+Private.h
armcknight: Yep, makes sense! Will do.

Sources/Sentry/SentryTracer+Private.h
armcknight: Yes, good call 👍🏻

PR #2774, opened 2023-03-09
meta: update brew and gem bundles

armcknight:
There's one job that's failing in https://github.com/getsentry/sentry-cocoa/pull/2764, my guess is it's because dependabot doesn't have access to the private certificates repo. Seeing if it runs if I open a PR to update cocoapods instead, and while I'm at it just update all the things.

#skip-changelog


Gemfile.lock
```diff
@@ -309,4 +307,4 @@ DEPENDENCIES
   xcpretty
 
 BUNDLED WITH
-   2.4.7
```
armcknight, line 312: I also upgraded `bundler` on my machine, but it's still behind whatever version whoever did this last is on. What version of ruby are you running locally @brustolin and @philipphofmann ? I'm on `2.7.5`.

brustolin: 2.6.10

brustolin (approved):

Gemfile.lock
armcknight: Huh, they're still shipping newer patches for earlier minor releases. Interesting!
- 2.7.5 released 24 Nov 2021
- 2.6.10 released 12 Apr 2022

PR #2778, opened 2023-03-10
chore: Add package manager information 

brustolin:
Added the package manager information in the SDK interface.

This will help us understand how users install Sentry SDK.

_#skip-changelog_


Sentry.xcodeproj/project.pbxproj
```diff
@@ -607,6 +607,7 @@
 		8431EFE229B27BAD00D8DC56 /* SentryNSTimerWrapperTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 849472842971C41A002603DE /* SentryNSTimerWrapperTest.swift */; };
 		8431EFE529B27BAD00D8DC56 /* SentryNSProcessInfoWrapperTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 849472822971C2CD002603DE /* SentryNSProcessInfoWrapperTests.swift */; };
 		8431EFE829B27BAD00D8DC56 /* SentrySystemWrapperTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 849472802971C107002603DE /* SentrySystemWrapperTests.swift */; };
+		8431F00529B2849A00D8DC56 /* (null) in Sources */ = {isa = PBXBuildFile; };
```
armcknight, line 610: Looks like the ghost of the old process info wrapper is still hanging around

Sentry.xcodeproj/project.pbxproj
```diff
@@ -5039,6 +5045,7 @@
 				LD_DYLIB_INSTALL_NAME = "$(DYLIB_INSTALL_NAME_BASE:standardizepath)/$(EXECUTABLE_PATH)";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				ONLY_ACTIVE_ARCH = YES;
+				OTHER_CFLAGS = "-DCARTHAGE_$(CARTHAGE)";
```
armcknight, line 5048: Why `OTHER_CFLAGS` instead of `GCC_PREPROCESSOR_DEFINITIONS`?

Sources/Sentry/SentrySdkInfo.m
```diff
@@ -1,25 +1,25 @@
 #import "SentrySdkInfo.h"
 #import <Foundation/Foundation.h>
 
+#if SWIFT_PACKAGE
+static NSString *SENTRY_PACKAGE_INFO = @"spm:getsentry/%@";
+#elif COCOAPODS
+static NSString *SENTRY_PACKAGE_INFO = @"cocoapods:getsentry/%@";
+#elif CARTHAGE_YES
```
armcknight, line 8: Does `#if defined(CARTHAGE)` not work here? Why do we need to embed it into a new definition?

armcknight (approved):
Looks good, just a couple questions.

Sentry.xcodeproj/project.pbxproj
brustolin: It was my first idea but it didn't work, which for me is weird.

Sources/Sentry/SentrySdkInfo.m
brustolin: Nope, it does not. 
CARTHAGE is a xcodebuild build setting, the only way I found to make it work is by "converting" it into a new definition.

Sources/Sentry/SentrySdkInfo.m
```diff
@@ -60,7 +60,19 @@ - (instancetype)initWithDictInternal:(NSDictionary *)dict orDefaults:(SentrySdkI
 
 - (NSDictionary<NSString *, id> *)serialize
 {
-    return @{ @"sdk" : @ { @"name" : self.name, @"version" : self.version } };
+    NSMutableDictionary *sdk = @{
+        @"name" : self.name,
+        @"version" : self.version,
+    }
+                                   .mutableCopy;
+    if (SENTRY_PACKAGE_INFO != nil) {
```
philipphofmann, line 68: `m`: Please add some tests as pointed out by codecov.

philipphofmann (commented):
Can you explain how this works in our repo develop docs? In don't get it.

develop-docs/README.md
```diff
@@ -196,3 +196,11 @@ Date: January 18th, 2023
...
 
 We release experimental SentrySwiftUI cocoa package with the version 8.0.0 because all podspecs file in a repo need to have the same version. 
+
+
+## Tracking package managers
+
+To be able to identify the package manager(PM) being used by the user, we need that the PM identify itself.
+Luckily all of the 3 PMs we support do this in some way, mostly by exposing a compiler directive (SPM, COCOA)
+or a build setting (CARTHAGE). With this information we can create a conditional compilation that injects the name of
+the PM. You can found this at `SentrySDKInfo.m`.
```
philipphofmann, line 206:
```suggestion
the PM. You can find this in `SentrySDKInfo.m`.
```

Sources/Sentry/SentrySdkInfo.m
philipphofmann: `m`: Please add the info you posted ⬆️ as a comment @brustolin.

Sources/Sentry/SentrySdkInfo.m
```diff
@@ -1,25 +1,30 @@
 #import "SentrySdkInfo.h"
 #import <Foundation/Foundation.h>
 
+/**
+ * This is required to identify the package manager used when installing sentry.
+ */
+#if SWIFT_PACKAGE
+static NSString *SENTRY_PACKAGE_INFO = @"spm:getsentry/%@";
```
philipphofmann, line 8: `m`: I would prefer adding tests for these constants here. I think you need to restructure the code a bit to make it work. As we use them in `NSString stringWithFormat` a small typo could easily lead to a crash we only detect after releasing.

develop-docs/README.md
```diff
@@ -196,3 +196,11 @@ Date: January 18th, 2023
 Contributors: @brustolin and @philipphofmann
 
 We release experimental SentrySwiftUI cocoa package with the version 8.0.0 because all podspecs file in a repo need to have the same version. 
+
+
+## Tracking package managers
```
philipphofmann, line 201: Thanks 👏

Sources/Sentry/SentrySdkInfo.m
```diff
@@ -1,30 +1,43 @@
 #import "SentrySdkInfo.h"
 #import <Foundation/Foundation.h>
 
+typedef NS_ENUM(NSUInteger, SentryPackageManagerOption) {
+    SentrySwiftPackage,
+    SentryCocoaPods,
+    SentryCarthage,
+    SentryNoPackage
```
philipphofmann, line 8: `l`: In that case, we don't know the package manager:
```suggestion
    SentryPackageManagerUnkown
```

Sources/Sentry/SentrySdkInfo.m
```diff
@@ -1,30 +1,43 @@
 #import "SentrySdkInfo.h"
 #import <Foundation/Foundation.h>
 
+typedef NS_ENUM(NSUInteger, SentryPackageManagerOption) {
+    SentrySwiftPackage,
```
philipphofmann, line 5: `l`: SPM is Swift Package Manager
```suggestion
    SentrySwiftPackageManager,
```

Sources/Sentry/SentrySdkInfo.m
```diff
@@ -58,9 +78,36 @@ - (instancetype)initWithDictInternal:(NSDictionary *)dict orDefaults:(SentrySdkI
...
-    return @{ @"sdk" : @ { @"name" : self.name, @"version" : self.version } };
+    NSMutableDictionary *sdk = @{
+        @"name" : self.name,
+        @"version" : self.version,
+    }
+                                   .mutableCopy;
+    if (self.packageManager != SentryNoPackage) {
+        sdk[@"packages"] = @{
+            @"name" :
+                [NSString stringWithFormat:[self getPackageName:self.packageManager], self.name],
```
philipphofmann, line 105: `h`: `getPackageName` can return `nil` and if it does, we crash here. So either check if it doesn't return `nil` before calling `stringWithFormat` or return a non nil value in `getPackageName`.

Tests/SentryTests/Protocol/SentrySdkInfoTests.swift
```diff
@@ -32,6 +33,31 @@ class SentrySdkInfoTests: XCTestCase {
...
+            XCTFail("Serialization of SdkInfo doesn't contain sdk")
+        }
+    }
+
+    func testPackageNames () {
+        let actual = SentrySdkInfo(name: sdkName, andVersion: "")
+        XCTAssertEqual(Dynamic(actual).getPackageName(0), "spm:getsentry/%@")
+        XCTAssertEqual(Dynamic(actual).getPackageName(1), "cocoapods:getsentry/%@")
+        XCTAssertEqual(Dynamic(actual).getPackageName(2), "carthage:getsentry/%@")
+        XCTAssertNil(Dynamic(actual).getPackageName(3).asString)
```
philipphofmann, line 59: `m`: That test doesn't validate that `"cocoapods:getsentry/%@"` and `"carthage:getsentry/%@"` are proper format strings. Instead of asserting their values, it would be better to test the format strings like `testSPM_packageInfo` does.

philipphofmann (commented):
Thanks for doing this. We're almost at the finish line.

Tests/SentryTests/Protocol/SentrySdkInfoTests.swift
brustolin: This test ensures that the format string don't change, the test before this one ensures that we use the format in the correct way.

philipphofmann: Yes, but if I now change `carthage:getsentry/%@` both here in the tests and in `SentrySDKInfo` to `carthage%@:getsentry/%@`, the test are green, but it should crash.

brustolin: If you change the implementation and the test, of course it would be green, same would happen for your suggestion.

But just to end this discussion Im gonna do it. This one is pretty much ctrl+c, ctrl+v.

philipphofmann (approved):
LGTM

PR #2781, opened 2023-03-13
fix: View hierarchy not sent for crashes

brustolin:
## :scroll: Description

View hierarchy was not being sent for crashes because of the lack of a filename.

Unit test.
Samples.

## :pencil: Checklist


Sources/Sentry/SentryViewHierarchyIntegration.m
```diff
@@ -13,7 +13,8 @@
 void
 saveViewHierarchy(const char *path)
```
armcknight, line 14: - [ ] `m`: change `path` to something more descriptive like `reportDirectory`
- [ ] `l`: add headerdoc above the signature. it's obvious here now, but if it gets refactored away may not be
```suggestion
/**
* Function to call through to the ObjC method to save a view hierarchy, which can be passed around as
* a function pointer in the C crash reporting code.
* @param reportDirectoryPath The path to the directory containing crash reporting files, in which a new
* file will be created to store the view hierarchy description.
*/
void
saveViewHierarchy(const char *reportDirectoryPath)
```

armcknight (commented):
Straightforward fix looks good, but I think the codepath could benefit from some better naming/docs:

- [ ] `m`: add to the headerdoc comment above `sentrycrash_setSaveViewHierarchy` in [SentryCrashC.h](https://github.com/getsentry/sentry-cocoa/blob/f7154992f8138ce55782b4c5701a21408543c815/Sources/SentryCrash/Recording/SentryCrashC.h#L125-L127) that specifies that the callback expects a path to the directory containing the crash report files
    - even better, typedef that callback type, and add a headerdoc to the typedef that explains this. That way you can add an `@param` directive. If you do this, please still mention how that param is used in `sentrycrash_setSaveViewHierarchy`
- [ ] `m`: add headerdoc above `saveViewHierarchy` declaration in [SentryViewHierarchy.h](https://github.com/getsentry/sentry-cocoa/blob/f7154992f8138ce55782b4c5701a21408543c815/Sources/Sentry/include/SentryViewHierarchy.h#L11) explaining what it does and what `filePath` is. It seems obvious in isolation but since the codepath traverses some functions that accept a parameter pointing to the _directory_ and this is the path to the _file_ in that directory, I think it helps to document everything to keep them straight.

Sources/Sentry/SentryViewHierarchyIntegration.m
philipphofmann: I think we should also update `saveScreenShot` with a similar comment.

Tests/SentryTests/Integrations/ViewHierarchy/SentryViewHierarchyIntegrationTests.swift
```diff
@@ -53,6 +53,12 @@ class SentryViewHierarchyIntegrationTests: XCTestCase {
         XCTAssertFalse(sentrycrash_hasSaveViewHierarchyCallback())
     }
 
+    func test_integrationAddFileName() {
+        SentrySDK.start { $0.attachViewHierarchy = true }
+        saveViewHierarchy("")
+        XCTAssertTrue(fixture.viewHierarchy.saveFilePathUsed?.hasSuffix("view-hierarchy.json") ?? false)
```
philipphofmann, line 59: `m`: Instead of asserting that the `saveFilePathUsed` has a suffix, I would prefer to validate that the view hierarchy gets stored to the path we passed in. So basically I would turn this into an integration test.

philipphofmann (commented):
Doesn't the same problem apply for `saveScreenShot`?

Tests/SentryTests/Integrations/ViewHierarchy/SentryViewHierarchyIntegrationTests.swift
brustolin: We have such a test for SentryViewHierarchy, here I just want to know if `saveViewHierarchy` is calling the proper function from DependencyContainers`s viewHierarchy.

brustolin:
> Doesn't the same problem apply for saveScreenShot?

No, saveScreenShot is working because SentryScreenShot append the filename for every screen

Sources/Sentry/SentryViewHierarchyIntegration.m
```diff
@@ -10,10 +10,17 @@
...
+* a function pointer in the C crash reporting code.
+* @param reportDirectoryPath The path to the directory containing crash reporting files, in which a new
+* file will be created to store the view hierarchy description.
+*/
 void
-saveViewHierarchy(const char *path)
+saveViewHierarchy(const char *reportDirectoryPath)
 {
-    NSString *reportPath = [NSString stringWithUTF8String:path];
+    NSString *reportPath = [[NSString stringWithUTF8String:path]
```
brustolin, line 22:
```suggestion
    NSString *reportPath = [[NSString stringWithUTF8String:reportDirectoryPath]
```

Tests/SentryTests/Integrations/ViewHierarchy/SentryViewHierarchyIntegrationTests.swift
philipphofmann: But then this test should need some updates because it's not testing the logic properly, as we changed the code but not this test.

Sources/Sentry/include/SentryScreenshot.h
```diff
@@ -13,7 +13,13 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (nullable NSArray<NSData *> *)appScreenshots;
 
-- (void)saveScreenShots:(NSString *)path;
+/**
+ * Save the current app screen shots in the given directory.
+ * If an app has more than one screen, one image for each screen will be saved.
+ *
+ * @param filesDirectory The path where the images should be saved.
+ */
+- (void)saveScreenShots:(NSString *)filesDirectory;
```
philipphofmann, line 22: `l`: I think `filesDirectory` is not much better than `path`. What about `imageDirectoryPath`?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- View hierarchy not sent for crashes (#2781)
```
philipphofmann, line 7: `m`: We released 8.3.2, so this needs some updates now.

philipphofmann (approved):
LGTM, but please update the changelog before merging.

PR #2782, opened 2023-03-13
ref: Remove unused zombie crash monitor

philipphofmann:
The zombie crash monitor is disabled by default and never enabled anywhere in the code.
https://github.com/getsentry/sentry-cocoa/blob/f7154992f8138ce55782b4c5701a21408543c815/Sources/SentryCrash/Recording/SentryCrash.m#L144
It's unused code and we can remove it.

#skip-changelog


brustolin:
Can you share a little bit more information?
What is this and why are we removing it?

philipphofmann:
> Can you share a little bit more information? What is this and why are we removing it?

I updated the description. Is it clearer now @brustolin?

brustolin (approved):

Issue #2779, opened 2023-03-12
Crash in 8.3.0 in +[SentryProfiler slicedSamples:transaction:]


eric:
### Platform

iOS

### Installed

CocoaPods

### Version

8.3.0

### Steps to Reproduce

1. Open app
2. Tap around
3. App Crashes

### Expected Result

App shouldn't crash.

### Actual Result

<img width="868" alt="screenshot-2023-03-11-20 17 47@2x" src="https://user-images.githubusercontent.com/470/224524031-bfd5176f-b89c-4a81-b933-5493d38a7b54.png">

brustolin:
Hello @eric 
Can you share the options you are using to start SentrySDK? I will try to reproduce this.

philipphofmann:
@armcknight, I think the profiler keeps adding to the samples array while slicing them. We could use locking on the data structures we iterate on or copy them. I vote for copying as locks will slow us down.

https://github.com/getsentry/sentry-cocoa/blob/f7154992f8138ce55782b4c5701a21408543c815/Sources/Sentry/SentryProfiler.mm#L387-L398

We should also copy the data for the payload in places like the following, as the profiler could keep adding data before the payload gets stored on disk, leading to inaccurate data. The downside is, of course, increased memory usage.

https://github.com/getsentry/sentry-cocoa/blob/f7154992f8138ce55782b4c5701a21408543c815/Sources/Sentry/SentryProfiler.mm#L404-L410

armcknight:
Correct about the sampling profiler being able to still mutate the data structure by adding to it while these enumeration blocks are running. Easiest to do the copy for now, you're correct we don't want to have that additional lock in sampling profiler.

armcknight:
We talked about trimming out any frames or stacks added after slicing the samples array, but deprioritized it as a small optimization that probably won't save us much, as those data structures are relatively small and compactible in compression. We can have another look though.

eric:
Did this impact anyone else? Were there other reports?

Having my exception reporting library cause my app to crash is a very bad experience. To me I consider it on the same level as a serious outage. Was this release pulled? Were customers notified of the issue?

Is there anything in place in the Sentry platform to track and identify potential issues in the exception reporting libraries so they're caught before they impact many customers?

eric:
Is that the SLA for the clients? Beta features may cause app crashes? If so, I would request a flag on the options when creating a SentrySDK instance to opt out of all beta features. 

[...]

PR #2786, opened 2023-03-13
fix: mutation while enumerating samples array

armcknight:
For https://github.com/getsentry/sentry-cocoa/issues/2779

We must synchronize access to reads and writes of the `samples` array, because even after making a copy to work with, there's still an `EXC_BREAKPOINT`. See thread sanitizer output:
```
WARNING: ThreadSanitizer: race on NSMutableArray (pid=81275)
  Modifying access of NSMutableArray at 0x00010ff62ca0 by thread T3:
    #0 -[__NSArrayM addObject:] <null>:51916644 (CoreFoundation:arm64+0x1065c)
    #1 processBacktrace(sentry::profiling::Backtrace const&, NSMutableDictionary<NSString*, NSMutableDictionary*>*, NSMutableDictionary<NSString*, NSDictionary*>*, NSMutableArray<SentrySample*>*, NSMutableArray<NSMutableArray<NSNumber*>*>*, NSMutableArray<NSDictionary<NSString*, objc_object*>*>*, NSMutableDictionary<NSString*, NSNumber*>*, NSMutableDictionary<NSString*, NSNumber*>*) SentryProfiler.mm:157 (Sentry:arm64+0x109a68)
    #2 __56-[SentryProfilerTests testProfilerMutationDuringSlicing]_block_invoke.102 SentryProfilerTests.mm:153 (SentryProfilerTests:arm64+0x1b43c)
    #3 __NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__ <null>:51916644 (Foundation:arm64+0x48a8e4)
    #4 _dispatch_client_callout <null>:51916644 (libdispatch.dylib:arm64+0x3764)

  Previous read-only access of NSMutableArray at 0x00010ff62ca0 by thread T2:
    #0 -[__NSArrayM copy] <null>:51916644 (CoreFoundation:arm64+0x1272c)
    #1 slicedProfileSamples(NSArray<SentrySample*>*, SentryTransaction*) SentryProfileTimeseries.mm:47 (Sentry:arm64+0x9750c)
    #2 __56-[SentryProfilerTests testProfilerMutationDuringSlicing]_block_invoke SentryProfilerTests.mm:127 (SentryProfilerTests:arm64+0x1af84)
    #3 __NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__ <null>:51916644 (Foundation:arm64+0x48a8e4)
    #4 _dispatch_client_callout <null>:51916644 (libdispatch.dylib:arm64+0x3764)

  Location is heap block of size 48 at 0x00010ff62ca0 allocated by main thread:
    #0 calloc <null>:51916644 (libclang_rt.tsan_iossim_dynamic.dylib:arm64+0x55230)
    #1 class_createInstance <null>:51916644 (libobjc.A.dylib:arm64+0x1a5b8)
    #2 __invoking___ <null>:51916644 (CoreFoundation:arm64+0x127cfc)

  Thread T3 (tid=5869034, running) is a GCD worker thread

  Thread T2 (tid=5869033, running) is a GCD worker thread

SUMMARY: ThreadSanitizer: race on NSMutableArray (CoreFoundation:arm64+0x1065c) in -[__NSArrayM addObject:]+0x5c
```

Wrote a new test that reproduced the crash by building up a large set of profiling data structures, and then concurrently querying it while continuing to modify it.

Also ran tests with thread sanitizer, which helped catch a potential extra data race in the fast enumeration blocks, which are now performed concurrently to help speed them up, since they'll now be blocking downstream backtrace gathering since we're synchronizing the two operations.


armcknight:
OK, I actually had a separate branch refactoring profiler to pull this out into a separate component, so i might just do that now, then see if I can exercise it in this way.

philipphofmann:
To reproduce this in a test, I think you need to let the profiler run for a while, so the samples array is big enough that iterating over it takes enough time for the profiler to add new sample entries. I don't think that's the best way to test this, though.

To make it testable I think we need to extract some code from `createProfilingEnvelopeItemForTransaction` and or `slicedSamples:transaction` so we can emulate the scenario leading to the crash. The easiest way IMO would be to extract `slicedSamples` or make it public so you can directly call it from tests.

philipphofmann:
> OK, I actually had a separate branch refactoring profiler to pull this out into a separate component, so i might just do that now, then see if I can exercise it in this way.

Basically, what I wrote in https://github.com/getsentry/sentry-cocoa/pull/2786#issuecomment-1466594076 😄. It's great to see we have the same thoughts independently.

Sources/Sentry/Profiling/SentryProfileTimeseries.mm
```diff
@@ -0,0 +1,92 @@
...
+        firstSampleRelativeToTransactionStart, lastSampleRelativeToTransactionStart);
+}
+
+NSArray<SentrySample *> *_Nullable slicedProfileSamples(
+    NSArray<SentrySample *> *samples, SentryTransaction *transaction)
+{
+    NSArray<SentrySample *> *samplesCopy;
+    {
+        std::lock_guard<std::mutex> l(_gSamplesArrayLock);
+        samplesCopy = [samples copy];
```
philipphofmann: `h`: This fixes the problem, but adding a similar bug in the future is likely. I think we might still have a bug in `serializedSamplesWithRelativeTimestamps` as it also iterates the samples array, but it doesn't use a copy or a sync mechanism.

I would prefer moving this back to the profiler to have all locks in one file, or we write a class that hides the samples array and ensures thread-safe access.

Tests/SentryProfilerTests/SentryProfilerTests.mm
```diff
@@ -52,6 +55,119 @@ - (void)testProfilerCanBeInitializedOffMainThread
                                  handler:^(NSError *_Nullable error) { NSLog(@"%@", error); }];
 }
 
+- (void)testProfilerMutationDuringSlicing
```
philipphofmann, line 58: Great test case 👏

philipphofmann (commented):
Something's off for me in Xcode. It doesn't find these files.

Screenshot 2023-03-14 at 12 04 51

Sources/Sentry/Profiling/SentryProfileTimeseries.mm
philipphofmann: Okay, then we can refactor this later if needed. Most importantly, we ship a fix quickly.

philipphofmann (approved):
After you address https://github.com/getsentry/sentry-cocoa/pull/2786#pullrequestreview-1338997261 and https://github.com/getsentry/sentry-cocoa/pull/2786#discussion_r1135361146, LGTM. I don't want to block this until the next day.

Sources/Sentry/Profiling/SentryProfileTimeseries.mm
```diff
@@ -0,0 +1,92 @@
...
+ * Print a debug log to help diagnose slicing errors.
+ * @param start @c YES if this is an attempt to find the start of the sliced data based on the
+ * transaction start; @c NO if it's trying to find the end of the sliced data based on the
+ * transaction's end, to accurately describe what's happening in the log statement.
+ */
+void
+logSlicingFailureWithArray(
+    NSArray<SentrySample *> *array, SentryTransaction *transaction, BOOL start)
+{
+    if (!SENTRY_CASSERT(array.count > 0, @"Should not have attempted to slice an empty array.")) {
```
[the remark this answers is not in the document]
armcknight: I changed this function to C-style, so it must use NSCAssert.

philipphofmann (approved):
Again LGTM

armcknight:
The codecov annotations are all about the early returns in the slicing function, which I previously promised to write tests for and still plan to, but they aren't pertinent to this fix, so I'm gonna kick that can a little further down the road.

Issue #2791, opened 2023-03-14
sentry-cocoa is causing "App Hanging"


eric:
### Platform

iOS

### Installed

CocoaPods

### Version

8.2.0

### Steps to Reproduce

1. Use the app
2. Wait for App Hang

### Expected Result

Not to hang

### Actual Result

<img width="984" alt="screenshot-2023-03-13-18 27 13@2x" src="https://user-images.githubusercontent.com/470/224868738-9527f2bf-1fb9-407c-bd0c-d06ba0e160b5.png">

brustolin:
Hello @eric, with only the stacktrace is hard to tell what's going on.
What is the error message of this event?
Do you mind sharing the json of this event?

philipphofmann:
Thanks for reporting this, @eric. How often do you see this?

It could be that the Sentry SDK is not causing the app hangs if it's actually an app hang, but our logic might fire too late, and the SDK could attach a stacktrace that doesn't map to the root cause of the app hang, see https://github.com/getsentry/sentry-cocoa/discussions/2715#discussioncomment-5091072.

Either we improve our app hang logic to collect the correct stack or change it to only report app hangs when we can collect the proper stack trace.

philipphofmann:
We moved updating the app state to a background thread with https://github.com/getsentry/sentry-cocoa/pull/2793, so the SDK won't report AppHangs with the stack trace pointing to the sentry-cocoa SDK as you pointed out above, @eric. Before closing this issue I would like to ensure that our SDK is not causing app hangs, but instead attaching wrong stack traces to app hangs.

eric:
> How often do you see this?

It's hard to say how often, because these traces have been grouped in an "App Hanging" issue that contains 120 events all of which contain no app stack frames beyond `main()`. I saw it at least a few times when I was looking.

PR #2793, opened 2023-03-14
fix: Updating AppHang state on main thread

philipphofmann:
## :scroll: Description

Move updating the app hang state to a background thread for anrStopped.

It came up while investigating https://github.com/getsentry/sentry-cocoa/issues/2791

Unit tests and on a real device.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryANRTracker.m
+++ b/Sources/Sentry/SentryANRTracker.m
@@ -89,7 +89,13 @@ SentryANRTracker ()
 
             if (reported) {
                 SENTRY_LOG_WARN(@"ANR stopped.");
-                [self ANRStopped];
+
+                // The ANR stopped, don't block the main thread with calling ANRStopped listeners.
+                // While the ANR code reports an ANR and collects the stack trace, the ANR might
+                // stop simultaneously. In that case, the ANRs stack trace would contain the
+                // following code running on the main thread. To avoid this, we offload work to a
+                // background thread.
+                [self.dispatchQueueWrapper dispatchAsyncWithBlock:^{ [self ANRStopped]; }];
             }
 
             reported = NO;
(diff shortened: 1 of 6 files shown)


Sources/Sentry/SentryANRTracker.m
```diff
@@ -89,7 +89,13 @@ - (void)detectANRs
...
             if (reported) {
                 SENTRY_LOG_WARN(@"ANR stopped.");
-                [self ANRStopped];
+
+                // The ANR stopped, don't block the main thread with calling ANRStopped listeners.
+                // While the ANR code reports an ANR and collects the stack trace, the ANR might
+                // stop simultaneously. In that case, the ANRs stack trace would contain the
+                // following code running on the main thread. To avoid this, we offload work to a
+                // background thread.
+                [self.dispatchQueueWrapper dispatchAsyncWithBlock:^{ [self ANRStopped]; }];
```
armcknight, line 98: This will also delay the suspension of threads to gather stacktraces. It might be better to separate the listeners and to collect the stack traces immediately, and only report the app state update from an async context.

armcknight (commented):
I'm just not sure about whether we want to delay gathering stack traces. It might remove our own code frames, but might make the feature less useful in general.

Sources/Sentry/SentryANRTracker.m
brustolin: We don't getter stacktraces when reporting the the app hanging stopped

brustolin (approved):

Sources/Sentry/SentryANRTracker.m
philipphofmann: Exactly what @brustolin said, but thanks for raising the concern, @armcknight.

PR #2798, opened 2023-03-14
output xcodebuild logs from carthage build and archive if failed

armcknight:
I had the [xcframework validation build fail](https://github.com/getsentry/sentry-cocoa/actions/runs/4418446156/jobs/7745583784#step:4:13) but with no clues as to why. This tries to capture the build output.

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,6 +104,14 @@ jobs:
           path: |
             ${{ github.workspace }}/*.zip
 
+      - name: Archive build log if failed
+        uses: actions/upload-artifact@v3
+        if: ${{  failure() || cancelled() }}
+        with:
+          name: raw-build-output-build-xcframework
+          path: |
+            build-xcframework.log
+
   validate-xcframework:
     name: Validate XCFramework Xcode ${{ matrix.xcode }}
     runs-on: macos-12
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ analyze:
 # For more info check out: https://github.com/Carthage/Carthage/releases/tag/0.38.0
 build-xcframework:
 	@echo "--> Carthage: creating Sentry xcframework"
-	carthage build --use-xcframeworks --no-skip-current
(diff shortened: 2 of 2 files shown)


brustolin (approved):

Sources/Sentry/SentryProfiler.mm
```diff
@@ -484,6 +484,7 @@ + (void)useFramesTracker:(SentryFramesTracker *)framesTracker
 + (nullable NSArray<SentrySample *> *)slicedSamples:(NSArray<SentrySample *> *)samples
                                         transaction:(SentryTransaction *)transaction
 {
+    [[SentryViewHierarchy alloc] init];
```
brustolin, line 487: h: I guess you forgot something here
```suggestion
```

armcknight: it's really annoying me that changes in a file under `Sources/` isn't triggering a new CI run 😡

philipphofmann: > it's really annoying me that changes in a file under Sources/ isn't triggering a new CI run 😡

@armcknight, the tests should run https://github.com/getsentry/sentry-cocoa/blob/230ba67c4f59d08a881b0fc0c791ed90b5de1419/.github/workflows/test.yml#L10

So does build https://github.com/getsentry/sentry-cocoa/blob/230ba67c4f59d08a881b0fc0c791ed90b5de1419/.github/workflows/build.yml#L10

Which parts of CI are not running? You can adjust this in the workflow files easily.

armcknight: aha: 
> If you use both the branches filter and the paths filter, the workflow will only run when both filters are satisfied

armcknight:
Instead of trying to test it let's just see if it helps/hurts going forward.

Sources/Sentry/SentryProfiler.mm
armcknight: So I just discovered that checks won't run if there is a merge conflict on the branch, that must've been what was going on: https://github.com/orgs/community/discussions/26304

PR #2811, opened 2023-03-17
fix: App Hang report crashes with too many threads

### LINKED ISSUE #2794: Crash in `getCurrentThreadsWithStackTrace`
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.2.0

### Steps to Reproduce

Since I upgraded to 8.x and enabled the "App Hanging" feature, I'm some crashes in `getCurrentThreadsWithStackTrace` at this [line](https://github.com/getsentry/sentry-cocoa/blob/f7154992f8138ce55782b4c5701a21408543c815/Sources/Sentry/SentryThreadInspector.m#L132)

The crash is happening on both iOS and macOS. I was unable to reproduce it on my end, therefore unfortunately can't give ways to reproduce.

Here is the relevant part of the stack trace:
```
OS Version: iOS 16.2 (20C65)
Report Version: 104

Exception Type: EXC_BAD_ACCESS (SIGBUS)
Exception Codes: BUS_NOOP at 0x000000016d3bc958
Crashed Thread: 5

Application Specific Information:
<redacted> > kevent_id >
KERN_PROTECTION_FAILURE at 0x16d3bc958.

Thread 5 Crashed:
0   Structured                      0x2034bd388         -[SentryThreadInspector getCurrentThreadsWithStackTrace] (SentryThreadInspector.m:132)
1   Structured                      0x203470448         -[SentryANRTrackingIntegration anrDetected] (SentryANRTrackingIntegration.m:64)
2   Structured                      0x20346fbd8         -[SentryANRTracker ANRDetected] (SentryANRTracker.m:139)
3   Structured                      0x20346f69c         -[SentryANRTracker detectANRs] (SentryANRTracker.m:121)
4   Foundation                      0x33cfe55a4         __NSThread__start__
5   libsystem_pthread.dylib         0x3e49b36c8         _pthread_start
```

### Expected Result

No crash

### Actual Result

Crash

_No response_

brustolin:
## :scroll: Description

Add a limit of 70 threads to report App hang. 
More than 100 is causing the SDK to crash, we add an extra safety margin.

closes #2794 

Unit tests

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryThreadInspector.m
+++ b/Sources/Sentry/SentryThreadInspector.m
@@ -119,14 +119,15 @@ getStackEntriesFromThread(SentryCrashThread thread, struct SentryCrashMachineCon
         thread_act_array_t suspendedThreads = NULL;
         mach_msg_type_number_t numSuspendedThreads = 0;
 
-        //We add a limit of 70 threads because in every test up to a 100 seems fine.
-        //We are giving it an extra safety margin.
-        sentrycrashmc_suspendEnvironment_upToMaxSupportedThreads(&suspendedThreads, &numSuspendedThreads, 70);
+        // We add a limit of 70 threads because in every test up to a 100 seems fine.
+        // We are giving it an extra safety margin.
+        sentrycrashmc_suspendEnvironment_upToMaxSupportedThreads(
+            &suspendedThreads, &numSuspendedThreads, 70);
         // DANGER: Do not try to allocate memory in the heap or call Objective-C code in this
         // section Doing so when the threads are suspended may lead to deadlocks or crashes.
 
-        //If no threads was suspended we don't need to do anything.
-        //This may happen if there is more than max amount of threads (70).
+        // If no threads was suspended we don't need to do anything.
+        // This may happen if there is more than max amount of threads (70).
         if (numSuspendedThreads == 0) {
             return threads;
         }
--- a/Sources/Sentry/SentryThreadInspector.m
+++ b/Sources/Sentry/SentryThreadInspector.m
@@ -119,7 +119,8 @@ getStackEntriesFromThread(SentryCrashThread thread, struct SentryCrashMachineCon
         thread_act_array_t suspendedThreads = NULL;
         mach_msg_type_number_t numSuspendedThreads = 0;
 
-        // We add a limit of 70 threads because in every test up to a 100 seems fine.
+        // SentryThreadInspector is crashing when there is too many threads.
+        // We add a limit of 70 threads because in test with up to 100 threads it seems fine.
         // We are giving it an extra safety margin.
         sentrycrashmc_suspendEnvironment_upToMaxSupportedThreads(
             &suspendedThreads, &numSuspendedThreads, 70);
--- a/Sources/Sentry/SentryThreadInspector.m
+++ b/Sources/Sentry/SentryThreadInspector.m
@@ -119,9 +119,15 @@ getStackEntriesFromThread(SentryCrashThread thread, struct SentryCrashMachineCon
         thread_act_array_t suspendedThreads = NULL;
         mach_msg_type_number_t numSuspendedThreads = 0;
 
-        sentrycrashmc_suspendEnvironment(&suspendedThreads, &numSuspendedThreads);
+        sentrycrashmc_suspendEnvironment_upToMaxSupportedThreads(&suspendedThreads, &numSuspendedThreads, 70);
         // DANGER: Do not try to allocate memory in the heap or call Objective-C code in this
         // section Doing so when the threads are suspended may lead to deadlocks or crashes.
+
+        //If no threads was suspended we don't need to do anything.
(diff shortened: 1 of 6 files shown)


Tests/SentryTests/SentryCrash/SentryThreadInspectorTests.swift
```diff
@@ -80,6 +81,27 @@ class SentryThreadInspectorTests: XCTestCase {
...
+        for _ in 0..<expect.expectedFulfillmentCount {
+            Thread.detachNewThread {
+                expect.fulfill()
+                while self.fixture.keepThreadAlive {
+                    Thread.sleep(forTimeInterval: 0.001)
+                }
+            }
+        }
+
+        wait(for: [expect], timeout: 1)
```
philipphofmann, line 99: `m`: CI could be a bit slow sometimes to spawn 70 threads. I think we could be more generous with the timeout to avoid flakiness.
```suggestion
        wait(for: [expect], timeout: 5)
```

Tests/SentryTests/SentryCrash/SentryThreadInspectorTests.swift
```diff
@@ -80,6 +81,27 @@ class SentryThreadInspectorTests: XCTestCase {
...
+
+        let sut = self.fixture.getSut(testWithRealMachineContextWrapper: true)
+
+        for _ in 0..<expect.expectedFulfillmentCount {
+            Thread.detachNewThread {
+                expect.fulfill()
+                while self.fixture.keepThreadAlive {
+                    Thread.sleep(forTimeInterval: 0.001)
+                }
+            }
```
philipphofmann, line 96: TIL; fancy way of starting multiple threads 👍

philipphofmann (commented):
I would feel more comfortable with this change after adding a test case in one of our sample apps that launches over 100 threads and then crashes. Then we should test on an iPhone if the SDK can still report 100 threads when crashing.

Tests/SentryTests/SentryCrash/SentryThreadInspectorTests.swift
```diff
@@ -5,6 +5,7 @@ class SentryThreadInspectorTests: XCTestCase {
     private class Fixture {
         var testMachineContextWrapper = TestMachineContextWrapper()
         var stacktraceBuilder = TestSentryStacktraceBuilder(crashStackEntryMapper: SentryCrashStackEntryMapper(inAppLogic: SentryInAppLogic(inAppIncludes: [], inAppExcludes: [])))
+        var keepThreadAlive = true
```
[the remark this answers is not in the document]
brustolin: `Thread.detachNewThread` is `@sendable` therefore it does not allow to change the value of a local variable after using it in the block. This is the less ugly workaround I could think of.

Sources/Sentry/SentryThreadInspector.m
```diff
@@ -119,10 +119,19 @@ - (SentryStacktrace *)stacktraceForCurrentThreadAsyncUnsafe
         thread_act_array_t suspendedThreads = NULL;
         mach_msg_type_number_t numSuspendedThreads = 0;
 
-        sentrycrashmc_suspendEnvironment(&suspendedThreads, &numSuspendedThreads);
+        // We add a limit of 70 threads because in every test up to a 100 seems fine.
+        // We are giving it an extra safety margin.
```
brustolin, line 123:
```suggestion
        // SentryThreadInspector is crashing when there is too many threads.
        // We add a limit of 70 threads because in test with up to 100 threads it seems fine.
        // We are giving it an extra safety margin.
```

Tests/SentryTests/SentryCrash/SentryThreadInspectorTests.swift
brustolin: DispatchQueue has a managing system that don't allow to run too many threads in parallel.

Sources/Sentry/SentryThreadInspector.m
```diff
@@ -119,10 +119,20 @@ - (SentryStacktrace *)stacktraceForCurrentThreadAsyncUnsafe
         thread_act_array_t suspendedThreads = NULL;
         mach_msg_type_number_t numSuspendedThreads = 0;
 
-        sentrycrashmc_suspendEnvironment(&suspendedThreads, &numSuspendedThreads);
+        // SentryThreadInspector is crashing when there is too many threads.
```
philipphofmann, line 122: Thanks 👏

brustolin:
> I would feel more comfortable with this change after adding a test case in one of our sample apps that launches over 100 threads and then crashes. Then we should test on an iPhone if the SDK can still report 100 threads when crashing.

Yeap, still working.

image

philipphofmann:
@brustolin, do you mind adding the code that launches 100 threads to one of our sample apps? I think we are going to need that in the future. You can also do this in an extra PR.

philipphofmann (approved):
LGTM

PR #2819, opened 2023-03-21
accept an autoformat that precommit keeps forcing, preventing commits…

armcknight:
…; unformatted code committed in #2782

#skip-changelog

### DIFF
--- a/Sources/SentryCrash/Recording/SentryCrash.m
+++ b/Sources/SentryCrash/Recording/SentryCrash.m
@@ -377,7 +377,10 @@ getBasePath()
 // ============================================================================
 
 #define SYNTHESIZE_CRASH_STATE_PROPERTY(TYPE, NAME)                                                \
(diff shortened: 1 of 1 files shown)


brustolin:
Something went wrong. No files to review

armcknight:
The autoformat commit changed it back. Wonder if the runs in pre-commit and the github action are conflicting?

philipphofmann:
> The autoformat commit changed it back. Wonder if the runs in pre-commit and the github action are conflicting?

Sometimes clang format get's an update in CI and then your locale version and the one in CI format the code differently. Maybe `brew upgrade clang-format` helps.

armcknight:
Ah, good call @philipphofmann . I'll close this and revisit if it's a problem after upgrading clang-format locally. It bumped major versions fwiw from 15.x.x to 16.

PR #2821, opened 2023-03-21
ref: Use SentryTracerConfiguration to initialize SentryTracer

brustolin:
Instead of having multiple functions to initialize SentryTracer with different configurations we should we one complex object that contains all the configurations.

_#skip-changelog_


philipphofmann (commented):
I like the idea of using a builder pattern for the initialization of the tracer, as it's getting more and more complex. In languages like Swift, Java, C# the builder pattern would make sense, but for Objective-C, this doesn't work like that. I like the approach in this [article](https://annema.me/2014/04/04/the-builder-pattern-in-objective-c.html). Instead of the `SentryTracerConfiguration`, we could use a normal Objective-C class, which you then pass to the tracer on init. I wouldn't copy every parameter as the downside is that when adding a new parameter you need to change it in two locations. The advantage of an Objective-C class is that we can have default values for parameters, and create a constructor with the required parameters.

I would do this refactoring in an extra PR to merge it quickly and not in the INP PR.

philipphofmann (commented):
I like the direction. It's definitely an improvement.

Sources/Sentry/include/SentryTracer.h
```diff
@@ -70,55 +64,17 @@ static NSTimeInterval const SentryTracerDefaultTimeout = 3.0;
...
  *
  * @return SentryTracer
  */
 - (instancetype)initWithTransactionContext:(SentryTransactionContext *)transactionContext
                                        hub:(nullable SentryHub *)hub
-                   profilesSamplerDecision:
-                       (nullable SentryProfilesSamplerDecision *)profilesSamplerDecision
-                               idleTimeout:(NSTimeInterval)idleTimeout
-                      dispatchQueueWrapper:(SentryDispatchQueueWrapper *)dispatchQueueWrapper;
+                             configuration:(nullable SentryTracerConfiguration *)configuration;
```
armcknight, line 77: I like this refactor, but my one suggestion is to make this nonnull and provide a `SentryTracerConfiguration.default` that can be supplied. That makes it more obvious at callsites of this constructor what is going on, rather than having to go to the implementation of this init to see what happens if `nil` is passed in. This is also a common pattern throughout Cocoa APIs.

PR #2826, opened 2023-03-22
test: fix raw log output ids for ui tests

armcknight:
This was producing raw output logs from ui test runs that overwrote each other because they collided when no differentiating pieces existed to interpolate: 
image

#skip-changelog


philipphofmann (approved):
Good catch, thanks 🙏

PR #2827, opened 2023-03-22
test: highlight buttons that are tapped so it's more obvious what's happening in UI test run videos

armcknight:
I was trying to debug some UI test failures in sauce labs and couldn't tell what was happening in the videos.

#skip-changelog

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift/ViewController.swift
+++ b/Samples/iOS-Swift/iOS-Swift/ViewController.swift
@@ -70,21 +70,24 @@ class ViewController: UIViewController {
         }
     }
     
-    @IBAction func addBreadcrumb(_ sender: Any) {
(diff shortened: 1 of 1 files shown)


brustolin (approved):
Good Idea. 
I  wish we could find a more universal solution that would work for every button without the need to call highlightButton every time.

philipphofmann:
@armcknight, I prefer enabling this option via an environment variable or something only for the UI tests. Users often check this sample app when trying Sentry for the first time. Having every button highlighted blue kind of looks weird to me.

armcknight:
Maybe I will convert everything to table views in a later refactor as mentioned in https://github.com/getsentry/sentry-cocoa/pull/2828#issuecomment-1481931624 so that we get the default highlight behavior.

With that, I'd be fine to just close this PR.

PR #2828, opened 2023-03-23
test: make all buttons tappable

armcknight:
We've added enough buttons to the sample app now that on small devices like the iphone 8, they aren't all tappable.

I made a couple stopgap changes to fix this:
- add a third column of buttons
- make all button text smaller
- pin all edges of outermost stack view to safe area layout margins, to keep it from going up underneath the navbar (this was the actual issue, it was obscuring a button a ui test was trying to tap, thereby failing the test)

Long term we need to reorganize this. Either multiple vcs with related options that can be pushed/popped on the nav controller (my preference), or a long table view or embed the stack view in a scroll view.

#skip-changelog


philipphofmann:
I agree with @brustolin that the small font size is not the best idea, as I think tapping the buttons can get complicated, but a scroll view might also not be the best for UI tests. We could also split the screen into multiple ones and add some navigation.

armcknight:
I agree a scroll view will be complicated, that would be my last choice. 

Since this problem blocks #2823, would you be ok merging this for now and then circling back within a couple days for a more robust solution? I'm happy to break it up into screens with navigation, I can take that on in a few days. We can have it in by the middle of next week.

It will just take more significant refactoring of logic in ViewController.swift and of UI tests, which is why I hoped to just get by with the smaller buttons for the time being.

brustolin:
I dont mind. Sample should not be a blocker for the SDK.

Btw, is a good opportunity to introduce UITabViewController in the sample. 
Organize the actions by feature.

brustolin (approved):

PR #2804, opened 2023-03-16
docs: call out beta status for all public-facing profiling APIs

armcknight:
Add notes and beta warnings to all public-facing APIs for profiling.

#skip-changelog

PR #2829, opened 2023-03-23
docs: clean up headerdocs

armcknight:
After making some of the changes in #2804 , I noticed some other irregularities in nearby headerdocs. 

- set aside all declarations of default values in separate `@note` entries
- add `@c` to sentry and builtin declared symbols in headerdocs and other tokens that are better viewed in monospace, 
    - like values that can be supplied to a documented method: 0 is easy to mistake as O in variable-width fonts, but `0` and `O` are better because of the slash through the monospace `0`. 
    - Also, tokens containing underscores would cause all subsequent text to be rendered as italic due to the markdown rendering, which doesn't happen if it is preceded by a `@c`
- some light editing of typos, whitespace and extra newlines; reworded some phrasing to be more clear
- removed some, but probably not all, superfluous headerdocs. things like a doc over an init method for class Foo that states `@returns Foo` ([example](https://github.com/getsentry/sentry-cocoa/pull/2829/files#diff-b81825fe615ea83315bdeecc2430aa95f71375710fb503a2faffa7147798ca6dL43-L47)). I'd rather just save that line and vertical space.
- standardize on `@c` for monospaced elements in headerdocs. This gives it treatment in the source file and in doc popovers. 
	- while backticks do make it monospaced in source files, it doesn't in doc popovers
		- before: 
			Screenshot 2023-03-22 at 5 41 00 PM
		- after:
			Screenshot 2023-03-22 at 5 41 23 PM
    - `<code></code` doesn't appear to do anything:
    	- before:
			Screenshot 2023-03-22 at 5 44 19 PM
			Screenshot 2023-03-22 at 5 44 22 PM
    	- after:
			Screenshot 2023-03-22 at 5 44 32 PM
			Screenshot 2023-03-22 at 5 44 39 PM

#skip-changelog

### DIFF
--- a/Sources/Sentry/Public/SentryAttachment.h
+++ b/Sources/Sentry/Public/SentryAttachment.h
@@ -11,8 +11,7 @@ NS_SWIFT_NAME(Attachment)
 SENTRY_NO_INIT
 
 /**
- * Initializes an attachment with data. Sets the content type to "application/octet-stream".
- *
+ * Initializes an attachment with data. Sets the content type to @c "application/octet-stream".
  * @param data The data for the attachment.
  * @param filename The name of the attachment to display in Sentry.
  */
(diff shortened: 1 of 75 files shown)


armcknight:
I just noticed the new "Build Documentation" feature in Xcode, and realize now that instead of using `@c` it should be double backticks, which creates hyperlinks in doc popover windows as well as the documentation organizer. That should be a fairly straightforward search/replace.

philipphofmann:
> Hmm, I just noticed the Build Documentation feature in Xcode, which `@c` doesn't play well with. And while I see some cross-reference links created for tokens surrounded by double-backticks, that doesn't work perfectly. Since we didn't really use that feature, I think we should stick with `@c` for now.

We can revisit this once we start publishing code docs.

philipphofmann (approved):
Many thanks for doing this 👏 💯 🥇 

Please merge this quickly to avoid conflicts. LGTM 🌟

armcknight:
@philipphofmann This was the first time I encountered the high risk file check, note that I added the new sha to its output to tighten up the round trip in that workflow a bit: https://github.com/getsentry/sentry-cocoa/pull/2829/files#diff-9219d0ee596a5725e4e618eb75d7fb9497df6ad02b984e46b857855d03161922R13

PR #2833, opened 2023-03-24
ref: Use dispatch queue in dependency container

philipphofmann:
Avoid allocating one extra dispatch queue in the dependency container.

#skip-changelog


Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -8,7 +8,6 @@
 #import <SentryDebugImageProvider.h>
 #import <SentryDefaultCurrentDateProvider.h>
 #import <SentryDependencyContainer.h>
-#import <SentryDispatchQueueWrapper.h>
```
philipphofmann, line 11: We imported this header twice.

brustolin (approved):

Issue #2832, opened 2023-03-24
We are facing a lot of SDK crashes that have occurred after upgrading to version 8.x.x.


zhgchgli0718:
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.3.1

### Steps to Reproduce

unknown, seem to be random crashes.

### Expected Result

The app was stable before we updated to version 7.3.1, but we're experiencing issues now.

Our current crash-free rate is around `98.X%`, which is lower than our previous rate of `99.98%` before we upgraded the SDK version. :(

### Actual Result

We have received a lot of fatal issues as shown below, but unfortunately, the crash log doesn't provide any useful information, and we have no idea how to fix it:

<img width="759" alt="image" src="https://user-images.githubusercontent.com/33706588/227419180-75300fe5-565b-46e2-8e52-bdc42cedd22a.png">

### EXC_BAD_ACCESS
Exception 1, Code 1, Subcode 0 > Attempted to dereference null pointer.
<img width="833" alt="image" src="https://user-images.githubusercontent.com/33706588/227418832-544b34db-df4d-40f2-8e3a-c3813aadb797.png">

### EXC_BAD_ACCESS -> The most one
3c8a24d6f isn't waiting on children and needs idle timeout dispatched. >
KERN_INVALID_ADDRESS at 0xc41c9bf20.
<img width="836" alt="image" src="https://user-images.githubusercontent.com/33706588/227418705-ddc7bf62-672c-4d81-a0b7-4d47ef9cae1d.png">

---

- While we can't be sure that the Sentry SDK is responsible for all of our crashes, we have received many similar crashes since upgrading to version 8.x.x.
- We noticed that upgrading to the latest SDK version helped to reduce the crash rate
- We also found an interesting message in the crash logs that said 'isn't waiting on children and needs idle timeout dispatched' this messsage which is actually came from the Sentry SDK

As mentioned above, we believe that this message is related to the SDK and may be the source of the problem, If there is anything we can do to help, please let us know.

### Are you willing to submit a PR?

yes

philipphofmann:
This could be related to https://github.com/getsentry/sentry-cocoa/issues/2769.

Issue #2769, opened 2023-03-08
8.2.0 still crash


shenyinghao:
### Platform

iOS

### Installed

CocoaPods

### Version

8.2.0

### Steps to Reproduce

libobjc.A.dylib
objc_release_x0 + 16
1
libobjc.A.dylib
objc_release + 16
2

SentryDispatchQueueWrapper.m - 第 68 行
-[SentryDispatchQueueWrapper dispatchAfter:block:] + 68
3

SentryTracer.m - 第 223 行
-[SentryTracer dispatchIdleTimeout] + 223
4

SentryTracer.m - 第 354 行
-[SentryTracer spanFinished:] + 354
5

SentrySpan.m - 第 157 行
-[SentrySpan finishWithStatus:] + 157
6

SentryNetworkTracker.m - 第 290 行
-[SentryNetworkTracker urlSessionTask:setState:] + 290
7

SentryNetworkTrackingIntegration.m - 第 72 行
__57+[SentryNetworkTrackingIntegration swizzleURLSessionTask]_block_invoke_2.27 + 72

### Expected Result

no crash

### Actual Result

There are many crashes. Please solve this problem as soon as possible. Currently, the crashes caused by sentry SDK account for 90%

brustolin:
Hello @shenyinghao.  Do you have a reliable way to reproduce this error? Can you share you us some code?

shenyinghao:
> Hello @shenyinghao. Do you have a reliable way to reproduce this error? Can you share you us some code?

This is an online problem. Only online crash logs can be provided. There is no recurrence during the development process, but there are many online crashes. Please help me check it

brustolin:
Thanks @shenyinghao.
Does Sentry captures this crash?
If so, can you send the event json to dhiogo.brustolin (at) sentry.io it would help a lot.

shenyinghao:
0
libobjc.A.dylib
objc_retain + 16
1

SentryDispatchQueueWrapper.m - 第 59 行
-[SentryDispatchQueueWrapper dispatchAfter:block:] + 59
2

SentryTracer.m - 第 223 行
-[SentryTracer dispatchIdleTimeout] + 223
3

SentryTracer.m - 第 354 行
-[SentryTracer spanFinished:] + 354
4

SentrySpan.m - 第 157 行
-[SentrySpan finishWithStatus:] + 157
5

SentryNSDataTracker.m - 第 257 行
[...]

PR #2834, opened 2023-03-24
fix: Crash in Tracer for idle timeout

philipphofmann:
## :scroll: Description

dispatch_block_create can return NULL. If it does, we finish the transaction and don't schedule the timeout.

Fixes GH-2832, GH-2769

Unit tests, and on an iPhone to validate UI event transactions are still working properly.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryTracer.m
+++ b/Sources/Sentry/SentryTracer.m
@@ -227,6 +227,7 @@ static BOOL appStartMeasurementRead;
     if (_idleTimeoutBlock == NULL) {
         SENTRY_LOG_WARN(@"Couln't create idle time out block. Can't schedule idle timeout. "
                         @"Finishing transaction");
+        // If the transaction has no children, the SDK will discard it.
         [self finishInternal];
     } else {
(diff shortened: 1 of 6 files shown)


Sources/Sentry/SentryTracer.m
```diff
@@ -217,14 +217,20 @@ - (void)dispatchIdleTimeout
...
             return;
         }
         [weakSelf finishInternal];
-    });
-    [self.dispatchQueueWrapper dispatchAfter:self.idleTimeout block:_idleTimeoutBlock];
+    }];
+    if (_idleTimeoutBlock == NULL) {
+        SENTRY_LOG_WARN(@"Couln't create idle time out block. Can't schedule idle timeout. "
+                        @"Finishing transaction");
+        [self finishInternal];
```
brustolin, line 230: m: I would just add a comment here that the transaction will not be send because it does not have any child.

brustolin (approved):
Looks good.

Just one suggestion.

Sources/Sentry/SentryTracer.m
philipphofmann: Better now?

PR #2836, opened 2023-03-27
ref: Remove not needed check for SentryTracer

philipphofmann:
Remove not needed check for isKindOfClass SentryTracer in SentryPerformanceTracker as the code calls startTransactionWithContext, which always returns a SentryTracer.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryPerformanceTracker.m
+++ b/Sources/Sentry/SentryPerformanceTracker.m
@@ -85,9 +85,7 @@ SentryPerformanceTracker () <SentryTracerDelegate>
                                                   customSamplingContext:@{}
                                                            timerWrapper:nil];
 
-            if ([newSpan isKindOfClass:[SentryTracer class]]) {
-                [(SentryTracer *)newSpan setDelegate:self];
-            }
+            [(SentryTracer *)newSpan setDelegate:self];
         }];
     }
 


Sources/Sentry/SentryPerformanceTracker.m
```diff
@@ -85,9 +85,7 @@ - (SentrySpanId *)startSpanWithName:(NSString *)name
                                                   customSamplingContext:@{}
                                                            timerWrapper:nil];
 
-            if ([newSpan isKindOfClass:[SentryTracer class]]) {
-                [(SentryTracer *)newSpan setDelegate:self];
-            }
+            [(SentryTracer *)newSpan setDelegate:self];
```
philipphofmann, line 88: `startTransactionWithContext` always returns a `SentryTracer`
https://github.com/getsentry/sentry-cocoa/blob/1bbcb9c080dd28de14e3f8583ce96130f7b9f865/Sources/Sentry/SentryPerformanceTracker.m#L82-L90

see https://github.com/getsentry/sentry-cocoa/blob/1bbcb9c080dd28de14e3f8583ce96130f7b9f865/Sources/Sentry/SentryHub.m#L389-L417

brustolin (approved):

PR #2842, opened 2023-03-30
ref: mock clock for GPU metrics tests and fix slicing logic

armcknight:
#skip-changelog

I noticed recently that a frozen GPU frame reported during an ANR from iOS-Swift was not aligned to where it should be with the stack frames that likely caused it ([profile](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/b09733aa309b4b1f875eb987abf20363/flamechart/?colorCoding=by+system+vs+application+frame&query=&sorting=call+order&tid=12291&view=top+down)): 
image

This is an investigation to see if there's an SDK bug.

As part of this, I'm refactoring the profiling tests to validate the timestamp values of the reported metrics, originally mentioned [here](https://github.com/getsentry/sentry-cocoa/pull/2545#discussion_r1113599464) (and later, [here](https://github.com/getsentry/sentry-cocoa/pull/2786#issuecomment-1468537060) and [here](https://github.com/getsentry/sentry-cocoa/pull/2823#discussion_r1147962530))

The commits basically represent 3 different bodies of work:
1. https://github.com/getsentry/sentry-cocoa/commit/4618eb6d530498008403e02e699e7c4d947e05e6 .. https://github.com/getsentry/sentry-cocoa/commit/2ef0faed5579e90efe3d1c32b63d15a73639bf83 are some refactors to be able to mock everything in the profiler tests and make the test wrappers easier to work with
2. https://github.com/getsentry/sentry-cocoa/commit/0c33454ddfaddc3b58757f7b1b33601f131ac205 reworks the profiler tests to mock the date provider, set the expectations around what GPU frame info should be reported, and validate the values the profiler reports
3. everything after are fixes in the SDK to get the mocked test working again

**Updates**
- Moved refactor of other components and test helpers to https://github.com/getsentry/sentry-cocoa/pull/2849
- The bug being investigated might not be client-side. We're going to consider the investigation closed while fronted checks a few things. Updated the PR title to reflect the changes that will be merged.

**Update 2**
- We believe we found the bug in the client and fixed in #2856

### DIFF
--- a/Sources/Sentry/SentryMetricProfiler.mm
+++ b/Sources/Sentry/SentryMetricProfiler.mm
@@ -48,7 +48,7 @@ SentrySerializedMetricEntry *_Nullable serializeValuesWithNormalizedTime(
     [absoluteTimestampValues enumerateObjectsUsingBlock:^(
         SentryMetricReading *_Nonnull reading, NSUInteger idx, BOOL *_Nonnull stop) {
         // if the metric reading wasn't recorded until the transaction ended, don't include it
-        if (orderedChronologically(transaction.endSystemTime, reading.absoluteTimestamp)) {
+        if (!orderedChronologically(reading.absoluteTimestamp, transaction.endSystemTime)) {
             return;
         }
 
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -201,11 +201,13 @@ processFrameRenders(SentryFrameInfoTimeSeries *frameInfo, SentryTransaction *tra
         const auto frameRenderStart = obj[@"start_timestamp"].unsignedLongLongValue;
 
         if (!orderedChronologically(transaction.startSystemTime, frameRenderStart)) {
-            SENTRY_LOG_DEBUG(@"GPU frame render started before profile start, will not report it.");
+            SENTRY_LOG_DEBUG(@"GPU frame render started (%llu) before transaction start (%llu), "
+                             @"will not report it.",
+                frameRenderStart, transaction.startSystemTime);
             return;
         }
         const auto frameRenderEnd = obj[@"end_timestamp"].unsignedLongLongValue;
-        if (orderedChronologically(transaction.endSystemTime, frameRenderEnd)) {
+        if (!orderedChronologically(frameRenderEnd, transaction.endSystemTime)) {
             SENTRY_LOG_DEBUG(@"Frame render finished after transaction finished, won't record.");
             return;
         }
@@ -248,7 +250,7 @@ processFrameRates(SentryFrameInfoTimeSeries *frameRates, SentryTransaction *tran
         if (!orderedChronologically(transaction.startSystemTime, timestamp)) {
             return;
         }
-        if (orderedChronologically(transaction.endSystemTime, timestamp)) {
+        if (!orderedChronologically(timestamp, transaction.endSystemTime)) {
             return;
         }
 
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -162,7 +162,8 @@ std::mutex _gProfilerLock;
 SentryProfiler *_Nullable _gCurrentProfiler;
 SentryNSProcessInfoWrapper *_gCurrentProcessInfoWrapper;
 SentrySystemWrapper *_gCurrentSystemWrapper;
-SentryNSTimerWrapper *_gCurrentTimerWrapper;
+SentryNSTimerWrapper *_gMetricTimerWrapper;
+SentryNSTimerWrapper *_gTimeoutTimerWrapper;
 #    if SENTRY_HAS_UIKIT
 SentryFramesTracker *_gCurrentFramesTracker;
 #    endif // SENTRY_HAS_UIKIT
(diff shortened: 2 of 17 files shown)


philipphofmann:
> I noticed recently that a frozen GPU frame reported during an ANR from iOS-Swift was not aligned to where it should be with the stack frames that likely caused it

That's most likely because https://github.com/getsentry/sentry-cocoa/discussions/2715#discussioncomment-5091072.

philipphofmann:
Opening several PRs would be great, @armcknight. Then giving feedback is way easier IMO.

armcknight:
> > I noticed recently that a frozen GPU frame reported during an ANR from iOS-Swift was not aligned to where it should be with the stack frames that likely caused it
> 
> That's most likely because [#2715 (comment)](https://github.com/getsentry/sentry-cocoa/discussions/2715#discussioncomment-5091072).

Thanks for this link @philipphofmann 🥔

Sources/Sentry/SentryMetricProfiler.mm
```diff
@@ -48,7 +48,7 @@ @implementation SentryMetricReading
     [absoluteTimestampValues enumerateObjectsUsingBlock:^(
         SentryMetricReading *_Nonnull reading, NSUInteger idx, BOOL *_Nonnull stop) {
         // if the metric reading wasn't recorded until the transaction ended, don't include it
-        if (orderedChronologically(transaction.endSystemTime, reading.absoluteTimestamp)) {
```
armcknight, line 51: This would throw out a measurement even if it happened at the time the transaction ended. I figure we might want to keep that data instead. Flipping parameters and negating changes testing a <= b instead to a > b, which is really the case we want to skip (measurement came strictly after transaction).

Sources/Sentry/SentryProfiler.mm
```diff
@@ -248,7 +250,7 @@
         if (!orderedChronologically(transaction.startSystemTime, timestamp)) {
             return;
         }
-        if (orderedChronologically(transaction.endSystemTime, timestamp)) {
+        if (!orderedChronologically(timestamp, transaction.endSystemTime)) {
```
armcknight, line 253: Same as

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -84,31 +96,89 @@ class SentryProfilerSwiftTests: XCTestCase {
...
+                        "elapsed_since_start_ns": String(_previousSystemTime),
+                        "value": NSNumber(value: displayLinkWrapper.currentFrameRate.rawValue)
+                    ])
                 }
-                displayLinkWrapper.normalFrame()
-                if frozenFrames < mockFrozenFramesPerBatch {
+
+                switch type {
+                case .normal:
+                    currentDateProvider.advanceBy(nanoseconds: 1)
```
philipphofmann, line 119: `l`: I think testing this with actual values that frames usually have would be better. 16.6 ms for a normal frame, slow frame 18 ms, and frozen frame 701 ms for 60 fps.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -84,31 +96,89 @@ class SentryProfilerSwiftTests: XCTestCase {
...
             }
 
+            /*
+             * Mock a series of GPU frame renders of varying quality (normal/slow/frozen) and
+             * refresh rate changes. The refresh rate changes ("|") happen at the same time as
+             * the frame render they appear above. Time is not to scale; frozen frames last
+             * much longer than the lower end of slow frames.
+             *
+             * refresh rate:  |---60hz--------------|---120hz------|---60hz--------------------------------------|
+             * time:          N--S----N--F----------N-N-S--N-F-----N--N--S----N--F----------N--S----N--F----------
```
philipphofmann, line 159: 💯 for the comment

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -449,18 +507,59 @@ private extension SentryProfilerSwiftTests {
...
 #endif
     }
 
-    func assertMetricValue<T: Equatable>(measurements: [String: Any], key: String, numberOfReadings: Int, expectedValue: T?, transaction: Transaction) throws {
+    func assertMetricEntries(measurements: [String: Any], key: String, expectedEntries: [[String: Any]], transaction: Transaction) throws {
+        guard let metricContainer = measurements[key] as? [String: Any] else {
+            throw TestError.noMetricsReported
+        }
+        guard let actualEntries = metricContainer["values"] as? [[String: Any]] else {
+            throw TestError.malformedMetricValueEntry
```
philipphofmann, line 521: Throwing errors in tests. That seems smart. TIL; cc @brustolin

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -449,18 +507,59 @@ private extension SentryProfilerSwiftTests {
...
+            print(entries.reduce(into: [NSString](), { partialResult, entry in
+                partialResult.append(entry["elapsed_since_start_ns"] as! NSString)
+            }))
+        }
+        print("expected entry absolute timestamps for \(key)")
+        printTimestamps(entries: sortedExpectedEntries)
+        print("actual entry relative timestamps for \(key)")
+        printTimestamps(entries: sortedActualEntries)
+        print("transaction start time: \(transaction.startSystemTime)")
+        // /debug
```
philipphofmann, line 545: `m`: I think this is going to spam the test logs. It would be awesome to only print this as an assertion message when the test fails.

philipphofmann (approved):
LGTM, with some minor suggestions.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
armcknight: i like that idea. i was planning to just remove it but it would be helpful

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
armcknight: Yeah, I just use them like early returns with added context!

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
armcknight: Originally I did that, but it was tough to debug issues in the tests with lots of large, noninteger values.

Worth noting here that the clocks are still independent between `currentDateProvider` and `displayLinkWrapper`. I thought for a bit about unifying them for #2849 but since it technically works as-is, where `displayLinkWrapper` essentially performs the necessary modifications to cause a frame of the desired type to be recorded, I'm hesitant to modify it this go around. I agree it's confusing though, let me think about it a bit more today.

Also note that the profiling code that consumes these timestamps doesn't actually validate things like "this frame took X time to render, but its timestamp is only (X-y) time since the previous frame was rendered."

armcknight: Wound up doing this in #2856

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
philipphofmann: Now when I think about it again I think [XCTUnwrap](https://developer.apple.com/documentation/xctest/3380195-xctunwrap) would be better for that use case.

philipphofmann (approved):
LGTM, maybe it makes sense to replace the throwing of errors with XCTUnwrap; I leave this up to you.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
armcknight: Yeah you're right, i think I did these back before I got into the habit of using XCTUnwrap. I'll change them.

Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
```diff
@@ -18,6 +18,11 @@ class SentryFramesTrackerTests: XCTestCase {
     }
     
     private var fixture: Fixture!
+
+    override class func setUp() {
+        super.setUp()
+        SentryLog.configure(true, diagnosticLevel: .debug)
```
philipphofmann, line 24: `m`: Every test running after this will also have `diagnosticLevel: .debug`. Why do we need to change the log level anyways?

philipphofmann (commented):
Changing the log level in the tests seems a bit weird to me.

Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
armcknight: I needed it for debugging, couldn't tell what was happening in the internals when reading the raw test logs from CI.

Various tests reset this when they call SentrySDK.start, for instance, without the debug option set to true. So I've gotten into the habit of just doing this for any tests where I'm debugging. I can remove it here for now, we can figure out a better strategy for this later.

Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
```diff
@@ -18,6 +18,11 @@ class SentryFramesTrackerTests: XCTestCase {
...
+    }
```
armcknight, line 25:
```suggestion
```

Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
```diff
@@ -69,14 +74,17 @@ class SentryFramesTrackerTests: XCTestCase {
...
-        fixture.displayLinkWrapper.fastestFrozenFrame()
+        _ = fixture.displayLinkWrapper.fastestFrozenFrame()
 
         try assert(slow: 1, frozen: 1, total: 2, frameRates: 2)
     }
     
     func testAllFrames_ConcurrentRead() throws {
+        // this is a really long test and the logs slow it down
+        SentryLog.configure(false, diagnosticLevel: .error)
+
```
armcknight, line 87:
```suggestion
```

Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
```diff
@@ -92,12 +100,15 @@ class SentryFramesTrackerTests: XCTestCase {
...
-            fixture.displayLinkWrapper.fastestFrozenFrame()
+            _ = fixture.displayLinkWrapper.fastestSlowFrame()
+            _ = fixture.displayLinkWrapper.fastestFrozenFrame()
         }
         
         group.wait()
         try assert(slow: frames, frozen: frames, total: 3 * frames)
+
+        // ok, set logs back to debug for other tests
+        SentryLog.configure(true, diagnosticLevel: .debug)
```
armcknight, line 111:
```suggestion
```

PR #2849, opened 2023-04-01
ref: testability affordances for frames tracker and date provider

armcknight:
In preparation for https://github.com/getsentry/sentry-cocoa/pull/2842, make some changes to expose new information and consolidate some operations to new/single sources of truth.


armcknight:
There are some test failures caused introduced in https://github.com/getsentry/sentry-cocoa/actions/runs/4540460799/jobs/8001435622?pr=2778#step:11:90

Sources/Sentry/SentryFramesTracker.m
```diff
@@ -142,41 +152,44 @@ - (void)displayLinkCallback
     BOOL shouldRecordFrameRates = [SentryProfiler isRunning];
 #        endif // defined(TEST) || defined(TESTCI)
     BOOL hasNoFrameRatesYet = self.frameRateTimestamps.count == 0;
-    BOOL frameRateSignificantlyChanged
-        = fabs(self.frameRateTimestamps.lastObject[@"frame_rate"].doubleValue
-              - actualFramesPerSecond)
-        > 1e-10f; // these may be a small fraction off of a whole number of frames per second, so
-                  // allow some small epsilon difference
+    double previousFrameRate
+        = self.frameRateTimestamps.lastObject[@"frame_rate"].unsignedLongLongValue;
```
philipphofmann, line 156: `m`: `unsignedLongLongValue` and `double` don't match. I think it should be
```suggestion
    uint64_t previousFrameRate
        = self.frameRateTimestamps.lastObject[@"frame_rate"].unsignedLongLongValue;
```

SentryTestUtils/TestCurrentDateProvider.swift
```diff
@@ -17,10 +17,27 @@ public class TestCurrentDateProvider: NSObject, CurrentDateProvider {
...
         internalDate = date
     }
+
+    /// Advance the current date by the specified number of seconds.
+    public func advance(by seconds: TimeInterval) {
+        setDate(date: date().addingTimeInterval(seconds))
+        internalSystemTime += seconds.toNanoSeconds()
+    }
+
+    public func advanceBy(nanoseconds: UInt64) {
```
philipphofmann, line 37: Thanks for that 👏

philipphofmann (commented):
As said by @brustolin, CI is broken, apart from that LGTM.

Sources/Sentry/SentryFramesTracker.m
armcknight: Excellent catch. I read the type keyword as "const auto" (got too spoiled by ObjC++ 😆 )

armcknight:
Made all the changes, but CI isn't going to pass until I merge the upstream PRs into this one. Although, they're all now merged/rebased with latest main, so if something that merged recently is causing all the failures, it might not pass anymore :/

armcknight:
I merged #2856 into #2842, which is also approved, but wanted to wait to merge that into this one until this one is approved, so that it's easier to review this one.
 
The tests are passing in #2842, but because I had some trouble separating the changes in this PR from that one, the tests don't pass in this one. Once I merge #2842 into this one, it will fix the issues. So, waiting for your reviews here to do so @philipphofmann @brustolin.

Sources/Sentry/SentryFramesTracker.m
```diff
@@ -16,7 +18,7 @@
 typedef NSMutableArray<NSDictionary<NSString *, NSNumber *> *> SentryMutableFrameInfoTimeSeries;
 #    endif // SENTRY_TARGET_PROFILING_SUPPORTED
 
-static CFTimeInterval const SentryFrozenFrameThreshold = 0.7;
+CFTimeInterval const SentryFrozenFrameThreshold = 0.7;
```
brustolin, line 21: l: Why we don't need/want static here anymore?

Sources/Sentry/SentryProfiler.mm
```diff
@@ -136,7 +136,7 @@
     }
 
     const auto sample = [[SentrySample alloc] init];
-    sample.absoluteTimestamp = backtrace.absoluteTimestamp;
+    sample.absoluteTimestamp = SentryCurrentDate.getCurrentDateProvider.systemTime;
```
brustolin, line 139: m: Is this correct? Before you had a fixed time extracted from the backtrace and now you're getting the current time.

brustolin (commented):
It looks good to me.
Just a few questions.

Sources/Sentry/SentryProfiler.mm
armcknight: This is a good point. this would push the timestamps gathered for backtrace samples a bit later than when they were previously gathered. When I changed it I assumed it wouldn't be a big deal, but let's check with @indragiek. Good catch @brustolin!

philipphofmann (approved):
I think only https://github.com/getsentry/sentry-cocoa/pull/2849#discussion_r1162418942 is missing, so LGTM.

armcknight:
> I think only [#2849 (comment)](https://github.com/getsentry/sentry-cocoa/pull/2849#discussion_r1162418942) is missing, so LGTM.

Done in https://github.com/getsentry/sentry-cocoa/pull/2849/commits/93ce1c2aaadd257108c6f7c1e34233af0588fd89!

Sources/Sentry/SentryProfiler.mm
```diff
@@ -136,7 +136,7 @@
     }
 
     const auto sample = [[SentrySample alloc] init];
-    sample.absoluteTimestamp = SentryCurrentDate.systemTime;
+    sample.absoluteTimestamp = backtrace.absoluteTimestamp;
```
indragiek: I don't think all the changes in the C++ code is necessary -- `SamplingProfiler` and `Backtrace` can stay as they are.

`SentryProfiler` can be initialized with a block with the signature: `const Backtrace & -> uint64_t` (maybe this initializer needs to go in a test-only header, doesn't need to be part of the public API). This block would be called on this line to set `sample.absoluteTimestamp`

When you default initialize this outside of the test, the block should simply return `backtrace.absoluteTimestamp` to preserve the current behavior.

If you need to mock it for the test, the test can pass some other block to the `SentryProfiler` initializer that returns either a fixed value (as you're doing here with `SentryCurrentDate.systemTime`) or an offset added to `backtrace.absoluteTimestamp`

That means that the only changes would need to be in this file and everything else can go away.

indragiek (changes requested):

Sources/Sentry/SentryProfiler.mm
armcknight: I don't have a clear idea of what this would look like. If we do this, then we have to wait until after `f` is already finished executing for it to return the timestamp, right? In that case, we aren't getting a timestamp until after the rest of the backtrace has been processed. That will be functionally the same as my initial change, where we just set the sample's timestamp to the value returned by the date provider after processing the backtrace: https://github.com/getsentry/sentry-cocoa/pull/2849/files/446ca92e0bfccf4bf51152bee81c64ac361a5aef..93ce1c2aaadd257108c6f7c1e34233af0588fd89#diff-65ccfc43cdffcd8d20527ee954818c2298714c3790e8d55eb1ab06cee4c7beb9L111. This would be simpler than modifying the callback to return a value.

Maybe I'm just misunderstanding what is being asked for. Would you mind opening a PR into this branch to show me?

Sources/Sentry/SentryProfiler.mm
```diff
@@ -624,8 +624,15 @@ - (void)start
                 SENTRY_LOG_WARN(@"Profiler instance no longer exists, cannot process next sample.");
                 return;
             }
+#    if defined(TEST) || defined(TESTCI)
+            Backtrace backtraceCopy = backtrace;
+            backtraceCopy.absoluteTimestamp = SentryCurrentDate.systemTime;
+            processBacktrace(backtraceCopy, threadMetadata, queueMetadata, samples, stacks, frames,
```
indragiek, line 630: maybe add a comment to explain why this is necessary

indragiek (approved):

PR #2856, opened 2023-04-04
fix: profiling GPU information timestamp choice

armcknight:
We realized that we were working under an incorrect assumption when recording GPU frame timestamps. This PR changes the timestamp that we use when we record that a slow or frozen timestamp has been encountered.

It also contains some refactoring of the associated logic in SentryProfiler and SentryFramesTracker and refactors and fixes for the profiler tests. This also includes: 
- converting the remainder of the profiling tests to use the mock date provider instead of delayed dispatches to achieve timeouts... this solves a long-standing request to speed up the profiling tests @philipphofmann!
    - this let us remove _most_ of the delayed dispatches; the remainder were exercising the profiler's timeout logic; to complete this, also used the SentryNSTimerWrapper for the NSTimer the profiler uses to abort due to an elapsed timeout period so it can be manually triggered in the tests
- using the TestCurrentDateProvider in the TestDisplayLinkWrapper instead of having that component maintain its own internal time

Here's a before and after of what reported slow/frozen frames looks like, you can see how they seem to line up better with the tested ANR event after making the fix:

[Before](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/b09733aa309b4b1f875eb987abf20363/flamechart/?colorCoding=by+system+vs+application+frame&query=&sorting=call+order&tid=12291&view=top+down):

[After](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/479e2400b3084f1e8e7072717b4ec644/flamechart/?colorCoding=by+application+frame&fov=0%2C0%2C9130846208%2C23&query=&sorting=call+order&tid=8451&view=top+down):

Screenshot 2023-04-04 at 10 05 34 AM
Screenshot 2023-04-04 at 10 05 45 AM
Screenshot 2023-04-04 at 10 06 05 AM

#skip-changelog

### DIFF
--- a/SentryTestUtils/TestDisplayLinkWrapper.swift
+++ b/SentryTestUtils/TestDisplayLinkWrapper.swift
@@ -14,17 +14,21 @@ public enum FrameRate: UInt64 {
     case low  = 60
     case high = 120
 
-    public var tickDuration: Double {
-        return 1 / Double(self.rawValue)
+    public var tickDuration: CFTimeInterval {
+        return 1 / CFTimeInterval(self.rawValue)
     }
 }
 
 public class TestDisplayLinkWrapper: SentryDisplayLinkWrapper {
     public var target: AnyObject!
     public var selector: Selector!
-    var internalTimestamp = 0.0
     public var currentFrameRate: FrameRate = .low
     let frozenFrameThreshold = 0.7
+    public var dateProvider: TestCurrentDateProvider
+
+    public init(dateProvider: TestCurrentDateProvider = TestCurrentDateProvider()) {
+        self.dateProvider = dateProvider
+    }
 
     public override func link(withTarget target: Any, selector sel: Selector) {
         self.target = target as AnyObject
@@ -32,11 +36,11 @@ public class TestDisplayLinkWrapper: SentryDisplayLinkWrapper {
     }
 
     public override var timestamp: CFTimeInterval {
-        return internalTimestamp
+        return dateProvider.systemTime().toTimeInterval()
     }
 
     public override var targetTimestamp: CFTimeInterval {
-        return internalTimestamp + currentFrameRate.tickDuration
+        return dateProvider.systemTime().toTimeInterval() + currentFrameRate.tickDuration
     }
 
     public override func invalidate() {
@@ -53,34 +57,41 @@ public class TestDisplayLinkWrapper: SentryDisplayLinkWrapper {
     }
     
     public func normalFrame() {
-        internalTimestamp += currentFrameRate.tickDuration
+        dateProvider.advance(by: currentFrameRate.tickDuration)
         call()
     }
     
-    public func fastestSlowFrame() {
-        internalTimestamp += slowFrameThreshold(currentFrameRate.rawValue) + timeEpsilon
+    public func fastestSlowFrame() -> CFTimeInterval {
+        let duration: Double = slowFrameThreshold(currentFrameRate.rawValue) + timeEpsilon
+        dateProvider.advance(by: duration)
         call()
+        return duration
     }
 
-    public func middlingSlowFrame() {
-        internalTimestamp += (frozenFrameThreshold - (slowFrameThreshold(currentFrameRate.rawValue) + timeEpsilon)) / 2.0
+    public func middlingSlowFrame() -> CFTimeInterval {
+        let duration: Double = (frozenFrameThreshold - (slowFrameThreshold(currentFrameRate.rawValue) + timeEpsilon)) / 2.0
+        dateProvider.advance(by: duration)
         call()
+        return duration
     }
     
-    public func slowestSlowFrame() {
-        internalTimestamp += frozenFrameThreshold
+    public func slowestSlowFrame() -> CFTimeInterval {
+        dateProvider.advance(by: frozenFrameThreshold)
         call()
+        return frozenFrameThreshold
     }
 
-    public func fastestFrozenFrame() {
-        internalTimestamp += frozenFrameThreshold + timeEpsilon
+    public func fastestFrozenFrame() -> CFTimeInterval {
+        let duration: Double = frozenFrameThreshold + timeEpsilon
+        dateProvider.advance(by: duration)
         call()
+        return duration
     }
 
     /// There's no upper bound for a frozen frame, except maybe for the watchdog time limit.
     /// - parameter extraTime: the additional time to add to the frozen frame threshold when simulating a frozen frame.
     public func slowerFrozenFrame(extraTime: TimeInterval = 0.1) {
-        internalTimestamp += frozenFrameThreshold + extraTime
+        dateProvider.advance(by: frozenFrameThreshold + extraTime)
         call()
     }
     
@@ -88,11 +99,11 @@ public class TestDisplayLinkWrapper: SentryDisplayLinkWrapper {
         self.call()
 
         for _ in 0..<slow {
-            fastestSlowFrame()
+            _ = fastestSlowFrame()
         }
         
         for _ in 0..<frozen {
-            fastestFrozenFrame()
+            _ = fastestFrozenFrame()
         }
 
         for _ in 0..<(normal - 1) {
--- a/Sources/Sentry/SentryFramesTracker.m
+++ b/Sources/Sentry/SentryFramesTracker.m
@@ -162,6 +162,8 @@ slowFrameThreshold(uint64_t actualFramesPerSecond)
         [self recordTimestamp:thisFrameSystemTimestamp
                         value:@(currentFrameRate)
                         array:self.frameRateTimestamps];
+    } else {
+        SENTRY_LOG_DEBUG(@"Won't record frame rate at %llu", thisFrameSystemTimestamp);
     }
 #    endif // SENTRY_TARGET_PROFILING_SUPPORTED
 
--- a/Sources/Sentry/SentryFramesTracker.m
+++ b/Sources/Sentry/SentryFramesTracker.m
@@ -162,8 +162,6 @@ slowFrameThreshold(uint64_t actualFramesPerSecond)
         [self recordTimestamp:thisFrameSystemTimestamp
                         value:@(currentFrameRate)
                         array:self.frameRateTimestamps];
-    } else {
-        SENTRY_LOG_DEBUG(@"Won't record frame rate at %llu", thisFrameSystemTimestamp);
     }
 #    endif // SENTRY_TARGET_PROFILING_SUPPORTED
 
--- a/Sources/Sentry/SentryFramesTracker.m
+++ b/Sources/Sentry/SentryFramesTracker.m
@@ -153,16 +153,15 @@ slowFrameThreshold(uint64_t actualFramesPerSecond)
 #        endif // defined(TEST) || defined(TESTCI)
     BOOL hasNoFrameRatesYet = self.frameRateTimestamps.count == 0;
     uint64_t previousFrameRate
-        = self.frameRateTimestamps.lastObject[@"frame_rate"].unsignedLongLongValue;
+        = self.frameRateTimestamps.lastObject[@"value"].unsignedLongLongValue;
     BOOL frameRateChanged = previousFrameRate != currentFrameRate;
     BOOL shouldRecordNewFrameRate
         = shouldRecordFrameRates && (hasNoFrameRatesYet || frameRateChanged);
     if (shouldRecordNewFrameRate) {
-        SENTRY_LOG_DEBUG(@"Recording new frame rate at %llu.", self.previousFrameSystemTimestamp);
-        [self.frameRateTimestamps addObject:@{
-            @"timestamp" : @(self.previousFrameSystemTimestamp),
-            @"frame_rate" : @(currentFrameRate),
-        }];
+        SENTRY_LOG_DEBUG(@"Recording new frame rate at %llu.", thisFrameSystemTimestamp);
+        [self recordTimestamp:thisFrameSystemTimestamp
+                        value:@(currentFrameRate)
+                        array:self.frameRateTimestamps];
     }
 #    endif // SENTRY_TARGET_PROFILING_SUPPORTED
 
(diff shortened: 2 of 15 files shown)


SentryTestUtils/TestDisplayLinkWrapper.swift
```diff
@@ -14,29 +14,33 @@ public enum FrameRate: UInt64 {
...
     }
 }
 
 public class TestDisplayLinkWrapper: SentryDisplayLinkWrapper {
     public var target: AnyObject!
     public var selector: Selector!
-    var internalTimestamp = 0.0
     public var currentFrameRate: FrameRate = .low
     let frozenFrameThreshold = 0.7
+    public var dateProvider: TestCurrentDateProvider
```
armcknight, line 27: now using a date provider instead of maintaining an independent internal clock

SentryTestUtils/TestDisplayLinkWrapper.swift
```diff
@@ -53,46 +57,53 @@ public class TestDisplayLinkWrapper: SentryDisplayLinkWrapper {
...
         call()
     }
     
-    public func fastestSlowFrame() {
-        internalTimestamp += slowFrameThreshold(currentFrameRate.rawValue) + timeEpsilon
+    public func fastestSlowFrame() -> CFTimeInterval {
+        let duration: Double = slowFrameThreshold(currentFrameRate.rawValue) + timeEpsilon
+        dateProvider.advance(by: duration)
         call()
+        return duration
```
armcknight, line 68: Return the duration so that we can know what to assert on in the profiling tests; we record the duration of frames, so need to make sure the serialized value is as expected.

This required adding a few `_ = ` to other usages of this api in some tests.

Sources/Sentry/SentryFramesTracker.m
```diff
@@ -153,16 +153,15 @@ - (void)displayLinkCallback
...
     if (shouldRecordNewFrameRate) {
-        SENTRY_LOG_DEBUG(@"Recording new frame rate at %llu.", self.previousFrameSystemTimestamp);
-        [self.frameRateTimestamps addObject:@{
-            @"timestamp" : @(self.previousFrameSystemTimestamp),
-            @"frame_rate" : @(currentFrameRate),
-        }];
+        SENTRY_LOG_DEBUG(@"Recording new frame rate at %llu.", thisFrameSystemTimestamp);
+        [self recordTimestamp:thisFrameSystemTimestamp
+                        value:@(currentFrameRate)
+                        array:self.frameRateTimestamps];
```
armcknight, line 164: Previously slow/frozen frames were recorded with the same structure and dictionary keys (start and end timestamps), and refresh rates were different (timestamps and values). This PR unifies them so they all are recorded with a timestamp and a value: for slow/frozen frames, the value is the duration. For refresh rate changes, the value is the new refresh rate.

This also has a downstream effect of allowing us to consolidate `SentryProfiler.processFrameRenders` and `SentryProfiler.processFrameRates`, see later in the diff.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -193,75 +194,32 @@
  * didn't occur within the profile time.
  */
 NSArray<SentrySerializedMetricReading *> *
-processFrameRenders(SentryFrameInfoTimeSeries *frameInfo, SentryTransaction *transaction)
+sliceGPUData(SentryFrameInfoTimeSeries *frameInfo, SentryTransaction *transaction)
```
armcknight, line 197: Here's where the aforementioned consolidation of slicing frame renders and refresh rates is. Since they're not written in a uniform data structure with the same keys, they can all be handled with the same logic instead of two different functions. `processFrameRenders` used to not only slice the data to transaction bounds, but also had to calculate frame durations; that's now done in the frames tracker. After that, it does the same thing as `processFrameRenders` did, so they are combined in a single function now named `slicedGPUData`.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -343,12 +301,15 @@ + (void)startWithHub:(SentryHub *)hub
...
-        [NSTimer scheduledTimerWithTimeInterval:kSentryProfilerTimeoutInterval
-                                         target:self
-                                       selector:@selector(timeoutAbort)
-                                       userInfo:nil
-                                        repeats:NO];
+        [_gTimeoutTimerWrapper scheduledTimerWithTimeInterval:kSentryProfilerTimeoutInterval
+                                                       target:self
+                                                     selector:@selector(timeoutAbort)
+                                                     userInfo:nil
+                                                      repeats:NO];
```
armcknight, line 312: Using the `SentryNSTimerWrapper` so we can mock timeouts in unit tests.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -479,10 +440,16 @@ + (void)useProcessInfoWrapper:(SentryNSProcessInfoWrapper *)processInfoWrapper
...
+{
+    std::lock_guard<std::mutex> l(_gProfilerLock);
+    _gMetricTimerWrapper = timerWrapper;
+}
+
++ (void)useTimeoutTimerWrapper:(SentryNSTimerWrapper *)timerWrapper
 {
     std::lock_guard<std::mutex> l(_gProfilerLock);
-    _gCurrentTimerWrapper = timerWrapper;
+    _gTimeoutTimerWrapper = timerWrapper;
```
armcknight, line 452: We now have two timer wrappers here, one for metrics collection as before, and a new one for profile timeout. We'll use two separate instances.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -92,60 +97,48 @@ class SentryProfilerSwiftTests: XCTestCase {
 
             // gather mock cpu usages and memory footprints
             for _ in 0..<mockUsageReadingsPerBatch {
-                self.timerWrapper.fire()
+                self.metricTimerWrapper.fire()
             }
 
     #if !os(macOS)
-            var shouldRecordFrameRateExpectation = false
-            var previousSystemTime = currentDateProvider.systemTime()
```
armcknight, line 100: We don't have to track this now that we are changing from recording frame info with the "current" timestamp instead of the "previous" one.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -92,60 +97,48 @@ class SentryProfilerSwiftTests: XCTestCase {
...
 
             func renderGPUFrame(_ type: GPUFrame) {
-                if shouldRecordFrameRateExpectation {
-                    shouldRecordFrameRateExpectation = false
-                    print("will expect frame rate \(displayLinkWrapper.currentFrameRate.rawValue) at \(previousSystemTime)")
-                    expectedFrameRateChanges.append([
-                        "elapsed_since_start_ns": String(previousSystemTime),
-                        "value": NSNumber(value: displayLinkWrapper.currentFrameRate.rawValue)
-                    ])
-                }
```
armcknight, line 115: The rest of the diff in this hunk deals with ordering of advancing the mocked system clock (which is now managed by the display link wrapper) since we record the "current" timestamp instead of "previous" now. Since some of the granularity of control was lost in this file by putting it there, we needed to reorder a few things. One of which was moving the creation of a refresh rate expectation until _after_ mocking a slow/frozen frame (which advances the system clock).

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -175,6 +169,8 @@ class SentryProfilerSwiftTests: XCTestCase {
             renderGPUFrame(.normal)
             renderGPUFrame(.frozen)
             renderGPUFrame(.normal)
+            changeFrameRate(.high)
+            renderGPUFrame(.normal)
```
armcknight, line 173: Always start and end on opposite frame rates, so the frame rate change gets picked up by the frames tracker between its previous and current callback invocations.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -207,121 +203,92 @@ class SentryProfilerSwiftTests: XCTestCase {
...
+    }
+
     override func setUp() {
         super.setUp()
         fixture = Fixture()
-        SentryLog.configure(true, diagnosticLevel: .debug)
-        CurrentDate.setCurrentDateProvider(fixture.currentDateProvider)
-        fixture.options.profilesSampleRate = 1.0
-        fixture.options.tracesSampleRate = 1.0
-        fixture.prepareMetricsMocks()
```
armcknight, line 217: moved to fixture init

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -56,22 +77,6 @@ class SentryProfilerSwiftTests: XCTestCase {
         let mockFrameRateChangesPerBatch: [FrameRate] = [.high, .low, .high, .low]
 #endif
 
-        func prepareMetricsMocks() {
```
armcknight, line 59: moved to fixture init

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -27,18 +27,39 @@ class SentryProfilerSwiftTests: XCTestCase {
...
+#if !os(macOS)
+            SentryProfiler.useFramesTracker(framesTracker)
+            framesTracker.start()
+            displayLinkWrapper.call()
+#endif
+        }
+
         /// Advance the mock date provider, start a new transaction and return its handle.
         func newTransaction(testingAppLaunchSpans: Bool = false) throws -> SentryTracer {
-            currentDateProvider.advanceBy(nanoseconds: 2)
```
armcknight, line 41: Moved to test cases that need it, which is only the concurrent test cases.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -207,121 +203,92 @@ class SentryProfilerSwiftTests: XCTestCase {
...
-        SentryLog.configure(true, diagnosticLevel: .debug)
-        CurrentDate.setCurrentDateProvider(fixture.currentDateProvider)
-        fixture.options.profilesSampleRate = 1.0
-        fixture.options.tracesSampleRate = 1.0
-        fixture.prepareMetricsMocks()
     }
 
     override func tearDown() {
         super.tearDown()
+        fixture.timeoutTimerWrapper.fire()
```
armcknight, line 218: If a test early exits because of a thrown error, it may not finish the spans it created. This ensures any open spans are closed before starting the next test case.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -207,121 +203,92 @@ class SentryProfilerSwiftTests: XCTestCase {
...
-                XCTFail("Encountered error: \(error)")
-            }
-        }
-        waitForExpectations(timeout: 3)
+        self.fixture.currentDateProvider.advanceBy(nanoseconds: 1.toNanoSeconds())
+        span.finish()
+        try self.assertMetricsPayload()
     }
 
     func testConcurrentProfilingTransactions() throws {
```
armcknight, line 231: This test case is the only one with changes pertinent to the "current" vs "previous" timestamp change, but mostly they involve correctly setting up mock state from how it was previously done. 

The test already had the following structure:

1. set up 10 spans
2. make sure each one is reported with the correct metrics/samples sliced to tx bounds
3. reset everything
4. do it all again to make sure stopping and starting the profiler again doesn't break anything (basically, copy/paste 1 and 2)

The other change in this test case is that the bodies of the loops between 1/2 and 4 had slightly diverged; extract them to a common function so they run exactly the same way both times.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -207,121 +203,92 @@ class SentryProfilerSwiftTests: XCTestCase {
...
     /// ```
-    ///    time                0s                         1s     2s     2.5s     3s  (these times are adjusted to the 1s profile timeout for testing only)
+    ///                        time->
     ///    transaction A       |---------------------------------------------------|
     ///    profiler A          |---------------------------x  <- timeout
     ///    transaction B                                           |-------|
     ///    profiler B                                              |-------|  <- normal finish
     ///   ```
     func testConcurrentSpansWithTimeout() throws {
-        let originalTimeoutInterval = kSentryProfilerTimeoutInterval
```
armcknight, line 300: The rest of the test cases are modified to use the mocked clock and timeout timer instead of relying on DispatchQueue and test expectations.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
philipphofmann: Please add this info as a comment.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -536,7 +487,7 @@ private extension SentryProfilerSwiftTests {
         let sortedExpectedEntries = sortedByTimestamps(expectedEntries)
 
         let expectedAmountOfReadings = actualEntries.count == expectedEntries.count
-        XCTAssert(expectedAmountOfReadings, "Wrong number of values under \(key). expected: \(printTimestamps(entries: sortedExpectedEntries)); actual: \(printTimestamps(entries: sortedActualEntries)); transaction start time: \(transaction.startSystemTime)")
+        XCTAssert(expectedAmountOfReadings, "Wrong number of values under \(key).\nexpected: \(printTimestamps(entries: sortedExpectedEntries));\nactual: \(printTimestamps(entries: sortedActualEntries)); transaction start time: \(transaction.startSystemTime)")
```
philipphofmann, line 490: Thanks for the descriptive assertion message 😀.

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
```diff
@@ -715,36 +666,31 @@ private extension SentryProfilerSwiftTests {
...
+        case .yes:
+            guard let event = client.captureEventWithScopeInvocations.first else {
+                XCTFail("Expected to capture at least 1 event with a profile")
+                return
             }
-
-            exp.fulfill()
+            XCTAssertEqual(1, event.additionalEnvelopeItems.count)
+        @unknown default:
+            fatalError("Unexpected value for sample decision")
```
philipphofmann, line 692: `m`: I would like to avoid `fatalError` in tests, as they will cause the test suite to crash. Please use `XCTFail` instead.

philipphofmann (approved):
Thanks for all the comments explaining the refactorings, so it was easier to review the PR 😃 🙏

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
armcknight: Yep, good call!

Tests/SentryProfilerTests/SentryProfilerSwiftTests.swift
armcknight: i also added a function to clear the bookkeeping array in SentryTracerConcurrency

Issue #2850, opened 2023-04-02
Latest Version 8.3.3 still facing a lots of crashes caused by Sentry SDK


zhgchgli0718:
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.3.3

### Steps to Reproduce

unknown, seem to be random crashes.

ref #2832

### Expected Result

The app was stable before we updated to version 7.3.1, but we're experiencing issues now.

Our current crash-free rate is around 98.X%, which is lower than our previous rate of 99.98% before we upgraded the SDK version. :(

ref #2832

### Actual Result

previous issue: #2832

We are extremely disappointed with Sentry as stability is crucial to our users and their experience.

Unfortunately, the latest version of the Sentry SDK did not address the numerous crash issues we were experiencing, and we have reverted to version 7.3.1 to maintain stability.
Upon monitoring the data after the rollback, we can confirm that the Crash-Free Rate has returned to 99.X%.

We can conclude that Sentry 8.x.x provided an unstable SDK to clients, which is absurd as crashes occurred on the service that monitors issues and crashes.

brustolin:
Hello @zhgchgli0718 Im so sorry to hear about this. 
Right now we're focused on improving the SDK stability and to solve this problems asap.

Do you mind sharing the options you use to initialize Sentry SDK (no need to share the DSN).
Also, It would help if you could share the stack trace of one of those `EXC_BAD_ACCESS` errors mentioned in the previous PR.

FelixHerrmann:
Hey guys! We noticed the same increase of crashes when we upgraded from Sentry 8.2.0 to 8.3.1 and 8.3.3. We had profiling enabled since it is available with no (noticeable) issues but 8.3.* seems to introduced something that is producing crashes quite frequently.

These are the most common ones:
- `KERN_INVALID_ADDRESS` in combination with `[__NSArrayM enumerateObjectsWithOptions:usingBlock:]`
- `Invalid type in JSON write ((null))`
- `-[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[1]`
- `-[__NSDictionaryM setObject:forKeyedSubscript:]: key cannot be nil`

@philipphofmann I will email you the name of our organization too to give you insights!

brustolin:
[...]

PR #2355, opened 2022-11-03
ref: json serialization error reporting

armcknight:
I had a JSON validation error, but couldn't tell what it was since we simply check for the boolean from `isValidJSON`. I had to implement just attempting the serialization and inspecting the inout error it gives, and learned that at least in my case, serialization also throws exceptions in certain cases, but the exception told me exactly what the problem was. I figured it would be helpful to keep around, but also that we may not want the exception handler in production, so I used some conditional compilation to keep it behaving as it was in release configs, but in all other configs, to get the info from the exception/error.

To that end, added some new functions to construct NSErrors that can include an underlying NSError we may get from a Cocoa library method, or an NSException, with some light refactoring of the implementations in SentryError.

#skip-changelog

Just copying here some details from one of my commits for visibility:

```
before:
    - test if input is valid JSON with `-[NSJSONSerialization
    isValidJSON]` and if not, build our own error.
    - problem: no hints as to why the JSON is invalid

after:
    - attempt direct serialization with `-[NSJSONSerialization
    dataWithJSONObject:options:error:]` and use the library error
    - this can throw an exception in certain failure modes: surround
    the call with try/catch
    - use new functions in SentryError to create an NSError from
    either an NSException or the inout NSError
    - only compile like this for nonrelease builds so we don't incur
    the performance hit of throwing an exception (see apple's docs:
    "For best performance in 64-bit, you should throw exceptions only
    when absolutely necessary.")
```

PR #2858, opened 2023-04-04
fix: Serialization with invalid objects

brustolin:
## :scroll: Description

Serialization may fail and crash if there is an invalid objects in the data.

Related PR https://github.com/getsentry/sentry-cocoa/pull/2355.

May be related to #2850 

Unit tests

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentrySerialization.m
+++ b/Sources/Sentry/SentrySerialization.m
@@ -22,8 +22,8 @@ NS_ASSUME_NONNULL_BEGIN
         return nil;
     }
 
-    NSError * error = nil;
-    NSData * data = [NSJSONSerialization dataWithJSONObject:dictionary options:0 error:&error];
+    NSError *error = nil;
+    NSData *data = [NSJSONSerialization dataWithJSONObject:dictionary options:0 error:&error];
     if (error) {
         SENTRY_LOG_ERROR(@"Internal error while serializing JSON: %@", error);
     }


philipphofmann (approved):
Thanks, @brustolin. LGTM

Sources/Sentry/SentrySerialization.m
```diff
@@ -16,35 +16,17 @@
...
-#else
-    if ([NSJSONSerialization isValidJSONObject:dictionary]) {
-#endif
-        data = [NSJSONSerialization dataWithJSONObject:dictionary options:0 error:error];
-#if defined(DEBUG) || defined(TEST) || defined(TESTCI)
-    } @catch (NSException *exception) {
-        if (error) {
-            SENTRY_HANDLE_ERROR(NSErrorFromSentryErrorWithException(
-                kSentryErrorJsonConversionError, @"Event cannot be converted to JSON", exception));
-        }
```
armcknight, line 40: I understand this code was weird, but it did help me to diagnose some json serialization issues in the past. `isValidJSONObject` only returns `YES` or `NO` but the exception tells you what is invalid. Maybe we should consider just always doing the `@try`/`@catch` even in production and including that info in any errors we send to ourselves.

brustolin: We had to rewrite this part of the code to be able to understand it and solve a bug that affects everyone, and ultimately, the CI code was not the same as the release code, what was why we did not detect it early.

AFAIK is a good practice to avoid @try/catch for objc when possible. 
But I agree that having the actual error description would help a lot, something we dont get with `isValidJSONObject`.

@philipphofmann any thoughts?

armcknight: Agreed it's good practice to avoid @try/@catch, but it was only compiled that way for non-prod. It was only used in debug/test to help us diagnose things during development.

I'd be surprised if we ultimately were seeing different outcomes between catching an exception or looking at the inout error, vs calling isValidJSONObject, as the documentation states either method is appropriate to validate the input.

brustolin: The error was fixed not we have a better undestanding of what was happening, I dont see a problem to use the compiler declarative again, as long as we dont make a "spaghetti" code, is hard to understand. 
Is ok to duplicate code to have only one `#if/#else`, way better to read the code.

philipphofmann: Using compiler directives led to having untested code, which led to crashes in production. If we want extra context when debugging I would prefer using something like `options.debug` and test all the code properly. Then our users would also see the extra context when they are debugging.

PR #2860, opened 2023-04-04
added 'strong' attribute

bitsandfoxes:
## :scroll: Description

From what I can tell: The prebuilt engine seems to be very specific and treats some warnings as errors.
Basically the same issue we were running into with: https://github.com/getsentry/sentry-cocoa/pull/2698

```
/sentry-unreal/sample/Plugins/sentry/Source/ThirdParty/Mac/include/Headers/SentryTraceContext.h:53:1: Error  : no 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed [-Werror,-Wobjc-property-no-attribute]
```

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/include/SentryTraceContext.h
+++ b/Sources/Sentry/include/SentryTraceContext.h
@@ -50,7 +50,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Sample rate used for this trace.
  */
-@property (nullable, nonatomic) NSString *sampleRate;
+@property (nullable, nonatomic, strong) NSString *sampleRate;
 
 /**
  * Initializes a SentryTraceContext with given properties.


brustolin (approved):

Sources/Sentry/include/SentryTraceContext.h
```diff
@@ -50,7 +50,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Sample rate used for this trace.
  */
-@property (nullable, nonatomic) NSString *sampleRate;
+@property (nullable, nonatomic, strong) NSString *sampleRate;
```
armcknight, line 53: All the other string properties are `readonly` so they don't need a retain attribute since no objects will be set into them from outside this class. Is this property different than those? I don't see any assignments to this property.

brustolin: SampleRate cant be readonly because we only now the value used for this after the SentryTraceContext being used to initialize the transaction.

PR #2861, opened 2023-04-04
test: cache the downloaded disk image for iOS 12.4 simulator

armcknight:
It takes 5-10 minutes to download the iOS 12.4 disk image, and this happens for every run of the test.yml workflow. 

Cache the downloaded disk image file so we don't have to redownload it every time.

#skip-changelog

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -168,18 +168,25 @@ jobs:
       - run: ./scripts/ci-select-xcode.sh ${{matrix.xcode}}
 
       # GH action images don't have an iOS 12.4 simulator. Therefore we have to download and install the simulator manually.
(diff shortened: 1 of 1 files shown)


armcknight:
Unfortunately the `xcode-install` tool doesn't allow separating out the download of the large file in a way that is amenable to caching :/

philipphofmann:
Caching doesn't work for that, @armcknight, as the simulator files are too big. The cache has a limit of 10GB, see https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy.

armcknight:
Yep I read that, I just figured even with cache eviction if it could reduce the number of times it would need to be downloaded, that would be something. Alas.

PR #2863, opened 2023-04-05
perf: optimize some conversions of unsigned ints to NSString

armcknight:
Some work done in a tight loop on the profiler's sampling thread used inefficient methods to convert integers to strings. We sped this up by using `snprintf`.

Targets to optimize:

- boxing unsigned integers to NSNumbers and converting to NSString with `-[NSNumber stringValue]`
- boxing to NSNumber and calling `sentry_formatHexAddress` which called `+[NSString stringWithFormat:]`
- directly formatting an unsigned int to a string with `[NSString stringWithFormat:@"%llu"`

Observed a **75% reduction in cost of doing the conversions** using Instruments, just in the profiler's sampling thread: https://www.notion.so/sentry/formatHexAddress-perf-optimization-6ce9e406e53b4d99aed2de591428599b?pvs=4

Converted a couple other usages in the SDK to also use the new function, so we get some other perf wins for free. There are still numerous usages of the older `sentry_formatHexAddress(NSNumber *)` that are using values from JSON-decoded `NSDictionary` instances in SentryCrashReportConverter, which could be a possible target for future optimizations by the Cocoa team.

# TODO

- [x] send a profile from current implementation using iOS-Swift, and then one from this changeset for the same operation, and compare them to make sure this doesn't break anything in subtle ways. tested the "fully blocking anr" in a manual transaction
    - `main`: https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/d112e213125c483280ba883f15dd72b0/flamechart/?colorCoding=by+system+vs+application+frame&fov=0%2C0%2C14142140416%2C48&query=&sorting=call+order&tid=259&view=top+down
    - with changes: https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/6496c43271564104b8a997de3ea5a645/flamechart/?colorCoding=by+system+vs+application+frame&fov=28102656%2C0%2C14030778368%2C48&query=&sorting=call+order&tid=259&view=top+down
image

#skip-changelog


Tests/SentryTests/SentryFormatterTests.swift
```diff
@@ -0,0 +1,32 @@
...
+            (0, "0"),
+            (1, "1"),
+            (123_456, "123456"),
+            (UInt64.max, "18446744073709551615")
+        ] {
+            XCTAssertEqual(sentry_stringForUInt64(UInt64(input)), expected)
+        }
+    }
+
+    func testFormatHexAddressPerformance() {
```
philipphofmann, line 25: `m`: `Measure` doesn't work in CI, as the baselines only work on the same computer config as you set them. If two engineers have the same MacBook with the same CPU and RAM, `measure` will fail if some code gets too slow. You didn't commit any baseline files, so it wouldn't work anyway. This test now only slows down CI, so we could also remove it.

Sources/Sentry/include/SentryFormatter.h
```diff
@@ -0,0 +1,36 @@
+#import <Foundation/Foundation.h>
+
+// 2 for the 0x prefix, plus 16 for the hex value, plus 1 for the null terminator
```
philipphofmann, line 3: `m`: Please document here somewhere how much faster `snprintf` is than `[NSString stringWithFormat:@"0x%016llx", [value unsignedLongLongValue]];`, so we know why we did this and don't change it back in the future.

philipphofmann (commented):
I'm surprised that `snprintf` is much faster than `sentry_formatHexAddress`. Thanks for the improvement.

Tests/SentryTests/SentryFormatterTests.swift
armcknight: Ah yeah, I only had this in there to compare some alternatives to snprintf. I;ll remove this.

Sources/Sentry/include/SentryFormatter.h
armcknight: I'll add something like the PR description here!

armcknight:
> I'm surprised that `snprintf` is much faster than `sentry_formatHexAddress`. Thanks for the improvement.

I think it comes down to all the extra allocations around NSStrings (even for the literal format string) and conversion of NSNumber back to an unsigned long long, both of which are avoided in the new version (and avoiding boxing the ull's to NSNumbers in the first place solely to pass them to this function is an extra benefit).

Sources/Sentry/include/SentryFormatter.h
armcknight: added here: https://github.com/getsentry/sentry-cocoa/pull/2863/commits/8a345884c4e1a32453fc7dd832e5e576d9c66cc1

philipphofmann (commented):
@armcknight, what's stopping us from making this ready for review?

armcknight:
@philipphofmann this one's ready for you now!

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes 
 
 - Ensure the current GPU frame rate is always reported for concurrent transaction profiling metrics (#2929)
+- Improved performance serializing profiling data (#2863)
```
philipphofmann, line 8: `m`: I think that needs to move up a bit.

philipphofmann (approved):
Thanks, @armcknight, for this performance improvement 👏 😃 . The changelog needs some fixes, I believe. LGTM.

CHANGELOG.md
armcknight: Are you looking at ordering by PR number? I haven't really cared about it too much, but am happy to start.

armcknight:
The only CI failures were codecov, but only because the test coverage was not properly assessed for the header implementations. There is indeed test coverage for them.

CHANGELOG.md
philipphofmann: No, I thought the entry was in an already-released section, but I maybe was mistaken; sorry.

PR #2864, opened 2023-04-06
ref: remove some dead code

armcknight:
I was looking at a couple places where we use a dynamic key to insert values into dictionaries as possible candidates for an exception to be thrown, and found a couple macros that expand to such code, but don't appear to be used in the codebase.

#skip-changelog


armcknight:
tests should pass after merging https://github.com/getsentry/sentry-cocoa/pull/2867

philipphofmann (approved):
LGTM, it's always great to see dead code deleted 😀.

PR #2866, opened 2023-04-06
ref: convert an sprintf call to snprintf

armcknight:
I tried running some tests with the address sanitizer and compilation failed because of these usages of `sprintf`. `stdio.h` now declares it with the following deprecation notice:

> This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.

I'm not sure why that only surfaces when running with ASAN.


armcknight:
tests should pass after merging https://github.com/getsentry/sentry-cocoa/pull/2867

Sources/SentryCrash/Recording/SentryCrashReport.c
```diff
@@ -913,14 +913,16 @@ writeNotableStackContents(const SentryCrashReportWriter *const writer,
...
     }
     uintptr_t contentsAsPointer;
-    char nameBuffer[40];
+#define SENTRY_NAME_BUFFER_SIZE 40
+    char nameBuffer[SENTRY_NAME_BUFFER_SIZE];
     for (uintptr_t address = lowAddress; address < highAddress; address += sizeof(address)) {
         if (sentrycrashmem_copySafely(
                 (void *)address, &contentsAsPointer, sizeof(contentsAsPointer))) {
-            sprintf(nameBuffer, "stack@%p", (void *)address);
+            snprintf(nameBuffer, SENTRY_NAME_BUFFER_SIZE, "stack@%p", (void *)address);
```
philipphofmann, line 921: `m`: I would prefer adding a test to cover these lines of code to ensure we are not causing any crashes.

philipphofmann (commented):
Thanks for the refactoring, @armcknight. Let's please add a test ensuring we don't crash in `SentryCrashReport`.

Sources/SentryCrash/Recording/SentryCrashReport.c
armcknight: oof, the other one was pretty simple to write a test case for, but this one is tougher. so far we have no coverage of this entire function, and it's gonna be fun to mock up a fixture: it will basically be an entire description of a stack trace formatted in the structs used by the crash reporter. will work more on this.

Tests/SentryTests/SentryCrash/SentryCrashJSONCodec_Tests.m
```diff
@@ -1672,4 +1672,20 @@ - (void)testDontCloseLastContainer
     return result;
 }
 
+- (void)testFastUIntEncode
```
philipphofmann, line 1675: `m`: Thanks for adding the test, but I think it's not working properly in CI.

armcknight: fixed

Sources/SentryCrash/Recording/SentryCrashReport.c
armcknight: I'm going to back out this change and only leave in the one that I already covered in test. This one is too big of a lift right now.

Sources/SentryCrash/Recording/Tools/SentryCrashJSONCodec.c
```diff
@@ -342,8 +342,10 @@ sentrycrashjson_addUIntegerElement(
 {
     int result = sentrycrashjson_beginElement(context, name);
     unlikely_if(result != SentryCrashJSON_OK) { return result; }
-    char buff[30];
-    sprintf(buff, "%" PRIu64, value);
+#define SENTRY_UINT_BUFFER_SIZE 30
+    char buff[SENTRY_UINT_BUFFER_SIZE];
+    snprintf(buff, SENTRY_UINT_BUFFER_SIZE, "%" PRIu64, value);
+#undef SENTRY_UINT_BUFFER_SIZE
```
brustolin, line 348: l: what is the reasoning to use a define like this?

brustolin, line 348: l: for `addIntegerElement` and `addFloatingPointElemenet` we have this:
```C
char buff[50];
snprintf(buff, sizeof(buff), "%lg", value);    
```
Which is different from the `#define` approach. Can we use the same approach in all places? 
And if you choose to use the `#define` I prefer to have one for the file, instead of one every function.

brustolin (commented):
LGTM.
Just one suggestion

brustolin (approved):

Sources/SentryCrash/Recording/Tools/SentryCrashJSONCodec.c
armcknight: I like the `sizeof` approach better!

armcknight:
@philipphofmann I'm not really sure, I must have hit the button by accident.

philipphofmann:
So what's stopping us from merging it @armcknight 😃?

armcknight:
Nothing, tests pass so should be g2g!

PR #2876, opened 2023-04-07
fix: avoid screenshots with either dimension of 0

armcknight:
## :scroll: Description

I noticed a test error on a Runtime Issue breakpoint for the test case that tries to take a screenshot of a 0 width window, when setting up the graphics context. We should never even let it get that far. 

I'll still leave the check for the image size, just in case somehow a nonzero width/height window still produces an image like that. But I changed the `||` to a `&&` in the comparison to make sure that _both_ dimensions must be greater than zero; otherwise, there's not much customers could do with something like that.

Added new unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryScreenshot.m
+++ b/Sources/Sentry/SentryScreenshot.m
@@ -56,8 +56,7 @@
         if ([window drawViewHierarchyInRect:window.bounds afterScreenUpdates:false]) {
             UIImage *img = UIGraphicsGetImageFromCurrentImageContext();
             // this shouldn't happen now that we discard windows with either 0 height or 0 width,
-            // but still, we shouldn't send any images with either one. previously we'd allow a 0
-            // dimension. change the || to &&
+            // but still, we shouldn't send any images with either one.
             if (LIKELY(img.size.width > 0 && img.size.height > 0)) {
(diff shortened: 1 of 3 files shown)


brustolin (approved):
LGTM

Sources/Sentry/SentryScreenshot.m
```diff
@@ -43,11 +44,21 @@ - (void)saveScreenShots:(NSString *)imagesDirectoryPath
...
+            continue;
+        }
+        UIGraphicsBeginImageContext(size);
 
         if ([window drawViewHierarchyInRect:window.bounds afterScreenUpdates:false]) {
             UIImage *img = UIGraphicsGetImageFromCurrentImageContext();
-            if (img.size.width > 0 || img.size.height > 0) {
+            // this shouldn't happen now that we discard windows with either 0 height or 0 width,
+            // but still, we shouldn't send any images with either one. previously we'd allow a 0
+            // dimension. change the || to &&
```
armcknight, line 60:
```suggestion
            // this shouldn't happen now that we discard windows with either 0 height or 0 width,
            // but still, we shouldn't send any images with either one.
```

philipphofmann (approved):
Thanks, LGTM 👏

PR #2879, opened 2023-04-07
feat: Core data operation in the main thread

### LINKED ISSUE #2873: DB on the main thread: Core Data
Performing DB queries on the main thread can lead to performance issues. To enable a performance issue detector on the backend to detect such queries, extend existing DB instrumentation to report whether the query was executed on the main thread.

brustolin:
## :scroll: Description

Adding extra information to a core data span when running in the main thread

close #2873 

Unit tests

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/include/SentryNSProcessInfoWrapper.h
+++ b/Sources/Sentry/include/SentryNSProcessInfoWrapper.h
@@ -4,7 +4,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface SentryNSProcessInfoWrapper : NSObject
 
-@property (class, readonly) SentryNSProcessInfoWrapper *shared;
 @property (nonatomic, readonly) NSString *processDirectoryPath;
 @property (nullable, nonatomic, readonly) NSString *processPath;
 @property (readonly) NSUInteger processorCount;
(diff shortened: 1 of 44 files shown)


Sources/Sentry/include/SentryNSProcessInfoWrapper.h
```diff
@@ -4,6 +4,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface SentryNSProcessInfoWrapper : NSObject
 
+@property (class, readonly) SentryNSProcessInfoWrapper *shared;
```
armcknight, line 7: Is there any way we can do what you need without introducing more singletons? Can you let us know why it's better/necessary to convert this to a singleton?

brustolin: Yes its possible, by creating another instance of SentryNSProcessInfoWrapper.
`SentryNSProcessInfoWrapper` consist of 3 functions that pass forward the call to another API, it does not have any variables nor any behaviour. Why should we allocated memory to hold reference to it every time we need?

philipphofmann (commented):
Almost LGTM, thanks @brustolin 👍

philipphofmann (approved):
LGTM 🎉

PR #2885, opened 2023-04-11
chore: Fixing CI

brustolin:
Trying to understand what is going on with CI and iOS 12.

_#skip-changelog_


brustolin:
I change it too much just to revert and add the final solution. I will close this PR and open a new one.

PR #2889, opened 2023-04-11
Chore : Fixing CI

brustolin:
This PR changes two things:

- Remove UITests for iOS 12 from running in the GH simulators. We still run it with SauceLabs.
- Remove Unit tests for view hierarchy from running on iOS 12 tests. GH is hanging when running this tests. We have view hierarchy enabled for UI tests that run in Saucelabs and iOS 12.

_#skip-changelog_

### DIFF
--- a/Sentry.xcodeproj/xcshareddata/xcschemes/Sentry.xcscheme
+++ b/Sentry.xcodeproj/xcshareddata/xcschemes/Sentry.xcscheme
@@ -88,9 +88,6 @@
                <Test
                   Identifier = "SentryUIApplicationTests/test_applicationWithScenes_noWindow()">
                </Test>
-               <Test
-                  Identifier = "SentryViewHierarchyTests">
-               </Test>
             </SkippedTests>
          </TestableReference>
          <TestableReference
--- a/Sentry.xcodeproj/xcshareddata/xcschemes/Sentry.xcscheme
+++ b/Sentry.xcodeproj/xcshareddata/xcschemes/Sentry.xcscheme
@@ -88,6 +88,9 @@
                <Test
                   Identifier = "SentryUIApplicationTests/test_applicationWithScenes_noWindow()">
                </Test>
+               <Test
+                  Identifier = "SentryViewHierarchyTests">
+               </Test>
             </SkippedTests>
          </TestableReference>
          <TestableReference
--- a/Sources/Sentry/SentryAppStartTracker.m
+++ b/Sources/Sentry/SentryAppStartTracker.m
@@ -35,7 +35,9 @@ SentryAppStartTracker ()
 @property (nonatomic, assign) BOOL wasInBackground;
 @property (nonatomic, strong) NSDate *didFinishLaunchingTimestamp;
 @property (nonatomic, assign) BOOL enablePreWarmedAppStartTracing;
-
+#    if TEST
+@property (nonatomic) BOOL isRunning;
+#    endif
 @end
 
 @implementation SentryAppStartTracker
@@ -67,6 +69,9 @@ SentryAppStartTracker ()
         self.wasInBackground = NO;
         self.didFinishLaunchingTimestamp = [currentDateProvider date];
         self.enablePreWarmedAppStartTracing = enablePreWarmedAppStartTracing;
+#    if TEST
+        self.isRunning = NO;
+#    endif
     }
     return self;
 }
@@ -115,6 +120,10 @@ SentryAppStartTracker ()
 
 #    if SENTRY_HAS_UIKIT
     [self.appStateManager start];
+#    endif
+
+#    if TEST
+    self.isRunning = YES;
 #    endif
 }
 
@@ -208,8 +217,8 @@ SentryAppStartTracker ()
         SentrySDK.appStartMeasurement = appStartMeasurement;
     };
 
-    // With only running this once we know that the process is a new one when the following
-    // code is executed.
+// With only running this once we know that the process is a new one when the following
+// code is executed.
 // We need to make sure the block runs on each test instead of only once
 #    if TEST
     block();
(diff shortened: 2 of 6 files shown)


philipphofmann (approved):
LGTM

Tests/SentryTests/SentryViewHierarchyTests.swift
```diff
@@ -26,167 +34,185 @@ class SentryViewHierarchyTests: XCTestCase {
     }
 
     func test_Multiple_Window() {
-        let firstWindow = UIWindow(frame: CGRect(x: 0, y: 0, width: 10, height: 10))
-        let secondWindow = UIWindow(frame: CGRect(x: 0, y: 0, width: 10, height: 10))
+        if #available(iOS 14, *) {
```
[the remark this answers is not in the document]
brustolin: Im still testing, but it looks like the 13 also have the problem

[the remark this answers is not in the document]
armcknight: You might be able to do this in an `setUpWithError` override instead of inserting it at the beginning of each test method, which diffs every line due to indentation and could be forgotten for new test methods added in the future.

```swift
override func setUpWithError() throws {
	try super.setUpWithError()
	guard #available(iOS 14, *) else {
		throw XCTSkip("Skipping for iOS >14")
	}
}
```

armcknight (commented):
Great plan, no need to run the extra compute! A couple suggestions for you.

Tests/SentryTests/Integrations/Performance/AppStartTracking/SentryAppStartTrackingIntegrationTests.swift
brustolin, line 1: The only thing `SentryAppStartTrackingIntegration` do is starting an `SentryAppStartTracker`.
I changed this tests to verify it the integration is properly setting up the tracker. 
Tests for `SentryAppStartTracker` behavior are being done in `SentryAppStartTrackerTests`, here we have no control over `SentrySysctl` which causes flake tests.

Sentry.xcodeproj/xcshareddata/xcschemes/Sentry.xcscheme
```diff
@@ -88,6 +88,9 @@
                <Test
                   Identifier = "SentryUIApplicationTests/test_applicationWithScenes_noWindow()">
                </Test>
+               <Test
+                  Identifier = "SentryViewHierarchyTests">
```
philipphofmann, line 92: `m`: Why do we need to skip this?

Tests/SentryTests/SentryViewHierarchyTests.swift
```diff
@@ -20,6 +20,21 @@ class SentryViewHierarchyTests: XCTestCase {
...
+
+        /**
+         * This test is making iOS 13 simulator hang in GH workflow,
+         * thats why we need to check for iOS 13 or later.
+         * By testing this in the other versions of iOS we guarantee the behavior
+         * mean while, running an iOS 12 sample with Saucelabs ensures this feature
+         * is not crashing the app.
+         */
+        if #available(iOS 13, *) {
+            throw XCTSkip("Skipping for iOS < 13")
```
philipphofmann, line 34: TIL 😃

Tests/SentryTests/Integrations/Performance/AppStartTracking/SentryAppStartTrackingIntegrationTests.swift
philipphofmann: @brustolin, this PR already improves CI a lot. Why not merge it and fix the issues in `SentryAppStartTrackerTests` in an extra PR?

brustolin: I was frustrating because CI was not green. I guess we could have separated in different PRs, but since the change is already here I don't think it worth going back.

Sentry.xcodeproj/xcshareddata/xcschemes/Sentry.xcscheme
brustolin: Im still doing tests. Im not 100% confident that the changes applied here are reliable.

I'll let you know when I think we can merge this for a last review.

Tests/SentryTests/Integrations/Performance/AppStartTracking/SentryAppStartTrackingIntegrationTests.swift
```diff
@@ -40,57 +40,63 @@ class SentryAppStartTrackingIntegrationTests: NotificationCenterTestCase {
...
     
-    func testAppStartMeasuringEnabledAndSampleRate_DoesUpdatesAppState() {
+    func testAppStartMeasuringEnabledAndSampleRate_properlySetupTracker() throws {
         sut.install(with: fixture.options)
+
+        let tracker = try XCTUnwrap(Dynamic(sut).tracker.asObject as? SentryAppStartTracker, "SentryAppStartTrackingIntegration should have a tracker")
+        try assertTrackerSetupAndRunning(tracker)
+    }
+
+    func testUnistall_stopTracker() throws {
```
philipphofmann, line 50: `l`:
```suggestion
    func testUnistall_stopsTracker() throws {
```

Sources/Sentry/SentryAppStartTracker.m
```diff
@@ -35,7 +35,9 @@
 @property (nonatomic, assign) BOOL wasInBackground;
 @property (nonatomic, strong) NSDate *didFinishLaunchingTimestamp;
 @property (nonatomic, assign) BOOL enablePreWarmedAppStartTracing;
-
+#    if TEST
+@property (nonatomic) BOOL isRunning;
+#    endif
```
philipphofmann, line 40: `m`: This class is internal; I would make this a public readonly property for SentryAppStartTracker. Then we don't need to use Dynamic for testing. You can keep the `Test` compiler directive, but I think removing it would also be fine.

philipphofmann (approved):
One minor comment. Thanks @brustolin 👏

Issue #2862, opened 2023-04-04
Migrate from `xcode-install`


armcknight:
### Description

We use [`xcode-install`](https://github.com/xcpretty/xcode-install) to install some older iOS simulators for test runs [here](https://github.com/getsentry/sentry-cocoa/blob/ff5c1d83bf601bbcd0f5f1070c3abf05310881bd/.github/workflows/test.yml#L174) and [here](https://github.com/getsentry/sentry-cocoa/blob/ff5c1d83bf601bbcd0f5f1070c3abf05310881bd/.github/workflows/test.yml#L343)

That project is being sunset, so we'll have to upgrade to the recommended fastlane [xcodes](https://github.com/RobotsAndPencils/xcodes) tool or other alternative. See https://github.com/xcpretty/xcode-install/blob/master/MIGRATION.md

philipphofmann:
Thanks for pointing that out @armcknight 👍.

PR #2891, opened 2023-04-12
ci: Remove iOS 12 simulator

philipphofmann:
Remove running unit tests on iOS 12 simulator as the xcode-install tool is being sunset. Instead, we now run UI tests in saucelabs on iOS 12.

Other usages of iOS 12 simulators are removed with https://github.com/getsentry/sentry-cocoa/pull/2889.

Fixes GH-2862

#skip-changelog

### DIFF
--- a/.github/workflows/saucelabs-UI-tests.yml
+++ b/.github/workflows/saucelabs-UI-tests.yml
@@ -106,6 +106,9 @@ jobs:
           - xcode: '13.4.1'
             suite: 'iOS-13'
 
+          - xcode: '13.4.1'
+            suite: 'iOS-12'
+
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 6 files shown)


brustolin (commented):
I'm all for it, but then, shouldn't we run unit tests in sauce labs?
We only run UI tests there and I believe our UI tests dont cover every aspect of the SDK.
Some swizzlings like URL Session, may have different implementation from OS to OS.

philipphofmann:
> I'm all for it, but then, shouldn't we run unit tests in sauce labs? We only run UI tests there and I believe our UI tests dont cover every aspect of the SDK. Some swizzlings like URL Session, may have different implementation from OS to OS.

I think the sizzling of UIViewControllers shouldn't be a problem, as we have good UI test coverage for UIViewControllers. Do you think our UI tests would surface bugs when swizzling the URL session?

brustolin:
> Do you think our UI tests would surface bugs when swizzling the URL session?

Im not sure about it. But we know url session swizzling is work right now. We should be careful when changing it next time.

brustolin (approved):

brustolin (approved):
LGTM 🚀

PR #2895, opened 2023-04-13
Fix: SentryDependencyContainer processInfoWrapper

brustolin:
SentryDependencyContainer processInfoWrapper property was wrapped by `#if SENTRY_HAS_METRIC_KIT` and did not work for tvOS.

_#skip-changelog_


philipphofmann (approved):
LGTM

PR #2896, opened 2023-04-13
ref: Use sample decision instead of bool

philipphofmann:
Use sample decision instead of bool in SentrySpanContext.

#skip-changelog


philipphofmann:
Tests will succeed after merging https://github.com/getsentry/sentry-cocoa/pull/2895.

brustolin (approved):

PR #2898, opened 2023-04-13
fix: Start frame check removed

brustolin:
We stopped by mistake marking the start frame after an async frame marker.
Rolling back the code to fix this.

_#skip-changelog_


philipphofmann (approved):
LGTM

PR #2899, opened 2023-04-13
test: split crash reporter tests into separate target

armcknight:
- reimplement some ojbc classes needed to move into the shared SentryTestUtils lib in swift since you cant mix them in a static library
- reimplement some swift classes into objc since they were used from other objc files and i couldn't get the "-Swift.h" importation working
- fixed some test assertions that started failing after making changes, not sure how they ever worked in the first place
- fixed a bug with async frames in stack trace builder, also not sure how the associated test was passing

TODO: 
- [ ] figure out why SentryCrashInstallationTests.testUninstall` is failing, need some logging because it passes locally
- [ ] give the tests a little more time to run, the Xcode 13 variants are timing out

#skip-changelog


Tests/SentryTests/SentryOptionsTest.m
```diff
@@ -1,8 +1,13 @@
 #import "SentryOptions.h"
+#import "SentryDSN.h"
 #import "SentryError.h"
+#import "SentryEvent.h"
+#import "SentryHTTPStatusCodeRange.h"
 #import "SentryOptions+HybridSDKs.h"
+#import "SentryOptions+Private.h"
 #import "SentrySDK.h"
-#import "SentryTests-Swift.h"
```
armcknight, line 5: This stopped working for some reason, which is why i had to rewrite UrlSessionDelegateSpy in objc.

Tests/SentryTests/Integrations/SentryCrash/SentryCrashIntegrationTests.swift
```diff
@@ -328,7 +328,7 @@ class SentryCrashIntegrationTests: NotificationCenterTestCase {
     }
     
     private func givenSutWithGlobalHubAndCrashWrapper() -> (SentryCrashIntegration, SentryHub) {
-        let sut = fixture.getSut(crashWrapper: SentryCrashWrapper.sharedInstance())
+        let sut = fixture.getSut(crashWrapper: SentryCrashWrapper()) // ???: should this be using TestSentryCrashWrapper?
```
armcknight, line 331: Question here. At some point we should also split out a separate test target for integration tests, which do not use mock subclasses of live SDK components.

Tests/SentryCrashTests/SentryStacktraceBuilderTests.swift
```diff
@@ -14,6 +14,11 @@ class SentryStacktraceBuilderTests: XCTestCase {
 
     private var fixture: Fixture!
 
+    override class func setUp() {
+        super.setUp()
+        SentryLog.configure(true, diagnosticLevel: .debug)
+    }
+
```
armcknight, line 21: TODO: i can remove all the logging here now that we figured out what the latent bug was that was causing this test case to fail after i made changes. 

I'm scratching my head trying to figure out how the test hadn't failed before 🤔 I didn't make any changes to any SDK logic as part of this PR. I did fix the bug in this PR though, see https://github.com/getsentry/sentry-cocoa/pull/2899/files#diff-84fe6a903bd6efcc5209502d4c66c80e68a8313d0d94927cb347d3b0015eda1fR49

Tests/SentryCrashTests/SentryCrashSysCtl_Tests.m
```diff
@@ -198,7 +198,7 @@ - (void)testSysCtlCurrentProcessStartTime
 - (void)testGetProcessInfo
 {
     int pid = getpid();
-    struct kinfo_proc procInfo = { { { { 0 } } } };
+    struct kinfo_proc procInfo = { { 0 }, { 0 } };
```
armcknight, line 201: This was a weird one that started failing the build.

Tests/SentryCrashTests/SentryCrashInstallationTests.swift
```diff
@@ -0,0 +1,99 @@
+import SentryTestUtils
+import XCTest
+
+class SentryCrashTestInstallation: SentryCrashInstallation {
+    override init() {
+        super.init(requiredProperties: [])
+    }
+}
+
+final class SentryCrashInstallationTests: XCTestCase {
```
armcknight, line 10: Rewritten in swift, see my other comment on the old objc implementation that was removed.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitorType.h
```diff
@@ -92,8 +94,6 @@ typedef enum {
 #define SentryCrashMonitorTypeRequired                                                             \
     (SentryCrashMonitorTypeSystem | SentryCrashMonitorTypeApplicationState)
 
-#define SentryCrashMonitorTypeNone 0
```
armcknight, line 95: This wouldn't work in the new swift implementation of an objc impl that was using this, so i changed it to be a member of the enum instead.

Sources/Sentry/SentryStacktraceBuilder.m
```diff
@@ -33,15 +34,20 @@ - (SentryStacktrace *)retrieveStacktraceFromCursor:(SentryCrashStackCursor)stack
...
             if (frame != nil) {
                 frame.stackStart = @(YES);
             }
             // skip the marker frame
+            SENTRY_LOG_DEBUG(@"Skipping async continuation marker frame.");
             continue;
         }
         if (stackCursor.symbolicate(&stackCursor)) {
-            [frames addObject:[self.crashStackEntryMapper mapStackEntryWithCursor:stackCursor]];
+            frame = [self.crashStackEntryMapper mapStackEntryWithCursor:stackCursor];
```
armcknight, line 49: Fixes a bug where it never detected async start frames.

SentryTestUtils/TestSentryUIDeviceWrapper.swift
```diff
@@ -0,0 +1,26 @@
+import Sentry
+
+public class TestSentryUIDeviceWrapper: SentryUIDeviceWrapper {
```
armcknight, line 3: This wasn't rewritten, it was only renamed/moved, but didn't diff correctly. No changes in this file.

SentryTestUtils/TestCrashWrapper.swift
```diff
@@ -0,0 +1,64 @@
+import Foundation
+import Sentry
+
+public class TestSentryCrashWrapper: SentryCrashWrapper {
```
armcknight, line 4: Rewritten in swift, see later comment on the old objc implementation.

brustolin:
> fixed a bug with async frames in stack trace builder, also not sure how the associated test was passing

This was fixed at #2898

philipphofmann:
I'm sorry I didn't get to that PR yet. I will give you a review on Monday.

Sources/Sentry/SentryCrashWrapper.m
```diff
@@ -13,14 +13,6 @@
 
 @implementation SentryCrashWrapper
 
-+ (instancetype)sharedInstance
```
philipphofmann, line 16: `m`: I think this change could be in an extra PR as it makes sense without all the other changes. Even if we agree not to proceed with this PR, this change would make it to the main branch.

Sources/Sentry/SentryStacktraceBuilder.m
philipphofmann: Can we put this in an extra PR, please?

Tests/SentryCrashTests/SentryCrashReportFilter_Tests.m
```diff
@@ -155,7 +155,6 @@ - (void)testPipelineInit
     id<SentryCrashReportFilter> filter = [[SentryCrashReportFilterPipeline alloc]
         initWithFilters:[SentryCrashReportFilterPassthrough filter],
         [SentryCrashReportFilterPassthrough filter], nil];
-    filter = filter;
```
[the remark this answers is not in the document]
philipphofmann: Thanks for the comment 😃

Tests/SentryTests/Helper/UrlSessionDelegateSpy.swift
```diff
@@ -1,17 +0,0 @@
-import Foundation
-
-class UrlSessionDelegateSpy: NSObject, URLSessionDelegate {
```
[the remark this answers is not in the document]
philipphofmann: It's such a small class so it should be fine.

Tests/SentryTests/Integrations/SentryCrash/SentryCrashIntegrationTests.swift
philipphofmann: Yes, that could make sense. If the test succeeds with `TestSentryCrashWrapper`, please use it.

Tests/SentryTests/SentryOptionsTest.m
philipphofmann: Thanks for the info. We can fix it in another PR.

Tests/SentryCrashTests/SentryFrameRemoverTests.swift
```diff
@@ -60,16 +60,31 @@ class SentryFrameRemoverTests: XCTestCase {
...
     func testOnlySdkFrames_AllFramesRemoved() {
         let actual = SentryFrameRemover.removeNonSdkFrames(fixture.sentryFrames)
-        XCTAssertEqual(fixture.sentryFrames, actual)
+        XCTAssert(equivalent(expected: fixture.sentryFrames, actual: actual))
+    }
+}
+
+private extension SentryFrameRemoverTests {
+    /// Compare elements pairwise to see if the two input arrays are equivalent or not.
+    func equivalent(expected: [Frame], actual: [Frame]) -> Bool {
```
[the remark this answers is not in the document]
philipphofmann: What you did is okay. I also don't understand why the tests worked before.

Tests/SentryCrashTests/SentryFrameRemoverTests.swift
```diff
@@ -60,16 +60,31 @@ class SentryFrameRemoverTests: XCTestCase {
...
+        guard expected.count == actual.count else { return false }
+        for i in 0..<expected.count {
+            let nextExpected = expected[i]
+            let nextActual = actual[i]
+            if nextExpected.package != nextActual.package {
```
philipphofmann, line 84: `m`: I guess we shouldn't only just compare the package only. You could use: https://github.com/getsentry/sentry-cocoa/blob/74192858dff8cf729ce015b06fe9d3ff64fe7b74/Tests/SentryTests/Protocol/SentryThreadEquality.swift#L35-L36

philipphofmann (commented):
Thanks @armcknight for this great refactoring. We need to call all test targets in CI. It think it makes sense to run the test targets subsequentially for each matrix config for the unit tests https://github.com/getsentry/sentry-cocoa/blob/74192858dff8cf729ce015b06fe9d3ff64fe7b74/.github/workflows/test.yml#L86-L90

otherwise, we end up with many more GH jobs. Maybe it makes sense to create multiple raw log artifacts per test target. We also need to ensure that the DerivedData Logs still work properly, and that the code coverage is not messed up.

FYI, I'm going to remove this piece of bash script as it's not required anymore this PR https://github.com/getsentry/sentry-cocoa/pull/2918 https://github.com/getsentry/sentry-cocoa/blob/74192858dff8cf729ce015b06fe9d3ff64fe7b74/scripts/xcode-test.sh#L61-L69

Sources/Sentry/SentryStacktraceBuilder.m
armcknight: It was already in #2898, just need to merge main into this branch[

Sources/Sentry/SentryCrashWrapper.m
armcknight: see https://github.com/getsentry/sentry-cocoa/pull/2941

brustolin:
@armcknight As discussed during a meeting. Can we close this PR?

PR #2901, opened 2023-04-14
test: Make TestLogOutput thread safe

philipphofmann:
When running the tests locally, I got a SIGABRT in the TestLogOutput when writing to the loggedMessages. I added a test to reproduce the same error by writing and reading the loggedMessages from different threads. The problem is fixed now by using a dispatch queue.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/TestLogOutput.swift
+++ b/Tests/SentryTests/TestLogOutput.swift
@@ -1,9 +1,50 @@
 import Foundation
 
 class TestLogOutput: SentryLogOutput {
-    var loggedMessages: [String] = []
+    
+    private let queue = DispatchQueue(label: "TestLogOutput", attributes: .concurrent)
+    
+    private var _loggedMessages: [String] = []
+    
+    var loggedMessages: [String] {
+        get {
+            queue.sync {
+                return _loggedMessages
+            }
+        }
+    }
+    
     override func log(_ message: String) {
         super.log(message)
-        loggedMessages.append(message)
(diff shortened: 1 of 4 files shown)


Tests/SentryTests/TestLogOutput.swift
```diff
@@ -1,9 +1,35 @@
...
+        }
+    }
+}
+
+class TestLogOutPutTests: XCTestCase {
+    
+    func testLoggingFromMulitpleThreads() {
+        let sut = TestLogOutput()
+        testConcurrentModifications(writeWork: { i in
+            sut.log("Some message \(i)")
```
brustolin, line 30: m: This will add too much unnecessary information in the actual log we use to debug CI.
Can we do something about it?

brustolin (approved):
Thanks for fixing this.

I have only one concern.

Tests/SentryTests/TestLogOutput.swift
philipphofmann: `sut` is an extra instance of the `TestLogOutput`. It isn't a global one. All the other tests making use of it use their own instance, see https://github.com/search?q=repo%3Agetsentry%2Fsentry-cocoa%20%3D%20TestLogOutput()&type=code.

brustolin: I check the CI log and is really not there, so we're good to go.

But it does write to the log, and running locally I can see it in the console.
```
super.log(message)
```

I dont understand why this is not in the CI log

PR #2904, opened 2023-04-14
test: Remove subclass of NSTimer

philipphofmann:
We subclassed NSTimer, which is not allowed see:
https://developer.apple.com/documentation/foundation/nstimer#1770465.
This sometimes leads to crashes in TestSentryNSTimerWrapper with EXC_BAD_ACCESS.
The subclass is now removed.

#skip-changelog

### DIFF
--- a/SentryTestUtils/TestSentryNSTimerWrapper.swift
+++ b/SentryTestUtils/TestSentryNSTimerWrapper.swift
@@ -9,14 +9,15 @@ public class TestSentryNSTimerWrapper: SentryNSTimerWrapper {
     
     public var timer: Timer {
         get {
-            _timer!
+            _timer ?? Timer()
         }
     }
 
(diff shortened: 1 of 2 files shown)


philipphofmann:
cc @armcknight I think that was the reason why the tests sometimes crashed here
https://github.com/getsentry/sentry-cocoa/blob/8c50edbfdb94d6de171a52917ed9e69144465cf5/SentryTestUtils/TestSentryNSTimerWrapper.swift#L25

and not because we used the lazy keyword.

brustolin (approved):

armcknight:
Huh, could you link to an example of the crash? I haven't seen that one. Good learning though, thanks!

philipphofmann:
> Huh, could you link to an example of the crash? I haven't seen that one. Good learning though, thanks!

I noticed that when running the tests locally, sometimes the tests crashed with `EXC_BAD_ACCESS`. If you checkout a commit previous to the merge of this PR and run all tests locally a couple of times, you should experience the issue.

PR #2906, opened 2023-04-14
test: give the xcode 13 tests a little more time

armcknight:
#skip-changelog

These test jobs are just barely timing out, see the latest run against `main`: https://github.com/getsentry/sentry-cocoa/actions/runs/4698327161/jobs/8330707336

in one case the test even succeeded but timed out right before the script finished.

PR #2907, opened 2023-04-15
test: use dispatch queue wrappers for all file managers in tests

armcknight:
#skip-changelog

Following a finding in #2906, init the remaining live file managers in the test suite with test dispatch queue wrappers, to keep them from delaying dispatches that cause issues in later test cases.

We should consider if some/all of these could also instead be converted to test file manager subclasses instead of the SDK file manager.


philipphofmann (approved):
LGTM

PR #2916, opened 2023-04-17
Chore: Removing integration tests

brustolin:
Both Alamofire and Home Assistant integration tests are causing more trouble then helping, we will remove this CI tests.

_#skip-changelog_


armcknight:
I'll admit I haven't really looked at their results; what issues do they cause? Is it something we could run only after merging to main instead of for every PR push?

brustolin:
> Is it something we could run only after merging to main instead of for every PR push?

TBH, I cant figure it out the problem, this tests went from working flawlessly to failing every time in CI with no apparent reason.
I don't think the headache worth keeping this tests.

philipphofmann:
@armcknight, we have had these integration tests for roughly a year, and they didn't surface any bugs in our code. We have them to ensure we are not breaking the functionality of UIViewControllers, NSURLSession, NSData, after swizzling. Right now, we know that the swizzling is working fine, and we didn't have to touch the swizzling logic for a year or so. Instead of maintaining these integration tests, we decided to lock these files with no-changes-in-high-risk-files. The next time we have to do some changes, we can add the integration tests again or run them manually.

develop-docs/README.md
```diff
@@ -225,3 +225,16 @@ For the swizzling of UIViewControllers and NSURLSession, we have UI tests runnin
 tests on iOS 12 simulators is acceptable. This decision reverts [manually installing iOS 12 simulators](#ios-12-simulators).
 
 Related to [GH-2862](https://github.com/getsentry/sentry-cocoa/issues/2862) and 
+
+### Remove integration tests from CI <a name="remove-integration-tests-from-ci"></a>
```
philipphofmann, line 229: 👏

scripts/no-changes-in-high-risk-files.sh
```diff
@@ -3,14 +3,19 @@ set -euo pipefail
...
+52cb473dcc8d13c0d4f6cd1429c3fc6e8588521660b714f4a2edb4eaf1401e9f  ./Sources/Sentry/SentryUIViewControllerSwizzling.m
+e95e62ec7363984f20c78643bb7d992a41a740f97e1befb71525ac34caf88b37  ./Sources/Sentry/SentryNSDataSwizzling.m
+9ad05dd8dd29788cba994736fdcd3bbde59a94e32612640d11f4f9c38ad6610e  ./Sources/Sentry/SentrySubClassFinder.m"
 
 if [ "$ACTUAL" = "$EXPECTED" ]; then
     echo "No changes in high risk files."
     exit 0
 else
-    echo "Changes in high risk files. If your changes are intended please update the sha in ./scripts/no-changes-in-high-risk-files.sh to $ACTUAL."
+    echo "Changes in high risk files. Test your changes by using `make test-alamofire` and `make test-homekit` to ensure the changes are safe to run on third party projects."
```
philipphofmann, line 17: `l`: I think we should change this to the following, as testing the changes with `make test-alamofire` and  `make test-homekit` might not make sense for every change in these files.
```suggestion
    echo "Changes in high risk files. You might want to test your changes by using `make test-alamofire` and `make test-homekit` to ensure the changes are safe to run on third party projects."
```

scripts/no-changes-in-high-risk-files.sh
```diff
@@ -3,14 +3,19 @@ set -euo pipefail
...
+    echo "If your changes are intended and everything is running properly, please update the sha in ./scripts/no-changes-in-high-risk-files.sh to: "
+    echo "$ACTUAL"
```
philipphofmann, line 19: Thanks for improving this 💯

scripts/no-changes-in-high-risk-files.sh
```diff
@@ -3,14 +3,19 @@ set -euo pipefail
...
 # To update the sha run the command in ACTUAL and copy the result in EXPECTED.
 
-ACTUAL=$(shasum -a 256 ./Sources/Sentry/SentryNSURLSessionTaskSearch.m ./Sources/Sentry/SentryNetworkTracker.m)
+ACTUAL=$(shasum -a 256 ./Sources/Sentry/SentryNSURLSessionTaskSearch.m ./Sources/Sentry/SentryNetworkTracker.m ./Sources/Sentry/SentryUIViewControllerSwizzling.m ./Sources/Sentry/SentryNSDataSwizzling.m ./Sources/Sentry/SentrySubClassFinder.m)
 EXPECTED="819d5ca5e3db2ac23c859b14c149b7f0754d3ae88bea1dba92c18f49a81da0e1  ./Sources/Sentry/SentryNSURLSessionTaskSearch.m
-58d5414b4f0a4c821b20fc1a16f88bda3116401e905b7bc1d18af828be75e431  ./Sources/Sentry/SentryNetworkTracker.m"
+58d5414b4f0a4c821b20fc1a16f88bda3116401e905b7bc1d18af828be75e431  ./Sources/Sentry/SentryNetworkTracker.m
+52cb473dcc8d13c0d4f6cd1429c3fc6e8588521660b714f4a2edb4eaf1401e9f  ./Sources/Sentry/SentryUIViewControllerSwizzling.m
+e95e62ec7363984f20c78643bb7d992a41a740f97e1befb71525ac34caf88b37  ./Sources/Sentry/SentryNSDataSwizzling.m
+9ad05dd8dd29788cba994736fdcd3bbde59a94e32612640d11f4f9c38ad6610e  ./Sources/Sentry/SentrySubClassFinder.m"
```
philipphofmann, line 11: `h`: I would also add `SentryCoreDataSwizzling.m`, `SentrySwizzleWrapper.m`, `SentrySwizzle.h`, and `SentrySwizzle.m` to the list.

scripts/test-alamofire.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
philipphofmann, line 1: `m`: It would be great to keep the explanation of 
https://github.com/getsentry/sentry-cocoa/blob/89b12ebb5452bf2787046df229892098a87e33d6/.github/workflows/integration-tests.yml#L16-L22

We could add it as a comment here.

scripts/test-homekit.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
philipphofmann, line 1: `m`: It would be great to keep the explanation of 
https://github.com/getsentry/sentry-cocoa/blob/89b12ebb5452bf2787046df229892098a87e33d6/.github/workflows/integration-tests.yml#L74-L77

We could add it as a comment here.

philipphofmann (commented):
Thanks for doing this @brustolin 👏

philipphofmann (approved):
Thanks @brustolin 👍. LGTM

PR #2919, opened 2023-04-17
separate the build and test xcodebuild tasks; print dates before/between

armcknight:
#skip-changelog

Saw some test timeouts even after merging #2907, which seemed to have fixed the issue.

Looking at the logs from the job that failed immediately after merging, the tests only had around 11 seconds to run before it timed out, so the build must've taken a long time. 

I broke up the build and test steps and print `date` before and after the build step to see if it's taking too long, and we can think about what to do from there.

### DIFF
--- a/scripts/xcode-test.sh
+++ b/scripts/xcode-test.sh
@@ -43,7 +43,7 @@ case $REF_NAME in
     "main")
         CONFIGURATION="TestCI"
         ;;
-    
+
     *)
         CONFIGURATION="Test"
         ;;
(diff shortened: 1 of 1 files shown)


philipphofmann:
Maybe `xcpretty` is swallowing something or causing delays. I would check the raw test log. If `xcpretty` is causing delays, I think we should ditch it.

philipphofmann:
@armcknight, I replaced xcpretty with xcbeautify in https://github.com/getsentry/sentry-cocoa/pull/2931. I'm unsure if this PR is still needed.

philipphofmann:
I have a workaround now with https://github.com/getsentry/sentry-cocoa/pull/2933. I just disabled the tests below iOS 15. They work fine locally on iOS 14.

I think we can close this PR @armcknight.

philipphofmann:
Again https://github.com/getsentry/sentry-cocoa/pull/2933 doesn't solve the problem. I'm now setting the log level to debug for all tests with https://github.com/getsentry/sentry-cocoa/pull/2938. I hope this will give us more detail on why the tests hang for so long.

PR #2924, opened 2023-04-18
ref: Remove SentrySwizzleWrapper.sharedInstance

philipphofmann:
Only the dependency container uses the shared instance. Therefore, we can remove it as the dependency container keeps track of the single instance.

#skip-changelog


philipphofmann:
The refactoring doesn't make sense, as the SentrySwizzleWrapper must only have one instance, cause the sizzling logic needs static access.

PR #2928, opened 2023-04-18
chore: Organize sample

brustolin:
Organize iOS-Swift sample by dividing diverent types of operations in different tabs given more room for more tests.

The result: 
sample

_#skip-changelog_


philipphofmann (commented):
Can we please make the buttons a bit bigger at the bottom?

Screenshot 2023-04-21 at 12 05 02

It would be great if it looked a bit more navigation like.

marandaneto:
@brustolin can we improve the labels to be a bit more clear of the outcome when clicking on the buttons?
* What is `NiB`?
* What should we see on Sentry when clicking on TableView, SplitView, CoreData, Performance scenarios, etc...

Maybe we could do always it with an intention as a prefix/suffix.
For example, CoreData spans, Table view (add intention), etc.
I'd need to read the source code to understand what they are doing as of now.

brustolin:
@marandaneto  I think we can improve this a little bit.

Most of this actions don't really matter for a booth sample, we use to open the various types of view controllers to make sure our code don't break anything, but most of the view controllers actions will just create a transaction.

brustolin:
> Can we please make the buttons a bit bigger at the bottom?

We can choose an Icon to go along, this is the default tab item font size

Samples/iOS-Swift/iOS-SwiftUITests/LaunchUITests.swift
```diff
@@ -11,15 +11,14 @@ class LaunchUITests: XCTestCase {
...
         waitForExistenceOfMainScreen()
-        checkSlowAndFrozenFrames()
     }
     
     override func tearDown() {
         app.terminate()
         super.tearDown()
     }
 
-    func testCrashRecovery() {
```
armcknight, line 22: is this a typo?

Samples/iOS-Swift/iOS-SwiftUITests/LaunchUITests.swift
```diff
@@ -11,15 +11,14 @@ class LaunchUITests: XCTestCase {
         app.launch()
         
         waitForExistenceOfMainScreen()
-        checkSlowAndFrozenFrames()
```
armcknight, line 14: Why won't we check this for every test case anymore?

brustolin: Because we would need to be in the last tab after every test.
Its this relevante? 
IMO we just need to check with one test whether slow and frozen frames are being recorded.

Samples/iOS-Swift/iOS-SwiftUITests/LaunchUITests.swift
```diff
@@ -11,15 +11,14 @@ class LaunchUITests: XCTestCase {
...
     }
     
     override func tearDown() {
         app.terminate()
         super.tearDown()
     }
 
-    func testCrashRecovery() {
+    func atestCrashRecovery() {
```
brustolin, line 21:
```suggestion
    func testCrashRecovery() {
```

Samples/iOS-Swift/iOS-SwiftUITests/LaunchUITests.swift
armcknight: Fair enough, one test check should be fine 👍🏻

armcknight (approved):
LGTM, I think we should forego any other cosmetic concerns, which can be iterated on in subsequent PRs. I still see this more as a test harness than demo app so am personally not concerned with that kind of thing.

brustolin:
Updated version of the button labels and tabs

sample

philipphofmann (approved):
Thanks @brustolin 👏

PR #2929, opened 2023-04-18
fix: always report a frame rate with GPU profiler metrics

armcknight:
## :scroll: Description

In cases where subsequent concurrent transactions will contain slow/frozen GPU frame metrics for profiles, but the frame rate would have already been recorded for the first concurrent transaction, the profile payloads for the subsequent ones would have discarded the earlier frame rate data in slicing.

This makes sure the current frame rate is reported for such cases, instead of nothing.

The UI test has been failing recently.

The current UI test now passes; the unit test for concurrent transactions still passes as well, with the additional change that we no longer manually reset the frame tracker's profiling data: that's now allowed to be managed by the component itself. There's also no longer a hardcoded logic to always record frame rate information during tests only in the frames tracker with conditional compilation.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryFramesTracker.m
+++ b/Sources/Sentry/SentryFramesTracker.m
@@ -146,22 +146,18 @@ slowFrameThreshold(uint64_t actualFramesPerSecond)
     }
 
 #    if SENTRY_TARGET_PROFILING_SUPPORTED
-#        if defined(TEST) || defined(TESTCI)
-    BOOL shouldRecordFrameRates = YES;
-#        else
-    BOOL shouldRecordFrameRates = [SentryProfiler isRunning];
-#        endif // defined(TEST) || defined(TESTCI)
-    BOOL hasNoFrameRatesYet = self.frameRateTimestamps.count == 0;
-    uint64_t previousFrameRate
-        = self.frameRateTimestamps.lastObject[@"value"].unsignedLongLongValue;
-    BOOL frameRateChanged = previousFrameRate != currentFrameRate;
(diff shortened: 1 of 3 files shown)


Sources/Sentry/SentryFramesTracker.m
```diff
@@ -146,22 +146,18 @@ - (void)displayLinkCallback
     }
 
 #    if SENTRY_TARGET_PROFILING_SUPPORTED
-#        if defined(TEST) || defined(TESTCI)
```
brustolin, line 149: l: Is this not meant to run only for tests anymore? Or you forgot to put this back by mistake?

brustolin:
Ci is broken, I believe some changes to tests are also needed.

Sources/Sentry/SentryFramesTracker.m
armcknight: No, this was deliberately removed, the change I made now makes it work correctly for tests. Previously, because slicing only took datapoints recorded strictly within the tx timeframe, there would be no frame rate info in a bunch of payloads, so this line made it such that we just had tons of them. It was a hack at the time, glad we could remove it.

armcknight: ah, i might've misremembered. needed this because frame rates are only recorded if the profiler is running, which it is not in the frames tracker tests. i'll try running it for those tests.

armcknight:
fixed @brustolin !

brustolin (approved):
Thanks for this!!
LGTM

PR #2936, opened 2023-04-19
test: Clean up starting SDK in sample apps

philipphofmann:
Remove turning on features that are on by default. Remove comment to adjust traces sample rate in production, as we recommend setting it to 1.0 anyways since having dynamic sampling.

#skip-changelog


brustolin (approved):

armcknight (approved):
👏🏻

Issue #1892, opened 2022-06-16
Xcode organizer reports crash on `writeBinaryImage`


jaroslavas:
### Platform

iOS

### Installed

CocoaPods

### Version

7.16.0

### Steps to Reproduce

Add sentry to app and start reporting issues.

```
SentrySDK.start { (options) in
            options.dsn = dsn
            options.debug = false
            options.enableAutoSessionTracking = false
            options.attachStacktrace = false
            options.maxBreadcrumbs = 25

            options.beforeSend = { event in
                // Some data is removed to make reports lighter
                event.debugMeta = nil
                return event
            }
        }
```

### Expected Result

Crashes should be seen in sentry as well as in XCode organizer.

### Actual Result

XCode organizer starts showing a lot of different crashes as one, where Sentrys `writeBinaryImage` is shown as the cause for the crash. 
![Screenshot 2022-06-16 at 10 11 22](https://user-images.githubusercontent.com/1903354/174014454-ccd3b806-cc97-426c-9dec-39af6cce7b8a.png)
![Screenshot 2022-06-16 at 10 11 46](https://user-images.githubusercontent.com/1903354/174014461-b5e00645-ef8b-4c69-b2cd-0ef6b18478d6.png)

philipphofmann:
What is the error message of the crash, @jaroslavas? When does this happen? Do you maybe have a repro for us?

jaroslavas:
Nope, I only have different stack traces ending in a crash by sentry and binary message, either writing to it or reading:
![Screenshot 2022-06-17 at 12 43 47](https://user-images.githubusercontent.com/1903354/174273458-c8666cb7-0cee-4d3d-b409-287c9c433b99.png)
![Screenshot 2022-06-17 at 12 43 59](https://user-images.githubusercontent.com/1903354/174273473-3edfc3d4-7005-4d7d-bf59-5f873e2d8b0c.png)

Unfortunately I can't reproduce this, only see a lot of reports in organizer from live users.

philipphofmann:
> a lot of reports in organizer from live users.

What exactly do you mean by that?

brustolin:
It means crash reports downloaded from iTunes Connect.

philipphofmann:
Did this start happening in `7.16.0` or did you see this problem for earlier versions as well, @jaroslavas?

philipphofmann:
[...]

PR #2939, opened 2023-04-20
fix: Cache binary images to be used for crashes

brustolin:
## :scroll: Description

During crashes we use 2 async-signal-unsafe functions `_dyld_get_image_header` and `_dyld_get_image_name`.
This may lead to a crash report not being properly saved to disk.

This proposal will cache binary images during runtime and use it when the app crashes.

Solves #1892 

## :pencil: Checklist


Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+static pthread_mutex_t binaryImagesMutex = PTHREAD_MUTEX_INITIALIZER;
+
+static void
+increaseBufferSize(void)
+{
+    pthread_mutex_lock(&binaryImagesMutex);
+    uint oldLength = binaryImagesBufferLength;
+    uint newLength = oldLength + CACHE_SIZE_INCREMENT;
+
+    SentryCrashBinaryImage *newBuffer = malloc(sizeof(SentryCrashBinaryImage) * newLength);
```
armcknight, line 23: I think we need `size_t` here instead of `uint`

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+
+    if (index >= 0) {
+        int amountToMove = binaryImagesAmount - index + 1;
+        int sizeToMove = amountToMove * sizeof(SentryCrashBinaryImage);
+        void *startPosition = binaryImagesBuffer + ((index + 1) * sizeof(SentryCrashBinaryImage));
+        void *moveTo = startPosition - sizeof(SentryCrashBinaryImage);
+
+        memcmp(moveTo, startPosition, sizeToMove);
+        binaryImagesAmount--;
+    }
```
armcknight, line 81: Could we also just early return for all conditions and do the work below that? Also you'll need to release the lock before early returning.
```suggestion
    if (index < 0 || index >= binaryImagesAmount) {
    	pthread_mutex_unlock(&binaryImagesMutex);
        return;
    }

     int amountToMove = binaryImagesAmount - index + 1;
     int sizeToMove = amountToMove * sizeof(SentryCrashBinaryImage);
     void *startPosition = binaryImagesBuffer + ((index + 1) * sizeof(SentryCrashBinaryImage));
     void *moveTo = startPosition - sizeof(SentryCrashBinaryImage);

     memcmp(moveTo, startPosition, sizeToMove);
     binaryImagesAmount--;
```

armcknight: Side note, in the past I've used a C macro that works like Swift/Go `defer` that is helpful for things like releasing locks for all exit paths... see https://web.archive.org/web/20200307094401/https://oded.blog/2017/10/05/go-defer-in-cpp/. I've come close to adding it a few times for us, but it's C++ so brings in some complications.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+int
+sentrycrashbic_imageCount(void)
+{
+    return binaryImagesAmount;
+}
+
+SentryCrashBinaryImage *
+sentrycrashbic_getBinaryImageCache(int index)
+{
+    return &binaryImagesBuffer[index];
```
armcknight, line 95: Do we need to bounds check `index` or take a lock to read this array?

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+        return;
+    }
+
+    if (index >= 0) {
+        int amountToMove = binaryImagesAmount - index + 1;
+        int sizeToMove = amountToMove * sizeof(SentryCrashBinaryImage);
+        void *startPosition = binaryImagesBuffer + ((index + 1) * sizeof(SentryCrashBinaryImage));
+        void *moveTo = startPosition - sizeof(SentryCrashBinaryImage);
+
+        memcmp(moveTo, startPosition, sizeToMove);
```
armcknight, line 79: Is this supposed to be `memcmp` (in which case we'd want to do something based on the comparison result) or `memcpy`?

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+#include <mach-o/dyld.h>
+#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#define CACHE_SIZE_INCREMENT 100
+static SentryCrashBinaryImage *binaryImagesBuffer = NULL;
+static uint binaryImagesBufferLength;
```
armcknight, line 12: Probably also needs to be `size_t`

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+    if (!sentrycrashdl_getBinaryImageForHeader(
+            (const void *)header, info.dli_fname, &binaryImage)) {
+        return;
+    }
+
+    if (binaryImagesAmount >= binaryImagesBufferLength) {
+        increaseBufferSize();
+    }
+
+    binaryImagesBuffer[binaryImagesAmount++] = binaryImage;
```
armcknight, line 54: Do we need to take a lock before reading `binaryImagesAmount` and modifying `binaryImagesBuffer `?
```suggestion
    pthread_mutex_lock(&binaryImagesMutex);
    if (binaryImagesAmount >= binaryImagesBufferLength) {
         increaseBufferSize();
     }
    binaryImagesBuffer[binaryImagesAmount++] = binaryImage;
    pthread_mutex_unlock(&binaryImagesMutex);
```

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+        binaryImagesAmount--;
+    }
+
+    pthread_mutex_unlock(&binaryImagesMutex);
+}
+
+int
+sentrycrashbic_imageCount(void)
+{
+    return binaryImagesAmount;
```
armcknight, line 89: Lock here?
```suggestion
    pthread_mutex_lock(&binaryImagesMutex);
    int amount = binaryImagesAmount;
    pthread_mutex_unlock(&binaryImagesMutex);
    return amount;
```

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+}
+
+void
+sentrycrashbic_stopCache(void)
+{
+    free(binaryImagesBuffer);
+    binaryImagesBuffer = NULL;
+    binaryImagesAmount = 0;
+    binaryImagesBufferLength = 0;
+}
```
armcknight, line 115: If somehow a customer called SentrySDK.start or SentrySDK.close multiple times, is there anything in the code paths that would prevent these from also being called multiple times, or do we need to lock/guard for that in these to prevent e.g. double-frees?

armcknight (commented):
Stoked that you're taking this on, nice work!

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin: `pthread_mutex_lock` is not async-signal safe, and since this is meant to be used during crashes we cannot lock it, but  for the same reason I dont think this will be a problem because there is no library load after a crash signal. But I will add a comment about this. 

Bounds check make sense.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin: Not because of [this](https://github.com/getsentry/sentry-cocoa/pull/2939/files#r1173472378)

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin:
```suggestion
        memcpy(moveTo, startPosition, sizeToMove);
```

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin: Is it possible to achieve this in C? The article shows a C++ suggestion.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin: No we dont, good point.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+    	pthread_mutex_unlock(&binaryImagesMutex);
+        return;
+    }
+
+     int amountToMove = binaryImagesAmount - index + 1;
+     int sizeToMove = amountToMove * sizeof(SentryCrashBinaryImage);
+     void *startPosition = binaryImagesBuffer + ((index + 1) * sizeof(SentryCrashBinaryImage));
+     void *moveTo = startPosition - sizeof(SentryCrashBinaryImage);
+
+     memcmp(moveTo, startPosition, sizeToMove);
```
brustolin, line 80:
```suggestion
     memcpy(moveTo, startPosition, sizeToMove);
```

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin: This is the number of elements that can fit in the cache, not the memory size. 
Would you still use size_t?

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,123 @@
...
+}
+
+void
+sentrycrashbic_startCache(void)
+{
+    pthread_mutex_lock(&binaryImagesMutex);
+    binaryImagesBufferLength = CACHE_SIZE_INCREMENT;
+    binaryImagesBuffer = malloc(sizeof(SentryCrashBinaryImage) * binaryImagesBufferLength);
+    binaryImagesAmount = 0;
+    _dyld_register_func_for_add_image(&binaryImageAdded);
```
philipphofmann, line 106: `h`: 

> When this function is first registered it is called for once for each image that is currently part of the process.

After calling `sentrycrashbic_stopCache` and clearing the `binaryImagesBuffer` `_dyld_register_func_for_add_image` will not call our registered function once for each image, so we won't get a full list of images. A couple of ideas on how to solve this problem: 

1. Never stopping updating our cache
2. Doing some check with `_dyld_image_count` and retrieve a full list if `_dyld_image_count` returns a different number.

Sources/SentryCrash/Recording/SentryCrashReport.c
```diff
@@ -1207,12 +1205,12 @@ writeBinaryImage(
...
 writeBinaryImages(const SentryCrashReportWriter *const writer, const char *const key)
 {
-    const int imageCount = sentrycrashdl_imageCount();
-
+    const int imageCount = sentrycrashbic_imageCount();
     writer->beginArray(writer, key);
     {
         for (int iImg = 0; iImg < imageCount; iImg++) {
-            writeBinaryImage(writer, NULL, iImg);
+            SentryCrashBinaryImage *image = sentrycrashbic_getBinaryImageCache(iImg);
```
philipphofmann, line 1212: `l`: We are getting a cached binary image, not the cache itself. 
```suggestion
            SentryCrashBinaryImage *image = sentrycrashbic_getCachedBinaryImage(iImg);
```

philipphofmann, line 1212: `h`: With the current approach, we either need a copy of all binary images or a lock on them cause the imageCount could change while we're iterating over the list, so we could get index out of bounds or end up writing an image twice.

philipphofmann (commented):
We should remove `sentrycrashdl_imageCount` and `sentrycrashdl_getBinaryImage` to ensure nobody is calling the async unsafe code anymore.

Thanks for tackling this, @brustolin. This is a tricky one.

Sources/SentryCrash/Recording/SentryCrashReport.c
brustolin: Its very unlikely to have a new lib being loaded after a crash and locks are not async-signal-safe.
We could add a nil check here because the lib has bounds protection.
I don't think Its possible to add the same library twice because they are added to the end of the array.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+#include <mach-o/dyld.h>
+#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+typedef struct SentryCrashBinaryImageNode {
+    SentryCrashBinaryImage image;
+    bool available;
```
philipphofmann, line 12: `l`: I would rename this to `removed` to stick to one naming pattern. The method flipping this flag is called `binaryImageRemoved`. I know I think I suggested `available`, but when looking at it now, I think `removed` is better.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+static SentryCrashBinaryImageNode *tailNode = NULL;
+static pthread_mutex_t binaryImagesMutex = PTHREAD_MUTEX_INITIALIZER;
+
+static void
+binaryImageAdded(const struct mach_header *header, intptr_t slide)
+{
+    pthread_mutex_lock(&binaryImagesMutex);
+    if (tailNode == NULL) {
+        pthread_mutex_unlock(&binaryImagesMutex);
+        return;
```
philipphofmann, line 26: `m`: If I'm not mistaken this will now always exit, as the initial value of `tailNode` is `NULL`.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+} SentryCrashBinaryImageNode;
+
+static SentryCrashBinaryImageNode rootNode = { 0 };
+static SentryCrashBinaryImageNode *tailNode = NULL;
+static pthread_mutex_t binaryImagesMutex = PTHREAD_MUTEX_INITIALIZER;
+
+static void
+binaryImageAdded(const struct mach_header *header, intptr_t slide)
+{
+    pthread_mutex_lock(&binaryImagesMutex);
```
philipphofmann, line 23: `h`: I think we could acquire the lock after all the checks that could return early in this method. If we keep it as it is now, we need to unlock it before every return, if I'm not mistaken.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,115 @@
...
+            (const void *)header, info.dli_fname, &binaryImage)) {
+        return;
+    }
+
+    SentryCrashBinaryImageNode *newNode = malloc(sizeof(SentryCrashBinaryImageNode));
+    newNode->available = true;
+    newNode->image = binaryImage;
+    newNode->next = NULL;
+
+    tailNode->next = newNode;
```
philipphofmann, line 45: `h`: I think tailNode should be `NULL` here. We need to set it to an initial state somewhere.

philipphofmann (commented):
@brustolin, I think the general approach looks fine, so I guess you can start adding tests now.

I think I found a few bugs in the code, so I wait until you add tests to give you a detailed review.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin: If is not a problem for you, I prefer to make the initialization of this struct `{0}` to be in a "non ready" state, so I know I need to activate it in order to make it available.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin: tailNode is set to rootNode at `sentrycrashbic_startCache`.

This is to prevent this function works before initializing the cache

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
armcknight: Probably, it would be interesting to try sometime!

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
armcknight: I agree with @brustolin 's approach. In any case, this is set not only from `binaryImageRemoved` but also `binaryImageAdded`, so the name can't really align with one of the other. I think the name is a correct description of what it's representing.

armcknight (approved):
Looks good, some small suggestions but no blockers as far as I can see. Nice work!

Sources/SentryCrash/Recording/SentryCrashReport.c
```diff
@@ -1204,15 +1207,10 @@ writeBinaryImage(
...
-
     writer->beginArray(writer, key);
-    {
-        for (int iImg = 0; iImg < imageCount; iImg++) {
-            writeBinaryImage(writer, NULL, iImg);
-        }
-    }
+    sentrycrashbic_iterateOverImages(&binaryImagesIteratorCallback, (void *)writer);
     writer->endContainer(writer);
+    printf("###FIM\n");
```
brustolin, line 1213:
```suggestion
```

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,156 @@
+#include "SentryCrashBinaryImageCache.h"
+#include "SentryCrashDynamicLinker.h"
+#include <mach-o/dyld.h>
+#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#if TEST || TESTCI
```
philipphofmann, line 10: `m`: I'm not a big fan of these compiler directives, as we are not testing the else branch. I would prefer to restructure the code so we don't need the compiler directives. Then we can test all the code.

Tests/SentryTests/SentryCrash/SentryBinaryImageCacheTests.m
```diff
@@ -0,0 +1,227 @@
+#import "SentryCrashBinaryImageCache.h"
+#import "SentryCrashWrapper.h"
+#import <XCTest/XCTest.h>
+#include <mach-o/dyld.h>
+
+// Exposing test only functions from `SentryCrashBinaryImageCache.m`
```
philipphofmann, line 6: Thanks for the comment 👍

Tests/SentryTests/SentryCrash/SentryBinaryImageCacheTests.m
```diff
@@ -0,0 +1,227 @@
...
+    _dyld_register_func_for_add_image(&cacheMachHeaders);
+}
+
+- (void)setUp
+{
+    sentry_setRegisterFuncForAddImage(&sentry_register_func_for_add_image);
+    sentry_setRegisterFuncForRemoveImage(&sentry_register_func_for_remove_image);
+
+    mach_headers_initial_array =
+        [mach_headers_test_cache subarrayWithRange:NSMakeRange(0, 5)].mutableCopy;
```
philipphofmann, line 76: `l`: Please add a constant/variable for the magic number 5 or a comment.

Tests/SentryTests/SentryCrash/SentryBinaryImageCacheTests.m
```diff
@@ -0,0 +1,227 @@
...
+    sentrycrashbic_stopCache();
+}
+
+- (void)testStartCache
+{
+    [[SentryCrashWrapper sharedInstance] startBinaryImageCache];
+    [self assertBinaryImageCacheLength:5];
+}
+
+- (void)testStartCacheTwice
```
philipphofmann, line 91: 💯

Tests/SentryTests/SentryCrash/SentryBinaryImageCacheTests.m
```diff
@@ -0,0 +1,227 @@
...
+
+- (void)testStopCache
+{
+    sentrycrashbic_startCache();
+    [self assertBinaryImageCacheLength:5];
+    sentrycrashbic_stopCache();
+    [self assertBinaryImageCacheLength:0];
+}
+
+- (void)testStopCacheTwice
```
philipphofmann, line 108: 💯

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,156 @@
...
+            break;
+        }
+        nextNode = nextNode->next;
+    }
+}
+
+void
+sentrycrashbic_iterateOverImages(sentrycrashbic_imageIteratorCallback callback, void *context)
+{
+    SentryCrashBinaryImageNode *nextNode = &rootNode;
```
philipphofmann, line 107: `m`: Please add a comment on why we don't need the mutex here and also for `binaryImageRemoved`.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,156 @@
...
+        return;
+    }
+
+    SentryCrashBinaryImageNode *newNode = malloc(sizeof(SentryCrashBinaryImageNode));
+    newNode->available = true;
+    newNode->image = binaryImage;
+    newNode->next = NULL;
+
+    pthread_mutex_lock(&binaryImagesMutex);
+    tailNode->next = newNode;
```
philipphofmann, line 85: `h`: I think we must check if the `tailNode` became `NULL` again here, as another thread could call `sentrycrashbic_stopCache` since the check above.
```suggestion
    if (tailNode == NULL) {
        return;
    }
    tailNode->next = newNode;
```

Tests/SentryTests/SentryCrash/SentryBinaryImageCacheTests.m
```diff
@@ -0,0 +1,227 @@
...
+{
+    NSMutableArray *array = (__bridge NSMutableArray *)context;
+    [array addObject:[NSValue valueWithPointer:image]];
+}
+
+@interface SentryBinaryImageCacheTests : XCTestCase
+
+@end
+
+@implementation SentryBinaryImageCacheTests
```
philipphofmann, line 61: `m`: As the `SentryBinaryImageCache` uses a mutex, it would be great to add some test cases checking if our locking strategy works as expected.

philipphofmann (commented):
@brustolin, great job 🥇 ; I think we are close to merging this. Apart from my comments, here are a couple of things to address, please

- [x] Add a changelog entry
- [ ] Remove `sentrycrashdl_imageCount`, `sentrycrashdl_getBinaryImage`, `SentryCrashDefaultBinaryImageProvider`
- [ ] I'm a bit worried, that this could blow up for apps having loads of binary images and maybe use a weird setup. How can we minimize the risk of it blowing up?

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin: I understand your concern.

You are worried about someone accidentally changing the 'else' part. However, the code is so specific that I can't see that happening without breaking the build, unless someone does something malicious, which should be caught during the PR review.

IMO this is the most elegant way to add a "Wrapper" for C code without the need to compile extra code that wont be required for release.

And we do have a test that uses the native function just to make sure it dont break. https://github.com/getsentry/sentry-cocoa/blob/9b2cb171e12b0cc4efe48a38b602e2dd6bf3e5bf/Tests/SentryTests/Integrations/SentryCrash/SentryCrashIntegrationTests.swift#L179-L208

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
brustolin: Sound right

brustolin:
> Remove sentrycrashdl_imageCount, sentrycrashdl_getBinaryImage, SentryCrashDefaultBinaryImageProvider

I rather do this in another PR. There still code for non crash captures that uses `SentryCrashDefaultBinaryImageProvider`.

brustolin:
>I'm a bit worried, that this could blow up for apps having loads of binary images and maybe use a weird setup. How can we minimize the risk of it blowing up?

I believe the amount of library loads that would cause a problem is too high to even be feasible in an App.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
philipphofmann: That works for me 👍

philipphofmann:
> > I'm a bit worried, that this could blow up for apps having loads of binary images and maybe use a weird setup. How can we minimize the risk of it blowing up?
> 
> I believe the amount of library loads that would cause a problem is too high to even be feasible in an App.

I think releasing a beta, would be a good strategy. Furthermore, if it blows up, you should notice it immediately after starting the app.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -0,0 +1,169 @@
...
+        }
+        nextNode = nextNode->next;
+    }
+}
+
+void
+sentrycrashbic_iterateOverImages(sentrycrashbic_imageIteratorCallback callback, void *context)
+{
+    /**
+     We can't use locks here because this is meant to be used during crashes,
```
philipphofmann, line 114: Golden 👍

Tests/SentryTests/SentryCrash/SentryBinaryImageCacheTests.m
```diff
@@ -0,0 +1,247 @@
...
+    [mach_headers_expect_array removeObjectAtIndex:0];
+    [self assertCachedBinaryImages];
+
+    addBinaryImage(removeItem.pointerValue, 0);
+    [self assertBinaryImageCacheLength:5];
+    [mach_headers_expect_array insertObject:removeItem atIndex:4];
+    [self assertCachedBinaryImages];
+}
+
+- (void)testAddBinaryImageInParallel
```
philipphofmann, line 205: `m`: I'm not too worried about calling `addBinaryImage` from multiple threads, but rather stopping from another thread while still adding. Ideally, we would test adding, iterating, and stopping from different threads.

philipphofmann (commented):
Almost finished. We just need to adapt the tests a bit, and then LGTM.

PR #2940, opened 2023-04-21
meta: split decision log into separate file

armcknight:
I was scanning through develop-docs/README.md for something and realized most of it is the decision log. Since that can accumulate lots of info fast, I figure it deserves its own doc.

Honestly I kind of like the idea of DECISIONS.md file along with the usual README.md, CONTRIBUTING.md, LICENSE.md etc for any project. I've often created them in notion and other KBs, just never as one of those top-level files. 

#skip-changelog


philipphofmann (approved):
Thanks, @armcknight.

PR #2941, opened 2023-04-21
ref: only declare conformance on SentryBaseIntegration superclass

armcknight:
I am working on converting SentryCrashWrapper from a singleton to fully dependency-injected (#2942). This involves changing SentryBaseIntegration.init to accept it as an argument, since it needs to use it in its implementation.

I noticed that all of our integration classes both subclass SentryBaseIntegration and conform to SentryIntegrationProtocol at their declarations. I lifted the protocol conformance declaration out of the subclasses and to the superclass, where it applies to any future subclass for free.

I also converted all the generic `id<SentryIntegrationProtocol>` to polymorphic declarations of `SentryBaseIntegration *` instead, like in array types. It would only really make sense to use the type-erased version if customers are able to create their own integrations conforming to `SentryIntegrationProtocol`, but as far as I can tell we don't support that kind of plug-in architecture. **Let me know if I've missed anything here and am wrong about this.**

Finally, I changed the implicit header import of SentryOptions from SentryIntegrationProtocol.h to a forward class declaration, and added the necessary missing imports of SentryOptions in the necessary places.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentrySDK.m
+++ b/Sources/Sentry/SentrySDK.m
@@ -381,7 +381,7 @@ static NSUInteger startInvocations;
     SentryHub *hub = SentrySDK.currentHub;
 
     // Uninstall all the integrations
-    for (NSObject<SentryIntegrationProtocol> *integration in hub.installedIntegrations) {
+    for (SentryBaseIntegration *integration in hub.installedIntegrations) {
         if ([integration respondsToSelector:@selector(uninstall)]) {
             [integration uninstall];
         }
--- a/Sources/Sentry/SentrySDK.m
+++ b/Sources/Sentry/SentrySDK.m
@@ -351,7 +351,8 @@ static NSUInteger startInvocations;
                 integrationName);
             continue;
         }
-        SentryBaseIntegration *integrationInstance = [[integrationClass alloc] init];
+        SentryBaseIntegration *integrationInstance = [integrationClass alloc];
+        integrationInstance = [integrationInstance init];
         BOOL shouldInstall = [integrationInstance installWithOptions:options];
 
         if (shouldInstall) {
--- a/Sources/Sentry/SentrySDK.m
+++ b/Sources/Sentry/SentrySDK.m
@@ -2,6 +2,7 @@
 #import "PrivateSentrySDKOnly.h"
 #import "SentryAppStartMeasurement.h"
 #import "SentryAppStateManager.h"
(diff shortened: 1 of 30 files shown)


Sources/Sentry/SentrySDK.m
```diff
@@ -350,7 +351,7 @@ + (void)installIntegrations
                 integrationName);
             continue;
         }
-        id<SentryIntegrationProtocol> integrationInstance = [[integrationClass alloc] init];
```
armcknight, line 353: This was an especially important change. Because the init is defined in SentryBaseIntegration and not SentryIntegrationProtocol, when I changed it from SentryBaseIntegration.init to SentryBaseIntegration.initWithCrashWrapper:, this line didn't break compilation. It would have gone on to crash at runtime. This is an unsafe code sector in general, so this makes it a little better.

brustolin (commented):
I think is nice to make `SentryBaseIntegration` implements `SentryIntegrationProtocol` and remove the protocol from our integrations, but I think we should continue to use id<SentryIntegrationProtocol> in the arrays and when interacting with integrations.
And Im pretty sure this is a breaking change, we dont see many customers using their own integrations, but its easy to do it.

armcknight:
After talking with @brustolin , it sounds like it is indeed a breaking change if we don't allow customers to supply the names of custom integrations they might create. 

My question is, how do custom integrations work? Is it actually a supported feature? I couldn't find documentation about it.

If it is not supposed to be supported, we should consider removing the ability. While we're at it, we should convert the SentryOptions.integrations string array into something that can supply a nonarbitrary set of integrations; something like an enum or option set.

brustolin:
> something like an enum or option set

I believe SentryOptions flags should be enough to enable/disable integrations. If we have an integration not controlled by a flag we should add it.

philipphofmann:
Custom integrations are an undocumented public feature. The mobile SDKs inherited enabling and disabling features via the list of integrations from languages like JavaScript and Python. On Cocoa, the code for removing one single integration is tedious, so we have simple feature flags for almost all features. I have no clue if somebody is writing custom integration, but our users could, and therefore changing them would be a breaking change. I don't have anything making them internal in the next major.

armcknight:
It sounds like this PR will not be merged until we are preparing the next major version. I'll close it and link to it from https://github.com/getsentry/sentry-cocoa/discussions/2946 so we can revive it later.

PR #2942, opened 2023-04-21
ref: convert SentryCrashWrapper usage from singleton to dependency injected

armcknight:
Per https://github.com/getsentry/sentry-cocoa/pull/2899#discussion_r1168773748

This will help with mocking things in tests. Also, we had many objects that still declared internal properties to retain references to the singleton, so we were halfway to DI anyways. This just cleans things up and makes it consistently so.

Also renamed a few things called `crashAdapter` to `crashWrapper`.

#skip-changelog


Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -76,18 +75,6 @@ - (SentryAppStateManager *)appStateManager
...
-{
-    if (_crashWrapper == nil) {
-        @synchronized(sentryDependencyContainerLock) {
-            if (_crashWrapper == nil) {
-                _crashWrapper = [SentryCrashWrapper sharedInstance];
-            }
-        }
-    }
-    return _crashWrapper;
-}
```
brustolin, line 89: h: I believe is better to keep access to crashWrapper here instead of using
```
[[[SentrySDK currentHub] getClient] crashWrapper]
```

Sources/Sentry/SentryBaseIntegration.m
```diff
@@ -1,14 +1,32 @@
...
 
+- (instancetype)initWithCrashWrapper:(SentryCrashWrapper *)crashWrapper
+{
+    if (self = [super init]) {
+        return nil;
+    }
+
+    _crashWrapper = crashWrapper;
+    return self;
+}
```
brustolin, line 28: h: I don't like this, most integration don't need crashWrapper (If Im not mistaken only one need it), and now all of them will have this constructor, and a extra crashWrapper property for no reason.

armcknight: There was actually only one place it was even used in this class, and i didn't even use this property for it, so i didn't even need to make the change for the constructor 🤦🏻

Sources/Sentry/SentryDependencyContainer.m
brustolin: So we have only one place with all our dependencies.

armcknight: The point of this PR is to stop using the singleton because it's harder to know what pieces are being used in tests. Forcing all components to declare their dependencies in their constructors makes it easier to mock. This way the compiler can enforce this; with the singleton, you can easily forget to set a mock object unless you already know all the internals of the component you're about to test. This presents a high barrier to entry for any newcomer.

This has been an ongoing problem with dispatch queue wrappers and file managers, to name two that i know of.

Sources/Sentry/SentryBaseIntegration.m
armcknight: Honestly, the entire function SentryBaseIntegration.shouldBeEnabledWithOptions should just be moved to SentryClient instead anyways. Then we can remove the param from the init and this is a nonissue.

Tests/SentryTests/Integrations/Session/SentryAutoSessionTrackingIntegrationTests.swift
```diff
@@ -4,7 +4,7 @@ import XCTest
 class SentryAutoSessionTrackingIntegrationTests: XCTestCase {
 
     func test_AutoSessionTrackingEnabled_TrackerInitialized() {
-        let sut = SentryAutoSessionTrackingIntegration()
+        let sut = SentryAutoSessionTrackingIntegration(crashWrapper: TestCrashWrapper())
```
philipphofmann, line 7: Having to provide the crash wrapper for so many tests not using it is another sign to me we shouldn't add it to the default constructor as discussed in https://github.com/getsentry/sentry-cocoa/pull/2942#discussion_r1173430878

philipphofmann (commented):
Thanks for tackling such issues, @armcknight.

armcknight:
After talking some time ago about our acceptance of singletons, I don't think it's worth pushing this PR forward.

PR #2952, opened 2023-04-24
Scope propagation improvements

kabiroberai:
## :scroll: Description

This PR updates the `SentryScope` copy constructor to copy over the parent's span. Additionally, it adds a `+[SentrySDK start]` overload that accepts an initial scope.

Previously, copied scopes (such as those created in `-[SentrySDK captureEvent:withScopeBlock:]`) did not inherit the original scope's `span`. This resulted in emitted events not linking back to the relevant span. Other SDKs do copy over the parent span/transaction (for example [here's](https://github.com/getsentry/sentry-java/blob/0be0d10a9fc61369a551c3b90134fe36c050a584/sentry/src/main/java/io/sentry/Scope.java#L85) the equivalent code in sentry-java).

The other change, enabling `SentrySDK` to be started with an initial scope, has two use cases for us:
1. It helps avoid race conditions when the scope is configured shortly after the SDK has started
2. It enables passing in a `SentryScope` subclass — we use this to override certain properties to be computed instead of stored.

I tested these changes in a production iOS app, ensuring that spans now propagate as expected.

## :pencil: Checklist


Tests/SentryTests/SentryScopeSwiftTests.swift
```diff
@@ -108,6 +108,36 @@ class SentryScopeSwiftTests: XCTestCase {
         XCTAssertNil(actual["transaction"])
         XCTAssertNotNil(actual["breadcrumbs"])
     }
+
+    func testInitWithScope() {
```
kabiroberai, line 112: this test is mostly a direct translation of the ObjC version, but with the addition of a check that calling `cloned.applyTo` has the same effect as calling it on the original scope.

philipphofmann:
Thanks for the PR, @kabiroberai. We didn't get to this today, but we'll have a look tomorrow.

philipphofmann:
@kabiroberai, could you elaborate on the race condition you experience between `start` and `configureScope`? Does you app do heavy work in the background already when you start the SDK?

kabiroberai:
👍 moved the `startWithOptions:andScope:` changes out of this PR. The race condition is speculative/us being extra cautious so it's possible that it doesn't actually get triggered, though I can stress-test to double check. Our main use case for passing our own scope is that we can replace the hub's scope with a subclass — to that end a `+[SentrySDK setScope:]` method would achieve the same goal. Unfortunately `configureScope` doesn't do the trick because we can't replace the `scope` instance entirely, only modify its properties.

Taking a step back, the reason we need to subclass `Scope` is because we're using Swift's `@TaskLocal` to allow different (concurrent) Tasks to have different "current" scopes. IIUC other Sentry SDKs achieve a similar goal through the use of thread-local hubs, but I don't see a way to do such a thing in the Cocoa SDK yet. I'd be happy to discuss alternative approaches here — maybe on Discord?

philipphofmann:
@kabiroberai, JavaScript the concept of the [initialScope on the options](https://docs.sentry.io/platforms/node/configuration/options/#initial-scope). We could do the same on Cocoa. I guess that would work for you.

> through the use of thread-local hubs

At some point, we should implement thread-local hubs, but if initialScope solves your problem, I would like to delay that a bit, as initialScope is also useful for other users.

Tests/SentryTests/SentryScopeSwiftTests.swift
philipphofmann: Please add some validation for assigning `self.span = scope.span` in the `initWithScope`. So basically if I remove `self.span = scope.span` from `initWithScope` this test should fail.

philipphofmann (commented):
Please add an entry to the changelog under a section Fixed with `Propagate span when copying scope`. CI fails for the benchmarks and release builds, cause that doesn't work for contributors yet. Just focus on the lints and unit tests. FYI, our unit test are a bit flaky at the moment. We did several fixes on the main branch, so merging main into this PR should help.

Tests/SentryTests/SentryScopeSwiftTests.swift
kabiroberai: Oops, this was what the `Event` comparison was supposed to validate, but I forgot to assign a span to the scope. Removing that line goes from green -> red as expected now.

brustolin (commented):
Thanks @kabiroberai !!

This looks good to me.

philipphofmann (approved):
Thanks @kabiroberai 💯; LGTM

philipphofmann:
I will fix the Changelog entry so that we can merge this PR.

PR #2956, opened 2023-04-25
perf: gather profiler metrics gathering on a low-pri queue

armcknight:
## :scroll: Description

Change the metric profiler from using an NSTimer, which dispatches blocks to the main thread, to a dispatch source that dispatches blocks to a queue with a utility quality of service.

We noticed some customer profiles that showed our own symbols doing lots of work on main threads. This is undesirable from a customer POV and isn't even really necessary for these readings.

Created mocking for the dispatch source and verified the preexisting tests still work.

This also introduces the aforementioned dispatch factory that vends our wrappers for dispatch queues and the new source wrapper, which we should eventually use in all the places in the SDK that internally alloc/init queue wrappers, for easier and more consistent mockability in tests.

## :pencil: Checklist


Sources/Sentry/SentryDispatchQueueWrapper.m
```diff
@@ -3,13 +3,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@implementation SentryDispatchQueueWrapper {
-    // Don't use a normal property because on RN a user got a warning "Property with 'retain (or
-    // strong)' attribute must be of object type". A dispatch queue is since iOS 6.0 an NSObject so
-    // it should work with strong, but nevertheless, we use an instance variable to fix this
-    // warning.
-    dispatch_queue_t queue;
-}
```
armcknight, line 12: This is no longer true since we bumped our min supported version. I believe we can also change this: https://github.com/getsentry/sentry-cocoa/blob/d257eb9dc69e0ffa45c1e4e526fa0f5d7199b83c/Sources/Sentry/SentryDispatchQueueWrapper.m#L16-L17

This is now exposed as a property in the public interface so it can be accessed when setting up a new `dispatch_source_t` in `SentryDispatchSourceWrapper`[

Sources/Sentry/SentryMetricProfiler.mm
```diff
@@ -22,19 +25,19 @@ @interface SentryMetricReading : NSObject
...
 
 NSString *const kSentryMetricProfilerSerializationUnitBytes = @"byte";
 NSString *const kSentryMetricProfilerSerializationUnitPercentage = @"percent";
 
+// Currently set to 10 Hz as we don't anticipate much utility out of a higher resolution when
+// sampling CPU usage and memory footprint, and we want to minimize the overhead of making the
+// necessary system calls to gather that information. This is currently roughly 10% of the
+// backtrace profiler's resolution.
+static uint64_t frequencyHz = 10;
+static uint64_t leewayNs = 50000; // 50 milliseconds
```
indragiek, line 39: nit: could derive this from `frequencyHz` at runtime instead of hardcoding it (i.e. interval / 2)

Sources/Sentry/SentryMetricProfiler.mm
```diff
@@ -156,12 +159,16 @@ - (void)stop
...
-                                                     }];
+    _timer =
+        [_dispatchFactory sourceWithInterval:intervalNs
+                                      leeway:(uint64_t)1e9 / frequencyHz
+                                   queueName:"io.sentry.metric-profiler"
+                                  attributes:dispatch_queue_attr_make_with_qos_class(
+                                                 DISPATCH_QUEUE_CONCURRENT, QOS_CLASS_UTILITY, 0)
+                                eventHandler:^{
+                                    [weakSelf recordCPUPercentagePerCore];
+                                    [weakSelf recordMemoryFootprint];
```
indragiek, line 170: `recordCPUPercentagePerCore` and `recordMemoryFootprint` were previously called on the main thread, are they thread safe?

philipphofmann:
Sorry, @armcknight, I didn't get to this today. I will try to have a look tomorrow. In the meantime maybe you have time to fix CI 😃?

Sources/Sentry/SentryMetricProfiler.mm
armcknight: ah yeah that's what it is in the benchmarker, and i changed it here. i'll put it back to that.

Sources/Sentry/SentryMetricProfiler.mm
armcknight: Good question. Their calls to `task_info` and `host_processor_info` are not synchronized, but their modifications of the data structure those readings are stored in are inside `@synchronized` blocks. Is this safe, or do you think we should move the `task_info`/`host_processor_info` calls inside the synchronization as well?

armcknight: i improved the method that reads from the data structure as well in https://github.com/getsentry/sentry-cocoa/pull/2956/commits/adef4bc78ea2111056102e91a9a8175b0ceece29

Sources/Sentry/SentryDispatchFactory.m
```diff
@@ -0,0 +1,36 @@
...
+- (SentryDispatchSourceWrapper *)sourceWithInterval:(uint64_t)interval
+                                             leeway:(uint64_t)leeway
+                                              queue:(SentryDispatchQueueWrapper *)queueWrapper
+                                       eventHandler:(void (^)(void))eventHandler
+{
+    return [[SentryDispatchSourceWrapper alloc] initTimerWithInterval:interval
+                                                               leeway:leeway
+                                                                queue:queueWrapper
+                                                         eventHandler:eventHandler];
+}
```
brustolin, line 22: m: This function is only used once, and internally in this same class. Can we remove it?

brustolin (approved):
LGTM.

A few suggestions to think about.

Sources/Sentry/SentryDispatchSourceWrapper.m
```diff
@@ -0,0 +1,32 @@
...
+    SentryDispatchQueueWrapper *_queueWrapper;
+    dispatch_source_t _source;
+}
+
+- (instancetype)initTimerWithInterval:(uint64_t)interval
+                               leeway:(uint64_t)leeway
+                                queue:(SentryDispatchQueueWrapper *)queueWrapper
+                         eventHandler:(void (^)(void))eventHandler
+{
+    if (!(self = [super init])) {
```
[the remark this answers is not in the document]
armcknight: I'll make the change but it's very annoying to have to change around code to satisfy a tool like this. IMO it'll actually be worse, because without the early return the failure path is now implicit. The same failure modes are still present, but the fact that the tooling treats it differently because of the way the code is written is not great :/

philipphofmann:
Not reviewing this, as @brustolin already LGTM.

PR #2960, opened 2023-04-25
feat: Swift error names

philipphofmann:
## :scroll: Description

Call into Swift to get the error description for Swift errors to get meaningful error names instead of only the error enum code. To avoid sending PII, the SDK strips parameter values and doesn't send the Swift error name for struct-based Swift errors.
This change has no impact on grouping.

I'm unsure about my data stripping approach in `SwiftDescriptor.getSwiftErrorDescription` to avoid sending PII. Maybe it's also OK to just attach the description. On the other hand, I feel safer changing that behavior in the next major.

Fixes GH-2958

Unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/SwiftDescriptor.swift
+++ b/Sources/Swift/SwiftDescriptor.swift
@@ -8,4 +8,20 @@ public class SwiftDescriptor: NSObject {
         return String(describing: type(of: object))
     }
     
+    @objc
+    public static func getSwiftErrorDescription(_ error: Error) -> String? {
+        let description = String(describing: error)
+        
+        // We can't reliably detect what is PII in a struct and what is not.
+        // Furthermore, we can't detect which property contains the error enum.
+        if description.contains(":") || description.contains(",") {
+            return nil
+        }
+        
+        // For error enums the description could contain PII in between (). Therefore,
+        // we strip the data.
+        let index = description.firstIndex(of: "(") ?? description.endIndex
+        return String(description[..<index])
+    }
+    
 }
--- a/Tests/SentryTests/SwiftDescriptorTests.swift
+++ b/Tests/SentryTests/SwiftDescriptorTests.swift
@@ -28,7 +28,61 @@ class SwiftDescriptorTests: XCTestCase {
         XCTAssertEqual(name, "InnerClass")
     }
     
+    func testgetSwiftErrorDescription_EnumValue() {
+        let actual = SwiftDescriptor.getSwiftErrorDescription(SentryTestError.someError)
+        XCTAssertEqual("someError", actual)
+    }
+    
+    func testgetSwiftErrorDescription_EnumValueWithData() {
+        let actual = SwiftDescriptor.getSwiftErrorDescription(SentryTestError.someOhterError(10))
+        XCTAssertEqual("someOhterError", actual)
+    }
+    
+    func testgetSwiftErrorDescription_StructWithData() {
+        let actual = SwiftDescriptor.getSwiftErrorDescription(XMLParsingError(line: 10, column: 12, kind: .internalError))
+        XCTAssertNil(actual)
+        
+        SentrySDK.capture(error: LoginError.wrongPassword)
+    }
+    
+    func testgetSwiftErrorDescription_StructWithOneParam() {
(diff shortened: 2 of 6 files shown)


Tests/SentryTests/SwiftDescriptorTests.swift
```diff
@@ -28,7 +28,61 @@ class SwiftDescriptorTests: XCTestCase {
...
+    func testgetSwiftErrorDescription_EnumValueWithData() {
+        let actual = SwiftDescriptor.getSwiftErrorDescription(SentryTestError.someOhterError(10))
+        XCTAssertEqual("someOhterError", actual)
+    }
+    
+    func testgetSwiftErrorDescription_StructWithData() {
+        let actual = SwiftDescriptor.getSwiftErrorDescription(XMLParsingError(line: 10, column: 12, kind: .internalError))
+        XCTAssertNil(actual)
+        
+        SentrySDK.capture(error: LoginError.wrongPassword)
```
brustolin, line 45: m: What is the purpose of this line in this test?

Sources/Swift/SwiftDescriptor.swift
```diff
@@ -8,4 +8,20 @@ public class SwiftDescriptor: NSObject {
...
     
+    @objc
+    public static func getSwiftErrorDescription(_ error: Error) -> String? {
+        let description = String(describing: error)
+        
+        // We can't reliably detect what is PII in a struct and what is not.
+        // Furthermore, we can't detect which property contains the error enum.
+        if description.contains(":") || description.contains(",") {
+            return nil
+        }
```
brustolin, line 19: h: This would also return nil for Enum errors with arguments making the stripping below unnecessary.

brustolin (commented):
It looks good, I have one question and one thing you should review.

Whats is preventing this from being ready for review?

Sources/Swift/SwiftDescriptor.swift
brustolin: I probably need a test with an enum like this

```Swift
enum LoginError: Error {
    case wrongUser(id : String)
    case wrongPassword
}
```

philipphofmann: Thanks will add one next week.

philipphofmann:
> Whats is preventing this from being ready for review?

What I mentioned above in the description: 

> I'm unsure about my data stripping approach in SwiftDescriptor.getSwiftErrorDescription to avoid sending PII. Maybe it's also OK to just attach the description. On the other hand, I feel safer changing that behavior in the next major.

Sources/Swift/SwiftDescriptor.swift
philipphofmann: I ditched the whole logic so this is comment is unrelevant.

Tests/SentryTests/SwiftDescriptorTests.swift
philipphofmann: It was there by accident. Good catch thanks.

Sources/Swift/SwiftDescriptor.swift
```diff
@@ -8,4 +8,9 @@ public class SwiftDescriptor: NSObject {
         return String(describing: type(of: object))
     }
     
+    @objc
+    public static func getSwiftErrorDescription(_ error: Error) -> String? {
+        return String(describing: error)
+    }
```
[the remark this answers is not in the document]
philipphofmann: I guess you missed my update in the description above 

Update: No need to worry about PII in a major as we already would send PII of Errors with https://github.com/getsentry/sentry-cocoa/blob/9d61bea77651d44e030326c34b2224d4076a9184/Sources/Sentry/SentryClient.m#L265-L267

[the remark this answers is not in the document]
brustolin: > I guess you missed my update in the description above

Yeap I did. I would like to suggest to only update description if something is wrong, change of direction in the PR should be describe in a new comment. Or both, since GH itself don't warn about description changing.

Thanks for clarifying!

brustolin (approved):

PR #2965, opened 2023-04-27
ref: Remove unused FileManager.storeDictionary

philipphofmann:
Remove the unused method storeDictionary:toPath.

#skip-changelog


brustolin (approved):

philipphofmann:
We only removed some code. The failing CI job has nothing to do with that.

PR #2966, opened 2023-04-27
ref: Move storeData into storeEnvelope

philipphofmann:
The storeData method was only used by storeEnvelope. Both methods used @synchronized. Now the storeData is moved into storeEnvelope and only one invocation of @ synchronized is needed.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/2965.

#skip-changelog


brustolin (approved):

philipphofmann:
Test failures have nothing to do with this change. They are due to tests timing out.

PR #2967, opened 2023-04-27
test: Speed up testDefaultMaxEnvelopesConcurrent

philipphofmann:
Set maxCacheItems to 1, so the testDefaultMaxEnvelopesConcurrent test doesn't need to store 30 envelope items but only 1 for the max envelopes items logic to run.

Test timed out here https://github.com/getsentry/sentry-cocoa/actions/runs/4807094860/jobs/8555451492.

#skip-changelog


brustolin (approved):

philipphofmann:
Test timed out. It has nothing to do with this PR.

PR #2982, opened 2023-05-02
feat: Allow starting the SDK with an initial scope

kabiroberai:
## :scroll: Description

This PR adds a `SentryOptions.initialScope` property, enabling simple configuration of the initial scope used when starting the SDK.

The reasons for this change are twofold:

1. Helps avoid the possibility of race conditions between starting the SDK and calling `configureScope`
2. Allows clients to pass a `SentryScope` subclass to override certain behaviors. For more context, see the discussion in #2952.

There is precedent for this option; see https://github.com/getsentry/sentry-cocoa/pull/2952#issuecomment-1527242186 which mentions a similar `initialScope` option in the JS SDK.

Unit tests + production app.

## :pencil: Checklist


brustolin (commented):
Thanks @kabiroberai for this PR.

Looks good to me.

Sources/Sentry/Public/SentrySDK.h
```diff
@@ -24,6 +24,12 @@ SENTRY_NO_INIT
  */
 @property (class, nonatomic, readonly) BOOL isEnabled;
 
+/**
+ * Inits and configures Sentry (SentryHub, SentryClient) and sets up all integrations. Make sure to
+ * set a valid DSN.
+ */
++ (void)startWithOptions:(SentryOptions *)options andScope:(nullable SentryScope *)scope NS_SWIFT_NAME(start(options:scope:));
```
philipphofmann, line 31: `h`: I prefer to align with JavaScript instead of adding an extra method. We should add a method `initialScope` to the options, basically what we have for `SentrySDK.configureScope`. The public-facing API should look like this
```Swift
SentrySDK.start { options in
    options.dsn = "dsn"
    options.initialScope { scope in
        scope.setEnvironment("debug")
        // ...
    }
}
```

I also think that the API above is cleaner than having to init the options and the scope yourself:
```
let options = Options()
options.dns = "dsn"

let initialScope = Scope()
scope.setEnvironment("debug")

SentrySDK.start(options: options, scope: initialScope)
```

philipphofmann (commented):
Thanks for doing this, @kabiroberai. We are close, but as pointed out in the comment below, this PR doesn't align with the approach we already have on JavaScript. I'm sorry if my communication wasn't clear enough, so you now have to adapt the PR.

Sources/Sentry/Public/SentrySDK.h
kabiroberai: Gotcha, done! I also made it possible to pass a whole `Scope` object by setting `initialScopeFactory` directly instead of calling `initialScope:`, which should solve for use case 2 mentioned in the PR description.

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -191,6 +191,19 @@ NS_SWIFT_NAME(Options)
...
+ * @note Return @c nil to allow the SDK to lazily create its own scope.
+ * @note The default simply returns @c nil .
+ */
+@property (nonatomic) SentryScope * _Nullable (^initialScopeFactory)(void);
+
+/**
+ * Provides an initial scope to use when starting the SDK.
+ * @param callback A block that configures a fresh scope as desired.
+ */
+- (void)initialScope:(void (^)(SentryScope *scope))callback;
```
kabiroberai, line 205: I wonder whether we should rename this to `-configureInitialScope:` and rename `initialScopeFactory` to `initialScope` — @philipphofmann what do you think?

philipphofmann (approved):
Thanks a lot, @kabiroberai 👏. LGTM

philipphofmann:
The failing CI jobs have nothing to do with the changes in this PR. We have some fixes on the main branch for https://github.com/getsentry/sentry-cocoa/actions/runs/4879105885/jobs/8705357840?pr=2982, but I don't want it to block this PR.

philipphofmann:
I will update the docs.

PR #2990, opened 2023-05-03
ci: Build test server on Ubuntu

philipphofmann:
As Ubuntu runners are cheaper and usually faster, build the test server on Ubuntu.

#skip-changelog


philipphofmann:
I had some [troubles](https://github.com/getsentry/sentry-cocoa/actions/runs/4869659733?pr=2990) making the test-server compiled on Ubuntu run on macOS. As the test-server uses caching, I decided it's not worth the effort of fixing the problem. There are more important issues at the moment.

PR #2991, opened 2023-05-03
test: Move codecov.yml to root directory

philipphofmann:
Move the codecov.yml so Codecov uses it. The [current yml](https://app.codecov.io/gh/getsentry/sentry-cocoa/settings/yaml) that Codecov uses is bogus:

```yml
codecov:
  branch: main
  notify:
    wait_for_ci: true
  require_ci_to_pass: true
comment:
  after_n_builds: 5
  behavior: default
  layout: reach,diff,flags,files,footer
  require_changes: false
coverage:
  range:
  - 85.0
  - 100.0
ignore:
- ^Tests/SentryTests.*
parsers:
  gcov:
    branch_detection:
      conditional: true
      loop: true
      macro: false
      method: false
```

Furthermore, remove some default values.

#skip-changelog


philipphofmann:
https://github.com/getsentry/sentry-cocoa/pull/2991#issuecomment-1532652350 is not correct. This should be fixed once we merge the yml to main.

PR #2992, opened 2023-05-03
fix: testNoBreadcrumbForTextFieldEditingChanged flakiness 

brustolin:
Solving testNoBreadcrumbForTextFieldEditingChanged flakiness in the CI simulator.

_#skip-changelog_

### DIFF
--- a/Samples/iOS-Swift/iOS-SwiftUITests/UIEventBreadcrumbTests.swift
+++ b/Samples/iOS-Swift/iOS-SwiftUITests/UIEventBreadcrumbTests.swift
@@ -24,6 +24,8 @@ class UIEventBreadcrumbTests: XCTestCase {
 
         //Trigger a change in textfield
         app.buttons["editingChangedButton"].tap()
+
+        Thread.sleep(forTimeInterval: 0.5)
         //Check the last breadcrumb is the button being pressed
         XCTAssertEqual(app.staticTexts["breadcrumbLabel"].label, "performEditingChangedPressed:")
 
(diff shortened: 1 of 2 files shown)


philipphofmann (approved):
I'm a bit surprised that this solves the problem, but if it does LGTM.

brustolin:
> I'm a bit surprised that this solves the problem, but if it does LGTM.

I should have make this a draft. Im trying options, and my assumption is that the simulator is slow and the test may execute faster than the app.

Samples/iOS-Swift/iOS-SwiftUITests/UIEventBreadcrumbTests.swift
```diff
@@ -22,17 +22,17 @@ class UIEventBreadcrumbTests: XCTestCase {
         app.buttons["Extra"].tap()
         app.buttons["UI event tests"].tap()
 
-        let label = app.staticTexts["breadcrumbLabel"]
-
         //Trigger a change in textfield
         app.buttons["editingChangedButton"].tap()
+
+        Thread.sleep(forTimeInterval: 0.5)
```
philipphofmann, line 28: `m`: Consider using `waitForExistence` with a timeout instead, please.

brustolin: the problem is that there is nothing to "waitForExistence"

brustolin: I'll rewrite the sample to dynamically create labels for every step. Overkill but it will solve

brustolin: So, I was using `accessibilityIdentifier` and totally forgot that it is possible to wait for a label with a specific text.
Now I have a proper solution.

brustolin: So, I was so fixed to use `accessibilityIdentifier` that I forgot that is possible to wait for a label with a specific text.
Now it works, but I had to update the macOS and Xcode version.

philipphofmann (approved):
LGTM

Issue #2993, opened 2023-05-03
Crashes reported on iOS Prod build 


Gauravprakash:

kahest:
Hi @Gauravprakash - thank you for reporting, we are investigating. Do you by any chance have the full stacktraces for those issues and could you share them with us?

krystofwoldrich:
I've moved this to `sentry-cocoa` because the RN code snippet showed `enabledNative: false`. Definitely, a whole stack trace would be helpful.

philipphofmann:
I think most of the issues should be fixed with https://github.com/getsentry/sentry-cocoa/pull/2995.

armcknight:
> I think most of the issues should be fixed with #2995.

I think that's going to be a relatively small proportion of the overall amount of crashes. The screenshot shows a total of 210 crashes, but #2995 only addresses the ones from `SentryProfiler.processBacktrace`, which is only 5% of the total. We would really need a stacktrace to see what's up with `+ [SentryProfiler isRunning]`. I'm not sure what's up with the most prevalent crash in the serializer. Stack traces will definitely be helpful.

brustolin:
I also would like to see the stacktrace for `+[SentrySerialization dataWithJSONObject:]`

@Gauravprakash You can also [send me](mailto:dhiogo.brustolin@sentry.io) your org name so we could better investigate.

Gauravprakash:
please check attached stacktrace, I have extracted it from 2 major crash for now 

[Archive 2.zip](https://github.com/getsentry/sentry-cocoa/files/11406174/Archive.2.zip)

Gauravprakash:
> I think most of the issues should be fixed with #2995.

but this should be fixed from library end

Gauravprakash:
What's update here , please share your explanation for crashes occurring

philipphofmann:
@Gauravprakash, could you please share the stacktrace of the crashes? That would help us to understand what's going on.

armcknight:
Yes, the stack traces for each of these would help. I looked at the archive you uploaded, which had a stack trace for one of them, which is very helpful thanks! But the only other report in there was for a crash not listed in your screenshot, and involved no sentry code. If at all possible, please provide stack traces for the other crashes.

armcknight:
@Gauravprakash we think we've fixed some of the crashes in #3018, based on the information in one of your stack traces and some other customer reports. Stay tuned for a patch release.

armcknight:
[...]

PR #2995, opened 2023-05-03
fix: nil-check values being added to dictionaries

armcknight:
## :scroll: Description

Check any values that might possibly be `nil` before adding them to a dictionary [literal].

Fixes some potential crashes in https://github.com/getsentry/sentry-cocoa/issues/2993.

I don't have a repro, but these are easy checks to add and shouldn't change any behavior/functionality.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -107,8 +107,11 @@ processBacktrace(const Backtrace &backtrace,
     }
     if (queueAddress != nil && queueMetadata[queueAddress] == nil
         && backtrace.queueMetadata.label != nullptr) {
-        queueMetadata[queueAddress] =
-            @ { @"label" : [NSString stringWithUTF8String:backtrace.queueMetadata.label->c_str()] };
+        const auto labelCString = backtrace.queueMetadata.label->c_str();
+        if (labelCString != nullptr) {
+            queueMetadata[queueAddress] =
+                @ { @"label" : [NSString stringWithUTF8String:labelCString] };
+        }
     }
 #    if defined(DEBUG)
     const auto symbols


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This change has no impact on grouping of the issues in Sentry.
 ### Fixes 
 
 - Propagate span when copying scope (#2952)
+- Prevent crashes in profiler when adding potentially `nil` references to dictionaries (#2995)
```
armcknight, line 29:
```suggestion
- Prevent potential crashes in profiler by nil-checking object references added to dictionaries (#2995)
```

brustolin (commented):
Looks good. I think we just need some tests to avoid regression. 
Is it possible to create tests for this scenarios?

armcknight:
@brustolin @philipphofmann I'm not actually sure how any of these could be nil, other than some allocator issue where the NSString allocations return nil. Do you have any ideas?

Sources/Sentry/SentryProfiler.mm
```diff
@@ -100,15 +100,20 @@
         threadMetadata[threadID] = metadata;
     }
     if (!backtrace.threadMetadata.name.empty() && metadata[@"name"] == nil) {
-        metadata[@"name"] = [NSString stringWithUTF8String:backtrace.threadMetadata.name.c_str()];
+        const auto name = [NSString stringWithUTF8String:backtrace.threadMetadata.name.c_str()];
+        if (name != nil) {
+            metadata[@"name"] = name;
```
indragiek, line 105: we don't need this for a subscript assignment. `metadata[key] = nil` will not crash like it will when you put it in a literal, it effectively just removes the element if it exists or does nothing otherwise

Sources/Sentry/SentryProfiler.mm
```diff
@@ -100,15 +100,20 @@
...
     if (backtrace.threadMetadata.priority != -1 && metadata[@"priority"] == nil) {
         metadata[@"priority"] = @(backtrace.threadMetadata.priority);
     }
     if (queueAddress != nil && queueMetadata[queueAddress] == nil
         && backtrace.queueMetadata.label != nullptr) {
-        queueMetadata[queueAddress] =
-            @ { @"label" : [NSString stringWithUTF8String:backtrace.queueMetadata.label->c_str()] };
+        const auto label = [NSString stringWithUTF8String:backtrace.queueMetadata.label->c_str()];
+        if (label != nil) {
+            queueMetadata[queueAddress] = @ { @"label" : label };
```
indragiek, line 115: this is legit 👍🏽

Sources/Sentry/SentryProfiler.mm
```diff
@@ -121,6 +126,10 @@
          backtraceAddressIdx < backtrace.addresses.size(); backtraceAddressIdx++) {
         const auto instructionAddress
             = sentry_formatHexAddressUInt64(backtrace.addresses[backtraceAddressIdx]);
+        if (instructionAddress == nil) {
```
indragiek, line 129: I don't see a way for the return value of `sentry_formatHexAddressUInt64` to be nil

Sources/Sentry/SentryProfiler.mm
```diff
@@ -219,16 +228,32 @@
         }
         const auto relativeTimestamp = getDurationNs(transaction.startSystemTime, timestamp);
 
+        const auto elapsed = sentry_stringForUInt64(relativeTimestamp);
```
indragiek, line 231: same here - this is more likely to crash than it is to return a nil value (e.g. if malloc fails)

Sources/Sentry/SentryProfiler.mm
```diff
@@ -249,10 +274,23 @@
             SENTRY_LOG_WARN(@"Filtered sample not chronological with transaction.");
             return;
         }
+
+        const auto elapsed = sentry_stringForUInt64(
+            getDurationNs(transaction.startSystemTime, sample.absoluteTimestamp));
+        if (elapsed == nil) {
```
indragiek, line 280: same here

Sources/Sentry/SentryProfiler.mm
```diff
@@ -249,10 +274,23 @@
...
+            SENTRY_LOG_WARN(@"Could not convert elapsed time to string.");
+            return;
+        }
+
+        const auto threadID = sentry_stringForUInt64(sample.threadID);
```
indragiek, line 285: same here

indragiek (changes requested):
Adding a lot of superfluous nil checks makes this hard to read, we should do this for functions that we know *can* return nil.

For `sentry_stringForUInt64` and `sentry_formatHexAddressUInt64` we should fix the implementations if it's actually possible for them to return nil vs. checking at every call site.

Sources/Sentry/SentryProfiler.mm
armcknight: ah yes, thank you for pointing this out

Sources/Sentry/SentryProfiler.mm
armcknight: This is the only dictionary literal in processBacktrace which was the blamed frame in the issue screenshot, so is the likely candidate actually needing this. I'll just remove all the others.

Sources/Sentry/SentryProfiler.mm
philipphofmann: Also [`stringWithUTF8String`](https://developer.apple.com/documentation/foundation/nsstring/1497379-stringwithutf8string) doesn't return `nil`. It raises an exception, though, when you pass `NULL` to it. You already check the `name` with `empty()`. So I don't think this will change any behavior.

Sources/Sentry/SentryProfiler.mm
philipphofmann: `h`: @armcknight, I think this won't work. As pointed out above, `stringWithUTF8String` is not supposed to return `nil`. Maybe the `label->c_str()` isn't returning a proper NULL-terminated C array of bytes in UTF8 encoding.

armcknight: I'm afraid if we don't think this check will be useful, then we will have to close this PR and go back to the drawing board. Perhaps there is something else going on and the crash reporter is errantly blaming this frame.

`std::string::c_str()` will never return `NULL` as long as the reference is not `NULL`, which we already guard against in the `if` condition right above these lines: https://github.com/getsentry/sentry-cocoa/blob/034be1c2ce49e44fc857dbda11657371e7cac85e/Sources/Sentry/SentryProfiler.mm#L108-L109

armcknight: I force pushed a new commit so that all the other nil checks in the PR are removed, and changed this one to check the return value of `c_str()` is `!= nullptr` before passing it to `stringWithUTF8String`: https://github.com/getsentry/sentry-cocoa/pull/2995/files#diff-9d7d24f33a2554811f36d8236cdd5b9b2c7c2786b12be31e4a05a3934193441cR110-R114

```
const auto labelCString = backtrace.queueMetadata.label->c_str();
if (labelCString != nullptr) {
    queueMetadata[queueAddress] =
        @ { @"label" : [NSString stringWithUTF8String:labelCString] };
}
```

but i'm not sure this is doing anything actually useful.

philipphofmann: But then I think the change doesn't make sense as you pointed out and we have to go back to the drawing board 🤔

armcknight: So should we close this PR @indragiek and @philipphofmann ?

philipphofmann: I agree with closing it. Although we can't merge it, it was still worth the effort as it helped us to understand the underlying problem.

PR #3005, opened 2023-05-05
fix: Remove "/" from crash report file name

### LINKED ISSUE #2975: No automatic Crash upload
### Platform

iOS

### Installed

CocoaPods

### Version

8.5.0

### Steps to Reproduce

1. First, I registered the SentrySDK in the first line of the `application(_:, didFinishLaunchingWithOptions:)` method:
```swift
public func start(environment: String? = nil, release: String? = nil) {
    guard !SentrySDK.isEnabled else { return }

    SentrySDK.start {
        $0.dsn = Config.sentry.dsn
        $0.debug = !Bool.isRelease
        $0.attachScreenshot = true
        $0.attachViewHierarchy = true
        $0.releaseName = release ?? Bundle.sak.shortVersionString()
        $0.dist = Bundle.sak.bundleVersionString()
        
        if let environment = environment {
            $0.environment = environment
        }
     }
}
```

2. Running App on iPhone XS Max with 14.7.1.
3. Click a pre-set button that executes the `SentrySDK.crash()` method after the button is clicked.

### Expected Result

App crashes and receives a crash log in our self-built Sentry panel.

### Actual Result

The app crashes, but the web panel is empty.

--------

I've tried calling the `capture()` method and manually reporting an `NSError`, and it shows up immediately in the web panel.

I also get `WatchdogTermination` in the web panel, but the only thing that doesn't show up is the crash.

Are there any configurations that would affect this behavior? Any help would be appreciated, thanks.

brustolin:
## :scroll: Description

Crash report uses bundleName to create the report file, but "/" is a valid character for bundle name but not for file name.
Replace "/" with "-" for the report file name.

Close #2975

Unit test and sample.

## :pencil: Checklist

### DIFF
--- a/Sources/SentryCrash/Recording/SentryCrash.m
+++ b/Sources/SentryCrash/Recording/SentryCrash.m
@@ -534,18 +534,20 @@ SYNTHESIZE_CRASH_STATE_PROPERTY(BOOL, crashedLastLaunch)
     sentrycrash_notifyAppTerminate();
 }
 
-- (NSString *) clearBundleName:(NSString *)filename
+- (NSString *)clearBundleName:(NSString *)filename
 {
     return [filename stringByReplacingOccurrencesOfString:@"/" withString:@"-"];
 }
 
-- (NSString *) getBundleName
+- (NSString *)getBundleName
 {
-    NSString *bundleName = [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleName"] ?: @"Unknown";
+    NSString *bundleName =
+        [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleName"] ?: @"Unknown";
     return [self clearBundleName:bundleName];
 }
 
-- (NSString *) getBasePath {
+- (NSString *)getBasePath
+{
     NSArray *directories
         = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
     if ([directories count] == 0) {
@@ -560,7 +562,6 @@ SYNTHESIZE_CRASH_STATE_PROPERTY(BOOL, crashedLastLaunch)
     NSString *pathEnd = [@"SentryCrash" stringByAppendingPathComponent:[self getBundleName]];
     return [cachePath stringByAppendingPathComponent:pathEnd];
 }
-
 
 @end
 
--- a/Sources/SentryCrash/Recording/SentryCrash.m
+++ b/Sources/SentryCrash/Recording/SentryCrash.m
@@ -67,10 +67,18 @@ SentryCrash ()
 static NSString *
 getBundleName(void)
 {
-    NSString *bundleName = [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleName"];
+    static NSString *bundleName = nil;
+
     if (bundleName == nil) {
-        bundleName = @"Unknown";
+#if TEST
+        bundleName = @"Sentry/Test";
+#else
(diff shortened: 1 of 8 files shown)


Sources/SentryCrash/Recording/SentryCrash.m
```diff
@@ -67,10 +67,19 @@
...
+    static NSString *bundleName = nil;
+
     if (bundleName == nil) {
-        bundleName = @"Unknown";
+#if TEST
+        bundleName = @"Sentry/Test";
+#else
+        bundleName =
+            [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleName"] ?: @"Unknown";
+#endif
```
philipphofmann, line 78: `m`: I would prefer finding a different way of testing this. One way would be to define an extra method or class that handles the edge case below and directly call it in tests.

Sources/SentryCrash/Recording/SentryCrash.m
```diff
@@ -67,10 +67,19 @@
 static NSString *
 getBundleName(void)
 {
-    NSString *bundleName = [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleName"];
+    static NSString *bundleName = nil;
+
     if (bundleName == nil) {
-        bundleName = @"Unknown";
+#if TEST
+        bundleName = @"Sentry/Test";
```
rakuyoMo, line 74: I think this is very dangerous. In many componentized projects, developers bring some macro definitions to pod components, and TEST is a very common one, used as a marker in test environments. In this case the bundleName will always be "Sentry/Test"

brustolin: Although I believe a dev should not add a macro definition to a external library pod this is a fair point, but we do have other places with TEST in the project.
We probably should rethink this.

rakuyoMo: > Although I believe a dev should not add a macro definition to a external library pod this is a fair point, but we do have other places with TEST in the project.
> We probably should rethink this.

In a componentized project, although we can use some hard coding to distinguish which are business components and which are third-party components, some inexperienced developers may ignore this and bring macro definitions to all pods.

I think the right approach is that the third-party pod and the main app each ensure the uniqueness of their own macro definitions.

Sources/SentryCrash/Recording/SentryCrash.m
```diff
@@ -562,6 +534,35 @@ - (void)applicationWillTerminate
     sentrycrash_notifyAppTerminate();
 }
 
+- (NSString *)clearBundleName:(NSString *)filename
+{
+    return [filename stringByReplacingOccurrencesOfString:@"/" withString:@"-"];
```
philipphofmann, line 539: `m`: Please briefly comment on why we added this logic.

philipphofmann (approved):
Thanks, @brustolin.

armcknight:
Where can a valid bundle ID be made with a forward slash? ADP just won't allow it (screenshot 1), and Xcode automatically converts my input (screenshot 2) to a hyphen after pressing enter (screenshot 3), and that change is reflected in the xcodeproj (screenshot 4): 
image
image
image
image

Sources/SentryCrash/Recording/SentryCrash.m
brustolin, line 539:
```suggestion
    //The bundle name is used as file name, therefore "/" is not allowed.
    return [filename stringByReplacingOccurrencesOfString:@"/" withString:@"-"];
```

PR #3010, opened 2023-05-08
ci: retry codecov steps one time if they fail

armcknight:
I've seen many codecov step failures, often due to network request failures (see [latest example](https://github.com/getsentry/sentry-cocoa/actions/runs/4913855191/jobs/8776836428?pr=2995)). Perform one retry just in case. Because there's no script invocation, we can't use the same kind of retry strategy as elsewhere where we use a bash for loop, and we can't use the [retry-step action from the GitHub Action marketplace](https://github.com/marketplace/actions/retry-step), so I used the duplicate action strategy mentioned here: https://www.thisdot.co/blog/how-to-retry-failed-steps-in-github-action-workflows

#skip-changelog

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -229,11 +229,22 @@ jobs:
       # We don't upload codecov for release branches, as we don't want a failing coverage check to block a release.
       # We don't upload codecov for scheduled runs as CodeCov only accepts a limited amount of uploads per commit.
       - name: Push code coverage to codecov
+        id: codecov_1
         uses: codecov/codecov-action@894ff025c7b54547a9a2a1e9f228beae737ad3c2 # pin@v3.1.3
         if: ${{ contains(matrix.platform, 'iOS') && !contains(github.ref, 'release') && github.event.schedule == '' }}
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -229,11 +229,22 @@ jobs:
...
         uses: codecov/codecov-action@894ff025c7b54547a9a2a1e9f228beae737ad3c2 # pin@v3.1.3
         if: ${{ contains(matrix.platform, 'iOS') && !contains(github.ref, 'release') && github.event.schedule == '' }}
         with:
           # Although public repos should not have to specify a token there seems to be a bug with the Codecov GH action, which can
           # be solved by specifying the token, see https://github.com/codecov/codecov-action/issues/557#issuecomment-1224970469
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: false
+          verbose: true
+
+      # Sometimes codecov uploads etc can fail. Retry one time to rule out e.g. intermittent network failures.
```
philipphofmann, line 242: 💯 for the comment on why we need this.

PR #3011, opened 2023-05-08
ci: only archive derived data logs if the build step fails

armcknight:
I noticed a case where it took several minutes to archive derived data logs [here](https://github.com/getsentry/sentry-cocoa/actions/runs/4914931513/jobs/8777452063?pr=3009): 
image

This happened due to a test failure. Now that we've split up the build and runtest steps, I think we only need derived data logs if the build step fails. Once the build succeeds, these logs can't help for test run failures, so the step linked above didn't actually need to run.

#skip-changelog

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -194,7 +194,16 @@ jobs:
 
       # We split building and running tests in two steps so we know how long running the tests takes.
       - name: Build tests
+        id: build_tests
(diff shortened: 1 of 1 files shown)


armcknight:
Well, got a natural experiment with a flake, at least the changes are validated 😆 https://github.com/getsentry/sentry-cocoa/actions/runs/4915463064/jobs/8778134993?pr=3011

brustolin (approved):

PR #3016, opened 2023-05-09
feat: Remove PII from screenshots

brustolin:
## :scroll: Description

Use iOS Vision to remove text from Screenshots.

image

Samples

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ This change has no impact on grouping of the issues in Sentry.
 ### Fixes 
 
 - Propagate span when copying scope (#2952)
+- Remove "/" from crash report file name (#3005)
 
 ## 8.6.0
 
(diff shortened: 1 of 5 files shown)


brustolin:
This solution is a little bit slow, during tests in the simulator, the simple image in the description took 2 seconds to finish. We may have to consider running this in a background thread, but in order for this to work we would have to change the `captureError` approach as well.

philipphofmann (commented):
@brustolin, what about making this an experimental feature to get feedback? We would need to add an experimental option, a way to track the adoption, some tests maybe, and some docs. WDYT?

bruno-garcia:
> We have made the decision to close this solution due to its lack of maturity. The process of removing PII takes an excessive amount of time and even then, there is no guarantee that it will be completely removed. As a result, users would still need to exercise caution when using this solution.

Our server side PII stripping is also not a silver bullet. We just need to document the limitations and trade offs.

PR #3018, opened 2023-05-09
fix: synchronize and copy global profiling data structures

armcknight:
## :scroll: Description

We've had several customer reports of crashes involving an enumeration happening in json validation, which loops through a globally-modifiable data structure we should be copying before placing in the profiling payload.

Here's the representative stack trace reported by one customer:

image

This stack trace also appears in the report uploaded by a customer in https://github.com/getsentry/sentry-cocoa/issues/2993, although it was not the blamed frame/thread, but it's possible that Crashlytics just isn't analyzing the crash correctly and this is also the root cause of that crash.

## :pencil: Checklist


Sources/Sentry/include/SentryTransactionContext+Private.h
```diff
@@ -37,6 +37,9 @@ SentryTransactionContext (Private)
                parentSampled:(SentrySampleDecision)parentSampled;
 
 #if SENTRY_TARGET_PROFILING_SUPPORTED
+
+@property (nonatomic, strong) SentryThread *threadInfo;
```
armcknight, line 41: Needed to expose this so we can set it in the new test case.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -475,6 +533,25 @@ + (void)useFramesTracker:(SentryFramesTracker *)framesTracker
 
 #    pragma mark - Private
 
++ (NSDictionary<NSString *, id> *)serializeForTransaction:(SentryTransaction *)transaction
+                                                profileID:(SentryId *)profileID
+{
+    std::lock_guard<std::mutex> l(_gProfilerLock);
```
philipphofmann, line 539: `m`: If I'm not mistaken, this lock will be active until we finish serializing the profiling data below. I am unsure if that could cause some slowdowns in the code. I think we could unlock after copying the profiling data.

philipphofmann (commented):
I still need to have a look a the tests. That's my first pass.

Sources/Sentry/include/SentryTransactionContext+Private.h
philipphofmann: `l`: Please add a comment that it's only required for testing, then.

Tests/SentryProfilerTests/SentryProfilerTests.mm
```diff
@@ -168,12 +171,118 @@ - (void)testProfilerMutationDuringSlicing
...
+
+    const auto index =
+        [frames indexOfObjectPassingTest:^BOOL(NSDictionary<NSString *, id> *_Nonnull obj,
+            __unused NSUInteger idx, __unused BOOL *_Nonnull stop) {
+            NSString *address = obj[@"instruction_addr"];
+            const auto unexpected =
+                @[ @"0x0000000000000777", @"0x0000000000000777", @"0x0000000000000777" ];
+            return [unexpected containsObject:address];
+        }];
+    XCTAssertEqual(index, NSNotFound);
```
philipphofmann, line 277: `l`: An assert message describing the assertion failure would be nice.

Sources/Sentry/SentryProfiler.mm
armcknight: This isn't actually new logic, it just didn't diff well. It is the same behavior as is found here in the "removed" side of the diff, that came out of the refactor: https://github.com/getsentry/sentry-cocoa/pull/3018/files#diff-9d7d24f33a2554811f36d8236cdd5b9b2c7c2786b12be31e4a05a3934193441cL356

You might be on to something here, but I'm hesitant to change how this lock works in this PR. It forces exclusivity between querying whether the profiler is running, starting and stopping it (not only normal stops, but also due to elapsed timeout timer and app going to background), and querying for profiler info like is happening in this function.

Tests/SentryProfilerTests/SentryProfilerTests.mm
```diff
@@ -168,12 +171,118 @@ - (void)testProfilerMutationDuringSlicing
     [self waitForExpectationsWithTimeout:1 handler:nil];
 }
 
+- (void)testProfilerMutationDuringSerialization
```
philipphofmann, line 174: `m`: It looks like we could mutate the profiling data on another thread while serializing. Maybe it makes sense to add a test similar to 
https://github.com/getsentry/sentry-cocoa/blob/1bf8571322c66325de9718c4e6e850f784d22411/Tests/SentryTests/TestLogOutput.swift#L31-L38

armcknight: Can you share the process by which this thread-unsafe mutation can still occur? This PR is making the change to prevent exactly that situation. `_gDataStructureLock` guards the point where those data structures are mutated: https://github.com/getsentry/sentry-cocoa/blob/971805f9f02eef494b69328ccfe7a81f15fc5e28/Sources/Sentry/SentryProfiler.mm#L160-L161

armcknight: and then where those data structures are copied: https://github.com/getsentry/sentry-cocoa/blob/971805f9f02eef494b69328ccfe7a81f15fc5e28/Sources/Sentry/SentryProfiler.mm#L280C39-L285

Sources/Sentry/SentryProfiler.mm
```diff
@@ -264,6 +266,136 @@
...
+    NSDictionary<NSString *, id> *serializedMetrics, NSArray<SentryDebugMeta *> *debugMeta)
+{
+    NSArray<SentrySample *> *samplesCopy;
+    NSArray<NSArray<NSNumber *> *> *stacksCopy;
+    NSArray<NSDictionary<NSString *, id> *> *framesCopy;
+    NSDictionary<NSString *, NSDictionary *> *threadMetadataCopy;
+    NSDictionary<NSString *, NSDictionary *> *queueMetadataCopy;
+    {
+        std::lock_guard<std::mutex> d(_gDataStructureLock);
+        samplesCopy = [profileData[@"profile"][@"samples"] copy];
```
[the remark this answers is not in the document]
armcknight: Definitely. @indragiek and I have discussed a memory footprint optimization of using a ring-buffer, which will be a good thing to keep in mind for continuous profiling, I think it would integrate well with the needs to stream chunks of profiling data to the server.

Sources/Sentry/include/SentryProfiler+Test.h
```diff
@@ -4,18 +4,26 @@
...
+@class SentryId;
 @class SentrySample;
+@class SentryTransaction;
 
 NS_ASSUME_NONNULL_BEGIN
 
 void processBacktrace(const sentry::profiling::Backtrace &backtrace,
     NSMutableDictionary<NSString *, NSMutableDictionary *> *threadMetadata,
     NSMutableDictionary<NSString *, NSDictionary *> *queueMetadata,
-    NSMutableArray<SentrySample *> *samples, NSMutableArray<NSMutableArray<NSNumber *> *> *stacks,
```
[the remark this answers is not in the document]
indragiek: Yes, it can change for a thread between samples if we find that the metadata is not present on one sample but is present for a subsequent sample on the same thread. This occasionally happens when we can't read the metadata for a thread (unclear why this happens but it tends to happen sometimes for a newly created thread).

indragiek (changes requested):

Sources/Sentry/include/SentryProfiler+Test.h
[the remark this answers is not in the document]
armcknight: Cool, thanks for confirming!

Tests/SentryProfilerTests/SentryProfilerTests.mm
philipphofmann: My apologies. I suggested a solution to a problem I didn't explain. When I remove the following line of code https://github.com/getsentry/sentry-cocoa/blob/a337836ce7a68f2b8f9c1a304f6abbf011a44ca0/Sources/Sentry/SentryProfiler.mm#L280

we still get crashes when serializing the data. It would be great if we could develop a test ensuring that we have the proper mechanisms in place so that such crashes don't happen anymore.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -155,7 +157,7 @@
     }
 
     {
-        std::lock_guard<std::mutex> l(_gSamplesArrayLock);
+        std::lock_guard<std::mutex> l(_gDataStructureLock);
```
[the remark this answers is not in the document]
armcknight: Ouch, ok let me get a test written for this.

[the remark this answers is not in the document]
armcknight: Wrote a test that reproduced this behavior in https://github.com/getsentry/sentry-cocoa/pull/3018/commits/a51cf976b5ccfaf9d7cd2c54a020d2003de3c961 and fixed in the next commit 👍🏻

Tests/SentryProfilerTests/SentryProfilerTests.mm
armcknight: Thanks for following up Philipp! So this PR is really doing two things:
- doing the copies of the data structures to prevent modification of the data during serialization (the `copy` operations)
- forcing exclusive access to the data structures preventing a race condition (the new and expanded critical sections guarded by the locks)

If you were to convert the calls to `copy` right below that lock guard to simple assignments to the same data structures, then you'd see the new tests I wrote fail.

The locks are only there to prevent the data race on those data structures, which is only going to be a very small edge case that actually causes a crash. The stacktrace in the PR description shows that there's a block enumeration in the call to isValidJSONObject, where if these data structures are modified in `processBacktrace` at just that time, a crash will occur. I think there could even be a data race with the copy calls if there were no locks, because what happens if the data structures are modified from another thread _during_ the execution of the copy?

Sources/Sentry/SentryProfiler.mm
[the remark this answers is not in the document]
armcknight: Note that only the threadMetadata structure still had this issue, because it has a mutable subdictionary, where the copy of the top-level structure still referenced the same mutable instance that could later be modified. The solution was to do a copy one level deeper. None of the other data structures had this issue, so the calls to `copy` of the top-level structures is sufficient.

philipphofmann (commented):
From my side LGTM. I leave this up to Indragie, as he requested changes.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -155,7 +143,30 @@
     }
 
     {
-        std::lock_guard<std::mutex> l(_gSamplesArrayLock);
+        // critical section: here is where all mutable data structures are updated, which must be
```
indragiek, line 146: `stacks` is still being mutated outside of the critical section on line 154

indragiek (changes requested):

Sources/Sentry/SentryProfiler.mm
armcknight: Hm, for some reason I thought I wouldn't need to put the modification of `stacks` or `frames` here, but now I can't remember why, because we do put the modification for `samples` here and it's essentially the same structure. I'll add them.

indragiek (approved):

PR #3028, opened 2023-05-15
feat: Double check-lock improvement

brustolin:
## :scroll: Description

Created a class to help with a double-checking and blocking approach to avoiding race condition.
It also makes it easier to test for such a condition.

`testFlush_CalledMultipleTimes_ImmediatelyReturnsFalse` was Flaky, as most of the tests trying to simulate race condition.

Unit test

## :pencil: Checklist

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-Swift.xcscheme
+++ b/Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-Swift.xcscheme
@@ -46,6 +46,16 @@
                </Test>
             </SkippedTests>
          </TestableReference>
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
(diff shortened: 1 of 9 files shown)


philipphofmann (commented):
Generally, I like the idea, and thanks a lot for looking into this, @brustolin. One downside of this approach is that the test doesn't work with the actual Objective-C APIs to validate if flushing is working correctly. This change turns the test into a unit test, but we also need to have an integration test to validate if flush works properly. That was the initial aim of `testFlush_CalledMultipleTimes_ImmediatelyReturnsFalse`. Maybe a mix of this PR and https://github.com/getsentry/sentry-cocoa/pull/3040 could be the ideal solution. 

Please have a look at https://github.com/getsentry/sentry-cocoa/pull/3040 and let me know what you think.

philipphofmann:
Can be closed as we copied parts from it to https://github.com/getsentry/sentry-cocoa/pull/3040 and fixed the flaky test with https://github.com/getsentry/sentry-cocoa/pull/3040.

Issue #2980, opened 2023-05-02
Memory leaks


vaibhavTM:
### Platform

iOS

### Installed

CocoaPods

### Version

8.1.0

### Steps to Reproduce

Just run the app

### Expected Result

No memory leaks.

### Actual Result

<img width="1368" alt="Screenshot 2023-05-02 at 2 29 56 PM" src="https://user-images.githubusercontent.com/83169376/235624692-aea6ee96-8e93-4aa0-ab90-473d628a6dd6.png">

### Are you willing to submit a PR?

No

brustolin:
Thanks @vaibhavTM for reaching out. We going to investigate this.

vaibhavTM:
Thankyou @brustolin  for quick response.

brustolin:
Hello @vaibhavTM.
I experience a few leaks when profiling is enabled (@armcknight can you check this?).
But In the latest version of the SDK I could not produce any of the leaks in your screenshot.

Can you update your Sentry SDK version and let us know please?

vaibhavTM:
Okay.

brustolin:
Out of curiosity I tried the SDK 8.1 and I was able to see the leaks.
So, Im confident that the leaks in the screenshot are fixed. 

We need to fixed leaks regarding profiling tho.

philipphofmann:
@vaibhavTM, try run `pod repo update` and try again, please.

armcknight:
Thanks for writing in @vaibhavTM . I'm looking at the possible leaks from the profiler now. I tried an instruments run with our Swift-iOS test app, and started a manual transaction with profiling and let it go until the profiler timed out. There are some symbols for me to investigate related to profiling: 
<img width="1112" alt="image" src="https://github.com/getsentry/sentry-cocoa/assets/3241469/c7ace62b-8364-4843-b08b-69bcc00508e5">

I hope to get updates on this today.

vaibhavTM:
@armcknight okay.

indragiek:
@armcknight @brustolin I made some progress tracking this down. Basically what appears to be happening is that we're capturing `SentryProfilingState` (previously just a bunch of arrays/dictionaries, but same idea) in the `std::function` that we pass to the `SamplingProfiler` initializer, but that captured `state` is never deallocated. It's not clear what's holding the reference to it - in the Xcode memory graph it looks like this:

[...]

PR #3035, opened 2023-05-16
fix: additional profiling data races

indragiek:
## :scroll: Description

While debugging the memory leaks reported in #2980 I found additional data race issues on top of the ones fixed in #3018. This is a larger refactor that fixes those issues but also simplifies the way that synchronization works.

In #3018, there was a change to add locking where `processBacktrace` mutated data structures that could be accessed concurrently. It attempted to separate the code that did not access the shared data structures, and moved all mutating code to a section at the end of the function. However, it missed a few reads from those data structures that also needed to be synchronized (e.g. `frames.count` and `stacks.count`). Generally, separating the critical section from the rest of the logic in that function is not worth the pain because the taking the lock for the entire duration of the function is not expensive.

This refactor adds a new type, `SentryProfilingState`, that centralizes all logic for reading/writing the shared data structures. This has the additional benefit that you no longer need a `processBacktrace` function that takes many parameters where the lifetime of those parameters is unclear (this will be helpful for fixing the memory leaks in a subsequent PR).

All notable changes have been commented on inline.

Ran all existing tests, they pass.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -295,7 +295,8 @@ serializedProfileData(NSDictionary<NSString *, id> *profileData, SentryTransacti
 
 @implementation SentryProfilingMutableState
 
-- (instancetype)init {
+- (instancetype)init
+{
     if (self = [super init]) {
         _samples = [NSMutableArray<SentrySample *> array];
         _stacks = [NSMutableArray<NSArray<NSNumber *> *> array];
@@ -330,10 +331,11 @@ serializedProfileData(NSDictionary<NSString *, id> *profileData, SentryTransacti
     block(_mutableState);
 }
 
-- (void)appendBacktrace:(const Backtrace &)backtrace {
+- (void)appendBacktrace:(const Backtrace &)backtrace
+{
     [self mutate:^(SentryProfilingMutableState *state) {
         const auto threadID = sentry_stringForUInt64(backtrace.threadMetadata.threadID);
-        
+
         NSString *queueAddress = nil;
         if (backtrace.queueMetadata.address != 0) {
             queueAddress = sentry_formatHexAddressUInt64(backtrace.queueMetadata.address);
@@ -344,21 +346,23 @@ serializedProfileData(NSDictionary<NSString *, id> *profileData, SentryTransacti
             state.threadMetadata[threadID] = metadata;
         }
         if (!backtrace.threadMetadata.name.empty() && metadata[@"name"] == nil) {
-            metadata[@"name"] = [NSString stringWithUTF8String:backtrace.threadMetadata.name.c_str()];
+            metadata[@"name"] =
+                [NSString stringWithUTF8String:backtrace.threadMetadata.name.c_str()];
         }
         if (backtrace.threadMetadata.priority != -1 && metadata[@"priority"] == nil) {
             metadata[@"priority"] = @(backtrace.threadMetadata.priority);
         }
         if (queueAddress != nil && state.queueMetadata[queueAddress] == nil
             && backtrace.queueMetadata.label != nullptr) {
-            state.queueMetadata[queueAddress] =
-                @ { @"label" : [NSString stringWithUTF8String:backtrace.queueMetadata.label->c_str()] };
+            state.queueMetadata[queueAddress] = @ {
+                @"label" : [NSString stringWithUTF8String:backtrace.queueMetadata.label->c_str()]
+            };
         }
-    #    if defined(DEBUG)
+#    if defined(DEBUG)
         const auto symbols
             = backtrace_symbols(reinterpret_cast<void *const *>(backtrace.addresses.data()),
                 static_cast<int>(backtrace.addresses.size()));
-    #    endif
+#    endif
 
         const auto stack = [NSMutableArray<NSNumber *> array];
         for (std::vector<uintptr_t>::size_type backtraceAddressIdx = 0;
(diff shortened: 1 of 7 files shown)


Sources/Sentry/SentryProfiler.mm
```diff
@@ -49,7 +50,6 @@
 #    endif
 
 const int kSentryProfilerFrequencyHz = 101;
-NSString *const kTestStringConst = @"test";
```
indragiek, line 52: This constant was unused so I removed it

Sources/Sentry/SentryProfiler.mm
```diff
@@ -421,12 +293,150 @@
...
+- (void)mutate:(void (^)(SentryProfilingMutableState *))block
+{
+    NSParameterAssert(block);
+    std::lock_guard<std::mutex> l(_lock);
+    block(_mutableState);
+}
+
+- (void)appendBacktrace:(const Backtrace &)backtrace {
+    [self mutate:^(SentryProfilingMutableState *state) {
+        const auto threadID = sentry_stringForUInt64(backtrace.threadMetadata.threadID);
```
indragiek, line 337: The logic inside `mutate` is essentially unchanged from the original logic of `processBacktrace` prior to the changes in #3018. The original code is here: https://github.com/getsentry/sentry-cocoa/blob/b7bd71993d0f05fd733eb36c665b7048a476e4fd/Sources/Sentry/SentryProfiler.mm#L91-L160

We now take a single lock to make the entire mutation which simplifies the code, there is no fine grained locking.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -421,12 +293,150 @@
...
+    
+    return @{
+        @"profile": @{
+            @"samples": samples,
+            @"stacks": stacks,
+            @"frames": frames,
+            @"thread_metadata": threadMetadata,
+            @"queue_metadata": queueMetadata
+        }
+    };
```
indragiek: This logic is preserved from the changes in #3018

Sources/Sentry/SentryProfiler.mm
```diff
@@ -421,12 +293,150 @@
...
+
+@end
+
 @implementation SentryProfiler {
-    NSMutableDictionary<NSString *, id> *_profileData;
+    SentryProfilingState *_state;
     std::shared_ptr<SamplingProfiler> _profiler;
     SentryMetricProfiler *_metricProfiler;
     SentryDebugImageProvider *_debugImageProvider;
-    thread::TIDType _mainThreadID;
```
indragiek, line 429: This was unused

Tests/SentryProfilerTests/SentryProfilerTests.mm
```diff
@@ -60,15 +60,7 @@ - (void)testProfilerCanBeInitializedOffMainThread
...
-    const auto resolvedThreadMetadata =
-        [NSMutableDictionary<NSString *, NSMutableDictionary *> dictionary];
-    const auto resolvedQueueMetadata = [NSMutableDictionary<NSString *, NSDictionary *> dictionary];
-    const auto resolvedStacks = [NSMutableArray<NSArray<NSNumber *> *> array];
-    const auto resolvedSamples = [NSMutableArray<SentrySample *> array];
-    const auto resolvedFrames = [NSMutableArray<NSDictionary<NSString *, id> *> array];
-    const auto frameIndexLookup = [NSMutableDictionary<NSString *, NSNumber *> dictionary];
-    const auto stackIndexLookup = [NSMutableDictionary<NSString *, NSNumber *> dictionary];
-
+    SentryProfilingState *state = [[SentryProfilingState alloc] init];
```
indragiek, line 63: `SentryProfilingState` simplifies these tests

Sources/Sentry/SentryProfiler.mm
```diff
@@ -79,108 +79,6 @@
...
-            const auto frame = [NSMutableDictionary<NSString *, id> dictionary];
-            frame[@"instruction_addr"] = instructionAddress;
-#    if defined(DEBUG)
-            const auto symbols
-                = backtrace_symbols(reinterpret_cast<void *const *>(backtrace.addresses.data()),
-                    static_cast<int>(backtrace.addresses.size()));
-            frame[@"function"] = parseBacktraceSymbolsFunctionName(symbols[backtraceAddressIdx]);
-#    endif
-            [stack addObject:@(frames.count)];
-            frameIndexLookup[instructionAddress] = @(frames.count);
```
indragiek, line 111: In the old implementation the logic was changed to use the `framesToAdd` array as a temporary container before applying the append all at once in the critical section, but that means that `frames.count` will now be incorrect here.

brustolin (approved):
LGTM

Tests/SentryProfilerTests/SentryProfilerTests.mm
```diff
@@ -153,9 +145,8 @@ - (void)testProfilerMutationDuringSlicing
     mutateExpectation.expectedFulfillmentCount = operations;
 
     void (^mutateBlock)(void) = ^(void) {
-        processBacktrace(backtrace, resolvedThreadMetadata, resolvedQueueMetadata, resolvedSamples,
-            resolvedStacks, resolvedFrames, frameIndexLookup, stackIndexLookup);
-        [mutateExpectation fulfill];
+        [state mutate:^(
+            __unused SentryProfilingMutableState *mutableState) { [mutateExpectation fulfill]; }];
```
armcknight, line 149: @indragiek I just noticed this change when going to write some new tests. Is the last backtrace created above ([here](https://github.com/getsentry/sentry-cocoa/pull/3035/files#diff-bfe2315bfeb5647454465dc52d0a0ba09dd139140e3292db7054a329a946e22bL149)) no longer appended to the profile data?

PR #3040, opened 2023-05-16
test: Fix flaky flush integration tests

philipphofmann:
The test testFlush_CalledMultipleTimes_ImmediatelyReturnsFalse is flaky. This PR tries to fix this by not relying on measuring the flush duration to identify whether the flush logic exited early. Instead, we now use an enum to identify the flush result and assert against it.

This PR is a similar proposal to https://github.com/getsentry/sentry-cocoa/pull/3028.

#skip-changelog


Sources/Sentry/include/SentryTransport.h
```diff
@@ -6,14 +6,21 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef NS_ENUM(NSInteger, SentryTransportFlushResult) {
+    kSentryTransportFlushResultSuccess = 0,
+    kSentryTransportFlushResultTimedOut,
+    kSentryTransportFlushResultAlreadyFlushingBeforeLock,
+    kSentryTransportFlushResultAlreadyFlushingInLock
+};
```
brustolin, line 14: l: Im not a huge fan of this over-engineering just to be able to test it.
This values are not used anywhere in the project other than tests, and `kSentryTransportFlushResultAlreadyFlushingInLock` is never used.

brustolin (commented):
I took a first look, but there is something bothering me.
I will have to do a deep dive into SentryHttpTransport to understand what's happening.

brustolin (approved):
Please find below the corrected version of the text:

I needed some time to understand what is going on with `testFlush_BlocksCallingThread_FinishesFlushingWhenSent`. If I got it right, the problem is that we rely on the `SentryRequestManager` mock and a set of configurations to be able to test the race condition. That's why I prefer the straightforward approach of #3028. However, it seems that some requirements have also changed, so the other PR would also need to be modified.

After studying SentryHTTPTransport (and the entire Transport layer), I believe we could simplify it. But this is a topic for another time.

This changes works, therefore LGTM!

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -186,16 +186,12 @@ - (BOOL)flush:(NSTimeInterval)timeout
 
     intptr_t result = dispatch_group_wait(self.dispatchGroup, dispatchTimeout);
 
-    @synchronized(self) {
-        self.isFlushing = NO;
-    }
```
brustolin, line 191: m: Is this supposed to be removed?

philipphofmann: Good catch, no I added it again.

Tests/SentryTests/Networking/SentryHttpTransportTests.swift
```diff
@@ -742,24 +744,20 @@ class SentryHttpTransportTests: XCTestCase {
         ensureFlushingGroup.enter()
         ensureFlushingQueue.async {
             ensureFlushingGroup.leave()
```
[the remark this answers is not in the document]
philipphofmann: Not required with the callback.

Sources/Sentry/include/SentryTransport.h
```diff
@@ -6,14 +6,25 @@
...
 
 - (void)recordLostEvent:(SentryDataCategory)category reason:(SentryDiscardReason)reason;
 
-- (BOOL)flush:(NSTimeInterval)timeout;
+- (SentryFlushResult)flush:(NSTimeInterval)timeout;
+
+/**
+ * Only needed for testing.
+ */
+- (void)setStartFlushCallback:(void (^)(void))callback;
```
brustolin, line 27: m: I believe that everything that is meant only for test should be wrapper by `#if TEST` to save a few bytes, this applies to implementation as well.

Also, this function is only required for tests with `SentryHTTPTransport`. Therefore, it could be placed in the `SentryHTTPTransport` class, eliminating the need to add an empty function for TestTransport.

brustolin (approved):
LGTM. 

I added one suggestion.

PR #3042, opened 2023-05-18
fix: use-after-free ASAN warning

armcknight:
## :scroll: Description

A customer reported that ASAN reports a use-after-free in Sentry code after assigning a UIImage to an image view, which is initialized with an empty string for the asset name. This draft PR so far reproduces the issue in a test case for further investigation. 

image

## :pencil: Checklist

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-Swift.xcscheme
+++ b/Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-Swift.xcscheme
@@ -26,7 +26,8 @@
       buildConfiguration = "Test"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      enableAddressSanitizer = "YES">
       <Testables>
          <TestableReference
             skipped = "NO">
@@ -52,6 +53,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      enableAddressSanitizer = "YES"
       launchStyle = "0"
       useCustomWorkingDirectory = "NO"
       ignoresPersistentStateOnLaunch = "NO"
--- a/Sources/SentryCrash/Recording/SentryCrashReport.c
+++ b/Sources/SentryCrash/Recording/SentryCrashReport.c
@@ -917,7 +917,7 @@ writeNotableStackContents(const SentryCrashReportWriter *const writer,
     for (uintptr_t address = lowAddress; address < highAddress; address += sizeof(address)) {
         if (sentrycrashmem_copySafely(
                 (void *)address, &contentsAsPointer, sizeof(contentsAsPointer))) {
-            sprintf(nameBuffer, "stack@%p", (void *)address);
+            //            sprintf(nameBuffer, "stack@%p", (void *)address);
             writeMemoryContentsIfNotable(writer, nameBuffer, contentsAsPointer);
         }
     }
--- a/Samples/iOS-Swift/iOS-Swift/Base.lproj/Main.storyboard
+++ b/Samples/iOS-Swift/iOS-Swift/Base.lproj/Main.storyboard
@@ -359,10 +359,10 @@
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
                             <stackView opaque="NO" contentMode="scaleToFill" distribution="fillEqually" translatesAutoresizingMaskIntoConstraints="NO" id="fn9-mQ-2PY">
-                                <rect key="frame" x="25" y="235.5" width="270" height="112"/>
+                                <rect key="frame" x="25" y="232.5" width="270" height="118.5"/>
                                 <subviews>
                                     <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" distribution="equalSpacing" translatesAutoresizingMaskIntoConstraints="NO" id="Q2k-6Y-RnD">
-                                        <rect key="frame" x="0.0" y="0.0" width="135" height="112"/>
+                                        <rect key="frame" x="0.0" y="0.0" width="135" height="118.5"/>
                                         <subviews>
                                             <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="system" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="NFC-V3-lgW">
                                                 <rect key="frame" x="0.0" y="0.0" width="135" height="28"/>
@@ -373,7 +373,7 @@
                                                 </connections>
                                             </button>
                                             <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="system" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="Llv-Yz-cwF">
-                                                <rect key="frame" x="0.0" y="28" width="135" height="28"/>
+                                                <rect key="frame" x="0.0" y="30" width="135" height="28"/>
                                                 <fontDescription key="fontDescription" type="system" pointSize="13"/>
                                                 <state key="normal" title="Capture NSException"/>
                                                 <connections>
@@ -381,7 +381,7 @@
                                                 </connections>
                                             </button>
                                             <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="system" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="wJ4-gS-64G" userLabel="fatalError">
-                                                <rect key="frame" x="0.0" y="56" width="135" height="28"/>
+                                                <rect key="frame" x="0.0" y="60" width="135" height="28"/>
                                                 <fontDescription key="fontDescription" type="system" pointSize="13"/>
                                                 <state key="normal" title="Throw FatalError"/>
                                                 <connections>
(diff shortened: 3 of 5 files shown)


armcknight:
Not sure why the UI test doesn't fail when running with ASAN on CI: https://github.com/getsentry/sentry-cocoa/actions/runs/5018316893/jobs/8997577913?pr=3042#step:4:529

armcknight:
I pushed a commit that avoids reading crash info from debug images for non-crash scenarios, but the part about the test not failing under ASAN in CI I think needs another look. It could come down to a different build configuration being used, but the problem also reproduced locally for me with the TestCI config. Could also be a difference in SDK versions or OS architectures.

Sources/SentryCrash/Recording/SentryCrashReport.c
```diff
@@ -917,7 +917,7 @@ writeNotableStackContents(const SentryCrashReportWriter *const writer,
     for (uintptr_t address = lowAddress; address < highAddress; address += sizeof(address)) {
         if (sentrycrashmem_copySafely(
                 (void *)address, &contentsAsPointer, sizeof(contentsAsPointer))) {
-            sprintf(nameBuffer, "stack@%p", (void *)address);
+            //            sprintf(nameBuffer, "stack@%p", (void *)address);
```
armcknight, line 920:
```suggestion
            sprintf(nameBuffer, "stack@%p", (void *)address);
```

Sources/SentryCrash/Recording/SentryCrashReport.c
```diff
@@ -1166,7 +1166,7 @@ writeBinaryImage(
     const SentryCrashReportWriter *const writer, const char *const key, const int index)
 {
     SentryCrashBinaryImage image = { 0 };
-    if (!sentrycrashdl_getBinaryImage(index, &image)) {
+    if (!sentrycrashdl_getBinaryImage(index, &image, /*isCrash*/ true)) {
```
armcknight, line 1169: This is the one place where we actually need to specify to attempt grabbing the debug image's crash info.

Sources/Sentry/include/SentryCrashBinaryImageProvider.h
```diff
@@ -10,8 +10,21 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
 - (SentryCrashBinaryImage)getBinaryImage:(NSInteger)index;
 
+/**
+ * Returns information for the image at the specified index.
+ * @param isCrash @c YES if we're collecting binary images for a crash report, @c NO if we're
+ * gathering them for other backtrace information, like a performance transaction. If this is for a
+ * crash, each image's data section crash info is also included.
+ */
+- (SentryCrashBinaryImage)getBinaryImage:(NSInteger)index isCrash:(BOOL)isCrash;
```
armcknight, line 26: Same question here about backwards compat in an `include/` header.

Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
```diff
@@ -40,10 +40,21 @@ typedef void (^SentryOnAppStartMeasurementAvailable)(
...
 + (NSArray<SentryDebugMeta *> *)getDebugImages;
 
+/**
+ * Returns the current list of debug images. Be aware that the @c SentryDebugMeta is actually
+ * describing a debug image.
+ * @param isCrash @c YES if we're collecting binary images for a crash report, @c NO if we're
+ * gathering them for other backtrace information, like a performance transaction. If this is for a
+ * crash, each image's data section crash info is also included.
+ */
++ (NSArray<SentryDebugMeta *> *)getDebugImagesCrashed:(BOOL)isCrash;
```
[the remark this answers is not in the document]
brustolin: Hybrid SDKs use this header, this change will introduce a breaking change to them, but this we may discuss if is necessary or not.

Sources/Sentry/include/SentryCrashBinaryImageProvider.h
brustolin: This one is totally private, no need for backward compatibility.

Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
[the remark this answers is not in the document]
armcknight: You mean it would break if I weren't to keep this backwards-compatible change?

Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-Swift.xcscheme
```diff
@@ -24,8 +24,8 @@
    </BuildAction>
    <TestAction
       buildConfiguration = "Test"
-      selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
-      selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      selectedDebuggerIdentifier = ""
+      selectedLauncherIdentifier = "Xcode.IDEFoundation.Launcher.PosixSpawn"
```
armcknight, line 28:
```suggestion
      selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
      selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
```

Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-Swift.xcscheme
```diff
@@ -49,9 +49,10 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Debug"
+      buildConfiguration = "TestCI"
```
armcknight, line 52:
```suggestion
      buildConfiguration = "Debug"
```

Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-Swift.xcscheme
```diff
@@ -49,9 +49,10 @@
...
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      enableAddressSanitizer = "YES"
```
armcknight, line 55:
```suggestion
```

Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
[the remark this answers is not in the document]
brustolin: Yes. If we remove the `+ (NSArray<SentryDebugMeta *> *)getDebugImages;` it would break.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Convert one of the two remaining usages of `sprintf` to `snprintf` (#2866)
+- Fix a use-after-free that could occur after initializing a UIImage with an empty name string (#3042)
```
philipphofmann, line 8: `m`: I think something like the following would be a more descriptive changelog; I don't think we need to explain the details.
```suggestion
- Fix use-after-free ASAN warning (#3042)
```

Sources/Sentry/Public/SentryDebugImageProvider.h
```diff
@@ -7,30 +7,68 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
 @interface SentryDebugImageProvider : NSObject
 
 - (instancetype)init;
 
 /**
  * Returns a list of debug images that are being referenced in the given threads.
- * @param threads A list of @c SentryThread that may or may not contain stacktracse.
+ * @param threads A list of @c SentryThread that may or may not contain stacktraces.
+ * @warning This assumes a crash has occurred and attempts to read the crash information from each
```
philipphofmann, line 19: ``m`: Should we maybe deprecate this method then?

Sources/SentryCrash/Recording/SentryCrashReport.c
philipphofmann: `h`: As this is the only place where `isCrash` is `true`, I think we don't need to bubble up `isCrash` in all the layers. The `SentryDebugImageProvider` and `PrivateSentrySDKOnly` could choose the default of `false`. The PR would be smaller and consumers of the getBinaryImage API don't have to decide if they should pass in `true` or `false` for `isCrash`. What do you think @armcknight?

Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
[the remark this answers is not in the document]
armcknight: great, thanks for confirming!

Sources/Sentry/Public/SentryDebugImageProvider.h
armcknight: I thought about that. I worried about inserting a deprecation notice in case people use warnings-as-errors, then we'd break their build. I'm ok to do it though, just let me know.

Sources/SentryCrash/Recording/SentryCrashReport.c
armcknight: If we do that, then this becomes a breaking change not just at build time by removing the old method from `SentryDebugImageProvider`, but also at runtime because the current behavior of the public API always attempts to gather the crash info and we'd be changing that.

Another option is to decouple the image provider from collecting crash info, and extract a SentryCrashInfoProvider that would be used here on line 1172, after retrieving the debug images and before writing them. Then all the other codepaths to `sentrycrashdl_getBinaryImage` wouldn't have to specify anything, and we could even put calls to SentryCrashInfoProvider in the backwards-compat methods to maintain the current behavior.

What do you think?

Sources/Sentry/Public/SentryDebugImageProvider.h
philipphofmann: I'm pretty sure almost nobody uses this API except for hybrid SDKs controlled by us. It even says so in the comment:https://github.com/getsentry/sentry-cocoa/blob/4053ee90b75473db21ccfc2e1f361f22e2d0544c/Sources/Sentry/Public/SentryDebugImageProvider.h#L8-L11

So I think it's fine to deprecate these methods. Actually, this class shouldn't be public. I created an issue for that https://github.com/getsentry/sentry-cocoa/issues/3056

armcknight: Cool, i'm good with that! Will add.

armcknight: Done in https://github.com/getsentry/sentry-cocoa/pull/3042/commits/58921ff1664f6d40bbfefde567d15fea25545b63. Also added a note to remove them in https://github.com/getsentry/sentry-cocoa/discussions/2946#discussioncomment-5988220.

philipphofmann (approved):
LGTM, but CI is not happy.

PR #2983, opened 2023-05-03
feat: stitch async Swift Concurrency frames

kabiroberai:
## :scroll: Description

Adds support for stitching async Swift Concurrency frames using [backtrace_async](https://keith.github.io/xcode-man-pages/backtrace.3.html#backtrace_async).

## :bulb: Motivation and Context

Switching from `backtrace` to `backtrace_async` is super trivial and gets Sentry 90% of the way to offering good Swift Concurrency support. The man page mentions that backtrace_async is identical to backtrace in all circumstances except when Swift Concurrency `Task`s are involved, so IMO it's a no-brainer to make the jump.

Unfortunately the story isn't as easy for profiling, since the code there manually walks the stack — I think a good approach there could be to mimic how backtrace_async is [implemented](https://github.com/apple-oss-distributions/Libc/blob/dbde50970329ace8793fca2b63d708d280c33d88/gen/backtrace.c#L47-L53) in Darwin's libc.

Fixes #1919.

## :green_heart: How did you test it?

Unit tests + a production app

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I reviewed the submitted code.
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

After this, it'd be amazing to see support for Swift Concurrency in profiling — though I suspect that requires a greater eng lift.

PR #3051, opened 2023-05-22
feat: Experimental support for Swift Async stacktraces

### LINKED ISSUE #1919: Stacktraces for Swift Concurrency
### Platform

iOS

### Installed

Swift Package Manager

### Version

7.16.0

### Steps to Reproduce

1. Create an async function, with a suspension point (such as `await Task.sleep(…)`)
2. Capture a Sentry event after the suspension point

It should be possible to repro this with the sample introduced in #1504

### Expected Result

The Sentry dashboard shows the entire stack trace, including the async function’s caller

### Actual Result

The stack trace stops at the function inside which the error was logged, showing the caller as `swift::runJobInEstablishedExecutorContext`

Sentry would ideally stitch together parent Task stack frames, similar to how `stitchAsyncCode` from #998 works for libdispatch.

brustolin:
## :scroll: Description

> This PR was created By [kabiroberai](https://github.com/kabiroberai) at #2983, but CI was strange and I wanted to test this properly.

Adds support for stitching async Swift Concurrency frames using [backtrace_async](https://keith.github.io/xcode-man-pages/backtrace.3.html#backtrace_async).

Switching from `backtrace` to `backtrace_async` is super trivial and gets Sentry 90% of the way to offering good Swift Concurrency support. The man page mentions that backtrace_async is identical to backtrace in all circumstances except when Swift Concurrency `Task`s are involved, so IMO it's a no-brainer to make the jump.

Unfortunately the story isn't as easy for profiling, since the code there manually walks the stack — I think a good approach there could be to mimic how backtrace_async is [implemented](https://github.com/apple-oss-distributions/Libc/blob/dbde50970329ace8793fca2b63d708d280c33d88/gen/backtrace.c#L47-L53) in Darwin's libc.

Fixes #1919.

Unit tests + a production app

## :pencil: Checklist

## :crystal_ball: Next steps

After this, it'd be amazing to see support for Swift Concurrency in profiling — though I suspect that requires a greater eng lift.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -130,14 +130,6 @@ jobs:
             xcode: '14.3'
             test-destination-os: 'latest'
 
-          # MetricKit doesn't exist for tvOS, so we can still run unit tests with
-          # Xcode 12 for it.
-          # tvOS 14
-          - runs-on: macos-11
-            platform: 'tvOS'
-            xcode: '12.5.1'
-            test-destination-os: 'latest'
-
           # tvOS 15
           - runs-on: macos-12
             platform: 'tvOS'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Support for stitching async Swift Concurrency frames (#3051)
+- Experimental support for Swift Async stacktraces (#3051)
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Support for stitching async Swift Concurrency frames
+
 ### Fixes
 
 - Convert one of the two remaining usages of `sprintf` to `snprintf` (#2866)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Support for stitching async Swift Concurrency frames
+- Support for stitching async Swift Concurrency frames (#3051)
 
 ### Fixes
 
--- a/Sources/Sentry/SentrySwiftAsyncIntegration.m
+++ b/Sources/Sentry/SentrySwiftAsyncIntegration.m
@@ -1,0 +1,16 @@
+#import "SentrySwiftAsyncIntegration.h"
+#import "SentryCrashStackCursor_SelfThread.h"
+
+@implementation SentrySwiftAsyncIntegration
+
+- (BOOL)installWithOptions:(nonnull SentryOptions *)options
+{
+    sentrycrashsc_setSwiftAsyncStitching(options.stitchSwiftAsync);
+    return options.stitchSwiftAsync;
+}
+
+- (void)uninstall {
+    sentrycrashsc_setSwiftAsyncStitching(NO);
+}
+
+@end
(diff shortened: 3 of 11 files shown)


philipphofmann (commented):
If I remember correctly, we planned on rolling this out as an experimental trackable feature. Do you plan on adding this to this PR, @brustolin?

brustolin:
> f I remember correctly, we planned on rolling this out as an experimental trackable feature. Do you plan on adding this to this PR, @brustolin?

Sure, we may add a flag. First I need to understand way this is not working with CI

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Support for stitching async Swift Concurrency frames (#3051)
```
philipphofmann, line 7: `l`:
```suggestion
- Experimental support for Swift Async stacktraces (#3051)
```

Sources/Sentry/SentrySwiftAsyncIntegration.m
```diff
@@ -0,0 +1,17 @@
+#import "SentrySwiftAsyncIntegration.h"
+#import "SentryCrashStackCursor_SelfThread.h"
+
+@implementation SentrySwiftAsyncIntegration
+
+- (BOOL)installWithOptions:(nonnull SentryOptions *)options
+{
+    sentrycrashsc_setSwiftAsyncStitching(options.stitchSwiftAsync);
```
philipphofmann, line 8: `h`: This will enable the feature for everyone, if I'm not mistaken. We need to do what we do for every integration https://github.com/getsentry/sentry-cocoa/blob/4053ee90b75473db21ccfc2e1f361f22e2d0544c/Sources/Sentry/SentryScreenshotIntegration.m#L23-L25

philipphofmann (commented):
Neat, a couple of suggestions:

Sources/Sentry/SentrySwiftAsyncIntegration.m
brustolin: In the next line the code returns `stitchSWiftAsync` which means the integration will not be installed if the feature is disabled. The default is `NO`

philipphofmann: Isn't calling super essential for the list of integrations sent with every event?

brustolin: No. `installWithOptions` is just a check that returns true or false whether the integration should be installed.
SentrySDK is responsible for the magic

https://github.com/getsentry/sentry-cocoa/blob/2405ba598c9bae766901b68fd3cf1a9c5dca97af/Sources/Sentry/SentrySDK.m#L356-L361

.github/workflows/test.yml
```diff
@@ -130,14 +130,6 @@ jobs:
             xcode: '14.3'
             test-destination-os: 'latest'
 
-          # MetricKit doesn't exist for tvOS, so we can still run unit tests with
-          # Xcode 12 for it.
-          # tvOS 14
-          - runs-on: macos-11
-            platform: 'tvOS'
-            xcode: '12.5.1'
-            test-destination-os: 'latest'
```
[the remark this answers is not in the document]
brustolin: I did not found the proper path and parameters to create a tvOS simulator, but before I invest more time on this, do you really think this is necessary?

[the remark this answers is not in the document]
brustolin: BTW, I don't think is possible to publish a tvOS app with Xcode 12, [and in July Xcode 13 will be deprecated as well](https://developer.apple.com/tvos/submit/), so, I see no point running unit tests with this Xcodes.

philipphofmann (approved):
LGTM for an experimental feature. Awesome @brustolin. We just need to point it out in the docs, so people start using it.

Issue #3050, opened 2023-05-22
the iOS app crashed on Sentry SDK [SentrySerialization dataWithJSONObject:]


Mohsenkhodadadzadehthelist:
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.7.1

### Steps to Reproduce

we are using Sentry for our product since 3 days ago and the firebase crashlytics reported to us we had 3 crashes on the Sentry SDK

### Expected Result

work properly without crash

### Actual Result

The first one :
+[SentrySerialization dataWithJSONObject:]
NSInvalidArgumentException - Invalid type in JSON write ((null))
```
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x9d94 __exceptionPreprocess
1  libobjc.A.dylib                0x183d0 objc_exception_throw
2  Foundation                     0x2e61c _writeJSONValue
3  Foundation                     0xd4a70 ___writeJSONArray_block_invoke
4  CoreFoundation                 0xa2ac __NSARRAY_IS_CALLING_OUT_TO_A_BLOCK__
5  CoreFoundation                 0x25904 -[__NSArrayM enumerateObjectsWithOptions:usingBlock:]
6  Foundation                     0xa8d98 _writeJSONArray
7  Foundation                     0xc627c ___writeJSONObject_block_invoke
8  CoreFoundation                 0x1377c __NSDICTIONARY_IS_CALLING_OUT_TO_A_BLOCK__
9  CoreFoundation                 0xc0e94 -[__NSDictionaryI enumerateKeysAndObjectsWithOptions:usingBlock:]
10 Foundation                     0x27f70 _writeJSONObject
11 Foundation                     0xc627c ___writeJSONObject_block_invoke
12 CoreFoundation                 0x1377c __NSDICTIONARY_IS_CALLING_OUT_TO_A_BLOCK__
13 CoreFoundation                 0x9c1e0 -[__NSDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:]
14 Foundation                     0x27f70 _writeJSONObject
15 Foundation                     0x514cc -[_NSJSONWriter dataWithRootObject:options:]
16 Foundation                     0x50f00 +[NSJSONSerialization dataWithJSONObject:options:error:]
17 Sentry                         0x69678 +[SentrySerialization dataWithJSONObject:] + 27 (SentrySerialization.m:27)
[...]

PR #3054, opened 2023-05-23
fix: SIGSEGV crash in profiler's queue metadata access

armcknight:
## :scroll: Description

Add more defensive checks to try to fix the third crasher in https://github.com/getsentry/sentry-cocoa/issues/3050

    the crash report shows line 141 as the culprit. there are two
    operations there:

    1. idInfo->thread_handle != 0
        this is already checked on line 138 with
    sentrycrashmem_isMemoryReadable(idInfo, sizeof(*idInfo))

    2. *queuePtr != nullptr
        queuePtr is checked on line 140 with
    sentrycrashmem_isMemoryReadable(queuePtr, sizeof(*queuePtr))

    I'm wondering if sentrycrashmem_isMemoryReadable is somehow
    leaving the memory in a bad state? but in any case, put some extra
    defensive checks before these operations. One of them is a check
    for queuePtr == nullptr which is already done right before the
    call to sentrycrashmem_isMemoryReadable(queuePtr,
    sizeof(*queuePtr)) but again, if that function is somehow causing
    a side effect on queuePtr, we need to check again.

There is no repro so there's no way to test it currently.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryThreadHandle.cpp
+++ b/Sources/Sentry/SentryThreadHandle.cpp
@@ -137,8 +137,15 @@ namespace profiling {
         if (thread_info(handle_, THREAD_IDENTIFIER_INFO, info, &count) != KERN_SUCCESS) {
             return 0;
         }
+        if (info == nullptr) {
+            return 0;
+        }
+
         const auto idInfo = reinterpret_cast<thread_identifier_info_t>(info);
         if (!sentrycrashmem_isMemoryReadable(idInfo, sizeof(*idInfo))) {
+            return 0;
+        }
+        if (idInfo->dispatch_qaddr == 0) {
             return 0;
         }
 
@@ -150,10 +157,16 @@ namespace profiling {
             return 0;
         }
 
+        if (idInfo == nullptr) {
+            return 0;
+        }
         if (idInfo->thread_handle == 0) {
             return 0;
         }
 
+        if (queuePtr == nullptr) {
+            return 0;
+        }
         if (*queuePtr == nullptr) {
             return 0;
         }
--- a/Sources/Sentry/SentryThreadHandle.cpp
+++ b/Sources/Sentry/SentryThreadHandle.cpp
@@ -128,21 +128,37 @@ namespace profiling {
     ThreadHandle::dispatchQueueAddress() const noexcept
     {
         if (handle_ == THREAD_NULL) {
-            return {};
+            return 0;
         }
+
         integer_t infoBuffer[THREAD_IDENTIFIER_INFO_COUNT] = { 0 };
         thread_info_t info = infoBuffer;
         mach_msg_type_number_t count = THREAD_IDENTIFIER_INFO_COUNT;
-        const auto idInfo = reinterpret_cast<thread_identifier_info_t>(info);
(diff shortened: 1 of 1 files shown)


brustolin (approved):
LGTM

Sources/Sentry/SentryThreadHandle.cpp
```diff
@@ -128,21 +128,50 @@ namespace profiling {
...
+            return 0;
         }
+
         integer_t infoBuffer[THREAD_IDENTIFIER_INFO_COUNT] = { 0 };
         thread_info_t info = infoBuffer;
         mach_msg_type_number_t count = THREAD_IDENTIFIER_INFO_COUNT;
+        if (thread_info(handle_, THREAD_IDENTIFIER_INFO, info, &count) != KERN_SUCCESS) {
+            return 0;
+        }
+        if (info == nullptr) {
```
indragiek, line 140: `info` isn't a pointer, it's a stack allocated struct, what does the nullptr check do?

Sources/Sentry/SentryThreadHandle.cpp
```diff
@@ -128,21 +128,50 @@ namespace profiling {
...
+        const auto queuePtr = reinterpret_cast<dispatch_queue_t *>(idInfo->dispatch_qaddr);
+        if (queuePtr == nullptr) {
+            return 0;
         }
-        return 0;
+        if (!sentrycrashmem_isMemoryReadable(queuePtr, sizeof(*queuePtr))) {
+            return 0;
+        }
+
+        if (idInfo == nullptr) {
```
indragiek, line 160: `idInfo` is just a cast of `info`, it's not a pointer

indragiek (changes requested):
The new checks you added would not do anything because the nullptr checks are not checking values that could be pointers.

Sources/Sentry/SentryThreadHandle.cpp
armcknight: `info`'s type is `thread_info_t` which is typedef'd to `integer *`, so i thought maybe...

Sources/Sentry/SentryThreadHandle.cpp
armcknight: The pointer member access operator was used on it: `idInfo->thread_handle`, which is why i thought this one would need to be null checked before using

Sources/Sentry/SentryThreadHandle.cpp
indragiek: Ah, I see. Still quite unlikely if `thread_info` doesn't return a non-success code though.

indragiek: Technically yes, it's a pointer, but on line 135 set `info` to point to memory allocated on the stack. The only way this could be `nullptr` is if `info` was passed by reference to `thread_info` and `thread_info` set it to `nullptr`, but it is not passed by ref in this case.

Sources/Sentry/SentryThreadHandle.cpp
indragiek: See comment above - technically yes it's a pointer, but in this case it is pointing to a valid stack array and there's no possibility of it being null.

armcknight:
Looks like this won't be an easy problem to solve, and this PR isn't the right course of action.

PR #3055, opened 2023-05-24
fix: Profiling memory leaks

indragiek:
## :scroll: Description

This fixes most of the memory leaks reported in #2980. I checked before/after the fixes using Instruments.

**Before:**
CleanShot 2023-05-23 at 23 33 23@2x

**After:**
CleanShot 2023-05-23 at 23 33 28@2x

There is 1 memory leak left to fix that I couldn't find a solution for yet, but the other leaks are fixed.

The root of the problem seems to be that `std::thread` does not correctly deallocate parameters that are passed to the newly spawned thread. It's unclear if this is expected behavior or not, but the fix was simple: pass parameters that were being copied (and leaked) by reference.

There's also one more unrelated fix to a crash that I found while debugging the leaks - it turns out `-[NSString stringWithUTF8String:]` can return `nil` for malformed data and we weren't checking for nil.

Check before/after the fix in Instruments, run all existing unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -354,9 +354,13 @@ serializedProfileData(NSDictionary<NSString *, id> *profileData, SentryTransacti
         }
         if (queueAddress != nil && state.queueMetadata[queueAddress] == nil
             && backtrace.queueMetadata.label != nullptr) {
-            state.queueMetadata[queueAddress] = @ {
-                @"label" : [NSString stringWithUTF8String:backtrace.queueMetadata.label->c_str()]
-            };
+            NSString *const labelNSStr = [NSString stringWithUTF8String:backtrace.queueMetadata.label->c_str()];
+            // -[NSString stringWithUTF8String:] can return `nil` for malformed string data
+            if (labelNSStr != nil) {
+                state.queueMetadata[queueAddress] = @ {
+                    @"label" : labelNSStr
+                };
+            }
         }
 #    if defined(DEBUG)
         const auto symbols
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentryProfiler.mm
```diff
@@ -354,9 +354,12 @@ - (void)appendBacktrace:(const Backtrace &)backtrace
         }
         if (queueAddress != nil && state.queueMetadata[queueAddress] == nil
             && backtrace.queueMetadata.label != nullptr) {
-            state.queueMetadata[queueAddress] = @ {
-                @"label" : [NSString stringWithUTF8String:backtrace.queueMetadata.label->c_str()]
-            };
+            NSString *const labelNSStr =
+                [NSString stringWithUTF8String:backtrace.queueMetadata.label->c_str()];
+            // -[NSString stringWithUTF8String:] can return `nil` for malformed string data
```
indragiek, line 359: Here's a repro of this happening:
CleanShot 2023-05-23 at 22 48 47@2x

brustolin (approved):
LGTM

philipphofmann (approved):
Thanks, @indragiek. LGTM.

Sources/Sentry/SentrySamplingProfiler.cpp
```diff
@@ -107,8 +107,8 @@ namespace profiling {
         }
         isSampling_ = true;
         numSamples_ = 0;
-        thread_ = std::thread(samplingThreadMain, port_, clock_, delaySpec_, cache_, callback_,
-            std::ref(numSamples_), onThreadStart);
+        thread_ = std::thread(samplingThreadMain, port_, clock_, delaySpec_, std::cref(cache_),
```
armcknight, line 110: TIL `std::ref` and `std::cref`, nice! I was trying to do this with `&` which wouldn't compile 😭

armcknight (approved):
Nice, thanks @indragie

PR #3062, opened 2023-05-25
ci: Build UI tests with Xcode 14

philipphofmann:
As you can only upload apps to the app store with Xcode 14, there is no 
need to keep still compiling our UI tests with Xcode 13. We can bump to
Xcode 14 and use the macOS 13 GH actions runner, which is faster.

### DIFF
--- a/.github/workflows/saucelabs-UI-tests.yml
+++ b/.github/workflows/saucelabs-UI-tests.yml
@@ -7,7 +7,6 @@ on:
   push:
     branches:
       - main
-      - ci/ui-tests-xcode-14
 
   pull_request:
     paths:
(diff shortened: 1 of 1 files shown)


brustolin (approved):

brustolin:
For more context https://developer.apple.com/news/?id=jd9wcyov

brustolin:
I believe we can remove every job for iOS that is using Xcode 13 and below.

philipphofmann:
> I believe we can remove every job for iOS that is using Xcode 13 and below.

Only iOS 12 is failing, so maybe I need to change it back for that one.

PR #3073, opened 2023-05-30
Fix no changes in high risk files

brustolin:
Updated the script that check for changes in high risk files.

_#skip-changelog_


philipphofmann (approved):
LGTM

PR #3071, opened 2023-05-30
fix: Expose SentrySwiftUI and SentryPrivate schemes for Carthage

RomanTysiachnik:
## :scroll: Description

Fixes issue #2951 and, in addition, exposes SentrySwiftUI, since otherwise there's no way to build it for clients that use carthage.

## :bulb: Motivation and Context

This is the only way to allow `xcodebuild` prepare `SentryPrivate` as a separate xcframework, when the main bundle is build statically, not dynamically

## :green_heart: How did you test it?
- ran locally `carthage bootstrap --no-use-binaries --use-xcframeworks --platform iOS,macOS`

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #3078, opened 2023-05-31
test: add triggers to build/test if a change occurs in the sentry xcode project

armcknight:
I noticed that https://github.com/getsentry/sentry-cocoa/pull/3071 didn't run build or test checks, which probably should've happened. We didn't actually configure those checks to run those on a project config change.

#skip-changelog


brustolin (approved):

armcknight:
Merged main to get the XCFramework check failure fix.

PR #3079, opened 2023-06-05
feat: Symbolicate locally only when debug is enabled

### LINKED ISSUE #3043: Investigate performance of `symbolicate_internal`
### Description

A user reported this at [twitter](https://twitter.com/chrisvasselli/status/1659460805853847553?s=20): 
image

image

We should consider improvements to the symbolicate process.

brustolin:
## :scroll: Description

Only symbolicates stack trace if option.debug is true

closes #3043

Unit tests

## :pencil: Checklist


Sources/Sentry/SentrySDK.m
```diff
@@ -146,6 +148,9 @@ + (void)startWithOptions:(SentryOptions *)options
     [SentrySDK setCurrentHub:[[SentryHub alloc] initWithClient:newClient andScope:scope]];
     SENTRY_LOG_DEBUG(@"SDK initialized! Version: %@", SentryMeta.versionString);
     [SentrySDK installIntegrations];
+
+    [SentryCrashWrapper.sharedInstance startBinaryImageCache];
+    [SentryBinaryImageCache.shared start];
```
philipphofmann, line 153: `m`: I would prefer if the SentryCrashIntegration keeps this responsibility, cause it needs it if I'm not mistaken.

Sources/Sentry/include/SentryBinaryImageCache.h
```diff
@@ -0,0 +1,23 @@
...
+
+@interface BinaryImageInfo : NSObject
+@property (nonatomic, strong) NSString *name;
+@property (nonatomic) uint64_t address;
+@property (nonatomic) uint64_t size;
+@end
+
+@interface SentryBinaryImageCache : NSObject
+
+@property (nonatomic, readonly, class) SentryBinaryImageCache *shared;
```
philipphofmann, line 13: `m`: As there can only be one instance of this, I think we should either make everything static here or add `SENTRY_NO_INIT` here.

Sources/Sentry/include/SentryBinaryImageCache.h
```diff
@@ -0,0 +1,23 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BinaryImageInfo : NSObject
```
philipphofmann, line 5: `h`: To avoid naming collisions.
```suggestion
@interface SentryBinaryImageInfo : NSObject
```

Tests/SentryTests/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,121 @@
+import XCTest
+
+class SentryBinaryImageCacheTests: XCTestCase {
```
philipphofmann, line 3: `m`: Please rename this Swift file to `SentryBinaryImageCacheTests`.

Tests/SentryTests/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,121 @@
...
+
+        sut.binaryImageAdded(&binaryImage3)
+        XCTAssertEqual(sut.cache.count, 2)
+        XCTAssertEqual(sut.cache.last?.name, "Expected Name at 400")
+
+        sut.binaryImageAdded(&binaryImage2)
+        XCTAssertEqual(sut.cache.count, 3)
+        XCTAssertEqual(sut.cache.first?.name, "Expected Name at 100")
+        XCTAssertEqual(sut.cache[1].name, "Expected Name at 200")
+        XCTAssertEqual(sut.cache.last?.name, "Expected Name at 400")
```
philipphofmann, line 36: `m`: Instead of doing manual comparisons, we could simply call `soft` on the cache and compare the lists with something like
```
let sorted = sut.cache.sorted {
    return $0.address < $1.address
}

XCTAssertEqual(sorted, sut.cache)
```

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -0,0 +1,121 @@
...
+}
+
+- (void)stop
+{
+    sentrycrashbic_registerAddedCallback(NULL);
+    sentrycrashbic_registerRemovedCallback(NULL);
+    _cache = nil;
+}
+
+- (void)binaryImageAdded:(const SentryCrashBinaryImage *)image
```
philipphofmann, line 42: Excellent, this is `O(log n)`. Calling sort after adding an image would be `O(n log n)`, so this is way faster 👍

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -0,0 +1,121 @@
...
+
+- (nullable BinaryImageInfo *)imageByAddress:(const uint64_t)address;
+{
+    @synchronized(self) {
+        NSInteger index = [self indexOfImage:address];
+        return index >= 0 ? _cache[index] : nil;
+    }
+}
+
+- (NSInteger)indexOfImage:(NSUInteger)address
```
philipphofmann, line 85: `m`: As we are going to call this for every frame, this is `O(n log m)` where n is the amount of frames and m the amount of images. Ideally, we would improve the search somehow so we can get this close to `O(n)`, but I have to think about how to solve this.

philipphofmann (commented):
Some comments after a first glance. This looks great already 👍

Sources/Sentry/SentryCrashStackEntryMapper.m
```diff
@@ -31,18 +34,26 @@ - (SentryFrame *)sentryCrashStackEntryToSentryFrame:(SentryCrashStackEntry)stack
...
         frame.function = [NSString stringWithCString:stackEntry.symbolName
                                             encoding:NSUTF8StringEncoding];
     }
 
-    if (stackEntry.imageName != NULL) {
-        NSString *imageName = [NSString stringWithCString:stackEntry.imageName
-                                                 encoding:NSUTF8StringEncoding];
-        frame.package = imageName;
-        frame.inApp = @([self.inAppLogic isInApp:imageName]);
+    BinaryImageInfo *info = [_binaryImageCache imageByAddress:stackEntry.address];
```
philipphofmann, line 42: `h`: Can't we use `stackEntry.imageAddress` here? Then we could get the binaryImageInfo from a dictionary which would be `O(1)` instead of doing a binary search `O(log n)`. If not, please add a comment to explain why.

brustolin: `stackEntry.imageAddress` will be 0 if debug is disable, the only thing stackEntry has is `address`.
We also need the image name for the package property. 
But we could check if both informations are already available before looking into the cache.

Sources/Sentry/SentrySDK.m
brustolin: SentryCrashIntegration may be disabled, then symbolication would be doomed.

Sources/Sentry/include/SentryBinaryImageCache.h
brustolin: Good point, I create this as I static struct at first.

Sources/Sentry/SentryCrashStackEntryMapper.m
philipphofmann: Please add a comment for that.

Sources/Sentry/SentryBinaryImageCache.m
philipphofmann: I thought about it and think `O(n log m)` is acceptable. It's as fast as quicksort. That should be fine.

Tests/SentryTests/SentryBinaryImageCache.swift
brustolin: Well, this test is not meant to change anymore, doesn't  matter the search algorithm, so Im fine with this direct approach.

philipphofmann: > this test is not meant to change anymore

I don't really understand that part, but we can also keep it as it is.

Sources/Sentry/SentryCrashStackEntryMapper.m
```diff
@@ -31,18 +34,28 @@ - (SentryFrame *)sentryCrashStackEntryToSentryFrame:(SentryCrashStackEntry)stack
...
         frame.function = [NSString stringWithCString:stackEntry.symbolName
                                             encoding:NSUTF8StringEncoding];
     }
 
-    if (stackEntry.imageName != NULL) {
-        NSString *imageName = [NSString stringWithCString:stackEntry.imageName
-                                                 encoding:NSUTF8StringEncoding];
-        frame.package = imageName;
-        frame.inApp = @([self.inAppLogic isInApp:imageName]);
+    // If there is no symbolicatio, because debug was disabled
```
philipphofmann, line 42:
```suggestion
    // If there is no symbolication, because debug was disabled
```

Tests/SentryTests/SentryBinaryImageCache.swift
brustolin: What I mean is that does not make sense for this test to ever change now that it works, so I would avoid preemptively refactoring 😁

Tests/SentryTests/SentryCrash/SentryBinaryImageCacheTests.m
```diff
@@ -253,6 +255,31 @@ - (void)testCloseCacheWhileAdding
...
+// because is easier to control SentryCrashBinaryImageCache in an objc test
+- (void)testSentryBinaryImageCacheIntegration
+{
+    sentrycrashbic_startCache();
+
+    SentryBinaryImageCache *imageCache = SentryBinaryImageCache.shared;
+    [imageCache start];
+    // by calling start, SentryBinaryImageCache will register a callback with
+    // `SentryCrashBinaryImageCache` that should be called for every image already cached.
+    XCTAssertEqual(5, imageCache.cache.count);
```
philipphofmann, line 269: `m`:  Why is the image cache count precisely 5? Could that change in the future depending on the test setup? If yes, this test could easily break in the future.

Tests/SentryTests/SentryBinaryImageCache+Private.h
```diff
@@ -0,0 +1,13 @@
+#import "SentryBinaryImageCache.h"
+#import "SentryCrashBinaryImageCache.h"
+
+@interface
+SentryBinaryImageCache ()
+
+@property (nonatomic, strong) NSArray<SentryBinaryImageInfo *> *cache;
```
philipphofmann, line 7: `l`: If the class is internal, I think we don't have to add a personal category to make it testable. IMO, you could also just add these methods directly to the class itself and add a comment that they are only required for testing, but we can also keep it if you prefer.

Sources/Sentry/include/SentryBinaryImageCache.h
```diff
@@ -0,0 +1,25 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentryBinaryImageInfo : NSObject
+@property (nonatomic, strong) NSString *name;
+@property (nonatomic) uint64_t address;
+@property (nonatomic) uint64_t size;
+@end
+
+@interface SentryBinaryImageCache : NSObject
```
philipphofmann, line 12: `m`: I think a comment here explaining why we need to sync the binary image cache from the c layer to Objective-C would make sense.

Tests/SentryTests/SentryCrash/SentryBinaryImageCacheTests.m
brustolin: This is a magic number for tests. Im using 5 binary images for deterministic reason.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Experimental support for Swift Async stacktraces (#3051)
 - Cache binary images to be used for crashes (#2939)
+- Local symbolicate for debug only (#3079)
```
philipphofmann, line 9: `m`: Maybe a bit cleaner.
```suggestion
- Symbolicate locally only when debug is enabled (#3079)
```

Tests/SentryTests/SentryCrash/SentryBinaryImageCacheTests.m
brustolin: I start a mock collection (lines 93-94) with 5 items that you be used to test `SentryBinaryImageCache`.
When I start `SentryBinaryImageCache` it should have a copy of those 5 initials items.

This 5 items represent 5 binary images loaded in the app.

Sources/Sentry/include/SentryBinaryImageCache.h
brustolin, line 12:
```suggestion
/**
* This class listen to `SentryCrashBinaryImageCache` to keep a copy of the loaded binaries information in a sorted collection that will be used to symbolicate frames with better performance.
*/
@interface SentryBinaryImageCache : NSObject
```

Sources/Sentry/include/SentryBinaryImageCache.h
```diff
@@ -0,0 +1,30 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentryBinaryImageInfo : NSObject
+@property (nonatomic, strong) NSString *name;
+@property (nonatomic) uint64_t address;
+@property (nonatomic) uint64_t size;
+@end
+
+/**
+ * This class listen to `SentryCrashBinaryImageCache` to keep a copy of the loaded binaries
```
philipphofmann, line 13: Thanks 👍

philipphofmann, line 13:
```suggestion
 * This class listens to `SentryCrashBinaryImageCache` to keep a copy of the loaded binaries
```

philipphofmann (approved):
LGTM, but we should ship a beta for this to reduce the risk of a big bang 😄.

PR #3089, opened 2023-06-07
fix: SentryUIViewControllerSwizzlingTests

brustolin:
Fixed SentryUIViewControllerSwizzlingTests because it broke with latest swift/xcode updates.

_#skip-changelog_

### DIFF
--- a/Sources/Sentry/SentryNSProcessInfoWrapper.mm
+++ b/Sources/Sentry/SentryNSProcessInfoWrapper.mm
@@ -1,6 +1,26 @@
 #import "SentryNSProcessInfoWrapper.h"
 
-@implementation SentryNSProcessInfoWrapper
+@implementation SentryNSProcessInfoWrapper {
+#if TEST
+    NSString *_executablePath;
+}
+- (void)setProcessPath:(NSString *)path
+{
+    _executablePath = path;
+}
+#    define EXECUTABLE_PATH _executablePath;
+
+- (instancetype)init
+{
+    self = [super init];
+    _executablePath = NSBundle.mainBundle.bundlePath;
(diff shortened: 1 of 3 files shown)


markushi (approved):
LGTM!

Sources/Sentry/SentryNSProcessInfoWrapper.mm
```diff
@@ -1,6 +1,26 @@
...
-@implementation SentryNSProcessInfoWrapper
+@implementation SentryNSProcessInfoWrapper {
+#if TEST
+    NSString *_executablePath;
+}
+- (void)setProcessPath:(NSString *)path
+{
+    _executablePath = path;
+}
+#    define EXECUTABLE_PATH _executablePath;
```
armcknight, line 11: Please always prefix macro names with `SENTRY_`. While this instance won't break compilation, this name collides with a build setting, so it's easy to imagine it causing confusion later: 
image

Sources/Sentry/SentryNSProcessInfoWrapper.mm
```diff
@@ -1,6 +1,26 @@
...
+{
+    self = [super init];
+    _executablePath = NSBundle.mainBundle.bundlePath;
+    return self;
+}
+
+#else
+}
+#    define EXECUTABLE_PATH NSBundle.mainBundle.executablePath;
+#endif
```
armcknight, line 23: We should just avoid breaking up syntactically complete constructs with `#if`/`#else`. We only recently had a discussion about how complex it makes things here: https://github.com/getsentry/sentry-cocoa/pull/2858#discussion_r1159415510

Issue #3082, opened 2023-06-06
App crash in the `sliceGPUData`


anton-plebanovich:
### Platform

iOS

### Installed

CocoaPods

### Version

8.6.0

### Steps to Reproduce

The actual steps are unknown but I hope the crash stack trace should be enough. We started to receive a lot of those after we updated from the `8.1.0` version to the `8.6.0` version. It looks like `nearestPredecessorValue` might be `nil` in some cases when an object for `slicedGPUEntries` is created in the `sliceGPUData` function.

```
OS Version: iOS 16.5 (20F66)
Report Version: 104

Exception Type: EXC_CRASH (SIGABRT)
Crashed Thread: 0

Application Specific Information:
*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[1]

Thread 0 Crashed:
0   CoreFoundation                  0x32f1f3cb4         __exceptionPreprocess
1   libobjc.A.dylib                 0x32131c3cc         objc_exception_throw
2   CoreFoundation                  0x32f399f80         _CFThrowFormattedException
3   CoreFoundation                  0x32f3a54a8         -[__NSPlaceholderDictionary initWithObjects:forKeys:count:].cold.5
4   CoreFoundation                  0x32f20f458         -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]
5   CoreFoundation                  0x32f20f2f8         +[NSDictionary dictionaryWithObjects:forKeys:count:]
6   Sentry                          0x10473ab2c         sliceGPUData (SentryProfiler.mm:228)
7   Sentry                          0x10473c5a4         +[SentryProfiler createProfilingEnvelopeItemForTransaction:] (SentryProfiler.mm:417)
8   Sentry                          0x104756918         -[SentryTracer captureTransactionWithProfile:] (SentryTracer.m:511)
9   Sentry                          0x1047566ec         -[SentryTracer finishInternal] (SentryTracer.m:499)
10  Sentry                          0x10475502c         -[SentryTracer deadlineTimerFired] (SentryTracer.m:238)
11  Sentry                          0x104754e10         __34-[SentryTracer startDeadlineTimer]_block_invoke (SentryTracer.m:217)
12  Foundation                      0x3237e119c         __NSFireTimer
13  CoreFoundation                  0x32f2b5620         __CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__
[...]

PR #3092, opened 2023-06-12
fix: avoid inserting a nil into the gpu data dictionary literal

armcknight:
## :scroll: Description

Stops the crash from occurring that was reported in #3082 

## :pencil: Checklist

~- [ ] I added tests to verify the changes.~

### DIFF
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -143,7 +143,7 @@ sliceGPUData(SentryFrameInfoTimeSeries *frameInfo, SentryTransaction *transactio
             @"value" : obj[@"value"],
         }];
     }];
-    if (useMostRecentRecording && slicedGPUEntries.count == 0) {
+    if (useMostRecentRecording && slicedGPUEntries.count == 0 && nearestPredecessorValue != nil) {
         [slicedGPUEntries addObject:@ {
             @"elapsed_since_start_ns" : @"0",
             @"value" : nearestPredecessorValue,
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentryProfiler.mm
```diff
@@ -143,7 +143,7 @@
             @"value" : obj[@"value"],
         }];
     }];
-    if (useMostRecentRecording && slicedGPUEntries.count == 0) {
+    if (useMostRecentRecording && slicedGPUEntries.count == 0 && nearestPredecessorValue != nil) {
```
brustolin, line 146: m: Does it make sense to have a test to avoid regression?

brustolin (commented):
It looks good, but I have one question below.

Sources/Sentry/SentryProfiler.mm
armcknight: I don't think so for several reasons: 
1. the only way this will regress is if someone removes the nil check
2. the testing is disproportionately difficult to implement
3. i don't even have a repro on this yet
4. i'm working on a more holistic solution in #3090

armcknight:
I updated this PR to _only_ perform the nil check, without the refactor. I will make that other change in a separate PR.

Sources/Sentry/SentryProfiler.mm
indragiek: +1 to fixing the immediate crash right now and working on the correct fix + tests in the other PR @armcknight linked

brustolin (approved):

PR #3096, opened 2023-06-14
fix: Rename EXECUTABLE_PATH

brustolin:
Renaming a definition that may collide with build params.

_#skip-changelog_


philipphofmann (approved):
LGTM.

PR #3108, opened 2023-06-16
fix: report instances of dropped profiles

armcknight:
We'd like to know when we discard profiles due to insufficient amounts of samples so we can assist customers in the frontend.

#skip-changelog


Sources/Sentry/SentryProfiler.mm
```diff
@@ -186,21 +186,26 @@
...
+    SentryHub *hub)
 {
     NSMutableArray<SentrySample *> *const samples = profileData[@"profile"][@"samples"];
     // We need at least two samples to be able to draw a stack frame for any given function: one
     // sample for the start of the frame and another for the end. Otherwise we would only have a
     // stack frame with 0 duration, which wouldn't make sense.
     if ([samples count] < 2) {
         SENTRY_LOG_DEBUG(@"Not enough samples in profile");
+        [hub.getClient recordLostEvent:kSentryDataCategoryProfile
+                                reason:kSentryDiscardReasonEventProcessor];
```
Zylphrex, line 199: Thoughts on introducing a separate reason for this use case? Perhaps something like `"insufficient_samples"` or if that's too specific since the other reasons are more generic we could do something like `"failed_validation"`

armcknight: I did consider it, but was put off by the scope of work, since it'd require changes to snuba. I think this is ok for now because this is the only combination of profile category event processor reason, if we hit trouble with it in the future getting overloaded let's change it then.

armcknight: It'll also at least require changes to these docs: https://develop.sentry.dev/sdk/client-reports/#envelope-item-payload

armcknight: What should we call the new reason? `insufficient_data` is my suggestion

armcknight: https://github.com/getsentry/snuba/pull/4407

brustolin (approved):
LGTM

PR #3119, opened 2023-06-23
ref: centralize frames tracker access in SentryDependencyContainer

armcknight:
In service of reworking SentryProfiler's management of GPU frame data in https://github.com/getsentry/sentry-cocoa/pull/3090, I need to simplify how it keeps a reference to the frames tracker. 

This has bugged me for some time because we vend a singleton of the frames tracker, but then various other objects keep and pass references to it, which is unnecessary; it can just be accessed via SentryDependencyContainer. This seemed like a good opportunity to fix it everywhere.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
+++ b/Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
@@ -17,7 +17,6 @@ class SentryTimeToDisplayTrackerTest: XCTestCase {
         init() {
             framesTracker = SentryFramesTracker(displayLinkWrapper: displayLinkWrapper)
             SentryDependencyContainer.sharedInstance().framesTracker = framesTracker
-            print("Fixture.init.SentryDependencyContainer.sharedInstance().framesTracker = \(SentryDependencyContainer.sharedInstance().framesTracker)")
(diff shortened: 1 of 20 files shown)


Tests/SentryTests/Integrations/Performance/UIViewController/SentryTimeToDisplayTrackerTest.swift
```diff
@@ -16,15 +16,17 @@ class SentryTimeToDisplayTrackerTest: XCTestCase {
...
         }
 
         func getSut(for controller: UIViewController, waitForFullDisplay: Bool) -> SentryTimeToDisplayTracker {
-            return SentryTimeToDisplayTracker(for: controller, framesTracker: framesTracker, waitForFullDisplay: waitForFullDisplay)
+            return SentryTimeToDisplayTracker(for: controller, waitForFullDisplay: waitForFullDisplay)
         }
     }
 
-    private let fixture = Fixture()
+    private lazy var fixture = Fixture()
```
armcknight, line 29: It turns out that with a constant instance property, it was initialized as part of a pre-test process, and then when the tests ran, the dependency container shared instance was constructed again and had a difference instance of the frames tracker, so the tests would fail. Making it lazy defers initialization until the test process itself is running.

brustolin (approved):
LGTM

PR #3120, opened 2023-06-24
ref: rename SentryTimerWrapper to SentryTimerFactory 

armcknight:
since it's no longer a subclass after the recent fixes, it's now a factory. Noticed as I made the change in https://github.com/getsentry/sentry-cocoa/pull/3119

#skip-changelog


brustolin (approved):
LGTM

PR #3121, opened 2023-06-24
ref: use SentryDependencyContainer instead of static methods/vars

armcknight:
Move remaining global wrapper variables from SentryProfiler into SentryDependencyContainer.

This simplifies the interface and testing, and helps avoid issues we had in the past with forgetting to initialize wrappers. This will simplify some of the refactoring necessary to help fix https://github.com/getsentry/sentry-cocoa/pull/3090

#skip-changelog


brustolin (approved):
LGTM

PR #3124, opened 2023-06-27
ref: rename SentryProfiler+Private.h to SentryProfilerState.h

armcknight:
This has been throwing me off when trying to sort out some header imports. I don't think it should be called SentryProfiler+Private because it doesn't contain any interface extensions of `SentryProfiler`, but rather introduces only the new classes `SentryProfilerState` (and it's mutable counterpart).

#skip-changelog


indragiek (approved):

armcknight:
The lone test failure is either a flake or a very rare manifestation of the race condition being explored for #3090; in either case it doesn't have bearing on the changes in this PR so I'm not going to bother rerunning.

PR #3133, opened 2023-07-07
ref: profile sample mocking

armcknight:
This is the last component of a long-standing desire to mock components of the profiler. It extracts the objc++ code from SentryProfilerTests.mm into an interface that can be used from Swift tests as well. This will help write finer grained tests to validate an upcoming fix, as well as speed up the profiler tests.

#skip-changelog


Sources/Sentry/include/SentryProfiler+Test.h
brustolin, line 1: h: Does it make sense to have this test file in this folder?

brustolin (commented):
All good but one file that seems to be in the wrong place.

Tests/SentryProfilerTests/SentryProfilerTests.mm
```diff
@@ -292,49 +237,16 @@ - (void)testProfilerPayload
...
-
-    const auto addresses2 = std::vector<std::uintptr_t>({ 0x777, 0x888, 0x789 });
-
-    Backtrace backtrace2;
-    backtrace2.threadMetadata = threadMetadata2;
-    backtrace2.queueMetadata = queueMetadata2;
-    backtrace2.absoluteTimestamp = 5;
-    backtrace2.addresses = addresses2;
-
+    const auto backtrace2 = mockBacktrace(12345568910, 666, "testThread", 9876543210, "testQueue",
```
philipphofmann, line 245: Way cleaner 💯

SentryTestUtils/SentryProfilerMocksSwiftCompatible.h
```diff
@@ -0,0 +1,28 @@
...
+
+@class SentryProfilerState;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * This delivers a wrapper around the C++ function to create a mock backtrace for incorporation into
+ * profiler state that can be called from Swift tests.
+ */
+@interface SentryProfilerMocksSwiftCompatible : NSObject
```
philipphofmann, line 14: `l`: Mocks are usually test doubles that simulate the behavior of a real object with the added ability to verify interactions with that object. They are used to set expectations and assert that specific methods or properties of the mocked object are called or accessed correctly during testing. What you define here is a fake. Fakes are simplified implementations of a real object that mimics its behavior to some extent. They are typically used to provide a functional substitute for a real object that is too complex, slow, or unavailable for testing.

TLDR; if you want to follow the correct testing naming terms, we should rename all the mocks in this PR to fake, but I'm also fine with keeping it. I don't want to be nitpicking on testing terms.

philipphofmann (commented):
This ref improves the testability of the profiling code 💪 . Excellent. I'm only not approving cause of Dhiogos comment https://github.com/getsentry/sentry-cocoa/pull/3133#discussion_r1255341626.

SentryTestUtils/SentryProfilerMocksSwiftCompatible.h
armcknight: Sorry but I think this is more of a mock. I think we should be careful introducing too much terminology (and therefore cognitive overhead) like fixture, double, oracle, stub, mock, fake, dummy, spy, robot... I don't think these are all standardized terms in industry. 

Since this is just creating structs to use as input to a real implementation, and not actually providing an alternate implementation of an interface, there is really no doubling going on at all, so in an OOP sense, this is neither mocking nor stubbing (nor faking).

I think the intention of the code gets across so I'm hesitant to change it. We're providing predefined input to the SUT and testing the output of the system.

Sources/Sentry/include/SentryProfiler+Test.h
armcknight: I split renamed this one to SentryProfiler+Private because some of the declarations are needed in the SDK, and moved the actual test helper to a new one SentryProfiler+Test in the Tests/ directory.

SentryTestUtils/SentryProfilerMocksSwiftCompatible.h
philipphofmann: Fine by me 👍 .

philipphofmann (approved):
LGTM

PR #3135, opened 2023-07-07
fix: profiler timeout scheduling and data preservation

armcknight:
## :scroll: Description

Fixes an unbounded memory growth issue if a transaction, and therefore the profiler, is started from a non-main context, where the timeout timer needs to be scheduled on the main thread.

Making that small change led me to finding the design flaw that was causing data loss (originally a crash) in https://github.com/getsentry/sentry-cocoa/issues/3082. Because we didn't keep instances of profilers that had timed out at 30s, while their associated transactions would go on for up to 430 more seconds until timing out, those transactions could query for data that had been lost. I think there was a race at the moment a profiler was queried for its data and when another one was started, thus losing the data being queried as it was being processed.

So, `SentryTracerConcurrency` and `SentryProfiler` are refactored here to remove the assumption that only one profiler instance would ever need to be kept in `SentryProfiler->_gCurrentProfiler`, and instead hold references to all outstanding profilers in `SentryTracerConcurrency._gProfilersToTracers`.

Customer reported issue.

Reproduced it manually, by starting many txs on a non-main queue and never stopping them, so that the profiler should time out. Validated that the dispatch to schedule the timeout timer fixed the issue.

Wrote a regression test in #3145 (the new base for this PR) so we can see it fail there and pass here. It's not a new test, but rather, fixes a preexisting test that didn't actually function correctly until gaining the capability to mock backtraces in #3133.
- failure in #3145: https://github.com/getsentry/sentry-cocoa/actions/runs/5537602499/jobs/10106610925?pr=3145#step:12:95
- passes here: https://github.com/getsentry/sentry-cocoa/actions/runs/5537612410/jobs/10106674935?pr=3135

## :pencil: Checklist


Sources/Sentry/SentryProfiler.mm
```diff
@@ -155,7 +155,12 @@
 serializedProfileData(NSDictionary<NSString *, id> *profileData, SentryTransaction *transaction,
     SentryId *profileID, NSString *truncationReason, NSString *environment, NSString *release,
     NSDictionary<NSString *, id> *serializedMetrics, NSArray<SentryDebugMeta *> *debugMeta,
-    SentryHub *hub)
+    SentryHub *hub
+#    if SENTRY_HAS_UIKIT
+    ,
+    SentryScreenFrames *gpuData
+#    endif // SENTRY_HAS_UIKIT
+)
```
armcknight, line 163: This is ugly, I know. I have a future refactor planned to move this entire function into `SentryProfilerState`.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -231,28 +236,25 @@
     auto metrics = serializedMetrics;
 
 #    if SENTRY_HAS_UIKIT
-    const auto framesTracker = SentryDependencyContainer.sharedInstance.framesTracker;
```
armcknight, line 234: In this hunk, we now use a private copy of the data in  `SentryDependencyContainer.sharedInstance.framesTracker.currentFrames` that is copied/stored per profiler instance. That way, when the profiler running for the last in-flight transaction is stopped, we can reset the `SentryFramesTracker`'s version of the data, while keeping the copies in instances of profilers that may still be waiting in memory for their transactions to finish.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -306,126 +349,97 @@ + (void)startWithHub:(SentryHub *)hub
...
-                                target:self
-                              selector:@selector(timeoutAbort)
-                              userInfo:nil
-                               repeats:NO];
-#    if SENTRY_HAS_UIKIT
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(backgroundAbort)
-                                                 name:UIApplicationWillResignActiveNotification
-                                               object:nil];
-#    endif // SENTRY_HAS_UIKIT
```
armcknight, line 326: Lines 313-326 were moved from this class method into the (preexisting) instance method `-[SentryProfiler initWithHub:]` a bit above this in the diff..

Sources/Sentry/SentryProfiler.mm
```diff
@@ -283,20 +285,61 @@ - (instancetype)initWithHub:(SentryHub *)hub
...
+    [self start];
+    [self scheduleTimeoutTimer];
+
+#    if SENTRY_HAS_UIKIT
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(backgroundAbort)
+                                                 name:UIApplicationWillResignActiveNotification
+                                               object:nil];
+#    endif // SENTRY_HAS_UIKIT
+
```
armcknight, line 302: Lines 293-301 were originally in the class method `+ [SentryProfiler startWithHub:]`.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -306,126 +349,97 @@ + (void)startWithHub:(SentryHub *)hub
         return;
     }
 
-#    if SENTRY_HAS_UIKIT
-    [SentryDependencyContainer.sharedInstance.framesTracker resetProfilingTimestamps];
-#    endif // SENTRY_HAS_UIKIT
```
armcknight, line 311: This is removed from here to SentryTracerConcurrency, where we can know how many profilers are being held waiting for their associated tracers to finish. Only when there are no profilers left do we reset this data.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -306,126 +349,97 @@ + (void)startWithHub:(SentryHub *)hub
...
-    if (!_gCurrentProfiler) {
-        SENTRY_LOG_WARN(@"No current global profiler manager to stop.");
-        return;
-    }
-    if (![_gCurrentProfiler isRunning]) {
-        SENTRY_LOG_WARN(@"Current profiler is not running.");
-        return;
-    }
-
-    [self stopProfilerForReason:SentryProfilerTruncationReasonNormal];
```
armcknight, line 342: The class method `+[SentryProfiler stop]` is no longer needed. `SentryTracerConcurrency` directly calls the instance method `-[SentryProfiler stop]` when its last associated transaction finishes. It was confusing having both, I think this this is better!

Sources/Sentry/SentryProfiler.mm
```diff
@@ -306,126 +349,97 @@ + (void)startWithHub:(SentryHub *)hub
...
+    trackProfilerForTracer(_gCurrentProfiler, tracer);
```
armcknight, line 352: This doesn't diff well if you're viewing a unified diff (looks better in a split diff), but it's the last line of the `+[SentryProfiler startWithHub:tracer:]` method (starting on line 336).

Sources/Sentry/SentryProfiler.mm
```diff
@@ -306,126 +349,97 @@ + (void)startWithHub:(SentryHub *)hub
...
-+ (BOOL)isRunning
++ (BOOL)isCurrentlyProfiling
 {
     std::lock_guard<std::mutex> l(_gProfilerLock);
     return [_gCurrentProfiler isRunning];
 }
 
-+ (SentryEnvelopeItem *)createProfilingEnvelopeItemForTransaction:(SentryTransaction *)transaction
++ (nullable SentryEnvelopeItem *)createProfilingEnvelopeItemForTransaction:
+    (SentryTransaction *)transaction
```
armcknight, line 362: This was already declared nullable return type in the header, I just added it here too.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -306,126 +349,97 @@ + (void)startWithHub:(SentryHub *)hub
...
-        SENTRY_LOG_WARN(@"Current profiler is not running.");
-        return;
-    }
-
-    [self stopProfilerForReason:SentryProfilerTruncationReasonNormal];
+    trackProfilerForTracer(_gCurrentProfiler, tracer);
 }
 
-+ (BOOL)isRunning
++ (BOOL)isCurrentlyProfiling
```
armcknight, line 355: I renamed this because it kinda sounded like an instance method, but it's not.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -306,126 +349,97 @@ + (void)startWithHub:(SentryHub *)hub
...
 {
     std::lock_guard<std::mutex> l(_gProfilerLock);
     return [_gCurrentProfiler isRunning];
 }
 
-+ (SentryEnvelopeItem *)createProfilingEnvelopeItemForTransaction:(SentryTransaction *)transaction
++ (nullable SentryEnvelopeItem *)createProfilingEnvelopeItemForTransaction:
+    (SentryTransaction *)transaction
 {
-    const auto profileID = [[SentryId alloc] init];
```
armcknight, line 353: We're now storing a profile ID with each profiler instance right when we init them, so we can track the ID with tracers in `SentryTracerConcurrency`.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -306,126 +349,97 @@ + (void)startWithHub:(SentryHub *)hub
...
-    return [self envelopeItemForProfileData:payload profileID:profileID];
+    const auto JSONData = [SentrySerialization dataWithJSONObject:payload];
+    if (JSONData == nil) {
+        SENTRY_LOG_DEBUG(@"Failed to encode profile to JSON.");
+        return nil;
+    }
+
+    const auto header = [[SentryEnvelopeItemHeader alloc] initWithType:SentryEnvelopeItemTypeProfile
+                                                                length:JSONData.length];
+    return [[SentryEnvelopeItem alloc] initWithHeader:header data:JSONData];
```
armcknight, line 387: This was the only place this method was called, so I just inlined the implementation here.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -306,126 +349,97 @@ + (void)startWithHub:(SentryHub *)hub
...
+    [_metricProfiler stop];
+    _truncationReason = reason;
+
+    if (![self isRunning]) {
+        SENTRY_LOG_WARN(@"Profiler is not currently running.");
+        return;
+    }
+
+    _profiler->stopSampling();
+    SENTRY_LOG_DEBUG(@"Stopped profiler %@.", self);
```
armcknight, line 442: Converted the class methods to instance methods, which simplifies their implementations a bit: no more locking or nil checks required: `+[timeoutAbort]`, `+[backgroundAbort]` and `+[stopProfilerForReason]`.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -490,39 +504,10 @@ - (void)start
...
-    const auto JSONData = [SentrySerialization dataWithJSONObject:profile];
-    if (JSONData == nil) {
-        SENTRY_LOG_DEBUG(@"Failed to encode profile to JSON.");
-        return nil;
-    }
-
-    const auto header = [[SentryEnvelopeItemHeader alloc] initWithType:SentryEnvelopeItemTypeProfile
-                                                                length:JSONData.length];
-    return [[SentryEnvelopeItem alloc] initWithHeader:header data:JSONData];
-}
```
armcknight, line 521: `+[envelopeItemForProfileData:profileID:]` was inlined at the sole callsite on new lines 379-387

Sources/Sentry/SentryProfiler.mm
```diff
@@ -490,39 +504,10 @@ - (void)start
...
-    }
-    if (!_profiler->isSampling()) {
-        SENTRY_LOG_WARN(@"Profiler is not currently sampling.");
-        return;
-    }
-
-    _profiler->stopSampling();
-    [_metricProfiler stop];
-    SENTRY_LOG_DEBUG(@"Stopped profiler %@.", self);
-}
```
armcknight, line 507: The preexisting instance method `-[stop]` was combined with the eliminated class method `+[stopProfilerForReason:]` to create the only instance method now needed to stop a profiler: `-[stopForReason:]` on new lines 432-442.

Sources/Sentry/SentryTracer.m
```diff
@@ -151,8 +151,7 @@ - (instancetype)initWithTransactionContext:(SentryTransactionContext *)transacti
     if (_configuration.profilesSamplerDecision.decision == kSentrySampleDecisionYes) {
         _isProfiling = YES;
         _startSystemTime = SentryCurrentDate.systemTime;
-        [SentryProfiler startWithHub:hub];
-        trackTracerWithID(self.traceId);
```
armcknight, line 155: We now just call the bookkeeping function `trackTracerWithID` from `SentryProfiler`'s implementation (the function was renamed to `trackProfilerForTracer(SentryProfiler *profiler, SentryTracer *tracer)`

Sources/Sentry/SentryTracer.m
```diff
@@ -515,13 +514,12 @@ - (void)captureTransactionWithProfile:(SentryTransaction *)transaction
 {
     SentryEnvelopeItem *profileEnvelopeItem =
         [SentryProfiler createProfilingEnvelopeItemForTransaction:transaction];
+
     if (!profileEnvelopeItem) {
         [_hub captureTransaction:transaction withScope:_hub.scope];
         return;
     }
 
-    stopTrackingTracerWithID(self.traceId, ^{ [SentryProfiler stop]; });
```
armcknight, line 523: This is now handled automatically from `SentryTracerConcurrency.profilerForFinishedTracer` (which is only called from `+[SentryProfiler createProfilingEnvelopeItemForTransaction:]` called on line 517 above).

Sources/Sentry/SentryProfiler.mm
```diff
@@ -283,20 +285,61 @@ - (instancetype)initWithHub:(SentryHub *)hub
...
+#    endif // SENTRY_HAS_UIKIT
+
     return self;
 }
 
+/**
+ * Schedule a timeout timer on the main thread.
+ * @warning from NSTimer.h: Timers scheduled in an async context may never fire.
+ */
+- (void)scheduleTimeoutTimer
```
armcknight, line 310: This is one of the cruxes of the change. By not ensuring we were scheduling the timeout timer from the main thread, we created a situation where the profiler may never stop, leading to unbounded memory growth.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -312,12 +312,9 @@ + (void)startWithHub:(SentryHub *)hub
...
     [_gCurrentProfiler start];
 
-    _gCurrentProfiler->_timeoutTimer = [SentryDependencyContainer.sharedInstance.timerFactory
-        scheduledTimerWithTimeInterval:kSentryProfilerTimeoutInterval
-                                target:self
-                              selector:@selector(timeoutAbort)
-                              userInfo:nil
-                               repeats:NO];
+    // from NSTimer.h: Timers scheduled in an async context may never fire.
+    dispatch_async(dispatch_get_main_queue(), ^{ [self scheduleTimeoutTimer]; });
```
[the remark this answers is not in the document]
armcknight: Looks like it, yeah! See the updated implementation where I check `+[NSThread isMainThread]`, because if you dispatch_async to the main queue when you're already on it, and the caller blocks the main queue, it'll never even get scheduled!

philipphofmann (commented):
First pass, I still need to have a close look at `SentryTracerConcurrency.m`.

Sources/Sentry/SentryScreenFrames.m
```diff
@@ -43,6 +43,16 @@ - (instancetype)initWithTotal:(NSUInteger)total
 
     return self;
 }
+
+- (id)copy
```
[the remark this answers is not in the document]
armcknight: Hmm yes, I did try this, and Xcode autogenerated a bunch of weird stubs for conformance. Let me have another look at it.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -283,20 +285,61 @@ - (instancetype)initWithHub:(SentryHub *)hub
...
+    _profileId = [[SentryId alloc] init];
+
     SENTRY_LOG_DEBUG(@"Initialized new SentryProfiler %@", self);
     _debugImageProvider = [SentryDependencyContainer sharedInstance].debugImageProvider;
     _hub = hub;
+    [self start];
+    [self scheduleTimeoutTimer];
+
+#    if SENTRY_HAS_UIKIT
+    [[NSNotificationCenter defaultCenter] addObserver:self
```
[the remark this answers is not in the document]
armcknight: good catch! this has been like this a while. will do.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -283,20 +285,61 @@ - (instancetype)initWithHub:(SentryHub *)hub
...
+                                  selector:@selector(timeoutAbort)
+                                  userInfo:nil
+                                   repeats:NO];
+    };
+
+    if (NSThread.isMainThread) {
+        block();
+    } else {
+        dispatch_async(dispatch_get_main_queue(), block);
+    }
```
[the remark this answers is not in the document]
armcknight: did this in a separate PR: https://github.com/getsentry/sentry-cocoa/pull/3138

Sources/Sentry/SentryProfiler.mm
indragiek: Any reason why we wouldn't use a dispatch timer that lets you explicitly specify which queue it fires on?

Sources/Sentry/SentryTracerConcurrency.mm
```diff
@@ -1,49 +1,123 @@
...
+// we must keep the aborted one around until its associated transaction finishes.
+typedef NSMutableDictionary</* SentryProfiler.profileId */ NSString *,
+    NSMutableSet<SentryTracer *> *>
+    SentryProfilerTracerMapping;
+static SentryProfilerTracerMapping *_gProfilersToTracers;
+
+// provided for fast access to a profiler given a tracer
+typedef NSMutableDictionary</* SentryTracer.tracerId */ NSString *, SentryProfiler *>
+    SentryTracerProfilerMapping;
+static SentryTracerProfilerMapping *_gTracersToProfilers;
```
indragiek, line 30: nit: matter of opinion but I find this pattern of typedef'ing generic types makes code less readable at the site where the typedef is used, where it is not clear that it's just an alias for a specialized dictionary type. Even though repeating it is verbose, it is clearer to read.

indragiek (approved):
lgtm, just a question

Sources/Sentry/SentryTracerConcurrency.mm
philipphofmann: I agree with @indragiek.

Sources/Sentry/SentryTracerConcurrency.mm
```diff
@@ -1,49 +1,123 @@
...
+    } else {
+        [_gProfilersToTracers[profilerKey] addObject:tracer];
+    }
+
+    _gTracersToProfilers[tracerKey] = profiler;
 }
 
-void
-stopTrackingTracerWithID(SentryId *traceID, SentryConcurrentTransactionCleanupBlock cleanup)
+SentryProfiler *_Nullable profilerForTracer(SentryTracer *tracer)
```
philipphofmann, line 67: `m`: If I'm not mistaken we can remove this method, as it's not used anywhere.

Sources/Sentry/SentryTracerConcurrency.mm
```diff
@@ -1,49 +1,123 @@
...
-    const auto idString = traceID.sentryIdString;
-    SENTRY_LOG_DEBUG(@"Removing trace id %@", idString);
-    [_gInFlightTraceIDs removeObject:idString];
-    if (_gInFlightTraceIDs.count == 0) {
-        SENTRY_LOG_DEBUG(@"Last in flight tracer completed, performing cleanup.");
-        cleanup();
-    } else {
-        SENTRY_LOG_DEBUG(@"Waiting on %lu other tracers to complete: %@.", _gInFlightTraceIDs.count,
-            _gInFlightTraceIDs);
+SentryProfiler *_Nullable profilerForFinishedTracer(SentryTracer *tracer)
```
philipphofmann, line 75: `h`: This method only gets called in `SentryProfiler.createProfilingEnvelopeItemForTransaction`, which is called by, `SentryTracer.captureTransactionWithProfile`, which is called by `SentryTracer.finishInternal`. There is no guarantee that `SentryTracer.finishInternal` will actually call `SentryTracer.captureTransactionWithProfile`. Therefore, it can happen that `_gProfilersToTracers` and `_gTracersToProfilers` keep references to `SentryTracer` and `SentryProfile` which should have been deallocated. This can lead to infinite memory growth, or am I missing something? We could solve this by keeping weak references here or we need some other way of cleaning up the two dictionaries. I think this problem already existed before, so we can also address in an extra PR.

philipphofmann (commented):
I maybe found one important memory growth issue, apart from that, LGTM.

Sources/Sentry/SentryProfiler.mm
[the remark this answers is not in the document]
armcknight: Done in https://github.com/getsentry/sentry-cocoa/pull/3138

Sources/Sentry/SentryTracerConcurrency.mm
armcknight: We can't use weak refs because there won't be anything else keeping valid profilers alive. The simplest solution I can think of is to call some cleanup function that removes them from the dicts at every early return in the codepath that would otherwise lead to capturing a transaction. 

Are there any early returns in `finishInternal` that _shouldn't_ do such cleanup, because it will be called again later?

And conversely, are there any callers to `finishInternal` that have early returns where we'd also need to do this cleanup because there won't be another call to `finishInternal`?

Happy to tackle this in a separate PR as stated.

Sources/Sentry/SentryProfiler.mm
armcknight: There was no specific reason we didn't use a dispatch timer AFAIK. Do you think we should change it in this PR?

indragiek: It can be addressed separately 👍🏽

Sources/Sentry/SentryTracerConcurrency.mm
armcknight: I did this in two change: https://github.com/getsentry/sentry-cocoa/pull/3154 to clean up profilers for discarded transactions, and weak references in https://github.com/getsentry/sentry-cocoa/pull/3155

PR #3139, opened 2023-07-11
ref: asserts and warnings

armcknight:
Add an assert/warning to timer factory if attempting to schedule a timer from a non-main context. Also add more macros that don't need to return the result of the condition if only a warning log is needed. Use in a couple other places.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryNSTimerFactory.m
+++ b/Sources/Sentry/SentryNSTimerFactory.m
@@ -8,7 +8,8 @@
                                       block:(void (^)(NSTimer *timer))block
 {
     SENTRY_ASSERT([NSThread isMainThread],
-        @"Timers must be scheduled from the main thread, or they may never fire.");
+        @"Timers must be scheduled from the main thread, or they may never fire. See the attribute "
+        @"on the declaration in NSTimer.h.");
     return [NSTimer scheduledTimerWithTimeInterval:interval repeats:repeats block:block];
 }
 
@@ -19,7 +20,8 @@
                                     repeats:(BOOL)yesOrNo
 {
     SENTRY_ASSERT([NSThread isMainThread],
-        @"Timers must be scheduled from the main thread, or they may never fire.");
+        @"Timers must be scheduled from the main thread, or they may never fire. See the attribute "
+        @"on the declaration in NSTimer.h.");
     return [NSTimer scheduledTimerWithTimeInterval:ti
                                             target:aTarget
                                           selector:aSelector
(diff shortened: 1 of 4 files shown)


Sources/Sentry/SentryNSTimerFactory.m
```diff
@@ -1,11 +1,14 @@
...
+#import "SentryInternalDefines.h"
 
 @implementation SentryNSTimerFactory
 
 - (NSTimer *)scheduledTimerWithTimeInterval:(NSTimeInterval)interval
                                     repeats:(BOOL)repeats
                                       block:(void (^)(NSTimer *timer))block
 {
+    SENTRY_ASSERT([NSThread isMainThread],
+        @"Timers must be scheduled from the main thread, or they may never fire.");
```
brustolin, line 11: Im curious where this is coming from? Never read anything about this.

philipphofmann: @armcknight, I think you once linked some Apple docs for that. Please add this as a comment.

philipphofmann (approved):
LGTM.

Sources/Sentry/SentryNSTimerFactory.m
armcknight: After looking for where I saw it, I learned a little bit more. It has to do with run loops and their modes. 

The original thing I saw that tipped me off to this is only found in an attribute on the declarations in NSTimer.h, not in any other docs: `+ (NSTimer *)scheduledTimerWithTimeInterval:(NSTimeInterval)ti target:(id)aTarget selector:(SEL)aSelector userInfo:(nullable id)userInfo repeats:(BOOL)yesOrNo NS_SWIFT_UNAVAILABLE_FROM_ASYNC("Timers scheduled in an async context may never fire.");`

I found a couple things talking about the run loop modes, which is most easily seen in this SO answer: https://stackoverflow.com/a/12173900. They link to this blog post: http://bynomial.com/blog/?p=67. I also found this one: https://danielemargutti.medium.com/the-secret-world-of-nstimer-708f508c9eb.

How should I word this? Something like: `According to the declaration in NSTimer.h, scheduled timers cannot be added from async contexts. They should only be scheduled from the main run loop.`?

armcknight: See my update, I simply point to the attribute on the declaration in NSTimer.h.

philipphofmann: I think you could also add the link as a code comment.

PR #3143, opened 2023-07-12
(chore): Delete stale.yml

hubertdeng123:
This logic for marking issues as stale is now moved to eng-pipes so it's easier to change the logic instead of going through all stale.yml files across Sentry.

https://github.com/getsentry/eng-pipes/blob/3236a12d691c4bf26721b3aad95cc247daad3813/src/webhooks/pubsub/stalebot.ts

#skip-changelog


philipphofmann (approved):
LGTM

PR #3145, opened 2023-07-13
fix: profiling memory growth fix

armcknight:
Following on from #3133, write a test that exercises the situation that is causing the memory growth bug. We actually already had a test that sets up the scenario, but prior to being able to mock backtraces, which that linked PR provides the ability to do, we were not able to distinguish which profile data comes from which profiler–and therefore tell that one had been lost. Now that we can disambiguate them with different mock data, we can ensure both are preserved. This test should fail in this PR and then pass in #3135 with the fixes.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Renamed `enableTimeToFullDisplay` to `enableTimeToFullDisplayTracing` (#3106)
     - This is an experimental feature and may change at any time without a major revision.
+    
+### Fixes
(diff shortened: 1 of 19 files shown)


indragiek (approved):

philipphofmann (approved):
LGTM, but the tests are failing.

armcknight:
The test is supposed to fail in this PR. We'll merge #3135 which is the fix into this branch before merging this to main.

Issue #3151, opened 2023-07-14
SentryCoreDataTracker *still* crashes when passed a nil error parameter


bjhomer:
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.9.1

### Steps to Reproduce

1. Set up Sentry with Core Data tracing
2. In an ObjC function, call `[managedObjectContext save:nil]`
3. Sentry crashes.

This issue was reported in #2433, and it was _partially_ fixed, there, but not enough to actually resolve the problem.

### Expected Result

Sentry should not crash

### Actual Result

Sentry crashes in `SentryCoreDataTracker`. It is checking if `*error == nil`, but since `error` comes from the caller, it might be nil. In that case, it's dereferencing a null pointer, which causes a crash.

<img width="977" alt="Screenshot 2023-07-14 at 2 36 41 PM" src="https://github.com/getsentry/sentry-cocoa/assets/276263/b71c02f3-c4d8-4271-b23f-d3e21347e43f">

The correct way to check if there was an error is to look at the _result_ of the function. If the function returns non-nil (or `true` if it's a Bool-returning function), then the call succeeded. Only if it failed should you look at the error, and even then, only if you passed in a non-nil error pointer in the first place.

Note that according to the [Cocoa Error Handling Programming Guide](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-BAJIIGCC), success is determined by the **return value of the function**, not by the value of the error parameter. You should only look at the error parameter if you already know the function failed.
 
<img width="882" alt="Screenshot 2023-07-14 at 2 39 34 PM" src="https://github.com/getsentry/sentry-cocoa/assets/276263/3dbaf027-450c-4d0b-aa1d-159544883287">

In the case of the calls to `SENTRY_LOG_DEBUG` here, we don't actually need to know details about the error at all; we just need to know if there *was* an error.

### Are you willing to submit a PR?

Yes, at least as a starting point.

Issue #2433, opened 2022-11-22
SentryCoreDataTracker crashes when passed a nil error parameter


bjhomer:
### Platform

iOS

### Installed

Swift Package Manager

### Version

7.31.2

### Steps to Reproduce

1. Enable Core Data profiling
2. Call `[someManagedContext save:nil]`, **passing a `nil` error pointer**

### Expected Result

Since it's legal to call `[managedObjectContext save:nil]`, Sentry should handle that situation appropriately.

### Actual Result

You get a crash in `-[SentryCoreDataTracker managedObjectContext:save:originalImp:]` [here](https://github.com/getsentry/sentry-cocoa/blob/f781ecba1e2ccfc37e2156f1a9300452c0c8f4c8/Sources/Sentry/SentryCoreDataTracker.m), because it tries to dereference the nil `error` pointer:

```objc
    if (fetchSpan) {
        [fetchSpan
            finishWithStatus:*error != nil ? kSentrySpanStatusInternalError : kSentrySpanStatusOk];

        SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically finished span with status: %@",
            *error == nil ? @"ok" : @"error");
    }
```

Sentry should not crash if the user reported a `nil` error. In fact, it should not be looking at the `error` parameter at all if it just wants to know whether an error occurred. The Cocoa conventions specify that an error is indicated by a nil (or zero) return value; you should only check the error parameter if you already know that an error occurred:

> **Important**: Success or failure is indicated by the return value of the method. Although Cocoa methods that indirectly return error objects in the Cocoa error domain are guaranteed to return such objects if the method indicates failure by directly returning nil or NO, you should always check that the return value is nil or NO before attempting to do anything with the NSError object.
> 
> (From the [Cocoa Error Handling Programming Guide](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-BAJIIGCC) )

A more correct implementation would look like this:

```objc
    if (fetchSpan) {
        [fetchSpan
            finishWithStatus:result == false ? kSentrySpanStatusInternalError : kSentrySpanStatusOk];

        SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically finished span with status: %@",
            result == true ? @"ok" : @"error");
    }
```

[...]

PR #3152, opened 2023-07-14
Avoid crashing in SentryCoreDataTracker

bjhomer:
## :scroll: Description

Users can pass in a `nil` error parameter when calling Core Data methods like `save:` and `executeFetchRequest:`. We were incorrectly handling that nil error parameter and trying to dereference it inside SENTRY_LOG_DEBUG calls. This commit fixes that by instead checking the *result* of the call, as recommended by the [Cocoa Error Handling Guide](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-BAJIIGCC).

This fixes https://github.com/getsentry/sentry-cocoa/issues/3151.

Honestly, I didn't. But I'm making the logging logic match the same logic used in `[fetchSpan finishWithStatus:]`

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryCoreDataTracker.m
+++ b/Sources/Sentry/SentryCoreDataTracker.m
@@ -67,7 +67,7 @@
             finishWithStatus:result == nil ? kSentrySpanStatusInternalError : kSentrySpanStatusOk];
 
         SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically finished span with status: %@",
-            error == nil ? @"ok" : @"error");
+            result == nil ? @"error" : @"ok");
     }
 
     return result;
@@ -109,7 +109,7 @@
         [fetchSpan finishWithStatus:result ? kSentrySpanStatusOk : kSentrySpanStatusInternalError];
 
         SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically finished span with status: %@",
-            *error == nil ? @"ok" : @"error");
+            result ? @"ok" : @"error");
     }
 
     return result;


Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -67,7 +67,7 @@ - (NSArray *)managedObjectContext:(NSManagedObjectContext *)context
             finishWithStatus:result == nil ? kSentrySpanStatusInternalError : kSentrySpanStatusOk];
 
         SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically finished span with status: %@",
-            error == nil ? @"ok" : @"error");
+            result == nil ? @"error" : @"ok");
```
bjhomer, line 70: I swapped the order of the parameters here to make the logic in the logging mirror the logic in the `[span finishWithStatus:]` call above.

Previously, this logging was almost always incorrect; it was just checking whether the user had passed in an `error` pointer. If so, it logged "error".

Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -109,7 +109,7 @@ - (BOOL)managedObjectContext:(NSManagedObjectContext *)context
         [fetchSpan finishWithStatus:result ? kSentrySpanStatusOk : kSentrySpanStatusInternalError];
 
         SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically finished span with status: %@",
-            *error == nil ? @"ok" : @"error");
+            result ? @"ok" : @"error");
```
bjhomer, line 112: If `result` is true, then the call succeeded, and we should log `ok`. We have no guarantee about the state of `error` in this case.

If `result` is false, then there was an error. *If* the user passed in an `error` pointer, it will now point to the error. But if the user passed in nil, then `error` will still be nil, and dereferencing it would cause a crash. In either case, we don't need the actual details of the error here at all; we only need to know _whether_ an error happened, and that can be determined entirely by looking at the result.

bjhomer:
This was partially fixed in `https://github.com/getsentry/sentry-cocoa/issues/2433`, but the logging still retained the old logic, so the crash remained. This PR fixes the logging logic.

philipphofmann (approved):
Thanks, @bjhomer, for the fix. I will fix the Changelog and add some tests in a subsequent PR, as I want to get a hotfix out quickly.

PR #3153, opened 2023-07-14
ref: rename SentryTracerConcurrency -> SentryProfiledTracerConcurrency

armcknight:
It's a more descriptive name, since this only related to tracers with profiles. I wanted to make this change in #3135 but it messed up the diff too much.

#skip-changelog


philipphofmann (approved):
LGTM

PR #3155, opened 2023-07-14
ref: use weak references to store tracers and profilers

armcknight:

Following onto #3154, store all but the necessary objects as weak references.

## :pencil: Checklist


philipphofmann:
I'm unsure if we actually need weak references. The bookkeeping of tracers and profilers is quite complex. I didn't have enough time today to consider all our available options. I will add a comment tomorrow.

armcknight:
It sounds like we're converging on a good solution in #3154 so this can be closed.

PR #3156, opened 2023-07-15
ref: remove private reference to hub

armcknight:
Remove a redundant ivar reference to SentryHub, and simplify function signature that accepted a hub instance and other parameters that could've been derived from it.

#skip-changelog


brustolin (approved):
LGTM

philipphofmann (approved):
Slightly better thanks 👍

PR #3157, opened 2023-07-15
ref: compile out more code from macOS builds that only works for UIKit

armcknight:
## :scroll: Description

Compile out more class (and their imports) that are for features that only work with UIKit.

Improves type safety, reduces the size of the maOS SDK by 2% (~101 KB) and runs slightly fewer tests for macOS. 

## :pencil: Checklist


philipphofmann (approved):
LGTM. As this reduces the SDK size for macOS by 2% we could add an entry to the changelog under an improvement header.

armcknight:
> LGTM. As this reduces the SDK size for macOS by 2% we could add an entry to the changelog under an improvement section.

True that! Gotta take the opportunity to brag whenever possible :)

PR #3158, opened 2023-07-15
ref: compile out more code/tests from tvOS with SENTRY_HAS_METRIC_KIT

armcknight:
Like #3157. Improves type safety, reduces the size of the tvOS SDK and runs slightly fewer tests for tvOS.

#skip-changelog

PR #3159, opened 2023-07-15
ref: compile out more code from platforms that don't support profiling

armcknight:
Like #3157 and #3158. Improves type safety, reduces the size of the tvOS SDK and runs slightly fewer tests for tvOS.

#skip-changelog


philipphofmann (approved):
LGTM, but CI is complaining.

PR #3161, opened 2023-07-15
ref: compile out more code not supported in watchOS

armcknight:
Same as #3157, #3158 and #3159: remove a bit more code from watch targets/tests.

#skip-changelog


philipphofmann (approved):
LGTM

PR #3162, opened 2023-07-15
test: remove live hub and client

armcknight:
Use mock hub in one test, remove the other test and use similar assertions in another test that already duplicates its logic.

#skip-changelog


philipphofmann (approved):
LGTM

PR #3163, opened 2023-07-15
ref: move SentryDateProvider to SentryDependencyContainer

armcknight:
Simplifies the date provider logic. We had a protocol that should have just been a class that could be subclassed by the test mock date provider. The static interface of SentryCurrentDate has been moved to SentryDependencyContainer.

#skip-changelog


philipphofmann (approved):
Ups, LGTM.

PR #3164, opened 2023-07-15
fix: use correct format specifier for NSInteger (either an int or a long int)

armcknight:
Saw this when debugging a test failure. `%zd` specifies a `size_t`, but this is formatting an amount of seconds. It seemed inappropriate to mix a value of time type with a type meant for memory values.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -611,7 +611,7 @@ SentryFileManager ()
 
 - (void)storeTimezoneOffset:(NSInteger)offset
 {
-    NSString *timezoneOffsetString = [NSString stringWithFormat:@"%zd", offset];
+    NSString *timezoneOffsetString = [NSString stringWithFormat:@"%ld", (long)offset];
     SENTRY_LOG_DEBUG(@"Persisting timezone offset: %@", timezoneOffsetString);
     @synchronized(self.timezoneOffsetFilePath) {
         if (![self writeData:[timezoneOffsetString dataUsingEncoding:NSUTF8StringEncoding]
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentryFileManager.m
```diff
@@ -611,7 +611,7 @@ - (NSNumber *_Nullable)readTimezoneOffset
 
 - (void)storeTimezoneOffset:(NSInteger)offset
 {
-    NSString *timezoneOffsetString = [NSString stringWithFormat:@"%zd", offset];
+    NSString *timezoneOffsetString = [NSString stringWithFormat:@"%ld", (long)offset];
```
philipphofmann, line 614: Please add a test case for this change.

philipphofmann (commented):
Good catch, thanks.

Sources/Sentry/SentryFileManager.m
armcknight: There is an existing test case `testStoreAndReadTimezoneOffset`, but i added another test case that does negative numbers.

philipphofmann: Sorry, I should have been more specific. You change the code for a reason cause it doesn't work for a specific use case. It would be great to add a test that would fail when using `@"%zd"`, but passes when using `@"%ld"`. I guess that's what you did with the extra test case.

philipphofmann (approved):
LGTM, maybe it's worth adding a changelog entry?

brustolin (approved):

Sources/Sentry/SentryFileManager.m
armcknight: Since the API that works with the timezone offset all restrict it to NSInteger, it's not possible to overflow the value that'll ultimately be passed to this. I think the tests are good at this point.

PR #3165, opened 2023-07-17
test: Add tests for CoreDataCrash

philipphofmann:
Add a changelog entry and some tests for #3152.

#skip-changelog


Tests/SentryTests/Integrations/Performance/CoreData/SentryCoreDataMiddleware+Extension.swift
```diff
@@ -2,12 +2,20 @@ import Foundation
 
 extension SentryCoreDataMiddleware {
     
-    func fetchManagedObjectContext<T>(_ context: NSManagedObjectContext, request: NSFetchRequest<T>, originalImp: (NSFetchRequest<T>, NSErrorPointer) -> [T]?) throws -> [Any] {
+    func fetchManagedObjectContext<T>(_ context: NSManagedObjectContext, request: NSFetchRequest<T>, isErrorNil: Bool = false, originalImp: (NSFetchRequest<T>, NSErrorPointer) -> [T]?) throws -> [Any] {
```
armcknight, line 5: I'm confused on how the CoreDataTracker, middleware and swizzling work. Are the methods in `@protocol SentryCoreDataMiddleware` supposed to be called by SDK consumers, or are they called somewhere else by us in the SDK code? `SentryCoreDataSwizzling.h` is not a public header, so it must be coming from the SDK somewhere.

What I'm searching for is where the inout NSError actually originates.

I think we should add headerdocs to SentryCoreDataSwizzling.h to explain how it's supposed to work.

armcknight: I think I get it now, we swizzle `-[NSManagedObjectContext executeFetchRequest:error:]` et al and the protocol methods are substituted for the originals.

Seeing as how there is only one class conforming to SentryCoreDataMiddleware I'm not convinced we even need that protocol. I think I brought this up previously when it was being implemented. It's overly complicated.

Tests/SentryTests/Integrations/Performance/CoreData/SentryCoreDataTrackerTest.swift
```diff
@@ -236,6 +236,22 @@ class SentryCoreDataTrackerTests: XCTestCase {
...
+        let fetch = NSFetchRequest<TestEntity>(entityName: "TestEntity")
+        
+        let transaction = startTransaction()
+        let sut = fixture.getSut()
+        
+        let context = fixture.context
+        
+        let _ = try?  sut.fetchManagedObjectContext(context, request: fetch, isErrorNil: true) { _, _ in
+            return nil
+        }
```
armcknight, line 249: A lot of the test cases in here use `try?`, but could be strengthened with `XCTAssertNoThrow`/`XCTAssertThrowsError`.

```suggestion
        XCTAssertNoThrow(try sut.fetchManagedObjectContext(context, request: fetch, isErrorNil: true) { _, _ in
            return nil
        })
```

Tests/SentryTests/Integrations/Performance/CoreData/SentryCoreDataMiddleware+Extension.swift
philipphofmann: Yes, I was also confused. I can improve it a bit.

armcknight: Thanks for doing that! 🙏🏻

PR #3172, opened 2023-07-19
fix: Use "__has_include" instead of platform check

brustolin:
Instead of filtering features out by platform directives, I believe we should check whether the framework we need exists with `__has_include`.

_#skip-changelog_


brustolin:
This wont work because of WatchOS. Apple use the same library name "UIKit" but it does not contains every class iOS have.

philipphofmann:
I also tried `__has_include` for MetricKit, but it didn't work.

PR #3175, opened 2023-07-19
ref: prevent implicit UIKit linkage

armcknight:
Prevent UIKit from being implicitly linked through usage in Sentry.framework. This is helpful for customers who would like to use the framework in an iOS extension without using UIKit-based features, which can help cut down on the process' memory footprint. This may also be beneficial for anyone using Sentry in an iOS app in a similar fashion.

- Adds new build configurations `Debug_without_UIKit` and `Release_without_UIKit` that were cloned from `Debug` and `Release` 
    - set `CLANG_MODULES_AUTOLINK` to `NO` to prevent UIKit from being linked
    - add a script to run in build.yml to ensure that the build products from these do not link UIKit (using `otool -L`)
- `Debug`, `Release`, `Test` and `TestCI` configurations have a new `GCC_PREPROCESSOR_DEFINITIONS` entry `SENTRY_UIKIT_LINKED=1`; also define this in the CocoaPods and SPM specs
- Redefine `SENTRY_HAS_UIKIT` to additionally only be true if `SENTRY_UIKIT_LINKED` is true
- Move a few more UIKit-only API/implementations inside `SENTRY_HAS_UIKIT` conditional compilation so it doesn't get compiled for the wrong configurations
- remove a few redundant definitions in SentryCrashSystemCapabilities.h like `SentryCrashCRASH_HAS_UIKIT`, `SentryCrashCRASH_HAS_MESSAGEUI`, `SentryCrashCRASH_HAS_UIALERTCONTROLLER` and `SentryCrashCRASH_HAS_UIALERTVIEW` which are unused, and `SentryCrashCRASH_HAS_UIAPPLICATION` and `SentryCrashCRASH_HAS_UIDEVICE` which are the same as `SENTRY_HAS_UIKIT`


Sentry.xcodeproj/project.pbxproj
```diff
@@ -4558,6 +4566,8 @@
 				CLANG_ANALYZER_NONNULL = YES;
 				CLANG_ANALYZER_NUMBER_OBJECT_CONVERSION = YES;
 				CLANG_ENABLE_OBJC_ARC = YES;
+				CLANG_ENABLE_OBJC_WEAK = YES;
+				CLANG_MODULES_AUTOLINK = NO;
```
armcknight, line 4570: This is one of the crucial parts. It requires explicitly linking CoreData and SystemConfiguration frameworks.

armcknight:
closing this in favor of #3178

armcknight:
Reopening this because #3178 will place extra complexity on the part of SDK consumers, and may not work for all combinations of targets. This solution Just Works as far as customers are concerned.

armcknight:
https://github.com/getsentry/sentry-cocoa/pull/3175/commits/f1e92f5b53e087f53a13692134a7302141b46ee9 Makes it so that autocomplete still works with the macros. It adds a new build config with a preprocessor definition that allows us to selectively define the UIKit macros to either the actual class name token or NSClassFromString result. I tried a couple different setups but the current form with the undefs is the only one where autocomplete actually worked.

Sources/Sentry/Public/SentryDefines.h
```diff
@@ -18,6 +18,123 @@
 #    define SENTRY_HAS_METRIC_KIT 0
 #endif
 
+#if __has_include(<UIKit/UIKit.h>)
```
brustolin, line 21: h: `__has_include` wont work here. Even if the project is not liked with UIKit. this returns true.

armcknight: I think it'll be cleaner to use SENTRY_HAS_UIKIT here, although they're equivalent. What it does is only defines the macros below for iOS/tvOS and Catalyst (confirmed by putting a `#warning blah` right under it and compiling for each platform)

armcknight: We still need UIKit.h to be included for all the parameter/value types from UIKit to still work. Including the header alone doesn't cause UIKit to be linked, so it's ok to do so. It's only the explicit message sends to UIKit classes that causes the linker to include dyld load commands. But it is still possible to include the header even if it won't be linked. The header delivers declarations, the load command loads the binary implementation.

armcknight:
Something must've come in with the merge from main because now i see UIKit linkage again in otool output

Tests/SentryTests/Integrations/Breadcrumbs/SentryAutoBreadcrumbTrackingIntegrationTests.swift
```diff
@@ -68,7 +70,9 @@ class SentryAutoBreadcrumbTrackingIntegrationTests: XCTestCase {
         SentrySDK.setCurrentHub(hub)
         
         let crumb = TestData.crumb
+#if os(iOS)
         fixture.systemEventBreadcrumbs?.startWithdelegateInvocations.first?.add(crumb)
+#endif // os(iOS)
```
philipphofmann, line 75: `m`: I think this change could be responsible for `SentryAutoBreadcrumbTrackingIntegrationTests.testInstall` failing. The test succeeds for all iOS targets and fails for the other ones. The change here or in `SentryAutoBreadcrumbTrackingIntegration.m` most likely is responsible for this failing test.

Sources/Sentry/include/SentryUIApplication.h
```diff
@@ -2,7 +2,10 @@
 
 #if SENTRY_HAS_UIKIT
 
-#    import <UIKit/UIKit.h>
+@class UIApplication;
+@class UIScene;
+@class UIWindow;
+@protocol UIApplicationDelegate;
```
philipphofmann, line 8: Way better 🥳

.github/workflows/build.yml
```diff
@@ -166,3 +166,23 @@ jobs:
       - uses: actions/checkout@v4
       - run: swift build
         shell: sh
+
+  check-uikit-linkage-debug:
```
philipphofmann, line 170: 🥇 for adding some CI validation

Sources/Sentry/UIView+Sentry.m
```diff
@@ -1,34 +0,0 @@
-#import "UIView+Sentry.h"
-
-#if SENTRY_HAS_UIKIT
```
philipphofmann, line 3: `m`: Are you sure we don't need this file anymore?

Sources/Sentry/UIViewController+Sentry.m
```diff
@@ -1,32 +0,0 @@
-#import "UIViewController+Sentry.h"
-
-#if SENTRY_HAS_UIKIT
-
-@implementation
-UIViewController (Sentry)
-
-- (NSArray<UIViewController *> *)sentry_descendantViewControllers
```
philipphofmann, line 8: `h`: I would rather keep the file and its tests, please.

philipphofmann (commented):
That's an elegant solution. It doesn't require any changes from our customers 💪. I think we should add an entry to the Changelog, as it's a nice feature. Furthermore, I think it would be great to document the approach in our [internal develop docs](https://github.com/getsentry/sentry-cocoa/tree/main/develop-docs), so we don't have to find the PR to understand how this approach works.

I added some comments. Thanks a lot for tackling this @armcknight 👍

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -199,7 +199,6 @@ NS_SWIFT_NAME(Options)
  */
 @property (nonatomic) SentryScope * (^initialScope)(SentryScope *);
 
-#if SENTRY_HAS_UIKIT
```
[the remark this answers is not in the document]
armcknight: Oh hm, I thought I had added it in the middle of this PR and then decided not to do it here. Guess I was wrong 😆 

But, now that `SENTRY_HAS_UIKIT` has been extended to include whether a `GCC_PREPROCESSOR_DEFINITIONS` is set, I think we can't use this in headers any longer, since the header could be included into targets that don't get our `GCC_PREPROCESSOR_DEFINITIONS` values, this would never evaluate to true.

What I did instead was override the setters for these options as no-ops for `!SENTRY_HAS_UIKIT` and print a warning log that they do nothing for configurations that don't link UIKit.

Sources/Sentry/UIViewController+Sentry.m
armcknight: That is fine, I can do that. I removed the category because it didn't work with my first attempt, which was to conditionally define macros for every UIKit thing. Now that I can just gate compilation of UIKit things, I can do it here too 👍🏻

Sources/Sentry/UIView+Sentry.m
armcknight: see https://github.com/getsentry/sentry-cocoa/pull/3175#discussion_r1347755645

Sources/Sentry/Public/SentryOptions.h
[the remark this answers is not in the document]
armcknight: I'll reevaluate this claim. I was unable to build the sample apps because the options gated by this were not able to be located. But SentryDefines.h itself is also a public header, and seemed to evaluate the macros correctly.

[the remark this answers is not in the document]
armcknight: So there's something going on with how regions surrounded by `SENTRY_HAS_UIKIT` are visible to Swift code that imports the Sentry module. The last thing that's holding me up is that when I remove gated compilation here: https://github.com/getsentry/sentry-cocoa/blob/2fe817d73be7b83f45f72bd55adf132d0f594bcb/Samples/iOS-Swift/iOS-Swift/ExtraViewController.swift#L34-L38

the iOS-Swift build fails because the compiler can't see `PrivateSentrySDKOnly.currentScreenFrames`. The stranger thing now is that even if I remove the `#if SENTRY_HAS_UIKIT` surrounding that and other declarations in PrivateSentrySDKOnly.h, it _still_ isn't building. I'm digging in a little more but no idea why this is happening right now :/ 

This is essentially the same thing I saw in the other apps though, which is why I removed those, and those builds now work. I'm wondering if there's something special about PrivateSentrySDKOnly.h though, since it's in a different subdirectory.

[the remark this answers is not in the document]
armcknight: Yeah there's definitely something going on with Swift apps consuming Sentry being unable to use declarations from Sentry.framework/PrivateHeaders, I also tried it with another one and that fails too: 
image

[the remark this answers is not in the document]
armcknight: So the issue was that the GCC_PREPROCESSOR_DEFINITION setting being used to set the macro definition in SentryDefines.h had no effect in headers delivered in the framework bundle. So for headers exposed in the framework bundle, I reverted to the previous type of macro, under a new name `SENTRY_UIKIT_AVAILABLE`, and stubbed the implementations depending on `SENTRY_HAS_UIKIT` (which, again, means it's available _and_ linked). You can see https://github.com/getsentry/sentry-cocoa/pull/3175/commits/ccd24659d579657049fd56f63822fb5988329c02

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Sentry can now be used without linking UIKit; this is helpful for using the SDK in certain app extension contexts (#3175)
```
philipphofmann, line 7: 🎉

Sources/Sentry/Public/SentryOptions.h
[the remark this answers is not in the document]
philipphofmann: Thanks for explaining, but why do we now use a combination of both `SENTRY_UIKIT_AVAILABLE` and `SENTRY_HAS_UIKIT` https://github.com/getsentry/sentry-cocoa/blob/a697fca629f46b6296f29bf28ad4d303c7eecbef/Sources/Sentry/SentryOptions.m#L635-L645

Wouldn't only using `SENTRY_HAS_UIKIT` be sufficient? I don't understand why we want to expose functionality when UIKit is available on the platform but not linked and to provide a fallback that prints a warning. Only using `SENTRY_HAS_UIKIT` would reduce complexity, or is that not possible?

philipphofmann (commented):
Only the ongoing discussion about using `SENTRY_UIKIT_AVAILABLE` and `SENTRY_HAS_UIKIT` stops me from approving this: https://github.com/getsentry/sentry-cocoa/pull/3175/#discussion_r1348151964

Sources/Sentry/Public/SentryOptions.h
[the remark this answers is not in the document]
armcknight: The old way that `SENTRY_HAS_UIKIT` was defined only tells us if we're compiling for either an iOS or tvOS target. So this works to compile out UIKit code for eg macOS, but if we want to compile out UIKit code on iOS/tvOS, we need another option to switch over. The only way I thought of to do so was with the new build configs that deliver a `GCC_PREPROCESSOR_DEFINITION` that I called `SENTRY_UIKIT_LINKED`, which we now check for when defining `SENTRY_HAS_UIKIT`. This is why I renamed what used to be `SENTRY_HAS_UIKIT` (`TARGET_OS_IOS || TARGET_OS_TVOS`) to `SENTRY_UIKIT_AVAILABLE` and changed `SENTRY_HAS_UIKIT` to be not only `SENTRY_UIKIT_AVAILABLE`, but also `SENTRY_UIKIT_LINKED` for the build configs that do so. I think the changed naming is more accurate for both cases as well. Let me know if this makes sense to you; I also wrote this up in the develop-docs/README.md in https://github.com/getsentry/sentry-cocoa/pull/3175/commits/2f5c366c2f960f04d96d2df2e0cc4282f53005de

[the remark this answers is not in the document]
brustolin: I agree with Philipp, it seems to me that `#if SENTRY_HAS_UIKIT` is what we should use, because it addresses both situation, it's the right target and linking UIKit is enabled.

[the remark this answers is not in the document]
philipphofmann: > What I saw was our test apps would not compile for both Debug and Debug_without_UIKit, they'd break on the latter because the symbol was compiled out but the sample app doesn't inherit the GCC_PREPROCESSOR_DEFINITION.

Not compiling for Debug is suspicious. Our sample apps not compiling for `Debug_without_UIKit`  should be fine. I guess we need a new sample without UIKit to validate `Debug_without_UIKit`.

> We also don't want to make people have to write #if/#endif regions around code we conditionally compile. We figured it was better to not break their builds and just print warnings (and I added documentation to all such public API).

We absolutely can't break existing builds in a minor version and writing `#if/#endif regions` is also suboptimal.

I agree with @brustolin. @armcknight, can we somehow make it work without breaking customers? 

As you said, maybe it's better to discuss this over a call instead of ping-pong here on GH.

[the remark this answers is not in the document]
armcknight: >> our test apps would not compile for both Debug and Debug_without_UIKit

> Not compiling for Debug is suspicious. Our sample apps not compiling for Debug_without_UIKit should be fine

Sorry, I should've been clearer. This is actually the case: before, it built for debug, but not debug_without_uikit. After my changes, it built for both, and only emitted the warning logs in debug_without_uikit. I did this so no code changes would be necessary in the sample app. This is my definition of not breaking customers.

I can add assertions instead of just warning logs, so that debug builds crash, but not production. How does that sound? I will push a commit with this change.

philipphofmann (approved):
Sorry for the long discussion. I think it's Okay how it is. LGTM if the following is true: 

For platforms where UIKit is never available as macOS, you don't have access to UIKit options such as `enableUIViewControllerTracing`. On iOS, where UIKit can be available, you always have access to `enableUIViewControllerTracing`, but if UIKit is not linked, the SDK logs a warning and assert so it doesn't compile in debug. Ideally, the option wouldn't be available, but I'm happy to accept this tradeoff as it's for very few users, and this PR allows them to use Sentry. If `enableUIViewControllerTracing` is now available on macOS, I think it's not ideal, and we should try to fix it. I tried it locally, and `enableUIViewControllerTracing` is not available, so it should be fine.

Sources/Sentry/SentryAppStartMeasurement.m
```diff
@@ -41,8 +60,50 @@ - (instancetype)initWithType:(SentryAppStartType)type
...
+- (NSDate *)moduleInitializationTimestamp
+{
+    return _moduleInitializationTimestamp;
+}
+
+- (NSDate *)didFinishLaunchingTimestamp
+{
+    return _didFinishLaunchingTimestamp;
 }
+#    endif // SENTRY_HAS_UIKIT
```
philipphofmann, line 105: `l`: Do we need these getters?

armcknight: nope, you're right. forgot that i'd originally added them to override depending on the macro, but then realized that if i make the inits return nil, no overrides were necessary. they're gone!

Issue #2847, opened 2023-03-31
MetricKit ignore stacktraces from the SDK


philipphofmann:
### Description

Some MXDiskWriteExceptions contain Sentry stacktraces 

<img width="681" alt="Screenshot 2023-03-31 at 11 37 30" src="https://user-images.githubusercontent.com/2443292/229084580-18eb4a67-88d4-4cb7-a893-faa13b472c51.png">

Maybe we should ignore these as every interaction with NSData also calls `SentryNSDataTracker,` and it's noise for the user.

The same applies for the `SentryBreadcrumbTracker` and `SentryANRTracker detectANRs`.
<img width="458" alt="Screenshot 2023-03-31 at 11 44 35" src="https://user-images.githubusercontent.com/2443292/229086205-3e7a5521-6336-4ad1-8f9e-eab0a25d294a.png">

philipphofmann:
After some investigation in https://github.com/getsentry/sentry-cocoa/pull/3176, we can conclude that doing this reliably on the client side is complicated. We could discard such events on the backend, but then we kind of lie. If MetricKit detects an issue inside our SDK we shouldn't discard the issue, but report it. We can collect such issues alongside with our SDK crashes and try to fix them: https://github.com/getsentry/sentry/issues/57961.

PR #3176, opened 2023-07-20
feat: Ignore specific MetricKit reports

philipphofmann:
This is a proof of concept for ignoring MetricKit reports that stem from our SDK (https://github.com/getsentry/sentry-cocoa/issues/2847). For example, `SentryNSDataSwizzling` hooks into almost every file I/O operation. MetricKit detects this and assumes that it's responsible for a MXDiskWriteException, which doesn't make any sense. Therefore, we could ignore such events. 

The idea is to walk through the stacktrace and check if the topmost non-system frame is, for example, from  `SentryNSDataSwizzling`. As we don't have function names available on the device, cause symbolication won't work on the device in release, we have to work with memory addresses of functions. What could work is to get the `IMP`, which is the instruction address, then use local symbolication with `dladdr` to get the symbol address and do the same with the MetricKit stacktrace. Then, we can compare the symbol addresses. To know if it's a system frame, we can use the same approach as for SDK crash detection. We simply check if the path of the binary image starts with `/System/Library/` or `/usr/lib/`, but again we need local symbolication to work for this. It's worth noting that we won't get function names with local symbolication, but I hope we can get the symbol addresses and the full paths of the binary images.

One downside of this approach is, that it won't work after app updates or OS updates.

I hacked this PR together to get some initial feedback.


armcknight:
This could be a good reason to switch from the swizzle-based File I/O detection to profile-based. All backend, and no client changes required since we can do this today with profiling by just flipping a switch (except for removing the swizzle, which, viewed purely in client times, would be a strict improvement). I guess the hardest part would be migrating the current perf issues to the new prof issues.

cc @indragiek because we've talked about this in the past.

armcknight:
We're going to close this draft for now as we're not sure it makes sense for the client and @philipphofmann is on leave for a while.

Issue #3174, opened 2023-07-19
iOS SDK imports UIKit which conflicts with FileProvider


rpropst:
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.9.1

### Steps to Reproduce

1. Have a project that implements FileProvider
2. Add Sentry iOS SDK
3. Fails to compile

- Incorporating sentry using our modified fork of buck, with their own spec file based on Sentry's `spm/cocoapods` specs

- We have operational reasons to disinclude uikit from the extension. so while it's technically fine, we cannot have anything linking it :confused:

- we have a 20 or 25 mb limit for a file provider ext after which apple watchdog kills the process, this is part of the reason they don't link UIKit
![image (1)](https://github.com/getsentry/sentry-cocoa/assets/3869147/6ac66741-8b71-4ea2-a52c-08d9071127a6)

### Expected Result

Compiles

### Actual Result

Fails compilation

armcknight:
Let's try to get this complete by end of September.

PR #3178, opened 2023-07-20
feat: Add SENTRY_IGNORE_UIKIT compiler directive to avoid linking with UIKIT

brustolin:
## :scroll: Description

Added SENTRY_IGNORE_UIKIT compiler directive in order to remove from the build everything related to UIKIT. 

Users will be able to add the `-DSENTRY_IGNORE_UIKIT` compiler flag, or add a preprocessor macro `SENTRY_IGNORE_UIKIT=1` to their Sentry build step to avoid linking with UIKIT and saving a few bytes from the SDK.

image
_Only one of these two options required_

This could be the solution for #3174

Manual test with `otool`

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Reduced macOS SDK footprint by 2% (#3157) with similar changes for tvOS and watchOS (#3158, #3159, #3161)
 
+### Features
+
+- Add SENTRY_IGNORE_UIKIT compiler directive to avoid linking with UIKIT (#3178)
+
 ### Fixes
 
 - Fix a crash in SentryCoreDataTracker for nil error params (#3152)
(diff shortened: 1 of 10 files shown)


armcknight (approved):
This is a much cleaner approach. I think we should introduce a new build configuration with those preprocessor macros already set so that people can select them with their package manager of choice, instead of having to manually add them in our project's build settings

armcknight (changes requested):
Unfortunately we received some negative feedback from a potential customer on this approach. Even though it's a bit more complicated on our end, we should consider going with the solution in #3175

brustolin:
We wont continue with this solution in favor of https://github.com/getsentry/sentry-cocoa/pull/3175

PR #3184, opened 2023-07-24
test: allow starting/stopping txs from either main/bg threads

armcknight:
I wrote this to help debug a potential issue. It lets you start a transaction from the main thread or bg thread, keeps a list of transactions that you've started, and then when stopping a transaction, gives you a list to choose which one to stop, and then lets you stop it from either the main thread or a bg thread.

https://github.com/getsentry/sentry-cocoa/assets/3241469/9fcc5e37-eef8-4dea-a563-47fb47c72357

#skip-changelog


brustolin (commented):
Looks good. Just need to fix UI tests

brustolin (approved):

PR #3187, opened 2023-07-25
EXP: energy monitoring

armcknight:
Add more metric gathering in SentryBenchmarking. This is a combination of several disparate branches i had locally from various work in the past for metric profiling, benchmarking and now for the energy measurement investigation. I found it easiest to combine it all in a common pattern; we can leave measurements in SentryBenchmarking (which only compiles into the iOS-Swift sample app) unless/until we decide to release them to customers, then we can move them into the SDK.

Add a new tab to iOS-Swift for Profiling specific stuff. This allows manual starting/stopping of benchmarking, and test scenarios. The old benchmarking setup was removed in favor of this: it used to only run for a set duration on a VC load, which was a more fragile and finicky approach.

This branch contains some commits that will not be merged (enabling automatic codesigning to help on-device testing).

#skip-changelog


brustolin (commented):
First look. For now I have one comment.
I don't like the idea of increasing the size of our samples with resources. 
This PR adds some images and JSONs, and every user that uses Cocoapod or SPM will be getting all of that.

armcknight:
The huge diff is from some test fixtures i can remove, as they aren't needed anymore.

armcknight:
Closing this and will open a couple separate PRs with relevant things from it.

PR #3194, opened 2023-07-31
feat: allow profiling from hybrid SDKs

vaind:

Cocoa SDK built-in native profiler can be used to support profiling Flutter apps, see https://github.com/getsentry/sentry-dart/issues/1106

This PR refactors some profiler APIs so that they don't depend on the cocoa `Tracer` & `Transaction` and instead expect only the necessary arguments. 

#skip-changelog

All previous tests pass & adding new tests for new hybrid SDK APIs

## :pencil: Checklist


brustolin (commented):
This looks ok to me, but I would like for @armcknight to also review this.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -126,21 +126,20 @@
...
-        if (!orderedChronologically(transaction.startSystemTime, sample.absoluteTimestamp)) {
+        if (!orderedChronologically(startSystemTime, sample.absoluteTimestamp)) {
             SENTRY_LOG_WARN(@"Filtered sample not chronological with transaction.");
             return;
         }
         const auto dict = [NSMutableDictionary dictionaryWithDictionary:@ {
-            @"elapsed_since_start_ns" : sentry_stringForUInt64(
-                getDurationNs(transaction.startSystemTime, sample.absoluteTimestamp)),
+            @"elapsed_since_start_ns" :
+                sentry_stringForUInt64(getDurationNs(startSystemTime, sample.absoluteTimestamp)),
```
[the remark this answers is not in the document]
armcknight: No, it must be a string because the backend JSON implementation can't support unsigned 64 bit integers. Unless that has changed, cc @phacops

Sources/Sentry/PrivateSentrySDKOnly.m
```diff
@@ -117,6 +119,30 @@ + (NSDictionary *)getExtraContext
...
+                                                            since:(uint64_t)startSystemTime;
+{
+    NSMutableDictionary<NSString *, id> *payload = [SentryProfiler
+        collectProfileBetween:startSystemTime
+                          and:SentryDependencyContainer.sharedInstance.dateProvider.systemTime
+                     forTrace:traceId
+                        onHub:[SentrySDK currentHub]];
+
+    if (payload != nil) {
+        payload[@"platform"] = @"cocoa";
```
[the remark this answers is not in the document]
armcknight: Yeah, this is hardcoded the same way in SentryEvent.m. We could define it in SentryInternalDefines.h and use that constant in all three places, are you able to import that header here?

Sources/Sentry/PrivateSentrySDKOnly.m
```diff
@@ -117,6 +119,30 @@ + (NSDictionary *)getExtraContext
...
+    }
+
+    return payload;
+}
+#endif
```
armcknight, line 144: I like to always comment the conditional scope being closed:
```suggestion
#endif // SENTRY_TARGET_PROFILING_SUPPORTED
```

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -55,12 +54,12 @@
 std::mutex _gStateLock;
 
 void
-trackProfilerForTracer(SentryProfiler *profiler, SentryTracer *tracer)
```
armcknight, line 58: I purposely used SentryTracer here to keep type safety. SentryEvent, SentrySpan and SentryTracer all have a SentryId property, but they're not interchangeable, so I did this to prevent any mistakes.

Sources/Sentry/SentryMetricProfiler.mm
```diff
@@ -46,23 +46,22 @@ @implementation SentryMetricReading
  */
 SentrySerializedMetricEntry *_Nullable serializeValuesWithNormalizedTime(
     NSArray<SentryMetricReading *> *absoluteTimestampValues, NSString *unit,
-    SentryTransaction *transaction)
```
armcknight, line 49: Same thing here with type safety, in addition to cutting down the amount of parameters. Unless this is required for this PR due to some nuance with compilation for hybrid platforms, I'd prefer this be reverted.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -153,11 +152,11 @@
     return result;
 }
 
-NSDictionary<NSString *, id> *
-serializedProfileData(
-    NSDictionary<NSString *, id> *profileData, SentryTransaction *transaction, SentryId *profileID,
-    NSString *truncationReason, NSDictionary<NSString *, id> *serializedMetrics,
-    NSArray<SentryDebugMeta *> *debugMeta, SentryHub *hub
+NSMutableDictionary<NSString *, id> *
```
armcknight, line 155: Why was this changed from immutable to mutable?

Sources/Sentry/include/SentryProfiler.h
```diff
@@ -43,7 +44,7 @@ SENTRY_EXTERN_C_END
 /**
  * Start a profiler, if one isn't already running.
  */
-+ (void)startWithTracer:(SentryTracer *)tracer;
```
armcknight, line 46: Same note for type safety here.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -358,21 +341,21 @@ + (BOOL)isCurrentlyProfiling
 + (nullable SentryEnvelopeItem *)createProfilingEnvelopeItemForTransaction:
     (SentryTransaction *)transaction
 {
-    const auto profiler = profilerForFinishedTracer(transaction.trace);
-    if (!profiler) {
-        SENTRY_LOG_WARN(@"Expected a profiler for tracer id %@ but none was found",
-            transaction.trace.traceId.sentryIdString);
```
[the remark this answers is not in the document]
armcknight: Nice find 👍🏻

Sources/Sentry/SentryProfiler.mm
```diff
@@ -358,21 +341,21 @@ + (BOOL)isCurrentlyProfiling
...
+    const auto payload = [self collectProfileBetween:transaction.startSystemTime
+                                                 and:transaction.endSystemTime
+                                            forTrace:transaction.trace.traceId
+                                               onHub:transaction.trace.hub];
```
armcknight, line 347: Similar thought here w.r.t. parameters, why not just pass the transaction? This is more repetitive and verbose.

armcknight (commented):
The refactor of the profiling code to expose the `+[collectProfileBetween:...]` looks fine, the part I'm confused about is why things were changed from passing a SentryTracer or SentryTransaction, in favor more parameters that are all encapsulated by them already? There is a [SentryTracer](https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry_tracer.dart) and [SentryTransaction](https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/protocol/sentry_transaction.dart) in the dart codebase, so I assume we'd be able to call a function like `+[PrivateSentrySDKOnly startProfilingForTrace:(SentryTracer *)]` instead of `+[PrivateSentrySDKOnly startProfilingForTrace:(SentryId *)]`. Let me know if I'm missing something there, I don't have any experience writing Dart or with our SDK for it.

Sources/Sentry/SentryProfiler.mm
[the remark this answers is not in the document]
vaind: I haven't changed any of this here, just noticed and pointed out that the timestamp is sent as a string here but an integer in (some of the) other SDKs

armcknight:
Thanks for the thoughtful reply @vaind . There is an issue though:

> Profiler doesn't require a transaction to be running, so it should not depend on it in its argument

This is not true for our other platforms right now, so it would require backend changes to make it so. We currently must always have a transaction associated to each profile for dynamic sampling reasons, and profiles are uploaded as attachment items in transaction envelopes. We can't wind up with a profile which has had its transaction DS'd out, for instance. So we also can't have a profile that never had a transaction in the first place. 

The only way for profiler to start currently is when a tracer starts: https://github.com/getsentry/sentry-cocoa/blob/279841c5c6a89ce0f25896daef5a55567c534240/Sources/Sentry/SentryTracer.m#L164-L170

And then when a tracer ends is the only place a profile can possibly be uploaded:
https://github.com/getsentry/sentry-cocoa/blob/279841c5c6a89ce0f25896daef5a55567c534240/Sources/Sentry/SentryTracer.m#L526-L552

FYI, we do want to get to a place where we can separate profiles from transactions, but that is a rough plan for the future.

Does this change how you see the state of things?

vaind:
> This is not true for our other platforms right now, so it would require backend changes to make it so. We currently must always have a transaction associated to each profile for dynamic sampling reasons, and profiles are uploaded as attachment items in transaction envelopes. We can't wind up with a profile which has had its transaction DS'd out, for instance. So we also can't have a profile that never had a transaction in the first place.
> ...
> Does this change how you see the state of things?

No, not really, I am aware how profiling works, from other I've implemented it in. 

In case of Flutter, we need to capture a native profile. Therefore, the profile will be collected by the native (cocoa) profiler and then handed over to Flutter where it will be serialized and attached (as an additional envelope item) to the same envelope as the transaction. This is an important point - the transaction attached to the profile is a Dart (Flutter) transaction, being created directly in Dart and going through all the handling, user callbacks, etc. in Dart user code. This transaction has nothing to do with objective-c `SentryTransaction` & `SentryTracer`. Therefore, the profiler should be able to run & collect profiles for given periods of time regardless of a running (objective-c) transaction. Also note the distinction: I'm not saying it should run without a running transaction when triggered from objective-c; I'm just saying it should be able to be launched without a transaction, and this is exposed in the "private" hybrid SDK API. This PR changes nothing for native apps.

armcknight (approved):
OK, I thought that might be the case. Thanks for explaining @vaind  👍🏻

PR #3196, opened 2023-08-01
Distributed tracing without performance

### LINKED ISSUE #3136: Tracing without Performance
See Project: https://github.com/getsentry/team-sdks/issues/5
See Project Board: [github.com/orgs/getsentry/projects/110](https://github.com/orgs/getsentry/projects/110)

Note: implementation details/requirements for Mobile (vs. backend) SDKs TBD. PR for Java can act as reference: https://github.com/getsentry/sentry-java/pull/2788

brustolin:
## :scroll: Description

Pass through / add sentry-trace and baggage even if performance is disabled.

Closes #3136 

Unit tests

## :pencil: Checklist


brustolin:
Hey @adinauer, since you implemented this for Java could you take a look at my implementation? Thanks!

Sources/Sentry/SentryPropagationContext.m
```diff
@@ -0,0 +1,50 @@
...
+        self.spanId = [[SentrySpanId alloc] init];
+    }
+    return self;
+}
+
+- (SentryTraceHeader *)traceHeader
+{
+    return [[SentryTraceHeader alloc] initWithTraceId:self.traceId
+                                               spanId:self.spanId
+                                              sampled:kSentrySampleDecisionNo];
```
markushi, line 28: Not sure if this is relevant, but in `sentry-java` sampled can be `true`, `false` or `null`.

Sources/Sentry/SentryScope.m
```diff
@@ -558,6 +563,12 @@ - (SentryEvent *__nullable)applyToEvent:(SentryEvent *)event
             newContext[@"trace"] = [span serialize];
         }
     }
+
+    if (event.exceptions.count > 0 && newContext[@"trace"] == nil) {
```
markushi, line 567: Is there are reason this is being done depending on the exception count? `sentry-java` only checks [the trace context](https://github.com/getsentry/sentry-java/blob/fb296f0fa15044af53a7c6d62d168b732eb16500/sentry/src/main/java/io/sentry/SentryClient.java#L677-L685).

markushi (approved):
Nice one, looks good to me! Left two comments, but please take them with a grain of salt - I'm not too familiar with the cocoa SDK (yet) 😅

Sources/Sentry/SentryScope.m
brustolin: The count check is only to be sure this is an error event. If I got it right we dont send trace for capture message. But I just confirm that capture would not run this code anyway.

Sources/Sentry/SentryPropagationContext.m
brustolin: Maybe because java may run in the server and propagate whats comes from the client? I have no clue how this could have any other value here in the cocoa.

Sources/Sentry/SentryScope.m
brustolin, line 567:
```suggestion
    if (newContext[@"trace"] == nil) {
```

Sources/Sentry/SentryNetworkTracker.m
```diff
@@ -253,6 +212,65 @@ - (void)urlSessionTaskResume:(NSURLSessionTask *)sessionTask
...
+            [currentRequest setValue:traceHeader.value forHTTPHeaderField:SENTRY_TRACE_HEADER];
+
+            if (baggageHeader.length > 0) {
+                [currentRequest setValue:baggageHeader forHTTPHeaderField:SENTRY_BAGGAGE_HEADER];
+            }
+        } else {
+            // Even though NSURLSessionTask doesn't have 'setCurrentRequest', some subclasses
+            // do. For those subclasses we replace the currentRequest with a mutable one with
+            // the additional trace header. Since NSURLSessionTask is a public class and can be
+            // override, we believe this is not considered a private api.
```
armcknight, line 252: This is interesting. What subclasses are we talking about here? I checked all of the public subclasses and none of them have this method: `NSURLSessionDataTask`, `NSURLSessionUploadTask`, `NSURLSessionDownloadTask`, `NSURLSessionStreamTask` and `NSURLSessionWebSocketTask`.

This doesn't have to be addressed in this PR since it's outside scope, but this is my first time seeing this and it piqued my interest.

Sources/Sentry/SentryNetworkTracker.m
```diff
@@ -253,6 +212,65 @@ - (void)urlSessionTaskResume:(NSURLSessionTask *)sessionTask
...
+    [self addBaggageHeader:[[propagationContext traceContext] toBaggage]
+               traceHeader:[propagationContext traceHeader]
+                 toRequest:sessionTask];
+}
+
+- (void)addBaggageHeader:(SentryBaggage *)baggage
+             traceHeader:(SentryTraceHeader *)traceHeader
+               toRequest:(NSURLSessionTask *)sessionTask
+{
+    if ([self sessionTaskRequiresPropagation:sessionTask]) {
```
armcknight, line 227: Would prefer an early return on the inverse condition to bump the much larger piece of code below out one scope indent level:
```suggestion
    if (![self sessionTaskRequiresPropagation:sessionTask]) {
        SENTRY_LOG_DEBUG(@"Not adding trace_id and baggage headers for %@",
             sessionTask.currentRequest.URL.absoluteString);
        return;
    }
    
    NSString *baggageHeader = @"";
    ...
```

Sources/Sentry/SentryScope.m
```diff
@@ -558,6 +563,12 @@ - (SentryEvent *__nullable)applyToEvent:(SentryEvent *)event
             newContext[@"trace"] = [span serialize];
         }
     }
+
+    if (newContext[@"trace"] == nil) {
+        // If this is an error event we need to add the distributed trace context.
```
armcknight, line 568: Thanks for this comment, I'd have no idea what's going on here otherwise 👍🏻

armcknight (commented):
The main thing I don't understand about the feature, which is not to say the implementation is wrong in this PR, is why this still has any effect when network tracing is disabled? I responded to @markushi 's comment at the point where I got lost on this. I tried following through all the way to the [notion doc](https://www.notion.so/sentry/Tracing-without-performance-efab307eb7f64e71a04f09dc72722530) but it's not clear to me what's going on.

Other than that, I just have some small requests on form and maybe another question or two, but nothing blocking right now. I want to think on it for a bit more so I can better understand what's going on, it's the first time I've heard about this initiative.

Sources/Sentry/SentryNetworkTracker.m
brustolin: Apple's own subclasses that are returned from `NSURLSession:startTask`

Sources/Sentry/SentryScope.m
```diff
@@ -71,6 +72,8 @@
 
 @property (nonatomic, retain) NSMutableArray<id<SentryScopeObserver>> *observers;
 
+@property (atomic, strong) SentryPropagationContext *propagationContext;
+
```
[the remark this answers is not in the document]
brustolin: Yes, but properties declared in categories are not synthesized in the class.

Sources/Sentry/SentryPropagationContext.m
brustolin: >Wouldn't it make more sense to not even swizzle -[NSURLSessionTask resume] if network tracking is disabled, so that this isn't even a valid codepath?

Thats the thing, we want this code to run even with network tracking is disabled, we not tracking the network, just adding and extra header to the request to propagate to the server.

Sources/Sentry/SentryNetworkTracker.m
```diff
@@ -115,12 +116,19 @@ - (BOOL)isTargetMatch:(NSURL *)URL withTargets:(NSArray *)targets
...
 - (void)urlSessionTaskResume:(NSURLSessionTask *)sessionTask
 {
-    @synchronized(self) {
-        if (!self.isNetworkTrackingEnabled) {
-            return;
-        }
+    NSURLSessionTaskState sessionState = sessionTask.state;
+    if (sessionState == NSURLSessionTaskStateCompleted
+        || sessionState == NSURLSessionTaskStateCanceling) {
+        return;
```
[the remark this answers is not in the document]
brustolin: The mechanism that change the state of the session is not synchronized by the same lock, so there is no point in synchronizing here yet.

armcknight (approved):
LGTM

PR #3197, opened 2023-08-01
chore: scripts should propagate errors

vaind:
## :scroll: Description

In another PR, I've noticed some scripts don't error-out if there's a failure, e.g. [this test build failure](https://github.com/getsentry/sentry-cocoa/actions/runs/5726804230/job/15518005439#step:11:32).
This makes it harder to find actual cause of a failure down the road.

Checked that there are no CI failures.

## :pencil: Checklist


.github/workflows/test.yml
```diff
@@ -7,25 +7,25 @@ on:
...
-      - 'Tests/**'
-      - 'test-server/**'
-      - 'Samples/**'
-      - '.github/workflows/test.yml'
-      - 'fastlane/**'
-      - 'scripts/tests-with-thread-sanitizer.sh'
-      - 'scripts/ci-select-xcode.sh'
-      - 'scripts/xcode-test.sh'
-      - '.codecov.yml'
+      - "Sources/**"
```
vaind, line 10: Sorry about the quotes - it's auto-formatted by vscode. I can roll this back manually if needed.

scripts/tests-with-thread-sanitizer.sh
```diff
@@ -1,14 +1,11 @@
...
-# When enableThreadSanitizer is enabled and ThreadSanitizer finds an issue, 
+# When enableThreadSanitizer is enabled and ThreadSanitizer finds an issue,
 # the logs only show failing tests, but don't highlight the threading issues.
 # Therefore we print a hint to find the threading issues.
-
 env NSUnbufferedIO=YES xcodebuild -workspace Sentry.xcworkspace -scheme Sentry -configuration Test -enableThreadSanitizer YES \
-    -destination "platform=iOS Simulator,OS=latest,name=iPhone 11" \
-    # The test works when having thread sanitizer not enabled, we have no clue why it fails, and threading issues will be caught
-    # even when it is disabled.
-    -skip-testing:"Sentry/SentrySessionTestsSwift/testInitWithJson_IfJsonContainsWrongValues_SessionIsNil" \
```
brustolin, line 11: l: Is this not an issue anymore?

brustolin (approved):
LGTM

scripts/tests-with-thread-sanitizer.sh
vaind: Not sure if it still causes an issue, the whole thread-sanitizer test fails at the moment (see PR description) and needs to be fixed before enabling. I'm suggesting (in the description above) to disable TSAN build for the time being. Therefore, we will only see if this specific test is a problem when all the other thread-sanitizer errors are fixed.

vaind: I don't have the capacity to do that at the moment. As is on the `main` branch, the thread sanitizer doesn't work anyway so why not just merge the useful parts of this PR, i.e. fixing the scripts, and let the TSAN continue not working until someone has time to fix all the issues?

kahest: I agree, we can merge this, created an issue to fix TSAN: https://github.com/getsentry/sentry-cocoa/issues/3200

brustolin: Ok, so please go ahead and disable the Thread Sanitizer test so we can merge this PR.

vaind: thanks, disabled.

armcknight: The reason the thread sanitizer test was failing was that the destination was invalid. If you look at the [logs](https://github.com/getsentry/sentry-cocoa/actions/runs/5730391353/job/15551242053?pr=3197#step:7:6), xcodebuild warns that it received an invalid destination, shows a list of valid destinations (you can verify that the destination in this diff doesn't appear in the list), and says it's going with the first one. The first one happens to always be a macOS destination, so I'm guessing since the binary was compiled for iOS, it fails to launch.

In the future, please do not disable a check like this along with the rest of an unrelated PR. If there's an issue with an unrelated test, please fix or disable it in a separate PR. This is one of the most critical checks we have in CI, and it's been disabled for a while now.

PR #3202, opened 2023-08-03
fix: Free memory returned by backtrace_symbols() in debug builds

indragiek:
## :scroll: Description

This code is only compiled in debug builds, but we were forgetting to `free` the pointer returned by `backtrace_symbols`. The documentation (https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/backtrace_symbols.3.html)  says:

> It is allocated using malloc() and should be released using free()

Customer reported a memory leak in a call to `backtrace_symbols`, this is the cause.

Verify that the memory is freed and does not leak in a debug build.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Fixes
  
 - Remove linker settings from Package.swift (#3188)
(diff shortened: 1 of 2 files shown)


bruno-garcia (approved):
Checks out, only usage of that is under DEBUG builds.

PR #3206, opened 2023-08-04
feat: Add required reason API

### LINKED ISSUE #3192: New required reason API
### Problem Statement

I was doing scans of the newly marked API from Apple and found that the Sentry SDK makes use of a few.

From Apple's newly published document ["Describing use of required reason API"](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api):

> Your app or third-party SDK must declare one or more approved reasons that accurately reflect your use of each of these APIs and the data derived from their use. You may use these APIs and the data derived from their use for the declared reasons only. These declared reasons must be consistent with your app’s functionality as presented to users, and you may not use the APIs or derived data for tracking.

Will Sentry be either removing the use of these APIs or providing a privacy manifest file for the iOS SDK? Or perhaps are you in contact with Apple around this issue?

### Solution Brainstorm

_No response_

_No response_

brustolin:
## :scroll: Description

Added API reason manifest according to [Apple docs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

close #3192
## :pencil: Checklist


Sentry.xcodeproj/project.pbxproj
```diff
@@ -2005,7 +2005,6 @@
 				844DA80728246D5000E6B62E /* Gemfile */,
 				844DA80428246D5000E6B62E /* Makefile */,
 				844DA81F28246DE300E6B62E /* scripts */,
-				844DA80928246D5000E6B62E /* dangerfile.js */,
```
armcknight, line 2008: Why'd this get removed?

armcknight (approved):
Privacy stuff looks good. I think we need to add the dangerfile.js back to the project.

Sentry.xcodeproj/project.pbxproj
brustolin: This file don't exist. I just removed the reference

armcknight: Oh, LOL 😆

Sentry.xcodeproj/project.pbxproj
```diff
@@ -5219,10 +5230,11 @@
 				GCC_C_LANGUAGE_STANDARD = "compiler-default";
 				GCC_OPTIMIZATION_LEVEL = 0;
 				GCC_WARN_SHADOW = YES;
-				INFOPLIST_FILE = Sources/Sentry/Info.plist;
+				INFOPLIST_FILE = Sources/Resources/Info.plist;
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				LD_DYLIB_INSTALL_NAME = "$(DYLIB_INSTALL_NAME_BASE:standardizepath)/$(EXECUTABLE_PATH)";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				MODULEMAP_FILE = "$(SRCROOT)/Sources/Resources/Sentry.modulemap";
```
armcknight, line 5237: Just hit a build error locally because while this is the correct path, it shouldn't have been updated here. We should've updated the path at https://github.com/getsentry/sentry-cocoa/blob/3437454585c7d7f684776118db34b9018d83941c/Sources/Configuration/Sentry.xcconfig#L7

We really need to get all the build settings out of the pbxproj into xcconfigs and write a linter that makes sure we never add new settings to the pbxproj.

Sentry.xcodeproj/project.pbxproj
```diff
@@ -5219,10 +5230,11 @@
 				GCC_C_LANGUAGE_STANDARD = "compiler-default";
 				GCC_OPTIMIZATION_LEVEL = 0;
 				GCC_WARN_SHADOW = YES;
-				INFOPLIST_FILE = Sources/Sentry/Info.plist;
+				INFOPLIST_FILE = Sources/Resources/Info.plist;
```
armcknight, line 5233: Same thing here as with [here](https://github.com/getsentry/sentry-cocoa/pull/3206/files#r1339232979). With this one, the info plist was moved to a different path, but the old path is still reflected in the file navigator in Xcode.

Sentry.xcodeproj/project.pbxproj
armcknight: fixed in https://github.com/getsentry/sentry-cocoa/pull/3315

PR #3214, opened 2023-08-10
fix: cpu usage readings for profiling

armcknight:
We noticed as we were iterating that we weren't calculating the CPU usage correctly. After fixing it, we realized that this reading doesn't report CPU usage per core only for the process using the API; it shows usage across all processes. We decided to use the API that reports a total CPU usage per thread, to guarantee process confinement, and sum these into a single number. This has a benefit of matching how Android reports this metric.

Visual inspection of the Xcode debugger and matching it to what we see in the flamechart viewer. 

Xcode debugger:
image

From the [profile](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/1f4ef46ffa25455b8921ec1561f83335/flamechart/?colorCoding=by+application+frame&fov=0%2C0%2C30372902912%2C34&query=&sorting=call+order&tid=259&view=top+down):
image

A bit hard to compare due to limitations sizing them to match scale, but the profiler's version basically matches the same trends, except for the one spike shown in the Xcode debugger. Also watched the Xcode "speedometer" with the single CPU % and numbers logged from the profiling code throughout the test, and they basically matched while I changed the amount of CPU work to do in the test app:
image


brustolin (approved):
LGTM

Issue #3063, opened 2023-05-25
Handle underlyingError property from NSError


brustolin:
### Description

We should add the `underlyingError` of an NSError to the event exception list as defined by the [documentation](https://develop.sentry.dev/sdk/event-payloads/exception/).

armcknight:
I have a first pass at a solution in #3230, but i'm not sure if they're being serialized the right way. I don't really understand where they're supposed to go, just reading the docs–it's not clear to me where they should go and there is no example.

philipphofmann:
Fixed with https://github.com/getsentry/sentry-cocoa/pull/3230.

PR #3230, opened 2023-08-17
feat: add underlying error info to reported NSErrors

armcknight:
## :scroll: Description

Refactor SentryNSError to accept just the NSError instance in init, and don't expose properties that aren't used outside of its implementation.

Then, wrap all underlying errors in SentryNSError and serialize them.

https://github.com/getsentry/sentry-cocoa/issues/3063

## :pencil: Checklist


brustolin (commented):
According to the [docs](https://develop.sentry.dev/sdk/event-payloads/exception/#examples) the correct way to report underlying errors is to add them to the event exception array.
The [Exception Interface](https://develop.sentry.dev/sdk/event-payloads/exception/) has no `underlying_errors` property.

armcknight:
> According to the [docs](https://develop.sentry.dev/sdk/event-payloads/exception/#examples) the correct way to report underlying errors is to add them to the event exception array. The [Exception Interface](https://develop.sentry.dev/sdk/event-payloads/exception/) has no `underlying_errors` property.

Thanks for pointing me to this @brustolin! The pertinent part seems to be:

> the exception attribute may be a flat list of objects...Multiple values represent chained exceptions and should be sorted oldest to newest

armcknight:
One problem though @brustolin is that this is for SentryNSError, and you're talking about something pertaining to SentryException; these are different API. I'll look at more docs for errors and see if we have something regarding underlying error reporting.

armcknight:
So I found this [here](https://develop.sentry.dev/sdk/event-payloads/), which I think is why I figured I could get away with adding a key that isn't already specified: 
image

Sources/Sentry/Public/SentryEvent.h
```diff
@@ -138,8 +138,8 @@ NS_SWIFT_NAME(Event)
 @property (nonatomic, strong) NSArray<SentryThread *> *_Nullable threads;
 
 /**
- * General information about the @c SentryException, usually there is only one
- * exception in the array.
+ * General information about the @c SentryException. Multiple exception indicate a chain of
+ * exception encountered, starting with the oldest at the beginning of the array.
```
armcknight, line 142:
```suggestion
 * General information about the @c SentryException. Multiple exceptions indicate a chain of
 * exceptions encountered, starting with the oldest at the beginning of the array.
```

Tests/SentryTests/Protocol/SentryNSErrorTests.swift
```diff
@@ -1,3 +1,4 @@
+import SentryTestUtils
```
armcknight, line 1:
```suggestion
```

Tests/SentryTests/SentryClientTests.swift
```diff
@@ -1522,29 +1531,28 @@ class SentryClientTest: XCTestCase {
...
-        
-        guard let mechanism = exception.mechanism else {
-            XCTFail("Exception doesn't contain a mechanism"); return
-        }
+
+        let mechanism = try XCTUnwrap(exception.mechanism)
+        let meta = try XCTUnwrap(mechanism.meta)
+        let actualError = try XCTUnwrap(meta.error)
         XCTAssertEqual("NSError", mechanism.type)
-        XCTAssertNotNil(mechanism.meta?.error)
```
armcknight, line 1545: This gives us a more granular idea of which thing is missing; It is responsible for all of the `throws` additions in this file.

brustolin (approved):
LGTM

Issue #3212, opened 2023-08-09
[Starfish] Add db attributes to database span's span data


AbhiPrasad:
To get an overview of this work across all SDKs, please visit:
https://github.com/orgs/getsentry/projects/135

Parent issue: https://github.com/getsentry/team-sdks/issues/19

For the starfish db module it was determined that we need extra database attributes.

`db.name` should be set if possible, but is not required. `db.system` should match the list in [OpenTelemetry's well known conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#notes-and-well-known-identifiers-for-dbsystem).

| Attribute  | Type | Description  | Examples  | Requirement Level |
|---|---|---|---|---|
| `db.system` | string | An identifier for the database management system (DBMS) product being used. | `sqlite` | Required |
| `db.name` | string | This attribute is used to report the name of the database being accessed. For commands that switch the database, this should be set to the target database (even if the command fails). | `customers`; `main` | Conditionally Required: If applicable. |

These values should be set on `span.data`.

### Motivation

This is being done for the Sentry's [new performance views for databases](https://sentry.sentry.io/starfish/database/). See a WIP example of the view below:

<img width="1251" alt="image" src="https://github.com/getsentry/sentry-laravel/assets/18689448/962c5fea-1b38-481c-ba22-7046a33ea454">

armcknight:
Hey @AbhiPrasad thanks for opening the issue to track this here. My main question is if we already have a comprehensive list of expected db.system values? We could report this generally as `core-data`, but there are several options for backing stores for a core data deployment: `sqlite`, `file` and `in-memory`.

AbhiPrasad:
For `db.system` OpenTelemetry has a list of well known items here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#notes-and-well-known-identifiers-for-dbsystem. If those aren't enough, we can add more options and and attempt to contribute to the upstream spec!

> For a file-backed database, depending on how this span data will be used, there may be some overlap with the current File I/O on Main Thread performance issue type. We may also want to think about how we might want to get the file path into the span.

This is a great point! I think we need should invest some time into thinking about this considering we're looking at the mobile views right now. 

[...]

PR #3231, opened 2023-08-17
feat: add more db info on span data

armcknight:
## :scroll: Description

Add the persistent store types and URLs to the span data for Core Data tracking.

Also a couple refactors:
- fix a copypasta variable name and log statement that were "fetch" but should've been "save"
- the fetch tracker method grabs a span inside block scope and then the rest of the method proceeds afterwards; but in save tracking, it's all done in the block scope. bump out the save method's logic to match the fetch method's (less indentation -> more code per line).
- rename method to be more descriptive
- moved/combined `SentryCoreDataTrackerTests.TestNSManagedContext` to `TestCoreDataStack.managedObjectContext`

https://github.com/getsentry/sentry-cocoa/issues/3212

### DIFF
--- a/Sources/Sentry/SentryCoreDataTracker.m
+++ b/Sources/Sentry/SentryCoreDataTracker.m
@@ -44,23 +44,18 @@
     }];
 
     if (fetchSpan) {
-        [SentryLog
-            logWithMessage:[NSString stringWithFormat:
-                                         @"SentryCoreDataTracker automatically "
-                                         @"started a new span with description: %@, operation: %@",
-                                     fetchSpan.description, SENTRY_COREDATA_FETCH_OPERATION]
-                  andLevel:kSentryLevelDebug];
+        SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically started a new span with "
+                         @"description: %@, operation: %@",
+            fetchSpan.description, fetchSpan.operation);
     } else {
-        [SentryLog
-            logWithMessage:
-                @"managedObjectContext:executeFetchRequest:error:originalImp: fetchSpan is nil."
-                  andLevel:kSentryLevelError];
+        SENTRY_LOG_ERROR(
+            @"managedObjectContext:executeFetchRequest:error:originalImp: fetchSpan is nil.");
     }
 
     NSArray *result = original(request, error);
 
     if (fetchSpan) {
-        [self mainThreadExtraInfo:fetchSpan];
+        [self addExtraInfoToSpan:fetchSpan withContext:context];
 
         [fetchSpan setDataValue:[NSNumber numberWithInteger:result.count] forKey:@"read_count"];
         [fetchSpan
@@ -78,35 +73,34 @@
                  originalImp:(BOOL(NS_NOESCAPE ^)(NSError **))original
 {
 
-    __block id<SentrySpan> fetchSpan = nil;
+    __block id<SentrySpan> saveSpan = nil;
     if (context.hasChanges) {
         __block NSDictionary<NSString *, NSDictionary *> *operations =
             [self groupEntitiesOperations:context];
 
         [SentrySDK.currentHub.scope useSpan:^(id<SentrySpan> _Nullable span) {
-            fetchSpan = [span startChildWithOperation:SENTRY_COREDATA_SAVE_OPERATION
-                                          description:[self descriptionForOperations:operations
-                                                                           inContext:context]];
-            fetchSpan.origin = SentryTraceOriginAutoDBCoreData;
-            if (fetchSpan) {
-                [SentryLog
-                    logWithMessage:[NSString
-                                       stringWithFormat:@"SentryCoreDataTracker automatically "
-                                                        @"started a new span with description: %@, "
-                                                        @"operation: %@",
-                                       fetchSpan.description, SENTRY_COREDATA_FETCH_OPERATION]
-                          andLevel:kSentryLevelDebug];
-
-                [fetchSpan setDataValue:operations forKey:@"operations"];
-            }
+            saveSpan = [span startChildWithOperation:SENTRY_COREDATA_SAVE_OPERATION
+                                         description:[self descriptionForOperations:operations
+                                                                          inContext:context]];
+            saveSpan.origin = SentryTraceOriginAutoDBCoreData;
         }];
+
+        if (saveSpan) {
+            SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically started a new span with "
+                             @"description: %@, operation: %@",
+                saveSpan.description, saveSpan.operation);
+
+            [saveSpan setDataValue:operations forKey:@"operations"];
+        } else {
+            SENTRY_LOG_ERROR(@"managedObjectContext:save:originalImp: saveSpan is nil");
+        }
     }
 
     BOOL result = original(error);
 
-    if (fetchSpan) {
-        [self mainThreadExtraInfo:fetchSpan];
-        [fetchSpan finishWithStatus:result ? kSentrySpanStatusOk : kSentrySpanStatusInternalError];
+    if (saveSpan) {
+        [self addExtraInfoToSpan:saveSpan withContext:context];
+        [saveSpan finishWithStatus:result ? kSentrySpanStatusOk : kSentrySpanStatusInternalError];
 
         SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically finished span with status: %@",
             result ? @"ok" : @"error");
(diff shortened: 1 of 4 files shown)


Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -115,11 +109,24 @@ - (BOOL)managedObjectContext:(NSManagedObjectContext *)context
...
+        __kindof NSPersistentStore *_Nonnull obj, NSUInteger idx, BOOL *_Nonnull stop) {
+        [systems addObject:obj.type];
+        if (obj.URL != nil) {
+            [names addObject:obj.URL.path];
+        } else {
+            [names addObject:@"(null)"];
+        }
+    }];
+    [span setDataValue:[systems componentsJoinedByString:@";"] forKey:@"db.system"];
+    [span setDataValue:[names componentsJoinedByString:@";"] forKey:@"db.name"];
```
armcknight, line 129: The new `db.name` and `db.system` data.

Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -78,35 +73,34 @@ - (BOOL)managedObjectContext:(NSManagedObjectContext *)context
                  originalImp:(BOOL(NS_NOESCAPE ^)(NSError **))original
 {
 
-    __block id<SentrySpan> fetchSpan = nil;
+    __block id<SentrySpan> saveSpan = nil;
```
armcknight, line 76: Renamed this variable

Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -78,35 +73,34 @@ - (BOOL)managedObjectContext:(NSManagedObjectContext *)context
...
+
+        if (saveSpan) {
+            SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically started a new span with "
+                             @"description: %@, operation: %@",
+                saveSpan.description, saveSpan.operation);
+
+            [saveSpan setDataValue:operations forKey:@"operations"];
+        } else {
+            SENTRY_LOG_ERROR(@"managedObjectContext:save:originalImp: saveSpan is nil");
+        }
```
armcknight, line 96: Moved this out of the `useSpan:` scope, like the code for fetch operations.

Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -78,35 +73,34 @@ - (BOOL)managedObjectContext:(NSManagedObjectContext *)context
...
-                                          description:[self descriptionForOperations:operations
-                                                                           inContext:context]];
-            fetchSpan.origin = SentryTraceOriginAutoDBCoreData;
-            if (fetchSpan) {
-                [SentryLog
-                    logWithMessage:[NSString
-                                       stringWithFormat:@"SentryCoreDataTracker automatically "
-                                                        @"started a new span with description: %@, "
-                                                        @"operation: %@",
-                                       fetchSpan.description, SENTRY_COREDATA_FETCH_OPERATION]
```
armcknight, line 97: This should've been SENTRY_COREDATA_SAVE_OPERATION, but I did one better and just access SentrySpan.operation for both such logs (here and in fetches)

Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -115,11 +109,24 @@ - (BOOL)managedObjectContext:(NSManagedObjectContext *)context
     return result;
 }
 
-- (void)mainThreadExtraInfo:(SentrySpan *)span
+- (void)addExtraInfoToSpan:(SentrySpan *)span withContext:(NSManagedObjectContext *)context
```
armcknight, line 112: Renamed this method as it's doing more than just main thread stuff now.

Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -44,23 +44,18 @@ - (NSArray *)managedObjectContext:(NSManagedObjectContext *)context
...
     if (fetchSpan) {
-        [SentryLog
-            logWithMessage:[NSString stringWithFormat:
-                                         @"SentryCoreDataTracker automatically "
-                                         @"started a new span with description: %@, operation: %@",
-                                     fetchSpan.description, SENTRY_COREDATA_FETCH_OPERATION]
-                  andLevel:kSentryLevelDebug];
+        SENTRY_LOG_DEBUG(@"SentryCoreDataTracker automatically started a new span with "
+                         @"description: %@, operation: %@",
+            fetchSpan.description, fetchSpan.operation);
```
armcknight, line 49: Converted all the logs to the new macros. Also here's the second place I changed interpolating the operation constant for SentrySpan.operation.

Tests/SentryTests/Integrations/Performance/CoreData/SentryCoreDataTrackerTest.swift
```diff
@@ -298,48 +305,36 @@ class SentryCoreDataTrackerTests: XCTestCase {
...
+        assertDataAndFrames(dbSpan: dbSpan, expectedOperation: SENTRY_COREDATA_FETCH_OPERATION, expectedDescription: expectedDescription, mainThread: mainThread)
+    }
+
+    func assertDataAndFrames(dbSpan: Span, expectedOperation: String, expectedDescription: String, mainThread: Bool) {
+        XCTAssertEqual(dbSpan.operation, expectedOperation)
         XCTAssertEqual(dbSpan.spanDescription, expectedDescription)
-        XCTAssertEqual(dbSpan.data["read_count"] as? Int, 1)
         XCTAssertEqual(dbSpan.data["blocked_main_thread"] as? Bool ?? false, mainThread)
+        XCTAssertEqual(try XCTUnwrap(dbSpan.data["db.system"] as? String), "SQLite")
+        XCTAssert(try XCTUnwrap(dbSpan.data["db.name"] as? NSString).contains(TestCoreDataStack.databaseFilename))
```
armcknight, line 337: Assertions for the new properties

Tests/SentryTests/Integrations/Performance/CoreData/TestCoreDataStack.swift
```diff
@@ -99,3 +100,40 @@ class TestCoreDataStack {
...
+    }
+
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+    }
+
+    override var hasChanges: Bool {
+        return ((inserted?.count ?? 0) > 0) ||
+        ((deleted?.count ?? 0) > 0) ||
+        ((updated?.count ?? 0) > 0) || super.hasChanges
```
armcknight, line 137: Had to add `|| super.hasChanges` for SentryCoreDataTrackerTests.test_Save() to work again, because it doesn't rely on the subclass properties.

Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -115,11 +109,24 @@ - (BOOL)managedObjectContext:(NSManagedObjectContext *)context
...
     [span setDataValue:@(isMainThread) forKey:BLOCKED_MAIN_THREAD];
+    NSMutableArray<NSString *> *systems = [NSMutableArray<NSString *> array];
+    NSMutableArray<NSString *> *names = [NSMutableArray<NSString *> array];
+    [context.persistentStoreCoordinator.persistentStores enumerateObjectsUsingBlock:^(
+        __kindof NSPersistentStore *_Nonnull obj, NSUInteger idx, BOOL *_Nonnull stop) {
+        [systems addObject:obj.type];
+        if (obj.URL != nil) {
+            [names addObject:obj.URL.path];
+        } else {
+            [names addObject:@"(null)"];
```
brustolin, line 125: h: I believe we need a better option to fallback. This is not user friendly and I'm foreseeing complains.

armcknight: I'm open to any alternatives. Nobody has proposed any so far.

armcknight: I would also be interested to know the cases where the URL is nil. I expected it to be nil for in memory stores, but even that has a value like `memory://563428497`. It is probably nil if there is no persistent store/coordinator at all, but then I don't know what the complaint would be.

brustolin: > I'm open to any alternatives.

`<App Name>:coredata` anything really other than (null)

armcknight: We already gather the equivalent of that in the `type`. 

In this case, we should send some kind of sentinel value so the frontend knows to render a special state equivalent to "there was no database". This seems as good as any.

kahest: I agree with @armcknight here - in case this occurs a lot we can always re-iterate.

kahest (approved):
LGTM! Thanks for the opportunistic refactoring as well 💯

Tests/SentryTests/Integrations/Performance/CoreData/SentryCoreDataTrackerTest.swift
armcknight: good catch, i removed those by accident

PR #3235, opened 2023-08-18
ref: report viewDidAppear breadcrumbs via delegate

armcknight:
I noticed this while working on the breadcrumb logic. Looks like one avenue of reporting a breadcrumb via the newer delegate method was missed.

#skip-changelog


brustolin (commented):
The changes look right, but CI is not happy. Maybe some other tests  expecting something else.

brustolin (approved):
LGTM

PR #3236, opened 2023-08-18
ref: convert SentryHub (Private) category to an interface extension

armcknight:
Converts the Private category to an interface extension (aka an anonymous category).

This avoids having to redeclare properties in the "conforming" implementation. Categories should be reserved for interfaces that we don't own.

I moved the uninstallation of integrations into the hub method that used to be called right afterwards at the sole callsite.

This also caught an encapsulation logic error: the `installedIntegrations` and `installedIntegrationNames` collections were defined as `strong` properties, not `copy`, so they would've handed out the actual mutable underlying collections, which could have been modified if casted to their mutable counterparts. I converted those declarations to methods instead in the test interface extension (renamed from "TestInit" to just "Test").

Also, because I converted the import of SentryTracer to a forward class declaration, I had to import SentrySpan in SentryCoreDataTracker. This is cleaner as it reduces coupling.

Also removed declaration of a nonexistent method `setSampleRandomValue`.

#skip-changelog


brustolin (approved):
LGTM

PR #3237, opened 2023-08-18
ref: convert SentryScope (Private) category to an interface extension

armcknight:
Interface extensions > private categories (see #3236 for elaboration)

#skip-changelog


brustolin (approved):
LGTM

PR #3239, opened 2023-08-18
ref: convert simple private category to interface extensions

armcknight:
Interface extensions > private categories (see https://github.com/getsentry/sentry-cocoa/pull/3236 for elaboration)

This PR collects all such changes that didn't involve any other code changes other than removing the category name to make it anonymous.

#skip-changelog


brustolin (approved):
LGTM

PR #3240, opened 2023-08-18
ref: move decl/impl from SentryEvent (MetricKit) to SenryEvent .h/.m …

armcknight:
Another conversion of a category to interface extension (see https://github.com/getsentry/sentry-cocoa/pull/3236 for elaboration)

This also moved the extension interface and implementation to SentryEvent instead of being hidden in SentryMetricKitIntegration. It also declared isMetricKitEvent as a method since that's what it actually is, and converted callsites to use method syntax.

#skip-changelog


brustolin (approved):
LGTM

PR #3241, opened 2023-08-19
ref: centralize SentryBinaryImageCache singleton in SentryDependencyContainer

armcknight:
This was a singleton, but then we'd pass the reference around in various constructors and properties. Just centralize it in SentryDependencyContainer.

#skip-changelog


brustolin (approved):
LGTM

PR #3244, opened 2023-08-19
ref: move SentryExtraContextProvider singleton to SentryDependencyContainer

armcknight:
Another singleton that was passed in a few inits, that should be just accessed via the dependency container.

#skip-changelog


brustolin:
Something broke CI

brustolin (approved):
LGTM

PR #3245, opened 2023-08-19
ref: access SentrySwizzleWrapper via existing SentryDependencyContainer property

armcknight:
We already had a property on SentryDependencyContainer for this, but that still accessed SentrySwizzleWrapper.sharedInstance and we still also passed around the reference.

#skip-changelog


brustolin (approved):
lgtm

PR #3246, opened 2023-08-19
ref: use preexisting SentryDependencyContainer.notificationCenterWrapper

armcknight:
We already had a centralized notification center wrapper. No need to pass around references.

#skip-changelog


armcknight:
The tests are failing because of a crash, and the reports I looked at involve both `SentryFileManager` and `Invocations`.

SentryTestUtils/Invocations.swift
```diff
@@ -42,8 +42,8 @@ public class Invocations<T> {
     }
     
     public func record(_ invocation: T) {
-        queue.async(flags: .barrier) {
-            self._invocations.append(invocation)
+        queue.async(flags: .barrier) { [weak self] in
```
philipphofmann, line 45: `m`: That seems like a fix that could be in an extra PR, so the whole test suite immediately benefits from this improvement.

Sources/Sentry/SentryAppStateManager.m
```diff
@@ -53,21 +50,23 @@ - (instancetype)initWithOptions:(SentryOptions *)options
 - (void)start
 {
     if (self.startCount == 0) {
-        [self.notificationCenterWrapper
+        SentryNSNotificationCenterWrapper *notificationCenterWrapper
```
philipphofmann, line 53: `m`: I'm not a real fan of this change. I prefer passing in dependencies, so you know what a class requires to work correctly, and it increases testability. Now I have no clue when using this class that it needs the `SentryNSNotificationCenterWrapper`, and I need to modify the dependencyContainer if I want to use the TestNotificationCenterWrapper.

armcknight: FWIW I agree with you. If you recall some time ago this is why I actually wanted to move away from SentryDependencyContainer and do full dependency injection: https://github.com/getsentry/sentry-cocoa/pull/2942#discussion_r1174075139

After that I figured we should move things to go to the singleton, and after ironing out the inconsistencies, if we decided we wanted DI later, it'd be straigthforward to change (even if it would be some effort, of course). 

Regardless of which approach we take, I'd prefer we be consistent. If I never know what to expect in a given class, or if there is a mixture of approaches in a given codepath, it makes it harder to reason about.

armcknight: Rereading that thread, I'm a little confused why you say the client can't use DI here: https://github.com/getsentry/sentry-cocoa/pull/2942#discussion_r1175314869

armcknight: In either case, regarding tests, you're either reading the list of dependencies and mocks in a constructor, or in test init where all the mocks are set on SDC.

philipphofmann: DI is a bit tricky for classes our users interact with. We can't ask our users for all the dependencies the client requires when initializing it cause that would be a nightmare for usability, and most classes required are internal. For internal classes, though, we can pass in all dependencies via the constructor. That way we would be consistent.

armcknight: OK, that makes sense to me I think. Although we could take a page out of Apple's book and provide a method `defaultFileManager`, `defaultHttpTransport` etc. It shouldn't be a usability nightmare for _everyone_ because most people should not be initializing their own instances of `SentryClient`, they should just use `SentrySDK.start...`

philipphofmann: But there are plenty of dependencies for the SentryClient; see https://github.com/getsentry/sentry-cocoa/blob/e0f077c32b92169e5d4ad45e0c0a4ae6bbc2a74b/Sources/Sentry/SentryClient.m#L132-L139

Passing in all dependencies is not user-friendly. I don't want to make FileManager, HttpTransport etc, public. I don't want users to know that these classes exist, and making them public reduces the flexibility.

I would like to stick to passing in all dependencies for internal classes and using the dependency container for public classes, to provide dependencies for our users.

armcknight: OK, I think I actually had it backwards before. A strategy of DI for internal interfaces and singletons for public interface dependencies sounds good, and easy to keep consistent 👍🏻

philipphofmann: I opened a PR to our decision log so this info doesn't get lost: https://github.com/getsentry/sentry-cocoa/pull/3396.

PR #3247, opened 2023-08-19
ref: move SentryCrash singleton to SentryDependencyContainer

armcknight:
Same deal, move the singleton to SentryDependencyContainer.

After making this change, SentryCrashInstallationTests.testUninstall fails when running the full test suite locally, but not when run by itself. Seems like there's a bug in the tests. Haven't sussed it out yet though.

#skip-changelog


brustolin:
Please fix CI.

brustolin:
We need CI fixed.

brustolin (approved):
LGTM.

Just make sure CI is working properly before merging this.

armcknight:
Still not sure why the test is failing for tvOS to test uninstalling the crash handler. But, I was unable to even run it locally, which #3382 should fix. It could be related to dirty state between tests, which #3381 attempts to fix more.

philipphofmann (approved):
Thanks @armcknight 👍

armcknight:
Found it: SentryCrashInstallationReporterTests never uninstalled the crash reporters it installed.

PR #3258, opened 2023-08-28
ci: always run setup_ci

armcknight:
For some reason if these jobs are both run on ci sequentially, the second one fails in the match phase. I think setup_ci lets the previous run's keychain get used for the subsequent run's, and then something fails. In any case, these are the only jobs that don't call this with force: true.

This is also failing nondeterminstically; sometimes it works on PRs, sometimes not.

#skip-changelog

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -263,8 +263,9 @@ platform :ios do
 
   desc "Build Perf-test app without Sentry"
   lane :build_perf_test_app_plain do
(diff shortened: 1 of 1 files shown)


armcknight:
The issue this PR sets out to solve is fixed, as the test apps build now. There is a downstream issue with how the test is actually run.

brustolin (approved):
LGTM

PR #3259, opened 2023-08-29
feat: Add data use in privacy manifests

brustolin:
## :scroll: Description

Added "data use" in privacy manifests.

Apple requires this information from Apps and Frameworks


brustolin:
Im wondering whether it make sense to ship this manifest with the SDK. Each feature may be disabled by the user and therefore it should be the developer responsibility to add in its manifest which features they use.

armcknight:
> Im wondering whether it make sense to ship this manifest with the SDK. Each feature may be disabled by the user and therefore it should be the developer responsibility to add in its manifest which features they use.

It's a tough question, because either way there are risks affecting developers who either don't know about the new requirements and/or don't know we deliver functionality that requires it.

If we don't ship it, we risk customers' apps being rejected and not understanding why, and then later finding out it is our "fault".

If we do ship it, then developers' users may wonder why an app requires "unnecessary" functionality, and if the developers aren't aware of what's going on, they're going to wonder what we're doing to their app.

I feel like the former is better, that we put a hard stop at the developers, and not let the issue trickle down to users, because that's going to cause a lot more confusion.

This is also another argument in favor of modularizing the SDK, to isolate components that would require these declarations, so our customers can still use some parts of Sentry even if they can't/won't use these.

brustolin:
Adding to what @armcknight said. All of the information on the manifest is regarding capturing crash and performance data. It's extremely unlikely for a user to choose to use Sentry and disable this things.

I believe it's ok to ship the manifest.

kahest:
It's also highly likely that app devs need to list more data usages for their apps, so they should be aware of how this works and what the requirements are. We should also follow up with better guidance on the [Data Privacy page](https://docs.sentry.io/product/security/mobile-privacy/) about what kinds of data usage they need to enter if they use certain data/features the SDK provides, see https://github.com/getsentry/sentry-docs/issues/5724 and https://github.com/getsentry/rfcs/pull/70

armcknight (approved):

PR #3272, opened 2023-09-08
fix: gather metric samples at profile start/end

armcknight:
## :scroll: Description

Gather a metric sample right at the beginning and end of a profile.

We previously gathered metrics samples at 100 Hz, but not until the first 100 ms of a profile had elapsed, and there could've been up to 99ms without a sample at the end of a profile. 
Unit test update.


brustolin (approved):
LGTM

PR #3273, opened 2023-09-09
fix: ensure GPU frame tracker always runs for profiles

armcknight:
## :scroll: Description

Always start the frames tracker when starting the profiler, and stop it when stopping the profiler if `enableAutoPerformanceTracing` is `NO`.

I noticed after reading through the frame tracking integration and options that if `enableAutoPerformanceTracing` is `NO`, the frames tracker doesn't run and we don't get GPU frame data for profiles.


Sources/Sentry/SentryProfiler.mm
```diff
@@ -458,6 +466,14 @@ - (void)stopForReason:(SentryProfilerTruncationReason)reason
         return;
     }
 
+#    if SENTRY_HAS_UIKIT
+    // if SentryOptions.enableAutoPerformanceTracing is NO, then we need to stop the frames tracker
+    // from running outside of profiles because it isn't needed for anything else
+    if (![[[[SentrySDK currentHub] getClient] options] enableAutoPerformanceTracing]) {
```
brustolin, line 472: l: We should find a better way to get the options sentry was initialized with. 
But this is a thing for another PR.

brustolin (approved):
LGTM

PR #3282, opened 2023-09-15
fix: decouple dual-purpose profiling IDs

armcknight:
one was used to profile concurrent spans in the client, and that was reused for the ID needed to identify profiles in the backend, and we wound up in a situation where one profile ID used for multiple concurrent spans was reused for each of their profiles


Sources/Sentry/SentryProfiler.mm
```diff
@@ -211,7 +211,7 @@
         @"model" : isEmulated ? sentry_getSimulatorDeviceModel() : sentry_getDeviceModel()
     };
 
-    payload[@"profile_id"] = profileID.sentryIdString;
```
indragiek, line 214: 💯

indragiek (approved):

PR #3283, opened 2023-09-15
ci: update unit test iOS/Xcode versions

armcknight:
## :scroll: Description

I noticed our 'latest iOS test' job [failed](https://github.com/getsentry/sentry-cocoa/actions/runs/6128877602/job/16636452342?pr=3273#step:11:97) for a reason that indicated UIKit wasn't present, and when i fixed that build for other macos builds, it passed but produced [this warning](https://github.com/getsentry/sentry-cocoa/actions/runs/6203483094/job/16844260477?pr=3273#step:12:27): `xcodebuild: WARNING: Using the first of multiple matching destinations:`, and the first destination is a macOS one. 

Looks like maybe the macOS image we're using removed the iPhone 8 at some point. 

This adds a new pinned job for 16.4 and "latest" now will test 17.0 (as those are the two versions mentioned in the xcodebuild logs) and then use iPhone 14 for later iOS versions and iPhone 8 for older versions, with a new test script parameter.

#skip-changelog


.github/workflows/test.yml
```diff
@@ -71,7 +71,7 @@ jobs:
   unit-tests:
     name: Unit ${{matrix.platform}} - Xcode ${{matrix.xcode}} - OS ${{matrix.test-destination-os}}
     runs-on: ${{matrix.runs-on}}
-    timeout-minutes: 20
+    timeout-minutes: 30 # something is taking much longer for this job, maybe the beta simulator is slow to launch (armcknight 19 Sept 2023)
```
armcknight, line 74:
```suggestion
    timeout-minutes: 40 # something is taking much longer for this job, maybe the beta simulator is slow to launch (armcknight 19 Sept 2023)
```

.github/workflows/test.yml
```diff
@@ -71,7 +71,7 @@ jobs:
   unit-tests:
     name: Unit ${{matrix.platform}} - Xcode ${{matrix.xcode}} - OS ${{matrix.test-destination-os}}
     runs-on: ${{matrix.runs-on}}
-    timeout-minutes: 20
+    timeout-minutes: 40 # something is taking much longer for this job, maybe the beta simulator is slow to launch (armcknight 19 Sept 2023)
```
philipphofmann, line 74: `h`: If this change requires us to increase this value to `40`, I would like to find out why. This looks like it will slow down CI significantly. Maybe using different devices fixes the problem.

armcknight: Good idea, i'll try a different device. This is probably why the steps to prepare various other simulators were originally added. Looks like the tests run fine once they do get going.

armcknight: Looks like this fixed the issue. Now to unflake the tests with a couple reruns!

armcknight: Actually i was wrong, looking at the other test runs it should only take 4-5 minutes. I just saw this in the Xcode 15.1 beta release notes: 
> Fixed: Executing Unit/UI tests from Xcode on the iOS Simulator takes an extended time to launch on first run of the suite. (115187363) (110330776)
So we can wait for that or we can try a similar approach as we do with the other earlier simulators: https://github.com/getsentry/sentry-cocoa/blob/7fa1b2f9254f7fc4336496140cf6363c5d3db02c/.github/workflows/test.yml#L159-L165
but i don't know if that's the same problem as this one.

philipphofmann: > Fixed: Executing Unit/UI tests from Xcode on the iOS Simulator takes an extended time to launch on first run of the suite. (115187363) (110330776)

Good catch 👏, @armcknight. I would be OK to increase the timeout to 40m and reduce it once Xcode 15.1 is out, if we add a comment to the 40m timeout explaining why we did it. It is not worth trying a workaround if the problem should be fixed soon.

philipphofmann (approved):
LGTM

armcknight:
I'm going to decouple this to just update the existing ones, and we'll push off the iOS 17 addition to a later PR since it's the one causing issues.

.github/workflows/test.yml
armcknight: I'm actually going to decouple this to just update the existing ones, and we'll push off the iOS 17 addition to a later PR since it's the one causing issues. I need these fixes now for another PR.

armcknight: https://github.com/getsentry/sentry-cocoa/issues/3329

PR #3284, opened 2023-09-15
fix: remove warning that spammed logs when called from display link callbacks

armcknight:
This method is now called from every display link callback and if there's no profiler running it emits large amounts of logs.


brustolin (approved):
LGTM

PR #3286, opened 2023-09-18
feat: Add Sampling Decision to Trace Envelope Header

### LINKED ISSUE #3201: Add Sampling Decision to Trace Envelope Header
### Description

See project description: https://github.com/getsentry/team-sdks/issues/8

brustolin:
## :scroll: Description

Adds `sampled` flag to all trace envelope headers

closes #3201

Unit tests

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/Public/SentryDefines.h
+++ b/Sources/Sentry/Public/SentryDefines.h
@@ -118,8 +118,8 @@ static DEPRECATED_MSG_ATTRIBUTE(
 
 static NSUInteger const defaultMaxBreadcrumbs = 100;
 
-static NSString * _Nonnull const kSentryTrueString = @"true";
-static NSString * _Nonnull const kSentryFalseString = @"false";
+static NSString *_Nonnull const kSentryTrueString = @"true";
+static NSString *_Nonnull const kSentryFalseString = @"false";
 
 /**
  * Transaction name source.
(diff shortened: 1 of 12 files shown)


Sources/Sentry/include/SentryTraceContext.h
```diff
@@ -52,6 +52,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nullable, nonatomic, strong) NSString *sampleRate;
 
+/**
+ * Value indicating whether the trace was sampled.
+ */
+@property (nullable, nonatomic, strong) NSString *sampled;
```
armcknight, line 58: These properties should probably be `readonly` and/or `copy` if they're only settable via the init method

armcknight (approved):
Changes look fine. Though I wonder if an `NSNumber` encapsulating a boolean wouldn't be more appropriate than a string. 

If we want to leave it as a string, we should consider using defined constants for the values.

Sources/Sentry/Public/SentryDefines.h
```diff
@@ -22,6 +22,9 @@
     -(instancetype)init NS_UNAVAILABLE;                                                            \
     +(instancetype) new NS_UNAVAILABLE;
 
+#define SENTRY_TRUE_STRING @"true"
+#define SENTRY_FALSE_STRING @"false"
```
brustolin, line 26: @armcknight Do you think a `#define` is fine or a `const` would be better?

I will create another PR later to replace every instance of "true" and "false"

armcknight: I prefer `static NSString *const` for these over macros, but it's not a huge deal.

armcknight (approved):
Updates look good, thanks!

PR #3290, opened 2023-09-19
fix: App hang with race condition for tick counter

### LINKED ISSUE #3274: ticksSinceUiUpdate is not thread safe
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.11.0

### Steps to Reproduce

0) Install Xcode 15b8
1) Install Sentry
2) Enable ThreadSanitizer
3) Configure Sentry in code as usual
4) Run app
5) Get stop on `Data race in -[SentryANRTracker detectANRs] `

### Expected Result

Without stop on Data racing

### Actual Result

Stopped on Data racing

Снимок экрана 2023-09-11 в 08 59 15

_No response_

brustolin:
## :scroll: Description

Thread sanitizer was complaining about a race condition in the ANR tracker.
Added atomic operations to avoid the race condition

closes #3274 

Instruments

## :pencil: Checklist

- [x]  I added tests to verify the changes.


armcknight:
I'm trying to get the thread sanitizer CI job to run.

armcknight:
From slack convo, the thread sanitizer test has been disabled at some point in the past. I think we need to fix that before merging this.

armcknight:
Realized I have such a task already assigned to me. see #3303

philipphofmann (approved):
LGTM, but the changelog needs fixes. I don't think we need the thread sanitizer job fix merged before merging this. If `SentryANRTrackerTests` are green, we should be good to go. The https://github.com/getsentry/sentry-cocoa/pull/3303 will be a mitigation for future data races. WDYT, @armcknight and @brustolin.

armcknight:
I don't think green ANR tracker tests are enough, they've been green all along. Without changes to them here to somehow ensure the data race is fixed, I don't see how we're validating the fix. I would take an exported Instruments trace here.

brustolin:
>  I don't see how we're validating the fix

In this case I had to validate manually, by running the sample multiple times comparing the outcome. Without a CI is hard to prevent regression, but as far this PR goes, the problem is fixed.

PR #3294, opened 2023-09-20
ref: Reduce complexity for SentryDependencyContainer

brustolin:
Making `SentryDependencyContainer` less complex and more predictable by removing some unnecessary lazy loads.

The goal would be to have all necessary objects initialized according to SentryOptions and remove all the @syncronized locks.

_#skip-changelog_

### DIFF
--- a/Sources/Sentry/SentryDependencyContainer.m
+++ b/Sources/Sentry/SentryDependencyContainer.m
@@ -58,7 +58,8 @@ static NSObject *sentryDependencyContainerLock;
     instance = [[SentryDependencyContainer alloc] init];
 }
 
-- (instancetype)init {
+- (instancetype)init
+{
     if (self = [super init]) {
         _dispatchQueueWrapper = [[SentryDispatchQueueWrapper alloc] init];
         _random = [[SentryRandom alloc] init];
--- a/Sources/Sentry/SentryDependencyContainer.m
+++ b/Sources/Sentry/SentryDependencyContainer.m
@@ -44,24 +44,30 @@ static NSObject *sentryDependencyContainerLock;
 {
     if (self == [SentryDependencyContainer class]) {
         sentryDependencyContainerLock = [[NSObject alloc] init];
+        instance = [[SentryDependencyContainer alloc] init];
     }
 }
 
 + (instancetype)sharedInstance
 {
-    @synchronized(sentryDependencyContainerLock) {
-        if (instance == nil) {
-            instance = [[self alloc] init];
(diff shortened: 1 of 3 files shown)


Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -44,24 +44,30 @@ + (void)initialize
 {
     if (self == [SentryDependencyContainer class]) {
         sentryDependencyContainerLock = [[NSObject alloc] init];
+        instance = [[SentryDependencyContainer alloc] init];
```
brustolin, line 47: The next step would be to initialize the `instance` when SentrySDK is started. This approach allows us to pass the options and eliminate any dependencies on `SentryDependencyContainer` from both the hub and client, resolving the unusual cyclic dependency.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -44,24 +44,30 @@ + (void)initialize
...
+}
+
+- (instancetype)init {
+    if (self = [super init]) {
+        _dispatchQueueWrapper = [[SentryDispatchQueueWrapper alloc] init];
+        _random = [[SentryRandom alloc] init];
+        _threadWrapper = [[SentryThreadWrapper alloc] init];
+        _binaryImageCache = [[SentryBinaryImageCache alloc] init];
+        _debugImageProvider = [[SentryDebugImageProvider alloc] init];
+        _dateProvider = [[SentryCurrentDateProvider alloc] init];
```
brustolin, line 69: These objects are either always required or too small to justify having an "double check/lock" path for them.

brustolin:
Hey @armcknight.
What do you think about this PR, does it make sense?

I will leave it as draft so we can discuss about it.

Sources/Sentry/SentryDependencyContainer.m
armcknight: Yes, I agree with that approach 👍🏻

brustolin:
I believe we should iterate over this ideas slowly.

I suggest that we merge this first changes, it will reduce complexity on some main objects, and later we improve to initialize all objects during `init`.

armcknight (approved):
Agreed on the iterative approach. Ship it!

Sources/Sentry/SentryDependencyContainer.m
philipphofmann: Thanks for the improvement, @brustolin.

PR #3298, opened 2023-09-25
Fix: App hang being reported from background

### LINKED ISSUE #3287: Increased 'App Hanging' Sentry Reports with Background Updates
### Platform

iOS

### Installed

CocoaPods

### Version

8.6.0

#### Background
We recently added "background update" support to our app based on [Apple's docs](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/pushing_background_updates_to_your_app).  Most notably, we add `content-available` to our push notification payload and then handle the [application(_:didReceiveRemoteNotification:fetchCompletionHandler:)](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application) delegate function, which iOS will call when the app is in the background.

#### Observation
Since releasing this feature we've seen a huge rise in "app hanging" sentry reports:
image

#### Details
But all of our background update code runs asynchronously off of the main queue, and the callstacks in the sentry events are widespread and mostly not in our app code at all.  Here's an [example callstack](https://gist.github.com/mikelehen/c7d3e35efba24bf30379517701a37da0) with 607 hits (~28% of the total hits from the chart above). The entire callstack except for the root main() function is iOS internal code, not our app code.  It's also worth noting that the "background update code" we're running (to fetch some network data and persist it to disk) is the same code we run routinely when the app is in the foreground, and it doesn't seem to trigger "app hang" reports in that case.

#### Analysis
My guess is what's happening is that the app is being CPU throttled since it's running in the background, and this causes the queue to be blocked for >2s even though it's doing "normal" things.  Another possibility is that iOS is scheduling work on the main queue more aggressively since the app is in the background and the UI is not expected to be interactive (total speculation).

The net result is we're getting tons of app hang reports via sentry, but not a single user complaint. And I expect these are all "red herrings."  I'm wondering if this is a known issue and/or if there is any suggested workaround.

#### Suggestion:
Naively, maybe the app hang detection logic should be conditioned on `UIApplication.shared.applicationState == UIApplication.State.active` since it doesn't matter if the app is not very responsive when it is running in the background.

### Expected Result

App hangs should only be reported when the app is in the foreground?

### Actual Result

Lots of app hangs reported when running in the background.

Maybe?  If it's as simple as conditioning on `UIApplication.shared.applicationState == UIApplication.State.active` then I could try.

brustolin:
## :scroll: Description

Some Apps may run in the background after receiving a notification or as part of background fetch.
We should not report app hang in this cases.

#close #3287 

Unit test

## :pencil: Checklist


Samples/iOS-Swift/iOS-Swift/Info.plist
```diff
@@ -28,6 +28,10 @@
 	<string>Testing location permissions</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>Testing location permissions</string>
+	<key>UIBackgroundModes</key>
+	<array>
+		<string>fetch</string>
+	</array>
```
brustolin, line 34:
```suggestion
```

Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-Swift.xcscheme
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
    LastUpgradeVersion = "1400"
-   version = "1.3">
+   version = "2.0">
```
brustolin, line 4:
```suggestion
   version = "1.3">
```

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
```diff
@@ -288,6 +288,7 @@ class MockApplication: NSObject, SentryUIApplicationProtocol {
     init(_ delegate: UIApplicationDelegate?) {
         self.delegate = delegate
     }
+
```
brustolin, line 291:
```suggestion
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- App hang being reported from background (#3298)
```
philipphofmann, line 7: `l`: Active voice is usually better
```suggestion
- Reporting app hangs reported from background (#3298)
```

philipphofmann (commented):
The approach LGTM. What's missing to merge this, @brustolin?

philipphofmann:
Okay, then ping me for a review once you're ready, @brustolin.

Tests/SentryTests/Integrations/ANR/SentryANRTrackingIntegrationTests.swift
```diff
@@ -108,7 +108,23 @@ class SentryANRTrackingIntegrationTests: SentrySDKIntegrationTestsBase {
...
+#if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
+    func testANRDetected_ButBackground_EventNotCaptured() {
+
+        class BackgroundSentryUIApplication: SentryUIApplication {
+            override var applicationState: UIApplication.State { .background }
+        }
+
+        givenInitializedTracker()
+        setUpThreadInspector()
+        //SentryDependencyContainer.sharedInstance().application = BackgroundSentryUIApplication()
```
brustolin, line 120:
```suggestion
        SentryDependencyContainer.sharedInstance().application = BackgroundSentryUIApplication()
```

Tests/SentryTests/Integrations/ANR/SentryANRTrackingIntegrationTests.swift
```diff
@@ -108,7 +108,23 @@ class SentryANRTrackingIntegrationTests: SentrySDKIntegrationTestsBase {
         
         assertNoEventCaptured()
     }
-    
+#if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
+    func testANRDetected_ButBackground_EventNotCaptured() {
+
+        class BackgroundSentryUIApplication: SentryUIApplication {
+            override var applicationState: UIApplication.State { .background }
+        }
```
philipphofmann, line 116: Excellent defining the class directly in here. This improves the test readability a lot 🥇

philipphofmann (approved):
Ship it 🚢

PR #3303, opened 2023-09-25
ci: fix thread sanitizer job

armcknight:
fixes #3200 and #3069

Fixed the races that were detected locally. Here are the outputs from each: 
[data race in ANR tracker.txt](https://github.com/getsentry/sentry-cocoa/files/12969522/data.race.in.ANR.tracker.txt)
[data race in SentryNetworkTracker.isCaptureFailedRequestsEnabled.txt](https://github.com/getsentry/sentry-cocoa/files/12969523/data.race.in.SentryNetworkTracker.isCaptureFailedRequestsEnabled.txt)
[data race in SentrySpan.isFinished.txt](https://github.com/getsentry/sentry-cocoa/files/12969524/data.race.in.SentrySpan.isFinished.txt)
[data race in SentryThreadInspectorTests.txt](https://github.com/getsentry/sentry-cocoa/files/12969525/data.race.in.SentryThreadInspectorTests.txt)
[data race in TestThread (1).txt](https://github.com/getsentry/sentry-cocoa/files/12969526/data.race.in.TestThread.1.txt)
[data race in TestThread (2).txt](https://github.com/getsentry/sentry-cocoa/files/12969527/data.race.in.TestThread.2.txt)
data race in binaryImageAdded
data race in SentryFileManager currentFileCounter

# TODO

- [ ] i just discovered the [suppressions](https://github.com/getsentry/sentry-cocoa/blame/98cca71ab965606c68f46b2f6a0a9c710ce6cdb2/Tests/ThreadSanitizer.sup) file listing some skipped tests. Let's work on trying to fix those in a subsequent PR.

#skip-changelog

PR #3318, opened 2023-09-29
ref: use xcode-test.sh to run sanitizers

armcknight:
#skip-changelog


philipphofmann (commented):
This looks great. I like the idea of removing the extra script. Once it works, LGTM.

armcknight:
Going to close this for now, because if the method of running tests diverges too much given our plan from #3303

PR #3324, opened 2023-10-03
meta: update ruby version and gem/brew bundles

armcknight:
I just set up a new machine so had to `make init` again, and noticed two things:

1. the gem bundle and brew bundle had updates, not surprising
2. got this warning about our ruby version (2.7.5):
> WARNING: ruby-2.7.5 is past its end of life and is now unsupported.
It no longer receives bug fixes or critical security updates.

So I figured I'd submit an update to that. Let me know if you have concerns about impact on any of our infra!

#skip-changelog


brustolin (approved):
CI is passing for build, I'm fine with this.

philipphofmann (approved):
Thanks @armcknight

PR #3327, opened 2023-10-04
ref: Replace deprecated function

### LINKED ISSUE #3317: Update os version retrieval
### Description

We are still using the gestalt api which is deprecated: https://github.com/getsentry/sentry-cocoa/blob/e32423038e99f0c5d868744421d664cdd55c7383/Sources/Sentry/SentryDevice.mm#L195-L200

recommend to upgrade to https://developer.apple.com/documentation/foundation/processinfo/1410906-operatingsystemversion which is now supported as of our minimum versions for all platforms: https://github.com/getsentry/sentry-cocoa/blob/e32423038e99f0c5d868744421d664cdd55c7383/Sentry.podspec#L11-L14
image
https://developer.apple.com/documentation/foundation/processinfo/1410906-operatingsystemversion

brustolin:
Replaced macOS OS version retrieval from a deprecated API to a new one.

closes #3317

_#skip-changelog_

### DIFF
--- a/Sources/Sentry/SentryDevice.mm
+++ b/Sources/Sentry/SentryDevice.mm
@@ -185,7 +185,7 @@ sentry_getOSVersion(void)
 #else
     const auto version = [[NSProcessInfo processInfo] operatingSystemVersion];
     return [NSString stringWithFormat:@"%ld.%ld.%ld", (long)version.majorVersion,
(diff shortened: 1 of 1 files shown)


philipphofmann (approved):
LGTM, nice 👏

PR #3331, opened 2023-10-06
Issue with relative path in import blocking me from integrating Sentry using Buck

jboulter11:
## :scroll: Description

Remove relative path to `SentryId.h` in import in the file `SentryProfiledTracerConcurrency.h`

## :bulb: Motivation and Context

This relative path in this particular import is causing our previously working buck builds to fail, and other files in the same folder seem fine to omit the relative path before the header name.
Screenshot of error stating ../Public/SentryId.h file not found

## :green_heart: How did you test it?

I'm hoping your CI will do that for me, so I don't have to mess with my work computer's dev environment to make this change.

## :pencil: Checklist

You have to check all boxes before merging:

- [ ] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #3332, opened 2023-10-06
fix: remove relative component of import

armcknight:
Reported with a proposed fix in https://github.com/getsentry/sentry-cocoa/pull/3331, but that can't run all the necessary CI checks

#skip-changelog


armcknight:
@vaind Would you mind reviewing this one? I'm not sure why the import was originally written that way. Is this going to break the Dart SDK somehow?

jboulter11:
Thank you very much for getting this taken care of for me. 😄

Issue #3320, opened 2023-10-02
EXC_BAD_ACCESS in SentryTracer cancelDeadlineTimer


philipphofmann:
### Description

The SDK crashes with `EXC_BAD_ACCESS ` here https://github.com/getsentry/sentry-cocoa/blob/e32423038e99f0c5d868744421d664cdd55c7383/Sources/Sentry/SentryTracer.m#L268-L272

<img width="973" alt="Screenshot 2023-10-02 at 09 37 31" src="https://github.com/getsentry/sentry-cocoa/assets/2443292/844ed259-f21c-4cd9-ba0b-d1e0d2c8cb50">

The [SDK crash in Sentry](https://sentry.sentry.io/issues/4300298551/?project=4505469596663808) has more context. This is happening across multiple versions see 
![image](https://github.com/getsentry/sentry-cocoa/assets/2443292/8ff97bbf-ef2c-4543-bb3b-eda4f13c59d2)

This is related to https://github.com/getsentry/sentry-cocoa/issues/3002.

gid-sentry:
Hey team looks like we're potentially seeing this again in 8.43.0

kahest:
reopening for investigation. our internal monitoring shows a low volume of these coming in for more recent versions

philipphofmann:
We also see this in our [internal SDK crash detection](https://sentry.sentry.io/issues/6070510978/?project=4505469596663808&query=is%3Aunresolved%20deadline&referrer=issue-stream&sort=user&statsPeriod=14d&stream_index=0) in 8.41.0 to 8.45.0.

philipphofmann:
The root cause is in cancelDeadlineTimer

https://github.com/getsentry/sentry-cocoa/blob/b38050392f837ffaf920f504c4fe5ee80220dea7/Sources/Sentry/SentryTracer.m#L308-L327

The method `cancelDeadlineTimer` can be called from any thread, but the `deadlineTimer invalidate` runs on the main thread. It seems like we're having a race condition there, which causes a crash. I tried to reproduce the problem with a test in SentryTracerTests, but I haven't been successful yet.
```Swift
func testFinishTracerFromMultipleThreads() {
    SentryLog.withoutLogs {
        
        SentryDependencyContainer.sharedInstance().dispatchQueueWrapper = SentryDispatchQueueWrapper()
        
        let sut = fixture.getSut(idleTimeout: 0.001)
[...]

PR #3333, opened 2023-10-06
fix: Crash in SentryTracer when cancelling timer

philipphofmann:
## :scroll: Description

The SentryTracer starts the timer for the deadline on the main thread and invalidates the timer, not necessarily on the main thread, but invalidate has to be called on the same thread on which the timer was started. This is fixed by starting and invalidating the NSTimer on the main thread.

Fixes GH-3320

It isn't straightforward to reproduce the EXC_BAD_ACCESS in the SentryTracer, cause violating calling invalidate on a different thread doesn't necessarily lead to crashes, see [Apple Docs](EXC_BAD_ACCESS). Unit tests. 

## :pencil: Checklist


Sources/Sentry/SentryTracer.m
```diff
@@ -267,8 +272,12 @@ - (void)deadlineTimerFired
 
 - (void)cancelDeadlineTimer
 {
-    [self.deadlineTimer invalidate];
-    self.deadlineTimer = nil;
+    // The timer must be invalidated from the thread on which the timer was installed, see
+    // https://developer.apple.com/documentation/foundation/nstimer/1415405-invalidate#1770468
+    [_configuration.dispatchQueueWrapper dispatchOnMainQueue:^{
+        [self.deadlineTimer invalidate];
+        self.deadlineTimer = nil;
```
armcknight, line 279: Can this cause a retain cycle? I guess we sort of need it to hang around so we can actually cancel the timer, but will the tracer be released after this block exits?

armcknight (approved):
Good catch. Looks basically right, I had one quick question. Ship when ready!

Sources/Sentry/SentryTracer.m
philipphofmann: Yes, it definitely could. When the main thread is busy, the tracer is not deallocated as long as this block runs on the main thread. Good catch thanks 🥇. I fixed the code and added a test.

PR #3342, opened 2023-10-12
test: Fix flaky Reachability test

philipphofmann:
The testMultipleReachabilityObservers sometimes times out in CI. As no threading is involved, there is no need for test expectations. So we can replace these expectations by an int keeping track of how often the connectivity changes. This speeds up the tests significantly and reduces the complexity. Furthermore, the reset of SentryReachbility can now only be achieved by calling removeAllObservers.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryReachability.m
+++ b/Sources/Sentry/SentryReachability.m
@@ -165,6 +165,10 @@ SentryConnectivityCallback(
 
         sentry_reachability_queue
             = dispatch_queue_create("io.sentry.cocoa.connectivity", DISPATCH_QUEUE_SERIAL);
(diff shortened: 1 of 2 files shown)


brustolin (approved):

brustolin:
It still looks flaky 😞

brustolin:
The code looks good, that's why I approved.

But It may be flaky still

image

philipphofmann:
> The code looks good, that's why I approved.
> 
> But It may be flaky still
> 
> image

Yes, sadly. I will figure it out.

armcknight:
of course after adding logs in the test to see how things are interleaving, the test keeps passing. let's merge this and see if it pops back up, then we'll have the logs to investigate further.

PR #3346, opened 2023-10-17
feat: Added screen name to app context

### LINKED ISSUE #3140: Report as app context what's the current screen during the error
see https://github.com/getsentry/team-sdks/issues/18

brustolin:
## :scroll: Description

Add screens name to app context

close #3140 

Unit tests

## :pencil: Checklist


Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,90 @@ - (UIApplication *)sharedApplication
...
+        // The top view controller is meant for navigation and not content
+        if ([self isHierarchicViewController:topVC]) {
+            topVC = [self relevantViewControllerFromHierarchy:topVC];
+            continue;
+        }
+
+        UIViewController *relevantChild = NULL;
+        for (UIViewController *childVC in topVC.childViewControllers) {
+            // Sometimes a view controller is used as container for a navigation controller
+            // If the navigation is ocuppaing the whole view controller we will consider this the
```
philipphofmann, line 98: `l`:
```suggestion
            // If the navigation is occupating the whole view controller we will consider this the
```

Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,90 @@ - (UIApplication *)sharedApplication
...
+
+        break;
+    }
+
+    return topVC;
+}
+
+- (BOOL)isHierarchicViewController:(UIViewController *)viewController
+{
+    return [viewController isKindOfClass:UINavigationController.class] ||
```
philipphofmann, line 120: `m`: What about `UIPageViewController`? Shouldn't we also add it here?

Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,90 @@ - (UIApplication *)sharedApplication
...
+            continue;
+        }
+
+        break;
+    }
+
+    return topVC;
+}
+
+- (BOOL)isHierarchicViewController:(UIViewController *)viewController
```
philipphofmann, line 118: `l`: UIKit docs use the wording container for these UIViewControllers, see https://developer.apple.com/documentation/uikit/view_controllers/managing_content_in_your_app_s_windows#3370390.
```suggestion
- (BOOL)isContainerViewController:(UIViewController *)viewController
```

Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,90 @@ - (UIApplication *)sharedApplication
     return result;
 }
 
+- (NSArray<UIViewController *> *)relevantViewControllers
```
philipphofmann, line 59: `m`: I'm a bit worried that there are no tests for this.

philipphofmann (commented):
As mentioned on [Slack](https://sentry.slack.com/archives/G01PW6HRQNN/p1698136685601969?thread_ts=1698133259.138989&cid=G01PW6HRQNN), I'm unsure about the general approach of this PR, and we need to discuss it before continuing.

Sources/Sentry/SentryClient.m
```diff
@@ -784,6 +785,19 @@ - (void)applyExtraDeviceContextToEvent:(SentryEvent *)event
                     key:@"app"
                   block:^(NSMutableDictionary *app) {
                       [app addEntriesFromDictionary:extraContext[@"app"]];
+#if SENTRY_HAS_UIKIT
+                      if ([NSThread isMainThread]) {
```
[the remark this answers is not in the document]
brustolin: Sounds good, a sync call with timeout seems acceptable.

Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,90 @@ - (UIApplication *)sharedApplication
...
+{
+    NSMutableArray *result = [NSMutableArray array];
+
+    NSArray<UIWindow *> *windows = [self windows];
+    if ([windows count] == 0)
+        return nil;
```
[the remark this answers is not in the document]
brustolin: I prefer nil and not include the data in the envelope

Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,90 @@ - (UIApplication *)sharedApplication
...
+
+    for (UIWindow *window in windows) {
+        UIViewController *vc = [self relevantViewControllerFromWindow:window];
+        if (vc != NULL) {
```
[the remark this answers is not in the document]
brustolin: Yeap, I keep forgetting which null type to use in each project 😂

Sources/Sentry/SentryUIApplication.m
brustolin: Sure. I think I never used it.

philipphofmann: Me neither; I checked the Apple docs for container view controllers. That's how I found it.

Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,105 @@ - (UIApplication *)sharedApplication
...
+}
+
+- (NSArray<UIViewController *> *)relevantViewControllerFromWindow:(UIWindow *)window
+{
+    UIViewController *rootViewController = window.rootViewController;
+    if (rootViewController == nil) {
+        return nil;
+    }
+
+    NSMutableArray<UIViewController *> *result = @[ rootViewController ].mutableCopy;
```
armcknight, line 84: I think this would be less allocations:
```suggestion
    NSMutableArray<UIViewController *> *result = [NSMutableArray<UIViewController *> arrayWithObject:rootViewController];
```

Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,105 @@ - (UIApplication *)sharedApplication
...
+        if ([self isContainerViewController:topVC]) {
+            [result removeObjectAtIndex:index];
+            [result addObjectsFromArray:[self relevantViewControllerFromContainer:topVC]];
+            continue;
+        }
+
+        UIViewController *relevantChild = nil;
+        for (UIViewController *childVC in topVC.childViewControllers) {
+            // Sometimes a view controller is used as container for a navigation controller
+            // If the navigation is occupating the whole view controller we will consider this the
```
armcknight, line 105:
```suggestion
            // If the navigation is occupying the whole view controller we will consider this the
```

Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,105 @@ - (UIApplication *)sharedApplication
...
+            // case.
+            if ([self isContainerViewController:childVC]
+                && CGRectEqualToRect(childVC.view.frame, topVC.view.bounds)) {
```
armcknight, line 108: Why are you comparing `frame` and `bounds`? `bounds` will always have an origin of `{0,0}`, but if a vc has multiple child vcs layed out in different locations, their `frame` origins will be nonzero in x or y for some of them; there might not be any at `{0,0}`.

Sources/Sentry/include/SentryUIApplication.h
```diff
@@ -41,6 +42,12 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (NSArray<UIScene *> *)getApplicationConnectedScenes:(UIApplication *)application
     API_AVAILABLE(ios(13.0), tvos(13.0));
+
+/**
+ * Return the most relevant view controller being presented in the app.
+ */
+- (nullable NSArray<UIViewController *> *)relevantViewControllers;
```
armcknight, line 49: Could you explain a bit more here? The comment speaks about a singular vc, but this returns an array.

Tests/SentryTests/SentryClientTests.swift
```diff
@@ -277,6 +277,68 @@ class SentryClientTest: XCTestCase {
         eventId.assertIsEmpty()
     }
     
+#if os(iOS) || targetEnvironment(macCatalyst)
```
armcknight, line 280: `SENTRY_HAS_UIKIT` will also run in tvOS, do we need to run some of these tests for that platform?

Sources/Sentry/SentryUIApplication.m
```diff
@@ -56,6 +56,105 @@ - (UIApplication *)sharedApplication
...
+    }
+
+    NSMutableArray<UIViewController *> *result = @[ rootViewController ].mutableCopy;
+    NSUInteger index = 0;
+
+    while (index < result.count) {
+        UIViewController *topVC = result[index];
+        // If the view controller is presenting another one, usually in a modal form.
+        if (topVC.presentedViewController != nil) {
+            [result replaceObjectAtIndex:index withObject:topVC.presentedViewController];
```
armcknight, line 91: Two questions:
1. Why replace instead of reporting them both? The combination might be interesting information.
2. Even if we replace presenters with modals, do we also want to do that for UIAlertControllers? (dialogs and share sheets are both presented via that API which is also a UIVC subclass)

armcknight (commented):
Good work on this, a couple questions for you!

Sources/Sentry/SentryUIApplication.m
armcknight: I guess breadcrumbs could be used to reconstruct the trail of presenters and other navigation sequences that led to the topmost vc(s) currently visible.

Samples/iOS-Swift/iOS-Swift/ErrorsViewController.swift
```diff
@@ -26,7 +26,11 @@ class ErrorsViewController: UIViewController {
     }
 
     @IBAction func crash(_ sender: UIButton) {
-        SentrySDK.crash()
+        //SentrySDK.crash()
+        
+        DispatchQueue.global().async {
+            SentrySDK.capture(message: "dae")
+        }
```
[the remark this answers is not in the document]
brustolin: This is a test that was left behind 🤦🏻

brustolin, line 33:
```suggestion
       SentrySDK.crash()
```

Samples/iOS-Swift/iOS-Swift/ErrorsViewController.swift
```diff
@@ -26,7 +26,7 @@ class ErrorsViewController: UIViewController {
     }
 
     @IBAction func crash(_ sender: UIButton) {
-        SentrySDK.crash()
+       SentrySDK.crash()
```
brustolin, line 29:
```suggestion
        SentrySDK.crash()
```

Sources/Sentry/SentryTracer.m
```diff
@@ -384,6 +388,33 @@ - (SentryTraceContext *)traceContext
...
+        }
+    };
+
+    // We need to retrieve the current view controller on the main thread.
+    // If the transaction is called from a background thread and finishes
+    // before retrieving the current view controller, the event will not have this information.
+    [SentryDependencyContainer.sharedInstance.dispatchQueueWrapper
+        dispatchAsyncOnMainQueue:saveViewNames];
+#endif
+}
```
brustolin, line 416:
```suggestion
}
#endif
```

Sources/Sentry/SentryUIApplication.m
brustolin: I believe we need UI Tests for this.

Sources/Sentry/SentryUIApplication.m
brustolin: Because I want to compare the child view controller view frame to the bounds of its parent view controller view. 

This check is a big stretch, but im my experience it happens from time to time.

Sources/Sentry/SentryClient.m
```diff
@@ -808,9 +809,40 @@ - (void)applyExtraDeviceContextToEvent:(SentryEvent *)event
...
 }
 
+#if SENTRY_HAS_UIKIT
+- (void)addViewNamesToContext:(NSMutableDictionary *)appContext event:(SentryEvent *)event
+{
+    if ([event isKindOfClass:[SentryTransaction class]]) {
+        SentryTransaction *transaction = (SentryTransaction *)event;
+        if ([transaction.screens count] > 0) {
+            appContext[@"view_names"] = transaction.screens;
+        }
```
philipphofmann, line 825: `h`: If I'm not mistaken, we must only attach `view_names` for transactions when they have one. If not, we should exit here.
```suggestion
        }
        return;
```

philipphofmann (commented):
Thanks for applying all the changes. I added a couple of comments.

Sources/Sentry/SentryClient.m
brustolin: Well, this feature was defined by us. Do you think this is the best approach?

Sources/Sentry/SentryClient.m
```diff
@@ -808,9 +809,40 @@ - (void)applyExtraDeviceContextToEvent:(SentryEvent *)event
...
+                = SentryDependencyContainer.sharedInstance.application.relevantViewControllers;
+            NSMutableArray *vcsNames = [NSMutableArray array];
+            for (id vc in viewControllers) {
+                [vcsNames addObject:[SwiftDescriptor getObjectClassName:vc]];
+            }
+            appContext[@"view_names"] = vcsNames;
+        };
+
+        [[SentryDependencyContainer.sharedInstance dispatchQueueWrapper]
+            dispatchSyncOnMainQueue:addViewName
```
brustolin, line 840:
```suggestion
            dispatchSyncOnMainQueue:addViewNames
```

Sources/Sentry/SentryClient.m
```diff
@@ -808,9 +809,40 @@ - (void)applyExtraDeviceContextToEvent:(SentryEvent *)event
...
                   }];
 }
 
+#if SENTRY_HAS_UIKIT
+- (void)addViewNamesToContext:(NSMutableDictionary *)appContext event:(SentryEvent *)event
+{
+    if ([event isKindOfClass:[SentryTransaction class]]) {
+        SentryTransaction *transaction = (SentryTransaction *)event;
+        if ([transaction.screens count] > 0) {
+            appContext[@"view_names"] = transaction.screens;
```
brustolin, line 824:
```suggestion
            appContext[@"view_names"] = transaction.viewNames;
```

Sources/Sentry/SentryTracer.m
```diff
@@ -641,6 +645,10 @@ - (SentryTransaction *)toTransaction
 
 #if SENTRY_HAS_UIKIT
     [self addMeasurements:transaction];
+
+    if ([viewNames count] > 0) {
```
philipphofmann, line 649: Thanks. That's way better than adding more code to the tracer.

philipphofmann, line 649: `m`: Do we have tests for this new logic?

Sources/Sentry/include/SentryUIApplication.h
```diff
@@ -41,6 +42,21 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (NSArray<UIScene *> *)getApplicationConnectedScenes:(UIApplication *)application
     API_AVAILABLE(ios(13.0), tvos(13.0));
+
+/**
+ * Retrieve the topmost view controllers currently presented in the app.
+ * In the case of a multi-screen iPad app, it will contain the most relevant
+ * view controller for each screen. If the topmost view controller
+ * is a split view controller, all of its sides are reported.
+ */
+- (nullable NSArray<UIViewController *> *)relevantViewControllers;
```
philipphofmann, line 52: `m`: I think this method can now be internal to `SentryUIApplication`. No other class is using it, they all use `relevantViewControllersNames`.

philipphofmann (commented):
We still miss some UITests or tests for the code in SentryUIApplication. Apart from that, this already looks great.

Sources/Sentry/SentryTracer.m
brustolin: Yeap!!

Samples/iOS-Swift/iOS-Swift/ViewControllers/PageViewController.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+import UIKit
+
+class PageViewController: UIPageViewController, UIPageViewControllerDataSource {
+    
+    class RedViewController: UIViewController {
+        override func viewDidLoad() {
+            super.viewDidLoad()
+            view.backgroundColor = .red
```
philipphofmann, line 9: Thanks for making them easy to differentiate.

Samples/iOS-Swift/iOS-SwiftUITests/TopViewControllerTests.swift
```diff
@@ -0,0 +1,89 @@
...
+    
+    override func tearDown() {
+        app.terminate()
+        super.tearDown()
+    }
+    
+    func testTabBarViewController() {
+        openInspector()
+        
+        let getTopBT = app.buttons["BTN_TOPVC"]
```
philipphofmann, line 24: `l`: I think we don't need to shorten button. For me the following is easier to read. Just a personal preference. The same applies to lbl. I would call it `LABEL_TOPVC`
```suggestion
        let getTopBT = app.buttons["BUTTON_TOPVC"]
```

Samples/iOS-Swift/iOS-SwiftUITests/TopViewControllerTests.swift
```diff
@@ -0,0 +1,89 @@
+import XCTest
+
+class TopViewControllerTests: XCTestCase {
```
philipphofmann, line 3: `m`: When you use `BaseUITest` added with https://github.com/getsentry/sentry-cocoa/pull/3373 you can remove the `setUp`, `tearDown` and `waitForExistenceOfMainScreen`.

philipphofmann, line 3: Thanks for adding some UITests.

Samples/iOS-Swift/iOS-Swift/Tools/SentryUIApplication.h
```diff
@@ -0,0 +1,58 @@
...
+@protocol UIApplicationDelegate;
+
+typedef NS_ENUM(NSInteger, UIApplicationState);
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * A helper tool to retrieve informations from the application instance.
+ */
+@interface SentryUIApplication : NSObject
```
philipphofmann, line 14: `m`: Why do we need this header file here? Please add a short comment with the explanation or remove it.

philipphofmann (approved):
LGTM with a couple of minor suggestions.

Samples/iOS-Swift/iOS-Swift/Tools/SentryUIApplication.h
armcknight: If it is needed here for some reason, we should use the one already in /Sources. See how SentryDevice.h/.mm are included in iOS-Swift/tvOS-Swift.

Sources/Sentry/SentryUIApplication.m
```diff
@@ -86,6 +88,148 @@ - (UIApplication *)sharedApplication
...
+        if (tbController.viewControllers.count > selectedIndex) {
+            return @[ [tbController.viewControllers objectAtIndex:selectedIndex] ];
+        } else {
+            return nil;
+        }
+    }
+    if ([containerVC isKindOfClass:UISplitViewController.class]) {
+        UISplitViewController *splitVC = (UISplitViewController *)containerVC;
+        if (splitVC.viewControllers.count > 0) {
+            return [splitVC viewControllers];
```
armcknight, line 221: So are these false positives? I see the UI test for split VC. Maybe we're not generating coverage from sample app UI tests? Otherwise we might want to raise this with codecov.

Samples/iOS-Swift/iOS-Swift/Tools/SentryUIApplication.h
brustolin: Adding the comment.
This is exposing private API for testing.

I rather have a copy of the header instead of using the same .h file because we have a better control of what is being exposed, and any change in the `#imports` in the original file wont affect the sample.

Sources/Sentry/SentryUIApplication.m
brustolin: Exactly, I don't think we generate coverage from UI Tests.

Samples/iOS-Swift/iOS-SwiftUITests/TopViewControllerTests.swift
brustolin: Sorry, I will let this one slide.

philipphofmann:
One final question just popped into my head. How expensive is the traversing of the UIViewControllers? Could it be that we slow down apps if they have a complex hierarchy of UIViewControllers, @brustolin? Are there any other risks involved with that PR? Should we maybe do a beta or are you confident to ship this PR right away?

If all answers to my questions are no, let's ship this 🚢 @brustolin

armcknight:
> One final question just popped into my head. How expensive is the traversing of the UIViewControllers? Could it be that we slow down apps if they have a complex hierarchy of UIViewControllers, @brustolin? Are there any other risks involved with that PR? Should we maybe do a beta or are you confident to ship this PR right away?
> 
> If all answers to my questions are no, let's ship this 🚢 @brustolin

We could get a quick idea on this if we create a perf test that measures the routine and create a full tree with N children per node and check it with various depths. Some quick numbers:

- 5 children per VC
    - 1 level: 5 vcs
    - 2 levels: 25 vcs
    - 3 levels: 125 vcs
    - 4 levels: 625 vcs

I can't quite imagine a view hierarchy with that many children VCs and layers, but iterating through 625 pointer references shouldn't take very long.

It's not possible to push a nav controller onto itself, but what if someone accidentally adds a view controller as a child of itself? Infinite loop? Any other possibilities for that here?

brustolin:
- Its impossible to create a infinite loop of UI Elements.
- The speed to navigate the view controller tree is as fast as accessing properties of an objects, and vc trees are not big.
- I believe this is good to go

philipphofmann:
> I can't quite imagine a view hierarchy with that many children VCs and layers, but iterating through 625 pointer references shouldn't take very long.

That's also my assumption. I would be surprised if it takes very long.

PR #3352, opened 2023-10-19
Fix: SDK detects OutOfMemory after device reboot

sepbehroozi:
## :scroll: Description

If the host app has been terminated due to device reboot, SentrySDK assumes it as a WatchdogTermination crash. In such cases, mostly, the boot time will be mentioned as "a few seconds before this event". 

Screenshot 2023-10-19 at 09 52 27

The idea behind this fix is that if the system boot time has been changed since the previous app launch, most likely the app has been terminated due to the reboot and the SentrySDK should not consider it as a WatchdogTermination crash.

This change helps reducing the amount of false WatchdogTermination events detected by the SDK.

There's a new test case added in the `SentryWatchdogTerminationsTrackerTests.swift` file, which sets the previous app-launch time to different value and expects not sending any OOM event.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Improve OOM detection by ignoring system reboot (#3352)
+
 ## 8.14.1
 
 ### Fixes
--- a/Sources/Sentry/SentryWatchdogTerminationLogic.m
+++ b/Sources/Sentry/SentryWatchdogTerminationLogic.m
@@ -62,7 +62,7 @@ SentryWatchdogTerminationLogic ()
 
     // The app may have been terminated due to device reboot
     if (previousAppState.systemBootTimestamp != currentAppState.systemBootTimestamp) {
-      return NO;
+        return NO;
     }
 
     // This value can change when installing test builds using Xcode or when installing an app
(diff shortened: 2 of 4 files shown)


sepbehroozi:
> Thanks @sepbehroozi for contributing. I believe this change makes sense.
> Would you mind taking a look at the tests that are failing?

Hello @brustolin 
Thanks for checking the PR.
Of course. I'll check them later today 👍

sepbehroozi:
> Thanks @sepbehroozi for contributing. I believe this change makes sense. Would you mind taking a look at the tests that are failing?

Hey @brustolin. I believe the reason for the failed tests was that the `Fixture` class of the `SentryCrashIntegrationTests` doesn't introduce a mock `SysctlWrapper` (In this case, the `TestSysctl()`) to the `SentryDependencyContainer`. That's why the `buildCurrentAppState` function in the `SentryAppStateManager` class uses the real `SentrySysctl` class when getting called from the `SentryCrashIntegrationTests` tests.
What I thought would be a good solution in this case, setting the `sysctlWrapper` of `SentryDependencyContainer.sharedInstance()` to `TestSysctl()` in the Fixture's initializer. I hope I could make sense of it :))

philipphofmann (commented):
Good catch, thanks, @sepbehroozi 🥳.  I also think this change makes sense. I leave it up to @brustolin to give you a final review.

brustolin (approved):
LGTM

PR #3356, opened 2023-10-20
fix: improvements resetting Reachability and other test state

armcknight:
Noticed in some test logs that there were many instances of reachability observers being notified where it didn't make sense to have so many. I realized that as part of `clearTestState()`, we would try to access the dependency container's reachability singleton to reset observers, but if the test had never activated initializing the instance in the first place, then we'd allocate one as a side effect of trying to reset the observers. Those instances then hung around.

I reworked the reset pathway to always remove reachability observers on the reachability instance, if it exists, when calling SentryDependencyContainer.reset.

Also noticed there were some test cases that would call clearTestState in both setUp and tearDown. So we'd have a sequence like

clearTestState
test1
clearTestState
clearTestState
test2
clearTestState
clearTestState
....

I removed it from setUp to avoid the duplicate calls between test cases, but to be on the safe side, added it to the _class_ method setUp, so state would be cleared before the _set_ of test cases ran, just in case there was a legitimate reason to do so. There are probably plenty of test sets that don't call clearTestState in their teardown.

Also at some point I went looking for the logs I'd added to testMultipleReachabilityObservers but didn't see them; I'm not sure NSLog works from both the tests and system under test. I think printf fixes that so all the logs appear in CI.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Networking/SentryReachabilityTests.m
+++ b/Tests/SentryTests/Networking/SentryReachabilityTests.m
@@ -20,7 +20,8 @@
 
 - (void)connectivityChanged:(BOOL)connected typeDescription:(nonnull NSString *)typeDescription
 {
-    NSLog(@"Received connectivity notification: %i; type: %@", connected, typeDescription);
+    printf("Received connectivity notification: %i; type: %s\n", connected,
+        typeDescription.UTF8String);
     self.connectivityChangedInvocations++;
 }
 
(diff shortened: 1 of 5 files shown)


Tests/SentryTests/Networking/SentryReachabilityTests.m
```diff
@@ -69,41 +70,39 @@ - (void)testConnectivityRepresentations
...
 
-    NSLog(@"[Sentry] [TEST] reachability callback to set to back to reachable");
+    printf("[Sentry] [TEST] reachability callback set back to reachable\n");
     SentryConnectivityCallback(
         self.reachability.sentry_reachability_ref, kSCNetworkReachabilityFlagsReachable, nil);
-    NSLog(@"[Sentry] [TEST] reachability callback to set to back to intervention required");
-    SentryConnectivityCallback(self.reachability.sentry_reachability_ref,
-        kSCNetworkReachabilityFlagsInterventionRequired, nil);
+    printf("[Sentry] [TEST] reachability callback set back to unreachable\n");
+    SentryConnectivityCallback(self.reachability.sentry_reachability_ref, 0, nil);
```
armcknight, line 93: I realized that `kSCNetworkReachabilityFlagsInterventionRequired` shouldn't even be detected by our reachability monitoring because we only return `YES` for "reachable" (and with UIKit, also "wwan"). Not sure "intervention required" should've ever worked, but trying it with 0 instead.

brustolin (approved):
LGTM

Curious about why are you favouring printf instead of NSLog?

philipphofmann (approved):
Thanks @armcknight 👍

armcknight:
I thought i'd seen logs where the NSLogs weren't showing up, but don't remember where that was now. I'll put them back to NSLog and if it happens again we can see what's up.

PR #3357, opened 2023-10-20
fix: remove redundant nilcheck

armcknight:
Just a small thing I noticed and figured could be improved. Sending `boolValue` to `nil` will result in a falsy value.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryHub.m
+++ b/Sources/Sentry/SentryHub.m
@@ -642,7 +642,7 @@ SentryHub ()
 
             SentryLevel level = sentryLevelForString(eventJson[@"level"]);
             if (level >= kSentryLevelError) {
-                *handled = [self eventContainsOnlyHandledErrors:eventJson];
+                *handled = [self eventContainsUnhandledError:eventJson];
                 return YES;
             }
         }
@@ -650,14 +650,14 @@ SentryHub ()
     return NO;
 }
 
-- (BOOL)eventContainsOnlyHandledErrors:(NSDictionary *)eventDictionary
+- (BOOL)eventContainsUnhandledError:(NSDictionary *)eventDictionary
 {
     NSArray *exceptions = eventDictionary[@"exception"][@"values"];
     for (NSDictionary *exception in exceptions) {
         NSDictionary *mechanism = exception[@"mechanism"];
         NSNumber *handled = mechanism[@"handled"];
 
-        if (handled != nil && [handled boolValue] == NO) {
+        if ([handled boolValue] == NO) {
             return NO;
         }
     }
--- a/Sources/Sentry/SentryHub.m
+++ b/Sources/Sentry/SentryHub.m
@@ -642,7 +642,7 @@ SentryHub ()
 
             SentryLevel level = sentryLevelForString(eventJson[@"level"]);
             if (level >= kSentryLevelError) {
-                *handled = [self eventContainsUnhandledError:eventJson];
+                *handled = [self eventContainsOnlyHandledErrors:eventJson];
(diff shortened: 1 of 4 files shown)


Sources/Sentry/SentryHub.m
```diff
@@ -657,7 +657,7 @@ - (BOOL)eventContainsOnlyHandledErrors:(NSDictionary *)eventDictionary
         NSDictionary *mechanism = exception[@"mechanism"];
         NSNumber *handled = mechanism[@"handled"];
 
-        if (handled != nil && [handled boolValue] == NO) {
+        if (handled.boolValue == NO) {
```
brustolin, line 660: This was not a redundant check. If handled is nil it means handled, otherwise check its value.
In this change, if handled is nil, boolValue will default no NO which means not handled.

```suggestion
       //If either the mechanism or the mechanism handled property is not  present, 
       //it means this is a user manual capture, which means it was handled.
       //That's why we need to check if handled is nil, to consider it handled.
        if (handled != nil && [handled boolValue] == NO) {
```

armcknight: Ah, that's where I misunderstood. Quite unintuitive that not having a variable holding a result indicating that it was handled... to mean it was handled 😆 We could pepper the code with comments like this, but I'd be more inclined to add another property called something like `userReported` that we could check a truthy value for here, instead of having a tribool check with nil/true/false. Where would we set such a property, somewhere in the codepath for `- [SentryClient captureError...]`?

I see this was just added and released :/ https://github.com/getsentry/sentry-cocoa/pull/3353/files

armcknight: Oh, of course, we could always just set `handled` to `YES` in such locations.

armcknight: Is it just flutter apps that can result in an error event with a `nil` for the entire `mechanism`? That seems to really be the case being handled here. Not just a `nil` value for `handled`. @brustolin @krystofwoldrich

armcknight: So like, making this nonnull: https://github.com/getsentry/sentry-cocoa/blob/ddfd670e1d40ed2ea47c70b44b16a4be8ac02014/Sources/Sentry/Public/SentryMechanism.h#L36

and setting the handled property here: https://github.com/getsentry/sentry-cocoa/blob/ddfd670e1d40ed2ea47c70b44b16a4be8ac02014/Sources/Sentry/SentryClient.m#L270

If this really is just a flutter specific thing, then I'd say we add another property for that somewhere. That is easier to reason about vs possible nil values, and the compiler can help us out more.

armcknight: So like, making this nonnull: https://github.com/getsentry/sentry-cocoa/blob/ddfd670e1d40ed2ea47c70b44b16a4be8ac02014/Sources/Sentry/Public/SentryMechanism.h#L36

and setting the handled property here: https://github.com/getsentry/sentry-cocoa/blob/ddfd670e1d40ed2ea47c70b44b16a4be8ac02014/Sources/Sentry/SentryClient.m#L270

If this really is just a flutter specific thing, then I'd say we add another property for that somewhere. That is easier to reason about vs possible nil values, and the compiler can help us out more.

brustolin: An event created or an exception capture by the user also does not have the mechanism. We could refactor it all, but there is no immediate benefit for all the extra work. Imo the comment that I suggested, or a little refactoring of this `eventContainsOnlyHandledErrors` function would be enough to avoid future mistakes.

Issue #3355, opened 2023-10-20
Tag all spans with thread info


philipphofmann:
### Description

See https://github.com/getsentry/team-mobile/issues/153.

PR #3359, opened 2023-10-25
feat: Add thread id and name to span data

philipphofmann:
## :scroll: Description

Add current thread name and thread id to every span when the SDK initializes the span.

Fixes GH-3355

Unit tests and with the sample app.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentrySpan.m
+++ b/Sources/Sentry/SentrySpan.m
@@ -45,7 +45,9 @@ SentrySpan ()
         } else {
             NSString *threadName = [SentryDependencyContainer.sharedInstance.threadInspector
                 getThreadName:currentThread];
-            _data[SPAN_DATA_THREAD_NAME] = threadName;
+            if (threadName.length > 0) {
+                _data[SPAN_DATA_THREAD_NAME] = threadName;
+            }
         }
 
         _tags = [[NSMutableDictionary alloc] init];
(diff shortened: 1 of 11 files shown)


Sources/Sentry/SentrySpan.m
```diff
@@ -33,6 +36,20 @@ - (instancetype)initWithContext:(SentrySpanContext *)context
...
+
+        if ([NSThread isMainThread]) {
+            _data[SPAN_DATA_THREAD_NAME] = @"main";
+        } else {
+            NSString *threadName = [SentryDependencyContainer.sharedInstance.threadInspector
+                getThreadName:currentThread];
+            if (threadName.length > 0) {
+                _data[SPAN_DATA_THREAD_NAME] = threadName;
+            }
+        }
```
armcknight, line 51: Any reason we couldn't put this block of logic into SentryThreadInspector, so that the code here would just read as
```
_data[SPAN_DATA_THREAD_NAME] = SentryDependencyContainer.sharedInstance.threadInspector.threadName;
```

armcknight (approved):
Looks good, one question on architecture but nothing blocking!

Sources/Sentry/SentrySpan.m
philipphofmann: ThreadInspector threadName returns an empty string when the thread name is `nil`, see https://github.com/getsentry/sentry-cocoa/blob/ae9c51b0d6e71205648e9d44d9f4d61827ccd1e6/Sources/Sentry/SentryThreadInspector.m#L211-L215

I will fix that in another PR.

Good catch, thanks @armcknight 👍

philipphofmann: I opened a PR https://github.com/getsentry/sentry-cocoa/pull/3361

PR #3361, opened 2023-10-27
impr: Stop sending empty thread names

philipphofmann:
## :scroll: Description

Set the thread name to nil when retrieving the thread name from the machine context wrapper in case it fails or the thread name is empty. This slightly reduces the payload size.

Came up here https://github.com/getsentry/sentry-cocoa/pull/3359#discussion_r1374232007.

Unit test and running the sample app on a simulator

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryCrashDefaultMachineContextWrapper.m
+++ b/Sources/Sentry/SentryCrashDefaultMachineContextWrapper.m
@@ -41,11 +41,16 @@ SentryCrashThread mainThreadID;
     return thread;
 }
 
-- (void)getThreadName:(const SentryCrashThread)thread
+- (BOOL)getThreadName:(const SentryCrashThread)thread
             andBuffer:(char *const)buffer
          andBufLength:(int)bufLength;
 {
-    sentrycrashthread_getThreadName(thread, buffer, bufLength);
+    bool result = sentrycrashthread_getThreadName(thread, buffer, bufLength);
+    if (result == true) {
+        return YES;
+    } else {
+        return NO;
+    }
 }
 
 - (BOOL)isMainThread:(SentryCrashThread)thread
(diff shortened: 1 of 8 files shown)


armcknight (commented):
Code looks fine, but does the server expect something to be in this field? What happens if the field isn't present?

Sources/Sentry/SentryCrashDefaultMachineContextWrapper.m
```diff
@@ -41,11 +41,16 @@ - (SentryCrashThread)getThread:(struct SentryCrashMachineContext *)context withI
...
             andBuffer:(char *const)buffer
          andBufLength:(int)bufLength;
 {
-    sentrycrashthread_getThreadName(thread, buffer, bufLength);
+    bool result = sentrycrashthread_getThreadName(thread, buffer, bufLength);
+    if (result == true) {
+        return YES;
+    } else {
+        return NO;
+    }
```
brustolin, line 53: pick:
```suggestion
    return sentrycrashthread_getThreadName(thread, buffer, bufLength) == true;
```

The result of a comparison will be objc BOOL.

brustolin (approved):
LGTM, I would just check what @armcknight  pointed out, about the server being ok with this.

philipphofmann:
`thread.name` is part of [span data](https://develop.sentry.dev/sdk/performance/span-data-conventions/#thread). You don't have to send it. You can omit it. This is when you skip it for spans.

CleanShot 2023-10-30 at 14 28 18@2x

And when you send it
CleanShot 2023-10-30 at 14 28 59@2x

For threads on events, you can omit the thread name. It's an optional field see https://develop.sentry.dev/sdk/event-payloads/threads/#attributes.

PR #3363, opened 2023-10-27
ref: remove header import in header, in favor of fcd already present

armcknight:
This prevented Sentry from being build on an Intel mac, causing weird errors in the umbrella header until i made this change, then it built.

In general we should not import a header in another header unless we need to subclass an interface defined in it, etc.

#skip-changelog


brustolin (approved):
LGTM

philipphofmann (approved):
Thanks, LGTM.

Issue #3354, opened 2023-10-19
`[SentryTransactionContext getThreadInfo]` causes EXC_BAD_ACCESS - `getThreadInfo` attempted to dereference null pointer


jozefizso:
### Platform

macOS

Operating system: macOS 11.7.8(20G1351)  
Model: MacBookPro11,3 (Intel)

### Installed

Swift Package Manager

### Version

8.9.4

### Steps to Reproduce

1. Use SentrySDK v8.9.4 in macOS application
2. Initialise SentrySDK
3. Call `SentrySDK.startTransaction()` (on main thread)
4. The application may crash with `EXC_BAD_ACCESS - getThreadInfo > attempted to dereference null pointer`

```swift
SentrySDK.startTransaction(
    name: "InstallFlow",
    operation: #function,
    bindToScope: true
)
```

### Expected Result

Application should not crash.

### Actual Result

```
sentry::profiling::ThreadHandle::current() (in Acme) (SentryThreadHandle.cpp:49)
-[SentryTransactionContext getThreadInfo] (in Acme) (SentryTransactionContext.mm:127)
-[SentryTransactionContext commonInitWithName:source:parentSampled:] (in Acme) (SentryTransactionContext.mm:147)
-[SentryTransactionContext initWithName:nameSource:operation:origin:] (in Acme) (SentryTransactionContext.mm:64)
-[SentryHub startTransactionWithName:operation:bindToScope:] (in Acme) (SentryHub.m:323)
+[SentrySDK startTransactionWithName:operation:bindToScope:] (in Acme) (SentrySDK.m:200)
InstallFlow.start() (in Acme) (InstallFlow.swift:22)
```

jozefizso:
Across 65 devices we saw this crash only on single Intel based MacBook.

jozefizso:
More information from the crash log:

```
System Integrity Protection: enabled

Notes:                 Translocated Process

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [26547]

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
[...]

PR #3364, opened 2023-10-28
fix: avoid call to possibly crashing `mach_thread_deallocate`

armcknight:
In an attempt to fix https://github.com/getsentry/sentry-cocoa/issues/3354, but unverified because we have no repro.

The crash report in the attached issue points to line 49 in SentryThreadHandle.cpp, but that is not actually the line crashing if you look at the source code, which is constructing a C++ unique pointer, the line above that is actually a macro: when it expands in place, the actual line of code crashing on line 49 is actually the call to mach_port_deallocate wrapped in the macro–see the definition for SENTRY_PROF_LOG_KERN_RETURN: https://github.com/getsentry/sentry-cocoa/blob/2124551ae0a2da085c69d3f0838e9be646fa4a70/Sources/Sentry/include/SentryMachLogging.hpp#L30-L38) 

We hope that avoiding the call to mach_port_deallocate avoids the whatever issue is occurring, and also that the new logic is potentially more efficient. See https://codereview.chromium.org/276043002/ for more information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Fixes
 
 - Stop sending empty thread names (#3361)
-- Work around edge case with a thread info syscall sometimes returning an invalid pointer, leading to a crash (#3364)
+- Work around edge case with a thread info kernel call sometimes returning invalid data, leading to a crash (#3364)
 
 ## 8.14.2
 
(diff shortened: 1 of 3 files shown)


indragiek (approved):

philipphofmann (approved):
Please add a changelog entry, @armcknight. LGTM

PR #3365, opened 2023-10-28
fix: edge case crashes with profiled traces

### LINKED ISSUE #3299: Uncaught exception in `profilerForFinishedTracer`
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.13.0

### Steps to Reproduce

Will work on more comprehensive steps to repro async, but for now at least wanted to get this on your radar.

### Expected Result

Ability to call `-[SentryTracer finishWithStatus:]` without triggering internal exceptions. This was working in 8.8.0, but tracing has been broken for us since that release for various reasons, first being a deadlock then with 8.13 we hit this exception.

### Actual Result

> Expected a profiler to be associated with tracer id dfd99f76d40d4af38488db3fff789426.

```
#0  __exceptionPreprocess ()
#1  objc_exception_throw ()
#2  -[NSAssertionHandler handleFailureInFunction:file:lineNumber:description:] ()
#3  profilerForFinishedTracer at /<redacted>/sentry-cocoa/Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm:118
#4  +[SentryProfiler collectProfileBetween:and:forTrace:onHub:] at /<redacted>/sentry-cocoa/Sources/Sentry/SentryProfiler.mm:394
#5  +[SentryProfiler createProfilingEnvelopeItemForTransaction:] at /<redacted>/sentry-cocoa/Sources/Sentry/SentryProfiler.mm:363
#6  -[SentryTracer captureTransactionWithProfile:] at /<redacted>/sentry-cocoa/Sources/Sentry/SentryTracer.m:540
#7  -[SentryTracer finishInternal] at /<redacted>/sentry-cocoa/Sources/Sentry/SentryTracer.m:528
#8  -[SentryTracer canBeFinished] at /<redacted>/sentry-cocoa/Sources/Sentry/SentryTracer.m:428
#9  -[SentryTracer finishWithStatus:] at /<redacted>/sentry-cocoa/Sources/Sentry/SentryTracer.m:400
```

Yes, if I can figure out a fix

armcknight:
fixes #3299

Thanks for @philipphofmann finding some [cases](https://github.com/getsentry/sentry-cocoa/issues/3299#issuecomment-1787556279) where this could potentially happen, we'll patch up those cases until we hear any other information from the customer that reported the original issue.

This introduces an immutable internal tracer ID for concurrency tracking purposes, and avoids querying for profiling data if the profiler failed initializing.

#skip-changelog


Sources/Sentry/include/SentryTracer+Private.h
```diff
@@ -5,4 +5,10 @@ SentryTracer ()
 
 @property (nonatomic, strong) SentryHub *hub;
 
+/**
+ * We need an unchanging identifier to track concurrent tracers against a static profiler.
+ * @c SentryTracer.traceId can be changed by consumers so is unfit for this purpose.
+ */
+@property (nonatomic, strong) SentryId *concurrencyID;
```
philipphofmann, line 12: `h`: This has to be immutable.
```suggestion
@property (nonatomic, readonly, strong) SentryId *concurrencyID;
```

philipphofmann, line 12: `l`: I think we could use the ID also for other purposes than concurrency. We could also call it something like `immutableID`, `immutableTraceID`, etc.

Sources/Sentry/include/SentryTracer+Private.h
```diff
@@ -5,4 +5,10 @@ SentryTracer ()
 
 @property (nonatomic, strong) SentryHub *hub;
 
+/**
+ * We need an unchanging identifier to track concurrent tracers against a static profiler.
```
philipphofmann, line 9: Thanks for this comment 👏

Sources/Sentry/include/SentryProfiler.h
```diff
@@ -44,7 +44,7 @@ SENTRY_EXTERN_C_END
 /**
  * Start a profiler, if one isn't already running.
  */
-+ (void)startWithTracer:(SentryId *)traceId;
++ (BOOL)startWithTracer:(SentryId *)traceId;
```
philipphofmann, line 47: `m`: I think we should rename `traceID` to `concurrencyID` to state it's not the traceID. I would also rename all occurrences in `traceID` in `SentryProfiledTracerConcurrency`.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -331,16 +331,17 @@ + (void)startWithTracer:(SentryId *)traceId
         trackProfilerForTracer(_gCurrentProfiler, traceId);
         // record a new metric sample for every concurrent span start
         [_gCurrentProfiler->_metricProfiler recordMetrics];
-        return;
+        return NO;
```
philipphofmann, line 334: `m`: Shouldn't this be `YES`? We are profiling in that case.

```suggestion
        return YES;
```

philipphofmann (commented):
I think we should add a changelog entry for this, as this PR solves a problem.

Sources/Sentry/include/SentryTracer+Private.h
armcknight: I tried this and redeclaring it as readwrite in SentryTracer.m and it failed with error: duplicate declaration

armcknight: i'll call it `internalID`

Sources/Sentry/include/SentryProfiler.h
armcknight: i'll call it `internalID` to match the declaration's name in SentryTracer

Sources/Sentry/SentryProfiler.mm
armcknight: ay, good catch!

Sources/Sentry/include/SentryTracer+Private.h
armcknight: nm, don't even need the readwrite redeclare, just have to set the ivar in init

indragiek (approved):

philipphofmann:
Thanks for the quick fix @armcknight 👏

PR #2711, opened 2023-02-21
Add cacheDirectory option

JacobHearst:
## :scroll: Description
This PR adds the ability for developers to specify a custom location for Sentry's cache location.

## :bulb: Motivation and Context

This option is useful in scenarios where the application hosting Sentry doesn't have read/write permissions for the default cache location. This is particularly relevant for macOS applications running prelogin

## :green_heart: How did you test it?
Built a security agent plugin with the fork, specified the cache location, and ran the plugin at the login window
## :pencil: Checklist

- [x] I reviewed the submitted code
- [x] I added tests to verify the changes
- [x] I updated the docs if needed
- [x] Review from the native team if needed
- [x] No breaking changes

PR #3369, opened 2023-11-01
Add cacheDirectory option

mlink:
## :scroll: Description

This PR adds the ability for developers to specify a custom location for Sentry's cache location.
Updated PR that was closed awhile ago here #2711 

This option is useful in scenarios where the application hosting Sentry doesn't have read/write permissions for the default cache location. This is particularly relevant for macOS applications running prelogin

Built a security agent plugin with the fork, specified the cache location, and ran the plugin at the login window

## :pencil: Checklist


philipphofmann:
Thanks for opening this PR. I'm sorry, @mlink, we did get to it this week. I will give you a review beginning next week.

Sources/Sentry/include/SentrySession.h
```diff
@@ -19,6 +19,7 @@ typedef NS_ENUM(NSUInteger, SentrySessionStatus) {
 SENTRY_NO_INIT
 
 - (instancetype)initWithReleaseName:(NSString *)releaseName;
+- (instancetype)initWithReleaseName:(NSString *)releaseName cacheDirectoryPath:(nullable NSString*)cacheDirectoryPath;
```
philipphofmann, line 22: `m`: We should only offer one initializer and remove the one with only one parameter. Only SentryHub calls this at the moment. As the hub has the options available we can pass in the `distinctID` here. So instead of the session calling `SentryInstallation` for the ID the SentryHub should do that and pass the distincID.

Sources/Sentry/SentryInstallation.m
```diff
@@ -7,7 +7,7 @@ @implementation SentryInstallation
 
 static NSString *volatile installationString;
 
-+ (NSString *)id
++ (NSString *)idWithCacheDirectoryPath:(NSString *)cacheDirectoryPath
```
philipphofmann, line 10: `m`: We can now pass in different values for `cacheDirectoryPath`, but the `installationString` stays the same once initialized. So depending on which `cacheDirectoryPath` gets passed in first you get different `installationString`. To fix that, I think we need a dictionary of `cacheDirectoryPath` to `installationString`.

Sources/SentryCrash/Installations/SentryCrashInstallation.h
```diff
@@ -43,12 +43,13 @@
  * WARNING: Only call async-safe functions from this function! DO NOT call
  * Objective-C methods!!!
  */
-@property (atomic, readwrite, assign) SentryCrashReportWriteCallback onCrash;
+@property (nullable, atomic, readwrite, assign) SentryCrashReportWriteCallback onCrash;
```
philipphofmann, line 46: `m`: Why did you add `nullable` here?

Sources/SentryCrash/Installations/SentryCrashInstallation.h
```diff
@@ -43,12 +43,13 @@
...
 
 /** Install this installation. Call this instead of -[SentryCrash install] to
  * install with everything needed for your particular backend.
+ * If you wish to use default cache directory, pass @c nil .
  */
-- (void)install;
+- (void)install:(nullable NSString *)customCacheDirectory;
```
philipphofmann, line 52: `m`: We only call this from `SentryCrashIntegration` where we can access the options, so I would prefer making this nonnullable so you always have to pass it in.
```suggestion
- (void)install:(NSString *)customCacheDirectory;
```

Sources/SentryCrash/Recording/SentryCrash.m
```diff
@@ -40,7 +40,10 @@
 #import "SentryDefines.h"
 #import "SentryDependencyContainer.h"
 #import "SentryNSNotificationCenterWrapper.h"
+#import "SentrySDK+Private.h"
 #import <NSData+Sentry.h>
+#import <SentryClient+Private.h>
+#import <SentryHub.h>
```
philipphofmann, line 46: `h`: SentryCrash has a `sharedInstance` that initializes a default basePath in `getBasePath`. We need to make some changes so we use the basePath passed to `SentryCrashInstallation`.

philipphofmann (commented):
Thanks for tackling this, @mlink 😃 . There are a couple of things to change, and please fix the failing unit tests.

Sources/SentryCrash/Installations/SentryCrashInstallation.h
mlink: Screenshot 2023-11-07 at 1 50 40 PM
It seems by adding nullability specifies elsewhere in the file Xcode now requires it everywhere. I guessed that the value could be `nullable`, but I can change it to `_Null_unspecified`.

Sources/SentryCrash/Recording/SentryCrash.m
mlink: It looks like the `basePath` is changed in `-[SentryCrashInstallation install:]`. Should it be moved somewhere else?

Sources/SentryCrash/Installations/SentryCrashInstallation.h
armcknight: That's a strange one, since `assign` isn't an object type. In any case, for this header, if you wind up needing to mark anything as `nullable`, could you instead use `NS_ASSUME_NONNULL_BEGIN`/`NS_ASSUME_NONNULL_END` and only mark the things that are `nullable`? (And, this property should not be `nullable`. Between this and https://github.com/getsentry/sentry-cocoa/pull/3369#discussion_r1385110224, it looks like you may not need this any more in this header.)

Sources/SentryCrash/Recording/SentryCrash.m
armcknight: There's no longer a `+[SentryCrash sharedInstance]`, see now `SentryDependencyContainer.crashReporter` calling `[[SentryCrash alloc] init]` (you may need to merge our `main` branch again to see this, as we just recently merged #3247 after your reply).

Sources/SentryCrash/Installations/SentryCrashInstallation.h
mlink: Changed to use `NS_ASSUME_NONNULL_BEGIN` / `NS_ASSUME_NONNULL_END`

philipphofmann:
Thanks for all your changes @mlink 🙏 . Sorry, I didn't get to the PR today. It was a bit busy. Will try to give you another review tomorrow or Monday.

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -8,6 +8,12 @@ NS_ASSUME_NONNULL_BEGIN
 NS_SWIFT_NAME(Options)
 @interface SentryOptions : NSObject
 
+/**
+ * Optional, custom cache directory. Use when default one can not be accessed, e.g. in security
+ * environment.
```
philipphofmann, line 13: `m`: A bit clearer, IMO.
```suggestion
 * The path to store SDK data, like events, transactions, profiles, raw crash data, etc. We
 recommend only changing this when the default, e.g., in security environments, can't be accessed.
 * 
 * @note The default is `NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask,
 YES)`.
```

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -8,6 +8,12 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
+@property (nonatomic, copy) NSString *cacheDirectoryPath;
```
philipphofmann, line 15: `l`: Please move the property `parsedDsn`. It shouldn't be the first one in the Options file.

Tests/SentryTests/SentrySessionTests.m
```diff
@@ -26,7 +26,7 @@ - (void)testInitDefaultValues
 
 - (void)testSerializeDefaultValues
 {
-    SentrySession *expected = [[SentrySession alloc] initWithReleaseName:@"1.0.0"];
+    SentrySession *expected = [[SentrySession alloc] initWithReleaseName:@"1.0.0" distinctId:@""];
```
philipphofmann, line 29: `l`: I would use something different than an empty string to improve the quality of the test. The same applies for all the other tests.
```suggestion
    SentrySession *expected = [[SentrySession alloc] initWithReleaseName:@"1.0.0" distinctId:@"some-id"];
```

Sources/Sentry/SentryInstallation.m
```diff
@@ -5,38 +5,40 @@
...
 @implementation SentryInstallation
 
-static NSString *volatile installationString;
++ (NSMutableDictionary*)installationStringsByCacheDirectoryPaths
+{
+    static dispatch_once_t once;
+    static NSMutableDictionary* dictionary;
+    
+    dispatch_once(&once, ^{
+        dictionary = [[NSMutableDictionary alloc] init];
```
philipphofmann, line 14: `l`:
```suggestion
        dictionary = [NSMutableDictionary dictionary];
```

Sources/Sentry/SentryInstallation.m
```diff
@@ -5,38 +5,40 @@
 
 @implementation SentryInstallation
 
-static NSString *volatile installationString;
++ (NSMutableDictionary*)installationStringsByCacheDirectoryPaths
```
philipphofmann, line 8: `l`: I prefer specifying the type explicitly:
```suggestion
+ (NSMutableDictionary<NSString *, NSString*> *)installationStringsByCacheDirectoryPaths
```

Sources/Sentry/SentryInstallation.m
```diff
@@ -5,38 +5,40 @@
...
+    static NSMutableDictionary* dictionary;
+    
+    dispatch_once(&once, ^{
+        dictionary = [[NSMutableDictionary alloc] init];
+    });
+    return dictionary;
+}
 
-+ (NSString *)id
++ (NSString *)idWithCacheDirectoryPath:(NSString *)cacheDirectoryPath
```
philipphofmann, line 19: `m`: I know there are no tests for this, but please add some for the new logic with different cacheDirPaths. Please write them in Swift.

Sources/Sentry/SentryFileManager.m
```diff
@@ -639,8 +639,7 @@ + (BOOL)createDirectoryAtPath:(NSString *)path withError:(NSError **)error
 
 - (void)createPathsWithOptions:(SentryOptions *_Nonnull)options
 {
-    NSString *cachePath
-        = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES).firstObject;
+    NSString *cachePath = options.cacheDirectoryPath;
```
philipphofmann, line 642: `m`: Please add a test to the SentryFileManagerTests to ensure that the SentyrFileManager uses the path of the options. That's the primary use case of this PR, and I think it's worth adding a test for it.

Sources/SentryCrash/Recording/SentryCrash.m
philipphofmann, line 46: `m`: You can remove these headers. We don't need them.

Sources/SentryCrash/Recording/SentryCrash.h
```diff
@@ -59,6 +59,11 @@ static NSString *const SENTRYCRASH_REPORT_ATTACHMENTS_ITEM = @"attachments";
 /** Init SentryCrash instance with custom base path. */
 - (id)initWithBasePath:(NSString *)basePath;
 
+/** Optional, custom cache directory. Use when default one can not be accessed, e.g. in security
+ * environment. Default: @c nil .
+ */
```
philipphofmann, line 64: `m`: That comment isn't entirely accurate. The class isn't public, so users don't interact with it. It should always be set to the customCacheDirecotyPath from the options anyways, and there is not default.

Sources/SentryCrash/Installations/SentryCrashInstallation.h
```diff
@@ -28,6 +28,8 @@
 #import "SentryCrashReportFilter.h"
 #import "SentryCrashReportWriter.h"
 
+NS_ASSUME_NONNULL_BEGIN
```
philipphofmann, line 31: Thanks 👍

philipphofmann (commented):
We're almost at the finish line. Thanks a lot for doing this, @mlink 👏

Sources/Sentry/Public/SentryOptions.h
mlink: I'm guessing you meant make `parsedDsn` the first one in the options file? I'll move `cacheDirectoryPath`.

Sources/Sentry/SentryInstallation.m
```diff
@@ -5,38 +5,40 @@
...
         if (nil == installationData) {
             installationString = [NSUUID UUID].UUIDString;
-            NSData *installationStringData =
-                [installationString dataUsingEncoding:NSUTF8StringEncoding];
+            NSData *installationStringData = [installationString dataUsingEncoding:NSUTF8StringEncoding];
             NSFileManager *fileManager = [NSFileManager defaultManager];
-            [fileManager createFileAtPath:installationFilePath
-                                 contents:installationStringData
-                               attributes:nil];
+            [fileManager createFileAtPath:installationFilePath contents:installationStringData attributes:nil];
```
philipphofmann, line 36: `m`: I think we should now check the return value, and if it is `NO`, meaning that we couldn't write to the path, we should log an error with `SENTRY_LOG_ERROR` to at least inform our users with a log statement that something is wrong. I'm sorry; I could have mentioned this already in a previous review.

philipphofmann (commented):
The PR looks good so far, but I found a blocking issue. The code now doesn't set the cache path properly for SentryCrash. Previously, it was something like
```
/Users/philipp.hofmann/Library/Developer/CoreSimulator/Devices/0E1EFAC8-A7EC-4B5B-88D6-68A3CD1E73AA/data/Containers/Data/Application/901E30F7-0741-4821-8FF9-ED7072792E53/Library/Caches/SentryCrash/iOS-Swift
```

And now it is 
```
/Users/philipp.hofmann/Library/Developer/CoreSimulator/Devices/0E1EFAC8-A7EC-4B5B-88D6-68A3CD1E73AA/data/Containers/Data/Application/901E30F7-0741-4821-8FF9-ED7072792E53/Library/Caches
```

With the current approach, crashes from previous SDK versions wouldn't end up in Sentry and stay on the devices forever. 

`SentryCrash.getBasePath` is called before setting the `basePath` in the `SentryCrashInstallation.m`. First, both paths need to match. Second, let's find a way to remove `SentryCrash.getBasePath` cause we need the customCacheDirectory of the options.

I found this bug by installing the iOS-Swift sample app of the main branch, crashing the sample app, checking this branch, launching the iOS-Swift sample app, and then the crash event never made it into Sentry.

philipphofmann (commented):
Thanks a lot for including all the feedback @mlink 🥇. Excellent work 👏 . We're close to merging this.

philipphofmann (approved):
LGTM, thanks a lot for the contribution, @mlink, and for being patient with all my feedback. 💯 😃

PR #3370, opened 2023-11-02
Revert "Fix: App hang being reported from background"

philipphofmann:
Revert #3298 because `ProfilingUITests.testProfilingGPUInfo` continuously fails after merging #3298 to the main branch.

I could reproduce `ProfilingUITests.testProfilingGPUInfo` failing when running all UITests locally. Only running `ProfilingUITests.testProfilingGPUInfo` succeeds. Commenting out the following code made the test work again.

https://github.com/getsentry/sentry-cocoa/blob/dd0557fb0fec40f5527e4f40ead6f82e092fde0f/Sources/Sentry/SentryANRTrackingIntegration.m#L69-L76

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,6 @@
 
 ### Fixes
 
-- Reporting app hangs from background (#3298)
 - Missing `mechanism.handled` is not considered crash (#3353)
 
 ## 8.14.1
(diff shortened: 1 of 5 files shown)


philipphofmann:
Will be fixed with https://github.com/getsentry/sentry-cocoa/pull/3372.

PR #3372, opened 2023-11-02
fix: Reading applicationState in the background

brustolin:
Keep track of application state by notification instead of reading directly from UIApplication. This way we avoid using UIApplication in the background thread when detecting an ANR.

_#skip-changelog_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@
 
 - Stop sending empty thread names (#3361)
 - Work around edge case with a thread info kernel call sometimes returning invalid data, leading to a crash (#3364)
-- Reading applicationState in the background (#3372)
 
 ## 8.14.2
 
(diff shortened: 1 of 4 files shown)


Sources/Sentry/SentryUIApplication.m
```diff
@@ -4,7 +4,34 @@
...
 
-@implementation SentryUIApplication
+@implementation SentryUIApplication {
+    UIApplicationState appState;
+}
+
+- (instancetype)init
+{
+    if (self = [super init]) {
+        [NSNotificationCenter.defaultCenter addObserver:self
```
philipphofmann, line 14: `m`: Please use the SentryNSNotificationCenterWrapper and add some tests when this solves the problem.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 - Stop sending empty thread names (#3361)
 - Work around edge case with a thread info kernel call sometimes returning invalid data, leading to a crash (#3364)
+- Reading applicationState in the background (#3372)
```
philipphofmann, line 13: `m`: I don't this requires a changelog entry as it fixes a fix 😄

philipphofmann (approved):
LGTM.

CHANGELOG.md
brustolin, line 13:
```suggestion
```

PR #3374, opened 2023-11-03
fix: Crash when initializing SentryHub manually

philipphofmann:
## :scroll: Description

Fix a crash when serializing the trace context after initializing the SentryHub manually.

A user reported a crash on Discord when running the following code
```
let sentryOption = Options()
sentryOption.dsn = dsn
        
let scope = Scope()
self.sentryHub = SentryHub(client: SentryClient(options: sentryOption), andScope: scope)

let error = NSError(domain: "abc", code: 12)
sentryHub.capture(error: error)
```

image

Unit tests.

## :pencil: Checklist


Tests/SentryTests/SentryHubTests.swift
```diff
@@ -67,8 +67,6 @@ class SentryHubTests: XCTestCase {
         fixture.fileManager.deleteAppState()
         fixture.fileManager.deleteTimestampLastInForeground()
         fixture.fileManager.deleteAllEnvelopes()
-        
-        sut = fixture.getSut()
```
armcknight, line 71: I tried running the new hub test case on main to see why it's crashing, because my mental model of this area of the SDK and the changes is not built up very much.

I noticed though that there were two live hub instances because this one is always set up; I made it lazy so it's not initialized for your new test case.

Sources/Sentry/SentryPropagationContext.h
```diff
@@ -1,19 +1,29 @@
...
-@class SentryId, SentrySpanId, SentryTraceContext, SentryTraceHeader;
+@class SentryId, SentrySpanId, SentryTraceContext, SentryTraceHeader, SentryOptions, SentryScope;
 
 NS_ASSUME_NONNULL_BEGIN
 
 @interface SentryPropagationContext : NSObject
 
 @property (nonatomic, strong) SentryId *traceId;
 @property (nonatomic, strong) SentrySpanId *spanId;
-@property (nonatomic, readonly, nullable) SentryTraceContext *traceContext;
```
armcknight, line 11: Let's please stop using properties to autosynthesize methods that never store a backing ivar, it's confusing and I don't believe what they were intended for.

Sources/Sentry/SentryPropagationContext.h
```diff
@@ -1,19 +1,29 @@
...
 
+/**
+ * We need to pass in the options and the user segment instead of retrieving them from the static
+ * API cause, for example, the SentryClient can have multiple instances with different options and
+ * scopes running and in that scenario we need can't retrieve these values from the static API.
+ *
+ *  @param options The current active options.
+ *  @param userSegment You can retrieve this usually from the `scope.userObject.segment`.
+ */
+- (SentryTraceContext *)getTraceContext:(SentryOptions *)options
```
armcknight, line 23: The way this is named makes me think it's retrieving something like from an ivar, but really it's creating a new thing every time it's called. Can we rename it to e.g. 'traceContextForOptions:userSegment:`?

armcknight: Really, I would prefer that we just call `[[SentryTraceContext alloc] initWithTraceId:...]` in the places where this method is called, I don't understand why this should be wrapped in `SentryPropagationContext` at all, the indirection at the call sites is confusing.

philipphofmann: That is an excellent point. I fully I agree 🥇. I moved it.

brustolin (approved):
LGTM

Sources/Sentry/SentryPropagationContext.h
```diff
@@ -1,15 +1,13 @@
 #import <Foundation/Foundation.h>
 
-@class SentryId, SentrySpanId, SentryTraceContext, SentryTraceHeader;
+@class SentryId, SentrySpanId, SentryTraceContext, SentryTraceHeader, SentryOptions, SentryScope;
```
armcknight, line 3: No longer needed
```suggestion
@class SentryId, SentrySpanId, SentryTraceContext, SentryTraceHeader;
```

armcknight (approved):
LGTM, just needs a changelog update and one small fixup!

PR #3378, opened 2023-11-03
test: clear test state in any test case using SentryDependencyContainer

armcknight:
When looking at continued flakes of reachability tests, I noticed there are still dangling reachability observer instances when new test cases start (which I believe were coming from SentryHttpTransportTests). When I looked further I noticed that there were several tests that use SentryDependencyContainer, but never clean it up afterwards, so I added that call to each test suite's `tearDown` method.

Also changed one tearDown to tearDownWithError that did a manual XCTFail in a do/try/catch.

#skip-changelog


brustolin (approved):
LGTM

philipphofmann (approved):
Thanks @armcknight. LGTM.

PR #2426, opened 2022-11-22
test: base unit test class

armcknight:
While working on failing tests in #2228, I was hitting various crashes due to issues that seemed to be related to statefulness in the test suites. 

This PR creates a new base unit test class that subsumes the old `clearTestState` function we had and automatically runs it on every `setUp`/`tearDown` call–most tests did this, but some would only call it in one or the other.

Other tests had other state management in their `setUp`/`tearDown` that didn't appear in `clearTestState`; those have been added so that they can apply to all tests using this.

Not all tests inherit from this base class yet, only the ones that called `clearTestState`. This is meant to be a step in the direction of getting all our state under control for all tests. I think another step will be moving all of the private `Fixture` classes into this base class so tests are all presented with a consistent mock system, which can be modified as needed depending on the test case. See https://github.com/getsentry/sentry-cocoa/issues/2425

# TODO

Fixed by including a separate class definition for the UI tests. ~- [ ] Figure out how to avoid it being used in UI test targets. Some test classes are present in both UI test and unit test targets.~

#skip-changelog

PR #3381, opened 2023-11-07
test: cleanup more test state

armcknight:
- added `tearDown` methods that call `clearTestState` in other test cases that use `SentryDependencyContainer` (I missed these in #3378 because I'd only searched for existing `tearDown` methods that were missing calls to `clearTestState` and not missing `tearDown` methods)
- in a few ObjC test cases, just call `[SentryDependencyContainer reset]` because I haven't figured out how to access `clearTestState` from ObjC code
- remove unused `TestURLSession` class

#skip-changelog


armcknight:
> LGTM. What about adding a base test class that clears the test state in setup and teardown that all our tests use? If you agree, I will open up a PR.

Yeah, I agree this would be ideal. I attempted it a while back as part of #2426 but the refactor proved to be too big. It would be better if it were scoped just to clearing test state at first, then maybe consider fixture/sut abstraction later.

PR #3385, opened 2023-11-07
test: Fix reporting TTFD for Swift sample

philipphofmann:
Some UIViewControllers called report TTFD in onViewWillAppear, which is before the UIViewController is shown. Now, we report TTFD onViewDidAppear.

#skip-changelog


armcknight (approved):
Looks good 👍🏻

Issue #3392, opened 2023-11-09
MetricKit reporting reports huge amount of events with no clear benefit


DenTelezhkin:
### Platform

macOS

### Installed

Swift Package Manager

### Version

8.14.1

### Steps to Reproduce

1. Enable MetricKit reporting
2. Deploy to production
3. Observe huge amounts of useless events

### Expected Result

MetricKit should report something useful, and in smaller amounts of events. Currently it sends so much events, that would blow any pricing plan in a very short time.

For example, just in a single day following event was reported 276000 times for just 10 users.

```
start_wqthread
New
Level: Warning
MXCPUException totalCPUTime:90 sec totalSampledTime:162.209 sec
```

![CleanShot 2023-11-09 at 13 44 20@2x](https://github.com/getsentry/sentry-cocoa/assets/853604/c2822f6c-aabc-4400-bb8f-b2a5cc913bd1)

(Actual events can be found here - but this link requires authorization https://macpaw.sentry.io/issues/4605658052/?project=5428010&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=7d&stream_index=0)

There are other events, reported in similar fashion, but this one is by far a leader.

Not only this is a huge amount of events, I can see no useful value in them it's not clear how to "fix" any of those warnings.

### Actual Result

If those events are expected to be delivered in such quantities, MetricKit integration should have some kind of spike protection, or sampling rate, because clearly analyzing 276000 events from 10 users is useless.

### Are you willing to submit a PR?

Not sure what PR would look like for this problem.

philipphofmann:
> MetricKit should report something useful and in smaller amounts of events.

We don't have any control over how valuable the MetricKit data is. All our integration does is subscribe to the data of MetricKit and convert it to Sentry events. You can use `beforeSend` to drop events like you specified in the docs: https://docs.sentry.io/platforms/apple/configuration/metric-kit/

> You can identify MetricKit events by looking at the `mx_hang_diagnostic`, `mx_cpu_exception`, and `mx_disk_write_exception mechanism.type`.

Anyways, it seems your application is using excessive CPU time over a short period. Otherwise, MetricKit wouldn't create the event.

I checked your events, and I think we create around 15 MetricKit events with MXCPUException for a user at the same time with the same stacktrace and the same breadcrumbs. It could be that we do something wrong when converting MetricKit data.

[...]

PR #3395, opened 2023-11-09
fix: Infinite loop when parsing MetricKit data

philipphofmann:
## :scroll: Description

Fix an infinite loop when parsing MetricKit data for stacktraces with callStackPerThread false and no sub frames.

Based on https://github.com/getsentry/sentry-cocoa/pull/3393.

Fixes GH-3392

Unit tests.

## :pencil: Checklist


Sources/Sentry/SentryMetricKitIntegration.m
```diff
@@ -282,9 +282,14 @@ - (void)buildAndCaptureMXEventFor:(NSArray<SentryMXFrame *> *)rootFrames
...
-                // Pop all siblings
-                for (int i = 0; i < parentFrame.subFrames.count; i++) {
+                if (parentFrame == nil) {
+                    // No parent frames
                     [stackTraceFrames removeLastObject];
+                } else {
+                    // Pop all siblings
+                    for (int i = 0; i < parentFrame.subFrames.count; i++) {
+                        [stackTraceFrames removeLastObject];
+                    }
```
armcknight, line 292: I don't quite understand what the objective is here, after reading the diagram in the headerdoc comment for the method (thanks to whomever created that btw!), i'm not sure why this example gets a result that preserved frames 4 and 5 and not 2:
```
 * | frame 0 |
 *      | frame 1 |
 *          | frame 2 |
 *          | frame 3 |
 *              | frame 4 |
 *              | frame 5 |
 *              | frame 6 |     -> stack trace consists of [0, 1, 3, 4, 5, 6]
 ```
---
Is it describing a the following call graph (a):
```
time ----->

|---frame 0------------------------------------------|
  |--frame 1-----------------------------------------|
    |-frame 2-| |-frame 3----------------------------|
                  |-frame 4-| |-frame 5-| |-frame 6-|
```
which would  be a tree like
```
     0
     |
     1
    / \
   2   3
      /|\
     4 5 6
```
---
or is it like this one (b)
```
time ----->

|---frame 0------------------------------------------|
  |--frame 1-----------------------------------------|
    |-frame 2-| |-frame 3----------------------------|
                  |-frame 4-----| 
                   |-frame 5---| 
                    |-frame 6-|
```
which would have a tree like
```
     0
     |
     1
    / \
   2   3
       |
       4
       |
       5
       |
       6
```
because the way the headerdoc diagram describes it sounds like (a) to me, but the result makes me think it's like (b). 
---
So I guess I'm just looking for a little more clarification on what "pop siblings" means, because I expected it to either remove 4 and 5 or preserve 2. It might be helpful to also include some diagrams describing what each test case is doing for an existing test and the new one you wrote.

armcknight (commented):
Nice find and nice work on this! I'm just looking for a little more explanation of the solution for some possible inputs.

Sources/Sentry/SentryMetricKitIntegration.m
philipphofmann: I created the diagram. It's b. They aren't siblings in terms of the tree but in terms of the stack trace level nesting. I'm going to update the code docs. Thanks for pointing that out so clearly and easy to understand @armcknight.

philipphofmann: I know it's a bit weird, but that's what my observations with MetricKit data revealed. I initially wrote the logic as you pointed out with A, but that didn't work properly. A also seemed more logical to me.

brustolin (approved):
LGTM

PR #3398, opened 2023-11-13
feat: Added support for arm64e

### LINKED ISSUE #3344: No member named '__fp' in 'struct __darwin_arm_thread_state64'
### Platform

macOS

### Installed

Swift Package Manager

### Version

8.13.1

### Steps to Reproduce

On macOS 13.5.2 using Xcode 15 running on an M1 processor
Open you Swift app
Use SPM to add sentry-cocoa to my macOS project.
Only add the Sentry and Sentry-Dynamic Package Package. I'm not using Swift-UI.
Compile the project

### Expected Result

Project successfully compiles

### Actual Result

Project fails to compile...  This is a dupe of https://github.com/getsentry/sentry-cocoa/issues/3210, which was closed without being resolved.

Sentry
SourcePackages/checkouts/sentry-cocoa/Sources/SentryCrash/Recording/Tools/SentryCrashCPU_arm64.c
SourcePackages/checkouts/sentry-cocoa/Sources/SentryCrash/Recording/Tools/SentryCrashCPU_arm64.c:56:41 No member named '__fp' in 'struct __darwin_arm_thread_state64'

SourcePackages/checkouts/sentry-cocoa/Sources/SentryCrash/Recording/Tools/SentryCrashCPU_arm64.c:66:41 No member named '__sp' in 'struct __darwin_arm_thread_state64'

SourcePackages/checkouts/sentry-cocoa/Sources/SentryCrash/Recording/Tools/SentryCrashCPU_arm64.c:76:41 No member named '__pc' in 'struct __darwin_arm_thread_state64'

SourcePackages/checkouts/sentry-cocoa/Sources/SentryCrash/Recording/Tools/SentryCrashCPU_arm64.c:86:41 No member named '__lr' in 'struct __darwin_arm_thread_state64'

SourcePackages/checkouts/sentry-cocoa/Sources/SentryCrash/Recording/Tools/SentryCrashCPU_arm64.c:126:45 No member named '__fp' in 'struct __darwin_arm_thread_state64'

SourcePackages/checkouts/sentry-cocoa/Sources/SentryCrash/Recording/Tools/SentryCrashCPU_arm64.c:128:45 No member named '__lr' in 'struct __darwin_arm_thread_state64'

SourcePackages/checkouts/sentry-cocoa/Sources/SentryCrash/Recording/Tools/SentryCrashCPU_arm64.c:130:45 No member named '__sp' in 'struct __darwin_arm_thread_state64'

SourcePackages/checkouts/sentry-cocoa/Sources/SentryCrash/Recording/Tools/SentryCrashCPU_arm64.c:132:45 No member named '__pc' in 'struct __darwin_arm_thread_state64'

I am willing but will likely not be able to.

brustolin:
## :scroll: Description

A developer may choose the arm64e architecture for their iOS app to benefit from enhanced security features. 
Sentry was not compiling for this architecture due to issues with accessing machine context properties

By using `arm_thread_state64_get_*` macros, we can read the necessary information.

closes #3344 

Samples

## :pencil: Checklist


brustolin:
Although the project is now compiling, when we run `profiling` it crashes, I couldn't figure it out the problem.

@armcknight maybe you would like to take a look at this?
Try the `iOS13-Swift` sample. It needs to run on a real device.

Samples/iOS-Swift/iOS13-Swift/SwiftUIView.swift
```diff
@@ -1,7 +1,7 @@
 import SentrySwiftUI
 import SwiftUI
 
-struct SwiftUI: View {
+struct SwiftUIView: View {
```
philipphofmann, line 4: `l`: What was the reason you added this change in this PR? Couldn't this be in another PR?

philipphofmann (commented):
The changes look good, but CI is complaining.

Samples/iOS-Swift/iOS13-Swift/SwiftUIView.swift
brustolin: It could be, but `iOS13-swift` sample was not compiling, and I want to use it to test arm64e instead of `the iOS-Swift` sample that we use for everything.

armcknight:
Having a look. For reference, here is the crash: 
image

It's happening when creating a profiler instance when there's already one in memory which gets destructed.

brustolin:
> It's happening when creating a profiler instance when there's already one in memory which gets destructed.

Thanks! Do we have plans to fix this already?

armcknight:
Sorry for the late reply @brustolin ... we're currently investigating the root cause.

indragiek:
@armcknight @brustolin how do I repro this? I'm running on device with the iOS-Swift sample app and starting/ending profiles and I can't get it to crash.

brustolin:
> @armcknight @brustolin how do I repro this? I'm running on device with the iOS-Swift sample app and starting/ending profiles and I can't get it to crash.

Are you running it on a real device? Simulators don't crash. Run the app and start a transaction, it crashes for me.

indragiek:
@brustolin I can't repro on an iPhone 14 Pro Max running iOS-Swift at the head of this branch. Are you testing with an arm64e or arm64 device? I tried with multiple concurrent transactions as well and everything seems to work fine. @armcknight do you have an older device you can test with?

brustolin:
> @brustolin I can't repro on an iPhone 14 Pro Max running iOS-Swift at the head of this branch. Are you testing with an arm64e or arm64 device? I tried with multiple concurrent transactions as well and everything seems to work fine. @armcknight do you have an older device you can test with?

Ow, sorry. iOS-Swift is not configured to run with arm64e, I must have roll it back. It only crashes when I compile the sample for arm64e. I will commit the change.

indragiek:
@brustolin Thanks, able to repro now. It appears this is crashing within profiling code any place where `shared_ptr` is used 
 and free'd (`QueueMetadata`, `ThreadMetadataCache`, and `SamplingProfiler`) so it isn't directly related to any specific piece of code. Replacing `shared_ptr` with `unique_ptr` in all of these places fixes the crash but I don't think it's the right way to solve the problem. Still looking into it.

indragiek:
There is some relevant documentation on Apple's page about [arm64e pointer authentication](https://developer.apple.com/documentation/security/preparing_your_app_to_work_with_pointer_authentication):

> Pointer authentication can also expose latent bugs in existing code. In C++, it’s incorrect to call a virtual method using a declaration that differs from its definition. In practice, such calls typically succeed in arm64, but trigger a pointer authentication failure in arm64e. You might encounter this bug when using OS_OBJECT types like [dispatch_queue_t](https://developer.apple.com/documentation/dispatch/dispatch_queue_t) and [xpc_connection_t](https://developer.apple.com/documentation/xpc/xpc_connection_t). You can’t pass instances of these types from C++ code to an Objective-C++ function (or vice versa) because they’re defined differently in Objective-C++ to support automatic reference counting (ARC).

I suspect something like this is causing the crash.

armcknight:
> I don't think [replacing shared_ptr with unique_ptr in all of these places is] the right way to solve the problem

Why not? We would potentially have to use `std::move` in places that create new references to these, if any, if my understanding of `std::move` is correct (based on my reading of https://stackoverflow.com/a/6876833). Are there other considerations?

indragiek:
@armcknight No doubt we can make the change, I just would not be confident that it actually fixes the issue without understanding why using a `shared_ptr` here causes a crash, and only on a single architecture. If this is urgent and we need to ship a quick patch then we certainly can do that though.

brustolin:
@armcknight @indragiek Any progress to solve this crash?

armcknight:
@brustolin check out the changes Indragie pushed up recently. Do you still see the crash with those?

brustolin:
Still getting an error at `SamplingProfiler::~SamplingProfiler`:

image

image

armcknight:
@brustolin Would you mind pulling the commit I just added and trying again? That was a nasty merge you had to do, and it overwrote some of indragie's changes. Sorry about that. I just tried my test and it no longer crashes. The profiles I created both appeared in the dash, so things appear to still function normally.

armcknight:
@brustolin would you mind capturing your repro steps in a UI test so we can eliminate any variation between our machines and see what happens when we run it in CI?

armcknight:
@brustolin Oh, I forgot that you were originally testing with iOS13-Swift. Is there no way to do it in the normal iOS-Swift sample app so we can test it in a simulator? That sample is still set up to build for the arm64e target. Is there something special about iOS13-Swift?

brustolin:
The problem only occurs when the target is arm64e. This is the whole issue.

I don't want to change iOS-Swift because of this; we would have to test on real devices anyway.

brustolin:
Hello @armcknight and @indragiek.

This have a higher priority now. It's affecting Xcode SwiftUI visualiser for macOS projects as described in [here](https://github.com/getsentry/sentry-cocoa/issues/4350).

Do you have any suggestions what could we do?
Maybe turn of profiling for arm64e?

brustolin:
Awesome, Thanks @indragiek for the fix!!

Samples/iOS-Swift/iOS13-Swift/AppDelegate.swift
```diff
@@ -21,10 +21,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                options.enableMetricKit = true
             }
+            // Sampling 100% - In Production you probably want to adjust this
+            options.tracesSampleRate = 1.0
+            options.sessionTrackingIntervalMillis = 5_000
+            options.profilesSampleRate = 1.0
+            options.attachScreenshot = true
+            options.attachViewHierarchy = true
+            options.environment = "test-app"
+            options.enableTimeToFullDisplayTracing = true
```
philipphofmann, line 34: `l`: Did you add this on purpose?

philipphofmann (commented):
I leave the review of the CPP code to @armcknight cause I'm not that knowledgeable in CPP.

Sentry.xcodeproj/project.pbxproj
```diff
@@ -6683,6 +6676,7 @@
 		D84DAD522B17428D003CF120 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALLOW_TARGET_PLATFORM_SPECIALIZATION = YES;
```
armcknight, line 6679: Please also move this to an xcconfig

Sources/Sentry/SentrySamplingProfiler.cpp
```diff
@@ -21,7 +21,7 @@ namespace profiling {
             mach_port_t port;
             clock_serv_t clock;
             mach_timespec_t delaySpec;
-            std::shared_ptr<ThreadMetadataCache> cache;
+            ThreadMetadataCache *cache;
```
armcknight, line 24: Shouldn't this be a `std::unique_ptr<ThreadMetadataCache>` @indragiek ?

Sources/Sentry/include/SentryBacktrace.hpp
```diff
@@ -57,8 +57,8 @@ namespace profiling {
      * @param f The function to call for each entry.
      * @param cache The cache used to look up thread metadata.
      */
-    void enumerateBacktracesForAllThreads(const std::function<void(const Backtrace &)> &f,
-        const std::shared_ptr<ThreadMetadataCache> &cache);
+    void enumerateBacktracesForAllThreads(
+        const std::function<void(const Backtrace &)> &f, ThreadMetadataCache *cache);
```
armcknight, line 61: same question here on naked pointer vs `std::unique_ptr`

Tests/SentryProfilerTests/SentryBacktraceTests.mm
```diff
@@ -230,7 +230,7 @@ - (void)testCollectsMultiThreadBacktrace
                     }
                 }
             },
-            cache);
+            &cache);
```
armcknight, line 233: also unsure of the ramifications of this change and the one on 202 (i can't group them into the same comment)

Tests/SentryProfilerTests/SentryBacktraceTests.mm
```diff
@@ -199,7 +199,7 @@ - (void)testCollectsMultiThreadBacktrace
     XCTAssertEqual(
         pthread_create(&thread2, nullptr, threadEntry, reinterpret_cast<void *>(bc_d)), 0);
 
-    const auto cache = std::make_shared<ThreadMetadataCache>();
+    ThreadMetadataCache cache;
```
armcknight, line 202: Why can't we use `make_unique` here as we do on https://github.com/getsentry/sentry-cocoa/pull/3398/files#diff-0397b590fd7ff98685a61fa9a99cceccf30bdf224754ba5f92448abd0852885bR80

Tests/SentryProfilerTests/SentryBacktraceTests.mm
armcknight: OK, I missed that this is a test, but I guess I still am wondering how it works...

armcknight (commented):
I'm not clear on why we can change `std::shared_ptr` to naked pointers vs using `unique_ptr` everywhere (partly because it's used in _some_ places (https://github.com/getsentry/sentry-cocoa/pull/3398/files#diff-0397b590fd7ff98685a61fa9a99cceccf30bdf224754ba5f92448abd0852885bR80)... does C++ autobox naked pointers?)

Samples/iOS-Swift/iOS13-Swift/AppDelegate.swift
brustolin: Yeap

Samples/iOS-Swift/iOS13-SwiftTests/LaunchUITest.swift
```diff
@@ -25,6 +25,11 @@ class LaunchUITests: XCTestCase {
         XCTAssertTrue(app.staticTexts["SwiftUI!"].waitForExistence(timeout: timeout), "SwiftUI not loaded.")
     }
     
+    func testNavigationTransaction() {
+        app.buttons["Test Navigation Transaction"].tap()
+        XCTAssertEqual(app.state, .runningForeground)
+    }
```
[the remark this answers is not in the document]
brustolin: Yes, this trigger the crash that was detected with arm64e

Sentry.xcodeproj/project.pbxproj
brustolin: Strange, I think Xcode added this. I dont remember changing this.

Tests/SentryProfilerTests/SentryBacktraceTests.mm
brustolin: I changed this test to make it compile, after @indragiek update on the code this was broken. 
I was hoping you or @indragiek to check if this is the intended use.

Sources/Sentry/SentrySamplingProfiler.cpp
indragiek: It can't be because the pointer is managed by `SentrySamplingProfiler`. We can't copy a `unique_ptr` from `SentrySamplingProfiler` into this struct, it can only be moved (which we don't want to do, since that makes the original `unique_ptr` invalid). So I'm falling back to a regular pointer here and the contract is that `SentrySamplingProfiler` will keep that valid for as long as this thread exists.

Sources/Sentry/include/SentryBacktrace.hpp
indragiek: See response above.

Tests/SentryProfilerTests/SentryBacktraceTests.mm
indragiek: Yes, you could do this ^

Tests/SentryProfilerTests/SentryBacktraceTests.mm
indragiek: You could either leave it the way it is here:

```
ThreadMetadataCache cache;
...
&cache (passing cache as a pointer to a stack allocated object)
```

or do this which is what @armcknight is suggesting above:

```
const auto cache = std::make_unique<ThreadMetadataCache>();
...
cache.get() (passing cache as a pointer to a heap allocated object)
```

In this case either will work because the pointer does not escape the implementation of `enumerateBacktracesForAllThreads`

brustolin: I will leave the way it is then. Thanks

Sources/Sentry/SentrySamplingProfiler.cpp
armcknight: Ah ok, that's what I was really curious about, why we couldn't move it. Good to know!

armcknight (approved):
I'm satisfied with the answers to my last questions, let's ship it!

PR #3399, opened 2023-11-13
test: Improve SentryReachabilityTests

philipphofmann:
The SentryReachabilityTests are still sometimes failing in CI. We now skip registering and unregistering the actual callbacks to SCNetworkReachability for all tests except one to minimize side effects.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryReachability.m
+++ b/Sources/Sentry/SentryReachability.m
@@ -44,6 +44,7 @@ NSString *const SentryConnectivityNone = @"none";
 void
 SentrySetReachabilityIgnoreActualCallback(BOOL value)
 {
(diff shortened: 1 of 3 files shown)


Sources/Sentry/SentryReachability.m
```diff
@@ -155,6 +156,15 @@ + (void)initialize
...
 }
 
+- (instancetype)init
+{
+    if (self = [super init]) {
+        self.skipRegisteringActualCallbacks = NO;
+    }
+
+    return self;
+}
```
brustolin, line 166: m: can we use `#if TEST` instead of this to avoid the extra bytes and processing during release?

brustolin (approved):

PR #3390, opened 2023-11-08
fix: dont call through to NSLog from profiler sampling thread

armcknight:
fixes #3336 

Let's just avoid calling through to NSLog from the profiler's sampling thread until we can get a different logging solution in place, if we decide we need these logs at some point in the future.

PR #3405, opened 2023-11-14
fix: un-invert compilation condition

armcknight:
#3390 attempted to compile some NSLogs out of production builds to avoid a crash, but the compilation condition was accidentally inverted, so that it was only compiled _into_ prod builds. This fixes that.


indragiek (approved):

philipphofmann (approved):
LGTM

PR #3414, opened 2023-11-15
test: Add Nimble for test assertions

philipphofmann:
Convert user feedback tests to see them in action. I had to increase the deployment target for tests to iOS 13 cause Nimble requires a minimum version of iOS 13 or higher. That shouldn't be a problem because we can only run unit tests on iOS 13 or higher on GH actions, see https://github.com/getsentry/sentry-cocoa/blob/1ce939e8ac0eb2a9c3ced2d46fbaeee39351b758/.github/workflows/test.yml#L83-L88

### DIFF
--- a/Tests/SentryTests/Protocol/SentryUserFeedbackTests.swift
+++ b/Tests/SentryTests/Protocol/SentryUserFeedbackTests.swift
@@ -1,3 +1,4 @@
+import Nimble
 import XCTest
 
 class SentryUserFeedbackTests: XCTestCase {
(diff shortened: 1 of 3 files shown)


Tests/SentryTests/Protocol/SentryUserFeedbackTests.swift
```diff
@@ -18,19 +19,19 @@ class SentryUserFeedbackTests: XCTestCase {
         
         let actual = userFeedback.serialize()
         
-        XCTAssertEqual(userFeedback.eventId.sentryIdString, actual["event_id"] as? String)
```
kahest, line 21: `as? String` might not be necessary, assuming Nimble checks if the types match?

kahest (commented):
FWIW, I think it's *far better*. Only callouts I have is to make sure we don't write any new tests without it and opportunistically refactor

Tests/SentryTests/Protocol/SentryUserFeedbackTests.swift
philipphofmann: The actual is of type `[String:Any]`. That's why you need to cast it. It didn't compile without `as ? String`. If you know how to do it without, please tell me.

kahest: no was just a hunch - I guess if UserFeedback is a bridged ObjC type it won't work - no need to invest more time

armcknight (approved):
Looks good to me, excited to try these out for our tests.

PR #3418, opened 2023-11-15
chore: Improve bug template

philipphofmann:
Add environment and allow selecting multiple platforms for the bug template.

#skip-changelog

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -3,15 +3,31 @@ description: Tell us about something that's not working the way we (probably) in
 labels: ["Platform: Cocoa", "Type: Bug"]
 body:
   - type: dropdown
-    id: environment
+    id: platform
     attributes:
       label: Platform
       description: Which platform do you use?
+      multiple: true
       options:
         - iOS
         - tvOS
         - macOS
         - watchOS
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -3,15 +3,31 @@ description: Tell us about something that's not working the way we (probably) in
 labels: ["Platform: Cocoa", "Type: Bug"]
 body:
   - type: dropdown
-    id: environment
+    id: platform
```
philipphofmann, line 6: I have no clue why this was called environment before. I think it was a mistake.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -3,15 +3,31 @@ description: Tell us about something that's not working the way we (probably) in
...
     attributes:
       label: Platform
       description: Which platform do you use?
+      multiple: true
       options:
         - iOS
```
kahest, line 12: `m`: I think we should also add iPadOS to be sure - now that multi-select is possible it makes even more sense IMO

kahest (commented):
Great, thanks for fixing/adding! 🚀

brustolin (approved):

PR #3424, opened 2023-11-16
fix: Crash in SentryFramesTracker

philipphofmann:
## :scroll: Description

The SentryFramesTracker registers a callback to the CADisplayLink. While the SentryFramesTrackingIntegration keeps a reference to the SentryFramesTracker and calls SentryFramesTracker.stop on uninstall, there could be situations where the SentryFramesTracker is deallocated, but not calling invalidate on the CADisplayLink. Then the CADisplayLink still calls  and crashes with EXC_BAD_ACCESS cause the frames tracker is deallocated. To fix this problem SentryFramesTracker.dealloc now calls invalidate on the CADisplayLink.

A customer reported this crash, and we also see this in our [SDK crashes](https://sentry.sentry.io/issues/?project=4505469596663808&query=is%3Aunresolved+displaylinkcallback&referrer=issue-list&sort=user&statsPeriod=14d).

The actual crash is hard to reproduce, but I added a unit test to validate that dealloc calls invalidate.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Infinite loop when parsing MetricKit data (#3395)
 - Fix incorrect implementation in #3398 to work around a profiling crash (#3405)
+- Fix crash in SentryFramesTracker (#3424)
 
 ### Improvements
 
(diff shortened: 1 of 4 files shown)


armcknight (approved):
Looks good. The Xcode 15.0.1 tests keep timing out but shouldn't block this.

PR #3425, opened 2023-11-16
test: Skip callbacks for Swift Reachability Tests

philipphofmann:
Calling the methods of SCNetworkReachability in a tight loop from multiple threads is not an actual use case, and it leads to flaky test results. With this test, we want to test if the adding and removing observers are adequately synchronized and not if we call SCNetworkReachability correctly.

The reason for this PR is that testNoObservers still is flaky sometimes. See https://github.com/getsentry/sentry-cocoa/actions/runs/6888982334.

#skip-changelog


brustolin (approved):
LGTM

PR #3430, opened 2023-11-20
test: Fix failing profilingGPU test

philipphofmann:
The testProfilingGPUInfo test asserts that the SDK adds a frozen frame to the last captured transaction. Previously, the test switched between UIViewControllers when triggering an ANR, which led to multiple transactions created. Now, there is a button to trigger an ANR on the same UIViewController to start and stop the transaction manually so that the frozen frame caused by the ANR ends up in the manually created transaction. This fixes the failing test.

#skip-changelog


brustolin (approved):
LGTM

PR #3431, opened 2023-11-20
test: Assert network errors network tracker

philipphofmann:
Add assert message to network tracker tests to get more insight in case they fail.

#skip-changelog


brustolin (approved):
Code looks fine.
CI is not happy

philipphofmann:
Yes, it's because of this https://github.com/getsentry/sentry-cocoa/pull/3430.

PR #3433, opened 2023-11-20
ci: Remove Xcode 15.0.1 for tests

philipphofmann:
Unit tests with Xcode 15.0.1 are more flaky compared to other Xcode versions. This could be due to a problem with GH action runners; see https://github.com/actions/runner-images/issues/8693. We will add Xcode 15 back once the next bugfix for Xcode 15.1 is released.

#skip-changelog


kahest (approved):
Thank you! 🚀

PR #3439, opened 2023-11-22
fix: Inaccurate number of frames for transactions

philipphofmann:
## :scroll: Description

The total frames were too high for transactions because the SDK subscribed twice to the CADisplayLink. This is fixed now by only subscribing once. Furthermore, this PR adds a UI test to validate that the number of total frames is not entirely off.

I also added Nimble to the SwiftUI tests and had to raise the minimum deployment target of the UI tests to iOS 13, which is okay because we don't run our UI tests lower than iOS 13.

Somebody reported too high frame rates for transactions.

Unit tests, UI tests, and manual testing with a simulator.

## :pencil: Checklist

### DIFF
--- a/Samples/iOS-Swift/iOS-SwiftUITests/LaunchUITests.swift
+++ b/Samples/iOS-Swift/iOS-SwiftUITests/LaunchUITests.swift
@@ -1,3 +1,4 @@
+import Nimble
 import XCTest
 
 class LaunchUITests: BaseUITest {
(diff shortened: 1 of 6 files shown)


Sources/Sentry/SentryFramesTracker.m
```diff
@@ -93,7 +93,12 @@ - (void)resetProfilingTimestamps
 
 - (void)start
 {
+    if (_isRunning) {
+        return;
+    }
+
```
kahest, line 99: that is subtle 🔍

krystofwoldrich (approved):
Looks good 🚀

kahest (approved):
good catch, also thanks for the (tricky) UI tests

philipphofmann:
Thanks, @krystofwoldrich, but I am also waiting for @brustolin for the iOS-related part.

Samples/iOS-Swift/iOS-SwiftUITests/LaunchUITests.swift
```diff
@@ -80,7 +81,48 @@ class LaunchUITests: BaseUITest {
...
+        // Therefore, we add a one-second tolerance for the expected amount of frames.
+        // We only want to validate if the number of total frames is entirely off.
+        let maximumFramesPerSecond = Double(UIScreen.main.maximumFramesPerSecond)
+        
+        let expectedMinimumTotalFrames = (secondsBetween - 1.0) * maximumFramesPerSecond
+        let expectedMaximumTotalFrames = (secondsBetween + 1.0) * maximumFramesPerSecond
+        
+        let actualTotalFrames = Double(endTotalFrames - startTotalFrames)
+        
+        expect(actualTotalFrames) >= expectedMinimumTotalFrames
```
brustolin, line 124: This is kinda weird codewise speaking. 
A function call that is compared to a value, but the comparison result is "not used", so weird.

Not advocating for changing back to not use nimble. 
Just that, this is the first time I actually seeing this being used, it was weird.

brustolin (approved):
LGTM

Samples/iOS-Swift/iOS-SwiftUITests/LaunchUITests.swift
kahest: with Nimble you can also write it like this

> expect(actual).to(beGreaterThanOrEqualTo(expected))

but I prefer the shorthand, even though it needs a bit getting used to

PR #3440, opened 2023-11-22
fix: Finish transaction for external view controllers

### LINKED ISSUE #3434: Transactions Missing for ViewControllers
### Platform

iOS

### Environment

Develop

### Installed

CocoaPods

### Version

8.15.2

### Steps to Reproduce

sdk setup

```
      let sampleRate = NSNumber(value: self.configurationProvider.sampleRateNormalized)
      options.dsn = self.configurationProvider.dsn
      options.dist = self.configurationProvider.buildNumber
      options.environment = self.configurationProvider.environment
      options.releaseName = self.configurationProvider.releaseName
      options.sampleRate = sampleRate
      options.sendDefaultPii = self.configurationProvider.sendDefaultPII
      options.enableTracing = true
      options.attachStacktrace = true
      options.tracesSampleRate = sampleRate
      options.enableAutoSessionTracking = true
      options.enableAutoPerformanceTracing = true
      options.debug = true // Enabled debug when first installing is always helpful
```

"seeing log messages like: 
(contact me for details as it lists their App's ViewController names)"
```
The following UIViewControllers will generate automatic transactions: 
...
```

"Couple of strange things here:
1. there are by my count >200 VC files in the app and immediate MyApp's dependencies.
2. The [docs for automatic view controller ](https://docs.sentry.io/platforms/apple/performance/instrumentation/automatic-instrumentation/)tracing say However, the SDK doesn't capture transactions for UIViewControllers of third-party libraries or SwiftUI. But a few of these are from the third party Google Places pod (the ones starting with GMS)"

"I suspect part of this is that Sentry thinks that some of MyApp's own cocoapods are external dependencies and is thus excluding them. When I exclude VCs from these modules from the count there are about 40 VCs total in the main part of the app. However only 18 of these are showing up in the list of automatically instrumented VCs"

"Also some of the items in the list above are from MyApp's pods, so I'm a bit confused"

"I'm also seeing quite a lot of `Was waiting for timeout for UI event trace but it had no children, will not keep transaction.` in the logs when I navigate around the app."

"I also had set up some filters on the integrations using options.integrations , but I've removed that for now."

"Also tried options.enableUIViewControllerTracing = true"

"Ok I think part of it is that we need to add includes for the VCs not part of the main app bundle. This still doesn't fully explain the patterns we're seeing though"

### Expected Result

Transactions for their ViewControllers

### Actual Result

Transactions for 18 out of 40 ViewControllers.

_No response_

brustolin:
## :scroll: Description

View controllers from frameworks in `inAppIncludes` option are not being swizzled, and so, their transactions are not being closed, but we do start the transaction, because start happens with swizzling of `UIViewController`.

close #3434 

Sample, Unit Test

## :pencil: Checklist


Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -99,6 +99,19 @@ - (NSInteger)indexOfImage:(uint64_t)address
...
+- (nullable NSString *)pathForImage:(NSString *)binaryName
+{
+    @synchronized(self) {
+        for (SentryBinaryImageInfo *info in _cache) {
+            if ([[info.name.lastPathComponent lowercaseString]
+                    isEqualToString:binaryName.lowercaseString]) {
+                return info.name;
+            }
+        }
+    }
```
kahest, line 111: `m`: since this uses a cache, I'm assuming this is not performance-heavy?

Sources/Sentry/SentrySDK.m
```diff
@@ -160,10 +160,11 @@ + (void)startWithOptions:(SentryOptions *)options
     SENTRY_LOG_DEBUG(@"Dispatching init work required to run on main thread.");
     [SentryThreadWrapper onMainThread:^{
         SENTRY_LOG_DEBUG(@"SDK main thread init started...");
-        [SentrySDK installIntegrations];
 
         [SentryCrashWrapper.sharedInstance startBinaryImageCache];
         [SentryDependencyContainer.sharedInstance.binaryImageCache start];
+
+        [SentrySDK installIntegrations];
```
kahest, line 167: `h`: I'm assuming this was moved because it now needs the cache. However, we should be **extremely** cautious with changes like these, please double- and triple-check that this doesn't break things

kahest (commented):
I'll leave the detailled reviews to @armcknight and @philipphofmann, but we shouldn't rush this

Sources/Sentry/SentrySDK.m
brustolin: TBH this makes more sense than before, we definitely need SentryDependencyContainer ready before installing all integrations since the integrations uses the dependency container.

Sources/Sentry/SentryBinaryImageCache.m
brustolin: I wouldn't consider this performance-heavy, but this is just used once, so definitely not a problem.

Sources/Sentry/SentryUIViewControllerSwizzling.m
```diff
@@ -69,6 +71,15 @@ - (instancetype)initWithOptions:(SentryOptions *)options
 
 - (void)start
 {
+    SentryBinaryImageCache *imageCache = SentryDependencyContainer.sharedInstance.binaryImageCache;
+
+    for (NSString *string in self.inAppLogic.inAppIncludes) {
```
philipphofmann, line 76: `h`: Just double-checking why this works now: Previously, we only looked for UIViewControllers in the image of the app delegate class. Now we use inAppIncludes which per default includes the `CFBundleExecutable` which is basically the same image of the app delegate class. Could it happen that the UIApplication returned by `findApp` is not in the `CFBundleExecutable`? If yes, this would be a breaking change and we still should call `swizzleAllSubViewControllersInApp` after swizzling the inApp ones. There is no risk of double swizzling because if we already swizzled the UIViewController again swizzling it does nothing if I'm not mistaken.

Sources/Sentry/SentryUIViewControllerSwizzling.m
```diff
@@ -95,19 +106,6 @@ - (void)start
                                  @"a rootViewController");
             }
         }
-
-        [self swizzleAllSubViewControllersInApp:app];
-    } else {
-        // If we can't find an UIApplication instance we may use the current process path as the
-        // image name. This mostly happens with SwiftUI projects.
-        NSString *processImage = self.processInfoWrapper.processPath;
```
philipphofmann, line 103: `h`: Is it save to remove this fallback?

Sources/Sentry/SentryBinaryImageCache.m
philipphofmann: FYI @kahest: The cache is for keeping a record of the binary images. It gets updated when you add and remove binary images.

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -99,6 +99,19 @@ - (NSInteger)indexOfImage:(uint64_t)address
     return -1; // Address not found
 }
 
+- (nullable NSString *)pathForImage:(NSString *)binaryName
+{
+    @synchronized(self) {
+        for (SentryBinaryImageInfo *info in _cache) {
+            if ([[info.name.lastPathComponent lowercaseString]
+                    isEqualToString:binaryName.lowercaseString]) {
```
philipphofmann, line 107: `h`: The inAppInclude is a prefix. We must change this to using `hasPrefix`.
```suggestion
                    hasPrefix:binaryName.lowercaseString]) {
```

The InAppLogic does the same see
https://github.com/getsentry/sentry-cocoa/blob/b0646651e1c1e526071883841a6412e4f75f3690/Sources/Sentry/SentryInAppLogic.m#L35

I think it's a good idea to refactor the code so both this code here and the `InAppLogic` use the same code so we only have to change it once.

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -99,6 +99,19 @@ - (NSInteger)indexOfImage:(uint64_t)address
     return -1; // Address not found
 }
 
+- (nullable NSString *)pathForImage:(NSString *)binaryName
+{
+    @synchronized(self) {
+        for (SentryBinaryImageInfo *info in _cache) {
```
philipphofmann, line 105: `l`: This is `n*m` where `n` is the number of `inAppIncludes` and `m` is the number of binary images. Having a large number of inAppIncludes in a large app could take some time. But I guess it should be fine for now, as we compare by prefix, so building a dictionary for faster lookup to get this to m is a bit tricky and not worth the effort right now, IMO. WDTY?

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -99,6 +99,19 @@ - (NSInteger)indexOfImage:(uint64_t)address
     return -1; // Address not found
 }
 
+- (nullable NSString *)pathForImage:(NSString *)binaryName
```
philipphofmann, line 102: `m`: Although most of the code here is covered by tests in `SentryUIViewControllerSwizzlingTests`, I think we should add some tests to `SentryBinaryImageCacheTests` to validate that we use `hasPrefix` and that this method returns `nil` if we don't find anything.

Sources/Sentry/SentryUIViewControllerSwizzling.m
```diff
@@ -69,6 +71,15 @@ - (instancetype)initWithOptions:(SentryOptions *)options
 
 - (void)start
 {
+    SentryBinaryImageCache *imageCache = SentryDependencyContainer.sharedInstance.binaryImageCache;
+
+    for (NSString *string in self.inAppLogic.inAppIncludes) {
+        NSString *pathToImage = [imageCache pathForImage:string];
+        if (pathToImage != nil) {
+            [self swizzleUIViewControllersOfImage:pathToImage];
+        }
```
philipphofmann, line 80: `l`: I think maybe logging a warning that we tried to swizzle UIViewControllers for the inAppInclude, but we couldn't find the image path for it, would be excellent.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
```diff
@@ -56,6 +63,20 @@ class SentryUIViewControllerSwizzlingTests: XCTestCase {
         super.tearDown()
         clearTestState()
     }
+    
+    func testExternalViewControllerImage() {
+        //Test to ensure ExternalUIViewController exists in an external lib
```
philipphofmann, line 68: Excellent idea 🥇

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
```diff
@@ -56,6 +63,20 @@ class SentryUIViewControllerSwizzlingTests: XCTestCase {
...
+        //just in case someone changes the settings of the `SentryTestUtils` lib
+        let imageName = String(
+            cString: class_getImageName(SentryUIViewControllerSwizzlingTests.self)!,
+            encoding: .utf8)! as NSString
+        
+        let externalImageName = String(
+            cString: class_getImageName(ExternalUIViewController.self)!,
+            encoding: .utf8)! as NSString
+        
+        XCTAssertNotEqual(externalImageName, imageName)
```
philipphofmann, line 78: `l`: I think we could add an assertion failure message explaining what the problem is.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
```diff
@@ -123,6 +153,14 @@ class SentryUIViewControllerSwizzlingTests: XCTestCase {
         }
     }
     
+    func testSwizzlingOfExternalLibs() {
```
philipphofmann, line 156: Wow so simple 👏

philipphofmann (commented):
The PR looks great. Thanks for fixing this. There are some issue to address before we can merge this @brustolin.

Sources/Sentry/SentryUIViewControllerSwizzling.m
```diff
@@ -69,6 +74,16 @@ - (instancetype)initWithOptions:(SentryOptions *)options
 
 - (void)start
 {
+    for (NSString *include in self.inAppLogic.inAppIncludes) {
+        NSString *pathToImage = [self.binaryImageCache pathForImage:include];
+        if (pathToImage != nil) {
+            [self swizzleUIViewControllersOfImage:pathToImage];
+        } else {
+            SENTRY_LOG_DEBUG(
+                @"UIViewControllerSwizzling: Failed to find binary image for %@ module.", include);
```
philipphofmann, line 83: `l`: I think this should be `SENTRY_LOG_WARN`. You don't need to add `UIViewControllerSwizzling` to the log message, as every log message includes the class and the line number. We could make the message more descriptive so the user knows what is going on: `Failed to find the binary image for inAppInclude <@%> and, therefore can't instrument UIViewControllers in that binary.`

Sources/Sentry/SentryBinaryImageCache.m
philipphofmann: I can't find any tests for what I specified above ⬆️. @brustolin, am I missing something, or did you resolve this comment by mistake?

brustolin:
>I can't find any tests for what I specified above ⬆️. @brustolin, am I missing something, or did you resolve this comment by mistake?

https://github.com/getsentry/sentry-cocoa/blob/cff8050429f3844e332dbf93aafd85769f2fbebc/Tests/SentryTests/SentryBinaryImageCacheTests.swift#L100-L109

Sources/Sentry/SentryBinaryImageCache.m
brustolin: I don't think we need the refactoring right now, this is called once when the app starts, even with 500 libs, and 10 items in the `inAppInclude` properties, this only takes microseconds.

If we ever use this function regularly during the session we may consider improving it.

Sources/Sentry/SentryBinaryImageCache.m
brustolin: I didn't notice this behaviour of inAppLogic, but I have a problem with it.

If I add `Awesome` to my inAppLogic it will match the `Awesome` lib and also `Core-Awesome` `Data-Awesome` and `Private-Awesome`.

Sources/Sentry/SentryUIViewControllerSwizzling.m
brustolin, line 83:
```suggestion
            SENTRY_LOG_WARN(
                @"Failed to find the binary image for inAppInclude <@%> and, therefore can't instrument UIViewControllers in that binary", include);
```

Sources/Sentry/SentryUIViewControllerSwizzling.m
brustolin: IMO it is. the app binary itself will be in the `inAppInclude` automatically.

Sources/Sentry/SentryBinaryImageCache.m
brustolin: https://github.com/getsentry/sentry-cocoa/blob/cff8050429f3844e332dbf93aafd85769f2fbebc/Tests/SentryTests/SentryBinaryImageCacheTests.swift#L100-L109

Sources/Sentry/SentryUIViewControllerSwizzling.m
```diff
@@ -69,6 +74,17 @@ - (instancetype)initWithOptions:(SentryOptions *)options
 
 - (void)start
 {
+    for (NSString *include in self.inAppLogic.inAppIncludes) {
+        NSString *pathToImage = [self.binaryImageCache pathForImage:include];
+        if (pathToImage != nil) {
+            [self swizzleUIViewControllersOfImage:pathToImage];
+        } else {
+            SENTRY_LOG_WARN(@"Failed to find the binary image for inAppInclude <@%> and, therefore "
```
brustolin, line 82:
```suggestion
            SENTRY_LOG_WARN(@"Failed to find the binary image for inAppInclude \<@%\> and, therefore "
```

Sources/Sentry/SentryBinaryImageCache.m
philipphofmann: No, it won't. The logic uses `hasPrefix` on the `lastPathComponent`. If you add `Awesome` to your inAppLogic `Core-Awesome`, `Data-Awesome`, and `Private-Awesome` won't be inApp, while `Awesome-Core` would be. I think we should stick to this logic.

Sources/Sentry/SentryUIViewControllerSwizzling.m
philipphofmann: I thought so, too. Just wanted to double-check.

Sources/Sentry/SentryUIViewControllerSwizzling.m
philipphofmann: @brustolin, please answer the question above ⬆️.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
```diff
@@ -11,30 +11,39 @@ class SentryUIViewControllerSwizzlingTests: XCTestCase {
         let objcRuntimeWrapper = SentryTestObjCRuntimeWrapper()
         let subClassFinder: TestSubClassFinder
         let processInfoWrapper = SentryNSProcessInfoWrapper()
+        let binaryImageCache: SentryBinaryImageCache
         
         init() {
             subClassFinder = TestSubClassFinder(dispatchQueue: dispatchQueue, objcRuntimeWrapper: objcRuntimeWrapper)
+            binaryImageCache = SentryDependencyContainer.sharedInstance().binaryImageCache
```
armcknight, line 18: What's the point of doing this? Why add another parameter to SentryUIViewControllerSwizzling where it was already using the singleton from SentryDependencyContainer, just to assign it to the same value again in these tests? It's not even accessed anywhere in the tests.

Sentry.xcodeproj/project.pbxproj
```diff
@@ -5261,7 +5265,7 @@
 				DEVELOPMENT_TEAM = "";
 				EXECUTABLE_PREFIX = lib;
 				GCC_C_LANGUAGE_STANDARD = gnu99;
-				MACH_O_TYPE = staticlib;
+				MACH_O_TYPE = mh_dylib;
```
armcknight, line 5268: Please revert these, this is not a dynamic library. I can't remember what it was anymore, but something was previously broken due to it being mistyped, until I changed it to `staticlib`.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
```diff
@@ -85,7 +108,16 @@ class SentryUIViewControllerSwizzlingTests: XCTestCase {
         fixture.sut.start()
         let controller = TestViewController()
         controller.loadView()
+        let span = SentrySDK.span
+        
+        //To finish the transaction we need to finish `initialDisplay` span
+        //by calling `viewWillAppear` and reporting a new frame
+        controller.viewWillAppear(false)
+        Dynamic(SentryDependencyContainer.sharedInstance().framesTracker).reportNewFrame()
```
armcknight, line 116: Is this to cause `SentryTimeToDisplayTracker.framesTrackerHasNewFrame` to be called, ending the span it's managing? If so, please mention it here, or whatever other side effect is desired.

armcknight (commented):
Some quick questions/suggestions!

Sentry.xcodeproj/project.pbxproj
brustolin: This is SentryTestUtils and we need it to be dynamic to test our integration with dynamic libs. 
This is just an auxiliary lib, no reason it could not be dynamic lib.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
brustolin: Following @philipphofmann suggestion to use injection in the constructor instead of making direct use of `SentryDependencyContainer`

philipphofmann: Yep, that's following our decision on how to handle [dependency injection](https://github.com/getsentry/sentry-cocoa/blob/main/develop-docs/DECISIONS.md#dependency-injection-strategy). I'm a fan of specifying all dependencies via their constructor to improve testability. If you disagree, I'm happy to discuss this again, but not here in this PR 😃

philipphofmann (approved):
LGTM, after you either include or ditch my suggestion to extract the logic of calling `hasPrefix` to a function in https://github.com/getsentry/sentry-cocoa/pull/3452.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
armcknight: i mean, we're not injecting anything, we just added it and aren't using it. yagni, do this if/when we actually need to inject a mock

Sentry.xcodeproj/project.pbxproj
armcknight: I just remembered what this was breaking, it causes the tvOS unit tests to crash when I run it locally: 
```
2023-11-28 11:21:45.982597-0900 xctest[15149:9584462] [general] Error loading /Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/SentryProfilerTests.xctest/SentryProfilerTests (177):  dlopen(/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/SentryProfilerTests.xctest/SentryProfilerTests, 0x0109): Library not loaded: /Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a
  Referenced from: <2D742D1E-D6A4-37DB-88B5-1F09B59C9473> /Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/SentryProfilerTests.xctest/SentryProfilerTests
  Reason: tried: '/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' (code signature in <F4A23AD7-1604-3FA6-94F5-463FFC70C956> '/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' not valid for use in process: Trying to load an unsigned library), '/Applications/Xcode15/Xcode.app/Contents/Developer/Platforms/AppleTVSimulator.platform/Developer/usr/lib/libSentryTestUtils.a' (no such file), '/Library/Developer/CoreSimulator/Volumes/tvOS_21J353/Library/Developer/CoreSimulator/Profiles/Runtimes/tvOS 17.0.simruntime/Contents/Resources/RuntimeRoot/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' (no such file), '/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' (code signature in <F4A23AD7-1604-3FA6-94F5-463FFC70C956> '/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' not valid for use in process: Trying to load an unsigned library), '/Library/Developer/CoreSimulator/Volumes/tvOS_21J353/Library/Developer/CoreSimulator/Profiles/Runtimes/tvOS 17.0.simruntime/Contents/Resources/RuntimeRoot/usr/lib/libSentryTestUtils.a' (no such file)
2023-11-28 11:21:46.265470-0900 xctest[15149:9584462] [Default] The bundle “SentryProfilerTests.xctest” couldn’t be loaded. Try reinstalling the bundle.
2023-11-28 11:21:46.265640-0900 xctest[15149:9584462] [Default] (dlopen(/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/SentryProfilerTests.xctest/SentryProfilerTests, 0x0109): Library not loaded: /Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a
  Referenced from: <2D742D1E-D6A4-37DB-88B5-1F09B59C9473> /Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/SentryProfilerTests.xctest/SentryProfilerTests
  Reason: tried: '/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' (code signature in <F4A23AD7-1604-3FA6-94F5-463FFC70C956> '/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' not valid for use in process: Trying to load an unsigned library), '/Applications/Xcode15/Xcode.app/Contents/Developer/Platforms/AppleTVSimulator.platform/Developer/usr/lib/libSentryTestUtils.a' (no such file), '/Library/Developer/CoreSimulator/Volumes/tvOS_21J353/Library/Developer/CoreSimulator/Profiles/Runtimes/tvOS 17.0.simruntime/Contents/Resources/RuntimeRoot/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' (no such file), '/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' (code signature in <F4A23AD7-1604-3FA6-94F5-463FFC70C956> '/Users/andrewmcknight/Library/Developer/Xcode/DerivedData/Sentry-gnnqqmcfqpqmmhcjxfzwadgjajnt/Build/Products/Test-appletvsimulator/libSentryTestUtils.a' not valid for use in process: Trying to load an unsigned library), '/Library/Developer/CoreSimulator/Volumes/tvOS_21J353/Library/Developer/CoreSimulator/Profiles/Runtimes/tvOS 17.0.simruntime/Contents/Resources/RuntimeRoot/usr/lib/libSentryTestUtils.a' (no such file))
Program ended with exit code: 83
``` 

this is then fixed when I change it back to a static lib. There are actual differences between static and dynamic libs, they are not interchangeable.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
armcknight: So the reason I asked about `SentryTimeToDisplayTracker.framesTrackerHasNewFrame` is because when I looked at `SentryTimeToDisplayTracker.reportNewFrame` and clicked through to the delegate call, one of the options that comes up is `SentryTimeToDisplayTracker.framesTrackerHasNewFrame`:
image

What is the actual codepath through which `SentryTimeToDisplayTracker.framesTrackerHasNewFrame` is called?

armcknight (dismissed):
Please do not merge the change declaring SentryTestUtils as a dynamic lib, as it breaks our ability to run some unit tests locally. Or let me know if that's not the case for anyone else, and I can look into why it's happening on my machine.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
brustolin: Yes, you're right, I misread the class name you wrote. the call reportNewFrame() is a function of `SentryFramesTracker` that will call the `framesTrackerHasNewFrame` from its delegates.

Sentry.xcodeproj/project.pbxproj
brustolin: Damn Apple!! Why does this only fail for tvOS?
Why they need a signed lib to run for tests. Such BS.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerSwizzlingTests.swift
```diff
@@ -85,7 +109,16 @@ class SentryUIViewControllerSwizzlingTests: XCTestCase {
...
```
brustolin, line 117:
```suggestion
        //This will call SentryTimeToDisplayTracker.framesTrackerHasNewFrame and finish the span its managing.
        Dynamic(SentryDependencyContainer.sharedInstance().framesTracker).reportNewFrame()
```

Sentry.xcodeproj/project.pbxproj
armcknight: Yep, mysterious to me as well as to why it's like this. Sorry it was an extra pain.

PR #3443, opened 2023-11-23
fix: check for null value returned from pthread_mach_thread_np

armcknight:
We didn't completely fix the problem reported in https://github.com/getsentry/sentry-cocoa/issues/3354 with https://github.com/getsentry/sentry-cocoa/pull/3364. This changes the callsites to check for `NULL` before dereferencing the pointer.

I checked the sources and it does appear that `NULL` is a possible return value from `pthread_mach_thread_np `: https://github.com/apple/darwin-libpthread/blob/2b46cbcc56ba33791296cd9714b2c90dae185ec7/src/pthread.c#L931-L937 and for `_pthread_is_valid `, https://opensource.apple.com/source/libpthread/libpthread-416.11.1/src/internal.h.

### DIFF
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -428,12 +428,16 @@ serializedProfileData(
               forTransaction:(SentryTransaction *)transaction;
 {
     payload[@"platform"] = transaction.platform;
+
+    const auto activeThreadID =
+        [transaction.trace.transactionContext sentry_threadInfo].threadId ?: @(-1);
     payload[@"transaction"] = @{
         @"id" : transaction.eventId.sentryIdString,
         @"trace_id" : transaction.trace.traceId.sentryIdString,
         @"name" : transaction.transaction,
-        @"active_thread_id" : [transaction.trace.transactionContext sentry_threadInfo].threadId
+        @"active_thread_id" : activeThreadID
     };
+
     const auto timestamp = transaction.trace.originalStartTimestamp;
     if (UNLIKELY(timestamp == nil)) {
         SENTRY_LOG_WARN(@"There was no start timestamp on the provided transaction. Falling back "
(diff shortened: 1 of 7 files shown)


Sources/Sentry/SentryProfiler.mm
```diff
@@ -428,12 +428,16 @@ + (void)updateProfilePayload:(NSMutableDictionary<NSString *, id> *)payload
...
     payload[@"platform"] = transaction.platform;
+
+    const auto activeThreadID =
+        [transaction.trace.transactionContext sentry_threadInfo].threadId ?: @(-1);
     payload[@"transaction"] = @{
         @"id" : transaction.eventId.sentryIdString,
         @"trace_id" : transaction.trace.traceId.sentryIdString,
         @"name" : transaction.transaction,
-        @"active_thread_id" : [transaction.trace.transactionContext sentry_threadInfo].threadId
+        @"active_thread_id" : activeThreadID
```
armcknight, line 438: I'm not sure what the right thing to put here would be.

Sources/Sentry/SentryThreadHandle.cpp
```diff
@@ -45,6 +45,9 @@ namespace profiling {
     ThreadHandle::current() noexcept
     {
         const auto thread = pthread_mach_thread_np(pthread_self());
+        if (thread == (mach_port_t)NULL) {
+            return nullptr;
+        }
```
armcknight, line 50: I don't think we actually need this, the crash was actually happening when `current()` is dereferenced (see the callsites in this diff)

Sources/Sentry/SentryProfiler.mm
philipphofmann: Do you mean if the thread ID is NULL? If yes, couldn't we just not set the `active_thread_id`?

philipphofmann (commented):
What's missing to finish this PR apart from tests, @armcknight?

armcknight:
> What's missing to finish this PR apart from tests, @armcknight?

The customer reported that this branch doesn't resolve the crash.

Sources/Sentry/SentryProfiler.mm
armcknight: @phacops Please let us know how we would proceed in this case. Maybe we just choose not to send the profile at all.

armcknight:
#3520 builds on this by adding more stringent and additional checks in other spots, so I'll close this in favor of moving that one forward.

PR #3445, opened 2023-11-23
ref: Add AutoSpanStarter

philipphofmann:
## :scroll: Description

Add a protocol `SentryAutoSpanStarter` for being able to change the strategy for how auto-generated spans are started. 

#skip-changelog

Currently, we start them from the span bound to the scope. I added this while [implementing carrier transactions](https://github.com/getsentry/sentry-cocoa/compare/feat/carrier-transactions?expand=1) https://github.com/getsentry/team-mobile/issues/157, but we decided to ditch them. Still, I don't want the refactoring to be lost because we are going to switch to span ingestion in the foreseeable future, and then adding a different `SentryAutoSpanStarter` will be handy.

Unit tests.

## :pencil: Checklist


Sources/Sentry/SentryAutoSpanOnScopeStarter.m
```diff
@@ -0,0 +1,17 @@
+#import "SentryAutoSpanOnScopeStarter.h"
+#import "SentryHub+Private.h"
+#import "SentrySDK+Private.h"
+#import "SentryScope+Private.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@implementation SentryAutoSpanOnScopeStarter
+
+- (void)startSpan:(SentrySpanCallback)callback
```
brustolin, line 10: m: Im not sure about this name.
The function actually don't start any span, it gives you the span in scope to use it as necessary, and not necessarily to start another span.

brustolin (approved):
The code looks good, but I don't get the necessity for it.

A new protocol was introduced, and with it the necessity to refactoring many classes with a new initialisation arguments and new properties, but in regard to behaviour, nothing changed, it didn't get more simple, on the other hand, now is more complicated.

Im approving to let you decide what to do about this. If you're sure that this will be good, you can go ahead and merge it.

philipphofmann:
That's a fair point 🤔 , @brustolin. I can use the changes of this PR for adding span ingestion if required. But only when we start adding it, this PR would make sense. Until then, it would complicate things for nothing. Therefore, I close it.

PR #3455, opened 2023-11-29
fix: Crash when UINavigationController without rootViewController

wbcyclist:
When a UINavigationController is not initialized with initWithRootViewController, it leads to a crash.
like this: `self.window.rootViewController = [[UINavigationController alloc] init];`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+
+### Fixes 
+
 - Crash when UINavigationController doesn't have rootViewController (#3455)
 
 ## 8.17.0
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Crash when UINavigationController doesn't have rootViewController (#3455)
```
brustolin, line 5:
```suggestion

### Fixes 

- Crash when UINavigationController doesn't have rootViewController (#3455)
```

brustolin (approved):

PR #3459, opened 2023-11-29
fix: Upload to TestFlight

brustolin:
Adding a new framework to iOS-Swift sample broke TestFlight upload because of some bundle identifier collision.

We don't really need this in the samples right now, we have a unit test for it.

In the future if the necessity changes we try to figure it out the identifier collision.

_#skip-changelog_


philipphofmann (approved):
LGTM

PR #3469, opened 2023-11-30
fix: Check for NULL in binary image cache

philipphofmann:
## :scroll: Description

Properly validate the binary image name when converting it to an NSString to avoid crashes.

I came across this when looking at https://github.com/getsentry/sentry-cocoa/pull/3468.

Unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryBinaryImageCache.m
+++ b/Sources/Sentry/SentryBinaryImageCache.m
@@ -2,6 +2,7 @@
 #import "SentryCrashBinaryImageCache.h"
 #import "SentryDependencyContainer.h"
 #import "SentryInAppLogic.h"
+#import "SentryLog.h"
 
 static void binaryImageWasAdded(const SentryCrashBinaryImage *image);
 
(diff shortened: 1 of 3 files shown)


Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -35,8 +36,21 @@ - (void)stop
 
 - (void)binaryImageAdded:(const SentryCrashBinaryImage *)image
 {
+    if (image->name == NULL) {
```
armcknight, line 39: Is it possible for `image` to also be `NULL`? If `nextNode->image` for whatever reason in NULL, then we could pass a NULL pointer and dereferencing here would crash. https://github.com/getsentry/sentry-cocoa/blob/24c239b8e33dc88bfc70f588069a029eb00b4165/Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c#L143-L147

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -35,8 +36,21 @@ - (void)stop
...
+        SENTRY_LOG_WARN(@"The image name was NULL. Can't add image to cache.");
+        return;
+    }
+
+    NSString *imageName = [NSString stringWithCString:image->name encoding:NSUTF8StringEncoding];
```
armcknight, line 44: Are we sure `image->name` will never be NULL? That can crash this constructor.

armcknight (commented):
Couple questions on possible additional NULL checks we might need.

Sources/Sentry/SentryBinaryImageCache.m
philipphofmann: I think it shouldn't but a check for NULL does not harm. Thanks for pointing that out.

philipphofmann: I think it shouldn't but a check for NULL does not harm. Thanks for pointing that out.

Sources/Sentry/SentryBinaryImageCache.m
philipphofmann: I added a check for that above `if (image->name == NULL)`. If it's `NULL` we return.

brustolin (approved):
LGTM

But, can this ever happen, the system calling binaryWasAdded but passing NULL?

Issue #3460, opened 2023-11-29
Provide a sentry SPM package without UIKit


Descartess:
### Problem Statement

We use sentry over SPM and we would like an out-of-the-box sentry package without UIKit available.  Is this possible?

### Solution Brainstorm

We could duplicate the current target in` Package.swift` and set relevant GCC_PRE_PROCESSOR_DEFINITIONS as follows 

```
.target(
            name: "SentryNoUIKit",
            dependencies: ["SentryPrivate"],
            path: "Sources",
            sources: [
                "Sentry/",
                "SentryCrash/"
            ],
            publicHeadersPath: "Sentry/Public/",
            cxxSettings: [
                .define("SENTRY_NO_UIKIT", to: "1"),
                .define("GCC_ENABLE_CPP_EXCEPTIONS", to: "YES"),
                .headerSearchPath("Sentry/include"),
                .headerSearchPath("Sentry/include/HybridPublic"),
                .headerSearchPath("Sentry/Public"),
                .headerSearchPath("SentryCrash/Installations"),
                .headerSearchPath("SentryCrash/Recording"),
                .headerSearchPath("SentryCrash/Recording/Monitors"),
                .headerSearchPath("SentryCrash/Recording/Tools"),
                .headerSearchPath("SentryCrash/Reporting/Filters"),
                .headerSearchPath("SentryCrash/Reporting/Filters/Tools"),
                .headerSearchPath("SentryCrash/Reporting/Tools")
            ]
        ),
        ```

### Are you willing to submit a PR?

Yes

philipphofmann:
Since v [8.14.1](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8140), you should be able to use the SDK without linking UIKit. You shouldn't have to use an extra SPM package. Please reopen the issue if that doesn't solve your problem.

philipphofmann:
@armcknight, you came up with the fix. I think nothing is required from the user side if I'm not mistaken. If not, can you please explain? We must put this info into the docs as well if we didn't do this already.

philipphofmann:
Yes, changelog makes sense 👍. 

> We might have to refactor UIKit-based code into a separate target from the Sentry framework target, something like SentryUIKit, as we initially had discussed as a possibility, so it can be omitted from SPM linkages.

[...]

PR #3470, opened 2023-11-30
meta: mention experimental nature of UIKitless feature

armcknight:
There is some confusion over the availability of the UIKitless feature, like in https://github.com/getsentry/sentry-cocoa/issues/3460

Point out this is still an experimental feature not yet compatible with SPM.


brustolin (approved):

philipphofmann (approved):
Thanks, @armcknight 👍

PR #3473, opened 2023-12-01
feat(internal): change collectProfilerForTrace result to mutable

vaind:
When testing the native profiling integration for iOS in .NET, same way I've done it in Flutter, I've found I could use the collectProfileOutput directly without copying, which would, just if it was exposed as the NSMutableDictionary that it actually is.

See https://github.com/getsentry/sentry-dotnet/pull/2930/files#diff-1b71762e3c3bd42c9d55060467f9fcbc82b733f625ee6310ee76a6ff35a9e2d1R39

#skip-changelog - this probably doesn't make sense to mention to standard SDK consumers

### DIFF
--- a/Sources/Sentry/PrivateSentrySDKOnly.mm
+++ b/Sources/Sentry/PrivateSentrySDKOnly.mm
@@ -129,9 +129,9 @@ static BOOL _framesTrackingMeasurementHybridSDKMode = NO;
     return SentryDependencyContainer.sharedInstance.dateProvider.systemTime;
 }
 
-+ (nullable NSDictionary<NSString *, id> *)collectProfileBetween:(uint64_t)startSystemTime
-                                                             and:(uint64_t)endSystemTime
-                                                        forTrace:(SentryId *)traceId;
++ (nullable NSMutableDictionary<NSString *, id> *)collectProfileBetween:(uint64_t)startSystemTime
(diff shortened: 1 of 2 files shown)


armcknight (approved):
Should be fine. Not sure about how C# works with it, but in ObjC, if you put a NSMutableDictionary inside a NSDictionary reference, you don't even have to call mutableCopy on it... you could just cast it back to a NSMutableDictionary reference... the underlying object type never changed, just the reference type you're using. (That's why it's so important to use the `copy` directive on ObjC `@property` declarations for types that have mutable counterparts; otherwise, you could internally store a NSMutableDictionary, and even if you only expose it as a NSDictionary, consumers could cast it back to the mutable type and mutate it.)

vaind:
> Should be fine. Not sure about how C# works with it, but in ObjC, if you put a NSMutableDictionary inside a NSDictionary reference, you don't even have to call mutableCopy on it... you could just cast it back to a NSMutableDictionary reference... the underlying object type never changed, just the reference type you're using. (That's why it's so important to use the `copy` directive on ObjC `@property` declarations for types that have mutable counterparts; otherwise, you could internally store a NSMutableDictionary, and even if you only expose it as a NSDictionary, consumers could cast it back to the mutable type and mutate it.)

Yeah, I tried casting first but that didn't compile. Something to do with the bridging code I guess.

PR #3477, opened 2023-12-05
fix: simplify dispatch queue wrapper access

### LINKED ISSUE #3476: TTFD timeout timer never scheduled
### Description

A customer reported that if they never call `reportFullyDisplayed`, they don't get a transaction. The headerdoc for `enableTimeToFullDisplayTracing` state:
```
 * If @c SentrySDK.reportFullyDisplayed() is not called, the transaction will finish
 * automatically after 30 seconds and the `Time to full display` Span will be
 * finished with @c DeadlineExceeded status.
```
But I was able to repro this, and it turns out that when scheduling the timeout timer for the span in `startDeadlineTimer`, `_configuration.dispatchQueueWrapper` is `nil` so no deadline timer ever fires.

armcknight:
A possible solution to fix #3476. There are several places that we start a transaction with a tracer configuration that lacks a dispatch queue wrapper here: https://github.com/getsentry/sentry-cocoa/blob/b9a9dcab37e2e5ebe18ca868743047c4d385e5ae/Sources/Sentry/SentryHub.m#L358-L365

and here: https://github.com/getsentry/sentry-cocoa/blob/b9a9dcab37e2e5ebe18ca868743047c4d385e5ae/Sources/Sentry/SentryPerformanceTracker.m#L83-L90 

In the spot where we do actually provide it here: https://github.com/getsentry/sentry-cocoa/blob/b9a9dcab37e2e5ebe18ca868743047c4d385e5ae/Sources/Sentry/SentryUIEventTrackerTransactionMode.m#L86-L95

that reference is actually just set to the instance held by the dependency container here: https://github.com/getsentry/sentry-cocoa/blob/b9a9dcab37e2e5ebe18ca868743047c4d385e5ae/Sources/Sentry/SentryUIEventTrackingIntegration.m#L28-L30

I know we have an ongoing debate about when and where to use DI vs singletons, but let's avoid having this mixture, and take time to completely refactor one dependency at a time to remove the reference completely from SentryDependencyContainer convert to DI in all codepaths.


Sources/Sentry/SentryTracer.m
```diff
@@ -195,32 +195,36 @@ - (void)dispatchIdleTimeout
 {
     @synchronized(_idleTimeoutLock) {
         if (_idleTimeoutBlock != NULL) {
-            [_configuration.dispatchQueueWrapper dispatchCancel:_idleTimeoutBlock];
+            [SentryDependencyContainer.sharedInstance.dispatchQueueWrapper
```
philipphofmann, line 198: `h`: I would either add a `dispatchQueueWrapper` property to the SentryTracer and set it to `SentryDependencyContainer.sharedInstance.dispatchQueueWrapper` or ask for it via the init methods of the tracer cause it's a required dependency. I prefer the second one for testability as we once aligned [here](https://github.com/getsentry/sentry-cocoa/blob/main/develop-docs/DECISIONS.md#dependency-injection-strategy), but I would also be OK with the first option.

philipphofmann (commented):
Good catch, @armcknight. Thanks for fixing this.

Sources/Sentry/SentryTracer.m
armcknight: I would approve a PR if you opened one showing me what you mean though

philipphofmann: I opened a PR https://github.com/getsentry/sentry-cocoa/pull/3497.

armcknight: Merged that!

philipphofmann (approved):
LGTM, if you add a changelog entry @armcknight. You could also change the PR's title to reflect the user benefit instead of the detailed technical change.

armcknight:
The thread sanitizer is in an unrelated test case, and failed recently on `main`: https://github.com/getsentry/sentry-cocoa/actions/runs/7263319984/job/19788493775#step:7:54

armcknight:
See https://github.com/getsentry/sentry-cocoa/issues/3528 inre the tsan failure

PR #3478, opened 2023-12-05
ref: Send frames for all spans

philipphofmann:
Initially, we wanted to send frame data only for spans started on the main thread, but we decided that the frame data makes sense for all spans. This PR fixes that.

### DIFF
--- a/Sources/Sentry/SentrySpan.m
+++ b/Sources/Sentry/SentrySpan.m
@@ -57,21 +57,20 @@ SentrySpan ()
 
         if ([NSThread isMainThread]) {
             _data[SPAN_DATA_THREAD_NAME] = @"main";
-
-#if SENTRY_HAS_UIKIT
-            // Only track frames if running on main thread.
-            _framesTracker = framesTracker;
-            if (_framesTracker.isRunning) {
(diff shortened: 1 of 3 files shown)


Sources/Sentry/SentrySpan.m
```diff
@@ -57,22 +57,21 @@ - (instancetype)initWithContext:(SentrySpanContext *)context
...
-                initFrozenFrames = currentFrames.frozen;
-            }
-#endif // SENTRY_HAS_UIKIT
         } else {
             _data[SPAN_DATA_THREAD_NAME] = [SentryDependencyContainer.sharedInstance.threadInspector
                 getThreadName:currentThread];
         }
 
+#if SENTRY_HAS_UIKIT
+        _framesTracker = framesTracker;
```
armcknight, line 66: Does this need any kind of synchronization now that it's multithreaded?

armcknight (approved):
Looks mostly fine, with one threading question

Sources/Sentry/SentrySpan.m
armcknight: @philipphofmann please look into this and open a subsequent PR if necessary.

Issue #3481, opened 2023-12-05
Add frames delay for transactions and spans


philipphofmann:
### Description

See https://github.com/getsentry/team-mobile/issues/160.

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-cocoa/pull/3482
- [ ] https://github.com/getsentry/sentry-cocoa/pull/3483
- [ ] https://github.com/getsentry/sentry-cocoa/pull/3487
- [ ] https://github.com/getsentry/sentry-cocoa/pull/3496
```

PR #3483, opened 2023-12-05
ref: Add nanosecondsToTimeInterval

philipphofmann:
SentryTime has timeIntervalToNanoseconds but not nanosecondsToTimeInterval. This PR adds nanosecondsToTimeInterval. 
Furthermore, added some tests for timeIntervalToNanoseconds  and fixed the max value check. This method is required for frame delay https://github.com/getsentry/sentry-cocoa/issues/3481.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryTime.mm
+++ b/Sources/Sentry/SentryTime.mm
@@ -10,9 +10,15 @@ uint64_t
 timeIntervalToNanoseconds(double seconds)
 {
     NSCAssert(seconds >= 0, @"Seconds must be a positive value");
-    NSCAssert(seconds <= UINT64_MAX / 1e9,
+    NSCAssert(seconds <= (double)UINT64_MAX / (double)NSEC_PER_SEC,
         @"Value of seconds is too great; will overflow if casted to a uint64_t");
-    return (uint64_t)(seconds * 1e9);
+    return (uint64_t)(seconds * NSEC_PER_SEC);
+}
+
+double
+nanosecondsToTimeInterval(uint64_t nanoseconds)
+{
+    return (double)nanoseconds / NSEC_PER_SEC;
 }
 
 uint64_t
--- a/Sources/Sentry/include/SentryTime.h
+++ b/Sources/Sentry/include/SentryTime.h
@@ -12,6 +12,12 @@ SENTRY_EXTERN_C_BEGIN
 uint64_t timeIntervalToNanoseconds(double seconds);
 
 /**
+ * Converts nanoseconds to a @c double from a Cocoa API like @c -[NSDate
+ * @c timeIntervalSinceDate:] .
+ */
+double nanosecondsToTimeInterval(uint64_t nanoseconds);
+
+/**
  * Returns the absolute timestamp, which has no defined reference point or unit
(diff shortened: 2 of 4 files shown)


Sources/Sentry/SentryTime.mm
```diff
@@ -10,9 +10,15 @@
 timeIntervalToNanoseconds(double seconds)
 {
     NSCAssert(seconds >= 0, @"Seconds must be a positive value");
-    NSCAssert(seconds <= UINT64_MAX / 1e9,
+    NSCAssert(seconds <= (double)UINT64_MAX / (double)NSEC_PER_SEC,
         @"Value of seconds is too great; will overflow if casted to a uint64_t");
-    return (uint64_t)(seconds * 1e9);
+    return (uint64_t)(seconds * NSEC_PER_SEC);
```
armcknight, line 15: Nice 👍🏻

Sources/Sentry/include/SentryTime.h
```diff
@@ -11,6 +11,12 @@ SENTRY_EXTERN_C_BEGIN
  */
 uint64_t timeIntervalToNanoseconds(double seconds);
 
+/**
+ * Converts nanoseconds to a @c double from a Cocoa API like @c -[NSDate
+ * @c timeIntervalSinceDate:] .
+ */
+double nanosecondsToTimeInterval(uint64_t nanoseconds);
```
armcknight, line 18: Let's use actual `NSTimeInterval`, which is a `double` typedef and has its own docs that people can read.

```suggestion
/** Converts integer nanoseconds to a @c NSTimeInterval. */
NSTimeInterval nanosecondsToTimeInterval(uint64_t nanoseconds);
```

Sources/Sentry/SentryTime.mm
```diff
@@ -10,9 +10,15 @@
...
+}
+
+double
```
armcknight, line 18:
```suggestion
NSTimeInterval
```

Sources/Sentry/SentryTime.mm
```diff
@@ -10,9 +10,15 @@
 timeIntervalToNanoseconds(double seconds)
 {
     NSCAssert(seconds >= 0, @"Seconds must be a positive value");
-    NSCAssert(seconds <= UINT64_MAX / 1e9,
+    NSCAssert(seconds <= (double)UINT64_MAX / (double)NSEC_PER_SEC,
```
armcknight, line 13:
```suggestion
    NSCAssert(seconds <= (NSTimeInterval)UINT64_MAX / (NSTimeInterval)NSEC_PER_SEC,
```

armcknight (approved):
Approved with some suggestions

Sources/Sentry/SentryTime.mm
philipphofmann: We can't use `NSTimeInterval` because of `SENTRY_EXTERN_C_BEGIN`. I think it's fine to keep the double. I'm not that familiar with the C and C++ interop. What would we need to do to remove the `SENTRY_EXTERN_C_BEGIN`, @armcknight?

PR #3501, opened 2023-12-15
fix: Marking manual sessions as crashed

philipphofmann:
## :scroll: Description

Fix marking manual sessions as crashed when capturing crash events.

## :bulb: Motivation and Context

Fixes GH-3498

## :green_heart: How did you test it?
Unit tests and with the simulator

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I reviewed the submitted code.
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #3502, opened 2023-12-15
chore: Script to create a hot fix

brustolin:
Added a script that helps creating a hot fix from a specific version with a specific PR commit.

usage (from repo root directory):

```
./scripts/hot-fix.sh <version from> <new version> <commit hash>
```
To release 8.17.2 with https://github.com/getsentry/sentry-cocoa/pull/3501
```
./scripts/hot-fix.sh 8.17.1 8.17.2 aea5987
```

After running the script we need to run the  `release` action with `hotfix/<new version>` branch for `Use workflow from`
_#skip-changelog_


armcknight (commented):
It sounds like this would be a new release process from just merging a PR to master, because we want to avoid shipping other new features along with a hotfix patch?

Could you add something to develop-docs/README.md that outlines the standard and hotfix processes?

Maybe it could also open a PR back into `main` after pushing to `$hotfix_branch` on line 22 using `gh pr create`, or would that be done automatically from the release job?

brustolin:
> It sounds like this would be a new release process from just merging a PR to master, because we want to avoid shipping other new features along with a hotfix patch?

From time to time, I am asked to do this, and I always forget how to. The script is designed to make it easy to create a hotfix release from one PR. This is not a process that we need to follow.

>Maybe it could also open a PR back into main after pushing to $hotfix_branch on line 22 using gh pr create, or would that be done automatically from the release job?

We don't need a new PR for hotfixes, all the code should have being reviewed already.

philipphofmann:
@brustolin, I think we should rename the script to something like `create-hotfix-cherry-pick-branch.sh` to avoid confusion, as it only creates the branch. I agree with @armcknight that we should add a note to the internal develop docs or explain the usage in the script.
```
./scripts/hot-fix.sh <version from> <new version> <commit hash>
```

philipphofmann (approved):
LGTM

Issue #3503, opened 2023-12-18
EXC_BAD_ACCESS   sentry::profiling::ThreadHandle::dispatchQueue...


zhou934381967:
### Platform

iOS

### Environment

Production

### Installed

Swift Package Manager

### Version

8.9.6

### Steps to Reproduce

Errors monitored online

### Expected Result

The operation is normal without Crash

### Actual Result

Existential Crash，Is that a problem with this? https://github.com/getsentry/sentry-cocoa/pull/3054
<img width="814" alt="截屏2023-12-18 16 09 42" src="https://github.com/getsentry/sentry-cocoa/assets/13718172/e28b081c-b31a-40cf-ac9b-1b48d6a7dfa0">

philipphofmann:
Thanks for reporting this, @zhou934381967. This is a duplicate of https://github.com/getsentry/sentry-cocoa/issues/3223. Please add more information over there.

PR #3522, opened 2023-12-30
fix: Remove dispatch queue metadata collection to fix crash

indragiek:
## :scroll: Description

This PR removes the collection of dispatch queue metadata when profiling, and uses `main` as the hard-coded thread name for the main thread (in place of the queue name `com.apple.main-thread`) for consistency with other platforms.

We have crashes in production from trying to read `ThreadHandle::dispatchQueueAddress` when profiling: https://github.com/getsentry/sentry-cocoa/issues/3503

This appears to be a race condition with queue lifetime that is not possible to fully fix. There are similar problems encountered by other projects trying to use the same API:

* https://github.com/kstenerud/KSCrash/issues/207
* https://github.com/Instabug/Instabug-iOS/issues/245

Updated the tests, checked that the hardcoded `main` thread name is set.

## :pencil: Checklist

- [X] I reviewed the submitted code.


Sources/Sentry/Profiling/SentryProfilerState.mm
```diff
@@ -71,36 +79,38 @@ - (void)mutate:(void (^)(SentryProfilerMutableState *))block
...
-        }
         NSMutableDictionary<NSString *, id> *metadata = state.threadMetadata[threadID];
         if (metadata == nil) {
             metadata = [NSMutableDictionary<NSString *, id> dictionary];
             state.threadMetadata[threadID] = metadata;
         }
-        if (!backtrace.threadMetadata.name.empty() && metadata[@"name"] == nil) {
-            metadata[@"name"] =
-                [NSString stringWithUTF8String:backtrace.threadMetadata.name.c_str()];
+        if (metadata[@"name"] == nil) {
```
philipphofmann, line 102: `m`: I think we shoud first check if `backtrace.threadMetadata.threadID == self->_mainThreadID`. Otherwise, the main thread could have different names depending on the platform. Or is this the desired behavior?

philipphofmann (approved):
Good catch, thanks @indragiek 👍

Sources/Sentry/Profiling/SentryProfilerState.mm
armcknight: This looks like it's checking if it's already been set previously, as we reuse the metadata dictionary. The check you mention does happen before we actually go to set the value on line 107 below in this hunk.

armcknight:
I logged the thread sanitizer warning in a new issue, I don't think it's related.

Sources/Sentry/Profiling/SentryProfilerState.mm
indragiek: Yeah @armcknight is right, that was the intent here

Issue #3461, opened 2023-11-29
Fix end time of app start


philipphofmann:
### Description

Currently, the end of the app start is when the OS posts `UIWindowDidBecomeVisibleNotification`, which doesn't mean that some proper content is rendered for the user. When looking at an app start transaction, we can see that the initial frame render ends before the initial display span even starts. 
![CleanShot 2023-11-29 at 15 49 11](https://github.com/getsentry/sentry-cocoa/assets/2443292/2f3d8e9d-53e1-4d7a-a349-31711e76b32d)

Instead, the end of the initial frame render span and the app start should be the same as of the end of TTID. The SDK reports TTID when the app draws the next frame after `loadView` of the UIViewController is called, which correctly reflects when the user sees the first frame of proper content.

armcknight:
Is this example reporting that TTID == TTFD? It's a bit confusing because I expect TTFD to always come after TTID, even if they're equal, but that's not what the frontend is showing here.

philipphofmann:
No TTFD is shortly after TTID. You should call `SentrySDK.reportFullyDisplayed` once your UIViewController has done all the heavy lifting. This might occur before TTID because we report TTID once the next frame is drawn. If you have a UIViewController doing very little, and you manually call `SentrySDK.reportFullyDisplayed`, that can happen. I also see this in our iOS-Swift sample app. We could highlight this in the product to state that this view was rendered on time.

armcknight:
> TTFD is shortly after TTID
Definitely seems like there's a bug in the frontend since in your screenshot it shows TTFD and then TTID. That should never happen IIUC.

philipphofmann:
Ah you are talking about this ?

![CleanShot 2023-12-01 at 10 35 46@2x](https://github.com/getsentry/sentry-cocoa/assets/2443292/0c42cc31-8c84-4ed0-9db9-0847e8b6b132)

Yes, that seems like a bug.

armcknight:
I'm still a bit unclear as to what we expect to see in the screenshot. It seems like TTFD should be at the end of `ui.load.full_display`, is that right?

This might be breaking in the sense that the aggregates are going to be weird once a customer goes live with this update, but they will even out after some time. I'm not sure that's really a breaking change w.r.t. the SDK, this is a bugfix because we weren't correctly reporting the timestamps, IIUC.

[...]

PR #3530, opened 2024-01-03
fix: App start ends when first frame is drawn

philipphofmann:
When performanceV2 is enabled, the app start ends when the app draws the first frame instead of when it receives the
UIWindowDidBecomeVisibleNotification.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/3525.

This the change requested in https://github.com/getsentry/sentry-cocoa/issues/3461, but in a non-breaking way behind the feature flag `performanceV2`.

Unit tests and simulator.

## :pencil: Checklist


brustolin (approved):
LGTM

Just wondering what's the relevance of this now that it aligns with TTID, both info will be always the same.

philipphofmann:
> Just wondering what's the relevance of this now that it aligns with TTID, both info will be always the same.

Yeah, I had the same thought. I don't think this is necessarily bad, but it's redundant. Maybe we need a different concept for app starts. I will keep this in the back of my head and bring it up in our Mobile Starfish discussions.

PR #3538, opened 2024-01-04
fix: Move header reference out of `extern C`

brustolin:
## :scroll: Description

Removing header references from `extern C` because they are causing compile errors with Android Studio and flutter.

close https://github.com/getsentry/sentry-dart/issues/1781

## :pencil: Checklist


philipphofmann (commented):
Can we add something in CI that prevents that mistake in the future? What else is missing so we can merge this @brustolin?

brustolin:
> Can we add something in CI that prevents that mistake in the future? What else is missing so we can merge this @brustolin?

This is kind of a shot in the dark. Everything I read about it says it's okay (or we should) to move the includes out of extern C. CI is not complaining about it, and other tests where I forced the file to use a C++ compiler were okay. But I could not reproduce the error before the changes.

I'm waiting for the user to test this branch; I sent them instructions in the original issue.

kahest:
The user in https://github.com/getsentry/sentry-dart/issues/1781#issuecomment-1882507062 reported that this fixed their issue

philipphofmann (approved):
LGTM

PR #3542, opened 2024-01-05
test: sample app and ui test tweaks

armcknight:
- delete Samples/iOS-Swift/ClearTestState.swift as it isn't used
- remove SentryBenchmarking.mm from compiling into iOS-SwiftUITests as it's only needed in the actual test harness (iOS-Swift app), not the test runner
- remove SentryDevice.mm from compiling into iOS-Swift, as it's only needed in the test runner (iOS-SwiftUITests)
- add launch args to iOS-Swift to allow parameterizing trace and profile sample rates set via either numerical rate or sampling function (this will help with some UI tests that are coming in #3529 )
- some changes Xcode automatically made to tvOS-Swift.xcodeproj involving linking some dependencies like SentryPrivate, SentryTestUtils, etc
- remove scripts/set-device-tests-environment.patch and do the configuration directly in AppDelegate; this has caused merge conflicts a few times when i modify something in AppDelegate and causes CI to break

#skip-changelog


brustolin (approved):
LGTM

philipphofmann (approved):
Awesome thanks 💯

PR #3545, opened 2024-01-05
ref: Properly teardown frames tracker

philipphofmann:
The teardown of the `SentryFramesTracker` didn't work properly in tests. The `clearTestState` method continuously initialized a new frames tracker cause it accessed the property dependency container. This is fixed now by calling `SentryFramesTracker.stop` directly in `SentryDependencyContainer.reset` which is also called in `clearTestState`. Furthermore, the `SentryFramesTracker.stop` now resets all frames and removes all observers.

I identified this problem when looking at the logs of failing tests at the main branch here: https://github.com/getsentry/sentry-cocoa/actions/runs/7419191432/job/20188316830. The logs showed continuous init calls to SentryFramesTracker and also unwanted tracking of frames during tests, which could lead to failing tests.

#skip-changelog

### DIFF
--- a/SentryTestUtils/ClearTestState.swift
+++ b/SentryTestUtils/ClearTestState.swift
@@ -26,9 +26,6 @@ class TestCleanup: NSObject {
         setTestDefaultLogLevel()
 
         #if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
-        let framesTracker = SentryDependencyContainer.sharedInstance().framesTracker
-        framesTracker.stop()
-        framesTracker.resetFrames()
 
         setenv("ActivePrewarm", "0", 1)
         SentryAppStartTracker.load()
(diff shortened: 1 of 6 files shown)


SentryTestUtils/ClearTestState.swift
```diff
@@ -26,9 +26,6 @@ class TestCleanup: NSObject {
         setTestDefaultLogLevel()
 
         #if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
-        let framesTracker = SentryDependencyContainer.sharedInstance().framesTracker
```
philipphofmann, line 29: `SentryDependencyContainer.reset` now stops the frames tracker, which this method also calls.

brustolin (approved):
LGTM

Issue #3329, opened 2023-10-04
Add iOS 17 unit tests to matrix


armcknight:
### Description

Tried doing this a bit in #3283 but had some issues with test job length. We should wait until github updates their macOS image's Xcode version once 15.1 goes GA to see if this bug is what was causing it:
> Fixed: Executing Unit/UI tests from Xcode on the iOS Simulator takes an extended time to launch on first run of the suite. (115187363) (110330776)

philipphofmann:
> Fixed: Executing Unit/UI tests from Xcode on the iOS Simulator takes an extended time to launch on first run of the suite. (115187363) (110330776)

The fix is in [Xcode 15.0.1](https://developer.apple.com/documentation/xcode-release-notes/xcode-15_0_1-release-notes), which is available in the latest [macos-13 GH image](https://github.com/actions/runner-images/releases/tag/macos-13/20231018.14). So we can tackle this issue now.

armcknight:
https://github.com/getsentry/sentry-cocoa/pull/3433 removed iOS 17 from test runs, so reopening this issue until we get them back in.

PR #3328, opened 2023-10-04
feat: Add experimental visionOS support

brustolin:
## :scroll: Description

Added support for visionOS

## :bulb: Motivation and Context

close #3101, #3323 

## :green_heart: How did you test it?

Samples

## :pencil: Checklist

You have to check all boxes before merging:

- [ ] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #3548, opened 2024-01-08
feat: Add visionOS as device family

brustolin:
Added `visionOS` as device family type. This will be used to filter events.


armcknight (commented):
Following on from #3328. I think we'll also need to update a couple other things to use Xcode 15.2, some or all of these: 
- ci-select-xcode.sh to use Xcode 15.2 as a default: https://github.com/getsentry/sentry-cocoa/blob/94d8eb36ac3c56b44fecc9c4d13b41f9b24f34c2/scripts/ci-select-xcode.sh#L11
- the build GHA workflow to build and validate the xcframework: https://github.com/getsentry/sentry-cocoa/blob/94d8eb36ac3c56b44fecc9c4d13b41f9b24f34c2/.github/workflows/build.yml#L97
- the unit test GHA version matrix, 14.3 is the latest Xcode version we currently use (we should also update to run iOS 17 separately once we get xcode 15.2 support; see https://github.com/getsentry/sentry-cocoa/issues/3329): https://github.com/getsentry/sentry-cocoa/blob/94d8eb36ac3c56b44fecc9c4d13b41f9b24f34c2/.github/workflows/test.yml#L110-L114

brustolin:
>Following on from https://github.com/getsentry/sentry-cocoa/pull/3328. I think we'll also need to update a couple other things to use Xcode 15.2, some or all of these:

Yes, but I think we should create a different PR for this things.

brustolin:
@philipphofmann do you think we can go ahead with this?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+- Add visionOS as device family (#3548)
```
philipphofmann, line 7: `m`: This needs to move up to the unreleased section

philipphofmann (approved):
Yes, sure. LGTM.

PR #3557, opened 2024-01-13
meta: debug logging

armcknight:
These are some debug logs that helped me with #3529.

#skip-changelog


Sources/Sentry/SentryOptions.m
```diff
@@ -708,4 +709,136 @@ - (void)setEnablePreWarmedAppStartTracing:(BOOL)enablePreWarmedAppStartTracing
 
 #endif // SENTRY_UIKIT_AVAILABLE
 
+#if defined(DEGUG)
+- (NSString *)description
```
philipphofmann, line 713: `m`: What about using [`debugDescription`](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418703-debugdescription) instead?

Sources/Sentry/SentryOptions.m
```diff
@@ -708,4 +709,136 @@ - (void)setEnablePreWarmedAppStartTracing:(BOOL)enablePreWarmedAppStartTracing
...
+{
+    NSMutableArray<NSString *> *optionsDescription = [NSMutableArray<NSString *> array];
```
philipphofmann, line 715: `m`: I think we must find a way to ensure we add future properties here; otherwise, it can be misleading. One idea is to count the number of properties of SentryOptions with something like
```objc
void printPropertiesOfClass(Class cls) {
    unsigned int count;
    objc_property_t *properties = class_copyPropertyList(cls, &count);

    for (unsigned int i = 0; i < count; i++) {
        const char *propertyName = property_getName(properties[i]);
    }

    free(properties);
}
```

and then compare it to the size of `optionsDescription`. If they don't match, a test fails.

Sources/Sentry/SentryOptions.m
```diff
@@ -708,4 +709,136 @@ - (void)setEnablePreWarmedAppStartTracing:(BOOL)enablePreWarmedAppStartTracing
...
+                                                self.enableMetricKit ? @"YES" : @"NO"]];
+    }
+#    endif // SENTRY_HAS_METRIC_KIT
+    [optionsDescription addObject:[NSString stringWithFormat:@"enableTimeToFullDisplayTracing: %@",
+                                            self.enableTimeToFullDisplayTracing ? @"YES" : @"NO"]];
+    [optionsDescription addObject:[NSString stringWithFormat:@"swiftAsyncStacktraces: %@",
+                                            self.swiftAsyncStacktraces ? @"YES" : @"NO"]];
+    [optionsDescription
+        addObject:[NSString stringWithFormat:@"cacheDirectoryPath: %@", self.cacheDirectoryPath]];
+    return [optionsDescription componentsJoinedByString:@"\n"];
```
philipphofmann, line 840: `m`: I don't know where I read this, but I think adding the class name and memory address in the description is good practice. So something like
```
<SentryOptions: 0x7f98f2e01d20, {
    debug = "YES";
   // ...
}>
```

Sources/Sentry/SentryOptions.m
philipphofmann, line 713: `m`: I think adding a test for this would be great. First, it validates that the code doesn't crash, and it also validates that the description works correctly. It could cost a debugging developer plenty of time if it doesn't.

philipphofmann (commented):
That's a great idea, thanks for opening the PR.

Sources/Sentry/SentryOptions.m
armcknight: Yeah that would work 👍🏻

armcknight:
I had a more involved solution that would look at each type of property, but wound up not needing it for now, just call `valueForKey:` and then `description`. But in case it's needed later, i got started from https://stackoverflow.com/questions/754824/get-an-object-properties-list-in-objective-c

Sources/Sentry/SentryOptions.m
armcknight: I wrote a totally dynamic solution.

Sources/Sentry/SentryOptions.m
armcknight: I'm simply going to test that it returns non nil, so we know it doesn't crash. Testing for the actual value is going to be much harder, because several things could change that would result in a different string. I say let's cross that bridge if we get to it.

Sources/Sentry/SentrySDK.m
```diff
@@ -160,7 +160,7 @@ + (void)startWithOptions:(SentryOptions *)options
     // We accept the tradeoff that the SDK might not be fully initialized directly after
     // initializing it on a background thread because scheduling the init synchronously on the main
     // thread could lead to deadlocks.
-    SENTRY_LOG_DEBUG(@"Starting SDK...");
+    SENTRY_LOG_DEBUG(@"Starting SDK with options: %@", options.debugDescription);
```
philipphofmann, line 163: `m`: I thought it would be enough to just pass the `options` here, and the NSString.format would use the `debugDescription`, but it doesn't. The `debugDescription` works only when the debugger is attached. So I think it's better to revert to your old way of using `#if defined(DEGUG)` in the description, because now `debugDescription` will also run in release, and I don't think that's needed. Or we simply change `debugDesription` to `description` that's also fine. Up to you.

Tests/SentryTests/SentryOptionsTest.m
```diff
@@ -1233,6 +1233,14 @@ - (void)testInitialSwiftAsyncStacktracesYes
     XCTAssertTrue(options.swiftAsyncStacktraces);
 }
 
+- (void)testOptionsDebugDescriptionDoesntCrash
+{
+    SentryOptions *options = [self getValidOptions:@{}];
+    XCTAssertNotNil(options.debugDescription);
```
philipphofmann, line 1239: `l`: Writing a proper test for this is hard because we must change it every time we add a property. Maybe we could check if the description contains one or two properties such as `releaseName: io.sentry.sample.iOS-Swift@8.20.0+1` to know it's not completely broken.

philipphofmann (commented):
Almost LGTM.

Sources/Sentry/SentrySDK.m
armcknight: I'm going to leave it as `debugDescription` and will surround it with the ifdef

Tests/SentryTests/SentryOptionsTest.m
armcknight: To me it's not worthwhile to essentially write a test for `property_getName` and `[[self valueForKey:propertyName] description]`. This is back to being just debug code, honestly we should remove this test altogether, if it crashes it will be obvious what's going on. Let's revisit this if it actually causes problems, my guess is we won't have to.

philipphofmann: It won't cause problems now, but if somebody makes a change by mistake, it's still good to have one test invoking the code to tell you it's broken.

philipphofmann (approved):
LGTM. The main branch was broken. Merging main into this branch should fix the broken CI.

Tests/SentryTests/SentryOptionsTest.m
armcknight: What specifically are you worried about getting broken?

armcknight: fair enough, i sent in some nondefault values for tracesSampler and profilesSamplingRate and checked for expected values in them.

philipphofmann (approved):
LGTM.

armcknight:
Ahhaha it fails on tvos because i decided to use a profiling-only option (╯°□°)╯︵ ┻━┻

PR #3558, opened 2024-01-13
test: automatic app launch profiling

armcknight:
The integration tests for #3529 

    - previously, profiles would post to a notification containing the payload so UI tests could extract them; now, the profiling will write profiles to disk and there's a button in iOS-Swift/ProfilingViewController that retrieves them on demand from UI tests
    - the new integration test launches the app multiple times with different configurations of SentryOptions and on each launch, evaluates assertions relating to two things:
        1. the config file we write out in SentrySDK.startWithOptions for subsequent launches to read
        2. any profile that ran during the current launch; there's a test-only category in SentryProfiler.mm called `SlowLoad` that has a long-running task in its `+[load]` override; we check for this as well as ensuring there's no frame for the `main` call on the main thread, ensuring we've captured stack traces in the pre-main launch sequence

# TODO

- [ ] assert that the stack containing SlowLoad was on the main thread (end of ProfilingUITests.assertLaunchProfile)

#skip-changelog


Samples/iOS-Swift/iOS-Swift/Profiling/ProfilingViewController.swift
```diff
@@ -1,20 +1,24 @@
 import UIKit
 
-@available(iOS 13.0, *)
+@available(iOS 16.0, *)
```
philipphofmann, line 3: `m`: Why did you have to bump this to iOS 16? I'm curious.

Samples/iOS-Swift/iOS-SwiftUITests/BaseUITest.swift
```diff
@@ -1,26 +1,44 @@
 import XCTest
 
 class BaseUITest: XCTestCase {
-    internal let app: XCUIApplication = XCUIApplication()
+    internal lazy var app: XCUIApplication = newAppSession()
+    
+    //swiftlint:disable implicit_getter
+    var automaticallyManageAppSession: Bool { get { true } }
```
philipphofmann, line 7: `l`: Maybe a bit clearer
```suggestion
    var automaticallyLaunchAndTerminateApp: Bool { get { true } }
```

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -1,4 +1,4 @@
-#import "SentryLaunchProfiling.h"
+#import "SentryLaunchProfiling+Tests.h"
```
philipphofmann, line 1: `m`: I'm a bit confused. Why does production code include a test header?

Sources/Sentry/SentryProfiler.mm
```diff
@@ -50,6 +50,30 @@
 #        import <UIKit/UIKit.h>
 #    endif // SENTRY_HAS_UIKIT
 
+#    if defined(TEST) || defined(TESTCI)
```
philipphofmann, line 53: `m`: As this code is only required for tests, wouldn't it make sense to put that into a file in the `SentryTests/` folder?

Sources/Sentry/SentryTracer.m
```diff
@@ -48,6 +48,11 @@
 #    import <SentryScreenFrames.h>
 #endif // SENTRY_HAS_UIKIT
 
+#if defined(TEST) || defined(TESTCI)
+#    import "SentryFileManager+Test.h"
```
philipphofmann, line 52: `m`: Having a test header in production code feels weird. Either we have to rename the header or change our code so it doesn't rely on a test header.

Samples/iOS-Swift/iOS-SwiftUITests/ProfilingUITests.swift
```diff
@@ -1,39 +1,246 @@
...
+class ProfilingUITests: BaseUITest {    
+    override var automaticallyManageAppSession: Bool { false }
+    
+    // this will run before the non-async BaseUITest.setUp, so we can bail out before running any of the logic in there
+    override func setUp() async throws {
+        try await super.setUp()
+        try checkOSVersionForProfilingTest()
+    }
+    
+    func testProfiledAppLaunches() throws {
```
philipphofmann, line 14: That's a fancy integration test covering multiple scenarios 💯

philipphofmann (commented):
The integration test is extensive. I like that 👍

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.m
```diff
@@ -0,0 +1,157 @@
...
+#import "SentryProfilingConditionals.h"
+#import "SentrySDK+Tests.h"
+#import <XCTest/XCTest.h>
+
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+
+@interface SentryAppLaunchProfilingTests : XCTestCase
+@end
+
+@implementation SentryAppLaunchProfilingTests
```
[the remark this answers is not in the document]
armcknight: I started writing these in Swift, but it was so verbose and repetitive that it became hard to understand the differences between the tests or what they were actually testing :/

Samples/iOS-Swift/iOS-Swift/Profiling/ProfilingViewController.swift
armcknight: ah, i used a FileManager API that was too new. I was able to convert to something more backwards compatible.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
armcknight: Mm, I thought I'd have to do this because I needed to expose some C declarations to tests, but I can just redeclare them in SentryLaunchProfiling.m as well and there doesn't seem to be an issue 👍🏻

Sources/Sentry/SentryProfiler.mm
armcknight: This will be tricky. It's only used in the UI tests, not the unit tests, so exposing internal implementation from the SDK isn't as straightforward. If we bring in SentryProfiler, then we have to bring in every other import as well.

Let me see if just putting an ObjC category in the app can still be guaranteed to run its `+[load]` method after our ObjC classes in the SDK.

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.m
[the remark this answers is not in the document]
philipphofmann: Okay,, that's fine then. Please explain this by adding a comment so we know, in the future why we chose ObjC here.

Sources/Sentry/SentryProfiler.mm
philipphofmann: Again, this is acceptable. Please add a comment to explain this. Isolating this in a category is a great idea.

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.m
```diff
@@ -0,0 +1,178 @@
...
+
+#    define SENTRY_OPTION(name, value)                                                             \
+        NSStringFromSelector(@selector(name))                                                      \
+            : value
+
+- (void)testDefaultOptionsDoNotEnableLaunchProfiling
+{
+    XCTAssertFalse(shouldProfileNextLaunch([self defaultOptionsWithOverrides:nil]).shouldProfile,
```
philipphofmann, line 21: `m`: Nimble works for ObjC as well. I didn't try it yet, but it works great. Up to you.

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.m
```diff
@@ -0,0 +1,178 @@
+#import "SentryLaunchProfiling+Tests.h"
+#import "SentryOptions+HybridSDKs.h"
+#import "SentryOptions+Private.h"
+#import "SentryProfilingConditionals.h"
+#import "SentrySDK+Tests.h"
+#import <XCTest/XCTest.h>
+
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+
+@interface SentryAppLaunchProfilingTests : XCTestCase
```
philipphofmann, line 10: I think that covers all edge cases. Excellent 👍

philipphofmann (approved):
Should we maybe merge this into https://github.com/getsentry/sentry-cocoa/pull/3529 to have everything in one place? I think this should be good to go.

Sources/Sentry/SentryProfiler.mm
armcknight: Will be done in a later PR

PR #3585, opened 2024-01-29
fix: TARGET_OS_VISION not defined for macOS

brustolin:
Added

```objc
#ifndef TARGET_OS_VISION
#define TARGET_OS_VISION 0
#endif
```

to fix compilation on CI for macOS

_#skip-changelog_


philipphofmann (approved):
LGTM

Issue #3576, opened 2024-01-27
Remove force unwraps from test code


armcknight:
### Description

We have numerous force-unwraps in our Swift test code, which make it harder to iterate locally if you make a breaking change. This causes the test process to crash if an assumption is broken, disallowing the complete suite of tests to run, preventing you from seeing the total list of tests that would fail. This could mean many cycles of fixing one breakage, run the suite again, encounter the next crash. Of course, it's more painful if it happens in CI (and is harder to see what's going on there, too).

Convert such force unwraps to either `guard...else { XCTFail("message"); return }` or `try XCTUnwrap(...)`

philipphofmann:
@armcknight, to enforce this rule, we could enable the swiftlint rule for the tests again https://github.com/getsentry/sentry-cocoa/blob/faaf832acdece0118cf8e9e20d3e4cf3c568691f/Tests/.swiftlint.yml#L6. We could allow specific usages of it in the code by disabling the lining rule only locally with code comments.

armcknight:
@philipphofmann excellent idea, I'm all for it 👍🏻

cmanallen:
Closing because this was marked as "done" in the Mobile and Cross Platform SDKs project.

PR #3589, opened 2024-01-30
test: use XCTUnwrap instead of force unwraps or unchecked array access

armcknight:
I changed these while investigating a broken assumption that were crashing the test run instead of just failing these cases. Helps with #3576.

#skip-changelog


philipphofmann (commented):
Thanks 💪. CI needs some fixes.

Issue #3577, opened 2024-01-28
Could not get NSFileTypeDirectory, Couldn't load files in folder [envelopes]


lockieluke:
### Platform

macOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

8.19.0

### Steps to Reproduce

1. Start a sentry session using the following configuration
```swift
SentrySDK.start { options in
    options.dsn = Bundle.main.object(forInfoDictionaryKey: "SENTRY_DSN_URL") as? String
    #if DEBUG
    options.debug = true
    options.diagnosticLevel = .warning
    #endif
	// Diagnostic.isDebug is simply a #if DEBUG macro
    options.environment = Diagnostic.isDebug ? "debug" : "production"
    
    // Set tracesSampleRate to 1.0 to capture 100% of transactions for performance monitoring.
    // We recommend adjusting this value in production.
    options.tracesSampleRate = 1.0
    
    options.attachStacktrace = true
    options.enableMetricKit = true
    options.enableTimeToFullDisplayTracing = true
    options.swiftAsyncStacktraces = true
}
```
2. Run your app
3. Look at the logs

### Expected Result

These errors should not exist

### Actual Result

```
[Sentry] [warning] [SentryFileManager:175] Could not get NSFileTypeDirectory from /Users/[username]/Library/Caches/io.sentry/.DS_Store
[Sentry] [error] [SentryFileManager:231] Couldn't load files in folder /Users/[username]/Library/Caches/io.sentry/[id]/envelopes: Error Domain=NSCocoaErrorDomain Code=260 "The folder “envelopes” doesn’t exist." UserInfo={NSFilePath=/Users/[username]/Library/Caches/io.sentry/[id]/envelopes, NSUserStringVariant=(
    Folder
), NSUnderlyingError=0x600000f1d0b0 {Error Domain=NSOSStatusErrorDomain Code=-43 "fnfErr: File not found"}}
```

### Are you willing to submit a PR?

I have no idea what's causing this, I'm sorry I don't think I'll be able to fix this

brustolin:
Hello @lockieluke, thanks for reporting this. Does your project have any function that cleans the cache folder during runtime?

philipphofmann:
[...]

PR #3594, opened 2024-01-30
fix: FileManager log info when path doesn't exist

philipphofmann:
## :scroll: Description

Logging an error when getting all files for a folder that doesn't exist confuses users; see GH-3577. Instead, log an info message that the SentryFileManager tried getting a list of files for a folder that doesn't exist. We chose info because it's still an edge case that shouldn't occur often, but it doesn't break any functionality. If a path was deleted manually by the user that the SDK requires, the SDK will create the path on next launch. We don't want to add checks if all the necessary paths exist every time the SDK stores an envelope cause this adds some overhead. If users manually remove the required paths, we accept that they have to wait until the next time the SDK launches.

Fixes GH-3577

Unit tests.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Clarify FramesTracker log message (#3570)
 - Fix rare battery breadcrumbs crash (#3582)
 - Fix synchronization issue in FramesTracker (#3571)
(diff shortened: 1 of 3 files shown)


brustolin (approved):
LGTM

PR #3604, opened 2024-02-02
test(app launch profiling): wipe test data from disk for integration test

armcknight:
Following on from #3558 and followed by https://github.com/getsentry/sentry-cocoa/pull/3605

I had originally had the wipe-data option but thought I could remove it. Turns out it does help, if you run this test multiple times locally with failures, it helps to have a clean state each time. Also made it so the sample app removes the profile data persisted to a file, which only happens `#if defined(TEST)`

#skip-changelog


brustolin (approved):
LGTM

Samples/iOS-Swift/iOS-Swift/AppStartup.m
```diff
@@ -0,0 +1,26 @@
+#import "AppStartup.h"
+
+@implementation AppStartup
+
+// we must do this in objective c, because it's not permitted to be overridden in Swift
```
philipphofmann, line 5: Thanks for the comment 🥇

philipphofmann (approved):
LGTM

philipphofmann (approved):
Again LGTM.

PR #3605, opened 2024-02-02
test(app launch profiling): integration test fixes and additions

armcknight:
Follows from #3604 and followed by https://github.com/getsentry/sentry-cocoa/pull/3606

Somehow I had the wrong values for some of the integration tests. It had `shouldProfileThisLaunch: false` for cases where it should have been `true`

This also resolves the TODO to check that the stack it inspects actually did run on the main thread.

TODO: there's still an issue retrieving the profile data from the file for the case marked `true`

#skip-changelog


brustolin:
It looks good, just need to fix macOS tests

philipphofmann (approved):
LGTM

PR #3606, opened 2024-02-02
feat(app launch profiling): save sample rates config to backup

armcknight:
Follows from #3605 and followed by https://github.com/getsentry/sentry-cocoa/pull/3621

I realized that a launch has a profiler run, and then due to any changes in config options, it would lead to not configuring for subsequent launch profiles, then the config file would be removed _before_ the associated transaction would start, thus never starting the transaction, and then since there'd be no tx finish to stop the profiler, it would run until timing out, and then never have anywhere to put its data and hang around in memory forever.

Instead, if a call to SentrySDK.startWithOptions would not configure an app launch profile, instead of simply removing any present config file, check first if we're currently profiling the launch, and if so, save the config to a backup location. 

#skip-changelog


brustolin (commented):
It seems good.
But im missing tests for all this create/delete/move file.

Sources/Sentry/SentryFileManager.m
```diff
@@ -787,7 +811,39 @@ - (void)createPathsWithOptions:(SentryOptions *)options
...
+        return;
+    }
+
+    NSError *error;
+    SENTRY_CASSERT([fm removeItemAtPath:path error:&error],
+        @"Failed to remove launch profile config file: %@", error);
+}
+
+void
+saveAppLaunchProfilingConfigFile(void)
```
philipphofmann, line 832: `m`: The naming is a bit confusing to me. This method doesn't save anything, but instead, it moves the current config to the backup location. The name should reflect that. Something like `moveAppLaunchProfilingConfigFileToBackUp`.

Sources/Sentry/SentryFileManager.m
```diff
@@ -787,7 +811,39 @@ - (void)createPathsWithOptions:(SentryOptions *)options
...
+{
+    NSFileManager *fm = [NSFileManager defaultManager];
+    NSString *path = launchProfileConfigFileURL().path;
+    if (!SENTRY_CASSERT_RETURN([fm fileExistsAtPath:path],
+            @"Expect to have a current launch profile config to use for subsequent transaction.")) {
+        return;
+    }
+
+    NSError *error;
+    SENTRY_CASSERT([fm moveItemAtPath:path
```
philipphofmann, line 842: `h`: What if the backup file already exists? This method then throws an error. This could happen if an app crashes during the start process. I think we should overwrite the existing config file instead. A test for that in `SentryFileManagerTests` would be great.

Sources/Sentry/SentryFileManager.m
```diff
@@ -787,7 +811,39 @@ - (void)createPathsWithOptions:(SentryOptions *)options
 
     NSError *error;
     SENTRY_CASSERT([fm removeItemAtPath:path error:&error],
-        @"Failed to remove launch profile marker file: %@", error);
+        @"Failed to remove launch profile config file: %@", error);
+}
+
+void
+removeAppLaunchProfilingConfigBackupFile(void)
```
philipphofmann, line 818: `l`: This is almost the same as `removeAppLaunchProfilingConfigFile`. Maybe we can reuse some code here. The FileManager already has `removeFileAtPath`. Maybe we could reuse that.

Sources/Sentry/SentryFileManager.m
```diff
@@ -751,12 +755,32 @@ - (void)createPathsWithOptions:(SentryOptions *)options
...
+}
+
 NSDictionary<NSString *, NSNumber *> *_Nullable appLaunchProfileConfiguration(void)
 {
     NSError *error;
-    NSDictionary<NSString *, NSNumber *> *config = [NSDictionary<NSString *, NSNumber *>
-        dictionaryWithContentsOfURL:launchProfileConfigFileURL()
-                              error:&error];
+    NSURL *configFileURL;
+    if (isTracingAppLaunch) {
```
philipphofmann, line 776: `l`: It would be better to pass that in as a parameter to specify if you want the backup or the current config. This would give us more flexibility, and it communicates more clearly what it returns. Otherwise, you have to know that the global state `isTracingAppLaunch` changes the behavior of that method.

philipphofmann (commented):
I found a few issues, but the approach make sense to me 👍

Sources/Sentry/SentryFileManager.m
armcknight: I'll put this in this PR unless you'd like otherwise.

philipphofmann: Yes, please do this in this PR. An extra PR would be unnecessary overhead 👍

Sources/Sentry/SentryFileManager.m
armcknight: Yep, i think overwriting is the way to go as well.

Sources/Sentry/SentryFileManager.m
armcknight: renamed to `backupAppLaunchProfilingConfigFile`

Sources/Sentry/SentryFileManager.m
```diff
@@ -36,6 +36,28 @@
     return YES;
 }
 
+/**
+ * @warning This is called from a `@synchronized` context in instance methods, but doesn't require
+ * that when calling from other static functions. Make sure you pay attention to where this is used
+ * from.
+ */
+void
+_unsafe_removeFileAtPath(NSString *path)
```
philipphofmann, line 45: `l`: Maybe a bit clearer that it's only not thread safe.
```suggestion
_non_thread_safe_removeFileAtPath(NSString *path)
```

Sources/Sentry/SentryFileManager.m
```diff
@@ -778,16 +808,33 @@ - (void)createPathsWithOptions:(SentryOptions *)options
...
 void
 removeAppLaunchProfilingConfigFile(void)
+{
+    _unsafe_removeFileAtPath(launchProfileConfigFileURL().path);
+}
+
+void
+removeAppLaunchProfilingConfigBackupFile(void)
+{
+    _unsafe_removeFileAtPath(launchProfileConfigBackupFileURL().path);
```
philipphofmann, line 818: :chefkiss:

Tests/SentryTests/Helper/SentryFileManagerTests.swift
```diff
@@ -663,55 +663,202 @@ class SentryFileManagerTests: XCTestCase {
...
+    
     func testReadGarbageTimezoneOffset() throws {
         try "garbage".write(to: URL(fileURLWithPath: sut.timezoneOffsetFilePath), atomically: true, encoding: .utf8)
         XCTAssertNil(sut.readTimezoneOffset())
     }
+}
+
+#if os(iOS) || os(macOS) || targetEnvironment(macCatalyst)
+// MARK: App Launch profiling tests
+extension SentryFileManagerTests {
```
philipphofmann, line 717: I didn't know you could use extensions to group tests. I love this pattern. @brustolin, I think we could use this more often. WDYT?

Tests/SentryTests/Helper/SentryFileManagerTests.swift
```diff
@@ -663,55 +663,202 @@ class SentryFileManagerTests: XCTestCase {
...
+    }
+    
+    func testBackupAppLaunchProfilingConfigFile() throws {
+        try ensureAppLaunchProfileConfig(exists: true)
+        try ensureAppLaunchProfileConfig(exists: false, backup: true)
+        expect(NSDictionary(contentsOf: launchProfileConfigFileURL())) != nil
+        expect(NSDictionary(contentsOf: launchProfileConfigBackupFileURL())) == nil
+        backupAppLaunchProfilingConfigFile()
+        expect(NSDictionary(contentsOf: launchProfileConfigFileURL())) == nil
+        expect(NSDictionary(contentsOf: launchProfileConfigBackupFileURL())) != nil
```
philipphofmann, line 806: `l`: I think validating the contents of the backup file similar to the test above would improve the quality of this test, as we don't check the contents anywhere else if I'm not mistaken.

https://github.com/getsentry/sentry-cocoa/blob/357449478f751f2664dd01495dcfddc07a31a939/Tests/SentryTests/Helper/SentryFileManagerTests.swift#L776-L781

philipphofmann (approved):
Thanks for all the excellent tests @armcknight 💯. LGTM

Tests/SentryTests/Helper/SentryFileManagerTests.swift
armcknight: I went back and forth on it. Could probably just have made a separate XCTestCase subclass ¯\\_(ツ)_/¯ but it is nice to group away the private non-testcase functions in a `private extension`

Tests/SentryTests/Helper/SentryFileManagerTests.swift
armcknight: When I went back to read the codepaths, I realized that `appLaunchProfileConfiguration()` is only ever called from the spot where it's already reading from the backup location. We never actually read the sample rates at the time we check for whether to start the profiler, we only check for the existence of the file at that time. Here's the test that checks for the rates in the backup file: https://github.com/getsentry/sentry-cocoa/blob/357449478f751f2664dd01495dcfddc07a31a939/Tests/SentryTests/Helper/SentryFileManagerTests.swift#L730-L739

I'll make sure there's a high-level explanation of what's going on with this all. There are lots of headerdocs that explain each piece, like on the functions I've added to SentryFileManager and SentryLaunchProfiling, but nothing that really ties them together.

philipphofmann (approved):
Again LGTM 👍

PR #3603, opened 2024-02-01
fix: build library for distribution

nekrich:
## :scroll: Description

Enable library evolution for SentrySwiftUI and SentryPrivate frameworks.
`BUILD_LIBRARY_FOR_DISTRIBUTION` flag:
> Ensures that your libraries are built for distribution. For Swift, this enables support for **library evolution** and generation of a module interface file.

## :bulb: Motivation and Context

When importing prebuilt SentrySwiftUI.xcframework downloaded from the Release assets in the Xcode-15.3.beta, I got an error:
```
Compiled module was created by a different version of the compiler '5.9.2.2.56'; rebuild 'SentrySwiftUI' and try again: .../SentrySwiftUI.framework/Modules/SentrySwiftUI.swiftmodule/arm64-apple-ios-simulator.swiftmodule
```
Since Carthage builds using xcworkspace, setting BUILD_LIBRARY_FOR_DISTRIBUTION build setting will do the trick. 

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I reviewed the submitted code.
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #3611, opened 2024-02-05
Revert "fix: build library for distribution"

philipphofmann:
Sadly, the change didn't work. Running `make build-xcframework` fails with `Error: error: using bridging headers with module interfaces is unsupported` see https://github.com/getsentry/sentry-cocoa/actions/runs/7753600966/job/21145248629. I can reproduce this error also locally. 

Reverts getsentry/sentry-cocoa#3603

#skip-changelog


brustolin (approved):
The problem it is only with SentrySwiftUI, we could leave the option enabled for SentryPrivate.

Also, it's possible to fix this for SentrySwiftUI by using module instead of bridging headers. If this is something we should consider, let me know and I can set it up.

philipphofmann:
> The problem it is only with SentrySwiftUI, we could leave the option enabled for SentryPrivate.

Yes, leaving the flag on for `SentryPrivate` works.

> Also, it's possible to fix this for SentrySwiftUI by using module instead of bridging headers. If this is something we should consider, let me know and I can set it up.

Yes, that's also what I found out doing some research. It would be great if you could 😃 . 

I merged the PR now, so we have a healthy main branch again and time to come up with a proper fix.

Issue #3555, opened 2024-01-12
Continuous profiling


armcknight:
Today, a profile can last up to 30 seconds and must be transmitted attached to a transaction that ran concurrently with it. This was done in part to keep memory growth of the profiling data constrained.

This body of work will transition the SDK to being able to run the profiler continuously. Periodically, profiling data will be chunked and transmitted where it can be stitched together later in the backend using a profiling "instance" ID that is set each time the profiler is started. The profile payloads will record information about root spans that were active during the profile.

This will require refactoring the profiler itself, the interface between the profiler and the tracer, and the profile payload schema.

Internal tracking: https://www.notion.so/sentry/Continuous-Profiling-iOS-d354219c683545a6b2806de8eb8bde67?pvs=4

Planning doc: https://www.notion.so/sentry/Continuous-Profiling-9d64a809ed2d487b89e4aa0358bfac18?pvs=4

cmanallen:
Closing due to being marked as done in the SDK project board.

PR #3621, opened 2024-02-08
feat(launch profiling): use dedicated traces instead of auto perf traces

armcknight:
Follows #3606 and followed by https://github.com/getsentry/sentry-cocoa/pull/3633

Realized there are some unworkable edge cases with trying to use `ui.load` traces to attach launch profiles as generated from SentryUIViewControllerPerformanceTracker. If launch A evaluates a tracesSampleRate of 1, and stores that in the launch config file, then launch B runs the launch profile but tracesSampleRate evaluates to 0, then SentryUIViewControllerPerformanceTracker isn't installed and never tries to start the ui.load trace.

Instead, just create a dedicated transaction to attach the profile to. This greatly simplifies several places where we previously tried to thread the sampling decisions through to make sure the auto tx would not be sampled out. This also allows SwiftUI apps to get launch traces (as originally discussed in the TODO section of https://github.com/getsentry/sentry-cocoa/pull/3529 and the comment at https://github.com/getsentry/sentry-cocoa/pull/3529#discussion_r1453369323)

#skip-changelog


armcknight:
The saucelabs ui tests are currently failing because the dSYM uploads are encountering 500 responses.

philipphofmann (commented):
I think we should keep the old behavior of trying to attach the app launch profile to the app start transaction. Having both linked in Sentry is highly useful for our users, or do you think it's acceptable to have an empty transaction with the app launch profile? Am I missing something? I would only create a dedicated transaction for the edge case you described here:

> If launch A evaluates a tracesSampleRate of 1, and stores that in the launch config file, then launch B runs the launch profile but tracesSampleRate evaluates to 0, then SentryUIViewControllerPerformanceTracker isn't installed and never tries to start the ui.load trace.

Maybe it's also acceptable to just ignore this for now and not send a profile if the traces sample rate is 0. We could point that out in the docs. In the long run, we want to remove the dependency of profiles and transactions, if I'm not mistaken. I hope that continuous profiling doesn't rely on transactions anymore https://github.com/getsentry/sentry-cocoa/issues/3555. 

Side note: 
It's a bit strange that we change the logic in `SentryLaunchProfiling`, `SentryHub`, `SentrySDK`, `SentryTracer`, but the code doesn't require any changes in the tests. I would expect tests to fail when we change the behavior, or do you plan on adding them later?

Sources/Sentry/SentrySampling.m
```diff
@@ -84,7 +84,7 @@
 
 SentrySamplerDecision *
 sampleProfile(SentrySamplingContext *context, SentrySamplerDecision *tracesSamplerDecision,
-    SentryOptions *options)
+    SentryOptions *_Nullable options)
```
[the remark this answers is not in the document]
armcknight: This comes out of the discussion on how to deal with a `nil` client in SentryHub.startTransaction. We could early return from that method with a SentryNoOpSpan before calling this, as mentioned in Slack, so could change it back to nonnull. Sounds like the right call to me.

armcknight:
> Maybe it's also acceptable to just ignore this for now and not send a profile if the traces sample rate is 0

What I'd like to avoid is the situation where we're already running the profiler, incurring performance overhead during the launch, and then throwing out the profile because the new traces sample rate evaluates to 0; this also means there's no trace to stop, which would stop the profiler, which instead runs until it times out after 30 seconds of wasted work.

> the code doesn't require any changes in the tests. I would expect tests to fail when we change the behavior

There was one integration test still failing, which started working after i made this change, which sets up the scenario I describe. I also removed some other tests that deal with the "backup config file" that is no longer a thing.

Sources/Sentry/SentrySampling.m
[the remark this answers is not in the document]
philipphofmann: Yes, I would also change it back, and solve the `SentryNoOpSpan` problem in another PR.

philipphofmann:
> What I'd like to avoid is the situation where we're already running the profiler, incurring performance overhead during the launch, and then throwing out the profile because the new traces sample rate evaluates to 0; this also means there's no trace to stop, which would stop the profiler, which instead runs until it times out after 30 seconds of wasted work.

I fully understand that. What about my suggestion to attach the profile to the app start transaction per default and only start your fallback app launch profile transaction when the trace sample rate evaluates to 0 on a subsequent launch? Then, we attach the app launch profile most of the time to the app start transaction so users can see them linked in Sentry. Attaching the profile always to an extra transaction is not a good idea because of an edge case. Or am I missing something? 

I leave this decision up to you. I'm okay with always attaching the app launch profile to an extra transaction, as it will change when we have continuous profiling. This discussion shouldn't stop us from shipping this feature.

> There was one integration test still failing, which started working after i made this change, which sets up the scenario I describe. I also removed some other tests that deal with the "backup config file" that is no longer a thing.

Then, all good.

---
The open merge conflicts make this PR hard to review. I remember it being small, but we have now plenty of changes in here
CleanShot 2024-02-09 at 10 13 04@2x

Overall, the PR looked good to me, but I would like to have another proper look without the conflicts before approving it.

brustolin:
It looks good, but need to fix CI.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -135,28 +135,37 @@
...
 }
 
-BOOL
-injectLaunchSamplerDecisions(
-    SentryTransactionContext *transactionContext, SentryTracerConfiguration *configuration)
+void
+stopLaunchProfile(void)
 {
     NSDictionary<NSString *, NSNumber *> *rates = appLaunchProfileConfiguration();
     removeAppLaunchProfilingConfigBackupFile();
```
philipphofmann, line 160: `m`: I would prefer doing this on a background thread to avoid extra file IO on the main thread for everybody starting the SDK.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -135,28 +135,37 @@
...
     });
 }
 
-BOOL
-injectLaunchSamplerDecisions(
-    SentryTransactionContext *transactionContext, SentryTracerConfiguration *configuration)
+void
+stopLaunchProfile(void)
 {
     NSDictionary<NSString *, NSNumber *> *rates = appLaunchProfileConfiguration();
```
philipphofmann, line 159: `m`: We don't use the rates here. Please remove them to avoid extra file IO on the main thread.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -135,28 +135,37 @@
...
+    if (launchTracer == nil) {
+        SENTRY_LOG_DEBUG(@"No launch tracer present to stop.");
     }
-    configuration.profilesSamplerDecision =
-        [[SentrySamplerDecision alloc] initWithDecision:kSentrySampleDecisionYes
-                                          forSampleRate:profilesRate];
-    transactionContext.sampleRate = rates[kSentryLaunchProfileConfigKeyTracesSampleRate];
-    return YES;
+
+    SENTRY_LOG_DEBUG(@"Finishing launch tracer.");
```
philipphofmann, line 167: `m`: This code will also run if the `launchTracer` is `nil`. Please add a return statement above or put this into an else statement.

philipphofmann (commented):
I found a few issues, and CI complains that the code doesn't compile.

Sources/Sentry/SentrySDK.m
```diff
@@ -164,6 +164,10 @@ + (void)setStartTimestamp:(NSDate *)value
 
 + (void)startWithOptions:(SentryOptions *)options
 {
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+    stopLaunchProfile();
```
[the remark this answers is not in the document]
armcknight: Hmm, yeah that is an interesting use case. They'd have to be selectively calling startWithOptions because they'd have had to call it on a prior launch to enable the launch profile in the first place. This is a tricky one to handle because although now we have a profile timeout, it will go away with continuous profiling.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
armcknight: It's now on a background thread in https://github.com/getsentry/sentry-cocoa/pull/3634

Sources/Sentry/Profiling/SentryLaunchProfiling.m
armcknight: Wound up putting a check like that in a subsequent PR

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -135,28 +135,37 @@
 
         SENTRY_LOG_INFO(@"Starting app launch profile at %llu", appLaunchSystemTime);
 
-        // don't worry about synchronizing the write here, as there should be no other tracing
-        // activity going on this early in the process. this codepath is also behind a dispatch_once
-        isTracingAppLaunch = [SentryProfiler startWithTracer:appLaunchTraceId];
+        SentryTransactionContext *context =
```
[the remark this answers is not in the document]
armcknight: In order to use the required initializer I also have to provide a nameSource. I used `kSentryTransactionNameSourceCustom` since that's what's provided as a default in the cascaded inits in `SentryTransactionContext.m`. I'm not totally clear on what these mean, and all but `kSentryTransactionNameSourceCustom` and `kSentryTransactionNameSourceComponent` appear to be unused in the SDK, so I wasn't sure if I should even add another one or not. Couldn't find anything talking about it on develop.sentry.dev or docs.sentry.io.

I did this in https://github.com/getsentry/sentry-cocoa/pull/3636/files

Sources/Sentry/SentrySDK.m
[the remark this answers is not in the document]
armcknight: https://github.com/getsentry/sentry-cocoa/issues/3637

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -135,28 +117,33 @@
...
+        NSNumber *tracesRate = rates[kSentryLaunchProfileConfigKeyTracesSampleRate];
+        if (profilesRate != nil && tracesRate != nil) {
+            config.profilesSamplerDecision =
+                [[SentrySamplerDecision alloc] initWithDecision:kSentrySampleDecisionYes
+                                                  forSampleRate:profilesRate];
+            context.sampleRate = tracesRate;
+        }
+        launchTracer = [[SentryTracer alloc] initWithTransactionContext:context
+                                                                    hub:nil
+                                                          configuration:config];
```
philipphofmann, line 134: `m`: I wonder if it makes sense to start the tracer when the `profilesRate` is `nil` or the `tracesRate` is `nil`. Currently, we would start the tracer.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -135,28 +117,33 @@
 
         SENTRY_LOG_INFO(@"Starting app launch profile at %llu", appLaunchSystemTime);
 
-        // don't worry about synchronizing the write here, as there should be no other tracing
-        // activity going on this early in the process. this codepath is also behind a dispatch_once
-        isTracingAppLaunch = [SentryProfiler startWithTracer:appLaunchTraceId];
+        SentryTransactionContext *context =
+            [[SentryTransactionContext alloc] initWithName:@"launch" operation:@"app.lifecycle"];
+        SentryTracerConfiguration *config = [SentryTracerConfiguration defaultConfiguration];
+        NSDictionary<NSString *, NSNumber *> *rates = appLaunchProfileConfiguration();
```
philipphofmann, line 123: `m`: The `appLaunchProfileConfiguration` could return `nil`, which should only happen in edge cases when the configuration file is corrupt or something similar. Maybe it makes sense to check for `nil` here and do an early return.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -135,28 +117,33 @@
...
+        NSNumber *profilesRate = rates[kSentryLaunchProfileConfigKeyProfilesSampleRate];
+        NSNumber *tracesRate = rates[kSentryLaunchProfileConfigKeyTracesSampleRate];
+        if (profilesRate != nil && tracesRate != nil) {
+            config.profilesSamplerDecision =
+                [[SentrySamplerDecision alloc] initWithDecision:kSentrySampleDecisionYes
+                                                  forSampleRate:profilesRate];
+            context.sampleRate = tracesRate;
+        }
+        launchTracer = [[SentryTracer alloc] initWithTransactionContext:context
+                                                                    hub:nil
```
philipphofmann, line 133: I just noticed that this won't work, because on this branch we never assign a hub to the `launchTracer`, but that's fixed with https://github.com/getsentry/sentry-cocoa/pull/3634 💪

philipphofmann (commented):
I identified a few more issues. This is close to LGTM.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
armcknight: Ah! I refactored this area to help write tests for this one https://github.com/getsentry/sentry-cocoa/pull/3636#discussion_r1489050618

Check out where I early return there now: https://github.com/getsentry/sentry-cocoa/blob/2fdf6e96851e948183a7f0998aec546e68a2ee74/Sources/Sentry/Profiling/SentryLaunchProfiling.m#L137-L149

Sources/Sentry/Profiling/SentryLaunchProfiling.m
armcknight: See https://github.com/getsentry/sentry-cocoa/pull/3621#discussion_r1490604276

armcknight: You can see it right in this diff now.

develop-docs/README.md
```diff
@@ -111,3 +111,19 @@ This feature is experimental and is currently not compatible with SPM.
...
+
+App launches can be automatically profiled if configured with `SentryOptions.enableAppLaunchProfiling`. If enabled, when `SentrySDK.startWithOptions` is called, `SentryLaunchProfiling.configureLaunchProfiling` will get a sample rate for traces and profiles with their respective options, and store those rates in a file to be read on the next launch. On each launch, `SentryLaunchProfiling.startLaunchProfile` checks for the presence of that file is used to decide whether to start an app launch profiled trace, and afterwards retrieves those rates to initialize a `SentryTransactionContext` and `SentryTracerConfiguration`, and provides them to a new `SentryTracer` instance, which is what actually starts the profiler. There is no hub at this time; also in the call to `SentrySDK.startWithOptions`, any current profiled launch trace is attempted to be finished, and the hub that exists by that time is provided to the `SentryTracer` instance via `SentryLaunchProfiling.stopLaunchProfile` so that when it needs to transmit the transaction envelope, the infrastructure is in place to do so.
```
philipphofmann, line 127: Thanks for the readme entry 💯

philipphofmann (approved):
LGTM

PR #3623, opened 2024-02-08
ref: Remove SentryPrivate

### LINKED ISSUE #3622: SentryPrivate framework shared library gets removed during build
**Description**

Building for iOS, using Gradle Xcode plugin (https://github.com/openbakery/gradle-xcodePlugin) SentryPrivate.framework/SentryPrivate gets removed and post-process signing steps fail as required file is missing. 

Sentry.Framework and it's library is ok, and does not get removed. 

**When does the problem happen**

- [x] During build
- [ ] During run-time
- [ ] When capturing a hard crash

**Environment**

- OS: MacOS
- Compiler: Xcode 15.2, default compilers
- CMake version and config: Cmake 3.28.2

**Steps To Reproduce**

Build project that has SentryPrivate.framework embedded, with following command

xcodebuild -project xx.xcodeproj CODE_SIGNING_ALLOWED=NO  

**Log output**

No errors on log and build succeeds, but if you peek inside the final IPA, Sentry.Framework and it's dynamic library is there, byt SentryPrivate.framework/SentryPrivate is missing.

brustolin:
## :scroll: Description

We have SentryPrivate because SPM dont support two languages in the same project.
This two frameworks are causing some problems for projects that download Sentry code and build it themselves.

close #3622 

Ci

## :pencil: Checklist


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Improvements
 
 - Cache installationID async to avoid file IO on the main thread when starting the SDK (#3601)
+- Remove SentryPrivate from Carthage and self build (#3623)
```
brustolin, line 8:
```suggestion
- Remove SentryPrivate from Carthage and self build and Cocoapod (#3623)
```

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Improvements
 
 - Cache installationID async to avoid file IO on the main thread when starting the SDK (#3601)
+- Remove SentryPrivate from Carthage and self build and Cocoapod (#3623)
```
philipphofmann, line 8: `m`:
```suggestion
- Remove SentryPrivate from Carthage, self built, and CocoaPods (#3623)
```

philipphofmann (approved):
Thanks, @brustolin 👍. LGTM

Sources/Sentry/include/SentrySwift.h
```diff
@@ -1,6 +1,12 @@
 #ifndef SentrySwift_h
 #define SentrySwift_h
 
+#if SWIFT_PACKAGE
 @import SentryPrivate;
+#elif __has_include("Sentry-Swift.h")
+#    import "Sentry-Swift.h"
+#else
+#    import "Sentry/Sentry-Swift.h"
+#endif
```
brustolin, line 10:
```suggestion
#if __has_include("Sentry-Swift.h")
#    import "Sentry-Swift.h"
#else
#    import "Sentry/Sentry-Swift.h"
#endif
```

Tests/Perf/metrics-test.yml
```diff
@@ -11,4 +11,4 @@ startupTimeTest:
 
 binarySizeTest:
   diffMin: 200 KiB
-  diffMax: 420 KiB
+  diffMax: 600 KiB
```
brustolin, line 14: Because the test is loading a dynamic framework now, its initial size is bigger than the previous solution with SPM.

Tests/SentryTests/Integrations/MetricKit/SentryMXCallStackTreeTests.swift
```diff
@@ -1,5 +1,5 @@
 #if os(iOS) || os(macOS)
-import SentryPrivate
+@testable import Sentry
```
brustolin, line 2: This is what I need from now on in order to test internal Swift code.

scripts/update-package-sha.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
+
+NEW_CHECKSUM_STATIC=$(shasum -a 256 Carthage/Sentry.xcframework.zip | awk '{print $1}')
+NEW_CHECKSUM_DYNAMIC=$(shasum -a 256 Carthage/Sentry-Dynamic.xcframework.zip | awk '{print $1}')
+
+os=$(uname)
+
+if [ "$os" == "Linux" ]; then
```
brustolin, line 9: Craft pre-release command runs on an ubuntu machine,
`sed` works a little bit different there

.github/last-release-runid
brustolin, line 1: We will use this file to save the id of the release workflow that generated the `xcframeworks` so we can use the same Id in the build workflow that runs after and retrieve the artefacts.

.github/workflows/build.yml
```diff
@@ -97,7 +97,28 @@ jobs:
...
+      - uses: actions/download-artifact@v4
+        if: startsWith(github.ref, 'refs/heads/release/')
+        with:
+          name: xcframeworks.zip
+          path: Carthage/
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ env.FRAMEWORKRUNID }}
+      
+      - name: Check XCFramework
+        if: startsWith(github.ref, 'refs/heads/release/')
```
brustolin, line 111: If this is a release, we dont compile it, we use the one downloaded from the artifact in the step before.

Package.swift
```diff
@@ -6,57 +6,34 @@ let package = Package(
...
+        .binaryTarget(
+                    name: "Sentry",
+                    url: "https://github.com/getsentry/sentry-cocoa/releases/download/8.22.0-alpha.0/Sentry.xcframework.zip",
+                    checksum: "86156301aee5c8774a8cd5c240286f914f6e7721aaac5a7c9d049ea613a4b730" //Sentry-Static
+                ),
+        .binaryTarget(
+                    name: "Sentry-Dynamic",
+                    url: "https://github.com/getsentry/sentry-cocoa/releases/download/8.22.0-alpha.0/Sentry-Dynamic.xcframework.zip",
+                    checksum: "86156301aee5c8774a8cd5c240286f914f6e7721aaac5a7c9d049ea613a4b730" //Sentry-Dynamic
+                ),
```
brustolin, line 22: IMO we should bump a major and kill one of this two (remove static and change the name of the Dynamic to Sentry only)

scripts/update-package-sha.sh
philipphofmann: `m`: Please add this essential info as a comment here.

.github/workflows/release.yml
```diff
@@ -13,16 +13,45 @@ on:
...
+  prepare_framework:
+    runs-on: macos-13-xlarge
+    name: 'Create XCFramework'
+    steps:
+      - uses: actions/checkout@v4
+    
+      - name: 'Generate XCFramework'
+        run: |
+          ./scripts/ci-select-xcode.sh 15.2
+          make build-xcframework
```
philipphofmann, line 25: `m`: I think it would be good to explain why we need to build the XCFramework here with a comment, and can't rely on building it in `build.yml.

Sources/Configuration/SentrySwiftUI.xcconfig
```diff
@@ -5,3 +5,4 @@ MACOSX_DEPLOYMENT_TARGET = 10.15
 IPHONEOS_DEPLOYMENT_TARGET = 13.0
 WATCHOS_DEPLOYMENT_TARGET = 6.0
 TVOS_DEPLOYMENT_TARGET = 13.0
+MACH_O_TYPE = mh_dylib
```
philipphofmann, line 8: `m`: Was SwiftUI a dynamic library before? Does this change have any implications for SwiftUI?

scripts/update-package-sha.sh
```diff
@@ -0,0 +1,17 @@
...
+    sed -i "s/checksum: \".*\" \/\/Sentry-Static/checksum: \"$NEW_CHECKSUM_STATIC\" \/\/Sentry-Static/" Package.swift
+    sed -i "s/checksum: \".*\" \/\/Sentry-Dynamic/checksum: \"$NEW_CHECKSUM_DYNAMIC\" \/\/Sentry-Dynamic/" Package.swift
+else
+    sed -i "" "s/checksum: \".*\" \/\/Sentry-Static/checksum: \"$NEW_CHECKSUM_STATIC\" \/\/Sentry-Static/" Package.swift
+    sed -i "" "s/checksum: \".*\" \/\/Sentry-Dynamic/checksum: \"$NEW_CHECKSUM_DYNAMIC\" \/\/Sentry-Dynamic/" Package.swift
+fi
+
+echo $GITHUB_RUN_ID > .github/last-release-runid
```
philipphofmann, line 17: `m`: Failing the script with an error message if `$GITHUB_RUN_ID` isn't set would be great because if you don't set it, it won't work, and if we ever change that by mistake, it will be a bit complicated to debug this problem.

scripts/build-xcframework.sh
```diff
@@ -0,0 +1,42 @@
...
+        else
+            echo "${sdk} SDK not found"
+        fi
+    done
+    
+    createxcframework+="-output Carthage/${scheme}${sufix}.xcframework"
+    $createxcframework
+}
+
+# Create Sentry.xcframework
```
philipphofmann, line 36: `l`: IMO, you could ditch that comment. Your code is self explaining 😃.

.github/workflows/build.yml
philipphofmann: `m`: Please add this vital info as a comment.

philipphofmann (commented):
This is the most exiting PR in 2024 so far 🥇😃.

I think we should add an entry to the internal decision log to explain why we chose this approach and that the biggest benefit is that we can mix Swift with ObjC code. Furthermore, I think we should somewhere in the code, maybe in `build.yml`, explain how this process exactly works, because it's not obvious.

.github/last-release-runid
```diff
@@ -0,0 +1 @@
+12
```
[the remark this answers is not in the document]
brustolin: I thought about it. But reading and writing in this file without a comment is so much easier and less prone to bugs

Package.swift
brustolin: Yes. We are keeping two. My suggestion is for the next major

Sources/Configuration/SentrySwiftUI.xcconfig
brustolin: It was not static before, I just move the configuration from the xcodeproj to a config file.

.github/workflows/build.yml
```diff
@@ -97,7 +97,28 @@ jobs:
...
-      - run: make build-xcframework
+      - run: echo "FRAMEWORKRUNID=$(./scripts/xcframework-generated-run.sh)" >> $GITHUB_ENV
+      
+      - uses: actions/download-artifact@v4
+        if: startsWith(github.ref, 'refs/heads/release/')
+        with:
+          name: xcframeworks.zip
+          path: Carthage/
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ env.FRAMEWORKRUNID }}
```
brustolin, line 108:
```suggestion
          run-id: ${{ env.FRAMEWORK_RUN_ID }}
```

Samples/SPM-Dynamic/Package.swift
```diff
@@ -7,16 +7,16 @@ let package = Package(
     name: "SPM-Dynamic",
     platforms: [.macOS(.v10_13)],
     products: [
-        .library(name: "SPM-Dynamic", type: .dynamic, targets: ["SPM-Dynamic"])
+        .executable(name: "SPM-Dynamic", targets: ["SPM-Dynamic"])
     ],
     dependencies: [
         // branch is replaced in CI to the current sha
-        .package(name: "Sentry", url: "https://github.com/getsentry/sentry-cocoa", .branch("main") )
+        .package(name: "Sentry", path: "../../../sentry-cocoa")
```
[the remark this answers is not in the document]
brustolin: No. We dont change this anymore, we change the Package.swift that this is targeting to.
During release we wont have the the final URL yet to be able target to that artefact, but we target to the artefact that was saved to be used as the final asset.

scripts/build-xcframework.sh
```diff
@@ -0,0 +1,42 @@
...
+generate_xcframework "Sentry" "-Dynamic"
+
+generate_xcframework "Sentry" "" staticlib
+
+# Create SentrySwiftUI.xcframework
+generate_xcframework "SentrySwiftUI"
```
brustolin, line 42:
```suggestion
generate_xcframework "Sentry" "-Dynamic"

generate_xcframework "Sentry" "" staticlib

generate_xcframework "SentrySwiftUI"
```

scripts/update-package-sha.sh
brustolin, line 9:
```suggestion
# Craft pre-release command runs on an ubuntu machine
# and `sed` needs an extra argument for macOS 
if [ "$os" == "Linux" ]; then
```

scripts/update-package-sha.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
+
```
brustolin, line 3:
```suggestion

if [ -z "$GITHUB_RUN_ID" ]; then
  echo "Error: GITHUB_RUN_ID is not set. Exiting script."
  exit 1
fi

```

.github/workflows/release.yml
```diff
@@ -13,16 +13,45 @@ on:
...
           token: ${{ secrets.GH_RELEASE_PAT }}
           fetch-depth: 0
-
+          
+      - uses: actions/download-artifact@v4
+        with:
+          name: xcframeworks.zip
+          path: Carthage/
+          
+      #update-package-sha.sh uses this env variable to update Package.swift
```
brustolin, line 52:
```suggestion
      #update-package-sha.sh uses this env variable to update Package.swift.
      #During release Craft calls bump.sh that uses update-package-sha.sh.
```

.github/workflows/release.yml
brustolin:
```suggestion
      - name: 'Generate XCFramework'
        run: |
          ./scripts/ci-select-xcode.sh 15.2
          #We need to build the framework during release to get it's SHA value
          #the framework will be saved as an artefact and we will use the same
          #binary for the entire release process to avoid the SHA to change
          make build-xcframework 
```

.github/workflows/build.yml
brustolin, line 111:
```suggestion
       #We dont compile the framework during releases to not change the artefact SHA value
       #instead we use the one archive as an artefact
        if: startsWith(github.ref, 'refs/heads/release/') 
```

.github/last-release-runid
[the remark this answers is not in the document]
philipphofmann: When you add other comments we don't need it here then 👍

Samples/SPM-Dynamic/Package.swift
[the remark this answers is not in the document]
philipphofmann: But then sample tests kind of get obsolete. The purpose of this sample is to include sentry-cocoa via SPM, similar to what our users would. Can't we use `binaryTarget` with the URL pointing to the local built XCFramework? That would be excellent.

.github/workflows/release.yml
```diff
@@ -49,7 +52,8 @@ jobs:
           name: xcframeworks.zip
           path: Carthage/
           
-      #update-package-sha.sh uses this env variable to update Package.swift
+      #update-package-sha.sh uses this env variable to update Package.swift.
+      #During release Craft calls bump.sh that uses update-package-sha.sh.
```
philipphofmann, line 56: `l`:
```suggestion
      # update-package-sha.sh uses this env variable to update Package.swift.
      # During release Craft calls bump.sh that uses update-package-sha.sh.
```

philipphofmann (commented):
Another quick pass.

Samples/SPM-Dynamic/Package.swift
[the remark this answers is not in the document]
brustolin: >But then sample tests kind of get obsolete.

IMO it was always obsolete. If `swift build` works, so will targeting the Package.swift, at least I never saw this tests failing if `swift build` is working.

> Can't we use binaryTarget with the URL pointing to the local built XCFramework? That would be excellent.

This is what the test is doing. The Package.swift file is now fixed to point to the final Package.swift (git URL or path does not matter). Instead of targeting an external ZIP URL, our Package.swift is now targeting a local ZIP file. In the end, I believe this replicates the same experience our users will have

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- Remove SentryPrivate from Carthage, self built, and CocoaPods (#3623)
```
brustolin, line 7:
```suggestion
- Remove SentryPrivate library (#3623)
```

philipphofmann (approved):
LGTM, but please wait with merging until we release 8.21.0.

Issue #3590, opened 2024-01-30
Remove usage of NSFileSystemFreeSize and NSFileSystemSize


tamwei:
### Problem Statement

According to the [Apple Docs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api?language=objc) we can't use these APIs for our purpose of sending system file size information along with events to Sentry, and therefore have to remove them.

https://github.com/getsentry/sentry-cocoa/blob/720779678ec34ec3e7773ca28019a8d7a0ef2c27/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m#L497-L513

brustolin:
The following reason is the closest we could use:


>7D9E.1
Declare this reason to include disk space information in an optional bug report that the person using the device chooses to submit. The disk space information must be prominently displayed to the person as part of the report.
Information accessed for this reason, or any derived information, may be sent off-device only after the user affirmatively chooses to submit the specific bug report including disk space information, and only for the purpose of investigating or responding to the bug report.

But the bug report is not optional, and not displayed to the user. 
So, I believe we will need to remove the storage size information which is not good.

philipphofmann:
> I believe we will need to remove the storage size information which is not good.

I think 7D9E.1 doesn't apply to us. I agree that we have to remove the storage size information. It's not good, but it's better than some apps not getting accepted by the App Store anymore because of us.

tamwei:
Well, even derived information is not available. Removing these APIs is an appropriate solution at the moment.

PR #3627, opened 2024-02-12
fix: Remove non-allowed file APIs

philipphofmann:
## :scroll: Description

Remove usage of NSFileSystemFreeSize and NSFileSystemSize because Apple forbids our usage; see
https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api?language=objc.

Most of the diffs are generated crash reports.

Fixes GH-3590

Unit tests and a manual test with a simulator.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryCrashReportConverter.m
+++ b/Sources/Sentry/SentryCrashReportConverter.m
@@ -113,7 +113,15 @@ SentryCrashReportConverter ()
 
         event.dist = self.userContext[@"dist"];
         event.environment = self.userContext[@"environment"];
-        event.context = self.userContext[@"context"];
+
+        NSMutableDictionary<NSString *, id> *userContext =
+            [[NSMutableDictionary alloc] initWithDictionary:self.userContext[@"context"]];
+        NSMutableDictionary<NSString *, id> *deviceContext =
+            [[NSMutableDictionary alloc] initWithDictionary:userContext[@"device"]];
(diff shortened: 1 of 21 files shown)


Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -147,6 +148,36 @@ - (SentryEvent *_Nullable)convertReportToEvent
...
+
+    NSMutableDictionary<NSString *, id> *deviceContext =
+        [[NSMutableDictionary alloc] initWithDictionary:context[@"device"]];
+    [deviceContext removeObjectForKey:@"storage_size"];
+    [deviceContext removeObjectForKey:@"free_storage"];
+
+    context[@"device"] = deviceContext;
+
+    return context;
+}
```
brustolin, line 179: l: I would not include this. The report was created by a version of the app that was not restricted by this new Apple requirements. The new binary of the app will not make use of the free space API, which could be a reason for the app to be rejected in the app store. Additionally, this part of the code will be useless after a while, and we will forget it here forever.

brustolin (approved):
All good. 
Just one suggestion to talk about

Sources/Sentry/SentryCrashReportConverter.m
philipphofmann: Yes, that makes sense. I wanted to be extra accurate, but I think we can remove this. Removes some complexity. Thanks 👍

PR #3632, opened 2024-02-14
fix(profiling): don't use concurrent enumeration when slicing samples

armcknight:
Something I noticed while debugging for launch profiling. It's possible with concurrency that an index beyond the first may be selected. These arrays are in chronological order, so searching just for the first instance and stopping should be performant enough. Concurrency seems like overkill and overcomplicating.

#skip-changelog


philipphofmann (approved):
LGTM, I agree that `NSEnumerationConcurrent` seems like an overkill.

If you want to make it even more performant, we could use a binary search as the array is in chronological order, but I would only do that if it's a bottleneck and in an extra PR.

PR #3633, opened 2024-02-14
fix(launch profiling): fix data wipe between integration test runs

armcknight:
Following https://github.com/getsentry/sentry-cocoa/pull/3621 and followed by https://github.com/getsentry/sentry-cocoa/pull/3634

I realized that wiping the app data didn't happen early enough to get before the app launch profile; instead of relying on it to happen automatically, be a little more careful of what is left behind in the first place in the integration test. 

#skip-changelog


Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
```diff
@@ -162,3 +163,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // swiftlint:enable force_cast
     }
 }
+
+private extension AppDelegate {
+    // previously tried putting this in an AppDelegate.load override in ObjC, but it wouldn't run until after a launch profiler would have an opportunity to run, since SentryProfiler.load would always run first due to being dynamically linked in a framework module. it is sufficient to do it  before calling SentrySDK.startWIthOptions to clear state for testProfiledAppLaunches because we don't make any assertions on a launch profile the first launch of the app in that test
```
philipphofmann, line 168: `;`:
```suggestion
    // previously tried putting this in an AppDelegate.load override in ObjC, but it wouldn't run until after a launch profiler would have an opportunity to run, since SentryProfiler.load would always run first due to being dynamically linked in a framework module. it is sufficient to do it before calling SentrySDK.startWithOptions to clear state for testProfiledAppLaunches because we don't make any assertions on a launch profile the first launch of the app in that test 
```

philipphofmann, line 168: Thanks for the comment 😃

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
```diff
@@ -162,3 +163,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // swiftlint:enable force_cast
     }
 }
+
+private extension AppDelegate {
+    // previously tried putting this in an AppDelegate.load override in ObjC, but it wouldn't run until after a launch profiler would have an opportunity to run, since SentryProfiler.load would always run first due to being dynamically linked in a framework module. it is sufficient to do it  before calling SentrySDK.startWIthOptions to clear state for testProfiledAppLaunches because we don't make any assertions on a launch profile the first launch of the app in that test 
+    func maybeWipeData() {
```
philipphofmann, line 169: `l`: We could rename this to wipeAppData and call it above when `ProcessInfo.processInfo.arguments.contains("--io.sentry.wipe-data")` is true. The name maybeWipeData is a bit confusing to me.

philipphofmann (approved):
LGTM

philipphofmann (approved):
Again LGTM.

PR #3634, opened 2024-02-14
fix(launch profiling): ensure launch profile can be uploaded

armcknight:
follows https://github.com/getsentry/sentry-cocoa/pull/3633 and followed by https://github.com/getsentry/sentry-cocoa/pull/3635

The dedicated trace was being created with a nil hub, so we need to inject it when it becomes available, before it is transmitted, or it will simply be lost. Also needed to inject the sampling decision in another parameter in the transaction context.

#skip-changelog


philipphofmann (approved):
LGTM

PR #3635, opened 2024-02-14
fix(launch profiling): debug logs and cleanup

armcknight:
Follows https://github.com/getsentry/sentry-cocoa/pull/3634 and followed by https://github.com/getsentry/sentry-cocoa/pull/3636

Remove some global variables that are no longer needed now that we use a dedicated transaction for app launch. 

Remove some old API from the backup config file strategy we no longer use. It was causing some build failures in tests.

Also some debug logs that were helpful to me.

#skip-changelog

PR #3636, opened 2024-02-14
fix(launch profiling): set a new trace origin for app launch profiles

armcknight:
follows https://github.com/getsentry/sentry-cocoa/pull/3635

It was requested to make a new origin for these transactions: https://github.com/getsentry/sentry-cocoa/pull/3621#discussion_r1486012768

I wasn't sure what to do for `nameSource`, another parameter required on the init I needed to use, so I stuck with `custom` as that's the default value used in some of the other init implementations.

#skip-changelog

### DIFF
--- a/Sources/Sentry/Profiling/SentryLaunchProfiling.m
+++ b/Sources/Sentry/Profiling/SentryLaunchProfiling.m
@@ -15,9 +15,10 @@
 #    import "SentrySamplerDecision.h"
 #    import "SentrySampling.h"
 #    import "SentrySamplingContext.h"
+#    import "SentryTraceOrigins.h"
 #    import "SentryTracer+Private.h"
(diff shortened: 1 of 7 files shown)


Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -130,7 +131,9 @@
 
         SentryTransactionContext *context =
             [[SentryTransactionContext alloc] initWithName:@"launch"
+                                                nameSource:kSentryTransactionNameSourceCustom
                                                  operation:@"app.lifecycle"
+                                                    origin:SentryTraceOriginAutoAppStartProfile
```
philipphofmann, line 136: `m`: A test validating the origin and name source would be nice.

philipphofmann (approved):
LGTM

Issue #3630, opened 2024-02-13
Fix NSMallocException when writing envelope to disk


philipphofmann:
### Description

We see a couple of `NSMallocException: Failed to grow buffer` in our [internal SDK crash detection](https://sentry.sentry.io/issues/?project=4505469596663808&query=is%3Aunresolved+NSMallocException&referrer=issue-list&sort=user&statsPeriod=14d) and also a customer reported this error.

The crash happens when serializing the envelope item data

https://github.com/getsentry/sentry-cocoa/blob/ac8fd5353ed60b08a13a4ad891b7ca374e258f01/Sources/Sentry/SentrySerialization.m#L77

The `SentrySerialization.dataWithEnvelope` puts the whole content of an envelope into NSData and then the `SentryFileManager` files the envelope to disk. https://github.com/getsentry/sentry-cocoa/blob/ac8fd5353ed60b08a13a4ad891b7ca374e258f01/Sources/Sentry/SentryFileManager.m#L284

When customers use large attachments, the SDK duplicates the memory footprint before writing the envelope to disk, which then causes the `NSMallocException`. To reduce the memory footprint of the SDK, especially with attachments, we can do the following improvements independently:

1. Use a `NSFileHandle` to write envelopes to disk in `SentryFileManager.storeEnvelope` instead of putting the contents of the envelope into NSData before writing to disk. The improvement should write envelope item directly to disk.
2. Instead of reading the data of attachments with a path into memory when initializing an envelope item, we could directly write the contents to disk: https://github.com/getsentry/sentry-cocoa/blob/ac8fd5353ed60b08a13a4ad891b7ca374e258f01/Sources/Sentry/SentryEnvelope.m#L159-L184

PR #3638, opened 2024-02-14
feat(Session Replay): ReplayEvent, ReplayRecording and Envelope handling

brustolin:
Added SentryReplayEven, SentryReplayRecording and Envelope handling for this new types.

_#skip-changelog_

Relates to:
* https://github.com/getsentry/sentry/issues/63255


Sources/Sentry/SentryClient.m
```diff
@@ -472,13 +476,53 @@ - (void)captureSession:(SentrySession *)session
...
+    if (replayEvent == nil) {
+        return;
+    } else if (![replayEvent isKindOfClass:SentryReplayEvent.class]) {
+        SENTRY_LOG_DEBUG(@"The event preprocessor didn't update the replay event in place. The "
+                         @"replay was discarded.");
+        return;
+    }
+
+    // breadcrumbs for replay will be send with ReplayRecording
+    replayEvent.breadcrumbs = nil;
```
bruno-garcia, line 501: could skip adding the crumbs here to spare some cycles?

Sources/Sentry/SentryDateUtil.m
```diff
@@ -38,6 +38,11 @@ + (NSDate *_Nullable)getMaximumDate:(NSDate *_Nullable)first andOther:(NSDate *_
     }
 }
 
++ (long)javascriptDate:(NSDate *)date
```
bruno-garcia, line 41: we're taking a Date and returning an long representing that date in milliseconds, right?
How about:

```suggestion
+ (long)toMillisecond:(NSDate *)date
```

Sources/Sentry/SentryReplayEvent.m
```diff
@@ -0,0 +1,28 @@
...
+        [trace_ids addObject:traceId.sentryIdString];
+    }
+
+    result[@"urls"] = self.urls;
+    result[@"replay_start_timestamp"] =
+        @([SentryDateUtil javascriptDate:self.replayStartTimestamp]);
+    result[@"trace_ids"] = trace_ids;
+    result[@"replay_id"] = self.replayId.sentryIdString;
+    result[@"segment_id"] = @(self.segmentId);
+    result[@"replay_type"] = @"buffer";
```
bruno-garcia, line 23: missing a `// TODO:` here? In JS there are two values;

https://github.com/getsentry/sentry-javascript/blob/bc79d32932d8d8a96f5238989a5d9ab3db915164/packages/types/src/replay.ts#L27

We send to Sentry either value to indicate whether the session started due to session sampling, or if it was buffering (keeping 30 seconds in memory 'in case' there's an error, or another trigger to start capturing the recording)

image
[Session example](https://sentry.sentry.io/replays/db7635c5e8b34421887ecc606fb6f186/?project=11276&query=&referrer=%2Freplays%2F&statsPeriod=24h&t_main=tags&yAxis=count%28%29)

[buffer](https://sentry.sentry.io/replays/9320fe38194f47629b028e11d77358dd/?project=11276&query=&referrer=%2Freplays%2F&statsPeriod=24h&t_main=tags&yAxis=count%28%29) example:
image

Sources/Sentry/SentryReplayRecording.m
```diff
@@ -0,0 +1,64 @@
...
+        @"type" : @5,
+        @"timestamp" : @(timestamp),
+        @"data" : @ {
+            @"tag" : @"video",
+            @"payload" : @ {
+                @"segmentId" : @(self.segmentId),
+                @"size" : @(self.size),
+                @"duration" : @(self.duration),
+                @"encoding" : @"h264",
+                @"container" : @"mp4",
```
bruno-garcia, line 49: how about moving this to some ReplayConstant we can refer also at the place where we do the recording. In the event we need to change anything it's easy to find out what needs to change

Sources/Sentry/SentryClient.m
brustolin: It's possible, but I don't think it is necessary

Sources/Sentry/SentryReplayRecording.m
brustolin: We can move this, but being very pragmatic, this will very unlikely ever change when it comes to video recording on iOS, and probably not used anywhere else.

Sources/Sentry/SentryReplayRecording.m
```diff
@@ -0,0 +1,64 @@
...
+    }
+    return self;
+}
+
+- (nonnull NSArray<NSDictionary<NSString *, id> *> *)serialize
+{
+
+    long timestamp = [SentryDateUtil millisecondsSince1970:self.start];
+
+    NSDictionary *metaInfo = @{
```
romtsn, line 33: note: we only have to send this with the first segment, and whenever the viewport/orientation changes, maybe leave a todo to not forget?

Sources/Sentry/include/SentryReplayRecording.h
```diff
@@ -0,0 +1,46 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class SentryId;
+
+static NSString *const SentryReplayEncoding = @"h264";
+static NSString *const SentryReplayContainer = @"mp4";
+static NSString *const SentryReplayFrameRateType = @"constant";
+
+@interface SentryReplayRecording : NSObject
```
romtsn, line 12: so this will also have other events in the future (e.g. breadcrumbs, spans, etc, see here https://github.com/getsentry/sentry-javascript/blob/8af205eba4171462c4b7f1d2d45044c97e748e37/packages/replay/src/types/replayFrame.ts#L207-L241)

You probably don't wanna have their properties all toplevel at the ReplayRecording, I'm kind of modelling the hierarchy in the following way:
- RRWebEvent
  - type
  - timestamp
  - data
  - RRWebMetaEvent
    - href
    - height
    - width
  - RRWebVideoEvent
    - size
    - ....

Sources/Sentry/SentryReplayRecording.m
brustolin: I was not aware of it. 
Is it a problem to send it with every recording tho?

Sources/Sentry/include/SentryReplayRecording.h
brustolin: Objc is so verbose that I prefer to have all this properties in the top level, especially because some of this "children" events share some properties.

I believe we will need only 3 more properties, Interactions, Breadcrumbs and Spans.

Sources/Sentry/SentryReplayRecording.m
romtsn: shouldn't be

Sources/Sentry/SentryClient.m
```diff
@@ -472,13 +476,53 @@ - (void)captureSession:(SentrySession *)session
...
+                                                withScope:[[SentryScope alloc] init]
+                                   alwaysAttachStacktrace:NO];
+
+    if (replayEvent == nil) {
+        return;
+    } else if (![replayEvent isKindOfClass:SentryReplayEvent.class]) {
+        SENTRY_LOG_DEBUG(@"The event preprocessor didn't update the replay event in place. The "
+                         @"replay was discarded.");
+        return;
+    }
```
philipphofmann, line 498: `m`: That looks a bit weird to me. Where does the event preprocessor update the replay event? I don't get this part.

Sources/Sentry/SentryClient.m
philipphofmann: The scope is currently empty. So there are no crumbs to add. If that's not on purpose, I think it would be great to do that improvement in follow-up PR.

Sources/Sentry/SentryClient.m
```diff
@@ -472,13 +476,53 @@ - (void)captureSession:(SentrySession *)session
...
                                                            singleItem:item];
     [self captureEnvelope:envelope];
 }
 
+- (void)captureReplayEvent:(SentryReplayEvent *)replayEvent
+           replayRecording:(SentryReplayRecording *)replayRecording
+                     video:(NSURL *)videoURL
+{
+    replayEvent = (SentryReplayEvent *)[self prepareEvent:replayEvent
+                                                withScope:[[SentryScope alloc] init]
```
philipphofmann, line 489: `h`: Do we use an empty scope here on purpose? If yes, can you please explain?

Sources/Sentry/SentryClient.m
```diff
@@ -472,13 +476,53 @@ - (void)captureSession:(SentrySession *)session
...
+    SentryEnvelopeItem *eventEnvelopeItem = [[SentryEnvelopeItem alloc] initWithEvent:replayEvent];
+
+    NSData *recording = [SentrySerialization dataWithJSONObject:[replayRecording serialize]];
+    SentryEnvelopeItem *recordingEnvelopeItem = [[SentryEnvelopeItem alloc]
+        initWithHeader:[SentryReplayEnvelopeItemHeader
+                           replayRecordingHeaderWithSegmentId:replayRecording.segmentId
+                                                       length:recording.length]
+                  data:recording];
+
+    NSData *video = [NSData dataWithContentsOfURL:videoURL];
```
philipphofmann, line 512: `m`: That part could be a bit heavy. Should we maybe put this into a background thread?

Furthermore, I think we need to fix this issue https://github.com/getsentry/sentry-cocoa/issues/3630 so we can use `NSFileHandle` or something similar to avoid copying the video into memory twice. We already get sometimes `NSMallocException`.

Sources/Sentry/SentryReplayRecording.m
```diff
@@ -0,0 +1,64 @@
...
+        @"type" : @4,
+        @"timestamp" : @(timestamp),
+        @"data" : @ { @"href" : @"", @"height" : @(self.height), @"width" : @(self.width) }
```
philipphofmann, line 36: `l`: Is `href` empty on purpose?

Sources/Sentry/SentryReplayRecording.m
```diff
@@ -0,0 +1,64 @@
...
+        self.duration = duration;
+        self.frameCount = frameCount;
+        self.frameRate = frameRate;
+        self.height = height;
+        self.width = width;
+    }
+    return self;
+}
+
+- (nonnull NSArray<NSDictionary<NSString *, id> *> *)serialize
```
philipphofmann, line 28: `l`: Please use `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` instead.

Sources/Sentry/include/SentryReplayRecording.h
philipphofmann: I would also prefer the hierarchy @romtsn pointed out even if it's verbose. I think you could even write those new classes in Swift as they shouldn't have any dependencies.

philipphofmann (commented):
That looks great. I'm excited to see session reply land on Cocoa soonish 😃. A few points to consider:

1. Some new classes miss `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`. Please add them. I think we should write a linter for that.
2. You could consider writing classes without dependencies in Swift.
3. I think you could set the base of this PR to `main`, so we include it sooner in the code base. Less potential for merge conflicts. I'm OK with the tradeoff that there is a bit of code that isn't used yet in there.

Sources/Sentry/SentryReplayEnvelopeItemHeader.m
```diff
@@ -0,0 +1,35 @@
...
++ (instancetype)replayVideoHeaderWithSegmentId:(NSInteger)segmentId length:(NSUInteger)length
+{
+    return [[self alloc] initWithType:SentryEnvelopeItemTypeReplayVideo
+                            segmentId:segmentId
+                               length:length];
+}
+
+- (NSDictionary *)serialize
+{
+    return @{ @"type" : self.type, @"length" : @(self.length), @"segment_id" : @(self.segmentId) };
```
romtsn, line 32: the `segment_id` is kind of an internal header within the payload, so it shouldn't be on the envelope item header afaiu, i.e.:

```json
// item header
{"type": "replay_video", "length": 5000}
// payload
{"segment_id": 0}\n{"type": 5, "timestamp": 1234, ...}
```

vaind: that's correct

Sources/Sentry/include/SentrySerialization.h
```diff
@@ -8,7 +8,7 @@ static int const SENTRY_BAGGAGE_MAX_SIZE = 8192;
 
 @interface SentrySerialization : NSObject
 
-+ (NSData *_Nullable)dataWithJSONObject:(NSDictionary *)dictionary;
++ (NSData *_Nullable)dataWithJSONObject:(id)jsonObject;
```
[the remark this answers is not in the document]
brustolin: To also serialize arrays instead of only dictionaries.

Sources/Sentry/include/HybridPublic/SentryEnvelopeItemType.h
```diff
@@ -5,3 +5,5 @@ static NSString *const SentryEnvelopeItemTypeTransaction = @"transaction";
 static NSString *const SentryEnvelopeItemTypeAttachment = @"attachment";
 static NSString *const SentryEnvelopeItemTypeClientReport = @"client_report";
 static NSString *const SentryEnvelopeItemTypeProfile = @"profile";
+static NSString *const SentryEnvelopeItemTypeReplayVideo = @"replay_video";
+static NSString *const SentryEnvelopeItemTypeReplayRecording = @"replay_recording";
```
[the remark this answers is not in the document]
brustolin: Can you point me in the right direction, Im not sure if I get this. Any example in the SDK of other envelope types?

Sources/Sentry/SentryClient.m
brustolin: Oh, missing a TODO here, we need to grab the current scope.
Thanks for the reminder.

Sources/Sentry/SentryClient.m
brustolin: We can make use of a background thread. 
The envelope format will change but I will look into avoiding load the video into memory.

Sources/Sentry/SentryReplayRecording.m
brustolin: Yes.  RRWebJson needs to have the property, but it can be empty

Sources/Sentry/SentryReplayEnvelopeItemHeader.m
brustolin: Yeah, I saw the new envelope format and realised that.

Sources/Sentry/include/SentryReplayRecording.h
brustolin: Using Swift would be nice, but programmatically speaking, it does not make sense to split this objects, they wont be used alone anywhere else, and as a con we will need to duplicate some information, like dimentions for example.

vaind (commented):
LGTM so far

Sources/Sentry/include/HybridPublic/SentryEnvelopeItemType.h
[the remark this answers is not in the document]
philipphofmann: Every envelope item type also has a data category that it maps to. Just check the code of the envelope item types and the code of the data categories. Ideally, an envelope item type should also map to a data category, which we use for rate limiting and client reports. Otherwise, rate limiting and client reports won't work for your new envelope items.

Sources/Sentry/include/SentryReplayRecording.h
philipphofmann: I'm a bit confused about why they chose inheritance for TS. The duplicated information should be in the base class. The rest of the children's classes inherit from the base class, as pointed out by @romtsn.

Sources/Sentry/SentryClient.m
brustolin: fixed

Sources/Sentry/SentryClient.m
brustolin: `testCaptureReplayEvent_WrongEventFromEventProcessor` test inside SentryClientTests explain this.

Sources/Sentry/SentryMsgPackSerializer.m
```diff
@@ -0,0 +1,87 @@
...
+
+            uint8_t buffer[1024];
+            NSInteger bytesRead;
+
+            while ([inputStream hasBytesAvailable]) {
+                bytesRead = [inputStream read:buffer maxLength:sizeof(buffer)];
+                if (bytesRead > 0) {
+                    [outputStream write:buffer maxLength:bytesRead];
+                } else if (bytesRead < 0) {
+                    break;
```
philipphofmann, line 44: `m`: Shouldn't we do some error handling here? Simply breaking leaves the written file in a dirty state.

Sources/Sentry/SentryMsgPackSerializer.m
```diff
@@ -0,0 +1,87 @@
...
+    NSOutputStream *outputStream = [[NSOutputStream alloc] initWithURL:path append:NO];
+    [outputStream open];
+
+    uint8_t mapHeader = (uint8_t)(0x80 | dictionary.count); // Map up to 15 elements
+    [outputStream write:&mapHeader maxLength:sizeof(uint8_t)];
+
+    [dictionary
+        enumerateKeysAndObjectsUsingBlock:^(NSString *key, id<SentryStreameble> value, BOOL *stop) {
+            NSData *keyData = [key dataUsingEncoding:NSUTF8StringEncoding];
+            uint8_t str8Header = (uint8_t)0xD9; // String up to 255 characteres
```
philipphofmann, line 18: `l`:
```suggestion
            uint8_t str8Header = (uint8_t)0xD9; // String up to 255 characters
```

Sources/Sentry/SentryEnvelope.m
```diff
@@ -198,6 +206,35 @@ - (_Nullable instancetype)initWithAttachment:(SentryAttachment *)attachment
...
+        [[videoURL URLByDeletingPathExtension] URLByAppendingPathExtension:@"dat"];
+
+    [SentryMsgPackSerializer serializeDictionaryToMessagePack:@{
+        @"replay_event" : replayEventData,
+        @"replay_recording" : recording,
+        @"replay_video" : videoURL
+    }
+                                                     intoFile:envelopeContentUrl];
+
+    // TODO: Create and envelope item that accepts and URL as content so we dont need to load the
```
philipphofmann, line 229: I think it's acceptable to release a version without this improvement. I would definitely not add this improvement to this PR.

Sources/Sentry/SentryClient.m
```diff
@@ -472,13 +475,40 @@ - (void)captureSession:(SentrySession *)session
...
+    SentryEnvelopeItem *videoEnvelopeItem =
+        [[SentryEnvelopeItem alloc] initWithReplayEvent:replayEvent
+                                        replayRecording:replayRecording
+                                                  video:videoURL];
+
+    SentryEnvelope *envelope = [[SentryEnvelope alloc] initWithHeader:[SentryEnvelopeHeader empty]
+                                                                items:@[ videoEnvelopeItem ]];
+
+    [self captureEnvelope:envelope];
+}
```
philipphofmann, line 510: `m`: There is some heavy serialization going on here. Maybe it makes sense to move all of this to a background thread.

Sources/Sentry/SentryMsgPackSerializer.m
```diff
@@ -0,0 +1,87 @@
...
+
+- (NSInputStream *)asInputStream
+{
+    return [[NSInputStream alloc] initWithURL:self];
+}
+
+- (NSInteger)streamSize
+{
+    NSFileManager *fileManager = [NSFileManager defaultManager];
+    NSDictionary *attributes = [fileManager attributesOfItemAtPath:self.path error:nil];
```
philipphofmann, line 67: `m`: I think we should handle a possible error here. If we can't get the file size of the path, we must act accordingly.

philipphofmann (commented):
A few comments

Sources/Sentry/include/SentryDataCategory.h
```diff
@@ -14,7 +14,8 @@ typedef NS_ENUM(NSUInteger, SentryDataCategory) {
     kSentryDataCategoryAttachment = 5,
     kSentryDataCategoryUserFeedback = 6,
     kSentryDataCategoryProfile = 7,
-    kSentryDataCategoryUnknown = 8
+    kSentryDataCategoryUnknown = 8,
+    kSentryDataCategoryReplay = 9
```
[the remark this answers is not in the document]
brustolin: I was worried to change it as it could be envelopes in the device at the moment of update.

Sources/Sentry/SentryEnvelope.m
brustolin: Yes, Im not adding to this PR.
Maybe on a follow up PR.

Sources/Sentry/SentryClient.m
brustolin: The code calling captureReplay will be calling from a background thread.

Sources/Sentry/SentryMsgPackSerializer.m
brustolin: Yes!!

Sources/Sentry/SentryMsgPackSerializer.m
```diff
@@ -64,7 +82,11 @@ - (NSInputStream *)asInputStream
 - (NSInteger)streamSize
 {
     NSFileManager *fileManager = [NSFileManager defaultManager];
-    NSDictionary *attributes = [fileManager attributesOfItemAtPath:self.path error:nil];
+    NSError *error;
+    NSDictionary *attributes = [fileManager attributesOfItemAtPath:self.path error:&error];
+    if (attributes == nil) {
+        SENTRY_LOG_DEBUG(@"Could not read file attributes - File: %@ - %@", self, error);
```
philipphofmann, line 88: `l`: W could return `-1` or `0` here.
```suggestion
        SENTRY_LOG_DEBUG(@"Could not read file attributes - File: %@ - %@", self, error);
        return -1;
```

Tests/SentryTests/SentryMsgPackSerializerTests.m
```diff
@@ -0,0 +1,88 @@
+#import "SentryMsgPackSerializer.h"
+#import <Foundation/Foundation.h>
+#import <XCTest/XCTest.h>
+
+@interface SentryMsgPackSerializerTests : XCTestCase
+
+@end
+
+@implementation SentryMsgPackSerializerTests
```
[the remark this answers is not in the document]
philipphofmann: No, I mean using some predefined dictionary with a couple of different values, serializing it with another message pack library, store the output to a file, add this file to our tests, and then comparing the results of the predefined dictionary with the stored file.

Sources/Sentry/include/SentryDataCategory.h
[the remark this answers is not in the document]
philipphofmann: We don't serialize these numbers to disk, so I think we should be good, but I get your point. You can also change it back to be safe.

Tests/SentryTests/SentryMsgPackSerializerTests.m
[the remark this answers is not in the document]
brustolin: Although the reading works in another message pack library (I tested it). The writing is different since I didn't optimise as explained in the SentryMsgPackSerializer.m line 35

```
        // We will always use the 4 bytes data length for simplicity.
        // Worst case we're losing 3 bytes.
```
As long we dont change the tests in this file that check for format, there is nothing to worry about.

Sources/Sentry/SentryClient.m
```diff
@@ -472,13 +475,46 @@ - (void)captureSession:(SentrySession *)session
...
+                                   alwaysAttachStacktrace:NO];
+
+    if (![replayEvent isKindOfClass:SentryReplayEvent.class]) {
+        SENTRY_LOG_DEBUG(@"The event preprocessor didn't update the replay event in place. The "
+                         @"replay was discarded.");
+        return;
+    }
+
+    // breadcrumbs for replay will be send with ReplayRecording
+    replayEvent.breadcrumbs = nil;
```
romtsn, line 499: we probably also got to nullify other stuff that replays do not support currently:
```
replayEvent.extra = nil;
replayEvent.threads = nil;
replayEvent.debugMeta = nil;
```

romtsn: and also this part which will be executed for the replay event I believe, but it shouldn't: https://github.com/getsentry/sentry-cocoa/blob/7b5ebd9f178fc217ef715d615db466d5d77c52ec/Sources/Sentry/SentryClient.m#L557-L562

brustolin: Good catch. I will fix this.

romtsn (approved):
LGTM 🎉

Sources/Sentry/SentryEnvelope.m
```diff
@@ -198,6 +206,37 @@ - (_Nullable instancetype)initWithAttachment:(SentryAttachment *)attachment
     return [self initWithHeader:itemHeader data:data];
 }
 
+- (nullable instancetype)initWithReplayEvent:(SentryReplayEvent *)replayEvent
+                             replayRecording:(SentryReplayRecording *)replayRecording
+                                       video:(NSURL *)videoURL
+{
+    NSData *replayEventData = [SentrySerialization dataWithJSONObject:[replayEvent serialize]];
+    NSMutableData *recording = [NSMutableData data];
+    [recording appendData:[SentrySerialization
```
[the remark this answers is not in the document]
romtsn: The format is:
```
{"segment_id":0}\n[{rrweb_event},{rrweb_event}]
```

Tests/SentryTests/SentryMsgPackSerializerTests.m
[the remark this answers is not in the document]
philipphofmann: Then maybe add a snapshot. Take some sample data, serialize it with `SentryMsgPackSerializer`, validate the oupt against another message pack library, store the contents to a file, and use the file in the tests.

Sources/Sentry/include/SentryMsgPackSerializer.h
```diff
@@ -0,0 +1,33 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol SentryStreamable
```
philipphofmann, line 5: `m`: We could maybe have written that class in Swift. I think we should make it a habit to write new classes without dependencies on the rest of the SDK in Swift.

philipphofmann (approved):
LGTM

Sources/Sentry/include/SentryMsgPackSerializer.h
bruno-garcia: > write new classes without dependencies 

Even those with deps? Are we free now to write just swift unles we need swizzling?

philipphofmann: @brustolin and I had a chat yesterday. @brustolin is working on making it possible to have ObjC and Swift code mixed in our codebase. If we succeed with the setup then yes 😃 💪

brustolin: We can refactor this in swift in another PR

PR #3641, opened 2024-02-15
ref: Add multiple transports to TransportAdapter

philipphofmann:
Add the possibility to add multiple transports in the TransportAdapter, which is internal. This is a preparation for adding support for Spotlight.

#skip-changelog


brustolin (approved):
Personally, I'm missing a way to add/remove transports from the adapter. Let's say the developer wants to enable/disable spotlight in the middle of the session. However, pragmatically, this achieves its purpose.

Let's address spotlight when it becomes necessary.

LGTM! 🚀

PR #3643, opened 2024-02-16
test: reset file manager state in clearTestState()

armcknight:
Noticed there was some stale state causing test failures in #3529. I don't think we ever had something that clears disk state in between tests.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -710,6 +710,11 @@ SentryFileManager ()
     return YES;
 }
 
+- (void)clearDiskState
+{
+    [self removeFileAtPath:self.basePath];
+}
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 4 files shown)


Sources/Sentry/SentryFileManager.m
```diff
@@ -710,6 +710,11 @@ - (BOOL)createDirectoryIfNotExists:(NSString *)path error:(NSError **)error
     return YES;
 }
 
+- (void)clearDiskState
```
philipphofmann, line 713: `l`: This is a duplicate of the existing `deleteAllFolders`.

philipphofmann (approved):
I think we once did, but I can't recall why we don't do this anymore. LGTM

philipphofmann:
Ah, I think now I recall why we removed it. We had trouble because some time ago, the init method of `SentryFileManger` called `deleteOldEnvelopeItems`, which deleted some envelope items async impacting other tests. Now the SentryClient calls it when `deleteOldEnvelopeItems` is true.

Sources/Sentry/SentryFileManager.m
armcknight: hah, nice

armcknight: ah not quite though, deleteAllFolders only deletes the sentry subfolder based on the current DSN; my implementation went one level higher and deleted all dsn-based folders. looks like deleteAllFolders is called explicitly from some certain test classes; i'll just remove deleteAllFolders and leave my new one in place since it's getting called from all tests.

PR #3653, opened 2024-02-19
ci: Increase SauceLabs Concurrency to 4

philipphofmann:
Our plan allows running 4 UI tests in parallel. This should speed up CI.

#skip-changelog


brustolin (approved):
I dont mind this change, but isn't this tied to our SauceLabs account restrictions?

philipphofmann:
> I dont mind this change, but isn't this tied to our SauceLabs account restrictions?

Yes, it is, and we can use 4 in parallel.

PR #3654, opened 2024-02-19
ci: Change to SauceLabs Retry

philipphofmann:
Instead of manually retrying SauceLabs tests with a loop in the CI job, we now use the retry mechanism of SauceLabs, which allows us only to rerun failing tests, which should speed up CI.

#skip-changelog

### DIFF
--- a/.sauce/benchmarking-config.yml
+++ b/.sauce/benchmarking-config.yml
@@ -3,6 +3,7 @@ kind: xcuitest
 sauce:
   region: us-west-1
   concurrency: 2
+  retries: 3
 
 defaults:
   timeout: 20m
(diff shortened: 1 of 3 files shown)


.sauce/benchmarking-config.yml
```diff
@@ -13,14 +14,25 @@ xcuitest:
 
 suites:
   - name: "High-end device"
+    passThreshold: 2
```
philipphofmann, line 17: Unfortunately, you can only add the `passThreshold` and the `smartRetry` for every suite and not for all suites.

brustolin (approved):

philipphofmann:
I think we should remove the whole retry mechanism for SauceLabs instead and fix the underlying issue of failing tests. I'm closing this now.

PR #3657, opened 2024-02-19
ci: Replace iOS 11 with 17 for SauceLabs

philipphofmann:
Xcode 15 only supports debugging on iOS 12. We can remove running UI tests in SauceLabs on iOS 11 and instead run them on iOS 17.

#skip-changelog


philipphofmann:
Replaced by https://github.com/getsentry/sentry-cocoa/pull/3658.

PR #3662, opened 2024-02-20
test: Fix failing UIViewController tests

philipphofmann:
The with GH-3529 added test method writeProfileFile crashed when a profile file already existed. Now, the writeProfileFile method exists and prints a debug message when a profiling file already exists and logs an error when it can't write the profile.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryProfiler.mm
+++ b/Sources/Sentry/SentryProfiler.mm
@@ -469,15 +469,19 @@ writeProfileFile(NSDictionary<NSString *, id> *payload)
     }
 
     if ([fm fileExistsAtPath:pathToWrite]) {
-        SENTRY_LOG_DEBUG(@"Already a%@ profile file present; make sure to remove them right after "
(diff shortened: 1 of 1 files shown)


brustolin (approved):

Sources/Sentry/SentryProfiler.mm
```diff
@@ -469,15 +469,19 @@ + (nullable SentryEnvelopeItem *)createEnvelopeItemForProfilePayload:
     }
 
     if ([fm fileExistsAtPath:pathToWrite]) {
-        SENTRY_LOG_DEBUG(@"Already a%@ profile file present; make sure to remove them right after "
+        SENTRY_LOG_DEBUG(@"Already a %@ profile file present; make sure to remove them right after "
                          @"using them, and that tests clean state in between so there isn't "
                          @"leftover config producing one when it isn't expected.",
             isTracingAppLaunch ? @" launch" : @"");
+        return;
```
armcknight, line 476: Dang good catch, I meant to add an early return here. The extra space before the %@ is unnecessary though, I know it's a bit weird.

PR #3663, opened 2024-02-20
ci: Increase SauceLabs Timeout

philipphofmann:
The UI tests in SauceLabs frequently time out. Let's increase the value to 60m, as it's better to have delayed UI test results than failing CI.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/3662.

#skip-changelog


brustolin (approved):
Let's test it for a while. The timeout could be due to a problem that will cause it to hang regardless of the duration.

PR #3666, opened 2024-02-20
ci: Run UI tests on GH actions

philipphofmann:
Replace running UI tests in SauceLabs with GH actions because SauceLabs tests are unreliable; see [internal Notion decision](https://www.notion.so/sentry/Move-UI-tests-from-SauceLabs-to-GH-action-simulators-e85caa9167314b59ae77d1dd6390e0b3).

#skip-changelog


brustolin (approved):
Perfect, I would just append to the SDK DECISION log.

PR #3671, opened 2024-02-20
feat(Session Replay): Session Replay Integration

brustolin:
## :scroll: Description

Adding Session replay integration

* https://github.com/getsentry/sentry/issues/63255
* 

Unit test

_#skip-changelog_


Sources/Sentry/SentrySessionReplay.m
```diff
@@ -0,0 +1,264 @@
...
+                                                         error:nil];
+        }
+
+        _replayMaker = [[SentryOnDemandReplay alloc] initWithOutputPath:_urlToCache.path];
+        _replayMaker.bitRate = _replayOptions.replayBitRate;
+        _replayMaker.cacheMaxSize = (NSInteger)(full ? _replayOptions.sessionSegmentDuration
+                                                     : _replayOptions.errorReplayDuration);
+        imageCollection = [NSMutableArray array];
+
+        NSLog(@"Recording session to %@", _urlToCache);
```
bruno-garcia, line 91: should we use option diagnostic logger when debug mode is on?

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -0,0 +1,264 @@
...
+
+        _isFullSession = full;
+        if (full) {
+            sessionReplayId = [[SentryId alloc] init];
```
bruno-garcia, line 95: we need a new session id everytime we end/start etc, should this be in some function like "resetReplayId" and invoke that from here?

Sources/Sentry/SentryViewPhotographer.m
```diff
@@ -0,0 +1,159 @@
...
+
+    return _shared;
+}
+
+- (instancetype)init
+{
+    if (self = [super init]) {
+#    if TARGET_OS_IOS
+        _ignoreClasses = @[ UISlider.class, UISwitch.class ].mutableCopy;
+#    endif
```
bruno-garcia, line 27:
```suggestion
#if TARGET_OS_IOS
        _ignoreClasses = @[ UISlider.class, UISwitch.class ].mutableCopy;
#endif
```

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -0,0 +1,264 @@
...
+
+- (void)stop
+{
+    [_displayLink invalidate];
+    _displayLink = nil;
+}
+
+- (void)replayForEvent:(SentryEvent *)event;
+{
+    if (_isFullSession) {
```
romtsn, line 106: I think you wanna get rid of this one, because it will return-early  and won't get to the one below (line 114)

brustolin: This is correct, the rest of the function it isn't.
The DSC would have the replayId already if the full session replay started.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -0,0 +1,164 @@
...
+        let interval = date.timeIntervalSince(_starttime)
+        let imagePath = (_outputPath as NSString).appendingPathComponent("\(interval).png")
+        try? data.write(to: URL(fileURLWithPath: imagePath))
+        _frames.append(SentryReplayFrame(imagePath: imagePath, time: date))
+        
+        while _frames.count > cacheMaxSize {
+            let first = _frames.removeFirst()
+            try? FileManager.default.removeItem(at: URL(fileURLWithPath: first.imagePath))
+        }
+    }
```
vaind, line 59: For Flutter, I think it would be useful to expose `asyncAddFrame(imagePath: NSString, date: Date)`

My idea is to:
- crate a SentryOnDemandReplay from a Flutter swift binding code
- create screenshots in Flutter
- write pngs to the cache directory directly in Flutter
- call asyncAddFrame(filename)
- call createVideoWith() to get the video file path
- send the replay envelope from Flutter

Sources/Sentry/SentrySessionReplay.m
romtsn: ok, but the `sample(errorSampleRate)` call is still missing here I believe

brustolin: Oh, yes. I had to revert back from Swift to Objc and it was lost in the process. Thanks!

Tests/SentryTests/Integrations/SessionReplay/SentryReplayEventTests.swift
```diff
@@ -13,14 +13,14 @@ class SentryReplayEventTests: XCTestCase {
...
-        sut.replayId = replayId
+        sut.eventId = replayId
         
         sut.segmentId = 3
         
         let result = sut.serialize()
         
         expect(result["urls"] as? [String]) == ["Screen 1", "Screen 2"]
-        expect(result["replay_start_timestamp"] as? Int) == 1_000
+        expect(result["replay_start_timestamp"] as? Int) == 1
```
armcknight, line 23: just looking for context here... did we change ms to seconds?

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -0,0 +1,250 @@
...
+        docs = [docs URLByAppendingPathComponent:SENTRY_REPLAY_FOLDER];
+
+        NSString *currentSession = [NSUUID UUID].UUIDString;
+        _urlToCache = [docs URLByAppendingPathComponent:currentSession];
+
+        if (![NSFileManager.defaultManager fileExistsAtPath:_urlToCache.path]) {
+            [NSFileManager.defaultManager createDirectoryAtURL:_urlToCache
+                                   withIntermediateDirectories:YES
+                                                    attributes:nil
+                                                         error:nil];
```
armcknight, line 77: we already have code in SentryFileManager that does this

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        _replayMaker = [[SentryOnDemandReplay alloc] initWithOutputPath:_urlToCache.path];
+        _replayMaker.bitRate = _replayOptions.replayBitRate;
+        _replayMaker.cacheMaxSize = (NSInteger)(full ? _replayOptions.sessionSegmentDuration
+                                                     : _replayOptions.errorReplayDuration);
```
armcknight, line 83: I don't quite understand what's being done here. What is a segment? Is this a customer-facing option? We should explain this in detail in a headerdoc there.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -0,0 +1,77 @@
...
+     */
+    let frameRate = 1
+    
+    /**
+     * The scale related to the window size at which the replay will be created
+     */
+    let sizeScale = 0.8
+   
+    /**
+     * The maximum duration of replays for error events.
```
armcknight, line 42: Is that the time starting when an error is detected? Or is this saying: if you want error replays, then you get 30 seconds from when the session started?

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -0,0 +1,77 @@
...
+     */
+    let errorReplayDuration = TimeInterval(30)
+    
+    /**
+     * The maximum duration of the segment of a session replay.
```
armcknight, line 47: What is a segment? What happens once this threshold is reached?

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -0,0 +1,77 @@
...
+     */
+    let sessionSegmentDuration = TimeInterval(5)
+    
+    /**
+     * Inittialize session replay options disabled
```
armcknight, line 52:
```suggestion
     * Initialize session replay options disabled
```

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public override init() {
+        self.replaysSessionSampleRate = 0
+        self.replaysOnErrorSampleRate = 0
+    }
+    
+    /**
+     * Inittialize session replay options
```
armcknight, line 60:
```suggestion
     * Initialize session replay options
```

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -0,0 +1,77 @@
...
+     *
+     *  sessionSampleRate Indicates the percentage in which the replay for the session will be
+     * created.
+     * @param errorSampleRate Indicates the percentage in which a 30 seconds replay will be send with
+     * error events.
```
armcknight, line 65:
```suggestion
     * - parameters:
     *   - sessionSampleRate: Indicates the percentage in which the replay for the session will be
     *     created.
     *   - errorSampleRate: Indicates the percentage in which a 30 seconds replay will be send with
     *     error events.
```

Sources/Swift/Integrations/SessionReplay/SentryVideoInfo.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+@objcMembers
+class SentryVideoInfo: NSObject {
```
armcknight, line 4: This could be a `struct` if `SentrySessionReplay` were written in Swift, same for some of the other things in Swift/Integrations/SessionReplay

Sources/Swift/Integrations/SessionReplay/SentryPixelBuffer.swift
```diff
@@ -0,0 +1,49 @@
...
+#if os(iOS) || os(tvOS)
+class SentryPixelBuffer {
+    private var pixelBuffer: CVPixelBuffer?
+    private let rgbColorSpace = CGColorSpaceCreateDeviceRGB()
+    private let size: CGSize
+    
+    init?(size: CGSize) {
+        self.size = size
+        let status = CVPixelBufferCreate(kCFAllocatorDefault, Int(size.width), Int(size.height), kCVPixelFormatType_32ARGB, nil, &pixelBuffer)
+        if status != kCVReturnSuccess {
```
armcknight, line 17: `guard`

Sources/Swift/Integrations/SessionReplay/SentryPixelBuffer.swift
```diff
@@ -0,0 +1,49 @@
+#if canImport(UIKit) && !SENTRY_NO_UIKIT
+
+import AVFoundation
+import CoreGraphics
+import Foundation
+import UIKit
+
+#if os(iOS) || os(tvOS)
+class SentryPixelBuffer {
```
armcknight, line 9: `struct`

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -0,0 +1,166 @@
...
+    private var _starttime = Date()
+    private var _frames = [SentryReplayFrame]()
+    private var _currentPixelBuffer: SentryPixelBuffer?
+    
+    var videoWidth = 200
+    var videoHeight = 434
+    
+    var bitRate = 20_000
+    var frameRate = 1
+    var cacheMaxSize = UInt.max
```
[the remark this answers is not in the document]
armcknight: Or even better, just use https://github.com/pointfreeco/swift-tagged

armcknight (commented):
Here's some stuff just from a quick glance through the diff. You could separate out the Swift rewrite of SentryReplayOptions to a separate PR. I don't mind the size otherwise.

How can we demo this, if it's at that stage? When it becomes ready for review, please add instructions with a demo. I wasn't able to see anything in the replays on the web dashboard when trying this. I also set the `sessionSampleRate` to 1 in the options in AppDelegate with no change. I used the button in the sample app to catch errors, but didn't see those in the errors timeline on the web.

Other stuff: 
- Please fix the typo on SentryOptions.h line 277: `@node` -> `@note`.
- Make a note there as well that it's not available for visionOS and only for iOS 16+ (and the tvOS equivalent)

Sources/Sentry/include/SentrySessionReplay.h
```diff
@@ -0,0 +1,35 @@
...
+#    import <UIKit/UIKit.h>
+
+@class SentryReplayOptions;
+@class SentryEvent;
+@class SentryCurrentDateProvider;
+
+NS_ASSUME_NONNULL_BEGIN
+
+API_AVAILABLE(ios(16.0), tvos(16.0))
+@interface SentrySessionReplay : NSObject
```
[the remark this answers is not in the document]
brustolin: We can't have SentryHub and SentryDependencyContainer available for Swift right now.
The alternative is to use wrappers, but its too late to do that now.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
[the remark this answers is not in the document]
brustolin: This properties are standardised already. Video size is always in pixel, frame rate is always hz, bitRate is always bytes per second, cache size is not in bytes, this is number of frames stored, this one could use a name improvement.

IMO we should not use external libraries and add dependencies in the SDK.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -0,0 +1,166 @@
...
+    
+    init(outputPath: String) {
+        self._outputPath = outputPath
+        _onDemandDispatchQueue = DispatchQueue(label: "io.sentry.sessionreplay.ondemand")
```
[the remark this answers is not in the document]
brustolin: Eventually, when we figure it out DependencyContainer for swift.

Sources/Sentry/include/SentrySessionReplayIntegration.h
```diff
@@ -0,0 +1,12 @@
+#import "SentryBaseIntegration.h"
+#import "SentryDefines.h"
+#import "SentrySwift.h"
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+#if SENTRY_HAS_UIKIT && !TARGET_OS_VISION
+@interface SentrySessionReplayIntegration : SentryBaseIntegration <SentryIntegrationProtocol>
```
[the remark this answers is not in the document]
brustolin: Because of dependencies. 
Right now it will be very trick to write integrations in swift

Sources/Sentry/include/SentrySessionReplay.h
```diff
@@ -0,0 +1,35 @@
...
+ * If full is @c YES, we transmit the entire session to sentry.
+ */
+- (void)start:(UIView *)rootView fullSession:(BOOL)full;
+
+/**
+ * Stop recording the session replay
+ */
+- (void)stop;
+
+- (void)replayForEvent:(SentryEvent *)event;
```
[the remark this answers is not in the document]
brustolin: Sents a replay for an event.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+    
+    private func asyncAddFrame(image: UIImage) {
+        guard let data = resizeImage(image, maxWidth: 300)?.pngData() else { return }
+        
+        let date = Date()
+        let interval = date.timeIntervalSince(_starttime)
+        let imagePath = (_outputPath as NSString).appendingPathComponent("\(interval).png")
+        try? data.write(to: URL(fileURLWithPath: imagePath))
```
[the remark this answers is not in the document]
brustolin: Personally I believe SentryFileManager for this class is unnecessary complexity. 
I plan to revisit this later, we need a working alpha asap right now.


> that's like a do...catch with an empty catch block

And sometimes that is fine.
But I guess I could have a earlier exit in here.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -0,0 +1,166 @@
...
+    func releaseFramesUntil(_ date: Date) {
+        _onDemandDispatchQueue.async {
+            while let first = self._frames.first, first.time < date {
+                self._frames.removeFirst()
+                try? FileManager.default.removeItem(at: URL(fileURLWithPath: first.imagePath))
+            }
+        }
+    }
+    
+    func createVideoWith(duration: TimeInterval, beginning: Date, outputFileURL: URL, completion: @escaping (SentryVideoInfo?, Error?) -> Void) throws {
```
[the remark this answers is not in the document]
brustolin: Result does not work with objc.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -0,0 +1,166 @@
...
+            if let image = UIImage(contentsOfFile: imagePath) {
+                let presentTime = CMTime(seconds: Double(frameCount), preferredTimescale: CMTimeScale(self.frameRate))
+                frameCount += 1
+                
+                if self._currentPixelBuffer?.append(image: image, pixelBufferAdapter: pixelBufferAdaptor, presentationTime: presentTime) != true {
+                    completion(nil, videoWriter.error)
+                }
+            }
+            
+            if frameCount >= frames.count {
```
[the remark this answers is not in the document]
brustolin: Good that you point this out. This condition needs to reach its conclusion, but we do have a problem  elsewhere.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -0,0 +1,88 @@
...
+#import "SentryDependencyContainer.h"
+#import "SentryGlobalEventProcessor.h"
+#import "SentryHub+Private.h"
+#import "SentryOptions.h"
+#import "SentryRandom.h"
+#import "SentrySDK+Private.h"
+#import "SentrySessionReplay.h"
+#import "SentrySwift.h"
+
+#if SENTRY_HAS_UIKIT && !TARGET_OS_VISION
```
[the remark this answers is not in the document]
brustolin: `#if defined(__IPHONE_16_0)` in SentryCoreGraphicsHelpers its a hack for our CI, It does not really matter for production.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -0,0 +1,166 @@
...
+        videoWriterInput.requestMediaDataWhenReady(on: _onDemandDispatchQueue) { [weak self] in
+            guard let self = self else { return }
+            
+            let imagePath = frames[frameCount]
+            if let image = UIImage(contentsOfFile: imagePath) {
+                let presentTime = CMTime(seconds: Double(frameCount), preferredTimescale: CMTimeScale(self.frameRate))
+                frameCount += 1
+                
+                if self._currentPixelBuffer?.append(image: image, pixelBufferAdapter: pixelBufferAdaptor, presentationTime: presentTime) != true {
+                    completion(nil, videoWriter.error)
```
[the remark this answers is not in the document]
brustolin: `guard` is more verbose when used just for an early return; usually, I prefer to use it only for optional binding.

And since the use of `guard` over `if` is a matter of preference, I won't be retroactively changing it in this PR, but I will be more aware of it from now on (despite not liking it).

Sources/Sentry/SentryCoreGraphicsHelper.m
```diff
@@ -0,0 +1,18 @@
+#import "SentryCoreGraphicsHelper.h"
+#if SENTRY_HAS_UIKIT
+@implementation SentryCoreGraphicsHelper
++ (CGMutablePathRef)excludeRect:(CGRect)rectangle fromPath:(CGMutablePathRef)path
+{
+#    if (TARGET_OS_IOS || TARGET_OS_TV)
```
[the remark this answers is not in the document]
brustolin: Yes, but I believe the original approach is more clear.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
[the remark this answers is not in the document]
armcknight: I'm fine with not bringing in the dependency, but please record the assumptions somewhere. Don't assume others will understand what the units are. Having a variable named "bitRate" that has a unit of "byte," which is 8 bits, is confusing. 20,000 bits vs 20,000 bytes is quite a difference. it should be `byteRate` at the very least.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
vaind: @brustolin WDYT?

brustolin:
CI errors in the PR will be fixed in the next one: https://github.com/getsentry/sentry-cocoa/pull/3816

SentryTestUtils/TestCurrentDateProvider.swift
```diff
@@ -43,6 +43,10 @@ public class TestCurrentDateProvider: SentryCurrentDateProvider {
         setDate(date: date().addingTimeInterval(TimeInterval(nanoseconds) / 1e9))
         internalSystemTime += nanoseconds
     }
+    
+    public func advanceBy(interval: TimeInterval) {
+        setDate(date: date().addingTimeInterval(interval))
+    }
```
philipphofmann, line 49: Oh yeah, that makes sense. Thanks 🥇

Sources/Sentry/SentryDateUtil.m
```diff
@@ -40,7 +40,7 @@ + (NSDate *_Nullable)getMaximumDate:(NSDate *_Nullable)first andOther:(NSDate *_
 
 + (long)millisecondsSince1970:(NSDate *)date
 {
-    return (NSInteger)([date timeIntervalSince1970] * 1000);
+    return (long)([date timeIntervalSince1970] * 1000);
```
philipphofmann, line 43: 👍

philipphofmann (approved):
LGTM, for merging this to main hidden behind a feature flag. As discussed, I'm going to give this a full review after it's merged to main, and I didn't check the functionality of the code yet.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
brustolin: Yeah, if its useful to Flutter we can make it public.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -102,20 +111,8 @@ class SentryOnDemandReplay: NSObject {
...
-        var actualEnd = Date()
-        
-        for frame in _frames {
-            if frame.time < beginning { continue } else if frame.time > end { break }
-            if frame.time < start { start = frame.time }
-            
-            actualEnd = frame.time
-            frames.append(frame.imagePath)
-        }
+        let (frames, start, end) = filterFrames(beginning: beginning, end: beginning.addingTimeInterval(duration))
```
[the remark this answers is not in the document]
brustolin: I understand. But for private functions inside the class itself I think it's fine. No need to declare another struct for this.

PR #3673, opened 2024-02-20
test: Enable App Launch Profiling UI tests

philipphofmann:
GH-3665 didn't enable the app launch profiling tests because they were still disabled in Xcode. This PR fixes that and also uses more accessibility identifiers in the profiling tests.

#skip-changelog


brustolin (approved):
Thanks for this @philipphofmann

PR #3674, opened 2024-02-21
Feat(Session Replay): Replay Settings

brustolin:
Settings for session replay in the Options.

_#skip-changelog_


Sources/Sentry/Public/SentryReplayOptions.h
```diff
@@ -0,0 +1,38 @@
...
+@property (nonatomic) float replaysSessionSampleRate;
+
+/**
+ * Indicates the percentage in which a 30 seconds replay will be send with error events.
+ * @discussion Specifying @c 0 means never, @c 1.0 means always.
+ * @note The value needs to be >= 0.0 and \<= 1.0. When setting a value out of range the SDK sets it
+ * to the default.
+ * @note The default is @c 0.
+ */
+@property (nonatomic) float replaysOnErrorSampleRate;
```
bruno-garcia, line 23: makes sense to prefix it given, objc

bruno-garcia (approved):

Sources/Sentry/Public/SentryReplayOptions.h
brustolin: I dont see why the property name would be an issue.

PR #3679, opened 2024-02-23
Fix: Missing argument for Craft step

brustolin:
It was missing an argument for Craft step in order for the target merge to work.

_#skip-changelog_


philipphofmann (approved):
LGTM

PR #3693, opened 2024-02-29
feat(Session Replay): Prepare Session event

brustolin:
Handling session event during `prepareEvent`.

_#skip-changelog_


Sources/Sentry/SentryScope.m
```diff
@@ -496,7 +496,7 @@ - (SentryEvent *__nullable)applyToEvent:(SentryEvent *)event
         event.fingerprint = fingerprints;
     }
 
-    if (event.breadcrumbs == nil) {
+    if (event.breadcrumbs == nil && ![event.type isEqualToString:SentryEnvelopeItemTypeReplayVideo]) {
```
brustolin: @romtsn, @bruno-garcia
Apart from breadcrumb, should we clean replay event of any other data that usually goes with events?

romtsn: I dunno if you set these ones right now, but I'm clearing them out:
```
replayEvent.extra = nil;
replayEvent.threads = nil;
replayEvent.debugMeta = nil;
```

brustolin: Perfect! I will just save some cycles and not even copy them to the event.

philipphofmann (approved):
Some tests for this would be nice. If they already exist, LTGM. Otherwise, please add some.

PR #3698, opened 2024-03-01
ref: Uncouple DependencyContainer from SentryHub

brustolin:
Removed SentryHub dependency from SentryDependencyContainer to avoid cyclic reference that caused a problem in this PR.

_#skip-changelog_


philipphofmann (approved):
That makes sense. Thanks 👍

PR #3699, opened 2024-03-01
fix: Add context to event with CrashHandler disabled

### LINKED ISSUE #3694: Make enriched Scope independent of SentryCrashIntegration
### Platform

iOS

### Environment

Production, Develop

### Installed

CocoaPods

### Version

8.20.0

_No response_

### Steps to Reproduce

1. disable `SentryCrashIntegration` with `enableCrashHandler` or by removing the integration
2.  configure a scope
3. serialize the scope => empty

### Expected Result

The scope should still be augmented with all the device infos.

### Actual Result

empy scope

could be but not sure about the wanted refactor

brustolin:
## :scroll: Description

If crashHandler is disabled the SDK does not add device and app context to events.

⚠️ First review [this PR](https://github.com/getsentry/sentry-cocoa/pull/3698)

close #3694 

Unit test

## :pencil: Checklist


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Make SentryFramesTracker available for HybridSDKs ([#3683](https://github.com/getsentry/sentry-cocoa/pull/3683))
 - Make SentrySwizzle available for HybridSDKs ([#3684](https://github.com/getsentry/sentry-cocoa/pull/3684))
+- Add context to event with CrashHandler disabled (#3699)
```
philipphofmann, line 9: `l`: I think it's the crash integration.
```suggestion
- Add context to event with CrashIntegration disabled (#3699)
```

Sources/Sentry/SentryHub.m
```diff
@@ -77,6 +86,114 @@ - (instancetype)initWithClient:(nullable SentryClient *)client
     return self;
 }
 
++ (void)enrichScope:(SentryScope *)scope crashWrapper:(SentryCrashWrapper *)crashWrapper
```
philipphofmann, line 89: `m`: The hub is already huge. I would prefer to have this code in an extra file. It can be a helper method somewhere.

Sources/Sentry/SentrySDK.m
```diff
@@ -62,6 +63,11 @@ + (void)initialize
         sentrySDKappStartMeasurementLock = [[NSObject alloc] init];
         startInvocations = 0;
         _detectedStartUpCrash = NO;
+
+        // The system monitor api is a data bag for system info
+        // that we need for every event, and because this is static info
+        // we need to start it as soon as possible.
+        sentrycrashcm_system_getAPI()->setEnabled(YES);
```
philipphofmann, line 70: `m`: I'm unsure if calling this in `initialize` isn't a bit too early. We only need it when inializing the hub. I think it would be better to call this before enriching the scope. We can call this code multiple times, as when it's initialized it's almost a NoOp. Furthermore, I think we should add a test that fails if we don't call this method in enrich scope, because if we don't the damage is big.

Tests/SentryTests/SentryHubTests.swift
```diff
@@ -178,6 +178,11 @@ class SentryHubTests: XCTestCase {
         XCTAssertNil(hub.scope.serialize()["breadcrumbs"])
     }
     
+    func testScopeEnriched() {
+        let hub = fixture.getSut(fixture.options, Scope())
+        expect(hub.scope.contextDictionary.allValues.isEmpty) == false
```
philipphofmann, line 183: `m`: Testing for some actual values would improve this test case.

philipphofmann (commented):
The `SentryCrashIntegrationTests` contains the tests for enriching the scope context. I think it would be great to move these tests to a different location.

Sources/Sentry/SentrySDK.m
```diff
@@ -6,6 +6,7 @@
 #import "SentryBreadcrumb.h"
 #import "SentryClient+Private.h"
 #import "SentryCrash.h"
+#import "SentryCrashMonitor_System.h"
```
brustolin, line 9:
```suggestion
```

PR #3708, opened 2024-03-04
ref: Convert SentryId to Swift

brustolin:
I can see this being the beginning of a project refactoring to Swift. 
Also, I want to try converting something to see how much effort it is.

_#skip-changelog_


Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -2,8 +2,9 @@
 
 #if SENTRY_TARGET_PROFILING_SUPPORTED
 
-#    import "SentryId.h"
+#    import "SentrySwift.h"
```
brustolin, line 5:
```suggestion
```

Tests/SentryTests/Protocol/SentryEventTests.swift
```diff
@@ -1,4 +1,4 @@
-import Sentry
+@testable import Sentry
```
brustolin, line 1:
```suggestion
import Sentry
```

Sources/Swift/Extensions/StringExtensions.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+
+extension String {
+    /// Retrieve a character at a specific index in the string.
+    subscript (value: Int) -> Character {
```
philipphofmann, line 5: `m`: Codecov shows that we don't use all of these methods here. Either we remove them, or we keep them and we also add tests for these, please.

philipphofmann (approved):
I think it would be smart to add refactorings of public ObjC classes to Swift as an improvement to the Changelog, because we communicate this then clearly and also if something isn't working properly it's easier to find the corresponding release. WDYT?

LGTM

brustolin:
> I think it would be smart to add refactorings of public ObjC classes to Swift as an improvement to the Changelog.

IMO, a changelog is a way to communicate product changes and fixes to the user. I believe this suggestion would clutter the changelog with information that nobody cares about.

This is my 50 cents. If you still think we should do it we do it.

philipphofmann:
> > I think it would be smart to add refactorings of public ObjC classes to Swift as an improvement to the Changelog.
> 
> IMO, a changelog is a way to communicate product changes and fixes to the user. I believe this suggestion would clutter the changelog with information that nobody cares about.
> 
> This is my 50 cents. If you still think we should do it we do it.

Also fine, then skip it.

PR #3710, opened 2024-03-05
ref: EnvelopeItemHeader: Init without filename

philipphofmann:
Add an extra initializer without a filename, which is required for DDM.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryEnvelopeItemHeader.m
+++ b/Sources/Sentry/SentryEnvelopeItemHeader.m
@@ -26,9 +26,8 @@
                    filenname:(NSString *)filename
                  contentType:(NSString *)contentType
 {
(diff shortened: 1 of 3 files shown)


Sources/Sentry/SentryEnvelopeItemHeader.m
```diff
@@ -11,6 +11,16 @@ - (instancetype)initWithType:(NSString *)type length:(NSUInteger)length
     return self;
 }
 
+- (instancetype)initWithType:(NSString *)type
```
brustolin, line 14: l: The initializer with  filename could use this initializer instead of the other one.

brustolin (approved):
LGTM!!

PR #3714, opened 2024-03-06
chore: Set HEADER_SEARCH_PATHS to expose _SentryPrivate to Objc

brustolin:
Setting HEADER_SEARCH_PATHS to `$(SRCROOT)/Sources/Sentry/include` recursive, so Objc compiler can use module.modulemap

_#skip-changelog_


philipphofmann (approved):
LGTM

PR #3716, opened 2024-03-07
ref: Use SentryPrivate for SentryRandom

philipphofmann:
Add SentryRandom to SentryPrivate.h and remove it from the bridging headers. This is a validation to see if we can add headers to SentryPrivate, which we use in Swift code, and then write tests for that Swift code.

#skip-changelog


Sources/Configuration/Sentry.xcconfig
```diff
@@ -5,3 +5,7 @@ PRODUCT_BUNDLE_IDENTIFIER = io.sentry.Sentry
 CURRENT_PROJECT_VERSION = 8.21.0
 
 MODULEMAP_FILE = $(SRCROOT)/Sources/Resources/Sentry.modulemap
+
+// This config is required so the test code can access the SentryPrivate module. Removing this setting
+// leads to an error: no such module '_SentryPrivate' when including the XCFramework with Carthage
+HEADER_SEARCH_PATHS = $(SRCROOT)/Sources/Sentry/include/**
```
brustolin, line 11: l: This setting is not only for test but also for the framework to find the module.

philipphofmann: Better now, @brustolin?

brustolin (approved):
LGTM

Next step is to convert SentryRandom to swift :D
But I know that you did this to understand the process.

PR #3718, opened 2024-03-07
ref: Convert CurrentDate to Swift

philipphofmann:
Using CurrentDate in Swift code is essential. Instead of exposing the existing ObjC class to Swift, I converted the class to Swift.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/3716 and is required for DDM.

#skip-changelog


Sources/Swift/Helper/SentryCurrentDateProvider.swift
```diff
@@ -0,0 +1,20 @@
+@_implementationOnly import _SentryPrivate
```
brustolin, line 1: l: We dont need _SentryPrivate here
```suggestion
```

Sources/Swift/Helper/SentryCurrentDateProvider.swift
```diff
@@ -0,0 +1,20 @@
...
+
+@objcMembers class SentryCurrentDateProvider: NSObject {
+
+    func date() -> Date {
+        return Date()
+    }
+
+    func dispatchTimeNow() -> DispatchTime {
+        return .now()
+    }
```
brustolin, line 11: l: Are we really not using this? At least we dont touch it during tests.
If so I would remove it.

brustolin (approved):
LGTM

Sources/Swift/Helper/SentryCurrentDateProvider.swift
philipphofmann: Yes we do, we need it because of `getAbsoluteTime()`, which is in `SentryTime.h`

PR #3723, opened 2024-03-11
ref: Remove SentrySwift import for trace context

philipphofmann:
The SentryTraceContext doesn't need to SentrySwift import. It only needs a reference to SentryId.

#skip-changelog

PR #3726, opened 2024-03-12
ref: Cleanup PrivateSDKOnly imports

philipphofmann:
This PR is based on https://github.com/getsentry/sentry-cocoa/pull/3723.

#skip-changelog

### DIFF
--- a/Sources/Sentry/PrivateSentrySDKOnly.mm
+++ b/Sources/Sentry/PrivateSentrySDKOnly.mm
@@ -1,4 +1,5 @@
 #import "PrivateSentrySDKOnly.h"
+#import "SentryAppStartMeasurement.h"
 #import "SentryBreadcrumb+Private.h"
 #import "SentryClient.h"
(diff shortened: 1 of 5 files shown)


Sources/Sentry/PrivateSentrySDKOnly.mm
```diff
@@ -1,4 +1,5 @@
 #import "PrivateSentrySDKOnly.h"
+#import "SentryAppStartMeasurement.h"
```
brustolin, line 2: When we're doing some changes like this to prepare to another PR, I would suggest to add a little bit more information on the PR's description to explain the motivation, because based on this PR alone I dont know why you're moving headers around.

brustolin (approved):
LGTM

Sources/Sentry/PrivateSentrySDKOnly.mm
philipphofmann: Ups sorry. Yes will do that.

PR #3732, opened 2024-03-12
ref: Remove forward declarations for Envelope

philipphofmann:
Having unused forward declarations in the SentryEnvelope.h doesn't make sense and makes it harder to expose that specific class to Swift. Therefore, we can remove them.

#skip-changelog

### DIFF
--- a/Sources/Sentry/include/HybridPublic/SentryEnvelope.h
+++ b/Sources/Sentry/include/HybridPublic/SentryEnvelope.h
@@ -8,8 +8,12 @@
 
 #endif
 
-@class SentryEvent, SentrySession, SentryId, SentryUserFeedback, SentryAttachment,
-    SentryTransaction, SentryClientReport, SentryEnvelopeItemHeader;
+@class SentryEvent;
+@class SentrySession;
+@class SentryId;
+@class SentryUserFeedback;
+@class SentryAttachment;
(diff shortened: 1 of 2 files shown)


Sources/Sentry/include/HybridPublic/SentryEnvelope.h
```diff
@@ -8,8 +8,12 @@
 
 #endif
 
-@class SentryEvent, SentrySession, SentryId, SentryUserFeedback, SentryAttachment,
-    SentryTransaction, SentryClientReport, SentryEnvelopeItemHeader;
+@class SentryEvent;
```
philipphofmann, line 11: I removed `SentryTransaction` and `SentryClientReport`.

brustolin (approved):
LGTM.

Have you encountered any issues with forward class declaration? This should serve as a workaround to avoid importing the actual header, and it worked for me.

Of course, if its not being used we should remove it.

Sources/Sentry/include/HybridPublic/SentryEnvelope.h
armcknight: Thank you for also splitting it into individual declarations 🙏🏻

armcknight (approved):

PR #3734, opened 2024-03-13
feat: reuse the vendorIdentifier if already fetched.

leejunhui:
## :scroll: Description

Reuse the `vendorIdentifier` if already fetched beforehand.Due to the China mainland privacy policy, apps cannot call the UIDevice.currentDevice.identifierForVendor multiple times in a short period of time.So to simultaneously use Sentry SDK and  meet the privacy policy.It's maybe the simplest solution that I could figure out right now😂

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Improvements
 
 - Remove SentryPrivate library (#3623)
+- Cache "vendorIdentifier" (#3734)
 
 ### Fixes
 
--- a/Sources/Sentry/SentryAppStateManager.m
+++ b/Sources/Sentry/SentryAppStateManager.m
@@ -181,17 +181,15 @@ SentryAppStateManager ()
     bool isDebugging = self.crashWrapper.isBeingTraced;
 
     UIDevice *device = [UIDevice currentDevice];
-    
-    // Prevent multiple calls to UIDevice.currentDevice.identifierForVendor in a short period of time, thereby violating privacy compliance requirements in China Mainland.
-    NSString *vendorId = self.vendorIdentifier;
-    if (self.vendorIdentifier.length == 0) {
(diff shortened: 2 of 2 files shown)


kahest:
Hey @leejunhui thanks for your contribution - could you provide a source for this requirement? We'd like to make sure we understand the requirement correctly, and see if there's other things we need to be aware of.

leejunhui:
hi @kahest , the policy source is the [**Information security technology-Basic requirements for
collecting personal information in mobile internet applications**](https://openstd.samr.gov.cn/bzgk/gb/newGbInfo?hcno=977D9EBB32ABF0A7DD6A1215969FE57A), Please note that this is a Chinese webpage, you can use Google's automatic translation function to access it

leejunhui:
hi @brustolin , I am very gald to write the tests to prevent problems, but due to the development of business needs, it may take a while to complete the writing of tests. Please give me some time, thanks a lot

kahest:
For clarity - the passages are about ensuring that the "Frequency of automatic collection of personal information should be the minimum frequency necessary to achieve the business functions of the product or service". I wouldn't relate this to the technical implementation of whether personal information is temporarily cached in memory or retrieved from the operating system API instead - this is more about only collecting/sending data if and when you need it.

That said, it's fair to apply the change you contributed if we can ensure it doesn't break functionality. I don't think it's required to comply with the terms above.

kahest:
We will [remove usage of the id for some use cases](https://github.com/getsentry/sentry-cocoa/issues/4022) and we'll keep the current behaviour for the once [usage in our OOM heuristics](https://github.com/getsentry/sentry-cocoa/blob/5e579afa2abba914448ad2b79995cbfbb142f5f3/Sources/Sentry/SentryWatchdogTerminationLogic.m#L71). For this, the id is not sent off device, so this PR is obsolete.

Issue #2155, opened 2022-09-14
Use absolute time for tracing duration


philipphofmann:
### Description

Currently, our transactions and spans use NSDate, which is not a safe API for time-sensitive measurements because
the time can be changed in iOS settings or by the system when it recalibrates with a remote
source of truth. Instead, we could switch to absolute system timestamps as used by profiling.
https://github.com/getsentry/sentry-cocoa/blob/eefdeedf0d7edd17ac03e89ab7a4da75f24bc1ff/Sources/Sentry/include/SentryTime.h#L10-L14

We could achieve this by internally keeping track of absolute timestamps and converting these to NSDate when finishing a transaction. We might also have to adjust span timestamps, as the system time could have changed during the lifetime of a transaction.

armcknight:
See also https://github.com/getsentry/sentry-cocoa/discussions/1094#discussioncomment-3521972

philipphofmann:
Consider updating the develop docs to use wall time for tracing.

armcknight:
The problem is that you cannot map system time to NSDate in the presence of the edge cases I've described. My proposition is that we keep both, using system time for anything requiring reliable timing like performance transactions and profiling. Then, when we go to construct an envelope, we'll record both, and perform checks so we can set a flag on any payload where we detect they have drifted due to NSDate being changed out from underneath.

As for `clock_gettime`, we already use that API here: https://github.com/getsentry/sentry-cocoa/blob/411a940bcf9f7f6808f0282bc583827284d4e88b/Sources/Sentry/SentryTime.mm#L12-L15

armcknight:
Just saw this bit of UI on a transaction, which is like what I had in mind to notify if we detect a jump in the time reported by NSDate: 
<img width="916" alt="image" src="https://user-images.githubusercontent.com/3241469/217341627-70304e60-a36c-4291-9585-8da8791b9b8f.png">

PR #3736, opened 2024-03-14
feat: append launch profile data to app start transactions/spans

### LINKED ISSUE #3725: Integrate launch profile data into app start transactions/spans
### Description

We launched the app launch profile feature to only create a dedicated transaction to attach profile data to. It starts the profiled trace in SentryTracer.load and stops it in SentrySDK.startWIthOptions.

But, any automatic perf txn that gets the app start spans won't even be started until some time after SentrySDK.startWIthOptions. So, we need to special case this to allow the profile to continue, and also to discard the dedicated txn, since we'll attach all the launch profile data to the app start txn.

Part of this is reverting the changes made in #2772, where we avoided carrying over the modified start time of that app start txn for profile data slicing.

Essentially, we want to make this:
image
include this data:
image

# Testing

- The profiling unit tests check the `originalStartTimestamp` that will be removed, so we'll check `transaction.startTimestamp` instead.
- Add a scenario to simulate this in the launch profiling integration UI test

armcknight:
Following on app launch profiling that only used a dedicated span to attach the profiles, we will now attach the profile data to an automatic app start transaction if one will be recorded, otherwise continue with the same dedicated txn.

Closes #3725

# Changes

- Change SentrySDK.startWithOptions from always stopping the launch profile to only stopping it if `SentryUIViewControllerPerformanceTracker.shared.enableWaitForFullDisplay` is `NO`, otherwise, that trace will eventually end and get all the profile data. Then, `SentryTimeToDisplayTracker` will call `stopLaunchProfile` passing `nil` as the hub argument which will stop the dedicated tracer without actually sending a duplicate dedicated transaction, since the profile data we previously attached to it is now on the app start txn.
- Records the system time `SentrySysctl.load` is called and keeps that as the `processStartSystemTimestamp` to report to `SentryAppStartMeasurement` as the `appStartSystemTimestamp`, which is used when SentryTracer requests a slice of profile data to attach to the app start transaction. Profile slicing needs a _system_ timestamp which it uses to keep relevant samples and compute relative timestamps. it's set as the SentryTransaction.startSystemTime for app start transactions.
- We no longer need to keep track of `originalStartTimestamp` in `SentryTracer`. We now take either the actual start time of the tracer (which was always the case outside of app start transactions, which modifies `SentryTracer.startTimestamp`) or  `SentryAppStartMeasurement.runtimeInitTimestamp` in the case of app start txns, since this corresponds to the time that we're recording `SentrySysctl.processStartSystemTimestamp` as mentioned above.
	- Note that for prewarmed launches, we don't record a span using `runtimeInitTimestamp`. But, we have no reliable way to test what will happen for these launch types, so we are going to postpone any further treatment of that edge case.

Before (the profile data starts at the extreme right end): 

image

After:  

image


Sources/Sentry/SentryTracer.m
```diff
@@ -607,7 +601,24 @@ - (void)finishInternal
...
+        }
+#    endif // SENTRY_HAS_UIKIT
+
+        if (startTimestamp == nil) {
+            startTimestamp = self.startTimestamp;
+        }
+        if (!SENTRY_ASSERT_RETURN(startTimestamp != nil,
+                @"A transaction with a profile should have a start timestamp already. We will "
+                @"assign the current time but this will be incorrect.")) {
+            startTimestamp = [SentryDependencyContainer.sharedInstance.dateProvider date];
```
brustolin, line 618: m: Does it make sense if its incorrect?

brustolin (commented):
Looking good.
Just a couple of comments.

Sources/Sentry/SentryTracer.m
```diff
@@ -686,7 +698,19 @@ - (SentryTransaction *)toTransaction
...
+        // if we have an app start span, use its app start timestamp. otherwise use the tracer's
+        // start system time as we currently do
+        SENTRY_LOG_DEBUG(@"Tracer start time: %llu", self.startSystemTime);
+
         transaction.startSystemTime = self.startSystemTime;
+#    if SENTRY_HAS_UIKIT
+        if (appStartMeasurement != nil) {
+            SENTRY_LOG_DEBUG(@"Assigning transaction start time as app start system time (%llu)",
+                appStartMeasurement.appStartSystemTimestamp);
+            transaction.startSystemTime = appStartMeasurement.appStartSystemTimestamp;
```
philipphofmann, line 710: `h`: The app `transaction.startSystemTime` can be quite different from the `transaction.startTimestamp` for app start transactions when the app start is not prewarmed, because then the `transaction.startTimestamp` is the process start timestamp and the `transaction.startSystemTime` then is the runtime init timestamp. I'm not sure if this is a problem.

Sources/Sentry/SentryTracer.m
philipphofmann: I agree; maybe it makes sense to ditch the profile in that case instead of using an incorrect time.

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -128,13 +133,19 @@ - (void)framesTrackerHasNewFrame:(NSDate *)newFrameDate
         [self.initialDisplaySpan finish];
         if (!_waitForFullDisplay) {
             [SentryDependencyContainer.sharedInstance.framesTracker removeListener:self];
+#    if SENTRY_TARGET_PROFILING_SUPPORTED
+            stopLaunchProfile(nil);
```
philipphofmann, line 137: `h`: When TTFD is enabled, this stops the launch profile and the app start transaction when we report TTID. Instead, when TTFD is enabled the app start transaction must continue until the user calls TTFD. When the user doesn't call TTFD the tracer finishes after some timeout. It would be great to add tests for these use cases to `SentryTimeToDisplayTrackerTest`.

Sources/Sentry/SentrySDK.m
```diff
@@ -214,8 +218,19 @@ + (void)startWithOptions:(SentryOptions *)options
 #endif // TARGET_OS_IOS && SENTRY_HAS_UIKIT
 
 #if SENTRY_TARGET_PROFILING_SUPPORTED
+        BOOL shouldStopLaunchProfile = YES;
+#    if SENTRY_HAS_UIKIT
```
philipphofmann, line 222: `m`: This code changes quite some logic. IMO, we should add tests for this or have to change existing tests.

Sources/Sentry/SentryTracer.m
```diff
@@ -607,7 +601,24 @@ - (void)finishInternal
 
 #if SENTRY_TARGET_PROFILING_SUPPORTED
     if (self.isProfiling) {
-        [self captureTransactionWithProfile:transaction];
+        NSDate *startTimestamp;
+
+#    if SENTRY_HAS_UIKIT
```
philipphofmann, line 606: `m`: I don't see updated tests or new tests for this logic. Please add some.

Sources/Sentry/SentryTimeToDisplayTracker.m
philipphofmann: I think calling ` stopLaunchProfile` in the `finishCallback` of the tracer here should do the job:

https://github.com/getsentry/sentry-cocoa/blob/69fcf6f477b69599124a017d5b819db58fa20629/Sources/Sentry/SentryTimeToDisplayTracker.m#L83-L84

Sources/Sentry/SentrySDK.m
armcknight: That's exactly what the changes here are for: https://github.com/getsentry/sentry-cocoa/pull/3736/files#diff-e97a92e3a9df5dec5f35a1ee0603359d5857c05051313cc2c55897e54e7da05bL798-R813

These changes change what the start timestamp will be for the profile. The tests were changed to account for that. They now ensure that if the profile is attached to an app start txn, that the profile start time matches that of the app start txn.

Sources/Sentry/SentryTracer.m
armcknight: It doesn't necessarily make sense, but this is how we previously handled it. It's just that the same decision had to move to this new location in code. I would say let's not worry about it because this will not matter once we move to continuous profiling: we won't have to deal with the potentially nil timestamp from any txn, we will be able to just use our own.

Sources/Sentry/include/SentrySpan.h
```diff
@@ -68,6 +68,8 @@ SENTRY_NO_INIT
  */
 @property (nullable, nonatomic, strong) NSDate *startTimestamp;
 
+@property (nonatomic, assign) uint64_t startSystemTimestamp;
```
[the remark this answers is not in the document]
armcknight: @philipphofmann which platform did you build for? When I remove it and build for iOS, it fails here: 
image
Added the gate in https://github.com/getsentry/sentry-cocoa/pull/3736/commits/d6d8ca9945d92d59abbad1bd17f20b808829e589

Sources/Sentry/Public/SentrySpanProtocol.h
```diff
@@ -65,6 +66,13 @@ NS_SWIFT_NAME(Span)
  */
 @property (nullable, nonatomic, strong) NSDate *startTimestamp;
 
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+/**
+ * The system time the span started. Needed to slice profiling data to the span bounds.
+ */
+@property (nonatomic, assign) uint64_t startSystemTimestamp;
```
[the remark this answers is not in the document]
armcknight: It's used as mentioned [here](https://github.com/getsentry/sentry-cocoa/pull/3736#discussion_r1534818007), I'll try to move it somewhere that is not public.

Sources/Sentry/SentryAppStartTracker.m
```diff
@@ -213,6 +213,7 @@ - (void)buildAppStartMeasurement:(NSDate *)appStartEnd
             [[SentryAppStartMeasurement alloc] initWithType:appStartType
                                                 isPreWarmed:isPreWarmed
                                           appStartTimestamp:appStartTimestamp
+                                    appStartSystemTimestamp:sysctl.processStartSystemTimestamp
```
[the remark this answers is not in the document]
armcknight: Yes you're right... I changed things a few times around these. Thanks for suggesting the correct name.

Sources/Sentry/SentryTimeToDisplayTracker.m
armcknight: I don't understand this feedback, if TTFD is enabled, this branch won't be taken because of the check on line 134 for `if (!_waitForFullDisplay)`, right?

Sources/Sentry/SentryTracer.m
armcknight: We don't use the NSDate-based timestamps for any alignment of transaction/span data and profile data. Also, those two timelines are unrelated, as we've discussed previously in https://github.com/getsentry/sentry-cocoa/issues/2155

Sources/Sentry/Public/SentrySpanProtocol.h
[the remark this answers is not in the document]
armcknight: Oh, so actually, I could remove this one, while keeping the one in SentrySpan.h. 💥

Sources/Sentry/include/SentrySpan.h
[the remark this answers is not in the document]
philipphofmann: When you also remove setting the start system timestamp from building app start spans the code still compiles; see https://github.com/getsentry/sentry-cocoa/pull/3783.

Sources/Sentry/SentryTracer.m
philipphofmann: Just wanted to double-check 👍

Sources/Sentry/SentrySDK.m
philipphofmann: I am sorry I missed these updates. My bad 👍

Sources/Sentry/SentryTimeToDisplayTracker.m
philipphofmann: > if TTFD is enabled, this branch won't be taken because of the check on line 134 for if (!_waitForFullDisplay), right?

Yes, that's true. Sorry, I had to finish my review yesterday quickly, cause I needed to watch my son. 

> When the user doesn't call TTFD the tracer finishes after some timeout.

This still applies. There is no guarantee that the user calls the manual TTFD API. When they don't call the manual TTFD API, the tracer times out and the `stopAndDiscardLaunchProfileTracer` after `[self.fullDisplaySpan finish];` will never be called. I think calling `stopAndDiscardLaunchProfileTracer` in the `finishCallback` should be the better approach.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -152,16 +152,21 @@
 }
 
 void
-stopLaunchProfile(SentryHub *_Nullable hub)
+stopAndTransmitLaunchProfile(SentryHub *hub)
```
philipphofmann, line 155: Way cleaner, thanks a lot 🥇

Sources/Sentry/SentryAppStartMeasurement.m
```diff
@@ -13,7 +13,7 @@ @implementation SentryAppStartMeasurement
     BOOL _isPreWarmed;
     NSTimeInterval _duration;
     NSDate *_appStartTimestamp;
-    uint64_t _appStartSystemTimestamp;
+    uint64_t _runtimeInitSystemTimestamp;
```
philipphofmann, line 16: Thanks 💯

philipphofmann (commented):
Thanks for including most of the feedback 😃

Sources/Sentry/include/SentrySpan.h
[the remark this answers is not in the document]
armcknight: Well how about that 😅

Sources/Sentry/SentryTimeToDisplayTracker.m
armcknight: Is the TTFD timeout managed via `SentryTracer.deadlineTimerFired`? If so then that will stop the profiler as well, unless some early return will be taken in `canBeFinished` or `finishInternal` before it gets a chance to call `captureTransactionWithProfile` here: https://github.com/getsentry/sentry-cocoa/blob/0addc419e43e1ded284a76db8c7d2a64840b939e/Sources/Sentry/SentryTracer.m#L621
This would be a good test case.

armcknight: I verified that the profiler is still stopped in this case. Slightly before its own 30 second timeout. I think this will be simpler than trying to do it in a closure, which brings along the added complexity of weakself and cycles.

philipphofmann (approved):
Thanks, @armcknight. LGTM

PR #3737, opened 2024-03-14
meta: accept pre-commit clang-format fixups

armcknight:
This keeps happening on my local branches even though I didn't modify these sources, preventing me from forming commits.

#skip-changelog


armcknight:
OK, must've been my local tooling was outdated.

PR #3739, opened 2024-03-14
meta: remove brew lockfile from git tracking

armcknight:
after updating my local tooling to fix the issue I was experiencing in #3737 

#skip-changelog


brustolin:
Can we just remove this two lock files from git?

philipphofmann (approved):
LGTM

armcknight:
> Can we just remove this two lock files from git?

Honestly yeah, at this point I don't see why not. Let's try it and see what happens.

armcknight:
@philipphofmann @brustolin please re-review the removal of the lockfiles here when you get a chance 🙏🏻

brustolin (approved):
No problem here.

Please just add this files to git ignore.

philipphofmann:
I'm unsure if we should remove the lock files. They ensure we all use the same version of dependencies locally and in CI.

brustolin:
> I'm unsure if we should remove the lock files. They ensure we all use the same version of dependencies locally and in CI.

It does not work. Every time I run `make init`, the lock files are updated.

armcknight:
Makes sense @philipphofmann , let's give that a shot

PR #3741, opened 2024-03-14
ref: Swift NSLock return value

philipphofmann:
Use generics to allow a return value for NSLock synchronized.

#skip-changelog


brustolin (approved):

brustolin:
I dont know why the tests are failing. I tried the code and all looks good.
Perhaps older versions of Xcode, with older Swift versions, don't support this.

PR #3745, opened 2024-03-14
ref: Expose DispatchQueueWrapper to Swift

philipphofmann:
I also changed the usage of SentryDispatchQueueWrapper to TestSentryDispatchQueueWrapper in the test if possible.

This is required for DDM to send metrics on a background thread and for scheduling a repeating timer on the main thread.

#skip-changelog


brustolin (approved):
LGTM

PR #3754, opened 2024-03-15
fix: Avoid duplicating loadView / viewDidLoad events

### LINKED ISSUE #3753: Re-entrant loadView / viewDidLoad
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.21.0

### Steps to Reproduce

1. Setup Sentry with `PerformanceTracking` and a profilingRate of 0.5
2. Launch the app

50% of the times we launch our app, Sentry's swizzled `loadView` method invokes the `.view` property of the relevant ViewController.

This ends up triggering two `viewDidLoad` invocations in our ViewController, and is causing duplication of UI objects (+ performance degradation).

### Expected Result

No `loadView` / `viewDidLoad` duplication

### Actual Result

Our main viewController is getting x2 `viewDidLoad` invocations. Please review the screenshots below:

<br><br>

Of course, PR will be up shortly. Thank you =)

jleandroperez:
## :scroll: Description
When performance tracking is enabled, `loadView`'s swizzled implementation may result in a duplicate chain of events, that result in multiple `viewDidLoad` invocations, in our main app's VC/

Thanks in advance!!

Fixes #3753

I'm afraid I did not test this change. Sentry's `Package.swift` points towards a binary distribution, and because of time constraints, I've been unable to pull together a Package Description for local testing.

## :pencil: Checklist

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift/ViewControllers/ReplaceContentViewController.swift
+++ b/Samples/iOS-Swift/iOS-Swift/ViewControllers/ReplaceContentViewController.swift
@@ -1,0 +1,99 @@
+//
+//  ReplaceContentViewController.swift
+//  iOS-Swift
+//
+//  Created by BJ Homer on 3/26/24.
+//  Copyright © 2024 Sentry. All rights reserved.
+//
+
+import UIKit
+
+class ReplaceContentViewController: UIViewController {
+
+    @IBOutlet var containerView: UIView!
+    private var currentChild: UIViewController!
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        // Add the initial child controller
+        let childVC = LoadCountReportingViewController(color: UIColor.systemBlue)
+        self.addChild(childVC)
+        childVC.view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        childVC.view.frame = containerView.bounds
+        containerView.addSubview(childVC.view)
+        childVC.didMove(toParent: self)
+
+        currentChild = childVC
+    }
+    
+    @IBAction func tappedReplaceContent(_ sender: Any) {
+        swapChild()
+    }
+
+    private func swapChild() {
+        let oldChild = currentChild!
+        let newChild = LoadCountReportingViewController(color: .systemPurple)
+
+        self.addChild(newChild)
+        oldChild.willMove(toParent: nil)
+
+        self.transition(from: oldChild,
+                        to: newChild,
+                        duration: 0.15,
+                        options: .transitionCrossDissolve,
+                        animations: {
+            newChild.view.frame = self.containerView.bounds
+        },
+                        completion: { (_) in
+            self.currentChild = newChild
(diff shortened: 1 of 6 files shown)


philipphofmann (commented):
Thanks a lot for this PR, @jleandroperez. The change makes sense to me, but I think we should add some tests for this. @brustolin, do you think we can add a UI test to `TopViewControllerTests` or somewhere to test this edge case?

brustolin:
>Thanks a lot for this PR, @jleandroperez. The change makes sense to me, but I think we should add some tests for this. @brustolin, do you think we can add a UI test to TopViewControllerTests or somewhere to test this edge case?

The code does make sense, but I can't reproduce.
With mock we probably can write a test.

jleandroperez:
Hi there @brustolin  + @philipphofmann!!

I'm afraid I spent some time trying to replicate the issue in a PoC, and it's not straightforward. 
The gist of what's happening / our setup would be:

1. Sentry is initialized in `willFinishLaunching` AppDelegate's callback
2. A placeholder ViewController is presented (as a child VC)
3. When the scene is ready, we transition into a secondary VC
   - Old API `transitionFromViewController:toViewController:duration:animations:completion` 
4. This `transition` API triggers the loadView / viewDidLoad sequence

Problem is that we are seeing this `viewDidLoad` callback sequence, coming from the `relevantViewControllerFromWindow:` API itself. 

Please note that L178 will trigger a loadView / viewDidLoad, right there:

image

After releasing LLDB, we end up breaking, again, in `viewDidLoad`. But notice this time the stack trace is different:

image


**Again, thank you both** for taking a look!!

bjhomer:
It seems like some of the checks are failing, but they seem to be authentication errors unrelated to this patch.

armcknight:
@bjhomer Yes, unfortunately this happens with PRs from forks. None of the credentials in our GHA settings can be accessed :/  I pushed the branch to our repo and opened a PR to run the tests here: https://github.com/getsentry/sentry-cocoa/pull/3797

Samples/iOS-Swift/iOS-Swift/ViewControllers/ReplaceContentViewController.swift
```diff
@@ -0,0 +1,99 @@
+//
+//  ReplaceContentViewController.swift
+//  iOS-Swift
+//
+//  Created by BJ Homer on 3/26/24.
+//  Copyright © 2024 Sentry. All rights reserved.
+//
+
```
philipphofmann, line 8: `l`: We always remove the copyright in our repo.
```suggestion
```

philipphofmann (commented):
Thanks a lot for the test @bjhomer🥇 and @jleandroperez, for the fix. Please fix one minor issue and add an entry to the Changelog. Then we can merge this PR 😃.

philipphofmann (approved):
The changelog still needs some fixes, but I will do that in a follow-up PR, so we can merge this. LGTM. Thanks again 😃

PR #3760, opened 2024-03-19
fix: XCFramework Path

### LINKED ISSUE #3759: Checksums for 8.22.0 binaries seem to be incorrect
### Description

Screenshot 2024-03-18 at 17 30 52

brustolin:
The path to the XCFramework in the package.json was incorrect before releasing the new version, so the bump tool did not properly update it.

Closes #3759

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Fixes 
+
+- Package.json xcframework path ()
+
 ## 8.22.0
+**Warning:** this version is not working with SPM
 
 ### Improvements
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Fixes 
+
+- Package.json xcframework path (#3760)
```
philipphofmann, line 7: `m`: I think we should communicate which error this solves.
```suggestion
- Checksum error when resolving the SDK via SPM (#3760)
```

philipphofmann (approved):
LGTM

Issue #3763, opened 2024-03-19
SDK start crashes when including via SPM or Carthage on 8.22.1 


philipphofmann:
### Platform

iOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

8.22.1

### Did it work on previous versions?

Yes, 8.21.0

### Steps to Reproduce

1. Create an empty sample iOS project
2. Add sentry-cocoa 8.22.1 as a SPM dependency
3. Start the SDK 

```swift
SentrySDK.start { options in
    options.dsn = "DSN"
    options.debug = true
}
```

### Expected Result

No crash.

### Actual Result

App crashes with
```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '+[NSDate sentry_fromIso8601String:]: unrecognized selector sent to class 0x1f16fc418'
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000180491128 __exceptionPreprocess + 172
	1   libobjc.A.dylib                     0x000000018008412c objc_exception_throw + 56
	2   CoreFoundation                      0x00000001804a5dd4 _CFAutoreleasePoolAddObject + 0
	3   CoreFoundation                      0x0000000180495278 ___forwarding___ + 1280
	4   CoreFoundation                      0x000000018049759c _CF_forwarding_prep_0 + 92
	5   SPM                                 0x0000000102745f98 -[SentryAppState initWithJSONObject:] + 436
	6   SPM                                 0x0000000102730c44 +[SentrySerialization appStateWithData:] + 168
	7   SPM                                 0x0000000102799454 -[SentryFileManager readAppStateFrom:] + 108
	8   SPM                                 0x000000010279939c -[SentryFileManager readPreviousAppState] + 72
	9   SPM                                 0x0000000102760f78 -[SentryAppStateManager loadPreviousAppState] + 32
	10  SPM                                 0x000000010273bf34 -[SentryWatchdogTerminationLogic isWatchdogTermination] + 84
	11  SPM                                 0x0000000102786da4 -[SentrySessionCrashedHandler endCurrentSessionAsCrashedWhenCrashOrOOM] + 88
	12  SPM                                 0x0000000102770590 __44-[SentryCrashIntegration startCrashHandler:]_block_invoke + 392
	13  libdispatch.dylib                   0x0000000103a5d93c _dispatch_client_callout + 16
	14  libdispatch.dylib                   0x0000000103a5f3dc _dispatch_once_callout + 84
[...]

PR #3764, opened 2024-03-19
fix: Remove categories functions that are striped of static library

brustolin:
Part of a fixes for #3763 

_#skip-changelog_


philipphofmann (approved):
LGTM, but we need to merge https://github.com/getsentry/sentry-cocoa/pull/3765 before merging this PR.

PR #3766, opened 2024-03-19
chore: Move XCFrameworks to root after building

philipphofmann:
Running `make build-xcframework-sample` relies on having the Sentry.xcframework.zip file in the repository root folder see
https://github.com/getsentry/sentry-cocoa/blob/6bc31df2a007b41dcdac3c456dc9a97d6a2c49bb/scripts/create-carthage-json.sh#L4

In CI, we archive the XCFramework zips here 
https://github.com/getsentry/sentry-cocoa/blob/6bc31df2a007b41dcdac3c456dc9a97d6a2c49bb/.github/workflows/build.yml#L132

and then downloading the artifacts puts them in the repository root folder before running `make build-xcframework-sample`

https://github.com/getsentry/sentry-cocoa/blob/6bc31df2a007b41dcdac3c456dc9a97d6a2c49bb/.github/workflows/build.yml#L145-L153

Running `make build-xcframework` locally doesn't put the XCFrameworks zip files into the repository root folder, which breaks running  `make build-xcframework-sample` locally. This PR fixes this problem.

#skip-changelog


brustolin (commented):
I believe you need to update 
- `update-package-sha.sh` script.
-  Lines 106 and 116 of Build.yml
- release.yml lines 34-36

IMO, it would be better to update the sample to point to the right place.

philipphofmann:
> IMO, it would be better to update the sample to point to the right place.

The problem is that GH actions download the artifact into the root folder of the repository. I could also then move the downloaded artifacts to `Carthage/` on GH actions, but that also feels kind of weird.

Makefile
```diff
@@ -65,6 +65,9 @@ build-xcframework:
 	ditto -c -k -X --rsrc --keepParent Carthage/Sentry.xcframework Carthage/Sentry.xcframework.zip
 	ditto -c -k -X --rsrc --keepParent Carthage/Sentry-Dynamic.xcframework Carthage/Sentry-Dynamic.xcframework.zip
 	ditto -c -k -X --rsrc --keepParent Carthage/SentrySwiftUI.xcframework Carthage/SentrySwiftUI.xcframework.zip
+	mv Carthage/Sentry.xcframework.zip Sentry.xcframework.zip
+	mv Carthage/Sentry-Dynamic.xcframework.zip  Sentry-Dynamic.xcframework.zip
+	mv Carthage/SentrySwiftUI.xcframework.zip SentrySwiftUI.xcframework.zip
```
armcknight, line 70: The `mv` destination arguments could just be the destination args in the `ditto` commands

armcknight: Couldn't modify unchanged lines in a suggestion comment, so i made a PR: https://github.com/getsentry/sentry-cocoa/pull/3807

philipphofmann:
@brustolin, please have a look at https://github.com/getsentry/sentry-cocoa/pull/3766#issuecomment-2007125832. I either want to merge or close this PR.

brustolin:
> @brustolin, please have a look at [#3766 (comment)](https://github.com/getsentry/sentry-cocoa/pull/3766#issuecomment-2007125832). I either want to merge or close this PR.

I dont mind, we can change it, but you still need to address the 3 points in [this comment](https://github.com/getsentry/sentry-cocoa/pull/3766#pullrequestreview-1945990991)

brustolin (approved):

PR #3771, opened 2024-03-20
fix: Hybrid SDKs without SentryPrivate

brustolin:
## :scroll: Description

We need to update some references in order for HybridSDKs to work properly


philipphofmann (approved):
LGTM

Issue #3631, opened 2024-02-13
[cocoa] Add support for Sentry Metrics


kahest:
### Description
Metrics! Sentry will support Metrics finally! First we need all of our SDKs to support Metrics, including iOS

#### Requirements:

- [x] Metrics API
- [x] Tag normalization
- [x] Metrics Summary

#### Document(s)
[Develop Specs](https://develop.sentry.dev/delightful-developer-metrics/)
[SDK Develop Specs](https://develop.sentry.dev/delightful-developer-metrics/sending-metrics-sdk/)
[Product Specs](https://develop.sentry.dev/sdk/metrics/)

```[tasklist]
### Tasks
- [x] https://github.com/getsentry/sentry-cocoa/pull/3676
- [x] https://github.com/getsentry/sentry-cocoa/pull/3744
- [ ] https://github.com/getsentry/sentry-cocoa/pull/3762
- [ ] https://github.com/getsentry/sentry-cocoa/pull/3782
- [ ] https://github.com/getsentry/sentry-cocoa/pull/3773
- [ ] https://github.com/getsentry/sentry-cocoa/pull/3791
- [ ] https://github.com/getsentry/sentry-cocoa/pull/3792
```

PR #3772, opened 2024-03-20
chore(metrics): Add Set, Gauge, Distribution metric classes

philipphofmann:
Add classes for Set, Gauge, and Distribution.

This is part of adding DDM to the SDK https://github.com/getsentry/sentry-cocoa/issues/3631 and this PR is based on https://github.com/getsentry/sentry-cocoa/pull/3762.

#skip-changelog


brustolin (commented):
I believe you need to merge main into this PR before I could review it.

PR #3773, opened 2024-03-20
chore(metrics): Add LocaleMetricsAggregator

philipphofmann:
Add LocalMetricsAggregator, which is the foundation for correlating metrics to spans.

This is part of adding DDM to the SDK https://github.com/getsentry/sentry-cocoa/issues/3631, and this PR is based on https://github.com/getsentry/sentry-cocoa/pull/3772.

I will call the `LocalMetricsAggregator` in an extra PR, so this PR doesn't get too big.

#skip-changelog


.swiftlint.yml
```diff
@@ -35,7 +35,6 @@ only_rules:
   - force_try
   - force_unwrapping
   - function_body_length
-  - function_parameter_count
```
philipphofmann, line 38: I disabled this rule cause it only allows 5 parameters per method, which doesn't make sense IMO.

Sources/Swift/Metrics/LocalMetricsAggregator.swift
```diff
@@ -0,0 +1,72 @@
...
+                metric = Metric(min: newMin, max: newMax, count: count, sum: newSum, tags: tags)
+            }
+            
+            bucket[tagsKey] = metric
+            metricBuckets[exportKey] = bucket
+        }
+    }
+    
+    func serialize() -> [String: [[String: Any]]] {
+        var returnValue: [String: [[String: Any]]] = [:]
```
brustolin, line 45: m: Since you're declaring `typealias`, I think we should reuse it wherever we expect that type of data. The biggest benefit of typealias is to easily update types, and we don't have that here. For this, you need to make ExportKey not private.

Otherwise, using `typealias` in this context with a base type is not adding any value; on the contrary, it requires a second look to understand what's going on.

```suggestion
    func serialize() -> [ExportKey: [[String: Any]]] {
        var returnValue: [ExportKey: [[String: Any]]] = [:]
```

brustolin (approved):
Looks good.

Only a few suggestions.

PR #3777, opened 2024-03-20
test: move the slow load test out of the SDK and into the sample app

armcknight:
There was some concern that having this in the SDK and using it in any examples would make it appear that the SDK has some issue with a slow method that runs during pre-main. Moving it to the sample app makes it more realistic that it is simulating something that would happen in a real world app.

The test failed locally until I disabled compiler optimizations for the test configs, because it wasn't seeing the function name in the frame; I guess that info had been stripped out of the compiled binary.


brustolin (approved):
LGTM

philipphofmann (approved):
That totally makes sense. Thanks 💯

PR #3791, opened 2024-03-26
feat: Add Metrics API

philipphofmann:
## :scroll: Description

Add beta implementation of MetricsAPI.

## :bulb: Motivation and Context

https://github.com/getsentry/sentry-cocoa/issues/3631

Docs PR https://github.com/getsentry/sentry-docs/pull/9554.

## :green_heart: How did you test it?
Unit tests and simulator.

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I reviewed the submitted code.
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #3792, opened 2024-03-26
feat(metrics): Change adding set to string

philipphofmann:
The MetricsAPI for adding a set should accept a string, which gets converted to a CRC32 hash instead of an integer.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/3791.

#skip-changelog


Sources/Swift/Metrics/SentryMetricsAPI.swift
```diff
@@ -1,5 +1,6 @@
 @_implementationOnly import _SentryPrivate
 import Foundation
+import zlib
```
brustolin, line 3: h: I believe you forgot this.

Sources/Swift/Metrics/SetMetric.swift
```diff
@@ -2,18 +2,21 @@ import Foundation
...
     }
 
-    init(first: Int32, key: String, unit: MeasurementUnit, tags: [String: String]) {
+    init(first: UInt, key: String, unit: MeasurementUnit, tags: [String: String]) {
         set = [first]
         super.init(type: .set, key: key, unit: unit, tags: tags)
     }
 
+    // This doesn't work with the full range of UInt.
+    // We still need to fix this.
```
brustolin, line 16: l: If we really need the function signature to expect a Double, this is a limitation we will need to accept.

But now that I'm understanding this a little better, having `Metric.Add` always expect a double seem a little bit wrong.

brustolin (approved):

Sources/Swift/Metrics/SetMetric.swift
philipphofmann: > But now that I'm understanding this a little better, having Metric.Add always expect a double seem a little bit wrong.

Yes, it is. It can be a double or an int.

I already have a PR for that https://github.com/getsentry/sentry-cocoa/pull/3795.

Sources/Swift/Metrics/SentryMetricsAPI.swift
philipphofmann: Oh, yeah thanks.

PR #3793, opened 2024-03-26
feat: Release framework with no link to UIKit

### LINKED ISSUE #3768: Ship an XCFramework without UIKit linked
### Problem Statement

At Dropbox, we currently build Sentry both with and without UIKit.  We require that Sentry in our File Provider extension not import UIKit because the UIKit symbols take up too much space in memory with FP's super low memory limitations.

We'd love to simplify our build by using your new XCFramework distributions instead, but we need both the ones you're releasing now and also ones built without the UIKit link in order to switch over.

### Solution Brainstorm

Build Sentry without a link to UIKit and provide those XCFramework(s) as well in your releases page.

Maybe

brustolin:
## :scroll: Description

Creates a framework with no link to UIKit to distribute as a GH release asset.

closes #3768


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Don't run onCrashedLastSession for nil Events (#3785)
 
+### Features
+
+- Release framework with no link to UIKit (#3793)
```
philipphofmann, line 11: `l`:
```suggestion
- Release framework without to UIKit/AppKit (#3793)
```

scripts/build-xcframework.sh
```diff
@@ -66,3 +67,5 @@ generate_xcframework "Sentry" "-Dynamic"
 generate_xcframework "Sentry" "" staticlib
 
 generate_xcframework "SentrySwiftUI"
+
+generate_xcframework "Sentry" "-NoUIKit" mh_dylib Release_without_UIKit
```
philipphofmann, line 71: `m`: I think that's a bit cleaner
```suggestion
generate_xcframework "Sentry" "-WihoutUIKitOrAppKit" mh_dylib Release_without_UIKitOrAppKit
```

philipphofmann (commented):
We don't have UIKit on macOS, but instead, we have AppKit. Our SDK includes headers to Cocoa.h, but we only use NSApplication notification names. Therefore, we should be able to replace all Cocoa.h imports with `#import <AppKit/AppKit.h>`. So, let's reflect that in the framework name, as pointed out in the comments.

scripts/build-xcframework.sh
brustolin: The forth parameter is the configuration name. It cannot be changed

philipphofmann: I thought that's how the XCFramework is called. I only care about the name of the XCFramework. How is that one called then?

brustolin: Second parameter.
It will be `Sentry-WihoutUIKitOrAppKit`

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ more about how to use the Metrics API.
 - Don't run onCrashedLastSession for nil Events (#3785)
 - Redistributable static libraries should never be built with module debugging enabled (#3800)
 
+### Features
```
philipphofmann, line 16: There is an already existing features header in the current unreleased section. I believe this needs to move up a couple of lines.

philipphofmann (approved):
LGTM

PR #3795, opened 2024-03-26
ref: Avoid casting for set metric

philipphofmann:
For the set metric, the SDK casts the value back and forth from UInt to Double, which doesn't work correctly with large UInt numbers as Double can't represent them accurately. This is fixed now by only using UInt for set metrics.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/3792.

#skip-changelog


brustolin (approved):
I like this approach better!!

PR #3808, opened 2024-03-29
chore(profiling): place legacy compilation gates around profiling code that will need to be changed/removed

armcknight:
To kick off work for #3555, introduce two new preprocessor definitions to help transition the profiling implementation from the current form (henceforth known as "legacy") to continuous mode.

Gate declarations and implementation that will change or be removed with the legacy flag. This is not necessarily exhaustive, but serves as a guidepost for further work.

#skip-changelog

### DIFF
--- a/Sources/Sentry/PrivateSentrySDKOnly.mm
+++ b/Sources/Sentry/PrivateSentrySDKOnly.mm
@@ -124,7 +124,7 @@ static BOOL _framesTrackingMeasurementHybridSDKMode = NO;
     return [SentryDependencyContainer.sharedInstance.extraContextProvider getExtraContext];
 }
 
-#if SENTRY_TARGET_PROFILING_SUPPORTED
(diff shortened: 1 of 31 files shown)


philipphofmann (commented):
I'm not a fan of using compiler flags to switch between the legacy and new profiling modes because you need to recompile the whole code. I prefer a feature flag because that's easier for testing and customers to try. I know I once said fine by be in Notion, but I changed my mind when looking at the PR now and would vote for a feature flag.

philipphofmann (approved):
I don't want https://github.com/getsentry/sentry-cocoa/pull/3808#pullrequestreview-1973523671 to block this PR. If you would like to stick to compiler flags, which I don't think is a great idea, do so @armcknight.

armcknight:
OK, we're going to go with the feature flag approach in #3831

PR #3831, opened 2024-04-05
feat(profiling): add private option to feature flag continuous profiling mode

armcknight:
A private (for now) feature flag that will be used to switch between implementations of legacy and continuous profiling.

For #3555 

#skip-changelog


philipphofmann (approved):
LGTM :shipit:

PR #3833, opened 2024-04-05
ref(profiling): prepare for new public API

armcknight:
- rename SentryProfiler.h to SentryProfiler+Private.h in case we need a public-facing SentryProfiler.h without all the interface that was already in that file
- there was already a SentryProfiler+Private.h; split up what was in there between the new one (nee SentryProfiler.h) and SentryProfiler+Test.h, depending on where the declarations are actually needed
- there were also two SentryProfiler+Test.h files: one in the SDK, and one in the test utils lib; remove the one in the SDK, and move its declarations to the one in the test lib along with the appropriate ones moved from the old SentryProfiler+Private.h that was deleted to make way for renaming SentryProfiler.h to SentryProfiler+Private.h

#skip-changelog

for #3555


philipphofmann (approved):
LGTM 🚀

PR #3834, opened 2024-04-05
feat(profiling): add API to SentrySDK to start/stop a continuous profiling session

armcknight:
And thread it through to a stub for new implementation for the continuous profiler.

Add a check for where the legacy profiler is started from SentryTracer for the new feature flag. 

# TODO

~- [ ] Haven't decided what to do yet with the launch profiler; I think it will also need the feature flag at the spots that automatically stop the tracer managing it, so it is left running throughout the app session or until SentrySDK.stopProfiler is called.~
~- [ ] just learned (April 5) that we will automatically start profiling from SentrySDK.startWithOptions (if launch profiling isn't enabled) which will respect profiles sample rate, so this needs to be added here or in a subsequent PR~
I am just going to leave this PR as the stub for the new continuous profiler and leave it to a future PR to deliver the implementation of automatic/manual start/stop.

#skip-changelog

for #3555


Sources/Sentry/SentryTracer.m
```diff
@@ -188,8 +189,9 @@ - (instancetype)initWithTransactionContext:(SentryTransactionContext *)transacti
 #endif // SENTRY_HAS_UIKIT
 
 #if SENTRY_TARGET_PROFILING_SUPPORTED
-    if (_configuration.profilesSamplerDecision.decision == kSentrySampleDecisionYes
-        || isTracingAppLaunch) {
+    if (!hub.client.options.enableContinuousProfiling
```
philipphofmann, line 192: `m`: A test or updating an existing test for this new check would be great.

philipphofmann (approved):
LGTM

Sources/Sentry/include/SentrySDK+Private.h
```diff
@@ -42,6 +43,20 @@ SentrySDK ()
  */
 + (void)captureEnvelope:(SentryEnvelope *)envelope;
 
+#if SENTRY_TARGET_PROFILING_SUPPORTED
```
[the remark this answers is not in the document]
armcknight: Yep!

Sources/Sentry/SentryTracer.m
armcknight: This is true. I think this logical change would be more appropriate in a later branch where I'm actually changing implementation, which will include tests, so I'll move it there.

Issue #3805, opened 2024-03-28
Metrics Rate Limits


kahest:
see https://github.com/getsentry/team-sdks/issues/77

PR #3837, opened 2024-04-08
ref: CurrentDate as dependency for rate limits

philipphofmann:
Replace access to the DependencyContainer in the RateLimits code by passing the current date as a dependency to reduce global state and the need to call clearTestState.

This came up when tackling https://github.com/getsentry/sentry-cocoa/issues/3805.

#skip-changelog


brustolin (approved):
As far or architecture goes, this looks good, but I hate this overuses of DI over initialization. Specially for basic things like Date, DispatchQueue, Random and more.

I believe we should talk about this architecture.

PR #3845, opened 2024-04-10
meta: bump deployment targets to match earliest SDK versions available in CI

armcknight:
These are the earliest versions available to test with in GitHub actions: https://github.com/actions/runner-images/blob/91e422a825d3941ea357a13a91b13d4f4fdc326d/images/macos/macos-11-Readme.md?plain=1#L267-L296


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changes
+
+- Update deployment targets for macOS/iOS/tvOS/watchOS (#3845)
```
bruno-garcia, line 7: Best to be specific here to what version we changed to

```suggestion
- Update deployment targets for: macOS (10.13 -> 10.15), iOS/tvOS (11 -> 13), watchOS (4 -> 6) (#3845)
```

brustolin:
I agree, I believe we should be even more aggressive with this, but our approach so far it is to support the minimum os version Xcode also supports, and right now this is iOS 12.

philipphofmann (changes requested):
Xcode 15 allows you to target macOS 10.13, iOS 12, tvOS 12, and watchOS 4. We must support these versions. This is also in our [SDK Philosophy and Design Principles](https://develop.sentry.dev/sdk/philosophy/#compatibility-is-king).

armcknight:
> Xcode 15 allows you to target macOS 10.13, iOS 12, tvOS 12, and watchOS 4. We must support these versions. This is also in our [SDK Philosophy and Design Principles](https://develop.sentry.dev/sdk/philosophy/#compatibility-is-king).

I will post a reply in the issue you opened: https://github.com/getsentry/sentry-cocoa/issues/3846

PR #3851, opened 2024-04-12
feat: Link errors with session replay

brustolin:
## :scroll: Description

Added replay id as error event tag and to baggage and trace headers.

Unit tests

## :pencil: Checklist

## :crystal_ball: Next steps

_#skip-changelog_


Tests/SentryTests/Transaction/SentryTraceStateTests.swift
```diff
@@ -131,27 +136,30 @@ class SentryTraceContextTests: XCTestCase {
...
-        XCTAssertEqual(baggage.sampleRate, fixture.sampleRate)
-        XCTAssertEqual(baggage.sampled, fixture.sampled)
+        expect(baggage.traceId) == fixture.traceId
+        expect(baggage.publicKey) == fixture.publicKey
+        expect(baggage.releaseName) == fixture.releaseName
+        expect(baggage.environment) == fixture.environment
+        expect(baggage.userSegment) == fixture.userSegment
+        expect(baggage.sampleRate) == fixture.sampleRate
+        expect(baggage.sampled) == fixture.sampled
+        expect(baggage.replayId) == fixture.replayId
```
philipphofmann, line 151: Thanks for converting these 👍

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -98,8 +99,13 @@ - (void)start:(UIView *)rootView fullSession:(BOOL)full
 
 - (void)startFullReplay
 {
-    _sessionStart = _lastScreenShot;
-    _isFullSession = YES;
+    @synchronized(self) {
```
philipphofmann, line 102: `m`: Why do we need to synchronize here? What problem does it solve? Does it make sense to add a test for this?

philipphofmann (commented):
Almost LGTM.

Sources/Sentry/SentrySessionReplay.m
brustolin: To avoid a race condition where two errors are captured very close to each other.
Honestly, I don't  trust tests for race conditions; they require too much 'magic' that makes them so flaky

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -112,7 +118,12 @@ - (void)stop
 
 - (void)captureReplayForEvent:(SentryEvent *)event;
 {
-    if (_isFullSession || !_isRunning) {
+    if (!_isRunning) {
+        return;
+    }
+
+    if (_isFullSession) {
```
romtsn, line 125: FYI: I reordered this with the if-check below (line 130), I think JS sdk only tags error/crashed events with replay_id but not messages

romtsn (approved):
LGTM!

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -98,8 +99,13 @@ - (void)start:(UIView *)rootView fullSession:(BOOL)full
...
 {
-    _sessionStart = _lastScreenShot;
-    _isFullSession = YES;
+    @synchronized(self) {
+        _sessionStart = _lastScreenShot;
+        _isFullSession = YES;
+        [SentrySDK.currentHub configureScope:^(SentryScope *_Nonnull scope) {
+            scope.replayId = [self->_sessionReplayId sentryIdString];
+        }];
+    }
```
brustolin, line 108:
```suggestion
      _sessionStart = _lastScreenShot;
      _isFullSession = YES;
      [SentrySDK.currentHub configureScope:^(SentryScope *_Nonnull scope) {
          scope.replayId = [self->_sessionReplayId sentryIdString];
      }];
```

Sources/Sentry/SentrySessionReplay.m
bruno-garcia: synchronization doesn't seem needed here though. DOesn't take any argument, two concurrent calls won't do anything. Scope changing is already thread safe.
And reading of the fields are isnt' atomic anyway

Sources/Sentry/SentrySessionReplay.m
bruno-garcia: should we tag messages though? @billyvg 
We don't need to trigger sampling, but if the recording is already ongoing, I imagine it's helpful to link any events captured with it.

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -124,15 +135,25 @@ - (void)captureReplayForEvent:(SentryEvent *)event;
...
 
     [self createAndCapture:finalPath
                   duration:_replayOptions.errorReplayDuration
                  startedAt:replayStart];
+}
 
-    [self startFullReplay];
+- (void)setEventTag:(SentryEvent *)event
+{
+    NSMutableDictionary *tags = @{ @"replayId" : [_sessionReplayId sentryIdString] }.mutableCopy;
```
bruno-garcia, line 152: we should plan to remove this tag with proper field. Not working on `context['replay'].replayId`?

bruno-garcia (approved):
This is a step forward but we can have some follow ups from it

Sources/Sentry/SentrySessionReplay.m
cmanallen: Correct context name is `contexts['replay'].replay_id`.

Sources/Sentry/SentrySessionReplay.m
brustolin: We should fix the DSC context, and then this won't be required anymore; replay ID will be shared in every type of event after a session mode is started.

Sources/Sentry/SentrySessionReplay.m
bruno-garcia: oh context has `replay_id` and tag has `replayId` ?? 😢

Sources/Sentry/SentrySessionReplay.m
billyvg: I don't think we need to tag messages

PR #3854, opened 2024-04-12
feat(profiling): implement continuous profiling

armcknight:
Begin implementing the behavior of the continuous profiler in `+ [SentryContinousProfiler start]` and `+ [SentryContinuousProfiler stop]`.

Split the rest of the class methods in SentryProfiler into a new `SentryLegacyProfiler` class, such that all that remains in SentryProfiler is the instance-based state and internal reference to the SamplingProfiler. The logic in this file is unchanged from how it appeared in SentryProfiler with the exception that it uses the new init method that passes a SentryProfilerModeLegacy enum value as a parameter.

So, now we have completely separated the control API for running a profiler into either a legacy (`SentryLegacyProfiler`) set of commands or the new continuous (`SentryContinuousProfiler`) set of commands, from the _state_ of a profiler (`SentryProfiler`).

`SentryProfiler` gets a new enum `SentryProfilerMode` which is used to initialize it with either the legacy or continuous mode.

#skip-changelog

for #3555


Sources/Sentry/Profiling/SentryLegacyProfiler.mm
```diff
@@ -0,0 +1,92 @@
...
+
+    if (_unsafe_isRunning()) {
+        SENTRY_LOG_DEBUG(@"A legacy profiler is already running.");
+        trackProfilerForTracer(_unsafe_gLegacyProfiler, traceId);
+        // record a new metric sample for every concurrent span start
+        [_unsafe_gLegacyProfiler.metricProfiler recordMetrics];
+        return YES;
+    }
+
+    _unsafe_gLegacyProfiler = [[SentryProfiler alloc] initWithMode:SentryProfilerModeLegacy];
```
armcknight, line 44: This is the only line that actually changed when moving this code here from SentryProfiler.

Tests/SentryTests/SentryLegacyProfiler+Test.h
```diff
@@ -0,0 +1,35 @@
...
+#    import <Foundation/Foundation.h>
+
+@class SentryProfiler;
+
+@interface
+SentryLegacyProfiler ()
+
+#    if defined(TEST) || defined(TESTCI)
+
++ (SentryProfiler *_Nullable)getCurrentProfiler;
```
armcknight, line 15: This wasn't previously `_Nullable` but it should've been, since the underlying global var is nullable. This affecting just a couple callsites in test code, some of which now use `XCTUnwrap`.

Issue #3836, opened 2024-04-05
App crash due to a background call to `-[UIApplication applicationState]` in `-[SentryUIApplication init]`


diogot:
### Platform

tvOS

### Environment

Other

### Installed

Swift Package Manager

### Version

8.23.0

### Did it work on previous versions?

It works up to version 8.21.0

### Steps to Reproduce

Call `+ (SentryId *)captureError:(NSError *)error withScopeBlock:(void (^)(SentryScope *_Nonnull))block` in a background thread crashes the app due to `Main Thread Checker: UI API called on a background thread: -[UIApplication applicationState]`.

This is happening in our unit tests, we didn't released to production.
We are using Xcode 15.3.

### Expected Result

App did not crash.

### Actual Result

App crash with this stack trace:

Main Thread Checker: UI API called on a background thread: -[UIApplication applicationState]
PID: 16779, TID: 3695907, Thread name: (none), Queue name: tv.standard.watch-NetworkService, QoS: 25
Backtrace:
4   Services                            0x000000010842a4b4 -[SentryUIApplication init] + 244
5   Services                            0x000000010841eacc -[SentryDependencyContainer application] + 68
6   Services                            0x00000001083d32b4 -[SentryCrashWrapper enrichScope:] + 1096
7   Services                            0x00000001083eafc8 -[SentryHub initWithClient:andScope:] + 736
8   Services                            0x00000001084107e0 +[SentrySDK currentHub] + 60
9   Services                            0x000000010841183c +[SentrySDK captureError:withScopeBlock:] + 80

philipphofmann:
Thanks for reporting this, @diogot. This should be an easy fix.

There is no guarantee that `SentryUIApplication init` is called on the main thread. It usually gets called on the main thread like this:
```
-[SentryUIApplication init] at SentryUIApplication.m:31
-[SentryDependencyContainer application] at SentryDependencyContainer.m:261
-[SentryCrashWrapper enrichScope:] at SentryCrashWrapper.m:187
-[SentryHub initWithClient:andScope:] at SentryHub.m:80
+[SentrySDK startWithOptions:] at SentrySDK.m:209
+[SentrySDK startWithConfigureOptions:] at SentrySDK.m:249
```

But if for example the SDK gets started from a background thread or if somebody interacts with the hub without starting the SDK on a background thread, `SentryUIApplication init` might get called on a background thred.

mlch911:
I'm experiencing this crash as well. 
[...]

PR #3855, opened 2024-04-15
fix: crash when call SentryUIApplication in background thread

mlch911:
## :scroll: Description

Fix crash when call SentryUIApplication in background thread. #3836

## :pencil: Checklist


brustolin:
Hello @mlch911, thanks for the help. 
Since the problem occurs only during initialization, and, the responsibility to call every other SentryUIApplication function from the main thread lies with the caller, I believe we should only change the init function.

I need to check tho what is triggering the initialization. because just pushing it to a background thread may not be enough.

mlch911:
> Hello @mlch911, thanks for the help. Since the problem occurs only during initialization, and, the responsibility to call every other SentryUIApplication function from the main thread lies with the caller, I believe we should only change the init function.
> 
> I need to check tho what is triggering the initialization. because just pushing it to a background thread may not be enough.

This is not only occurs during initialization. Our app encounter this crash rarely when the network monitoring occurs at background thread just after we init sentry in main thread.

Here is the crash stack.
```
-[MaxWindow initWithFrame:] (MaxWindow.m:24)
-[AppDelegate window] (AppDelegate.m:184)
-[SentryUIApplication windows] (SentryUIApplication.m:84)
-[SentryCrashWrapper enrichScope:] (SentryCrashWrapper.m:187)
-[SentryHub initWithClient:andScope:] (SentryHub.m:80)
+[SentrySDK currentHub] (SentrySDK.m:76)
-[SentryRequestOperation initWithSession:request:completionHandler:]_block_invoke (SentryRequestOperation.m:37)
```

Sources/Sentry/SentryDispatchQueueWrapper.m
```diff
@@ -56,23 +56,41 @@ - (void)dispatchSyncOnMainQueue:(void (^)(void))block
...
+}
 
-        dispatch_time_t timeout_t
-            = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(timeout * NSEC_PER_SEC));
-        return dispatch_semaphore_wait(semaphore, timeout_t) == 0;
-    }
-    return YES;
+- (id)dispatchSyncOnMainQueueWithResult:(id (^)(void))block timeout:(NSTimeInterval)timeout
+{
+	if ([NSThread isMainThread]) {
```
balavor, line 75: nit: worth using `DispatchSpecificKey` for the main queue instead.

mlch911: Is there any benefit using `DispatchSpecificKey` instead of `[NSThread isMainThread]`?
I think the goal here is to ensure this is running on main thread. Setting a specificKey to main dispatch-queue and check the current dispatch-queue's specificKey may be excrescent.

brustolin (approved):

Issue #3856, opened 2024-04-15
Update to 8.21.0 causes a crash in my app in runtime


iwan-ua:
### Platform

macOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

8.21.0

### Did it work on previous versions?

8.20.0

### Steps to Reproduce

(I know these aren't really steps, please see actual result)
1. Update the library
2. Launch tests that previously worked
3. Crash in NETunnelProvider

### Expected Result

tests succeed as before the update

### Actual Result

After updating to 8.21.0 I noticed this warning on my CI, that wasn't there before

ld: tentative definition of '_context' with size 296 from '/Users/distiller/Library/Developer/Xcode/DerivedData/MyCompany-elgqgghuekrytrhiucwhimezasdq/Build/Products/Debug Dev/libMyCompanyTunnelProviderMac.a(netbios.c.o)' is being replaced by real definition of smaller size 208 from '/Users/distiller/Library/Developer/Xcode/DerivedData/MyCompany-elgqgghuekrytrhiucwhimezasdq/Build/Products/Debug Dev/Sentry.o'

My NETunnelProvider crashes, probably because it uses sentry code instead of my own code

8.21.0 had this change, so I think it might be responsible, but I'm not sure
https://github.com/getsentry/sentry-cocoa/pull/3690

iwan-ua:
I think [this part](https://github.com/getsentry/sentry-cocoa/blame/d9308fd7531e4b4a45e32ecb21406736ad9fd632/Sources/Sentry/Profiling/SentryLaunchProfiling.m#L95) conflicts with this part of our code:

struct {
    ....
} context;

We'll rename it and see if it fixes it. But I think for third-party libs it's better to give unique names to its global variables

brustolin:
>I think [this part](https://github.com/getsentry/sentry-cocoa/blame/d9308fd7531e4b4a45e32ecb21406736ad9fd632/Sources/Sentry/Profiling/SentryLaunchProfiling.m#L95) conflicts with this part of our code:

Makes sense. context is a very generic symbol. We probably should rethink this. 
What do you think @armcknight ?

iwan-ua:
Fixed by changing our
`struct {
....
} context;`
into
`static struct {
....
} context;`

brustolin:
Thanks for the feedback @iwan-ua.

I believe we should change the name internally to avoid future naming conflict.

PR #3862, opened 2024-04-18
fix(profiling): prefix some functions with "sentry_"

armcknight:
We should've been prefixing any functions in the global namespace to avoid potential collisions with other apps' functions. Fixes the immediate problem reported in https://github.com/getsentry/sentry-cocoa/issues/3856, but we may need to do more here with all other functions. These generic functions names present the most likely possibility for collisions, though, so this should practically fix most of the issue.

Closes https://github.com/getsentry/sentry-cocoa/issues/3856


brustolin (commented):
I believe prefixing is good enough, but CI is not happy.

indragiek (approved):

PR #3863, opened 2024-04-19
test: use correct a11y property and skip if not correct OS version

armcknight:
I noticed this fail locally due to a XCUIElement query timeout, but it worked when I changed `crashTheApp` from being the accessibilityIdentifier to the accessibilityLabel for the button.

I also set it up to report the test case as skipped if the OS version req is not met.

I've noticed this test flake in the past but not sure this will fix that.

#skip-changelog


Samples/iOS-Swift/iOS-Swift/Base.lproj/Main.storyboard
```diff
@@ -750,15 +750,15 @@
                                                     </button>
                                                     <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="system" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="NNd-Ec-zXw">
                                                         <rect key="frame" x="0.0" y="56" width="160" height="28"/>
-                                                        <accessibility key="accessibilityConfiguration" identifier="crashTheApp"/>
+                                                        <accessibility key="accessibilityConfiguration" label="crashTheApp"/>
```
armcknight, line 753: the only relevant change in the file.

indragiek (approved):

PR #3865, opened 2024-04-22
ref: Remove unused property in SentryUIApplication

philipphofmann:
#skip-changelog


brustolin (approved):
Looks good.

PS:  Personally I would recommend to refrain from doing this type of refactoring.
- SentryUIApplication is already private
- You had to create a new branch, change the code, open a new PR
- There is no benefit

philipphofmann:
> Personally I would recommend to refrain from doing this type of refactoring.
> 
> SentryUIApplication is already private
> You had to create a new branch, change the code, open a new PR
> There is no benefit

Yes, but `sharedApplication` is still visible internally in the project, although it doesn't need to be. That's one benefit and less code. I don't mind the little extra overhead, cause it makes the codebase a bit cleaner.

PR #3866, opened 2024-04-22
fix: Replay Duration

brustolin:
## :scroll: Description

The end of the  segment was equal the time the last frame was taken, by the end should be the last frame date plus the duration of the frame

Unit Tests

_#skip-changelog_


Sources/Sentry/SentrySessionReplay.m
```diff
@@ -291,9 +291,7 @@ - (void)takeScreenshot
 
     _processingScreenshot = NO;
 
-    dispatch_queue_t backgroundQueue
-        = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
-    dispatch_async(backgroundQueue, ^{ [self->_replayMaker addFrameWithImage:screenshot]; });
+    [self->_replayMaker addFrameAsyncWithImage:screenshot];
```
brustolin, line 294: This function is already called from a background thread

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -21,52 +21,50 @@ enum SentryOnDemandReplayError: Error {
...
             return
         }
-        _frames.append(SentryReplayFrame(imagePath: imagePath, time: date))
+        frames.append(SentryReplayFrame(imagePath: imagePath, time: date))
         
-        while _frames.count > cacheMaxSize {
-            let first = _frames.removeFirst()
+        while frames.count > cacheMaxSize {
+            let first = frames.removeFirst()
             try? FileManager.default.removeItem(at: URL(fileURLWithPath: first.imagePath))
```
philipphofmann, line 65: `m`: Does it make sense to do some error handling here?

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -21,52 +21,50 @@ enum SentryOnDemandReplayError: Error {
...
 class SentryOnDemandReplay: NSObject {
     private let _outputPath: String
-    private let _onDemandDispatchQueue: DispatchQueue
-    
-    private var _starttime = Date()
-    private var _frames = [SentryReplayFrame]()
     private var _currentPixelBuffer: SentryPixelBuffer?
+    private var _totalFrames = 0
     
+    var frames = [SentryReplayFrame]()
```
philipphofmann, line 30: `m`: When we expose this property, we need to make it thread-safe. We currently only access it in tests. Maybe we can find a way to test the code, so we don't have to expose it to avoid making it thread safe when accessing it.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -87,9 +85,9 @@ class SentryOnDemandReplay: NSObject {
     }
     
     func releaseFramesUntil(_ date: Date) {
-        _onDemandDispatchQueue.async {
-            while let first = self._frames.first, first.time < date {
-                self._frames.removeFirst()
+        workingQueue.async {
+            while let first = self.frames.first, first.time < date {
```
philipphofmann, line 89: `m`: We also need to make this and other accesses to `frames` thread safe.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -21,52 +21,50 @@ enum SentryOnDemandReplayError: Error {
...
     private let _outputPath: String
-    private let _onDemandDispatchQueue: DispatchQueue
-    
-    private var _starttime = Date()
-    private var _frames = [SentryReplayFrame]()
     private var _currentPixelBuffer: SentryPixelBuffer?
+    private var _totalFrames = 0
     
+    var frames = [SentryReplayFrame]()
+    var workingQueue: DispatchQueue
```
philipphofmann, line 31: `m`: I think it would be better to add a second initializer to keep this private. Now, you could change this after initializing the class while a couple of blocks are already running async, which opens a door for bugs. The same applies to `dateProvider`.

Tests/SentryTests/Integrations/SessionReplay/SentryOnDemandReplayTests.swift
```diff
@@ -0,0 +1,102 @@
...
+            
+            expect(FileManager.default.fileExists(atPath: output.path)) == true
+            videoExpectation.fulfill()
+            try? FileManager.default.removeItem(at: output)
+        }
+        
+        wait(for: [videoExpectation], timeout: 1)
+    }
+    
+    private func waitWorkingQueue() {
```
philipphofmann, line 85: `m`: What made you choose using a real dispatch queue over using `SentryDispatchQueueWrapper` in combination with `TestSentryDispatchQueueWrapper`, which executes everything on the main thread, so you don't need this here? 
Obviously, you need to test your code with a real dispatch queue as well, but when you need this you simply can use `SentryDispatchQueueWrapper`.

Tests/SentryTests/Integrations/SessionReplay/SentryOnDemandReplayTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        let sut = getSut()
+        sut.addFrameAsync(image: UIImage.add)
+       
+        waitWorkingQueue()
+        
+        guard let frame = sut.frames.first else {
+            fail("Frame was not saved")
+            return
+        }
+        expect(FileManager.default.fileExists(atPath: frame.imagePath)) == true
```
philipphofmann, line 29: `m`: If I got it right, other classes rely on the sut writing files to the imagePath you pass in when initializing the sut. It would be great if this test validated the behavior of the class in the same way as other classes use it. Meaning instead of exposing the frames to get the image path, we could check if a file exists at the path that we pass to the constructor of the sut or we can test the behavior similarly as we already do in testGenerateVideo.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
brustolin: This is exposed for test reasons, I will look into some alternative.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -21,52 +21,50 @@ enum SentryOnDemandReplayError: Error {
...
-        _onDemandDispatchQueue.async {
-            self.asyncAddFrame(image: image)
+    func addFrameAsync(image: UIImage) {
+        workingQueue.async {
+            self.addFrame(image: image)
         }
     }
     
-    private func asyncAddFrame(image: UIImage) {
+    private func addFrame(image: UIImage) {
```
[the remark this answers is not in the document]
brustolin: This method is private and its always called from workingQueue, which makes it thread safe.
I created a second method to not have to write`self.` to access every property and function

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
brustolin: We could flood even more the console log.

Honestly I dont think it matter, I clean up function to delete the entire replay folder is coming soon.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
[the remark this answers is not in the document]
brustolin: For `SentryOndemandReplay` im using dispatchqueue as synicing mechanism.
This method is private and its always called from workingQueue (line 46), which makes it thread safe.
I created a second method to not have to write `self.` to access every property and function.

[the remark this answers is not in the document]
philipphofmann: When you call `addFrameAsync` / `workingQueue.async` then `addFrame` runs on a background thread. If you call that multiple times in a row, you will run into race conditions. The way the code uses the dispatchqueue isn't synchronized, as far as I know. Am I missing something? Sorry I don't get it how this code should be thread-safe.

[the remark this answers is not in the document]
brustolin: The default behaviour of a dispatch queue is `serial`

Try the following code and you will understand what Im doing here:

```swift
import Foundation

let queue = DispatchQueue(label: "My Queue")

queue.async {
    print("Starting waiting queue")
    Thread.sleep(forTimeInterval: 5)
}

for i in 0..<10 {
    queue.async {
        print("Queue \(i)")
    }
}
```

And this is the output you will get 100% of the time
```
Starting waiting queue
Queue 0
Queue 1
Queue 2
Queue 3
Queue 4
Queue 5
Queue 6
Queue 7
Queue 8
Queue 9
```

This is the same reason our ANR approach works.

Tests/SentryTests/Integrations/SessionReplay/SentryOnDemandReplayTests.swift
brustolin: I forgot that SentryDispatchQueuWrapper when not used from the DependencyContainer would work as a DispatchQueue.

Changed it already

Tests/SentryTests/Integrations/SessionReplay/SentryOnDemandReplayTests.swift
brustolin: The output instruction helps the class to do a clean up latter. But the classes that use sut dont read the files directly. 

But I agree that we should check whether the sut is writing in the right place.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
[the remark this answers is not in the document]
philipphofmann: Oh, sorry, my bad 🙈 ; I completely overlooked that the dispatch queue isn't concurrent.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -158,13 +165,14 @@ class SentryOnDemandReplay: NSObject {
...
         var start = dateProvider.date()
         var actualEnd = start
-        
-        for frame in frames {
-            if frame.time < beginning { continue } else if frame.time > end { break }
-            if frame.time < start { start = frame.time }
-            
-            actualEnd = frame.time
-            framesPath.append(frame.imagePath)
+        workingQueue.dispatchSync {
```
philipphofmann, line 172: `m`: When this is called from the main thread, this won't work, if I'm not mistaken, because we access the frames on the main thread, but they could be modified on a background thread in the meantime.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -158,13 +165,14 @@ class SentryOnDemandReplay: NSObject {
...
+            for frame in self.frames {
+                if frame.time < beginning { continue } else if frame.time > end { break }
+                if frame.time < start { start = frame.time }
+                
+                actualEnd = frame.time
+                framesPath.append(frame.imagePath)
```
philipphofmann, line 178: `l`: It's not the path of frames, but paths of frames.
```suggestion
                framePaths.append(frame.imagePath)
```

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -21,40 +21,48 @@ enum SentryOnDemandReplayError: Error {
...
     var cacheMaxSize = UInt.max
     
-    init(outputPath: String) {
+    convenience init(outputPath: String) {
+        self.init(outputPath: outputPath,
+                  workingQueue: SentryDispatchQueueWrapper(name: "io.sentry.onDemandReplay", attributes: nil),
+                  dateProvider: SentryCurrentDateProvider())
+    }
+    
+    init(outputPath: String, workingQueue: SentryDispatchQueueWrapper, dateProvider: SentryCurrentDateProvider) {
```
philipphofmann, line 49: Thanks 💯

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -112,17 +121,17 @@ class SentryOnDemandReplay: NSObject {
         videoWriter.startSession(atSourceTime: .zero)
         
         var frameCount = 0
-        let (frames, start, end) = filterFrames(beginning: beginning, end: beginning.addingTimeInterval(duration))
+        let (framesPath, start, end) = filterFrames(beginning: beginning, end: beginning.addingTimeInterval(duration))
```
philipphofmann, line 124: `l`: This name better reflects that it's an array.
```suggestion
        let (framePaths, start, end) = filterFrames(beginning: beginning, end: beginning.addingTimeInterval(duration))
```

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -21,40 +21,48 @@ enum SentryOnDemandReplayError: Error {
...
-    private var _starttime = Date()
-    private var _frames = [SentryReplayFrame]()
     private var _currentPixelBuffer: SentryPixelBuffer?
+    private var _totalFrames = 0
+    private let dateProvider: SentryCurrentDateProvider
+    private let workingQueue: SentryDispatchQueueWrapper
+    private var _frames = [SentryReplayFrame]()
+    
+    var frames: [SentryReplayFrame] {
+        _frames
```
philipphofmann, line 34: `m`: I think this still needs some synchorinization via the dispatch queue or an NSLock, because _frames could be modified on a background thread with the workingQueue, and this property could be called from any thread.

Tests/SentryTests/Integrations/SessionReplay/SentryOnDemandReplayTests.swift
```diff
@@ -0,0 +1,77 @@
...
+
+#if os(iOS) || os(tvOS)
+class SentryOnDemandReplayTests: XCTestCase {
+    
+    let dateProvider = TestCurrentDateProvider()
+    let outputPath = FileManager.default.temporaryDirectory
+    
+    func getSut() -> SentryOnDemandReplay {
+        let sut = SentryOnDemandReplay(outputPath: outputPath.path, 
+                                       workingQueue: TestSentryDispatchQueueWrapper(),
```
philipphofmann, line 15: `m`: Thanks for using the DispatchQueueWrapper. I think we should also add one test with the actual SentryDispatchQueueWrapper to validate that our code is thread safe.

philipphofmann (commented):
Thanks for the detailed explanation of how I misunderstood your use of the DispatchQueue. Either my understanding again is wrong, or we still have some synchronization issues.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
brustolin: So, yes and no.

If we call this from the main thread, it will be executed in the main thread but only after all other dispatch code finishes execution. Moreover, if we call any asynchronous code while this one is running, the new one will have to wait.

Once again, the dispatch queue only executes code serially

brustolin: Oh, SentryDispatchSync is not the same as DispatchQueueSync 😒

I will add a true sync function

philipphofmann: > Oh, SentryDispatchWrapper.dispatchSync is not the same as DispatchQueue.sync 😒

I think we must rename `SentryDispatchWrapper.dispatchSync` to avoid confusion.

philipphofmann: Ah you added a new true sync function. You must not rename this one. Sorry didn't see it.

Tests/SentryTests/Integrations/SessionReplay/SentryOnDemandReplayTests.swift
brustolin: Honestly I dont think we need.
We dont have a reliable way to test race condition, and this tests are often source of flakiness.

bruno-garcia: you can test a lot by making the code testable. I use signals so i can control when each thread does what so the code becomes deterministic. I can find some examples when i'm on the computer 
I'd say we can merge this as add later though

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -21,40 +21,54 @@ enum SentryOnDemandReplayError: Error {
...
-    private var _starttime = Date()
-    private var _frames = [SentryReplayFrame]()
     private var _currentPixelBuffer: SentryPixelBuffer?
+    private var _totalFrames = 0
+    private let dateProvider: SentryCurrentDateProvider
+    private let workingQueue: SentryDispatchQueueWrapper
+    private var _frames = [SentryReplayFrame]()
+    
+    #if TEST
+    var frames: [SentryReplayFrame] {
```
philipphofmann, line 34: Acceptable workaround.

philipphofmann (approved):
LGTM

Tests/SentryTests/Integrations/SessionReplay/SentryOnDemandReplayTests.swift
philipphofmann: There is never a reliable way to check if your code doesn't have race conditions. For some, you need to have some test code run for minutes to surface. What I like to do is to call my code that should be thread-safe from multiple threads in a tight loop to ensure it doesn't crash right away by using all real implementations and not fakes. I did that for the metrics and it helped me to surface some threading issues https://github.com/getsentry/sentry-cocoa/blob/c02bb1f4252fce2f69a5a6e36bde70f0211663fb/Tests/SentryTests/Swift/Metrics/BucketMetricsAggregatorTests.swift#L341-L365

Ideally, we would also test if it's working correctly, but not crashing is the minimum I like to do.

brustolin: Test Added

PR #3868, opened 2024-04-22
chore: Remove BaseIntegrationProtocol from Replay integration

brustolin:
Removing a swift protocol from SentrySessionReplayIntegration for hybrid SDK usage.

We really dont need the protocol, maybe we should reconsider its use.

_#skip-changelog_


krystofwoldrich:
Thanks, tested it with the RN sample, works.

brustolin:
> The hub uses it. Are you sure you can safely remove it?

It's working, no compiler error, no runtime error.

PR #3872, opened 2024-04-23
ref(profiling): move sliceGPUData to timeseries code unit; prefix C functions

armcknight:
Taking some time to clean up the implementation of SentryProfiler and organize it a little better before moving forward with adding the continuous profiling implementation.

This moves the `sliceGPU` function out of SentryProfiler.mm and into SentryProfilerTimeseries.mm. Also prefix the C function names with `_sentry` as in #3862 

for #3555 #skip-changelog


philipphofmann (approved):
LGTM 🚀

PR #3873, opened 2024-04-23
ref(profiling): move test helpers to separate code unit; prefix with _sentry

armcknight:
Following on https://github.com/getsentry/sentry-cocoa/pull/3872, for #3555, #skip-changelog

Move some functions only used for testing purposes to a separate code unit `SentryProfilerTestHelpers`


brustolin (approved):
Personally, I don't like this approach of adding so many test compiler declaratives; it makes it harder to read, and we will be testing code that will not be the same as what we ship. 

That being said, it works, so I won't oppose it.

philipphofmann (approved):
> Personally, I don't like this approach of adding so many test compiler declaratives; it makes it harder to read, and we will be testing code that will not be the same as what we ship.

I agree with you @brustolin 👍 , but this PR only moves the existing stuff around. We should discuss this somewhere else.

PR #3874, opened 2024-04-23
ref(profiling): rename more functions to prefix with `sentry_` and remove leading underscores on public members

armcknight:
Following on https://github.com/getsentry/sentry-cocoa/pull/3873, for https://github.com/getsentry/sentry-cocoa/issues/3555, #skip-changelog

Just renaming these functions to namespace then the same as was done in #3862


brustolin (approved):

philipphofmann (approved):
LGTM

PR #3877, opened 2024-04-23
ref: Use average color for redact masking

brustolin:
Use the text color or the average color of the image as the redaction mask.

_#skip-changelog_

### DIFF
--- a/Sources/Sentry/SentrySessionReplay.m
+++ b/Sources/Sentry/SentrySessionReplay.m
@@ -260,13 +260,12 @@ SentrySessionReplay ()
         _processingScreenshot = YES;
     }
 
-    UIImage *screenshot = [_screenshotProvider imageWithView:_rootView options:_replayOptions];
-
-    _processingScreenshot = NO;
-
-    dispatch_queue_t backgroundQueue
-        = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
-    dispatch_async(backgroundQueue, ^{ [self->_replayMaker addFrameWithImage:screenshot]; });
+    [_screenshotProvider imageWithView:_rootView
+                               options:_replayOptions
+                            onComplete:^(UIImage *screenshot) {
+                                self->_processingScreenshot = NO;
+                                [self->_replayMaker addFrameWithImage:screenshot];
+                            }];
 }
 
 @end
--- a/Sources/Swift/Tools/SentryViewPhotographer.swift
+++ b/Sources/Swift/Tools/SentryViewPhotographer.swift
@@ -9,7 +9,7 @@ import UIKit
 @available(iOS, introduced: 16.0)
 @available(tvOS, introduced: 16.0)
 @objcMembers
-class SentryViewPhotographer: NSObject {
+class SentryViewPhotographer: NSObject, SentryViewScreenshotProvider {
     
     private struct RedactRegion {
         let rect: CGRect
@@ -68,31 +68,47 @@ class SentryViewPhotographer: NSObject {
         ].compactMap { NSClassFromString($0) }
     }
     
-    @objc(imageWithView:options:)
-    func image(view: UIView, options: SentryRedactOptions) -> UIImage? {
-        
+    func image(view: UIView, options: SentryRedactOptions, onComplete: @escaping ScreenshotCallback ) {
         let image = UIGraphicsImageRenderer(size: view.bounds.size).image { _ in
             view.drawHierarchy(in: view.bounds, afterScreenUpdates: false)
         }
         
-      //  let mask = self.mask(view: view, options: options)
-        let maskSize = view.bounds.size.applying(CGAffineTransformMakeScale(0.005, 0.005))
+        let redact = self.mask(view: view, options: options)
         
-        let maskColor = UIGraphicsImageRenderer(size: maskSize).image { _ in
-            image.draw(in: CGRect(x: 0, y: 0, width: maskSize.width, height: maskSize.height))
+        DispatchQueue.global().async {
+            let screenshot = UIGraphicsImageRenderer(size: view.bounds.size, format: .init(for: .init(displayScale: 1))).image { context in
+                image.draw(at: .zero)
+                
+                for region in redact {
+                    (region.color ?? self.averageColor(of: context.currentImage, at: region.rect)).setFill()
+                    context.fill(region.rect)
+                }
+            }
+            onComplete(screenshot)
+        }
+    }
+    
+    private func averageColor(of image: UIImage, at region: CGRect) -> UIColor {
+        let colorImage = UIGraphicsImageRenderer(size: CGSize(width: 1, height: 1), format: .init(for: .init(displayScale: 1))).image { context in
+            guard let croppedImage = image.cgImage?.cropping(to: region) else {
+                UIColor.black.setFill()
+                context.fill(CGRect(x: 0, y: 0, width: 1, height: 1))
+                return
+            }
+            
+            context.cgContext.draw(croppedImage, in: CGRect(x: 0, y: 0, width: 1, height: 1), byTiling: false)
         }
         
-        let screenshot = UIGraphicsImageRenderer(size: view.bounds.size).image { context in
-            image.draw(at: .zero)
-            
-            context.cgContext.setFillColor(UIColor.red.cgColor)
-            
-            context.cgContext.clip()
-            maskColor.draw(in: view.bounds)
-            context.cgContext.restoreGState()
-        }
+        guard
+            let pixelData = colorImage.cgImage?.dataProvider?.data,
+            let data = CFDataGetBytePtr(pixelData) else { return .black }
         
-        return screenshot
+        let red = CGFloat(data[0]) / 255.0
+        let green = CGFloat(data[1]) / 255.0
+        let blue = CGFloat(data[2]) / 255.0
+        let alpha = CGFloat(data[3]) / 255.0
+        
+        return UIColor(red: red, green: green, blue: blue, alpha: alpha)
     }
     
     private func mask(view: UIView, options: SentryRedactOptions?) -> [RedactRegion] {
--- a/Sources/Swift/Tools/SentryViewPhotographer.swift
+++ b/Sources/Swift/Tools/SentryViewPhotographer.swift
@@ -11,6 +11,45 @@ import UIKit
 @objcMembers
 class SentryViewPhotographer: NSObject {
     
+    private struct RedactRegion {
+        let rect: CGRect
+        let color: UIColor?
+        
+        func splitBySubtracting(region: CGRect) -> [RedactRegion] {
+            guard rect.intersects(region) else { return [self] }
+            guard !region.contains(rect) else { return [] }
+            
+            let intersectionRect = rect.intersection(region)
+            var resultRegions: [CGRect] = []
+            
+            // Calculate the top region.
+            resultRegions.append(CGRect(x: rect.minX,
+                                        y: rect.minY,
+                                        width: rect.width,
+                                        height: intersectionRect.minY - rect.minY))
+            
+            // Calculate the bottom region.
+            resultRegions.append(CGRect(x: rect.minX,
+                                        y: intersectionRect.maxY,
+                                        width: rect.width,
+                                        height: rect.maxY - intersectionRect.maxY))
+            
+            // Calculate the left region.
+            resultRegions.append(CGRect(x: rect.minX,
+                                        y: max(rect.minY, intersectionRect.minY),
+                                        width: intersectionRect.minX - rect.minX,
+                                        height: min(intersectionRect.maxY, rect.maxY) - max(rect.minY, intersectionRect.minY)))
+            
+            // Calculate the right region.
+            resultRegions.append(CGRect(x: intersectionRect.maxX,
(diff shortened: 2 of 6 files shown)


Sources/Sentry/SentrySessionReplay.m
```diff
@@ -306,11 +306,12 @@ - (void)takeScreenshot
...
-    UIImage *screenshot = [_screenshotProvider imageWithView:_rootView options:_replayOptions];
-
-    _processingScreenshot = NO;
-
-    [self->_replayMaker addFrameAsyncWithImage:screenshot];
+    [_screenshotProvider imageWithView:_rootView
+                               options:_replayOptions
+                            onComplete:^(UIImage *screenshot) {
+                                self->_processingScreenshot = NO;
+                                [self->_replayMaker addFrameAsyncWithImage:screenshot];
```
philipphofmann, line 313: `m`: Can this lead to a retain cycle? Should we maybe use `__weak` to self here?

Sources/Swift/Tools/RedactRegion.swift
```diff
@@ -0,0 +1,48 @@
+#if canImport(UIKit)
+import Foundation
+import UIKit
+
+struct RedactRegion {
+    let rect: CGRect
+    let color: UIColor?
+    
+    init(rect: CGRect, color: UIColor? = nil) {
```
philipphofmann, line 9: `m`: What made you choose to make the color nullable? Only the tests omit this parameter. The source code always passes in the color.

Tests/SentryTests/RedactRegionTests.swift
```diff
@@ -0,0 +1,102 @@
+import Foundation
+import Nimble
+@testable import Sentry
+import XCTest
+#if os(iOS) || os(tvOS)
+class RedactRegionTests: XCTestCase {
```
philipphofmann, line 6: `m`: Those are neat test cases. I think it would make sense to add some test cases when the CGRects don't overlap exactly. For example, what happens if the region to subtract is bigger than the sut region.
For example
```
func testSplitBySubtracting_BottomIsWider() {
    let sut = RedactRegion(rect: CGRect(x: 0, y: 0, width: 100, height: 100), color: .red)
    
    let result = sut.splitBySubtracting(region: CGRect(x: 0, y: 50, width: 150, height: 50))
    
    expect(result.count) == 1
    expect(result.first?.rect) == CGRect(x: 0, y: 0, width: 100, height: 50)
    expect(result.first?.color) == .red
}
```

Sources/Swift/Tools/RedactRegion.swift
```diff
@@ -0,0 +1,48 @@
...
+    }
+    
+    func splitBySubtracting(region: CGRect) -> [RedactRegion] {
+        guard rect.intersects(region) else { return [self] }
+        guard !region.contains(rect) else { return [] }
+        
+        let intersectionRect = rect.intersection(region)
+        var resultRegions: [CGRect] = []
+        
+        // Calculate the top region.
```
philipphofmann, line 21: Those are useful comments. Thanks 💯

Sources/Swift/Tools/SentryViewPhotographer.swift
```diff
@@ -1,16 +1,12 @@
...
 import CoreGraphics
 import Foundation
 import UIKit
 
-@available(iOS, introduced: 16.0)
-@available(tvOS, introduced: 16.0)
 @objcMembers
-class SentryViewPhotographer: NSObject {
-    
+class SentryViewPhotographer: NSObject, SentryViewScreenshotProvider {
```
philipphofmann, line 9: `m`: This class has a low test coverage. Does it make sense to add some tests for these changes in this PR?

Sources/Swift/Tools/SentryViewPhotographer.swift
```diff
@@ -29,23 +25,62 @@ class SentryViewPhotographer: NSObject {
...
+        let redact = self.mask(view: view, options: options)
+        let imageSize = view.bounds.size
+        DispatchQueue.global().async {
+            let screenshot = UIGraphicsImageRenderer(size: imageSize, format: .init(for: .init(displayScale: 1))).image { context in
+                image.draw(at: .zero)
+                
+                for region in redact {
+                    (region.color ?? self.averageColor(of: context.currentImage, at: region.rect)).setFill()
+                    context.fill(region.rect)
+                    //region.image?.draw(in: region.rect)
```
philipphofmann, line 42: `l`: Commented code.

Sources/Sentry/SentrySessionReplay.m
brustolin: It can hold the reference for a little longer. But will not create a retain cycle, once the callback returns everything will be disposed.

Sources/Swift/Tools/SentryViewPhotographer.swift
brustolin, line 42:
```suggestion
```

Sources/Swift/Tools/SentryViewPhotographer.swift
brustolin: This bugs me too, but this class relies too much on UI and drawing, testing the Image results will most likely be flaky.

I will try to extract some code from here that can be tested.

Sources/Swift/Tools/RedactRegion.swift
brustolin: Color can be nil, see `color` function from `SentryViewPhotographer`

```swift
    private func color(for view: UIView) -> UIColor? {
         return (view as? UILabel)?.textColor
     }
```

Sources/Sentry/SentrySessionReplay.m
philipphofmann: But we call onComplete with a strong reference to self async here https://github.com/getsentry/sentry-cocoa/blob/cb3bca31cafa521d1bfe3e3e627d9284d95c7e6e/Sources/Swift/Tools/SentryViewPhotographer.swift#L35-L45

We have to be careful not to introduce retains cycles in the future. Currently, it doesn't matter much, because we always have a strong reference to `SentrySessionReplay` once session replay is enable. Maybe it's safer to use a `__weak` reference now to avoid problems in the future. Up to you.

Tests/SentryTests/UIImageHelperTests.swift
```diff
@@ -0,0 +1,65 @@
...
+    private let testFrame = CGRect(x: 0, y: 0, width: 100, height: 100)
+    
+    func testAverageColorRed() {
+        let begin = Date()
+            let image = UIGraphicsImageRenderer(size: testFrame.size).image { context in
+                UIColor.red.setFill()
+                context.fill(testFrame)
+            }
+            
+            expect(UIImageHelper.averageColor(of: image, at: self.testFrame)) == .red
```
philipphofmann, line 18: `l`:  The code formatting here seems to be a bit off.

Tests/SentryTests/UIImageHelperTests.swift
```diff
@@ -0,0 +1,65 @@
+#if canImport(UIKit)
+import Foundation
+import Nimble
+@testable import Sentry
+import XCTest
+
+class UIImageHelperTests: XCTestCase {
```
philipphofmann, line 7: Neat tests 👏

Tests/SentryTests/RedactRegionTests.swift
```diff
@@ -0,0 +1,143 @@
...
+        
+        let result = sut.splitBySubtracting(region: CGRect(x: 0, y: 25, width: 100, height: 50))
+        
+        expect(result.count) == 3
+        expect(result[0].rect) == CGRect(x: 50, y: 0, width: 100, height: 25)
+        expect(result[1].rect) == CGRect(x: 50, y: 75, width: 100, height: 25)
+        expect(result[2].rect) == CGRect(x: 100, y: 25, width: 50, height: 50)
+    }
+
+    func testSplitBySubtracting_TopIsWider() {
```
philipphofmann, line 101: Thanks for adding more tests for these edge cases 💯

Tests/SentryTests/UIRedactBuilderTests.swift
```diff
@@ -0,0 +1,130 @@
+#if os(iOS)
+import Foundation
+import Nimble
+@testable import Sentry
+import UIKit
+import XCTest
+
+class UIRedactBuilderTests: XCTestCase {
```
philipphofmann, line 8: Those tests are the specification to the sut. I don't need to read the code of the sut to understand what the class does. Excellent 🥇

Tests/SentryTests/UIRedactBuilderTests.swift
```diff
@@ -0,0 +1,130 @@
...
+    
+    private class RedactOptions: SentryRedactOptions {
+        var redactAllText: Bool = true
+        var redactAllImages: Bool = true
+    }
```
philipphofmann, line 13: `m`: If I'm not mistaken we always test with both values true. It would be great to test with different combinations of these too.

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -0,0 +1,136 @@
...
+    //This is a list of UIView subclasses that will be ignored during redact process
+    var ignoreClasses: [AnyClass] 
+    //This is a list of UIView subclasses that need to be redacted from screenshot
+    var redactClasses: [AnyClass]
+    
+    init() {
+        redactClasses = [ UILabel.self, UITextView.self, UITextField.self ] + [
+            "_TtCOCV7SwiftUI11DisplayList11ViewUpdater8Platform13CGDrawingView",
+            "_TtC7SwiftUIP33_A34643117F00277B93DEBAB70EC0697122_UIShapeHitTestingView",
+            "SwiftUI._UIGraphicsView", "SwiftUI.ImageLayer"
```
philipphofmann, line 61: `m`: Where do these class names come from? Can you add a comment to explain this?

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -0,0 +1,136 @@
...
+    
+    private func shouldRedact(view: UIView, redactText: Bool, redactImage: Bool) -> Bool {
+        if redactImage, let imageView = view as? UIImageView {
+            return shouldRedact(imageView: imageView)
+        }
+        return redactText && redactClasses.contains { view.isKind(of: $0) }
+    }
+    
+    private func shouldRedact(imageView: UIImageView) -> Bool {
+        // Checking the size is to avoid redact gradient backgroud that
```
philipphofmann, line 95: `l`
```suggestion
        // Checking the size is to avoid redact gradient background that
```

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+    }
+    
+    private func color(for view: UIView) -> UIColor? {
+        return (view as? UILabel)?.textColor
+    }
+    
+    private func hasBackground(_ view: UIView) -> Bool {
+        return view.backgroundColor != nil && (view.backgroundColor?.cgColor.alpha ?? 0) > 0.9
```
philipphofmann, line 131: `m`: Why only if the alpha is greater than `0.9`? Can you add this info as a comment?

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -0,0 +1,136 @@
...
+            return shouldRedact(imageView: imageView)
+        }
+        return redactText && redactClasses.contains { view.isKind(of: $0) }
+    }
+    
+    private func shouldRedact(imageView: UIImageView) -> Bool {
+        // Checking the size is to avoid redact gradient backgroud that
+        // are usually small lines repeating
+        guard let image = imageView.image, image.size.width > 10 && image.size.height > 10  else { return false }
+        return image.imageAsset?.value(forKey: "_containingBundle") == nil
```
philipphofmann, line 98: `l`: Out of curiosity. What is the `_containingBundle`?

philipphofmann (commented):
Thanks for adding more tests and extracting some code to make it testable 👍

Sources/Swift/Tools/UIImageHelper.swift
```diff
@@ -0,0 +1,36 @@
+#if canImport(UIKit) && !SENTRY_NO_UIKIT
+#if os(iOS) || os(tvOS)
+
+import Foundation
+import UIKit
+
+final class UIImageHelper {
+    private init() { }
+    
+    static func averageColor(of image: UIImage, at region: CGRect) -> UIColor {
```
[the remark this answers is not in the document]
brustolin: I prefer to organise helper functions into static classes. 
We dont have static class in Swift, so static function in a class that cannot be initialised it is.

Tests/SentryTests/UIRedactBuilderTests.swift
brustolin: True. Forgot to test no redacting when the flag is false

Sources/Swift/Tools/UIRedactBuilder.swift
brustolin: Information about the bundle that the image is coming from.

philipphofmann (approved):
LGTM. Thanks for including all the feedback.

bruno-garcia:
hope we can ship this one Monday to unblock customers trialing who are affected by high CPU

PR #3881, opened 2024-04-23
ci: retry UI tests up to 3 times

armcknight:
I've been seeing a lot of this lately, the iOS-SwiftUITests-Runner not reporting an app launch, so the tests run times out. One retry usually fixes it, but that's annoying to do manually.
> Testing failed: iOS-SwiftUITests-Runner (13303) encountered an error (The test runner failed to initialize for UI testing. (Underlying Error: Timed out while loading Accessibility.))

https://github.com/getsentry/sentry-cocoa/actions/runs/8806093882/job/24170123920?pr=3875#step:6:1014

#skip-changelog

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -141,8 +141,9 @@ jobs:
         if: ${{ matrix.device == 'iPhone 14 (16.4)' }}
         run: ./scripts/create-simulator.sh 14.3.1 16.4 16-4
 
+        # UI tests have been flaking lately due to: Testing failed: iOS-SwiftUITests-Runner (13303) encountered an error (The test runner failed to initialize for UI testing. (Underlying Error: Timed out while loading Accessibility.))
       - name: Run Fastlane
-        run: fastlane ui_tests_ios_swift device:"${{matrix.device}}" 
+        run: for i in {1..3}; do fastlane ui_tests_ios_swift device:"${{matrix.device}}" && break ; done
 
       - name: Archiving Raw Test Logs
         uses: actions/upload-artifact@v4


philipphofmann:
I think we should only retry the UI tests when this specific error occurs:
> iOS-SwiftUITests-Runner (13303) encountered an error (The test runner failed to initialize for UI testing. (Underlying Error: Timed out while loading Accessibility.))

We could use grep or something similar as we already do for Android https://github.com/getsentry/sentry-java/blob/a33b07683d6fe58c9eae5c68dae508d1565b4561/.github/workflows/integration-tests-ui.yml#L64-L71.

Retrying the UI tests blindly will lead to more flaky tests because we will ignore them. Also, it will increase CI time, which is a bummer for productivity. IMO, we should disable or remove the flaky tests or fix them.

armcknight:
> I think we should only retry the UI tests when this specific error occurs:
> 
> > iOS-SwiftUITests-Runner (13303) encountered an error (The test runner failed to initialize for UI testing. (Underlying Error: Timed out while loading Accessibility.))
> 
> We could use grep or something similar as we already do for Android [getsentry/sentry-java@`a33b076`/.github/workflows/integration-tests-ui.yml#L64-L71](https://github.com/getsentry/sentry-java/blob/a33b07683d6fe58c9eae5c68dae508d1565b4561/.github/workflows/integration-tests-ui.yml#L64-L71).
> 
> Retrying the UI tests blindly will lead to more flaky tests because we will ignore them. Also, it will increase CI time, which is a bummer for productivity. IMO, we should disable or remove the flaky tests or fix them.

People are already ignoring the flaky tests. I just looked through the last several days' worth of PRs that have been merged without green checks, and they contain such failures. Nobody seems to care to investigate the failures–it's not even possible to do in the UI tests because of the presence of the bug I just fixed in https://github.com/getsentry/sentry-cocoa/pull/3898#pullrequestreview-2020631944

We can fix this in an incremental fashion. All you are doing by asking for the world instead of taking it one step at a time is slowing down work for everyone else. You are also free to help us take some of those steps. But please do not simply stop me from unblocking myself on this, so I can get back to spending my time on more important work than babysitting CI.

I recorded more thoughts in https://github.com/getsentry/sentry-cocoa/issues/3897

philipphofmann:
> I just looked through the last several days' worth of PRs that have been merged without green checks, and they contain such failures. 

I think you don't necessarily need all CI checks to be green when hitting merge. We know some unit tests are flaky, and it's hard to fix all flakiness. When I change something and I see a completely unrelated test starting to fail, I can either investigate and check if I caused it, or I can decide that it has nothing to do with it and ignore the test failure. If we start using retries you won't see that a test failed, which will lead to even more flaky tests. 

> All you are doing by asking for the world instead of taking it one step at a time is slowing down work for everyone else. You are also free to help us take some of those steps. But please do not simply stop me from unblocking myself on this, so I can get back to spending my time on more important work than babysitting CI.

I'm only not approving this PR, because I believe it will make things worse. Nobody says that all CI checks have to be green to merge a PR as mentioned above. It's only super annoying when you have to do a release.

I created a helper function to rerun flaky tests directly in the test code. Maybe that's a better alternative. Please let me know what you think https://github.com/getsentry/sentry-cocoa/pull/3903.

PR #3883, opened 2024-04-23
ci: retry unit tests up to 3 times

armcknight:
Similar story as #3881. I have to retry unit tests on basically every single PR I open because some test will flake.

#skip-changelog

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -209,7 +209,8 @@ jobs:
         # passed to xcodebuild doesn't end up in the job name,
         # because GitHub Actions don't provide an easy way of
         # manipulating string in expressions.
-        run: ./scripts/xcode-test.sh ${{matrix.platform}} ${{matrix.test-destination-os}} $GITHUB_REF_NAME ci test-without-building "${{matrix.device}}" TestCI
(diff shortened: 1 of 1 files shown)


philipphofmann:
I don't think that's a great idea. Retrying unit tests 3 times will lead to ignoring flaky tests, which leads to more flaky tests and increases CI duration. Instead, we should disable the flaky tests and fix them.

armcknight:
> I don't think that's a great idea. Retrying unit tests 3 times will lead to ignoring flaky tests, which leads to more flaky tests and increases CI duration. Instead, we should disable the flaky tests and fix them.

We can actually do both. Any time multiple runs are needed and the same test doesn't fail in all 3, that's a great signal that that test is flaky. I captured this idea in https://github.com/getsentry/sentry-cocoa/issues/3897.

At the end of the day, no amount of automation will prevent people from ignoring issues. We need the tools _and_ the cultural discipline to use them in order to overcome this problem.

philipphofmann:
As you disable most test flakes with https://github.com/getsentry/sentry-cocoa/pull/3904, maybe we can close this. I would prefer disabling more flaky tests and https://github.com/getsentry/sentry-cocoa/issues/3897 over this.

PR #3884, opened 2024-04-24
chore: Rename SentryDispatchQueueWrapper function name for Swift

brustolin:
`SentryDispatchQueueWrapper` is confusing to use with Swift. It behaves as if it were a DispatchQueue but also includes functions that interact with the main queue. When converting to Swift, the functions that manipulate its own queue overlap with those that interact with the main queue, distinguished only by the argument name, which can be omitted because it's a block.

Renaming those functions in Swift would better reflect their respective actions.

_#skip-changelog_


philipphofmann (approved):
Thanks @brustolin 🥇

PR #3886, opened 2024-04-24
ci: Use replace sentry paths filter with dorny

philipphofmann:
We archived our fork of dorny/paths-filter. The sentry repo also uses dorny/path-filter, see https://github.com/getsentry/sentry/pull/64233.

#skip-changelog

### DIFF
--- a/.github/workflows/changes-in-high-risk-code.yml
+++ b/.github/workflows/changes-in-high-risk-code.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Get changed files
         id: changes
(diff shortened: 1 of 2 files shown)


.github/workflows/changes-in-high-risk-code.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Get changed files
         id: changes
-        uses: getsentry/paths-filter@4512585405083f25c027a35db413c2b3b9006d50 # v2.11.1
+        uses: dorny/paths-filter@0bc4621a3135347011ad047f9ecf449bf72ce2bd # v3.0.0
```
brustolin, line 22: Why are we moving await from Sentry's `paths-filter`, if we need functions that it does not offer should we propose an update?

Issue #2343, opened 2022-10-31
Use `paths-filter` action in CI


armcknight:
### Description

The missing functionality GitHub should have for its workflow triggers: https://github.com/getsentry/paths-filter

This will cut down on a lot of wasted compute time, and headaches with flaky tests failing when it’s not even necessary to run them.

brustolin:
Im curious to know where would this help? Do you have any scenarios in mind?

IMO, If we need a job or step to run with a specific path filter we could create a different workflow, no need to use another github action.

brustolin:
I thought
```yaml
  pull_request:
    paths:
```
was meant for this.

armcknight:
Sorry I didn't see your earlier message from some time ago. Check this out: https://github.com/dorny/paths-filter/issues/107#issuecomment-954215848

philipphofmann:
I tried using path-filters in https://github.com/getsentry/sentry-cocoa/pull/3887 and it turns out it's a bit more complicated than using GH actions builtin `paths:`. As we don't have any required checks the advantage described in https://github.com/dorny/paths-filter/issues/107#issuecomment-954215848 doesn't apply to our repository. I think we can continue using  GH actions builtin `paths:` instead to optimize our CI. We can close this.

PR #3887, opened 2024-04-24
ci: Use dorny paths-filter for tests.yml

philipphofmann:
#skip-changelog

### DIFF
--- a/.github/file-filters.yml
+++ b/.github/file-filters.yml
@@ -16,23 +16,41 @@ fastlane: &fastlane
 
 sources: &sources
   - 'Sources/**'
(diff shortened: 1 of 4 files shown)


brustolin (commented):
Can you write a short description why we need this and what is the benefits? 

It looks like a more complex way to control GitHub workflow.

philipphofmann:
I opened this to evaluate https://github.com/getsentry/sentry-cocoa/issues/2343, but we can close this as it complicates things, see https://github.com/getsentry/sentry-cocoa/issues/2343#issuecomment-2074720017.

PR #3888, opened 2024-04-24
ci: Remove duplicate sample paths in test.yml

philipphofmann:
We already include all sources via Sample/**.

https://github.com/getsentry/sentry-cocoa/blob/6813f7c156f1e842058549f1801b2eede804f37f/.github/workflows/test.yml#L14

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,14 +20,6 @@ on:
       - ".codecov.yml"
       - Sentry.xcodeproj
 
-      # run the workflow any time an Xcode scheme changes for a sample app
-      - "Samples/tvOS-Swift/tvOS-Swift.xcodeproj/xcshareddata/xcschemes/tvOS-Swift.xcscheme"
-      - "Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS13-Swift.xcscheme"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -20,14 +20,6 @@ on:
       - ".codecov.yml"
       - Sentry.xcodeproj
 
-      # run the workflow any time an Xcode scheme changes for a sample app
-      - "Samples/tvOS-Swift/tvOS-Swift.xcodeproj/xcshareddata/xcschemes/tvOS-Swift.xcscheme"
-      - "Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS13-Swift.xcscheme"
-      - "Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-SwiftUITests.xcscheme"
-      - "Samples/iOS-Swift/iOS-Swift.xcodeproj/xcshareddata/xcschemes/iOS-Swift.xcscheme"
```
brustolin, line 27: I believe we should keep at least iOS-Swift, otherwise we fix a problem in the app itself and the UI test wont run.

brustolin: I see. Perfect

brustolin (approved):

PR #3870, opened 2024-04-22
New Memory Termination Handling (OOM)

naftaly:
## :scroll: Description

This PR introduces `memory pressure` and `memory level` systems in order to effectively track memory terminations and correctly report them.

[Example Report](https://bedroomcode.sentry.io/issues/5235790069/events/latest/?query=is%3Aunresolved&referrer=latest-event&statsPeriod=24h&stream_index=0)

Screenshot 2024-04-22 at 3 34 56 PM
Screenshot 2024-04-22 at 3 35 08 PM

## :bulb: Motivation and Context

A lot has changed since [this article](https://engineering.fb.com/2015/08/24/ios/reducing-fooms-in-the-facebook-ios-app/) about handling memory terminations. In order to bring Sentry OOM handling up to date, some additional heuristics are needed. 

At this point, this is more of a proposal than a full PR since a few things would need to be updated to land this. For example, SentryAppMemoryTracker should be move to the dependency container.

## :green_heart: How did you test it?

Some new tests were added for new memory classes. More in depth testing is required as well as some changes to DI in order to allow for these tests.

## :pencil: Checklist

You have to check all boxes before merging:

- [ ] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

Issue #3518, opened 2023-12-22
Add memory info to low memory breadcrumb


barisyild:
### Problem Statement

![image](https://github.com/getsentry/sentry-cocoa/assets/25794892/8509aed0-dd30-4826-91ed-7c68d3b3ff45)

LOW_MEMORY does not provide enough information.
Is it possible that low memory warnings contain detailed information about memory?

### Solution Brainstorm

Things like memory usage may come up with the LOW_MEMORY warning.

philipphofmann:
We could use the same approach as we already do for getting memory statistics for profiles https://github.com/getsentry/sentry-cocoa/blob/3297d6e29cd37151f68e9c605aea8b44e5e432cb/Sources/Sentry/SentryMetricProfiler.mm#L176-L177

philipphofmann:
We could also add a memory footprint breadcrumb every x seconds for x seconds after receiving this warning, to see if the footprint goes down.

philprime:
Related to #2514

PR #3895, opened 2024-04-24
feat: Add SIGTERM support

naftaly:
## :scroll: Description

Added support for catching the SIGTERM signal.

Screenshot 2024-04-24 at 12 45 08 PM
Screenshot 2024-04-24 at 12 45 16 PM

All Apple OS's send SIGTERM (like any good unix based system) when they want to request a graceful termination of an application, when the app doesn't quit in those circumstances, it'll receive a SIGKILL. This often gives us the opportunity to get a stack trace where we know Apple wants the app to be terminated but don't know why (watchdog events). This is just one more piece of information that can be added to the puzzle of figuring out unexplained app terminations.

I ran the tests as well as [Tested on a real app](https://bedroomcode.sentry.io/issues/5231595050/?query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h&stream_index=1).

## :pencil: Checklist


Sources/SentryCrash/Recording/Tools/SentryCrashSignalInfo.c
```diff
@@ -115,7 +121,10 @@ static const SentryCrashSignalInfo g_fatalSignalData[] = {
     SIGNAL_INFO_NOCODES(SIGPIPE),
     SIGNAL_INFO(SIGSEGV, g_sigSegVCodes),
     SIGNAL_INFO_NOCODES(SIGSYS),
-    SIGNAL_INFO(SIGTERM, g_sigTrapCodes),
+#if SIGTRAP_ENABLED
+    SIGNAL_INFO(SIGTRAP, g_sigTrapCodes),
+#endif
+    SIGNAL_INFO_NOCODES(SIGTERM),
```
armcknight, line 127: I think we can actually leave out the SIGTRAP_ENABLED macro and just allow SIGTRAP monitoring as well. This was fixed upstream in https://github.com/kstenerud/KSCrash/pull/370. Also, a few lines down, a comment explains that:
> Dereferencing a NULL pointer ... causes SIGTRAP, 0 on arm
So we probably want to catch those.

armcknight (commented):
Thanks for the PR @naftaly! I think it's fine if we leave in the SIGTRAP logic, it looks like that was a bug that has since been fixed upstream. Would you mind making that small change?

Sources/SentryCrash/Recording/Tools/SentryCrashSignalInfo.c
naftaly: You're right. I pulled that whole part out. So now we have the small bug fix and addition of SIGTERM.

armcknight (commented):
The code changes look good to me, we might just want to push this diff up in one of our own branches to get all the other tests running.

naftaly:
@armcknight is there anything you need me to do to get this on in?

armcknight:
I'm just running the other tests really quick in https://github.com/getsentry/sentry-cocoa/pull/3946/, and then we can merge this! Thanks @naftaly for your patience.

armcknight (approved):
I'm happy with the results of the tests in #3946! 🙏🏻

philipphofmann (commented):
Just some missing tests. Apart from that LGTM. Thanks for doing this @naftaly 💯

Sources/SentryCrash/Recording/SentryCrashDoctor.m
```diff
@@ -351,6 +351,11 @@ - (BOOL)isMemoryCorruption:(NSDictionary *)report
     return NO;
 }
 
+- (BOOL)wasProgramTerminationRequested:(NSDictionary *)errorReport
+{
+    return [(errorReport[@SentryCrashField_Signal][@SentryCrashField_Signal]) intValue] == SIGTERM;
```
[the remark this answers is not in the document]
naftaly: Added the test in https://github.com/getsentry/sentry-cocoa/pull/3895/commits/4a8091fc633e7c5bdb1f1689ff5e2033afee241d

philipphofmann (approved):
LGTM, thanks for the test @naftaly 👏

eric:
What is this telling me when I get this on iOS and tvOS? Is this a good or a bad thing?

naftaly:
> What is this telling me when I get this on iOS and tvOS? Is this a good or a bad thing?

It’s the OS trying to tell you that it wants apps to exit gracefully. I’ve seen it happen in many different scenarios. Some of the interesting ones are when the device is shutdown, when an app will be terminated to update it and when something changes in traits that might require the OS to restart apps.

What’s interesting is that previously this would all go unnoticed to us the developers. If these happen when the user can see them, or perceive them, then we have an issue that need to be resolved. If they happen in the background, then we just hope that state restauration is well implemented and users don’t notice it.

> 
> What would Sentry have been reporting otherwise? The Watchdog event?

Likely nothing, it would have gone unnoticed.

eric:
We've always had a large number of `WatchdogTermination` events that didn't correspond to memory issues. Could it have been that these were actually receiving SIGTERM and it was not handled/reported and thus turned into the `WatchdogTermination` event?

naftaly:
> We've always had a large number of `WatchdogTermination` events that didn't correspond to memory issues. Could it have been that these were actually receiving SIGTERM and it was not handled/reported and thus turned into the `WatchdogTermination` event?

I guess it’s possible since the Watchdog code in Sentry (if I’m not mistaken) simply returns at the end of its heuristics and calls it an OOM/Watchdog event. So if the actual signal isn’t caught, then that might be the default. 

FWIW, I’ve also proposed a [PR to add real OOM recognition](https://github.com/getsentry/sentry-cocoa/pull/3870) so that could help with other false positives. It might end up in [KSCrash first though](https://github.com/kstenerud/KSCrash/pull/476) which is what Sentry uses for crash reporting.

naftaly:
@eric if you’re getting SIGTERM’s now, it would be interesting to know what type of event is going down (if any). That would help understand exactly what is happening or will happen for your app.

eric:
I just received the first SIGTERM on our Testflight beta today, and so wasn't sure what to do with it... it didn't really feel actionable.

eric:
screenshot-2024-05-21-15 55 38@2x

We do currently send memory usage with all of our breadcrumbs to try to give us a better idea of where we have actual memory issues. It always seemed silly to me that Sentry couldn't at least sample the memory usage once per second and store it somewhere to give a more informed message related to the Watchdog events.

naftaly:
> I just received the first SIGTERM on our Testflight beta today, and so wasn't sure what to do with it... it didn't really feel actionable.

If it’s background and it feels unactionable, then I would not prioritize it, but still keep in mind it’s there and happening. If it’s foreground, then I usually start adding breadcrumbs to make it actionable if the stack isn’t enough to give you an idea of why it is going on. These issues are definitely going to harder than your run of the mill exception since these happen based on what happened in the past vs. A mistake that might have happened at that point in time. If you can share the stack (of all threads) then I would be happy to take a look and see if I spot anything.

eric:
It doesn't seem like much of anything is happening here: https://fancy-bits-llc.sentry.io/issues/5386894502/

What is the signal that is sent when the OS decides that it's time for your app to stop running in the background when it needs to get more memory? Is SIGTERM ever going to be sent during normal operation?

naftaly:
> It doesn't seem like much of anything is happening here: https://fancy-bits-llc.sentry.io/issues/5386894502/
> 

I don’t have access. Maybe copy it to a gist or something.

> What is the signal that is sent when the OS decides that it's time for your app to stop running in the background when it needs to get more memory? 

That would be SIGKILL, we can’t catch it. Sometimes, if we’re really lucky, a SIGTERM comes before it, but I’ve not seen it often.

> Is SIGTERM ever going to be sent during normal operation?

Depends what normal operation is, but yes, it can happen at any time. It’s a vestige of unix where it’s meant to tell apps to do their cleanup and exit otherwise be “exited”.

philipphofmann:
> We do currently send memory usage with all of our breadcrumbs to try to give us a better idea of where we have actual memory issues.

We have that planned here: https://github.com/getsentry/sentry-cocoa/issues/3518#issuecomment-1909911490, but we haven't gotten to it yet, sorry.

> It always seemed silly to me that Sentry couldn't at least sample the memory usage once per second and store it somewhere to give a more informed message related to the Watchdog events.

Yes, that makes sense. I created an issue. Thanks for the input 👏 , @eric. https://github.com/getsentry/sentry-cocoa/issues/4003

Issue #2325, opened 2022-10-25
Support `hints`


marandaneto:
### Description

This is needed for [https://github.com/getsentry/sentry-cocoa/issues/1460](<https://github.com/getsentry/sentry-cocoa/issues/1460>) and [https://github.com/getsentry/sentry-cocoa/issues/2650](<https://github.com/getsentry/sentry-cocoa/issues/2650>)
Also for [https://develop.sentry.dev/sdk/features/#http-client-errors](<https://develop.sentry.dev/sdk/features/#http-client-errors>) -> [PR](<https://github.com/getsentry/sentry-cocoa/pull/2308>)
`Hints` are part of the [Unified API](<https://develop.sentry.dev/sdk/unified-api/#hints>).
Hints are generally a `Map<String, Any>`, on Java leverage the data bag with a [typed class with helper methods](<https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Hint.java>).

philipphofmann:
If we add an overload for `beforeSend` with a `hint` we can do this without a breaking change.

philipphofmann:
On Java, we have [typed hints](https://github.com/getsentry/sentry-java/tree/beba01476ab71745b0cae821bed705d582374161/sentry/src/main/java/io/sentry/hints). I think that would be better than having a `Map<String, Any>`.

marandaneto:
> On Java, we have [typed hints](https://github.com/getsentry/sentry-java/tree/beba01476ab71745b0cae821bed705d582374161/sentry/src/main/java/io/sentry/hints). I think that would be better than having a `Map<String, Any>`.

Yes, it used to be a `Map`, now it is a typed Hint, which is better.

philipphofmann:
Hints must allow capturing stacktraces only for one event if capture stacktraces is disabled in the options, as requested in https://github.com/getsentry/sentry-cocoa/issues/3386.

marcelofabri:
Thanks, @philipphofmann! Can you expand a little bit more on how much of a priority this is from Sentry's point of view and what should we expect in terms of timeline? 🙏

kahest:
@marcelofabri We bumped the priority because we have more data points that `hints` would be a useful addition for some use cases. We don't plan to tackle this within the next few weeks at the moment so don't have a specific timeline, but we'd want to land this within the next months.

sebastienboisvert:
[...]

Issue #1460, opened 2021-11-17
feat: Add Attachments in beforeSend


philipphofmann:
Currently, adding attachments in beforeSend is not possible, because the Cocoa SDK doesn't have hints as for example, [Python](https://docs.sentry.io/platforms/python/configuration/filtering/#event-hints) does. Either we add hints or the SDK should offer another way of adding attachments in beforeSend. Furthermore, it would be nice to also drop attachments in beforeSend.

This came up in https://github.com/getsentry/sentry-cocoa/issues/1459. A user wants to add an attachment to an OOM event.

philipphofmann:
Hints could be its own class extending a dictionary with methods for adding an attachment. Before implementing this we still need to investigate the use cases of hints in the Cocoa SDK. Look at Python and Android.

How this could look like with hints:
```Swift
SentrySDK.start { options in
    options.beforeSend { event, hint in
       hint.addAttachment(Attachment())
       hint.removeAttachment()
       hint.attachments
    }
}
```

eric:
How can we add attachments to any sort of crashes at all? We have logs that would be helpful in identifying why a crash happened and can't figure out how to send it.

tmm1:
What if we use the following existing api, what is the expected behavior?

```
  [SentrySDK configureScope:^(SentryScope * _Nonnull scope) {
    [scope addAttachment:...];
  }];

**Comment by philipphofmann** [comment]
@tmm1, your approach could work, but `SentrySDK.start` installs the `SentryCrashIntegration`, which then, on a background thread, will send the crash report. So it could work in most cases, but there is no guarantee. So you could do
```
SentrySDK.start { options in
    // ...
    
    options.onCrashedLastRun = { event in
        SentrySDK.configureScope { scope in
            scope.clearAttachments()
        }
    }
}

SentrySDK.configureScope { scope in
    scope.addAttachment(...)
}
```

philprime:
@itaybre I guess this issue can now be closed, right?

PR #3896, opened 2024-04-24
Add ability to set attachments on a sentry event

izackp:
## :scroll: Description

This just exposes a property on the event which will allow the user to add attachments to be sent.

This is so I can add log data along with crashes. Its incredibly necessary in resolving not so obvious bugs.

It's currently live in our app and seems to work.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/Public/SentryEvent.h
+++ b/Sources/Sentry/Public/SentryEvent.h
@@ -6,7 +6,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 @class SentryThread, SentryException, SentryStacktrace, SentryUser, SentryDebugMeta, SentryContext,
-    SentryBreadcrumb, SentryId, SentryMessage, SentryRequest;
+    SentryBreadcrumb, SentryId, SentryMessage, SentryRequest, SentryAttachment;
 
 NS_SWIFT_NAME(Event)
 @interface SentryEvent : NSObject <SentrySerializable>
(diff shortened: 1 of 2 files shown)


philipphofmann:
While this might work in your fork, we don't intend to add attachments to the event payload. Attachments are a separate thing. They get sent via extra envelope items. What is your use case for this @izackp?

If you only want to attach an attachment to a single event, you can use the local scope as mentioned in the [docs](https://docs.sentry.io/platforms/apple/enriching-events/attachments/):
```
SentrySDK.captureMessage(message: "my message", block: { scope in
	scope.addAttachment(fileAttachment)
})
```

izackp:
I want attachments on crash reports. Exposing a property allows me to add logs/attachments in beforeSend very easily. 

I looked at the proposed solution here and it doesn't seem to work:
https://github.com/getsentry/sentry-cocoa/issues/1460#issuecomment-1085515472

I guess what you really want implemented are 'hints' which are in the java version (I believe).. so I suppose this won't get accepted. 

```
options.beforeSend = { [weak self] (newEvent:Sentry.Event?) in
    guard let myEvent = newEvent else { return newEvent }
    if (myEvent.exceptions == nil && myEvent.error == nil) {
        return newEvent!;
    }
    //if (myEvent.isAppHangEvent) { return newEvent; }
    
    newEvent?.attachments = self?.attachments().map({
        return Sentry.Attachment(data: $0.data, filename: $0.fileName, contentType: $0.mimeType)
    })
    return newEvent
}
```

philipphofmann:
Yes, exactly. We need hints (https://github.com/getsentry/sentry-cocoa/issues/2325) for this. We don't want attachments to the event payload.

philipphofmann:
I'm closing this because, as mentioned here: https://github.com/getsentry/sentry-cocoa/pull/3896#issuecomment-2078810659, we will add hints at some point.

Issue #3897, opened 2024-04-24
Intelligent flakiness mitigation


armcknight:
### Description

We've had a persistent problem with flaky tests for a long time. 

One process we've tried using is to disable a flaky test with a PR, and open an issue to fix it. This has not eradicated the issues, and doesn't scale that well.

One automation we've attempted in the past is to allow up to a certain number of retries of any test invocation that is known to contain flakes. It was recently removed: https://github.com/getsentry/sentry-cocoa/pull/3660

However, we have unit and UI tests that basically always contain at least one flake. This requires each PR author to check the test run, read what failed, decide if it's a real failure related to their changes, and use the web UI to initiate a retry of the failed tests, and then wait for that round to complete again. This process adds hours to each PR.

I propose automating that manual process by adding the retry strategy back into the mix with https://github.com/getsentry/sentry-cocoa/pull/3881 and https://github.com/getsentry/sentry-cocoa/pull/3883.

Then, we should write some automation that will extract the name of any test that did not fail in every retry, where more than one test run was actually required, and report that in a PR comment so we can more easily perform our current flaky test process of disabling it and filing an issue to fix it, vs having to dive into the GHA logs for each one. 

Even that could potentially be automated, since we can open github issues with their API, and skipping a test is just inserting a line of text into the .xcscheme file. We could also pipe the event to slack so we are aware of it happening there, too.

kahest:
Thanks @armcknight, I agree with a lot here. A few thoughts/opinions:
- if we run these tests, they should be green, we shouldn't need to ignore failed checks when merging PRs
  - do we need to run these tests on every PR? IMO there should be no reason not to 
  - we have flaky tests, and the current process doesn't work well in surfacing/eliminating them
  - as a stopgap (!), it's IMO valid to add auto-retry (and keep manually disabling flaky tests as needed)
- we definitely need to iterate on that short to medium term
  - we're already planning to spend some time on improving CI/flakiness in the next few weeks/ in Q2
  - we should automatically surface fails/retries - e.g. in PR comments as suggested above
  - can we also capture fails/retries as errors in https://sentry.sentry.io/issues/?project=5899451? & we can setup threshold alerts and send to #team-cocoa etc.
- whatever we do, we will iterate

armcknight:
I like the idea to send these to sentry.sentry.io so we can get slack alerts, that would be nicer than rebuilding a slack alerter from scratch in the test scripts.

[...]

PR #3903, opened 2024-04-25
test: Add rerunFlakyTest method

philipphofmann:
Add a utility method to rerun a flaky test and only let the test fail if it fails two times in a row. 

This is an alternative to https://github.com/getsentry/sentry-cocoa/pull/3881,  https://github.com/getsentry/sentry-cocoa/pull/3881, and could fix https://github.com/getsentry/sentry-cocoa/issues/3897  and has advantages over rerunning all tests. First, it's faster as it only runs the flaky tests in a row instead of all tests. This will surface new flaky tests, as you must explicitly use this method to mark flaky tests. If you agree, @brustolin and @armcknight, I would add an update to our internal development docs on handling flaky tests. 

#skip-changelog

### DIFF
--- a/Sentry.xcodeproj/project.pbxproj
+++ b/Sentry.xcodeproj/project.pbxproj
@@ -121,6 +121,8 @@
 		62C316812B1F2E93000D7031 /* SentryDelayedFramesTracker.h in Headers */ = {isa = PBXBuildFile; fileRef = 62C316802B1F2E93000D7031 /* SentryDelayedFramesTracker.h */; };
 		62C316832B1F2EA1000D7031 /* SentryDelayedFramesTracker.m in Sources */ = {isa = PBXBuildFile; fileRef = 62C316822B1F2EA1000D7031 /* SentryDelayedFramesTracker.m */; };
 		62C3168B2B1F865A000D7031 /* SentryTimeSwiftTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 62C3168A2B1F865A000D7031 /* SentryTimeSwiftTests.swift */; };
+		62D4AC1D2BDA876000AEC0AE /* RerunFlakyTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 62D4AC1C2BDA876000AEC0AE /* RerunFlakyTest.swift */; };
+		62D4AC1F2BDA88D500AEC0AE /* RerunFlakyTestTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 62D4AC1E2BDA88D500AEC0AE /* RerunFlakyTestTests.swift */; };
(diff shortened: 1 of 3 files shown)


armcknight:
Good idea in general, but I don't think this is the way to do it, for a couple reasons:

1. we'll have to insert more code into each flaky test case
2. there'll be no easy way to centrally identify the set of flakes. you'll have to grep
3. this is more code (and tests!) we have to maintain
4. this is already built into `xcodebuild` with `-retry-tests-on-failure` and `-test-iterations number`

Also, this doesn't really fix #3897; it would be what comes after it. That issue is more about minimizing the impact of a priori unknown flakes, and identifying them. The general principle you propose is reasonable for running flaky tests after they've been identified.

I would propose we have a separate branches in xcode-test.sh that skips known flakes for normal runs, and then runs only flakes with the aforementioned `xcodebuild` options to allow rerunning them multiple times.

But, I still also think it'd be nice to have automatic flake detection, as I describe in #3897.

philipphofmann:
> That issue is more about minimizing the impact of a priori unknown flakes, and identifying them

That's the goal, yes. Instead of completely disabling the flaky tests, we could use the rerunFlakyTest method to keep them running. 

> 1. we'll have to insert more code into each flaky test case
> 2. there'll be no easy way to centrally identify the set of flakes. you'll have to grep
> 3. this is more code (and tests!) we have to maintain

I don't mind the bit of extra code and potential maintenance. I prefer a simple grep over using the Xcode test selector to find the disabled tests. I think that https://github.com/getsentry/sentry-cocoa/issues/3897 is more complex and harder to maintain than the rerunFlakyTest method. What do you think @brustolin?

> this is already built into xcodebuild with -retry-tests-on-failure and -test-iterations number

This approach reruns new flaky tests, which means new flaky tests would stay unnoticed. I think we need to get back to a point where you stop and take a step back when you see a failing unit test in CI and not just assume it's a flaky test.

brustolin:
The proposal in this PR is too "hacky" IMO. 
I also dont think that re-running multiple times is the solution.

> I think we need to get back to a point where you stop and take a step back when you see a failing unit test in CI and not just assume it's a flaky test.

I think we either do this or do nothing.

philipphofmann:
Yes, disabling everything is event better. @armcknight already did that. Let's be more strict again and disable test flakes https://github.com/getsentry/sentry-cocoa/pull/3904.

I'm going to close this for now.

armcknight:
> I think we need to get back to a point where you stop and take a step back when you see a failing unit test in CI and not just assume it's a flaky test.

As a general principle, yes, but it's usually very obvious when a test is a flake. Two big red flags are when something fails that's not remotely involved with the code that changed in the diff, and when it is a test that has failed and then passed on retry many times over a long period.

I wouldn't tell you that you need to stop and take step back to reevaluate whether this was really a flake or not: https://github.com/getsentry/sentry-cocoa/pull/3909. It's obvious it was a flake.

PR #3911, opened 2024-04-26
fix: Start session replay without a UIWindow

brustolin:
## :scroll: Description

If there is no UIWindow available during SDK initialization, wait for a UISceneDidActivateNotification

_#skip-changelog_

Unit Tests

## :pencil: Checklist


Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -68,6 +81,21 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
         expect(SentrySDK.currentHub().trimmedInstalledIntegrationNames().count) == 1
         expect(SentryGlobalEventProcessor.shared().processors.count) == 1
     }
+    
+    func testWaitForNotificationWithNoWindow() {
+        uiApplication.windowsMock = nil
+        startSDK(sessionSampleRate: 1, errorSampleRate: 0)
+        
+        guard let sut = SentrySDK.currentHub().installedIntegrations().first as? SentrySessionReplayIntegration else {
+            fail("Did not installed replay integration")
```
philipphofmann, line 90: `m`: That looks like it could get brittle because what if the order of the installed integration changes? I think filtering for one integration to be the `SentrySessionReplayIntegration` would be a better approach.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -68,6 +81,21 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
...
+            return
+        }
+        
+        expect(Dynamic(sut).sessionReplay.asObject) == nil
+        uiApplication.windowsMock = [UIWindow()]
+        NotificationCenter.default.post(name: UIScene.didActivateNotification, object: nil)
```
philipphofmann, line 96: `m`: Please use the `TestNSNotificationCenterWrapper` if possible because the NotificationCenter lead to flaky tests in the past, see:

https://github.com/getsentry/sentry-cocoa/blob/c8dbe735adc49b67a92bacb74930d826a23ca7ee/Sources/Sentry/include/SentryNSNotificationCenterWrapper.h#L5-L13

philipphofmann (commented):
Shouldn't we add a Changelog entry for this?

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
brustolin: If we change the order of the integrations in the test (it should not happen) this will immediately fail and then we fix the integration index.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
brustolin: I tried, and it was a nightmare. I prefer to leave it like this.

brustolin:
> Shouldn't we add a Changelog entry for this?

We haven't  release session replay yet, which means this is still under session replay entry in the changelog.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
philipphofmann: Up to you. I'm also fine with the way it is.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
philipphofmann: Why was it a nightmare? What was the problem?

brustolin: Many SIGNCRASH without a reason or proper stack trace.

The SessionReplayIntegration retrieves the notification wrapper from the Dependency Container, which is used by many other classes. After attempting to change the wrapper within the container, other objects were trying to remove observers from the wrapper during clean up, and the Invocation class was trying to access disposed objects.

philipphofmann: > The SessionReplayIntegration retrieves the notification wrapper from the Dependency Container, which is used by many other classes.

That's a major reason why I always advocate for passing in dependencies via constructors and not using the dependency container, which is global state and has side effects. Passing in dependencies via the constructor minimizes side effects. Anyways, we don't need to fix it in this PR.

philipphofmann (approved):
LGTM

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
brustolin: Integrations cannot have objects passed in the init function.
The problem here is with `Invocation` class, not the `SentryDependencyContainer`

philipphofmann: > Integrations cannot have objects passed in the init function.

Yes, I know, but most likely other classes can.

philipphofmann: > The problem here is the Invocation class, not SentryDependencyContainer

I think it's a mixture of both. So yes, the invocation class should use weak references in that case, and we should less of the `SentryDependencyContainer`. IMO.

Issue #3775, opened 2024-03-20
new iOS App Store Privacy Guidelines issue (System boot time APIs usage)


jtabraham123:
### Problem Statement

iOS App store guidelines that take in effect May 1st prevent applications from making calls to system boot time apis (like mach_absolute_time() and systemUptime) without declaring a reason in a privacy manifest file https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api. mach_absolute_time() is used in SentryTime.mm line 30. Their "acceptable reasons for using System Boot Time apis" list make it seem like you are not allowed to send this information without the user explicitly accepting they consent to this information being sent in a bug report. I dont want my app to make the user accept a message everytime I want to send something to sentry. You are allowed to send it if "There is an exception for information about the amount of time that has elapsed between events that occurred within the app, which may be sent off-device." But im not sure if you can send the boot time directly, just using it to calculate the elapsed time and sending that seems to be allowed

### Solution Brainstorm

- Give the option to the user to disable the sending of boot time in the options object in SentrySDK.start {}
- remove the api call? - seems like this would be harder to do

### Are you willing to submit a PR?

Yes I can submit a quick PR for this, it seems semi easy to fix

philipphofmann:
Thanks for bringing this up @jtabraham123 😃.

We already declared 35F9.1 in our `PrivacyInfo.xcprivacy`:

https://github.com/getsentry/sentry-cocoa/blob/fa409feccbfc87427acf396bfce16cac97df3640/Sources/Resources/PrivacyInfo.xcprivacy#L55-L60

[Apple docs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278394):
> 35F9.1
> Declare this reason to access the system boot time in order to measure the amount of time that has elapsed between events that occurred within the app or to perform calculations to enable timers.
> 
> Information accessed for this reason, or any derived information, may not be sent off-device. There is an exception for information about the amount of time that has elapsed between events that occurred within the app, which may be sent off-device.

We use `systemUptime` and `mach_absolute_time()` only for events that occurred within the app. In that case, we are allowed to send this information off-device. Therefore, we can keep using these APIs.

johnhatvani:
@philipphofmann is there a timeline for when 8.22.3 will be released? (from what I can see these changes were captured in 8.22.3-beta)

brustolin:
[...]

PR #3912, opened 2024-04-26
fix: dont add boot time to enriched scopes

armcknight:
Fixes an issue raised after reopening #3775, to avoid sending boot time off device, which is explicitly forbidden by Apple.

Still searching for an appropriate place to put a test for this. No test failed after making the change, and we have lots of test files for crash reporting, so it's not obvious to me ATM.


bruno-garcia (approved):
to unblock

bruno-garcia:
We could add a test later asserting this property isn't in a captured error. In the test we can expand the reasoning.

PR #3913, opened 2024-04-27
feat(profiling): continuous profiler basic implementation

armcknight:
For #3555; #skip-changelog

- further refactor the current profiling implementation in order to make space for the new one as a parallel to the legacy logic
- carry over some basic functionality from the legacy profiler controls so it can run
- create a test case that starts and stops it to ensure that it at least doesn't crash

Starting and stopping automatically on given events, including launch profiling, will be handled in a subsequent PR.

This extracts the remaining class functions from SentryProfiler.mm to a new class SentryLegacyProfiler, to parallel the previously created and stubbed SentryContinuousProfiler. 

All that then remains of SentryProfiler are the instance methods. Both the legacy and continuous classes will use an instance of SentryProfiler and control it with their respective specified behaviors. SentryProfiler is now initialized with an enum to specify whether it is running in legacy or continuous mode. It will use this to e.g. decide what to do when its timer expires (stop profiling in legacy, send a chunk in continuous–this part is also left for a later PR).

Also renamed SentryProfilerSwiftTests.swift to SentryLegacyProfilerTests and created a new SentryContinuousProfilerTests class, and factored out the fixture to a separate file in SentryProfileTestFixture.


brustolin (approved):
Looks good for me. Thanks!!

Sources/Sentry/Profiling/SentryLegacyProfiler.mm
```diff
@@ -0,0 +1,92 @@
...
+namespace {
+/** @warning: Must be used from a synchronized context. */
+std::mutex _unsafe_gLegacyProfilerLock;
+
+/** @warning: Must be used from a synchronized context. */
+SentryProfiler *_Nullable _unsafe_gLegacyProfiler;
+
+/** @warning: Must be called from a synchronized context. */
+BOOL
+_unsafe_isRunning(void)
```
philipphofmann, line 22: `l`: Maybe you could rename this and other unsafe prefixes to `non_synched_isRunning` or `call_from_synched` or something similar because unsafe doesn't reflect that you have to call it from a synchronized context.

Sources/Sentry/include/SentryLegacyProfiler.h
```diff
@@ -0,0 +1,35 @@
...
+
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+
+#    import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class SentryId;
+
+@interface SentryLegacyProfiler : NSObject
```
philipphofmann, line 11: `l`: What if this persists until we have another new profiling mode and the Continuous also becomes a legacy? Then we have to name this SentryLegacyLegacyProfiler, and SentryContinuousProfiler becomes SentryLegacyProfiler 😄 . Maybe we should call this `SentryTraceProfiler,` so its name reflects that it requires a traceId to work. But we can also keep it as is.

Sources/Sentry/SentryProfiler.mm
```diff
@@ -87,7 +84,10 @@ - (instancetype)init
 #    endif // SENTRY_HAS_UIKIT
 
     [self start];
-    [self scheduleTimeoutTimer];
+
+    if (mode == SentryProfilerModeLegacy) {
+        [self scheduleTimeoutTimer];
+    }
```
philipphofmann, line 90: `m`: What's stopping us from moving the `scheduleTimeoutTimer` logic to the `SentryLegacyProfiler`. Then, this class wouldn't even know that there are different profiling modes, and we could ditch the `SentryProfilerMode` enum.

Sources/Sentry/SentryTracer.m
```diff
@@ -193,7 +193,7 @@ - (instancetype)initWithTransactionContext:(SentryTransactionContext *)transacti
     if (_configuration.profilesSamplerDecision.decision == kSentrySampleDecisionYes
         || sentry_isTracingAppLaunch) {
         _internalID = [[SentryId alloc] init];
-        if ((_isProfiling = [SentryProfiler startWithTracer:_internalID])) {
+        if ((_isProfiling = [SentryLegacyProfiler startWithTracer:_internalID])) {
```
philipphofmann, line 196: I really like this refactoring approach. Almost no changes in the SentryTracer so far, which minimizes the risk of bugs for the existing implementation of the SentryTracer 💯

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
```diff
@@ -0,0 +1,53 @@
...
+private extension SentryContinuousProfilerTests {
+    func performContinuousProfilingTest() throws {
+        SentryContinuousProfiler.start()
+        
+        XCTAssert(SentryContinuousProfiler.isCurrentlyProfiling())
+        
+        try addMockSamples()
+        SentryContinuousProfiler.stop()
+        
+        XCTAssertFalse(SentryContinuousProfiler.isCurrentlyProfiling())
```
philipphofmann, line 35: `m`: Any reason you can't use Nimble assertions here?

Sources/Sentry/Profiling/SentryLegacyProfiler.mm
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    sentry_trackProfilerForTracer(_unsafe_gLegacyProfiler, traceId);
+    return YES;
+}
+
++ (BOOL)isCurrentlyProfiling
+{
+    std::lock_guard<std::mutex> l(_unsafe_gLegacyProfilerLock);
+    return [_unsafe_gLegacyProfiler isRunning];
```
philipphofmann, line 57: `l`: reusing the existing method:
```suggestion
    return _unsafe_isRunning();
```

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
```diff
@@ -0,0 +1,76 @@
...
+/** @warning: Must be called from a synchronized context. */
+BOOL
+_unsafe_isRunning(void)
+{
+    return _unsafe_gContinuousCurrentProfiler != nil &&
+        [_unsafe_gContinuousCurrentProfiler isRunning];
+}
+} // namespace
+
+@implementation SentryContinuousProfiler
```
philipphofmann, line 27: 💯 that you duplicated the implementation of `SentryContinuousProfiler` and `SentryLegacyProfiler`. Although we could add some abstractions and refactor them to have some base implementation, I prefer duplicating them because there are small differences between them, and the code is way easier to read.

philipphofmann (approved):
LGTM, with mostly nitpicks.

Sources/Sentry/Profiling/SentryLegacyProfiler.mm
armcknight: Yeah that's fair!

Sources/Sentry/include/SentryLegacyProfiler.h
armcknight: We can do that, although I hope continuous profiling is the last iteration we need :D

Sources/Sentry/SentryProfiler.mm
armcknight: Yes I did consider that. Nothing is stopping it, it should be fine enough to move the timer to a static object in the two control classes.

Sources/Sentry/Profiling/SentryLegacyProfiler.mm
armcknight: Thanks for catching this. Honestly we don't even need the nil check on `_unsafe_gLegacyProfiler` since it's safe to send a message to `nil` and it would return false in that case, so i might simplify this flow.

Sources/Sentry/Profiling/SentryLegacyProfiler.mm
armcknight: I went with a prefix of `_threadUnsafe_`

Sources/Sentry/include/SentryLegacyProfiler.h
armcknight: So I'm going to do this in a separate PR, because I have a few stacked at this point and resolving the conflicts through each will be annoying. https://github.com/getsentry/sentry-cocoa/pull/3952

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
armcknight: Habit

philipphofmann: Okay, then, let's revisit this. I like them a lot. The most important thing is that we all on the Cocoa SDK agree on using Nimble or not. Let's discuss this in Slack.

philipphofmann (approved):
Again LGTM

Issue #3702, opened 2024-03-04
Data race in -[SentryFramesTracker displayLinkCallback] at 0x10cd42f70


qweluke:
### Platform

iOS

### Environment

Production, Develop

### Installed

Swift Package Manager

### Version

8.20.0

### Did it work on previous versions?

No

### Steps to Reproduce

After enabling `thread sanitizer` i see 3 different data race errors in the xcode.

### Expected Result

There should be no `data race` error when a `thread sanitizer` is enabled

### Actual Result

<img width="1667" alt="image" src="https://github.com/getsentry/sentry-cocoa/assets/193701/85300218-f692-4b0b-b537-1b9f76ce1208">

<img width="1659" alt="image" src="https://github.com/getsentry/sentry-cocoa/assets/193701/96b1cbc5-bba0-4185-912a-f1e8ffe8bc4b">

brustolin:
Thanks @qweluke for reporting this. We will fix this, but I can give you an ETA right now.

philipphofmann:
The `displayLinkCallback` should be called on the main thread. I wonder how we get data races in there. I can easily reproduce this problem by activating the thread sanitizer for our iOS-Swift sample app. Thanks for reporting this, @qweluke.

philipphofmann:
The data races are false positives. We are going to disable them with https://github.com/getsentry/sentry-cocoa/pull/3922. For an explanation, visit https://github.com/getsentry/sentry-cocoa/pull/3922.

PR #3922, opened 2024-04-30
fix: Ignore FramesTracker thread data races

philipphofmann:
## :scroll: Description

The ThreadSanitizer reports data races for the SentryFramesTracker, which we can ignore.

I defined a macro for ignoring thread sanitizer warnings directly in the code. I will open another PR to replace the ThreadSanitizer.sup by ignoring the warnings directly in the code.

Fixes GH-3702

Running the iOS-Swift project with the thread sanitizer enabled.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Ignore SentryFramesTracker thread sanitizer data races (#3922)
+
 ## 8.25.0
 
 ### Features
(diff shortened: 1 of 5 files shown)


brustolin (approved):
Im ok with this solution, just wondering if someone could see this and say that we're hiding problems in our SDK.

philipphofmann:
> Im ok with this solution, just wondering if someone could see this and say that we're hiding problems in our SDK.

@brustolin, that's why I added the explanation: 

> This class ignores a couple of methods for the thread sanitizer. We intentionally accept several data races in this class, a decision that is driven by the fact that the code always writes on the main thread. This approach, while it may not provide 100% correct frame statistic for background spans, significantly reduces the overhead of synchronization, thereby enhancing performance.

Do you still believe users could think we're hiding problems?

brustolin:
> Do you still believe users could think we're hiding problems?

There will be always someone out there to complain. 😁
Lets merge this!!

PR #3923, opened 2024-04-30
ref: Ignore more thread sanitizer warnings in code

philipphofmann:
Ignore the thread sanitizer in the code where possible, which has the advantage of knowing a method is ignored when reading the code and not jumping to the ThreadSanitizer.sup file. Furthermore, use the ThreadSanitizer.sup file not only for tests but for the Run Xcode schemes.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/3922.

#skip-changelog


SentryTestUtils/TestSentryDispatchQueueWrapper.swift
```diff
@@ -13,8 +16,12 @@ public class TestSentryDispatchQueueWrapper: SentryDispatchQueueWrapper {
     public var dispatchAsyncInvocations = Invocations<() -> Void>()
     public var dispatchAsyncExecutesBlock = true
     public override func dispatchAsync(_ block: @escaping () -> Void) {
-        dispatchAsyncCalled += 1
-        dispatchAsyncInvocations.record(block)
+        
+        dispatchAsyncLock.synchronized {
```
philipphofmann, line 20: This was a data race happening frequently when running the tests and it was a small easy fix. That's why I added it in this PR.

brustolin (approved):

PR #3926, opened 2024-04-30
fix: reportFullyDisplayed data race

philipphofmann:
## :scroll: Description

Fix a data race when calling reportFullyDisplayed from a background thread by synchronizing the call to the main queue.

This came up when enabling the thread sanitizer for the iOS-Swift sample app and navigating to the LoremIpsum screen.

CleanShot 2024-04-30 at 17 36 56@2x

Unit tests and thread sanitizer in the iOS-Swift sample app.

## :pencil: Checklist


Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -38,10 +40,12 @@ @implementation SentryTimeToDisplayTracker {
 
 - (instancetype)initForController:(UIViewController *)controller
                waitForFullDisplay:(BOOL)waitForFullDisplay
+             dispatchQueueWrapper:(SentryDispatchQueueWrapper *)dispatchQueueWrapper
```
brustolin, line 43: m: I believe its time for us to talk about this. If "every" class need this, having this in the initializer for every class looks like a "code smell", every time we add a new dependency in the constructor, we need to go back and refactor every other class that depends on it.

All the changes that were added to `Sources/Sentry/SentryUIViewControllerPerformanceTracker.m` in this PR are really unnecessary.

I would write a notion page about this but I guess we could use a PR to discuss it.
My suggestion its to have a convenient initializer that get the dependencies from the `SentryDependencyContainer` and a full initializer that is used for tests where you can use test wrappers. 
This goes for DispatchQueueWrapper, CurrentDataProvider, Random or FileManager wrapper.

armcknight: I agree this is a code smell. An alternative is to use SentryDependencyContainer as an init parameter on each class instead of treating it as a singleton object. That way you get the configurable initialization and expandable dependencies without modifying call sites, while not having a more confusing singleton flow with possible state corruption in tests. Then we could have some vendor that builds readymade instances of it for tests. WDYT?

brustolin: Personally, I like singletons. I would even stop using SentryDependencyContainer and have a singleton for each of these basic wrappers used everywhere. That would reduce coupling.

I think adding SentryDependencyContainer as an init increases coupling. In the future, if we really plan to split Sentry into modules, it would make it harder. So, I prefer not to go this route.

philipphofmann: Thanks for bringing this up @brustolin. I think we shouldn't let this discussion block this PR. I created a [Notion page](https://www.notion.so/sentry/Dependency-Injection-Discussion-d1509755080245c280d4349b1da69237) to discuss our dependency injection approach.

brustolin (approved):

Issue #3893, opened 2024-04-24
Create profile envelope item on a background thread


philipphofmann:
### Description

Currently, the tracer creates the profile envelope item, which calls slicing the profiling data on the calling thread of `SentryTracer.finishInternal` here https://github.com/getsentry/sentry-cocoa/blob/c02bb1f4252fce2f69a5a6e36bde70f0211663fb/Sources/Sentry/SentryTracer.m#L627-L644

When the profile lasts for a few seconds, this can significantly noticeably block the main thread. With https://github.com/getsentry/sentry-cocoa/issues/3826 the captureTransaction in the hub already happens on a background thread. Moving `createProfilingEnvelopeItemForTransaction` to a background thread is a bit trickier, because if the tracer gets deallocated in the meantime, it might remove the profiling payload in dealloc:
https://github.com/getsentry/sentry-cocoa/blob/c02bb1f4252fce2f69a5a6e36bde70f0211663fb/Sources/Sentry/SentryTracer.m#L208-L215

While we can somehow fix this problem with the current solution, continuous profiling (https://github.com/getsentry/sentry-cocoa/issues/3555) will change our current approach, and the solution could get outdated. @armcknight, does it make sense to consider this point while moving towards continuous profiling (https://github.com/getsentry/sentry-cocoa/issues/3555)?

armcknight:
It does make sense to consider for the new implementation, definitely. For legacy profiling, I think it should be fine to move things to a bg thread once we have the transaction created by the tracer. There may be a few accesses of the tracer via the transaction from profiler code, but those could be replaced with parameters in the profiling methods/functions that are called so the values can be captured before offloading to the bg thread, risking deallocation.

philipphofmann:
[...]

PR #3932, opened 2024-05-02
feat(profiling): continuous profile serialization schema

armcknight:
#skip-changelog, for #3555

Implements the necessary schema changes for how continuous profiling chunks will be serialized, and their envelope structure. This was cribbed from the current functions that do the same for legacy profiles (`sentry_serializedProfileDataLegacy` and `sentry_profileEnvelopeItemLegacy`) to create `sentry_serializedContinuousProfileChunk` and `sentry_continuousProfileChunkEnvelope`.

Also implements the timer logic to transmit those envelopes every time it fires in `SentryProfiler.mm`.

Added assertions for the new schema to `SentryContinuousProfilerTests.swift`. Those tests will simulate sending multiple chunks per test, validating the contents of each, which have unique data per chunk to ensure the same data isn't carrying over, and that the profiler is still running after transmitting each until the profiler is explicitly stopped.


Sources/Sentry/Profiling/SentryProfilerSerialization.mm
```diff
@@ -196,9 +207,138 @@
...
+    profile[@"samples"] = _sentry_serializedSamplesWithRelativeTimestamps(
+        samples, startSystemTime, SentryProfilerModeContinuous);
+
+    payload[@"profile"] = profile;
+
+    payload[@"version"] = @"2";
+    const auto debugImages = [NSMutableArray<NSDictionary<NSString *, id> *> new];
+    for (SentryDebugMeta *debugImage in debugMeta) {
+        [debugImages addObject:[debugImage serialize]];
+    }
```
philipphofmann, line 240: `m`: We could significantly reduce the payload if we only add the debug images references in the samples payload; see `SentryDebugImageProvider.getDebugImagesForFrames`, but we don't have to do this in this PR.

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
```diff
@@ -196,9 +207,138 @@
...
+    if (debugImages.count > 0) {
+        payload[@"debug_meta"] = @ { @"images" : debugImages };
+    }
+
+    payload[@"chunk_id"] = [[[SentryId alloc] init] sentryIdString];
+    payload[@"profile_id"] = profileID.sentryIdString;
+    payload[@"environment"] = hub.scope.environmentString ?: hub.getClient.options.environment;
+    payload[@"release"] = hub.getClient.options.releaseName;
+    payload[@"platform"] = SentryPlatformName;
```
philipphofmann, line 249: `m`: Shouldn't we also add os and device data https://github.com/getsentry/sentry-cocoa/blob/4c0b15137001db76ed776f24dd430b9121a977ba/Sources/Sentry/Profiling/SentryProfilerSerialization.mm#L139-L152

If yes, I think we should extract this into a method, to reduce duplication.

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
```diff
@@ -196,9 +207,138 @@
...
+
+    // add the gathered metrics
+    auto metrics = serializedMetrics;
```
philipphofmann, line 252: `m`: The metrics code is a bit different from `sentry_serializedProfileDataLegacy`. Is this intended? If yes, why? If not, I think we should extract some method to get rid of some duplication instead of duplicating it.

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
```diff
@@ -196,9 +207,138 @@
...
+
+#    if SENTRY_HAS_UIKIT
+    const auto mutableMetrics =
+        [NSMutableDictionary<NSString *, id> dictionaryWithDictionary:metrics];
+    const auto slowFrames = sentry_sliceGPUData(gpuData.slowFrameTimestamps, startSystemTime,
+        endSystemTime, /*useMostRecentFrameRate */ NO, SentryProfilerModeContinuous);
+    if (slowFrames.count > 0) {
```
philipphofmann, line 259: `m`: Codecov complains quite a bit for uncovered lines in this method. Should we add tests for them?

Sources/Sentry/SentryProfiler.mm
```diff
@@ -232,6 +241,26 @@ - (BOOL)isRunning
...
+
+    const auto envelope = sentry_continuousProfileChunkEnvelope(
+        startSystemTime, endSystemTime, stateCopy, self.profilerId,
+        [self.metricProfiler serializeBetween:startSystemTime and:endSystemTime]
+#    if SENTRY_HAS_UIKIT
+        ,
+        self.screenFrameData
+#    endif // SENTRY_HAS_UIKIT
+    );
+    [SentrySDK captureEnvelope:envelope];
```
philipphofmann, line 261: `m`: Does it make sense to move this to a background thread to potentially unblock the calling thread, which could be the main thread to tackle https://github.com/getsentry/sentry-cocoa/issues/3893? Also, okay to do this in another PR.

Sources/Sentry/include/HybridPublic/SentryEnvelopeItemType.h
```diff
@@ -7,3 +7,4 @@ static NSString *const SentryEnvelopeItemTypeClientReport = @"client_report";
 static NSString *const SentryEnvelopeItemTypeProfile = @"profile";
 static NSString *const SentryEnvelopeItemTypeReplayVideo = @"replay_video";
 static NSString *const SentryEnvelopeItemTypeStatsd = @"statsd";
+static NSString *const SentryEnvelopeItemTypeProfileChunk = @"profile_chunk";
```
philipphofmann, line 10: `h`: We must add a data category related to this envelope item type or map this envelope item type to the data category profile so rate limiting and client reports work correctly. See https://github.com/getsentry/sentry-cocoa/blob/e072ad1b27ff5e6413edd2e4716a5ef0be4eee04/Sources/Sentry/SentryDataCategoryMapper.m#L35-L37

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
```diff
@@ -21,32 +21,125 @@ final class SentryContinuousProfilerTests: XCTestCase {
...
-        SentryProfilerMocksSwiftCompatible.appendMockBacktrace(to: state, threadID: threadMetadata.id, threadPriority: threadMetadata.priority, threadName: threadMetadata.name, addresses: addresses)
+    func assertValidData(expectedEnvironment: String, expectedAddresses: [NSNumber]?) throws {
+        let envelope = try XCTUnwrap(self.fixture.client?.captureEnvelopeInvocations.last)
+        XCTAssertEqual(1, envelope.items.count)
+        let profileItem = try XCTUnwrap(envelope.items.first)
+        XCTAssertEqual("profile_chunk", profileItem.header.type)
+        let data = profileItem.data
+        let profile = try XCTUnwrap(try JSONSerialization.jsonObject(with: data) as? [String: Any])
+
+        XCTAssertNotNil(profile["version"])
```
philipphofmann, line 84: `l`: Asserting for the expected version depending on continuous profiling or tracer based profiling would improve the tests.

philipphofmann (commented):
A couple of issues, but looking good already 👍

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
```diff
@@ -92,7 +103,7 @@
 }
 
 NSMutableDictionary<NSString *, id> *
-sentry_serializedProfileData(
+sentry_serializedProfileDataLegacy(
```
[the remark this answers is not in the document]
armcknight: I did rename it there, but it's stacked on top of this pr. it'll have to be either merged after this one, or combined into it (which I'd prefer not to do)

Sources/Sentry/Profiling/SentryProfilerDefines.h
```diff
@@ -0,0 +1,55 @@
+#import "SentryProfilingConditionals.h"
+
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+
+#    import "SentryDefines.h"
+#    import <Foundation/Foundation.h>
+
+typedef NS_ENUM(NSUInteger, SentryProfilerMode) {
+    SentryProfilerModeLegacy,
```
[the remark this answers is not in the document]
armcknight: See https://github.com/getsentry/sentry-cocoa/pull/3932#discussion_r1597527928

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
armcknight: This is good feedback. Maybe we send the full list of debug images in the first chunk, and then in subsequent chunks, send a list of images that have been added or removed since the last chunk. @phacops what do you think?

In any case I think this will need to be done in a subsequent PR, and any change is going to depend on a change in the backend as well, so we shouldn't hold up this PR for it. I've made a note in the [planning doc](https://www.notion.so/sentry/Continuous-Profiling-Cocoa-SDK-9d64a809ed2d487b89e4aa0358bfac18?pvs=4#c2725f3879c442729e8bd107dd56ea8c).

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
armcknight: we are no longer sending those in profile payloads, they're going to get it elsewhere in the backend.

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
armcknight: Here are the differences:
- instead of using dictionary literals, which have led to crashes in the past, I went with NSMutableDictionaries and added the same fields with assignment
- the units are in milliseconds instead of nanoseconds (this change applies to all of continuous profiling durations)

Sources/Sentry/SentryProfiler.mm
armcknight: Yes, I'm definitely on board to do that, but agree it is appropriate to do so in another PR. ([link](https://www.notion.so/sentry/Continuous-Profiling-Cocoa-SDK-9d64a809ed2d487b89e4aa0358bfac18?pvs=4#5454422467754884b59f496ddaa77b0b))

Sources/Sentry/include/HybridPublic/SentryEnvelopeItemType.h
```diff
@@ -1,3 +1,5 @@
+// each item type must have a data category name mapped to it; see SentryDataCategoryMapper
```
philipphofmann, line 1: Thanks for the comment

Sources/Sentry/SentryDataCategoryMapper.m
```diff
@@ -11,6 +11,7 @@
 NSString *const kSentryDataCategoryNameAttachment = @"attachment";
 NSString *const kSentryDataCategoryNameUserFeedback = @"user_report";
 NSString *const kSentryDataCategoryNameProfile = @"profile";
+NSString *const kSentryDataCategoryNameProfileChunk = @"profile_chunk";
```
philipphofmann, line 14: Didn't know that data category exists already 💯 
https://github.com/getsentry/relay/blob/23a48f2c7142ab940d5e6e3d3289c087eba2fcf2/py/sentry_relay/consts.py#L30

philipphofmann (approved):
LGTM, thanks 👍

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
armcknight: Oh yes definitely after another app launch. I was only thinking about doing it between chunk envelopes in the same profile session (another app launch implies a different profile session)

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
armcknight: I'm beefing up the tests around this.

armcknight: I have another branch at the top of the stack where i was already adding more tests, I will do these there because these particular tests involve a lot of work around timestamps, and we just made a decision that's going to affect that (moving to using NSDate for all profiling timestamps), which I'd like to merge before writing the tests to avoid potentially doing a lot of annoying debugging twice.

PR #3937, opened 2024-05-05
feat(profiling): starting and stopping continuous profiler

armcknight:
for #3555; #skip-changelog

After some more discussion, we're only going to offer manually starting and stopping the profiler (besides app launch profiling starting it automatically before any app code can actually run; but that must at least be manually configured by customers by setting the SentryOption for it). 

We originally planned to automatically start/stop with session start/stop that respected the profiles sampling definitions, but we won't move forward with that right now. Manual start/stop will not respect profiling sample rates, calls will _always_ start the profiler if one isn't already running.

Closing the SDK _will_ stop any profiler that's running.

This cleans up anything that tied profiles to sessions, and adds more tests around various start/stop scenarios.


philipphofmann (approved):
LGTM

PR #3938, opened 2024-05-05
feat(profiling): continuous profiling from launch

armcknight:
for #3555; #skip-changelog

This PR makes the necessary changes to app launch profiling logic to work with the continuous profiler, since we'll no longer be using SentryTracer to manage starting the profiler, instead starting it directly.


Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.swift
```diff
@@ -0,0 +1,190 @@
...
+     * Test how combinations of the following options interact to ultimately decide whether or not to start the profiler on the next app launch..
+     * - `enableLaunchProfiling`
+     * - `enableTracing`
+     * - `enableContinuousProfiling` (always profiles regardless of sample rate or trace options)
+     * - `tracesSampleRate`
+     * - `profilesSampleRate`
+     * - `profilesSampler`
+     */
+    func testShouldProfileLaunchBasedOnOptionsCombinations() {
+        for testCase: (enableAppLaunchProfiling: Bool, enableTracing: Bool, enableContinuousProfiling: Bool, tracesSampleRate: Int, profilesSampleRate: Int, profilesSamplerReturnValue: Int, shouldProfileLaunch: Bool) in [
```
armcknight, line 104: This looks nuts, but we now have a ton of combinations that can affect whether or not a launch profile starts. I originally tried writing this with Quick BDD style tests, but the code became unreadable (i've left the intermediate commits, if you're curious). I figured this is better because it explicitly lists all the options that lead to a launch profile starting. This PR has a sizable diff, but this file alone is half of it.

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -140,6 +142,9 @@ - (void)framesTrackerHasNewFrame:(NSDate *)newFrameDate
         if (!_waitForFullDisplay) {
             [SentryDependencyContainer.sharedInstance.framesTracker removeListener:self];
 #    if SENTRY_TARGET_PROFILING_SUPPORTED
+            if (SentrySDK.options.enableContinuousProfiling) {
+                return;
+            }
             sentry_stopAndDiscardLaunchProfileTracer();
```
philipphofmann, line 148: `h`: I think we must not return here and also not below, as it could happen that with a new frame both the TTFD and TTID spans need to be finished. It would be great to add test for this. Instead it should be:
```suggestion
            if (!SentrySDK.options.enableContinuousProfiling) {
                sentry_stopAndDiscardLaunchProfileTracer();
            }

```

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -140,6 +142,9 @@ - (void)framesTrackerHasNewFrame:(NSDate *)newFrameDate
         if (!_waitForFullDisplay) {
             [SentryDependencyContainer.sharedInstance.framesTracker removeListener:self];
 #    if SENTRY_TARGET_PROFILING_SUPPORTED
+            if (SentrySDK.options.enableContinuousProfiling) {
```
philipphofmann, line 145: `m`: We already pass in `enableWaitForFullDisplay` via the constructor, which stems from the options. It would be great to follow this pattern.

philipphofmann (commented):
Thanks, almost LGTM 😸

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.swift
armcknight: I haven't heard that term before, thanks for the link. I'm not sure this is directly applicable in the way you think though. We don't have just one partition, but multiple, so we need to test all permutations. Just because one test case with `enableContinuousProfiling` set to true "covers" the lines of code in the implementation, it doesn't necessarily cover all the paths that traverse those lines of code. It's possible that in the future, someone puts an early return before those lines of code, giving an invalid result, due to one of the other options. Testing all the combinations protects against that. Am I missing something with the technique you mention?

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -95,6 +103,60 @@
...
+sentry_willProfileNextLaunch(SentryOptions *options)
+{
+    return sentry_shouldProfileNextLaunch(options).shouldProfile;
+}
+#    endif // defined(TEST) || defined(TESTCI) || defined(DEBUG)
+
+#    pragma mark - Exposed only to tests
+
+void
+_sentry_nondeduplicated_startLaunchProfile(void)
```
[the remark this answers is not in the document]
armcknight: i was trying to take the same approach as before when we changed `_sentry_unsafe` to `_sentry_threadUnsafe`... i'll admit this isn't the best name, but i was going for descriptiveness, and i also left a headerdoc explaining it.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -41,16 +47,18 @@
...
     SentrySamplerDecision *_Nullable profilesDecision;
 } SentryLaunchProfileConfig;
 
-NSString *const kSentryLaunchProfileConfigKeyTracesSampleRate = @"traces";
-NSString *const kSentryLaunchProfileConfigKeyProfilesSampleRate = @"profiles";
-
 SentryLaunchProfileConfig
 sentry_shouldProfileNextLaunch(SentryOptions *options)
 {
+    if (options.enableContinuousProfiling) {
```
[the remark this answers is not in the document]
armcknight: Good catch, this was a mistake

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.swift
```diff
@@ -0,0 +1,190 @@
...
+        sentry_manageTraceProfilerOnStartSDK(options, TestHub(client: nil, andScope: nil))
+        XCTAssert(appLaunchProfileConfigFileExists())
+        let dict = try XCTUnwrap(appLaunchProfileConfiguration())
+        XCTAssertEqual(dict[kSentryLaunchProfileConfigKeyContinuousProfiling], true)
+
+        _sentry_nondeduplicated_startLaunchProfile()
+        XCTAssert(SentryContinuousProfiler.isCurrentlyProfiling())
+    }
+    
+    func testTraceProfilerStartsWhenBothSampleRatesAreSet() {
```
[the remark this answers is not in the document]
armcknight: This isn't due to randomness but due to trying to separate the `dispatch_once` invoked function and running it multiple times in the same process, where i'm not properly cleaning up some static global state 🤔

Sources/Sentry/SentryTimeToDisplayTracker.m
armcknight: did this one and the other one

Sources/Sentry/SentryTimeToDisplayTracker.m
armcknight: why not go the other way and pass in the options instance?

armcknight: Here it is in case that's ok with you: https://github.com/getsentry/sentry-cocoa/pull/3989

armcknight: So, that PR has some failing tests, and I'm guessing they're due to the mismatch I mentioned. I think that's the right way to make this change, but I don't currently have the time to focus on that. I unstuck this branch by making the new boolean init param and passing it from `SentryUIViewControllerPerformanceTracker`, which was already grabbing the options from SentrySDK.options in its own init method. That would go away in the other PR. Please have a look at the failing tests there and see if there are some quick fixes you could help with!

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.swift
armcknight: Also maybe that's what I was trying to do with the nested BDD stuff, if you look at those intermediate commits, but it got confusing to know what had been set for a given final test result. I found it better to just have each combination explicitly listed.

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.swift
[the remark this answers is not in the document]
armcknight: I fixed this in clearTestState and added a test to ensure that the state is correct in the real world case.

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.swift
philipphofmann: I'm fine with keeping it as it is. We can discuss strategies to reduce the amount of test cases another time. We only need to change the tests when the requirements change. I'm OK with how it is right now.

Sources/Sentry/SentryTimeToDisplayTracker.m
philipphofmann: I'm confused why you went with the refactoring in https://github.com/getsentry/sentry-cocoa/pull/3989, while just passing in the argument similar to `enableWaitForFullDisplay` would be way less work. I would ditch https://github.com/getsentry/sentry-cocoa/pull/3989 right now. You can also keep the access to `SentrySDK.options.enableContinuousProfiling` in this PR, if you're in a hurry. Most important would be some tests.

philipphofmann (approved):
LGTM

Sources/Sentry/SentryTimeToDisplayTracker.m
philipphofmann: Ah I see, because `SentryUIViewControllerPerformanceTracker` is a singleton, and we would need to refactor it. The `SentryTimeToDisplayTracker` isn't a singleton and requires its dependencies via the constructor, so it's easier to test. Nevermind.

PR #3940, opened 2024-05-06
test: Remove clearTestState for ViewHierarchyTests

philipphofmann:
The SentryViewHierarchyTests only need to reset the SentryDependencyContainer and not reset the whole global test state.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/SentryViewHierarchyTests.swift
+++ b/Tests/SentryTests/SentryViewHierarchyTests.swift
@@ -37,7 +37,7 @@ class SentryViewHierarchyTests: XCTestCase {
 
     override func tearDown() {
         super.tearDown()
-        clearTestState()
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/SentryViewHierarchyTests.swift
```diff
@@ -37,7 +37,7 @@ class SentryViewHierarchyTests: XCTestCase {
 
     override func tearDown() {
         super.tearDown()
-        clearTestState()
+        SentryDependencyContainer.reset()
```
brustolin, line 40: m: What is the impact of this change? Will we need to be aware of which teardown api to use for every test now?

brustolin (approved):
Code looks good. CI is not happy. Probably flaky.

PR #3957, opened 2024-05-08
fix: Unable to find class: SentryCrashExceptionApplication

### LINKED ISSUE #3949: Mac app crashes on launch when using Sentry static framework
### Platform

macOS

### Environment

Production, Develop, TestFlight

### Installed

Swift Package Manager

### Version

8.25.0

8.21.0

### Steps to Reproduce

1. Create a new macOS app
2. Add Sentry as an SPM dependency
3. Change the Principal Class to `SentryCrashExceptionApplication`, [as described here](https://docs.sentry.io/platforms/apple/guides/macos/usage/#capturing-uncaught-exceptions-in-macos).
4. Run the app

See the attached sample project, which should reproduce the issue.

### Expected Result

The app should launch when run.

### Actual Result

The app crashes on launch. Nothing appears in the Xcode console, but in the macOS console, we see this:

```
Unable to find class: SentryCrashExceptionApplication, exiting
```

I suspect that because there is no reference to `SentryCrashExceptionApplication` in the code, the class was never linked in from the static library, and thus is not present at runtime. If I add any reference to `SentryCrashExceptionApplication` in the code, it launches correctly. For example, I can add this:

```swift
let x = SentryCrashExceptionApplication.self
print(x)
```

Note, though, that if this reference were optimized out by the compiler, then the app might crash again. (Hence the need to actually `print` it.)

It would be nice if Sentry itself could reference `SentryCrashExceptionApplication` on macOS, so that it is seen as a transitive dependency and thus would be linked in when other Sentry APIs are referenced. Alternatively, if referencing `SentryCrashExceptionApplication` in code is necessary, this should be documented.

I'd be willing to, but it's not clear what route the team wants to take

brustolin:
## :scroll: Description

Added a reference `SentryCrashExceptionApplication` inside the SDK to prevent compiler from stripping it.

closes #3949 

Sample

## :pencil: Checklist


Sources/Sentry/SentrySDK.m
```diff
@@ -196,6 +200,11 @@ + (void)startWithOptions:(SentryOptions *)options
     SENTRY_LOG_DEBUG(@"Configured options: %@", options.debugDescription);
 #endif // defined(DEBUG) || defined(TEST) || defined(TESTCI)
 
+#if TARGET_OS_OSX
+    // Reference to SentryCrashExceptionApplication to prevent compiler from stripping it
```
philipphofmann, line 204: Thanks for the comment 🥇

philipphofmann (approved):
LGTM

Issue #3956, opened 2024-05-08
Sentry gets deadlocked during offline startup


molind:
### Platform

iOS

### Environment

Production

### Installed

Swift Package Manager

### Version

8.25.0

### Did it work on previous versions?

8.21.0

### Steps to Reproduce

1. Switch device to the flight mode. 
2. SentrySDK.start inside `application(_: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil)`

It's not guaranteed to happen. But in crashlog we can follow to the cause.

Main thread inside the dispatch_once tries to access `[SentryCrashWrapper systemInfo]` and at the same time second thread locks [SentrySDK currentHub], calls `[SentryCrashWrapper systemInfo]` and waits for main thread to finish. But main thread can't finish because inside of that dispatch_once it's also tries to access [SentrySDK options]. 

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Incident Identifier: 82C9C585-8649-42DA-A0F8-62C3D085EF43
CrashReporter Key:   3c8d5e561b90c78666cc1a6eb487bbdce220b9e7
Hardware Model:      iPhone14,2
Process:             Guru [3540]
Path:                /private/var/containers/Bundle/Application/D308A662-48B8-45D6-8F68-AD0817F04B3C/Guru.app/Guru
Identifier:          com.bodunov.galileo
Version:             5.5.6 (9247)
AppStoreTools:       15E204
AppVariant:          1:iPhone14,2:15
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           com.bodunov.galileo [1445]

Date/Time:           2024-05-08 13:39:25.2179 +0200
Launch Time:         2024-05-08 13:39:05.0456 +0200
OS Version:          iPhone OS 17.4.1 (21E236)
Release Type:        User
Baseband Version:    3.50.04
Report Version:      104

Exception Type:  EXC_CRASH (SIGKILL)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Termination Reason: FRONTBOARD 2343432205 
<RBSTerminateContext| domain:10 code:0x8BADF00D explanation:scene-create watchdog transgression: app<com.bodunov.galileo(1455943A-E1D8-4E68-9A07-72D45E6103AB)>:3540 exhausted real (wall clock) time allowance of 19.89 seconds
[...]

PR #3959, opened 2024-05-08
Fixed deadlock during offline start.

molind:
## :scroll: Description

See details in #3956 

Sentry deadlocked during offline startup.

I've added sleep into the crash wrapper. And when background thread tried to send envelope it locked SentrySdk @synchronized(self).

```
- (NSDictionary *)systemInfo
{
    static NSDictionary *sharedInfo = nil;
    dispatch_once(&onceToken,
                  ^{
        sleep(5);
        sharedInfo = SentryDependencyContainer.sharedInstance.crashReporter.systemInfo;
    });
    return sharedInfo;
}
```

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentrySDK.m
+++ b/Sources/Sentry/SentrySDK.m
@@ -73,9 +73,6 @@ static NSDate *_Nullable startTimestamp = nil;
 + (SentryHub *)currentHub
 {
     @synchronized(self) {
-        if (nil == currentHub) {
-            currentHub = [[SentryHub alloc] initWithClient:nil andScope:nil];
-        }
         return currentHub;
     }
 }
--- a/Sources/Sentry/SentryHttpTransport.m
+++ b/Sources/Sentry/SentryHttpTransport.m
@@ -18,6 +18,7 @@
 #import "SentryNSURLRequest.h"
 #import "SentryNSURLRequestBuilder.h"
 #import "SentryOptions.h"
+#import "SentrySDK+Private.h"
 #import "SentrySerialization.h"
 #import "SentrySwift.h"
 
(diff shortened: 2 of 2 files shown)


Sources/Sentry/SentrySDK.m
```diff
@@ -73,9 +73,6 @@ + (void)initialize
 + (SentryHub *)currentHub
 {
     @synchronized(self) {
-        if (nil == currentHub) {
-            currentHub = [[SentryHub alloc] initWithClient:nil andScope:nil];
-        }
```
brustolin, line 78: h:  We cant remove this right now. It will break a lot of tests, we need to find another solution.

molind: Could you explain motivation behind this?

molind: If it'll try to send envelopes to that hub, they'll be just dropped. And transport should wait for a hub with a client to send something.
May be we could check scope, before enrich it.

```
if (scope)
        [SentryDependencyContainer.sharedInstance.crashWrapper enrichScope:scope];
```

But that's strange solution and it could happen again in different part of code. Sentry should try to send something only when SentrySDK start finishes. With changes from this PR it's guaranteed.

molind:
I've checked 8.21, when it worked without issues.

It creates SentryHub without enriching scope:

```
- (instancetype)initWithClient:(nullable SentryClient *)client
                      andScope:(nullable SentryScope *)scope
{
    if (self = [super init]) {
        _client = client;
        _scope = scope;
        _sessionLock = [[NSObject alloc] init];
        _integrationsLock = [[NSObject alloc] init];
        _installedIntegrations = [[NSMutableArray alloc] init];
        _installedIntegrationNames = [[NSMutableSet alloc] init];
        _crashWrapper = [SentryCrashWrapper sharedInstance];
        _errorsBeforeSession = 0;
    }
    return self;
}
```

And there is no systemInfo call and no deadlock.

Sources/Sentry/SentrySDK.m
brustolin: > Could you explain motivation behind this?

This was probably put in place to make writing tests easier. 
As you can see, all unit tests are broken right now. 
But I agree that we need to check if the SDK is initialized before starting to use the hub.

philipphofmann:
Thanks for all the context in this PR and in https://github.com/getsentry/sentry-cocoa/issues/3956, which helped us reproduce the issue. As this PR is only a workaround and doesn't fix the underlying issue, I opened https://github.com/getsentry/sentry-cocoa/pull/3970 to fix the underlying issue. We can close this.

PR #3960, opened 2024-05-08
ref: consolidate SentryThreadWrapper.onMainThread to SentryDispatchQueueWrapper.dispatchSyncOnMainQueue

armcknight:
I did this because there wasn't yet a way to mock `SentryThreadWrapper.onMainThread` and I need to do that for a test in https://github.com/getsentry/sentry-cocoa/pull/3937. In any case it's bothered me for a while that we have a few ways to do this, so this will be better going forward. I also removed another method under `SentryDispatchQueueWrapper.dispatchOnMainQueue` that just called through to `SentryDispatchQueueWrapper.dispatchSyncOnMainQueue`

#skip-changelog


armcknight:
This turns out to cause all kinds of weird test failures and crashes, so I'm going to see if I can change just the spot I need, instead of doing a general refactor.

PR #3963, opened 2024-05-09
chore: Move SentryDefine enums to swift

brustolin:
We can't use Objective-C enums from Swift as a function argument without breaking `TestUtil` compilation, but we can use Swift int enums from Objective-C.

Session Replay will need `SentryLevel`, so a move it to swift, alongside with it I also moved `SentryTransactionNameSource`, both from `SentryDefines.h`

_#skip-changelog_


brustolin:
So, this causes another problem.
Swift enums used in objc class properties or function parameters cannot be accessed from Swift.
This does not look good but I believe this is a step into more Swift code, until almost everything is swift.

armcknight (approved):
What's keeping this in draft vs ready for review? You want to move more things?

brustolin:
> What's keeping this in draft vs ready for review? You want to move more things?

After discovering the following problem 
> Swift enums used in objc class properties or function parameters cannot be accessed from Swift.

Im trying to figure it out if this is ok or not.

brustolin:
> Im trying to figure it out if this is ok or not.

So, we still need to do this. Using objc enums from Swift cause cocoapod static linking to fail.
IMO we will need to move more classes to Swift faster.

philipphofmann:
> IMO we will need to move more classes to Swift faster.

Yes, indeed 💯

philipphofmann (approved):
LGTM

PR #3965, opened 2024-05-09
ref: rename iOS-SwiftUITests -> iOS-Swift-UITests

armcknight:
This has bothered me for a while. I can never tell if I'm looking at a test target for iOS-SwiftUI, or a UI test target for iOS-Swift 🙃 

#skip-changelog


philipphofmann (approved):
LGTM

Issue #3899, opened 2024-04-24
Hang during initialization when SentrySDK.addBreadcrumb() is called from different thread


juraj-blahunka:
### Platform

iOS

### Environment

TestFlight

### Installed

Carthage

### Version

8.24.0

### Did it work on previous versions?

8.21.0

### Steps to Reproduce

The issue was only reproducible in TestFlight / AppStore builds, not in DEBUG. 
It was reproduced with 8.24.0 and 8.23.0. Didn't try 8.22.0. Not reproducible with 8.21.0.

1. Launch the app
2. Force-quit (kill) the app
3. Run the app and wait as it hangs

The SDK is initialized through the usual Swift call:
```
SentrySDK.start { options in ... }
```

The breadcrumb is logged from a different thread (synchronous dispatch queue)
```
private lazy var cookiesSyncQueue = DispatchQueue(label: "CookieManagerSyncQueue")

cookiesChangeStream = NotificationCenter.default
    .publisher(for: .NSHTTPCookieManagerCookiesChanged, object: rawStorage)
    .throttle(for: 0.1, scheduler: cookiesSyncQueue, latest: true)
    .sink {  _ in
        let breadcrumb = ...
        SentrySDK.addBreadcrumb(breadcrumb)
    }
```

**Other notes**
* Commenting out the breadcrumb call from the other dispatch queue works.
* Since it was reproduced in 8.23.0, but not in 8.21.0, this PR might be related https://github.com/getsentry/sentry-cocoa/pull/3698

### Expected Result

The app launches without a hang.

### Actual Result

The app hangs and gets killed by the watchdog after 20 seconds.

In the ips file, both threads hang with `-[SentryCrashWrapper systemInfo]` at the top of the trace. Here are the excerpts:

Main thread:
```
0   libsystem_kernel.dylib        	       0x1e266293c __ulock_wait2 + 8
1   libsystem_platform.dylib      	       0x1f6028ea8 _os_unfair_lock_lock_slow + 188
2   libobjc.A.dylib               	       0x1920138b8 objc_sync_enter + 20
3   kphone                        	       0x102ac5224 +[SentrySDK options] (in kphone) + 24 + 7918116
4   kphone                        	       0x102ad5508 -[SentryDependencyContainer crashReporter] (in kphone) + 88 + 7984392
5   kphone                        	       0x102a7e10c __32-[SentryCrashWrapper systemInfo]_block_invoke (in kphone) + 40 + 7627020
[...]

PR #3970, opened 2024-05-13
fix: Potential deadlock when starting the SDK

philipphofmann:
## :scroll: Description

Fix a deadlock when two threads access SentrySDK.options and SentrySDK.currentHub, which used the same object in synchronized. This problem is fixed now by using two independent locks for SentrySDK.options and SentrySDK.currentHub.

Fixes GH-3956, Fixes GH-3899

Unit tests and as explained in https://github.com/getsentry/sentry-cocoa/issues/3956.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Fix data race when calling reportFullyDisplayed from a background thread (#3926)
 - Ensure flushing envelopes directly after capturing them (#3915)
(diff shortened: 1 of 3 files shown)


brustolin (approved):
This probably solve the problem. But Im wondering whether reading the options need to be synchronised. We never change it or one of its properties, and the two functions related to options changed in this PR are internal.

Issue #3700, opened 2024-03-01
SentryFramesTracker Background CPU Usage


abbottkilroy:
### Platform

iOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

8.20.0

### Steps to Reproduce

Run app in background

### Expected Result

Sentry seems to be taking a bunch of CPU time when our app is in the background. SentryFramesTracker seems to be the culprit. As our app is continuously running in the background this can be a significant issue.

### Actual Result

<img width="1706" alt="image001" src="https://github.com/getsentry/sentry-cocoa/assets/115741741/9628533a-226f-4613-ab0c-67f663d34338">

philipphofmann:
We could stop the frames tracker when the app goes into the background, which we currently don't do. This should be an easy fix. Thanks for pointing that out, @abbottkilroy.

PR #3968, opened 2024-05-12
ref: move TestNSNotificationCenterWrapper.swift from SentryTests to SentryTestUtils

armcknight:
Finally figured out the necessary changes to get this in so I can use it in the continuous profiling tests.

I still had to reimplement SentryNotificationCenterWrapperTests from objc to swift because I couldn't figure out how to get the test subclass to appear in SentryTestUtils-Swift.h (which isn't actually used anywhere as a thing, as yet) so it could be imported into the objc test implementation.

#skip-changelog

PR #3979, opened 2024-05-14
impr: Stop FramesTracker when app is in backround

philipphofmann:
## :scroll: Description

Pause the SentryFramesTracker when the app goes to the background, and unpause
when the app goes to the foreground to save CPU cycles.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/3968.

Fixes https://github.com/getsentry/sentry-cocoa/issues/3700.

Unit tests and running the iOS-Swift sample app on a real device.

See the CPU usage of SentryFramesTracker without pausing, as explained in #3700, compared to pausing it when the app moves to the background.

| Without Pausing | With Pausing |
|--------|--------|
| Without Pausing| With Pausing | 

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/include/HybridPublic/SentryFramesTracker.h
+++ b/Sources/Sentry/include/HybridPublic/SentryFramesTracker.h
@@ -4,9 +4,11 @@
 
 #    import "SentryProfilingConditionals.h"
 
-@class SentryOptions, SentryDisplayLinkWrapper, SentryScreenFrames;
+@class SentryDisplayLinkWrapper;
+@class SentryScreenFrames;
 @class SentryCurrentDateProvider;
 @class SentryDispatchQueueWrapper;
+@class SentryNSNotificationCenterWrapper;
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -32,6 +34,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithDisplayLinkWrapper:(SentryDisplayLinkWrapper *)displayLinkWrapper
                               dateProvider:(SentryCurrentDateProvider *)dateProvider
                       dispatchQueueWrapper:(SentryDispatchQueueWrapper *)dispatchQueueWrapper
+                        notificationCenter:(SentryNSNotificationCenterWrapper *)notificationCenter
                  keepDelayedFramesDuration:(CFTimeInterval)keepDelayedFramesDuration;
 
 @property (nonatomic, assign, readonly) SentryScreenFrames *currentFrames;
--- a/Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
+++ b/Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
@@ -11,9 +11,8 @@ class SentryFramesTrackerTests: XCTestCase {
         var displayLinkWrapper: TestDisplayLinkWrapper
         var queue: DispatchQueue
         var dateProvider = TestCurrentDateProvider()
+        var notificationCenter = TestNSNotificationCenterWrapper()
         let keepDelayedFramesDuration = 10.0
-        lazy var hub = TestHub(client: nil, andScope: nil)
-        lazy var tracer = SentryTracer(transactionContext: TransactionContext(name: "test transaction", operation: "test operation"), hub: hub)
         
         let slowestSlowFrameDelay: Double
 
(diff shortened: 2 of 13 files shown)


Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
```diff
@@ -146,6 +135,13 @@ class SentryFramesTrackerTests: XCTestCase {
     }
 
     func testFrameRateChange() throws {
+#if os(iOS) || os(macOS) || targetEnvironment(macCatalyst)
+        let hub = TestHub(client: nil, andScope: nil)
+        let tracer = SentryTracer(transactionContext: TransactionContext(name: "test transaction", operation: "test operation"), hub: hub)
+        
+        // the profiler must be running for the frames tracker to record frame rate info etc, validated in assertProfilingData()
+        SentryLegacyProfiler.start(withTracer: tracer.traceId)
```
philipphofmann, line 143: Starting the profiler is only required for this test. As running the profiler requires a hub and a trace it interfered with the test cases I added for this PR I did this change in this PR. Furthermore, it makes sense to only run the profiler for tests that need it.

armcknight: I'm good with this, thanks for pointing it out!

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
```diff
@@ -15,7 +15,7 @@ class SentrySessionTrackerTests: XCTestCase {
 
         let notificationCenter = TestNSNotificationCenterWrapper()
         let dispatchQueue = TestSentryDispatchQueueWrapper()
-        lazy var fileManager = try! SentryFileManager(options: options, dispatchQueueWrapper: dispatchQueue)
+        lazy var fileManager = try! SentryFileManager(options: options, dispatchQueueWrapper: SentryDispatchQueueWrapper())
```
armcknight, line 18: Why are we switching this to a live dispatch queue wrapper?

Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
```diff
@@ -588,13 +588,51 @@ class SentryFramesTrackerTests: XCTestCase {
...
+    func testFrameTrackerPauses_WhenAppGoesToBackground() {
+        let sut = fixture.sut
+        sut.start()
+        
+        var callbackCalls = 0
+        let listener = FrameTrackerListener()
+        listener.callback = {
+            callbackCalls += 1
+        }
+        sut.add(listener)
```
armcknight, line 607: I don't think this is actually necessary here? We don't assert the number of `callbackCalls`. The one below does, maybe we just add the assertion here too.

SentryTestUtils/WeakReference.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+public class WeakReference<T: AnyObject> {
```
armcknight, line 3: I don't quite understand why we need this, could you explain here and/or in a comment? Were there retain cycles in the tests that kept objects alive between test cases?

Sources/Sentry/include/HybridPublic/SentryFramesTracker.h
```diff
@@ -4,9 +4,11 @@
 
 #    import "SentryProfilingConditionals.h"
 
-@class SentryOptions, SentryDisplayLinkWrapper, SentryScreenFrames;
```
armcknight, line 7: ❤️

Sources/Sentry/include/HybridPublic/SentryFramesTracker.h
```diff
@@ -4,9 +4,11 @@
...
+@class SentryDisplayLinkWrapper;
+@class SentryScreenFrames;
 @class SentryCurrentDateProvider;
 @class SentryDispatchQueueWrapper;
+@class SentryNSNotificationCenterWrapper;
```
armcknight, line 11: Just alphabetizing
```suggestion
@class SentryCurrentDateProvider;
@class SentryDispatchQueueWrapper;
@class SentryNSNotificationCenterWrapper;
@class SentryScreenFrames;
```

armcknight (approved):
Good work, I'm glad to see this. Just a couple cleanup items and questions, nothing blocking.

Maybe now that we have a pause/unpause method, in the future it will be easier to implement stopping the frame tracker only when there are performance tracking spans in flight. I thought I'd opened an issue for this at some point, but can't find it.

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philipphofmann: Good catch. That's not intended. I forgot to undo that change.

SentryTestUtils/WeakReference.swift
philipphofmann: I added a comment. Is it clear now? If not, I'm going to update the comment in a follow up PR.

PR #3986, opened 2024-05-14
ref: move TestHub to SentryTestUtils

armcknight:
Need this to be able to write a new profiler test in https://github.com/getsentry/sentry-cocoa/pull/3938 that needs a hub instance, and I don't want to use the live hub.

Had to rewrite `testSpanFinishesAfterTracerReleased_NoCrash_TracerIsNil` in swift, added to SentryTracerTests.swift, for the same reason as in #3968, I can't figure out how to import SentryTests-Swift.h into objc test classes.

#skip-changelog


philipphofmann (approved):
LGTM, thanks 👍

PR #3985, opened 2024-05-14
test(profiling): more tests for continuous profiling

armcknight:
Aiming to maximize code coverage of the code that's been added for continuous profiling, alongside the minimal tests added along the way.

for #3555; #skip-changelog

PR #3987, opened 2024-05-15
fix(profiling): put back the missing insertion of initial frame rate

armcknight:
In #3932, this line of code got accidentally removed. I'm not sure if this codepath is ever actually taken. It's not currently covered in tests; to be honest the mocking around GPU timestamps is a mess, and I haven't been able to untangle it to get this particular codepath covered. I hope to make more progress on that in #3985.

But I would like to get this merged before any release goes out, in case something would actually break.

#skip-changelog, for #3555


philipphofmann (approved):
LGTM

PR #3989, opened 2024-05-15
ref: pass options object into SentryTimeToDisplayTracker

armcknight:
This will help us use the enableContinuousProfiling option in the implementation without adding too many new parameters to the initializer.

#skip-changelog; for #3555


armcknight:
This actually isn't good enough, because `SentryUIViewControllerPerformanceTracker` is a singleton 😭 we need to thread its initialization through to somewhere instead of having it init itself, so the options can be passed in.

philipphofmann:
As mentioned in https://github.com/getsentry/sentry-cocoa/pull/3938#discussion_r1602986537, I think we don't need this PR.

PR #3993, opened 2024-05-16
Don't initialize NSApplication subclass during initialization

### LINKED ISSUE #3992: Sentry registers own NSApplication subclass (8.26 regression)
### Platform

macOS

### Environment

Production, Develop, TestFlight

### Installed

Swift Package Manager

### Version

8.26

### Xcode Version

15.4

8.25

### Steps to Reproduce

Since 8.26 it is no longer possible to use a custom NSApplication subclass while initializing Sentry before that class is initialized, because Sentry calls `[SentryCrashExceptionApplication sharedApplication]` as part of its initialization and thereby registers `SentryCrashExceptionApplication` as NSApplication instance.

This was caused by #3957. The fix is to e.g. have Sentry call `[SentryCrashExceptionApplication class]` instead.

1. Create custom NSApplication subclass inheriting from `SentryCrashExceptionApplication`register it as "Principal class" in Info.plist
2. Initialize SentrySDK before application class is initialized (e.g. from `+initialize`)

### Expected Result

NSApplication subclass is initialized normally.

### Actual Result

SentryCrashExceptionApplication is registered as application instance.

### Demo project

The following project illustrates the issue. Notice how `[MyApplication init]` is never called, even tho `MyApplication` is registered as principal class.
[SentryApplicationTest.zip](https://github.com/getsentry/sentry-cocoa/files/15336262/SentryApplicationTest.zip)

yes

triplef:

Fixes #3992.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixes
 
 - Fix retrieving GraphQL operation names crashing ([#3973](https://github.com/getsentry/sentry-cocoa/pull/3973))
+- Fix SentryCrashExceptionApplication subclass problem (#3993)
 
 ## 8.26.0
 


armcknight (commented):
Is there a reason the mac exception reporting isn't considered an integration, or a component of the crash reporter integration? Because if crash reporting is disabled, then we're crashing apps for no good reason.

If we moved this line of code to SentryOptions.defaultIntegrations, then it'd be nicely consistent with the rest of the `-[NSObject class]` invocations, and we wouldn't need this weird side effect here.

triplef:
Thanks for taking a look @armcknight and feel free to discard this PR if there’s a better solution!

philipphofmann:
> Is there a reason the mac exception reporting isn't considered an integration, or a component of the crash reporter integration?

No reason, no. It has always been like that, and we hardly touched it.

> then it'd be nicely consistent with the rest of the -[NSObject class] invocations

What do you mean by that, @armcknight?

brustolin:
> Is there a reason the mac exception reporting isn't considered an integration, or a component of the crash reporter integration?

As far as I know, having a NSApplication subclass is the only way to capture unhandled objc errors for macOS.

brustolin:
This solution looks good to me. It also avoid the stripping of the SentryCrashExceptionApplication.
Having it as an integration will require more refactoring and options, this could be a new issue for a later improvement.

philipphofmann (approved):
LGTM, thanks.

philipphofmann:
> > Is there a reason the mac exception reporting isn't considered an integration, or a component of the crash reporter integration?
> 
> As far as I know, having a NSApplication subclass is the only way to capture unhandled objc errors for macOS.

Yes, you need the NSApplication subclass thingy, but we could move the logic of calling `[SentryCrashExceptionApplication class];` into the SentryCrashIntegration so it only gets called when you install the crash integration, but as most people have it enabled and we fixed the bug, I think it's okay to keep the code in the SDK.m.

I also tested the fix and it works. We could add some UI tests for the macOS sample to test the regression, but I don't think it's worth the effort.

PR #3994, opened 2024-05-16
Include frame addresses in SentryFrame when possible

kabiroberai:
## :scroll: Description

This PR adds a `-[SentryFrame frameAddress]` property which represents the base address of the frame. This address is only included when building stack traces using `MachineContext` (but this covers most use cases where it's needed). Afaik the `backtrace(3)` approach doesn't give us enough info to include frame addresses.

It's really useful to have frame address information when dealing with stack overflow-type crashes. We have some `EXC_BAD_ACCESS` crashes at Ramp due to stack overflows, because Swift Concurrency has a relatively small stack size of 512K. We're hoping to use this to determine which stack frames/allocations are excessively large and can be shrunk, by computing the difference between consecutive frame addresses.

This PR doesn't aim to serialize the frame address and push it to the backend: that would require additional work on the backend side to handle the `frame_addr` property. Instead, the changes here give the client the opportunity to consume the frame information (e.g. in `beforeSend`) and potentially add context such as a list of large frames to their `Event.extra`.

- Triggered some stack overflow crashes and ensured that the `frameAddress` was accessible by the caller.
- Unit tests

## :pencil: Checklist

- [x] I added tests to verify the changes. (note: let me know if more unit tests are needed and/or where I can add coverage)
- [x] Review from the native team if needed. (requested)

## :crystal_ball: Next steps

It would be great if we could support this property on the Sentry backend so that clients can consume it without additional code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Improvements
 
+- Include frame addresses in SentryFrame when possible ([#3994](https://github.com/getsentry/sentry-cocoa/pull/3994))
 - Stop FramesTracker when app is in background (#3979)
 
 ### Fixes
--- a/Sources/Sentry/Public/SentryFrame.h
+++ b/Sources/Sentry/Public/SentryFrame.h
@@ -49,6 +49,11 @@ NS_SWIFT_NAME(Frame)
 @property (nonatomic, copy) NSString *_Nullable instructionAddress;
 
 /**
+ * Address of the frame pointer.
+ */
+@property (nonatomic, copy) NSString *_Nullable frameAddress;
+
+/**
  * InstructionAddress of the frame
  */
 @property (nonatomic) NSUInteger instruction;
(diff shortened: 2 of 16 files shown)


philipphofmann:
Thanks a lot for this proposal @kabiroberai.

> It's really useful to have frame address information when dealing with stack overflow-type crashes. We have some EXC_BAD_ACCESS crashes at Ramp due to stack overflows, because Swift Concurrency has a relatively small stack size of 512K. We're hoping to use this to determine which stack frames/allocations are excessively large and can be shrunk, by computing the difference between consecutive frame addresses.

How would you expect Sentry to show this information in an event? I would prefer building a feature that helps all users without having to manually compute differences between frame addresses themselves.

philipphofmann:
I'm closing this due to inactivity. Please feel free to reopen it with an answer to the question above (https://github.com/getsentry/sentry-cocoa/pull/3994#issuecomment-2117253154).

PR #3996, opened 2024-05-17
Expose custom cxa throw and rethrow calls to Unity

munkiki7:
## :scroll: Description

I have added custom calls that can be later used by Unity SDK for capturing C++ exceptions on iOS.
Also, when the current exception is NULL (during the std::terminate call) the sentrycrashsc_initSelfThread call is made to enrich the stack trace.

The fix is related to this [issue](https://github.com/getsentry/sentry-unity/issues/1628).

I have added the code to the Unity SDK that calls these new exposed functions using the dlsym search:

image

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add C++ exception support for `__cxa_rethrow` (#3996)
+
 ### Improvements
 
 - Stop FramesTracker when app is in background (#3979)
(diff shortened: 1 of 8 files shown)


philipphofmann (commented):
This change makes sense to me, but I think someone of @getsentry/owners-native should have a look at this proposal, because my CPP on that topic isn't the best. @armcknight, maybe you could also have a look please?

Swatinem (approved):
lgtm

munkiki7:
Glad to be helpful, @philipphofmann

Issue #2616, opened 2023-01-16
Session Replay for iOS native apps


bruno-garcia:
Support for Session Replay for Native iOS apps

Our goal is to support UIKit and SwiftUI. That means that apps rendering directly with Metal for example would not be supported. 

Please let us know on the issue description if this aligns with what you need/expect.

We're working on it! Wanna join the early adopter release? Join the waitlist and discussion about the feature: 
* https://github.com/getsentry/sentry/discussions/63138

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-1782) by [Unito](https://www.unito.io)

bruno-garcia:
We're doing a PoC of this as we speak. We should have some updates after the holidays

bruno-garcia:
We released our first Alpha version of the SDK with support: https://github.com/getsentry/sentry-cocoa/releases/tag/8.24.0-alpha.0

To get access, it requires adding your Sentry org to our feature flag. This way data can be ingested and displayed in Sentry.
[Please let us know on the waitlist if you're interested](https://sentry.io/lp/mobile-replay-beta/)

bruno-garcia:
Reopening because we had a first alpha but we'll keep this open until it's more generally available.

vaind:
I've given this a try to integrate into Flutter and I'm facing a couple of roadblocks so far:
1. The APIs are currently private (`SentrySessionReplayIntegration` and/or `SentrySessionReplay` and its dependencies, e.g. `SentryOnDemandReplay`, ...).
2. `SentryOnDemandReplay` would need a `func addFrame(imagePath: NSString)`
[...]

PR #4002, opened 2024-05-22
feat(Session Replay): Add breadcrumbs to session replay

brustolin:
## :scroll: Description

Converting session breadcrumbs to replay breadcrumbs

Unit tests

## :pencil: Checklist


Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+        var data = [String: Any]()
+        
+        breadcrumb.data?.forEach({
+            guard supportedNetworkData.contains($0.key) else { return }
+            let newKey = $0.key == "response_body_size" ? "bodySize" : $0.key.replacingOccurrences(of: "http.", with: "")
+            data[newKey.snakeToCamelCase()] = $0.value
+        })
+        
+        //We dont have end of the request in the breadcrumb.
+        return SentryRRWebSpanEvent(timestamp: timestamp, endTimestap: timestamp, operation: "resource.http", description: description, data: data)
```
brustolin, line 76: timestamp and endTimestamp are the same because the network breadcrumb dont have the start. An improvement for network tracker is coming in the next PR

Sources/Sentry/SentryScope.m
```diff
@@ -30,11 +30,6 @@
  */
 @property (atomic, strong) NSMutableDictionary<NSString *, id> *extraDictionary;
 
-/**
- * Contains the breadcrumbs which will be sent with the event
- */
-@property (atomic, strong) NSMutableArray<SentryBreadcrumb *> *breadcrumbArray;
```
brustolin, line 36: Moved to SentryScope+Private

Sources/Sentry/SentrySerialization.m
```diff
@@ -328,6 +328,7 @@ + (NSData *)dataWithReplayRecording:(SentryReplayRecording *)replayRecording
                               dataWithJSONObject:[replayRecording headerForReplayRecording]]];
     [recording appendData:[@"\n" dataUsingEncoding:NSUTF8StringEncoding]];
     [recording appendData:[SentrySerialization dataWithJSONObject:[replayRecording serialize]]];
+
```
brustolin, line 331:
```suggestion
```

Sources/Swift/Extensions/StringExtensions.swift
```diff
@@ -37,4 +37,20 @@ extension String {
...
+    
+    func snakeToCamelCase() -> String {
+        var result = ""
+        
+        var toUpper = false
+        for char in self {
+            if char == "_" {
+                toUpper = true
+            } else {
+                result.append(toUpper ? char.uppercased() : char.lowercased())
```
philipphofmann, line 49: `m`: I'm not sure if this is a problem, but if you perform this method on `hello_worldHello`, the outcome is `helloWorldhello`, and I think it should be `helloWorldHello` because we only use uppercase if the previous char was _. This logic doesn't keep existing uppercase chars.

Tests/SentryTests/Integrations/SessionReplay/SentryBreadcrumbReplayConverterTests.swift
```diff
@@ -0,0 +1,162 @@
...
+        
+        let result = try XCTUnwrap(sut.replayBreadcrumbs(from: [breadcrumb]).first)
+        let crumbData = try XCTUnwrap(result.data)
+        let payload = try XCTUnwrap(crumbData["payload"] as? [String: Any])
+        let payloadData = try XCTUnwrap(payload["data"] as? [String: Any])
+        
+        XCTAssertEqual(result.timestamp, breadcrumb.timestamp)
+        XCTAssertEqual(crumbData["tag"] as? String, "performanceSpan")
+        XCTAssertEqual(payload["description"] as? String, "https://test.com")
+        XCTAssertEqual(payload["op"] as? String, "resource.http")
```
philipphofmann, line 96: `m`: Also asserting the `startTimesamp` and `endTimestamp` properties of the payload improves this test case, as no other test validates these properties.

Tests/SentryTests/Integrations/SessionReplay/SentryReplayRecordingTests.swift
```diff
@@ -39,4 +38,17 @@ class SentryReplayRecordingTests: XCTestCase {
...
+        let date = Date(timeIntervalSince1970: 5)
+        let sut = SentryReplayRecording(segmentId: 3, size: 200, start: Date(timeIntervalSince1970: 2), duration: 5_000, frameCount: 5, frameRate: 1, height: 930, width: 390, extraEvents: [
+            SentryRRWebEvent(type: .custom, timestamp: date, data: nil)
+        ])
+      
+        let data = sut.serialize()
+        
+        let extraInfo = data[2]
+        expect(extraInfo["type"] as? Int) == 5
+        expect(extraInfo["timestamp"] as? Int) == 5_000
```
philipphofmann, line 52: `l`: Please ditch Nimble

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+        "status_code",
+        "method",
+        "response_content_length",
+        "request_content_length",
+        "http.query",
+        "http.fragment"]
+    )
+    
+    func replayBreadcrumbs(from breadcrumbs: [Breadcrumb]) -> [SentryRRWebEvent] {
+        breadcrumbs.compactMap { replayBreadcrumb(from: $0) }
```
philipphofmann, line 17: Ah, so great. We finally can use `compactMap` instead of iterating over the list and throwing out the nil values in ObjC. 💯

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+    //replay front understands
+    private func replayBreadcrumb(from breadcrumb: Breadcrumb) -> SentryRRWebEvent? {
+        guard let timestamp = breadcrumb.timestamp else { return nil }
+        if breadcrumb.category == "http" {
+            return networkSpan(breadcrumb)
+        } else if breadcrumb.type == "navigation" {
+            return navigationBreadcrumb(breadcrumb)
+        } else if breadcrumb.category == "touch" {
+            return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: "ui.tap", message: breadcrumb.message)
+        } else if breadcrumb.type == "connectivity" && breadcrumb.category == "device.connectivity" {
```
philipphofmann, line 30: `l`: Does it make sense to define reusable constantans for things like `device.connectivity`, `BATTERY_STATE_CHANGE`, so we ensure both this class and `SentrySystemEventBreadcrumbs` will use the same keywords?

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+            guard let networkType = breadcrumb.data?["connectivity"] as? String, !networkType.isEmpty  else { return nil }
+            return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: "device.connectivity", data: ["state": networkType])
+        } else if let action = breadcrumb.data?["action"] as? String, action == "BATTERY_STATE_CHANGE" {
+            var data = breadcrumb.data?.filter({ item in item.key == "level" || item.key == "plugged" }) ?? [:]
+            
+            data["charging"] = data["plugged"]
+            data["plugged"] = nil
```
philipphofmann, line 37: `l`: Please comment on why we set this to `nil`. I don't get it.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+            
+            return SentryRRWebBreadcrumbEvent(timestamp: timestamp,
+                                              category: "device.battery",
+                                              data: data)
+        }
+        
+        let level = getLevel(breadcrumb: breadcrumb)
```
philipphofmann, line 44: `m`: Why do we only keep the level for this crumb here? Shouldn't we also keep the level for all others?

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+            return networkSpan(breadcrumb)
+        } else if breadcrumb.type == "navigation" {
+            return navigationBreadcrumb(breadcrumb)
+        } else if breadcrumb.category == "touch" {
+            return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: "ui.tap", message: breadcrumb.message)
+        } else if breadcrumb.type == "connectivity" && breadcrumb.category == "device.connectivity" {
+            guard let networkType = breadcrumb.data?["connectivity"] as? String, !networkType.isEmpty  else { return nil }
+            return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: "device.connectivity", data: ["state": networkType])
+        } else if let action = breadcrumb.data?["action"] as? String, action == "BATTERY_STATE_CHANGE" {
+            var data = breadcrumb.data?.filter({ item in item.key == "level" || item.key == "plugged" }) ?? [:]
```
philipphofmann, line 34: `m`: Why do we filter the data here? If we add extra info in the `SentrySystemEventBreadcrumbs`, that would also be useful for SR, the data would be dropped here.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+    private func replayBreadcrumb(from breadcrumb: Breadcrumb) -> SentryRRWebEvent? {
+        guard let timestamp = breadcrumb.timestamp else { return nil }
+        if breadcrumb.category == "http" {
+            return networkSpan(breadcrumb)
+        } else if breadcrumb.type == "navigation" {
+            return navigationBreadcrumb(breadcrumb)
+        } else if breadcrumb.category == "touch" {
+            return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: "ui.tap", message: breadcrumb.message)
+        } else if breadcrumb.type == "connectivity" && breadcrumb.category == "device.connectivity" {
+            guard let networkType = breadcrumb.data?["connectivity"] as? String, !networkType.isEmpty  else { return nil }
```
philipphofmann, line 31: `m`: Why do we drop crumbs for which the networkType is empty?

Tests/SentryTests/Integrations/SessionReplay/SentryBreadcrumbReplayConverterTests.swift
```diff
@@ -0,0 +1,162 @@
...
+        crumb.type = "navigation"
+        crumb.data = ["state": "foreground"]
+        let result = sut.replayBreadcrumbs(from: [crumb])
+        
+        XCTAssertEqual(result.count, 1)
+        let event = result.first?.serialize()
+        let eventData = event?["data"] as? [String: Any]
+        let eventPayload = eventData?["payload"] as? [String: Any]
+        
+        XCTAssertEqual(event?["type"] as? Int, 5)
```
philipphofmann, line 32: `l`: Isn't that a bit cleaner? The same applies to all other event["type"] assertions.
```suggestion
        XCTAssertEqual(event?["type"] as? Int, SentryRRWebEventType.custom.rawValue)
```

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+                                              category: "device.battery",
+                                              data: data)
+        }
+        
+        let level = getLevel(breadcrumb: breadcrumb)
+        return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: breadcrumb.category, message: breadcrumb.message, level: level, data: breadcrumb.data)
+    }
+    
+    private func navigationBreadcrumb(_ breadcrumb: Breadcrumb) -> SentryRRWebBreadcrumbEvent? {
+        guard let timestamp = breadcrumb.timestamp else { return nil }
```
philipphofmann, line 49: `l`: This check is redundant. We already check for that in `replayBreadcrumb`. We could pass `timestamp` from `replayBreadcrumb` as a nonnil argument to here to ditch it.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+        
+        if breadcrumb.category == "app.lifecycle" {
+            guard let state = breadcrumb.data?["state"] else { return nil }
+            return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: "app.\(state)")
+        } else if let position = breadcrumb.data?["position"] as? String, breadcrumb.category == "device.orientation" && (position == "landscape" || position == "portrait") {
```
philipphofmann, line 54: `m`: Do we really need to check if the position is `portrait` or `landscape`? Can't we just send what the SDK already created?

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
...
+        var data = [String: Any]()
+        
+        breadcrumb.data?.forEach({
+            guard supportedNetworkData.contains($0.key) else { return }
+            let newKey = $0.key == "response_body_size" ? "bodySize" : $0.key.replacingOccurrences(of: "http.", with: "")
+            data[newKey.snakeToCamelCase()] = $0.value
+        })
+        
+        //We dont have end of the request in the breadcrumb.
+        return SentryRRWebSpanEvent(timestamp: timestamp, endTimestamp: timestamp, operation: "resource.http", description: description, data: data)
```
philipphofmann, line 76: `m`: I'm a bit confused why we convert a breadcrumb to a span event. What made you choose this way?

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,83 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+@objcMembers
+class SentryBreadcrumbReplayConverter: NSObject {
```
philipphofmann, line 5: `h`: This class heavily relies on the implementation detail of classes such as `SentryBreadcrumbTracker`, `SentrySystemEventBreadcrumbs`, etc. I'm worried if we change something in our breadcrumbs implementation that we will break the logic here. Do you have suggestions to mitigate this risk?

philipphofmann (commented):
I found a couple of issues.

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -284,6 +287,20 @@ - (void)captureSegment:(NSInteger)segment
...
+        NSArray *breadcrumbs = [scope.breadcrumbArray
+            filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(
+                                            SentryBreadcrumb *evaluatedObject,
+                                            NSDictionary<NSString *, id> *bindings) {
+                return [evaluatedObject.timestamp compare:videoInfo.start] != NSOrderedAscending &&
+                    [evaluatedObject.timestamp compare:videoInfo.end] != NSOrderedDescending;
+            }]];
+
+        events = [self->_breadcrumbConverter replayBreadcrumbsFrom:breadcrumbs];
+    }];
```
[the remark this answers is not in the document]
brustolin: Configure scope is thread safe

Sources/Swift/Extensions/StringExtensions.swift
brustolin: I dont have a preference. I just followed android implementation,
but I believe the premises is that snake case, [by convention](https://en.wikipedia.org/wiki/Snake_case), dont have upper cases in the middle of the word.

> Snake case (stylized as snake_case) is the naming convention in which each space is replaced with an underscore (_) character, and words are written in lowercase


@romtsn Thoughts?

Tests/SentryTests/Integrations/SessionReplay/SentryBreadcrumbReplayConverterTests.swift
brustolin: Its clear, but we can change `SentryRRWebEventType.custom` by mistake and the test will succeed where the server side is expecting 5.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
brustolin: Because Network information for session replay has a different struct.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
brustolin: That's a good point, that I dont have an answer for other than we should be aware of it when review PRs.

Session replay is expecting a specific data struct different from the one used by our breadcrumbs, that's why we need the converter, the best approach would be strongly typed breadcrumbs, which we dont have.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
brustolin: Btw, this is a RRWebSpan, no relation with trace span

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
brustolin: Removing information that will not be used.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
brustolin: The extra data is not displayed in the UI, session replay only need data that they have a use for.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
brustolin: Session replay UI has a specific implementation, any information that is not known for session replay will not be displayed.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
brustolin: Custom breadcrumbs will be displayed in a different tab in the UI, we send all the information of the breadcrumb in this cases.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
brustolin: This is more of an unwrapping of timestamp than check if it is nil.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,89 @@
...
+        "request_content_length",
+        "http.query",
+        "http.fragment"]
+    )
+    
+    func replayBreadcrumbs(_ breadcrumbs: [Breadcrumb], from: Date, until: Date) -> [SentryRRWebEvent] {
+        breadcrumbs.filter {
+            guard let timestamp = $0.timestamp else { return false }
+            return timestamp >= from && timestamp <= until
+        }
```
philipphofmann, line 20: Thanks for converting this to Swift ❤️

Tests/SentryTests/Extensions/StringExtensionsTests.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+@testable import Sentry
+import XCTest
+
+class StringExtensionsTests: XCTestCase {
```
philipphofmann, line 5: 💯 Thanks

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
philipphofmann: I'm sorry I haven't been clear enough. I meant, please this comment to the code.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
philipphofmann: Please add this info as a comment in the code, as it's not obvious and subsequent readers of the code could ask the same question.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
philipphofmann: Again, please add this info as a comment to the code, as it's not obvious.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
philipphofmann: Again, please add this info as a comment to the code, as it's not obvious.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
philipphofmann: Fine by me.

philipphofmann (commented):
Almost, LGTM.

brustolin:
> I'm sorry I haven't been clear enough. I meant, please this comment to the code.

Oh no, I did understand, I just do see it. We check some information, if it dont match what we want we ditch it, pretty simple stuff.

But I will add a general comment in the convert function to explain this.

Sources/Swift/Extensions/StringExtensions.swift
romtsn: hm, I'm not sure the android implementation is the same, we only uppercase the first character after the underscore, but keep the case of the other characters. So we never lowercase anything

brustolin: Ow, so I got it wrong. Ok, I can change it.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,92 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+@objcMembers
+class SentryBreadcrumbReplayConverter: NSObject {
```
romtsn, line 5: Probably nitpicking, but: should we name it `SentryReplayBreadcrumbConverter` rather? Because now it's sounds like we're converting breadcrumb replays and not replay breadcrumbs 😅

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,92 @@
...
+    private let supportedNetworkData = Set<String>([
+        "status_code",
+        "method",
+        "response_content_length",
+        "request_content_length",
+        "http.query",
+        "http.fragment"]
+    )
+    
+    func replayBreadcrumbs(_ breadcrumbs: [Breadcrumb], from: Date, until: Date) -> [SentryRRWebEvent] {
```
romtsn, line 16: also nit: I'd expect the main method of the `XXXConverter` class to be just `convert` 🙂

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -21,8 +21,11 @@ class SentryBreadcrumbReplayConverter: NSObject {
         .compactMap { replayBreadcrumb(from: $0) }
     }
     
-    //Convert breadcrumb information into something
-    //replay front understands
+    /**
+     * This function will convert the SDK breadcrumbs to session replay breadcrumbs in a format that the front-end understands.
```
philipphofmann, line 25: Thanks 👍

philipphofmann (approved):
Thanks for adding tests to ensure changes in the SDK crumbs don't break SR crumbs. Maybe it would make sense to move all the SR crum tests from the SDK crumb test classes to the SR crumb converter tests class.

brustolin:
> Maybe it would make sense to move all the SR crum tests from the SDK crumb test classes to the SR crumb converter tests class

Some of those tests have a context already in place in their respective test classes. Its easier to leave it as it is.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,92 @@
...
+    private func networkSpan(_ breadcrumb: Breadcrumb) -> SentryRRWebSpanEvent? {
+        guard let timestamp = breadcrumb.timestamp,
+              let description = breadcrumb.data?["url"] as? String,
+              let startTimestamp = breadcrumb.data?["request_start"] as? Date
+        else { return nil }
+        var data = [String: Any]()
+        
+        breadcrumb.data?.forEach({
+            guard supportedNetworkData.contains($0.key) else { return }
+            let newKey = $0.key == "response_body_size" ? "bodySize" : $0.key.replacingOccurrences(of: "http.", with: "")
```
romtsn, line 81: > $0.key == "response_body_size" 

`h`: I think this condition will never be met, because `supportedNetworkData` does not contain `response_body_size`/`request_body_size`. You should probably add these keys to the list

romtsn: and I think in general this logic is slightly off - basically the keys should be `data['requestBodySize']` or `data['responseBodySize']` - I hope it makes more sense this way

brustolin: But If I got it right from java, you're using `bodySize`
 ```Java
 val breadcrumbData = mutableMapOf<String, Any?>()
            for ((key, value) in breadcrumb.data) {
                if (key in supportedNetworkData) {
                    breadcrumbData[
                        key
                            .replace("content_length", "body_size")
                            .substringAfter(".")
                            .snakeToCamelCase()
                    ] = value
                }
            }
```
Not sure If I got your comment.

Comment (romtsn) [reply Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift:81]: yeah, I guess I wasn't clear enough - so you have this line:
```
guard supportedNetworkData.contains($0.key) else { return }
```
but because `supportedNetworkData` doesn't contain `response_body_size`/`request_body_size` in it, this will always early-return, so you will never get to the next line for those keys. So you have to update `supportedNetworkData` with those keys.

to the second point - yes we use `bodySize` but here we actually replace a `content_length` substring with `body_size`, so `response_content_length` transforms to `response_body_size` and then we camelCase it to `responseBodySize`. I hope it makes it more clear

File: Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift [hunk 25]
```diff
@@ -0,0 +1,92 @@
...
+        }
+        .compactMap { replayBreadcrumb(from: $0) }
+    }
+    
+    /**
+     * This function will convert the SDK breadcrumbs to session replay breadcrumbs in a format that the front-end understands.
+     * Any deviation in the information will cause the breadcrumb or the information itself to be discarded
+     * in order to avoid unknown behavior in the front-end.
+     */
+    private func replayBreadcrumb(from breadcrumb: Breadcrumb) -> SentryRRWebEvent? {
```
vaind, line 29: For the RN SDK (and Flutter later), there should be a way to extend or override this so that it can implement mappings for its own breadcrumbs. AFAIU there's currently no way to instantiate `SentrySessionReplay` from a hybrid SDK so it can't provide its own implementation. Or has anything changed in that regard lately? It came up in the discussion in [this issue](https://github.com/getsentry/sentry-cocoa/issues/2616) but I haven't noticed any updates.

brustolin: Nothing new yet. Still need to find a proper way to make SentrySessionReplay Hybrid public

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,92 @@
...
+        guard let timestamp = breadcrumb.timestamp else { return nil }
+        
+        if breadcrumb.category == "app.lifecycle" {
+            guard let state = breadcrumb.data?["state"] else { return nil }
+            return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: "app.\(state)")
+        } else if let position = breadcrumb.data?["position"] as? String, breadcrumb.category == "device.orientation" {
+            return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: "device.orientation", data: ["position": position])
+        } else {
+            if let to = breadcrumb.data?["screen"] as? String {
+                return SentryRRWebBreadcrumbEvent(timestamp: timestamp, category: "navigation", message: to, data: ["to": to])
```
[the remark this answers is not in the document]
romtsn: alright, I guess we have similar problems for Fragments, but I'll leave it up to you to decide. We do the filtering here: https://github.com/getsentry/sentry-java/pull/3406/files#diff-2f6abcfaf76542e5e58a9fbd4458b23ca658763d9bc0f989abfdeeee5c5898ddR239

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
```diff
@@ -0,0 +1,92 @@
...
+            guard supportedNetworkData.contains($0.key) else { return }
+            let newKey = $0.key == "response_body_size" ? "bodySize" : $0.key.replacingOccurrences(of: "http.", with: "")
+            data[newKey.snakeToCamelCase()] = $0.value
+        })
+        
+        //We dont have end of the request in the breadcrumb.
+        return SentryRRWebSpanEvent(timestamp: startTimestamp, endTimestamp: timestamp, operation: "resource.http", description: description, data: data)
+    }
+    
+    private  func getLevel(breadcrumb: Breadcrumb) -> SentryLevel {
```
romtsn, line 89:
```suggestion
    private func getLevel(breadcrumb: Breadcrumb) -> SentryLevel {
```
swiftlint please :P

romtsn (approved):
LGTM!

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
brustolin: We do have one in place with git hooks. Not sure what is going on here.

Sources/Swift/Integrations/SessionReplay/SentryBreadcrumbReplayConverter.swift
vaind: Alright, with breadcrumbs this starts to be a blocker for react-native too so when do you think you could have a look into that?

brustolin: I guess @bruno-garcia can help us with the priorities. 
Right now Im working with touch tracker. 
I can work on exposing SessionReplay for hybrids after it, or I can give it a pause right now.

bruno-garcia: I'd say it's OK to push multi touch in order to unblock a react native breadcrumbs. We need to get that one out

PR #4007, opened 2024-05-23
feat(profiling): continuous profile absolute NSDate-based timestamps

armcknight:
In order to line up profiles with streamed spans or other performance transactions, we must have absolute timestamps. In order to comply with Apple's privacy manifest rules, we will use NSDate based timestamps for this purpose. As mentioned in https://github.com/getsentry/sentry-cocoa/pull/3932#discussion_r1600543307

This PR also adds the remaining testing for cpu/memory/GPU metrics in the serialized JSON payloads, originally mentioned in https://github.com/getsentry/sentry-cocoa/pull/3987#issue-2296890721. I wound up duplicating the fixture logic to mock metrics samples from the tracer to the continuous case because trying to get the same code to work for both cases, which have different slicing behavior and timestamp bases, was too difficult. Better to let them each stand on their own.

I also had to do the same duplication for more of the serialization and slicing logic for continuous profiling. It turns out that the timestamp-based slicing isn't even necessary, since all the components start and stop together, as opposed to keeping e.g. GPU timestamp collection going for multiple concurrent profilers, that then later need to be demultiplexed.

I also introduced a TestFramesTracker so I could mock the GPU timestamps for the new continuous profile tests, instead of simulating a device and letting the real tracker work on that. I didn't get the trace profile tests moved over to it, I wound up spending so much time on all this. I'd rather let that keep working as is and then hopefully discard it one day soon!

for #3555; #skip-changelog


Tests/SentryProfilerTests/SentryProfileTestFixture.swift
```diff
@@ -142,34 +182,34 @@ class SentryProfileTestFixture {
         func renderGPUFrame(_ type: GPUFrame) {
             switch type {
             case .normal:
-                let currentSystemTime: UInt64 = currentDateProvider.systemTime()
-                print("expect normal frame to start at \(currentSystemTime)")
+                let timestamp = String(currentDateProvider.systemTime())
+                log(#line, "expect normal frame to start at \(timestamp)")
```
philipphofmann, line 186: `m`: What's the benefit of logging all that information? Do we really need that? This is a lot of information. This refers to all log messages here.

Tests/SentryProfilerTests/SentryProfileTestFixture.swift
```diff
@@ -96,39 +100,75 @@ class SentryProfileTestFixture {
...
-        // gather mocked metrics readings
-        for _ in 0..<mockUsageReadingsPerBatch {
+    lazy var df: DateFormatter = {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZ"
+        return dateFormatter
+    }()
+    
+    func log(_ line: Int, _ message: String) {
+        print("\(df.string(from: Date())) [Sentry] [TEST] [\((#file as NSString).lastPathComponent):\(line)]: \(message)")
```
philipphofmann, line 163: `m`: What made you use `print` instead of `SentryLog.log` here?

Tests/SentryProfilerTests/SentryProfileTestFixture.swift
```diff
@@ -219,6 +259,110 @@ class SentryProfileTestFixture {
...
+        }
+        
+        /*
+         * Mock a series of GPU frame renders of varying quality (normal/slow/frozen) and
+         * refresh rate changes. The refresh rate changes ("|") happen at the same time as
+         * the frame render they appear above. Time is not to scale; frozen frames last
+         * much longer than the lower end of slow frames.
+         *
+         * refresh rate:  |---60hz--------------|---120hz------|---60hz--------------------------------------|
+         * time:          N--S----N--F----------N-N-S--N-F-----N--N--S----N--F----------N--S----N--F----------
```
philipphofmann, line 336: 🥇

Sources/Sentry/SentryMetricProfiler.mm
```diff
@@ -81,6 +75,27 @@ @implementation SentryMetricReading
...
+    [readings enumerateObjectsUsingBlock:^(
+        SentryMetricReading *_Nonnull reading, NSUInteger idx, BOOL *_Nonnull stop) {
+        const auto value = [NSMutableDictionary dictionary];
+        value[@"timestamp"] = @(reading.absoluteNSDateInterval);
+        value[@"value"] = reading.value;
+        [serializedValues addObject:value];
+    }];
+    if (serializedValues.count == 0) {
+        return nil;
+    }
```
philipphofmann, line 96: `l`: You could do an early return if `readings.count == 0`, instead of checking if `serializedValues.count == 0` here. Then, you would skip allocating an NSMutableArray in that case.

Sources/Sentry/SentryMetricProfiler.mm
```diff
@@ -146,25 +162,63 @@ - (void)stop
...
+- (NSMutableDictionary<NSString *, id> *)serializeContinuousProfileMetrics;
+{
+    NSArray<SentryMetricReading *> *memoryFootprint;
+    NSArray<SentryMetricReading *> *cpuEnergyUsage;
+    NSArray<SentryMetricReading *> *cpuUsage;
+    @synchronized(self) {
+        cpuEnergyUsage = [NSArray<SentryMetricReading *> arrayWithArray:_cpuEnergyUsage];
+        memoryFootprint = [NSArray<SentryMetricReading *> arrayWithArray:_memoryFootprint];
+        cpuUsage = [NSArray<SentryMetricReading *> arrayWithArray:_cpuUsage];
+    }
```
philipphofmann, line 192: 👍 for minimizing the synchronized scope size

philipphofmann (commented):
> I didn't get the trace profile tests moved over to it, I wound up spending so much time on all this. I'd rather let that keep working as is and then hopefully discard it one day soon!

Yes, let's ditch them as soon as we can.

philipphofmann (approved):
Ups, forgot to approve. LGTM.

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
```diff
@@ -36,19 +42,24 @@
...
+    const auto profilerState = [profiler.state copyProfilingData];
     [profiler.state clear]; // !!!: profile this to see if it takes longer than one sample duration
                             // length: ~9ms
 
+    const auto metricProfilerState = [profiler.metricProfiler serializeContinuousProfileMetrics];
+    [profiler.metricProfiler clear];
+
+#    if SENTRY_HAS_UIKIT
+    const auto framesTracker = SentryDependencyContainer.sharedInstance.framesTracker;
+    SentryScreenFrames *screenFrameData = [framesTracker.currentFrames copy];
```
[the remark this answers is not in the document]
armcknight: So this was actually copypasta'd from the trace profiler's usage pattern, where we need to keep a copy in case the entire structure is mutated by another profiler, to avoid either data loss or corruption.

Since the frames tracker starts/stops with the profiler, we probably don't have to do this anymore.

Tests/SentryProfilerTests/SentryProfileTestFixture.swift
armcknight: Hmm, i didn't even consider to use the sentry logger from tests lol 😅

Tests/SentryProfilerTests/SentryProfileTestFixture.swift
armcknight: it helped debug things, but moreso when i was still trying to get the simulation working. I needed to know where things actually went wrong before the eventual failed test assertion.

I typically find that helpful logs are much too voluminous to read line by line for understanding, but it's helpful to have the info logged if you know what to search for. In projects I've worked on in the past that have both more voluminous and mature logging solutions, they had a way to set log levels per component. So when you're debugging, you could set only replays/network/tests to debug level, and everything else to info/warn, to focus on just the components you're working on. 

I thought I'd already filed an issue for this, but couldn't find anything in GitHub or even Notion, so I created one: https://github.com/getsentry/sentry-cocoa/issues/4020

Sources/Sentry/SentryMetricProfiler.mm
```diff
@@ -193,6 +247,7 @@ - (void)recordMemoryFootprint
     }
 
     @synchronized(self) {
+        SENTRY_LOG_DEBUG(@"Recording memory footprint of %llu", footprintBytes);
```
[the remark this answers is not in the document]
armcknight: yeah this one is super verbose, agreed. it helped me debug some stuff in the tests but i can remove it. 

Maybe it'd be worthwhile introducing another "trace" log level for our purposes, but, i'd be hesitant to do so until  we had macros that didn't even _evaluate_ the log statements beneath the configured log level, to avoid building up huge strings that are eventually discarded. See again, https://github.com/getsentry/sentry-cocoa/issues/4020

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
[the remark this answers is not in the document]
armcknight: I think what happened is when I originally read the callsite in the past, `currentFrames` looks like a property that is maintaining state, instead of creating an instance of something to return, so I took a copy to avoid data loss/races/corruption. This is another casualty in the long battle over whether to use dot or bracket notation when calling an objc method. Also, it used to be declared as a `@property`, which I've changed in this PR. Never use `@property` synthesis to make a getter that does something unrelated–always use a method for that!

Tests/SentryProfilerTests/SentryProfileTestFixture.swift
armcknight: I think I'd prefer to keep them separate, at the very least, be able to have a separate instance of SentryLogger that can be used from test code, that doesn't interfere with the one in the SDK.

armcknight:
Not sure why that rule failed in the lint step, but it's unrelated to the diff here.

PR #4009, opened 2024-05-23
feat(profiling): include continuous profile IDs in spans

armcknight:
The last piece of the puzzle is to include any continuous profiling session IDs in span serialized payloads so they can be linked in the backend.

for #3555; #skip-changelog

### DIFF
--- a/Sources/Sentry/Profiling/SentryContinuousProfiler.mm
+++ b/Sources/Sentry/Profiling/SentryContinuousProfiler.mm
@@ -116,12 +116,6 @@ _sentry_threadUnsafe_transmitChunkEnvelope(void)
         disableTimer();
     }
 
-    [SentryDependencyContainer.sharedInstance.notificationCenterWrapper
-        postNotification:[[NSNotification alloc]
-                             initWithName:kSentryNotificationContinuousProfileStopped
-                                   object:nil
-                                 userInfo:nil]];
-
     {
         std::lock_guard<std::mutex> l(_threadUnsafe_gContinuousProfilerLock);
         [_threadUnsafe_gContinuousCurrentProfiler
--- a/Sources/Sentry/Profiling/SentryContinuousProfiler.mm
+++ b/Sources/Sentry/Profiling/SentryContinuousProfiler.mm
@@ -32,6 +32,9 @@ SentryProfiler *_Nullable _threadUnsafe_gContinuousCurrentProfiler;
 
 NSTimer *_Nullable _chunkTimer;
 
+/** @note: The session ID is reused for any profile sessions started in the same app session. */
+SentryId *_profileSessionID;
+
 void
 disableTimer()
 {
@@ -86,6 +89,10 @@ _sentry_threadUnsafe_transmitChunkEnvelope(void)
             SENTRY_LOG_WARN(@"Continuous profiler was unable to be initialized.");
             return;
         }
+
+        static dispatch_once_t onceToken;
+        dispatch_once(&onceToken, ^{ _profileSessionID = [[SentryId alloc] init]; });
+        _threadUnsafe_gContinuousCurrentProfiler.profilerId = _profileSessionID;
     }
 
     [SentryDependencyContainer.sharedInstance.notificationCenterWrapper
--- a/Sources/Sentry/Profiling/SentryContinuousProfiler.mm
+++ b/Sources/Sentry/Profiling/SentryContinuousProfiler.mm
@@ -6,6 +6,7 @@
 #    import "SentryDispatchQueueWrapper.h"
 #    import "SentryLog.h"
 #    import "SentryMetricProfiler.h"
+#    import "SentryNSNotificationCenterWrapper.h"
 #    import "SentryNSTimerFactory.h"
 #    import "SentryProfiler+Private.h"
 #    import "SentryProfilerSerialization.h"
@@ -87,6 +88,11 @@ _sentry_threadUnsafe_transmitChunkEnvelope(void)
         }
     }
 
+    [SentryDependencyContainer.sharedInstance.notificationCenterWrapper
+        postNotification:[[NSNotification alloc]
+                             initWithName:kSentryNotificationContinuousProfileStarted
+                                   object:nil
+                                 userInfo:nil]];
     [self scheduleTimer];
 }
 
@@ -98,16 +104,36 @@ _sentry_threadUnsafe_transmitChunkEnvelope(void)
 
 + (void)stop
 {
-    std::lock_guard<std::mutex> l(_threadUnsafe_gContinuousProfilerLock);
+    {
+        std::lock_guard<std::mutex> l(_threadUnsafe_gContinuousProfilerLock);
 
-    if (![_threadUnsafe_gContinuousCurrentProfiler isRunning]) {
-        SENTRY_LOG_DEBUG(@"No continuous profiler is currently running.");
-        return;
+        if (![_threadUnsafe_gContinuousCurrentProfiler isRunning]) {
+            SENTRY_LOG_DEBUG(@"No continuous profiler is currently running.");
+            return;
+        }
+
+        _sentry_threadUnsafe_transmitChunkEnvelope();
+        disableTimer();
     }
 
-    _sentry_threadUnsafe_transmitChunkEnvelope();
-    disableTimer();
-    [_threadUnsafe_gContinuousCurrentProfiler stopForReason:SentryProfilerTruncationReasonNormal];
+    [SentryDependencyContainer.sharedInstance.notificationCenterWrapper
+        postNotification:[[NSNotification alloc]
+                             initWithName:kSentryNotificationContinuousProfileStopped
+                                   object:nil
+                                 userInfo:nil]];
+
+    {
+        std::lock_guard<std::mutex> l(_threadUnsafe_gContinuousProfilerLock);
+        [_threadUnsafe_gContinuousCurrentProfiler
+            stopForReason:SentryProfilerTruncationReasonNormal];
+        _threadUnsafe_gContinuousCurrentProfiler = nil;
+    }
+}
+
++ (nullable SentryId *)currentProfilerID
+{
+    std::lock_guard<std::mutex> l(_threadUnsafe_gContinuousProfilerLock);
+    return _threadUnsafe_gContinuousCurrentProfiler.profilerId;
 }
 
 #    pragma mark - Private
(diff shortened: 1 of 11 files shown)


Sources/Sentry/SentrySpan.m
```diff
@@ -84,10 +96,50 @@ - (instancetype)initWithContext:(SentrySpanContext *)context
...
+                    addObject:SentryContinuousProfiler.currentProfilerID.sentryIdString];
+            }
+            [SentryDependencyContainer.sharedInstance.notificationCenterWrapper
+                addObserver:self
+                   selector:@selector(linkProfiler)
+                       name:kSentryNotificationContinuousProfileStarted];
+            [SentryDependencyContainer.sharedInstance.notificationCenterWrapper
+                addObserver:self
+                   selector:@selector(linkProfiler)
+                       name:kSentryNotificationContinuousProfileStopped];
```
philipphofmann, line 114: `m`: I guess using the NSNotificationCenter is fine for an experimental release, but I believe this could become a performance problem when you have hundreds of spans or more running simultaneously. I'm unsure how performant the NSNotifcationCenter is when it has to loop through.

Sources/Sentry/SentrySpan.m
```diff
@@ -84,10 +96,50 @@ - (instancetype)initWithContext:(SentrySpanContext *)context
...
+            _profileSessionIDs = [NSMutableSet<NSString *> set];
+            if (SentryContinuousProfiler.isCurrentlyProfiling) {
+                [_profileSessionIDs
+                    addObject:SentryContinuousProfiler.currentProfilerID.sentryIdString];
+            }
+            [SentryDependencyContainer.sharedInstance.notificationCenterWrapper
+                addObserver:self
+                   selector:@selector(linkProfiler)
+                       name:kSentryNotificationContinuousProfileStarted];
+            [SentryDependencyContainer.sharedInstance.notificationCenterWrapper
```
philipphofmann, line 111: `m`: Wouldn't subscribing to started be enough? If there is an ongoing profile running, this method already adds the profileID. Also, adding it when a profile stops would add it twice, or is there an edge case that I don't see?

Tests/SentryTests/Transaction/SentrySpanTests.swift
```diff
@@ -63,6 +63,131 @@ class SentrySpanTests: XCTestCase {
...
+    func testSpanWithProfileId() throws {
+        let span = fixture.getSut()
+        SentryContinuousProfiler.start()
+        let profileId = try XCTUnwrap(SentryContinuousProfiler.profiler()?.profilerId.sentryIdString)
+        span.finish()
+        
+        let serialized = span.serialize()
+        
+        let includedProfileIDs = try XCTUnwrap(serialized["profile_ids"] as? [String])
+        XCTAssert(includedProfileIDs.contains(profileId))
```
philipphofmann, line 87: `l`: I think we could assert that it only contains this one profileId and nothing else. That's what I would expect.

Tests/SentryTests/Transaction/SentrySpanTests.swift
```diff
@@ -63,6 +63,131 @@ class SentrySpanTests: XCTestCase {
...
+    }
+
+    // test that a span that starts before a continuous profile and ends after
+    // the profile ends includes that profile id in the span
+    func testSpanWithProfileId1() throws {
```
philipphofmann, line 92: `l`: The method name could better reflect what it's testing, so when a test fails, you know what's wrong without reading the comment above.

Tests/SentryTests/Transaction/SentrySpanTests.swift
```diff
@@ -63,6 +63,131 @@ class SentrySpanTests: XCTestCase {
         clearTestState()
     }
     
+    func testSpanDoesNotIncludeTraceProfilerID() throws {
```
philipphofmann, line 66: Those are neat test cases covering multiple edge cases 💯

Sources/Sentry/SentrySpan.m
```diff
@@ -84,10 +96,50 @@ - (instancetype)initWithContext:(SentrySpanContext *)context
         _spanId = context.spanId;
         _sampled = context.sampled;
         _origin = context.origin;
+
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+        if (SentrySDK.options.enableContinuousProfiling) {
```
philipphofmann, line 101: `m`: I think we miss a test to only subscribe if `enableContinuousProfiling` is enabled.

Sources/Sentry/SentrySpan.m
```diff
@@ -84,10 +96,50 @@ - (instancetype)initWithContext:(SentrySpanContext *)context
...
+}
+
+- (void)linkProfiler
+{
+    [_profileSessionIDs addObject:SentryContinuousProfiler.currentProfilerID.sentryIdString];
+}
+
+- (void)stopObservingContinuousProfiling
+{
+    [SentryDependencyContainer.sharedInstance.notificationCenterWrapper
```
philipphofmann, line 134: `m`: I think we should only call this if `enableContinuousProfiling` is enabled.

philipphofmann (commented):
I think I found a few issues.

Tests/SentryTests/Transaction/SentrySpanTests.swift
armcknight: yep i updated them some time after you left this review, check em out now, they also have ascii diagrams!

Sources/Sentry/SentrySpan.m
armcknight: yeah personally i don't love it either, this just got it up and running the quickest. I can move it to my preferred pattern of delegation. less spaghetti like.

armcknight:
After some team discussion we're going to make a slight change here to only link a single profile ID instead of an array of them, and also reuse the same ID for all profile sessions in an app session.

Sources/Sentry/SentrySpan.m
armcknight: Yes, so I was thinking about the case where the profiler is already running by the time the span starts and then use the stop notification to get the profile ID, but if the span stops before the profile, and then it would never get that one either. We should just check for any current profilers running when the span starts and grab the profile ID at that time.

Tests/SentryTests/Transaction/SentrySpanTests.swift
armcknight: This will change to exactly that when we make the change I mentioned: https://github.com/getsentry/sentry-cocoa/pull/4009#issuecomment-2136240595

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
```diff
@@ -98,16 +104,30 @@ + (BOOL)isCurrentlyProfiling
...
 
-    _sentry_threadUnsafe_transmitChunkEnvelope();
-    disableTimer();
-    [_threadUnsafe_gContinuousCurrentProfiler stopForReason:SentryProfilerTruncationReasonNormal];
+    {
+        std::lock_guard<std::mutex> l(_threadUnsafe_gContinuousProfilerLock);
+        [_threadUnsafe_gContinuousCurrentProfiler
+            stopForReason:SentryProfilerTruncationReasonNormal];
+        _threadUnsafe_gContinuousCurrentProfiler = nil;
+    }
```
brustolin, line 124: l: Curious question: Why are you splitting this function into two scopes that use the same lock?

Sources/Sentry/SentrySpan.m
```diff
@@ -84,10 +96,46 @@ - (instancetype)initWithContext:(SentrySpanContext *)context
...
+}
+
+- (void)linkProfiler
+{
+    [_profileSessionIDs addObject:SentryContinuousProfiler.currentProfilerID.sentryIdString];
+}
+
+- (void)stopObservingContinuousProfiling
+{
+    if (SentrySDK.options.enableContinuousProfiling) {
```
philipphofmann, line 131: `l`: Sorry, I that I didn't catch this in the first review. If somebody keeps a reference to the Options and changes `enableContinuousProfiling` while the span is running to `false`, we would never stop observing. That's unlikely, but maybe it's better to add property to the SentrySpan to know if we have to stop observing.

Sources/Sentry/SentrySpan.m
```diff
@@ -84,10 +96,46 @@ - (instancetype)initWithContext:(SentrySpanContext *)context
...
         _sampled = context.sampled;
         _origin = context.origin;
+
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+        if (SentrySDK.options.enableContinuousProfiling) {
+            _profileSessionIDs = [NSMutableSet<NSString *> set];
+            if (SentryContinuousProfiler.isCurrentlyProfiling) {
+                [_profileSessionIDs
+                    addObject:SentryContinuousProfiler.currentProfilerID.sentryIdString];
+            } else {
```
philipphofmann, line 106: `m`: Shouldn't we also add the observer if we are already profiling? Can it happen that a span has multiple profileSessionIDs?

The test case for this would be
```
    +---------span-----------+
 +--profile1--+    +--profile2--+
```

Tests/SentryTests/Transaction/SentrySpanTests.swift
```diff
@@ -63,6 +66,182 @@ class SentrySpanTests: XCTestCase {
...
+            $0.name.rawValue == kSentryNotificationContinuousProfileStarted
+        }
+        XCTAssertEqual(continuousProfileObservations.count, 1)
+    }
+
+    /// Test a span that starts before and ends before a continuous profile, includes profile id
+    ///
+    /// ```
+    /// +-------span-------+
+    ///     +----profile----+
```
philipphofmann, line 121: thanks for the diagrams 🥇

philipphofmann (commented):
> After some team discussion we're going to make a slight change here to only link a single profile ID instead of an array of them, and also reuse the same ID for all profile sessions in an app session.

Sorry, I read this after adding my comments. I guess some of them could be irrelevant now.

philipphofmann (approved):
I don't want to block this. LGTM. If you want another review for the single profileID changes, please let me know.

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
armcknight: I can't remember now why this happened. I might've been debugging a deadlock. I don't see anything bad happening if I combine the scopes, I think there's no issue there.

Sources/Sentry/SentrySpan.m
armcknight: Good idea 👍🏻

Sources/Sentry/SentrySpan.m
armcknight: Yep you are right for the original implementation, but this is now obviated.

Sources/Sentry/SentrySpan.m
armcknight: Honestly if the question is just one of performance, I don't know if we're going to be able to beat it by just iterating through an array of listener instances 🤔 It's fair to assume that Apple has sunk lots of resources into optimizing such a widely used API.

Issue #3706, opened 2024-03-04
Clearly point out missing reason for NSExceptions


philipphofmann:
### Description

Our events should clearly communicate when the NSException doesn't have a reason, because this can lead to confusion, such as https://github.com/getsentry/sentry-cocoa/issues/3265. 

We could use a placeholder reason such as `<Missing Reason>` when there is none, as suggested in https://github.com/getsentry/sentry-cocoa/issues/3265#issuecomment-1973276793.

We have to ensure this works for unhandled NSExceptions as well as for user-reported NSExceptions via `SentrySDK.captureException`.

philipphofmann:
Another proposal is to change the exception value to something like: `reason: "it crashed"` and `reason: nil`. So when there is no reason, this clearly communicates this.

PR #4011, opened 2024-05-23
fix: Point out missing reason for NSExceptions

philipphofmann:
## :scroll: Description

Stop sending an empty reason for fatal NSExceptions, which is confusing to users. When the SentryException doesn't contain a value, Sentry displays (no error message).

Fixes GH-3706

Unit tests and with the macOS sample app. Fatal NSExceptions now look like this in Sentry.
CleanShot 2024-05-23 at 13 48 55@2x

I will create an issue on Sentry to replace the placeholder (no error message) with (no reason) for NSExceptions.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 - Fix retrieving GraphQL operation names crashing ([#3973](https://github.com/getsentry/sentry-cocoa/pull/3973))
 - Fix SentryCrashExceptionApplication subclass problem (#3993)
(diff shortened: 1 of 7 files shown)


philipphofmann:
We first need to complete https://github.com/getsentry/sentry-cocoa/issues/4012 to make this work. As 9.0.0 is around the corner, we will reopen this PR once ehttps://github.com/getsentry/sentry-cocoa/issues/4012 is completed.

PR #4014, opened 2024-05-24
ref: Cleanup PropagationContext imports

philipphofmann:
Remove unused imports.

#skip-changelog


brustolin (approved):
If this is not what's causing CI to fail, it LGTM.

PR #4017, opened 2024-05-27
fix: Crash in baggageEncodedDictionary

philipphofmann:
## :scroll: Description

The baggageEncodedDictionary method sometimes crashes with a null pointer exception in CFCharacterSetAddCharactersInString. It seems mutableCopy of NSCharacterSet.alphanumericCharacterSet or when calling stringByAddingPercentEncodingWithAllowedCharacters sometimes doesn't work. Instead of trying to fix the underlying issue of a null pointer exception, we can convert the code to Swift, which is designed for safety to avoid such mistakes. This should prevent the crash.

This is a frequent crash in our [internal SDK crash detection](https://sentry.sentry.io/issues/4688158870/?project=4505469596663808&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=6).

Unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Helper/SentryBaggageSerialization.swift
+++ b/Sources/Swift/Helper/SentryBaggageSerialization.swift
@@ -5,7 +5,7 @@ class SentryBaggageSerialization: NSObject {
     
     private static let SENTRY_BAGGAGE_MAX_SIZE = 8_192
     
-    static func baggageEncodedDictionary(_ dictionary: [String: String]) -> String {
+    static func encodeDictionary(_ dictionary: [String: String]) -> String {
         var items: [String] = []
         items.reserveCapacity(dictionary.count)
         
@@ -31,7 +31,7 @@ class SentryBaggageSerialization: NSObject {
         return items.sorted().joined(separator: ",")
     }
     
-    static func decodeBaggage(_ baggage: String) -> [String: String] {
+    static func decode(_ baggage: String) -> [String: String] {
         guard !baggage.isEmpty else {
             return [:]
         }
--- a/Sources/Swift/Helper/SentryBaggageSerialization.swift
+++ b/Sources/Swift/Helper/SentryBaggageSerialization.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 @objcMembers
-class SentrySwiftSerialization: NSObject {
+class SentryBaggageSerialization: NSObject {
     
     private static let SENTRY_BAGGAGE_MAX_SIZE = 8_192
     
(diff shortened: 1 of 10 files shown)


Sources/Swift/Helper/SentrySwiftSerialization.swift
```diff
@@ -0,0 +1,51 @@
...
+        var currentSize = 0
+        
+        for (key, value) in dictionary {
+            let keyDescription = key.addingPercentEncoding(withAllowedCharacters: allowedSet) ?? ""
+            let valueDescription = value.addingPercentEncoding(withAllowedCharacters: allowedSet) ?? ""
+            let item = "\(keyDescription)=\(valueDescription)"
+            if item.count + currentSize <= SENTRY_BAGGAGE_MAX_SIZE {
+                currentSize += item.count + 1 // +1 is to account for the comma that will be added for each extra item
+                items.append(item)
+            }
```
brustolin, line 23: h: An empty key or value should not be allowed, there is not added value.

brustolin (commented):
Great! I believe converting things is even better than trying to fix objc. There is one thing to talk about tho.

Sources/Swift/Helper/SentrySwiftSerialization.swift
philipphofmann: Good catch. A test should fail for this 😉

brustolin: Yes, I believe we should add it. 
Maybe the previous understanding was that those methods don't return nil, and that is the cause of the error.

philipphofmann: That makes sense. I update the code.

Sources/Swift/Helper/SentryBaggageSerialization.swift
```diff
@@ -0,0 +1,56 @@
...
+        allowedSet.insert(charactersIn: "-_.")
+        var currentSize = 0
+        
+        for (key, value) in dictionary {
+            guard let keyDescription = key.addingPercentEncoding(withAllowedCharacters: allowedSet), !keyDescription.isEmpty else {
+                continue
+            }
+            guard let valueDescription = value.addingPercentEncoding(withAllowedCharacters: allowedSet), !valueDescription.isEmpty else {
+                continue
+            }
```
brustolin, line 22: l: 
```suggestion
            guard let keyDescription = key.addingPercentEncoding(withAllowedCharacters: allowedSet), 
            let valueDescription = value.addingPercentEncoding(withAllowedCharacters: allowedSet), !keyDescription.isEmpty && !valueDescription.isEmpty  else {
                continue
            }
```

brustolin (approved):
LGTM

PR #4024, opened 2024-05-30
feat(profiling): change how continuous profiling is enabled in options

armcknight:
After some team discussion, we decided to change how continuous profiling is enabled. Remove SentryOptions.enableContinuousProfiling, and make setting SentryOptions.profilesSampleRate = nil the way to do it.

ETA (3 Jun 2024): a profilesSampleRate of 0 also enables continuous profiling. If a sampler function is defined, always disable continuous profiling.

Second edit: we decided to go back on the decision to overload the 0 value. Now, only setting it to nil is the way to enable continuous profiling again.

#skip-changelog; for #3555


philipphofmann (commented):
PR looks about right. I'm going to give this a full review once it's marked as ready for review.

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -456,7 +456,17 @@ NS_SWIFT_NAME(Options)
  * the SDK sets it to the default of @c 0.
  * This property is dependent on @c tracesSampleRate -- if @c tracesSampleRate is @c 0 (default),
  * no profiles will be collected no matter what this property is set to. This property is
- * used to undersample profiles *relative to* @c tracesSampleRate
+ * used to undersample profiles *relative to* @c tracesSampleRate .
+ * @note Setting this value to @c nil enables an experimental new profiling mode, called continuous
```
philipphofmann, line 460: `m`: I think we should mention that nothing happens if you don't interact with the continuous profiling API. Otherwise, you could assume that this is a breaking change.

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -142,7 +142,7 @@ - (void)framesTrackerHasNewFrame:(NSDate *)newFrameDate
         if (!_waitForFullDisplay) {
             [SentryDependencyContainer.sharedInstance.framesTracker removeListener:self];
 #    if SENTRY_TARGET_PROFILING_SUPPORTED
-            if (!SentrySDK.options.enableContinuousProfiling) {
+            if (SentrySDK.options.profilesSampleRate != nil) {
```
philipphofmann, line 145: `m`: Why can't we use `isContinuousProfilingEnabled` here?

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -415,6 +415,47 @@ class SentrySDKTests: XCTestCase {
         XCTAssert(transaction === newSpan)
     }
     
+#if os(iOS) || os(macOS) || targetEnvironment(macCatalyst)
+
+    func testStartingContinuousProfilerWithSampleRateZero() throws {
```
philipphofmann, line 420: Those tests are excellent for understanding how the code works. 💯

philipphofmann (approved):
LGTM

Sources/Sentry/SentryTimeToDisplayTracker.m
armcknight: The team has decided to completely remove that option. It wasn't really my preference, but it allows it to work with less configuration code.

armcknight: ah, i missed this one

Sources/Sentry/Public/SentryOptions.h
armcknight: Otherwise, I think what you're saying is sort of implied by the explanation that's already there, which talks about needing to call SentrySDK.startProfiler

Issue #4013, opened 2024-05-23
Add the ability to disable SIGTERM reporting


sindresorhus:
### Problem Statement

Support for SIGTERM was added in https://github.com/getsentry/sentry-cocoa/pull/3895, but in practice, I don't find it very useful or actionable. I have just been spammed with a lot of notifications about SIGTERM, which turns out to be users simply rebooting their Mac (I have a crash report form, so some users filled it out with that reason), and the system sending SIGTERM to all apps.

### Solution Brainstorm

Add a setting to disable the SIGTERM handling.

### Are you willing to submit a PR?

No

philipphofmann:
You can drop these events in beforeSend for now. We're going to add a flag.

sindresorhus:
> Would you say it makes sense to disable this per default on macOS, @sindresorhus?

I think it could still be useful to attach more information to watchdog events (as mentioned in https://github.com/getsentry/sentry-cocoa/pull/3895). For example, if the app is frozen and macOS tries to terminate it, it would be ideal to get a stack trace of what's currently running, so I could find the culprit. But reporting all SIGTERM events is not useful. So maybe catch SIGTERM events internally, but only use it to enrich other events? (I don't have deep knowledge how the Sentry SDK works though).

philipphofmann:
> It’s the OS trying to tell you that it wants apps to exit gracefully. I’ve seen it happen in many different scenarios. Some of the interesting ones are when the device is shutdown, when an app will be terminated to update it and when something changes in traits that might require the OS to restart apps.

> What’s interesting is that previously this would all go unnoticed to us the developers. If these happen when the user can see them, or perceive them, then we have an issue that need to be resolved. If they happen in the background, then we just hope that state restauration is well implemented and users don’t notice it.

@naftaly you wrote this ⬆️ here https://github.com/getsentry/sentry-cocoa/pull/3895#issuecomment-2123550329. I'm starting to believe we shouldn't mark these events as fatal crashes but report them as normal events or do you think it would be better to attach this information to other events as suggested above? I think the story is different for macOS and iOS.

naftaly:
> @naftaly you wrote this ⬆️ here [#3895 (comment)](https://github.com/getsentry/sentry-cocoa/pull/3895#issuecomment-2123550329). I'm starting to believe we shouldn't mark these events as fatal crashes but report them as normal events or do you think it would be better to attach this information to other events as suggested above? I think the story is different for macOS and iOS.

[...]

PR #4025, opened 2024-05-31
feat: Disable SIGTERM reporting by default

philipphofmann:
## :scroll: Description

We added support for SIGTERM reporting in the last release and enabled it by default. For some users, SIGTERM events were verbose and not actionable. Therefore, we disable it per default.

Fixes GH-4013

Options via unit test. The signal handling installation manually with log messages cause this code isn't testable at all at the moment, and I don't think we should change that by adding a simple boolean check.

## :pencil: Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,24 +4,24 @@ on:
     branches:
       - main
     paths:
-      - 'Sources/**'
-      - 'Tests/**'
-      - 'test-server/**'
-      - 'Samples/**'
-      - '.github/workflows/lint.yml'
-      - 'scripts/ci-select-xcode.sh'
+      - "Sources/**"
+      - "Tests/**"
+      - "test-server/**"
+      - "Samples/**"
+      - ".github/workflows/lint.yml"
+      - "scripts/ci-select-xcode.sh"
 
   pull_request:
     paths:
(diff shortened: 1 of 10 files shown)


brustolin (approved):

.github/workflows/lint.yml
```diff
@@ -4,24 +4,24 @@ on:
...
-      - 'test-server/**'
-      - 'Samples/**'
-      - '.github/workflows/lint.yml'
-      - 'scripts/ci-select-xcode.sh'
+      - "Sources/**"
+      - "Tests/**"
+      - "test-server/**"
+      - "Samples/**"
+      - ".github/workflows/lint.yml"
+      - "scripts/ci-select-xcode.sh"
```
brustolin, line 12: l: Any practical reason for this change?
Just want to understand.

philipphofmann: Nope, I added fail-fast below and VSCode changed this.

eric:
If this is false is a SIGTERM shutdown considered graceful or is it a watchdog event?

brustolin:
Hello @eric, if `enableSigtermReporting` is false, no issue related to SIGTERM will be created.

eric:
So a SIGTERM will not create a watchdog event?

PR #4027, opened 2024-05-31
chore: Exposing SessionReplayIntegration to hybrid SDKs

brustolin:
Now its possible to get a instance of SentrySessionReplayIntegration from `[PrivateSentrySDKOnly createReplayIntegration]`

To start the replay just call:
```objc
- (void)startWithOptions:(SentryReplayOptions *)replayOptions
      screenshotProvider:(id<SentryViewScreenshotProvider>)screenshotProvider
     breadcrumbConverter:(id<SentryReplayBreadcrumbConverter>)breadcrumbConverter
             fullSession:(BOOL)shouldReplayFullSession;
```
with the objects that control screenshot and breadcrumb conversion.

_#skip-changelog_


Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
```diff
@@ -156,8 +157,16 @@ typedef void (^SentryOnAppStartMeasurementAvailable)(
...
 + (NSData *)captureViewHierarchy;
+
 #endif // SENTRY_UIKIT_AVAILABLE
 
+#if SENTRY_HAS_UIKIT && !TARGET_OS_VISION
+/**
+ * This is how hybrid SDKs can control session replay.
+ */
++ (nullable SentrySessionReplayIntegration *)createReplayIntegration;
+#endif
```
vaind, line 168: How should we use this in hybrid SDKs? There's already an instance installed among integrations by the time the SDK starts. 

Do you suggest removing the original integration from options during `init()`, like so:
```objective-c
        options.integrations = [options.integrations filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(NSString * integration, NSDictionary *bindings) {
            return ![integration isEqualToString:@"SentrySessionReplayIntegration"];
        }]];
```

and then adding a custom hybrid SDK replay integration that holds an instance of `SentrySessionReplayIntegration` (created by this function) as a property and delegates to it? Or did you have another scenario in mind?

Sources/Sentry/include/HybridPublic/SentrySessionReplayIntegration-Hybrid.h
```diff
@@ -0,0 +1,18 @@
...
+@class SentryReplayOptions;
+
+@protocol SentryViewScreenshotProvider;
+@protocol SentryReplayBreadcrumbConverter;
+
+@interface SentrySessionReplayIntegration : NSObject
+
+- (void)startWithOptions:(SentryReplayOptions *)replayOptions
+      screenshotProvider:(id<SentryViewScreenshotProvider>)screenshotProvider
+     breadcrumbConverter:(id<SentryReplayBreadcrumbConverter>)breadcrumbConverter
```
vaind, line 14: I don't know much about objective-c, but shouldn't this be:
```suggestion
     breadcrumbConverter:(id<SentryReplayBreadcrumbConverterProtocol>)breadcrumbConverter
```

vaind: Or, following the convention with `SentryViewScreenshotProvider` which is already a protocol, maybe rename `SentryReplayBreadcrumbConverter` to `SentryDefaultReplayBreadcrumbConverter` and `SentryReplayBreadcrumbConverterProtocol` to `SentryReplayBreadcrumbConverter`.

Actually, I'd also suggest renaming `SentryViewScreenshotProvider` to `SentryScreenshotProvider`

Sources/Swift/Integrations/SessionReplay/SentryReplayBreadcrumbConverter.swift
```diff
@@ -1,8 +1,13 @@
...
 import Foundation
 
+@objc(SentryReplayBreadcrumbConverter)
+protocol SentryReplayBreadcrumbConverterProtocol: NSObjectProtocol {
+    func convert(breadcrumbs: [Breadcrumb], from: Date, until: Date) -> [SentryRRWebEvent]
+}
+
 @objcMembers
-class SentryReplayBreadcrumbConverter: NSObject {
+class SentryReplayBreadcrumbConverter: NSObject, SentryReplayBreadcrumbConverterProtocol {
```
vaind, line 10: This base converter should also be exposed to hybrid SDKs so that we can delegate to it after custom processing is done. Like here: https://github.com/getsentry/sentry-react-native/pull/3846/files#diff-e405eb323a493cbd718d41b79f2642874fc60ff229e17b139f8516b07fe2a7a2R35

vaind (changes requested):
I've only reviewed the code without being able to test it in RN.

Sources/Sentry/include/HybridPublic/SentrySessionReplayIntegration-Hybrid.h
brustolin: > I don't know much about objective-c, but shouldn't this be

No, there is no protocol suffix is the objc side. 
We can work on the class naming.

vaind: > No, there is no protocol suffix is the objc side.

what about this then?
https://github.com/getsentry/sentry-cocoa/pull/4027/files#diff-75294e7bc3f1e95b081b2d9a83fc65e10468a8c4089a8c0aadd6decf1245a672R5

brustolin: This is swift
```Swift
@objc(SentryReplayBreadcrumbConverter) // <-- the name used in objc
 protocol SentryReplayBreadcrumbConverterProtocol: NSObjectProtocol {
     func convert(breadcrumbs: [Breadcrumb], from: Date, until: Date) -> [SentryRRWebEvent]
 }
 ```

Comment (brustolin) [reply Sources/Sentry/include/HybridPublic/SentrySessionReplayIntegration-Hybrid.h:14]: But lets change this

File: Sources/Sentry/PrivateSentrySDKOnly.mm [hunk 4]
```diff
@@ -294,29 +288,34 @@ + (void)addReplayIgnoreClasses:(NSArray<Class> *_Nonnull)classes
...
 #if SENTRY_HAS_UIKIT && !TARGET_OS_VISION
-    return [[SentrySessionReplayIntegration alloc] init];
+    NSArray *integrations = [[SentrySDK currentHub] installedIntegrations];
+    SentrySessionReplayIntegration *replayIntegration;
+    for (id obj in integrations) {
+        if ([obj isKindOfClass:[SentrySessionReplayIntegration class]]) {
+            replayIntegration = obj;
+            break;
+        }
+    }
```
vaind, line 310: Maybe deduplicate this? It's also in `captureReplay()`

Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
```diff
@@ -161,10 +164,15 @@ typedef void (^SentryOnAppStartMeasurementAvailable)(
 #endif // SENTRY_UIKIT_AVAILABLE
 
 #if SENTRY_HAS_UIKIT && !TARGET_OS_VISION
+
 /**
- * This is how hybrid SDKs can control session replay.
+ * Configure session replay with different breadcrumb converter
+ * and screeshot provider. Used by the Hybrid SDKs.
+ * If can pass nil to avoid changing the property.
```
vaind, line 171:
```suggestion
 * Passing nil will keep the previous value.
```

Sources/Swift/Integrations/SessionReplay/SentrySRDefaultBreadcrumbConverter.swift
```diff
@@ -1,8 +1,13 @@
 @_implementationOnly import _SentryPrivate
 import Foundation
 
+@objc
+protocol SentryReplayBreadcrumbConverter: NSObjectProtocol {
+    func convert(breadcrumbs: [Breadcrumb], from: Date, until: Date) -> [SentryRRWebEvent]
```
vaind, line 6: What do you think about only single breadcrumb conversion here (we have it the same in Java)
```suggestion
    func convert(breadcrumb: Breadcrumb) -> SentryRRWebEvent?
```

that would allow moving the timestamp filter to the integration so that it doesn't need to be duplicated for all implementations.

Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
```diff
@@ -156,8 +160,21 @@ typedef void (^SentryOnAppStartMeasurementAvailable)(
...
 
+#if SENTRY_HAS_UIKIT && !TARGET_OS_VISION
+
+/**
+ * Configure session replay with different breadcrumb converter
+ * and screeshot provider. Used by the Hybrid SDKs.
+ * If can pass nil to avoid changing the property.
+ */
+- (void)configureSessionReplayWith:(nullable id<SentryReplayBreadcrumbConverter>)breadcrumbConverter
+                screenshotProvider:(nullable id<SentryViewScreenshotProvider>)screenshotProvider;
```
vaind, line 174:
```suggestion
+ (void)configureSessionReplayWith:(nullable id<SentryReplayBreadcrumbConverter>)breadcrumbConverter
                screenshotProvider:(nullable id<SentryViewScreenshotProvider>)screenshotProvider;
```

Sources/Swift/Integrations/SessionReplay/SentryReplayBreadcrumbConverter.swift
vaind: What's your plan of exposing the base converter to RN? We cannot subclass the swift class from obj-c

vaind: https://github.com/getsentry/sentry-react-native/pull/3846/files#diff-0a664422c5522958ee625f390b58d6b833bab7d360e761b371a3815c827e206dR25

Sources/Swift/Integrations/SessionReplay/SentrySRDefaultBreadcrumbConverter.swift
brustolin: This was a design decision because of diference between Swift and Objc filtering. Right now I prefer not to change this.

vaind (commented):
There are some minor changes I've suggested but this already works for RN so they're not mandatory

vaind (approved):

Sources/Sentry/PrivateSentrySDKOnly.mm
```diff
@@ -304,24 +305,43 @@ + (SentryBreadcrumb *)breadcrumbWithDictionary:(NSDictionary *)dictionary
     return [[SentryBreadcrumb alloc] initWithDictionary:dictionary];
 }
 
+#if SENTRY_HAS_UIKIT && !TARGET_OS_VISION
++ (SentrySessionReplayIntegration *)getReplayIntegration
```
philipphofmann, line 309: `m`: The return value of this method should be nullable as there is no guarantee that we always find an `SentrySessionReplayIntegration`.

philipphofmann (approved):
LGTM

Sources/Sentry/PrivateSentrySDKOnly.mm
brustolin:
```suggestion
+ (nullable SentrySessionReplayIntegration *)getReplayIntegration
```

PR #4030, opened 2024-06-04
fix(profiling): set continuous profile id in each serialized transaction at keypath contexts.trace.data.profile_id

armcknight:
Following along with #4009 , we also need the profile ID in each transaction (which just takes the id already recorded by the SentrySpan additions in that last PR when creating the transaction from the trace, which uses the span's logic as it's the superclass).

for #3555; #skip-changelog

### DIFF
--- a/Sources/Sentry/SentryTransaction.m
+++ b/Sources/Sentry/SentryTransaction.m
@@ -2,6 +2,8 @@
 #import "SentryEnvelopeItemType.h"
 #import "SentryMeasurementValue.h"
 #import "SentryNSDictionarySanitize.h"
+#import "SentryProfilingConditionals.h"
+#import "SentrySpan+Private.h"
 #import "SentrySwift.h"
 #import "SentryTransactionContext.h"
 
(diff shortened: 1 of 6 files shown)


Sources/Sentry/SentryTransaction.m
```diff
@@ -54,6 +56,11 @@ - (instancetype)initWithTrace:(SentryTracer *)trace children:(NSArray<id<SentryS
         serializedData[@"_metrics_summary"] = metricsSummary;
         [serializedTrace removeObjectForKey:@"_metrics_summary"];
     }
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+    NSMutableDictionary *traceProfileEntry = [NSMutableDictionary dictionary];
+    traceProfileEntry[@"profile_id"] = self.trace.profileSessionID;
+    serializedTrace[@"data"] = traceProfileEntry;
```
philipphofmann, line 62: `h`: This overwrites existing fields in the `data` keyword. Instead, we must add the `profile_id` to the existing data dictionary. I'm surprised that no test fails for this.

philipphofmann (commented):
I found one major issue.

Sources/Sentry/SentryTransaction.m
armcknight: good catch. i wrote a test that failed, and then fixed the issue.

philipphofmann (approved):
LGTM

PR #4031, opened 2024-06-04
ci: XLarge runner for Address Sanitizer UI tests

philipphofmann:
The address sanitizer is a bit more resource-intensive, so the tests for UI frames sometimes fail.
We can avoid this problem by using a more powerful GH actions runner.

It turns out that there are still problems in the UI tests, which will be fixed by https://github.com/getsentry/sentry-cocoa/pull/4032. After merging https://github.com/getsentry/sentry-cocoa/pull/4032, we can merge those changes in here and validate if this PR still makes sense.

#skip-changelog

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -85,7 +85,7 @@ jobs:
 
   ui-tests-address-sanitizer:
     name: UI Tests with Address Sanitizer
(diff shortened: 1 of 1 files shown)


philipphofmann:
No, that doesn't solve any problems. I'm closing this.

PR #4032, opened 2024-06-04
tests: Write test profile only when data exists

philipphofmann:
Only write the test profile when the serialization succeeds. We had crashes in UI tests when trying to write empty serialized profile data, because the SENTRY_CASSERT failed. This PR avoids creating the profile paths and file when there is nothing to write.

I could reproduce the problem by running `TopViewControllerTests` locally. Now, the tests run smoothly.

The UI tests here failed because of other problems fixed with https://github.com/getsentry/sentry-cocoa/pull/4033.

#skip-changelog

### DIFF
--- a/Sources/Sentry/Profiling/SentryProfilerSerialization.mm
+++ b/Sources/Sentry/Profiling/SentryProfilerSerialization.mm
@@ -329,11 +329,11 @@ SentryEnvelope *_Nullable sentry_continuousProfileChunkEnvelope(
         return nil;
     }
 
-#    if defined(TEST) || defined(TESTCI) || defined(DEBUG)
+#    if defined(TEST) || defined(TESTCI)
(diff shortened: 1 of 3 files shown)


Sources/Sentry/Profiling/SentryProfilerSerialization.mm
```diff
@@ -389,6 +385,10 @@
         return nil;
     }
 
+#    if defined(TEST) || defined(TESTCI) || defined(DEBUG)
```
brustolin, line 388: h: 
We should refrain from using `defined(DEBUG)` for testing things, this will also run from cocoapods

philipphofmann: Excellent catch; I just copy-pasted the code. Thanks 🥇

brustolin (approved):

PR #4041, opened 2024-06-06
feat(Session Replay): Track touches for replay

brustolin:
## :scroll: Description

Added a touch tracker to send touch events with replays

Unit tests
Samples

## :pencil: Checklist


Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+        var point: CGPoint {
+            CGPoint(x: x, y: y)
+        }
+    }
+    
+    private class TouchInfo {
+        let id: Int
+        
+        var start: TouchEvent?
+        var end: TouchEvent?
```
philipphofmann, line 23: `m`: When reading the code, I thought start and end are timestamps. Therefore, I would rename this to
```suggestion
        var startTouchEvent: TouchEvent?
        var endToucheEvent: TouchEvent?
```

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+        var movements = [TouchEvent]()
```
philipphofmann, line 24: `l`: Again just when looking at the movements property in the code, I have no clue what it is. Maybe that's a bit cleaner:
```suggestion
        var movementToucheEvents = [TouchEvent]()
```

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+  
+    func flushFinishedEvents() {
+        trackedTouches = trackedTouches.filter { $0.value.end == nil }
+    }
+    
+    func replayEvents(from: Date, until: Date) -> [SentryRRWebEvent] {
+        let uptime = dateProvider.systemUptime()
+        let now = dateProvider.date()
+        let startTimeInterval = uptime - now.timeIntervalSince(from)
+        let endTimeInterval = uptime - now.timeIntervalSince(until)
```
philipphofmann, line 111: `m`: It took me quite a while to figure out why we need sytemUptime here. It's because of `UIEvent.timestamp`

> This property contains the number of seconds that have elapsed since system startup. For a description of this time value, see the description of the [systemUptime](https://developer.apple.com/documentation/foundation/processinfo/1414553-systemuptime) method of the [ProcessInfo](https://developer.apple.com/documentation/foundation/processinfo) class.
> from Apple Docs: https://developer.apple.com/documentation/uikit/uievent/1613790-timestamp

Instead of converting the `timestamp` around multiple times in this method with `now.addingTimeInterval(infoStart.timestamp - uptime)`, can't we just directly store the timestamp of type Date on the `TouchEvent` in `trackTouchFrom`?

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+    init(dateProvider: SentryCurrentDateProvider, scale: Float) {
+        self.dateProvider = dateProvider
+        self.scale = CGAffineTransform(scaleX: CGFloat(scale), y: CGFloat(scale))
+    }
+    
+    func trackTouchFrom(event: UIEvent) {
+        guard let touches = event.allTouches else { return }
+        for touch in touches {
+            guard touch.phase == .began || touch.phase == .ended || touch.phase == .moved || touch.phase == .cancelled else { continue }
+            let info = trackedTouches[touch] ?? TouchInfo(id: touchId++)
```
philipphofmann, line 45: `m`: `UITouch` doesn't conform to the `Hashable` protocol only via `NSObject`, so if I'm not mistaken the hash will be the memory address / pointer. Especially with touch events, I think two touch events might get the same memory address. What made you choose to store the touch event in a dict?

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+            let position = touch.location(in: nil).applying(scale)
+            let newEvent = TouchEvent(x: position.x, y: position.y, timestamp: event.timestamp, phase: touch.phase.toRRWebTouchPhase())
+            
+            switch touch.phase {
+            case .began:
+                info.start = newEvent
+            case .ended, .cancelled:
+                info.end = newEvent
+            case .moved:
+                if let last = info.movements.last, touchesDelta(last.point, position) < 10 { continue }
```
philipphofmann, line 55: `l`: A short comment explaining what the `touchesDelta` logic does here would be nice cause I don't get that immediately.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+    
+    private func touchesDelta(_ lastTouch: CGPoint, _ newTouch: CGPoint) -> CGFloat {
+        let dx = newTouch.x - lastTouch.x
+        let dy = newTouch.y - lastTouch.y
+        return sqrt(dx * dx + dy * dy)
+    }
+    
+    private func debounceEvents(in touchInfo: TouchInfo) {
+        guard touchInfo.movements.count >= 3 else { return }
+        let subset = touchInfo.movements.suffix(3)
```
philipphofmann, line 74: `m`: It would be great to explain the logic here a bit why we chose the last 3 items.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+        if subset[subset.startIndex + 2].timestamp - subset[subset.startIndex + 1].timestamp > 0.5 {
```
philipphofmann, line 75: `m`: Shouldn't we compare the last and the first item?
```suggestion
        if subset[subset.startIndex + 2].timestamp - subset[subset.startIndex].timestamp > 0.5 {
```

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+            //Dont debounce if the touch has a 1 second difference to show this pause in the replay.
```
philipphofmann, line 76: `l`: We actually check for 0.5 second difference, if I'm not mistaken, so this comment needs an update.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+            return
+        }
+        if arePointsCollinearSameDirection(subset[subset.startIndex].point, subset[subset.startIndex + 1].point, subset[subset.startIndex + 2].point) {
+            touchInfo.movements.remove(at: touchInfo.movements.count - 2)
+        }
+    }
+    
+    private func arePointsCollinearSameDirection(_ a: CGPoint, _ b: CGPoint, _ c: CGPoint, tolerance: CGFloat = 10) -> Bool {
```
philipphofmann, line 84: `l`: We never change the `tolerance` so we could remove the parameter.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,149 @@
...
+        let directionAB = CGPoint(x: b.x - a.x, y: b.y - a.y)
+        let directionBC = CGPoint(x: c.x - b.x, y: c.y - b.y)
+        let lengthAB = sqrt(pow(directionAB.x, 2) + pow(directionAB.y, 2))
+        let lengthBC = sqrt(pow(directionBC.x, 2) + pow(directionBC.y, 2))
+        let normalizedAB = CGPoint(x: directionAB.x / lengthAB, y: directionAB.y / lengthAB)
+        let normalizedBC = CGPoint(x: directionBC.x / lengthBC, y: directionBC.y / lengthBC)
+        let dotProduct = normalizedAB.x * normalizedBC.x + normalizedAB.y * normalizedBC.y
+        
+        // Directions are considered the same if the dot product is close to 1 (with some tolerance)
+        return abs(dotProduct - 1) < 0.001
```
philipphofmann, line 100: `m`: Out of curiosity? On which math or algorithm is this based?

philipphofmann (commented):
I didn't fully review this yet. There is plenty of logic in here, so I will give this another pass next week.

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
```diff
@@ -129,7 +129,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 scope.setTag(value: "swift", key: "language")
                
                 let user = User(userId: "1")
-                user.email = "tony@example.com"
+                user.email = "Dhiogo@example.com"
```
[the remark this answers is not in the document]
brustolin: True

Sources/Sentry/include/SentryScope+Private.h
```diff
@@ -25,6 +25,11 @@ SentryScope ()
 
 - (NSArray<SentryBreadcrumb *> *)breadcrumbs;
 
+/**
+ * Contains the breadcrumbs which will be sent with the event
+ */
+@property (atomic, strong) NSMutableArray<SentryBreadcrumb *> *breadcrumbArray;
```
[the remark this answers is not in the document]
brustolin: Thanks for the reminder

Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebTouchEvent.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+
+enum TouchEventPhase: Int {
+    case unknown = 0
+    case start   = 7
```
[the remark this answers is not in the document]
brustolin: Values defined in the RRWeb protocols, there is other values before this but we dont use it.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
brustolin: That's the only way to know if the touch is the same. According to documentation the touch object remains the same in between events. While we hold reference to it no other touch event will get the same pointer address.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
brustolin: We could, which technically would result in the same number of conversions in the long run. 
I chose not to because, with this approach, we're safeguarded against time drift—for example, if the timezone changes in the middle of the replay segment

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
brustolin: Some planar trigonometry. First half to detect if the lines are colinear (have the same angle), the second half is to detect if the lines go in the same direction.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
brustolin: I dont see why, it only matter to compare the time gap between the last two events.

.swiftlint.yml
```diff
@@ -24,7 +24,7 @@ only_rules:
   - custom_rules
   - cyclomatic_complexity
   - dynamic_inline
-  - empty_count
+  #- empty_count
```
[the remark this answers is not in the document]
brustolin: Yes

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,146 @@
...
+        var abAngle = atan2(b.x - a.x, b.y - a.y)
+        var bcAngle = atan2(c.x - b.x, c.y - b.y)
+
+        if abAngle * bcAngle < 0 { return false; }
+
+        abAngle += .pi
+        bcAngle += .pi
+
+        return abs(abAngle - bcAngle) < 0.05 || abs(abAngle - (2 * .pi - bcAngle)) < 0.05
+    }
```
brustolin, line 98: You can visually test the collinear tolerances in [here](https://jsfiddle.net/dhiogorb/8owgh1pb/3/)

.swiftlint.yml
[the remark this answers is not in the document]
philipphofmann: I would prefer removing the rules instead of commenting them out.

Sources/Sentry/SentrySessionReplay.m
```diff
@@ -289,13 +292,18 @@ - (void)captureSegment:(NSInteger)segment
     replayEvent.segmentId = segment;
     replayEvent.timestamp = videoInfo.end;
 
-    __block NSArray<SentryRRWebEvent *> *events;
+    __block NSArray<SentryBreadcrumb *> *breadcrumbs;
+    [SentrySDK.currentHub
+        configureScope:^(SentryScope *_Nonnull scope) { breadcrumbs = scope.breadcrumbs.copy; }];
```
philipphofmann, line 297: `m`: You don't need to copy the array as `scope.breadcrumbs` returns a new array anyways https://github.com/getsentry/sentry-cocoa/blob/3478fc541487bfd77886aa7d430d6242b9a6c679/Sources/Sentry/SentryScope.m#L204-L220

Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebTouchEvent.swift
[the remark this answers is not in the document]
philipphofmann: Please add this info as a comment cause I tend to forget such vital info.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philipphofmann: Please add this info as a comment.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philipphofmann: > I chose not to because, with this approach, we're safeguarded against time drift—for example, if the timezone changes in the middle of the replay segment

That makes sense. It would be great to add this info as a comment and also a short hint on why we have to use systemUptime at the beginning of this method would greatly improve the readability of this code.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -0,0 +1,146 @@
...
+            let newEvent = TouchEvent(x: position.x, y: position.y, timestamp: event.timestamp, phase: touch.phase.toRRWebTouchPhase())
+            
+            switch touch.phase {
+            case .began:
+                info.startEvent = newEvent
+            case .ended, .cancelled:
+                info.endEvent = newEvent
+            case .moved:
+                // If the distance between two points is smaller than 10 points, we don't record the second movement.
+                // iOS event polling is fast and will capture any movement; we don't need this granularity for replay.
```
philipphofmann, line 56: Thank you ❤️

.swiftlint.yml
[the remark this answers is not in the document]
brustolin: I saw other rules commented, so I follow the pattern. 
We can clean it then

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philipphofmann: Now I get it. The code now is easier to understand. Again if you add some comments it's way easier to understand.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philipphofmann: That's fancy. If the JS Fiddle is permanent it would be cool to add the link as a comment.

philipphofmann (approved):
LGTM, with mostly suggestions to improve the readability of the code.

Sources/Sentry/SentrySessionReplay.m
brustolin:
```suggestion
        configureScope:^(SentryScope *_Nonnull scope) { breadcrumbs = scope.breadcrumbs; }];
```

Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebTouchEvent.swift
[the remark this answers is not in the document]
brustolin: Line 3 above it ;)

PR #4053, opened 2024-06-12
feat: Resume session replay when app enters foreground

brustolin:
## :scroll: Description

The session replay pauses when the app enters background, now it resumes when back to the foreground.

UITest and samples

## :pencil: Checklist


Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -141,6 +139,11 @@ - (void)startWithOptions:(SentryReplayOptions *)replayOptions
                                                    name:UIApplicationDidEnterBackgroundNotification
                                                  object:nil];
 
+        [NSNotificationCenter.defaultCenter addObserver:self
```
philipphofmann, line 142: `m`: Again, I'm not a fan of using the NSNotificationCenter directly as it cause multiple problems with testing in the past. I would prefer to use `SentryNSNotificationCenterWrapper`.

philipphofmann (approved):
LGTM

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: Technically, there is no reason for NSNotificationCenter to cause any problems. 
Maybe the old flakiness was due to issues in the test. 
I will monitor this to ensure it does not become flaky.

philipphofmann: No clue. It's just what I experienced in the past.

PR #4057, opened 2024-06-12
fix: remove dead property/accessor/setter that seem to be duplicates of that in SentryCrash

armcknight:
Noticed while investigating something in the crash reporter. None of these appear in the static call graph or via textual search.

There is a parallel version at https://github.com/getsentry/sentry-cocoa/blob/d9280eec4f311096709ba84fc6a5d04423503c6c/Sources/SentryCrash/Recording/SentryCrash.h#L141 that is actually used.

Validated that crash reports are still sent using the iOS-Swift button to test a crash and checking the dashboard.

#skip-changelog


armcknight:
it also appears SentryCrashInstallation.addPreFilter is also unused, can add that to this PR.

philipphofmann (approved):
Always great to see code deleted 🗑️ 🔥

PR #4059, opened 2024-06-12
fix: remove unused DSN property in SentryNSURLRequest

armcknight:
Noticed this while debugging some network requests: I saw this _dsn ivar that was always nil. Turns out this prop is never used, the DSN is only a method parameter used in those method impls.

#skip-changelog


brustolin (approved):
Removed code where CI does not fail (apart from flakiness) should be automatically approved!!

philipphofmann (approved):
🥳

PR #4060, opened 2024-06-12
test: add more schema launch args and env vars to allow selectively enabling more sdk features

armcknight:
I needed to be able to easily disable other SDK features while debugging profiling network requests, so I added more launch args and environment variables to help zero in on exactly what I needed.

Also cleaned up a few that were previously added.

#skip-changelog

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
+++ b/Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
@@ -12,9 +12,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     //swiftlint:disable function_body_length cyclomatic_complexity
     static func startSentry() {
+        let args = ProcessInfo.processInfo.arguments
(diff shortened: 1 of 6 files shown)


Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
```diff
@@ -69,20 +68,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
             options.sessionTrackingIntervalMillis = 5_000
             options.attachScreenshot = true
             options.attachViewHierarchy = true
        
 #if targetEnvironment(simulator)
-            options.enableSpotlight = true
-            options.environment = "test-app"
+            options.enableSpotlight = !args.contains("--disable-spotlight")
 #else
-            options.environment = "device-tests"
```
armcknight, line 80: I think these environment settings were always effectively overwritten below by setting different values in options.initialScope?

philipphofmann (approved):
LGTM

PR #2440, opened 2022-11-23
fix: Reporting crashes when restarting the SDK

philipphofmann:
## :scroll: Description

The SDK did not report crashes after restarting the SDK. This is fixed by restoring the SentryCrashMonitors when installing SentryCrash after uninstalling it.

## :bulb: Motivation and Context

Fixes GH-2419

## :green_heart: How did you test it?

Manually with the steps explained in GH-2419, and some unit tests to validate the internal state of SentryCrash.
## :pencil: Checklist

- [x] I reviewed the submitted code
- [x] I added tests to verify the changes
- [x] I updated the docs if needed
- [x] Review from the native team if needed
- [x] No breaking changes

PR #4061, opened 2024-06-12
fix: reimplement crash logs written to file on disk

armcknight:
This is helpful to debug crash handling, which can't be done with the debugger attached. Reading log files is easier than getting the same info out of the system console.

I removed the file-based local level definitions, so that it's completely centralized. 

Also moved the macro def that effectively switches between console and file logging to the header so that it can be used to conditionally compile the other parts that set up file-based logging. Those pieces of code have been brought back after being removed in #2440 

#skip-changelog

### DIFF
--- a/Sources/SentryCrash/Recording/Tools/SentryCrashLogger.h
+++ b/Sources/SentryCrash/Recording/Tools/SentryCrashLogger.h
@@ -129,114 +129,138 @@
  * ===============
  *
  * The C logger changes its behavior depending on the value of the preprocessor
- * define SentryCrashLogger_CBufferSize.
+ * define @c SentryCrashLogger_CBufferSize.
  *
- * If SentryCrashLogger_CBufferSize is > 0, the C logger will behave in an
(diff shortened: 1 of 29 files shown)


philipphofmann (commented):
Looking good so far.

armcknight:
I'm not sure why I opened this as a draft, it should be ready for review. I'm not sure any of the test failures are related.

Sources/SentryCrash/Recording/Tools/SentryCrashLogger.h
```diff
@@ -129,115 +129,139 @@
...
+#define HDR_SentryCrashLogger_h
 
-#    ifdef __cplusplus
+#ifdef __cplusplus
 extern "C" {
-#    endif
+#endif
 
-#    include <stdbool.h>
+#include <stdbool.h>
```
brustolin, line 170: m: We had some problems before by declaring headers after `extern "C"`. Do you think we can move this up?

brustolin (approved):
Looks good!

philipphofmann (approved):
Again, LGTM.

Issue #4056, opened 2024-06-12
Hang detection causes app to hang


noahsmartin:
### Platform

iOS

### Environment

Production, Develop

### Installed

Swift Package Manager

### Version

N/A

### Xcode Version

N/A

### Steps to Reproduce

I've been debugging an app hang and it seems to be caused by deadlock triggered by the Sentry ANR detection. Here is the stack trace I get when sampling the threads in my app during the hang:

```
+ 7566 Thread_460801: io.sentry.app-hang-tracker
+ 7566 thread_start  (in libsystem_pthread.dylib) + 8  [0x10513a6f0]
+ 7566 _pthread_start  (in libsystem_pthread.dylib) + 104  [0x10513f4c0]
+ 7566 __NSThread__start__  (in Foundation) + 720  [0x180df3e70]
+ 7566 -[SentryANRTracker detectANRs]  (in MyApp) + 928  [0x1116fb804]
+ 7566 -[SentryANRTracker ANRDetected]  (in MyApp) + 220  [0x1116fbdf0]
+ 7566 -[SentryANRTrackingIntegration anrDetected]  (in MyApp) + 196  [0x111737924]
+ 7566 -[SentryThreadInspector getCurrentThreadsWithStackTrace]  (in MyApp) + 232  [0x1116f7d64]
+ 7566 getStackEntriesFromThread  (in MyApp) + 108  [0x1116f7718]
+ 7566 symbolicate_internal  (in MyApp) + 108  [0x11171a0d8]
+ 7566 sentrycrashdl_dladdr  (in MyApp) + 108  [0x11174f70c]
+ 7566 dyld4::APIs::_dyld_get_image_header(unsigned int)  (in dyld_sim) + 112  [0x10482f8ec]
+ 7566 dyld4::RuntimeLocks::withLoadersReadLock(void () block_pointer)  (in dyld_sim) + 56  [0x104816010]
+ 7566 _os_unfair_lock_lock_slow  (in libsystem_platform.dylib) + 204  [0x1051199c4]
+ 7566 __ulock_wait  (in libsystem_kernel.dylib) + 8  [0x10509e7dc]
```

And at the same time a different thread has this stack trace:
```
+ 7566 _os_log_impl_stream  (in libsystem_trace.dylib) + 528  [0x1800a72f0]
+ 7566 _os_activity_stream_reflect  (in libsystem_trace.dylib) + 320  [0x180098e94]
[...]

PR #4063, opened 2024-06-13
fix: Potential deadlock in app hang detection

philipphofmann:
## :scroll: Description

The app hang detection retrieves the stacktraces of all threads when it detects an app hang. The logic for retrieving the stacktrace uses sentrycrashdl_dladdr, which isn't async-safe, although it claims to be because it accesses _dyld_get_image_header, which acquires a lock. Therefore, in rare cases, retrieving the stacktrace for all threads can lead to a deadlock. We fix this by skipping symbolication, which calls sentrycrashdl_dladdr when debug is false, which we we thought we did in #3079 but somehow missed for all threads.

Fixes GH-4056

Setting debug=false for the iOS-Swift sample app to see if the symbolication still works. I wasn't able to reproduce the deadlock mentioned in GH-4056, but according to the description this fix should fix the problem.

## :pencil: Checklist


brustolin (approved):
This looks good. I just dont see how this could help in an AppHang deadlock.
Being AFAIK being async-safe is not a requirement to use in a background thread when the main thread is busy. We use objc during ANR detection which is also not async-safe

philipphofmann:
> This looks good. I just dont see how this could help in an AppHang deadlock. AFAIK being async-safe is not a requirement to use in a background thread when the main thread is busy. We use objc during ANR detection which is also not async-safe

Looking at the stracktrace posted in https://github.com/getsentry/sentry-cocoa/issues/4056 you can see that `_dyld_get_image_header` requires a lock. When we're not calling `symbolicate_internal`, we prevent acquiring a lock, which you shouldn't do when suspending all threads. Acquiring a lock with `symbolicate_internal` doesn't always end up in a deadlock but in rare cases it can.
```
+ 7566 -[SentryThreadInspector getCurrentThreadsWithStackTrace]  (in MyApp) + 232  [0x1116f7d64]
+ 7566 getStackEntriesFromThread  (in MyApp) + 108  [0x1116f7718]
+ 7566 symbolicate_internal  (in MyApp) + 108  [0x11171a0d8]
+ 7566 sentrycrashdl_dladdr  (in MyApp) + 108  [0x11174f70c]
+ 7566 dyld4::APIs::_dyld_get_image_header(unsigned int)  (in dyld_sim) + 112  [0x10482f8ec]
+ 7566 dyld4::RuntimeLocks::withLoadersReadLock(void () block_pointer)  (in dyld_sim) + 56  [0x104816010]
+ 7566 _os_unfair_lock_lock_slow  (in libsystem_platform.dylib) + 204  [0x1051199c4]
+ 7566 __ulock_wait  (in libsystem_kernel.dylib) + 8  [0x10509e7dc]
```

Is it clearer now @brustolin or am I missing something?

brustolin:
> which you shouldn't do when suspending all threads

Ow, this is the problem! Now I remember, thanks!!

PR #4064, opened 2024-06-13
chore: Pinning tvOS tests

brustolin:
Pinning tvOS tests to latest stable versions.
Github added beta simulators and those were being used as `latest`.

_#skip-changelog_


philipphofmann (approved):
LGTM

PR #4069, opened 2024-06-14
Fix swizzle loadView on UITableViewController

gabriellanata:
https://github.com/getsentry/sentry-cocoa/issues/4070

We've been getting a crash caused by Sentry in an IB file which is a subclass of `UITableViewController`. We've verified that this only happens when both `enableSwizzling` and `enableUIViewControllerTracing`, disabling either one fixes the issue.

We believe the crash is caused by sentry swizzling the `loadView` method, and the risk of this is acknowledged in a comment next to where this is done. This function has a check to test if the `loadView` method was overridden or not by comparing it to the default implementation from `UIViewController`. However, this does not include other subclasses of `UIViewController` from UIKit like `UITableViewController` that could have a different implementation internally than the one in UIViewController.

The proposed fix is to also check for `UITableViewController` in this function. If the theory mentioned above is correct, it means that just checking for these two might not be enough and it could be necessary to check for all possible view controllers provided by UIKit.

This fix has not been tested


brustolin:
Thank you very much for the help.
We will close this PR in favour of #4071 that its a broader solution.

PR #4073, opened 2024-06-14
fix(sessions): `storeEnvelope` updates session when passed unhandled event

krystofwoldrich:
## :scroll: Description

This PR add session update based on the stored envelope. 

Stored envelope with `handled: false` ends session as crashed and doe not start a new session. This is important because store envelope will be called by RN only for unhandled errors (not user set but based on RN error handler) these error are re-thrown by the JS engine on the native layers which will cause the application to crash.

~If the stored envelope from RN is actual crash the application crashes, when the app is opened a new session is started? Is this wanted behaviour? Does it matches native crashes?~

**RN on Android** `handled: false` increments errors count. When `handled:false` does not crash the app the same session is being updated. This is unwanted and will be updated in https://github.com/getsentry/sentry-java/pull/3480

- related PR: https://github.com/getsentry/sentry-cocoa/pull/3091
- related Android fix https://github.com/getsentry/sentry-java/pull/3480

- fixes: https://github.com/getsentry/sentry-react-native/issues/3614

RN sample app, unit tests

## :pencil: Checklist


Sources/Sentry/SentrySDK.m
```diff
@@ -385,7 +385,10 @@ + (void)captureEnvelope:(SentryEnvelope *)envelope
 + (void)storeEnvelope:(SentryEnvelope *)envelope
 {
     if (nil != [SentrySDK.currentHub getClient]) {
-        [[SentrySDK.currentHub getClient] storeEnvelope:envelope];
+        [[SentrySDK.currentHub getClient]
```
philipphofmann, line 388: `m`: What about adding a method `storeEnvelope` to the hub, which calls the `storeEnvelope` on the client and updates the session state? Then we don't need to expose the `updateSessionState` and we have the session logic encapsulated in the hub.

Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
```diff
@@ -3,6 +3,21 @@ import XCTest
 
 class PrivateSentrySDKOnlyTests: XCTestCase {
 
+    private class Fixture {
+        static func getUnhandledExceptionEnvelope() -> SentryEnvelope {
```
philipphofmann, line 7: `m`: We're not using the `Fixtures` pattern anymore, because it's so verbose. I would just add `getUnhandledExceptionEnvelope` as a method to `PrivateSentrySDKOnlyTests` without creating the extra class.

Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
```diff
@@ -18,7 +33,19 @@ class PrivateSentrySDKOnlyTests: XCTestCase {
         XCTAssertEqual(1, client?.storedEnvelopeInvocations.count)
         XCTAssertEqual(envelope, client?.storedEnvelopeInvocations.first)
     }
+    
+    func testStoreEnvelopeWithUndhandledMarksSessionAsCrashedAndDoesNotStartNewSession() {
```
philipphofmann, line 37: `l`: To slightly improve the readability:
```suggestion
    func testStoreEnvelopeWithUndhandled_MarksSessionAsCrashedAndDoesNotStartNewSession() {
```

Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
```diff
@@ -29,6 +56,18 @@ class PrivateSentrySDKOnlyTests: XCTestCase {
         XCTAssertEqual(1, client?.captureEnvelopeInvocations.count)
         XCTAssertEqual(envelope, client?.captureEnvelopeInvocations.first)
     }
+    
+    func testcaptureEnvelopeWithUndhandledMarksSessionAsCrashedAndStartsNewSession() {
```
philipphofmann, line 60: `l`:
```suggestion
    func testCaptureEnvelopeWithUndhandled_MarksSessionAsCrashedAndStartsNewSession() {
```

Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
```diff
@@ -29,6 +56,18 @@ class PrivateSentrySDKOnlyTests: XCTestCase {
...
+        let client = TestClient(options: Options())
+        let hub = TestHub(client: client, andScope: nil)
+        SentrySDK.setCurrentHub(hub)
+        hub.setTestSession()
+
+        let envelope = Fixture.getUnhandledExceptionEnvelope()
+        PrivateSentrySDKOnly.capture(envelope)
+
+        XCTAssertEqual(1, hub.startSessionInvocations)
```
philipphofmann, line 69: `m`: This test doesn't validate that the SDK marks the session as crashed. It would be great if it did.

Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
```diff
@@ -3,6 +3,21 @@ import XCTest
 
 class PrivateSentrySDKOnlyTests: XCTestCase {
 
+    private class Fixture {
```
philipphofmann, line 6: `m`: What made you choose to add these tests here instead of adding them to SentrySDKTests or SentryHubTests where the logic is?

philipphofmann (commented):
Thanks for fixing this in the Cocoa SDK.

Sources/Sentry/SentryHub.m
```diff
@@ -637,10 +632,13 @@ - (void)captureEnvelope:(SentryEnvelope *)envelope
...
     }
 
-    [client captureEnvelope:[self updateSessionState:envelope]];
+    // If captured envelope cointains not handled errors, these are not going to crash the app and
+    // we should create new session.
+    [client captureEnvelope:[self updateSessionState:envelope andStartNewSession:YES]];
 }
 
 - (SentryEnvelope *)updateSessionState:(SentryEnvelope *)envelope
+                    andStartNewSession:(BOOL)startNewSession
```
brustolin, line 641: l:
```suggestion
                    startNewSession:(BOOL)startNewSession
```

Sources/Sentry/include/SentryHub+Private.h
```diff
@@ -62,6 +62,9 @@ SentryHub ()
 
 - (void)captureEnvelope:(SentryEnvelope *)envelope;
 
+- (SentryEnvelope *)updateSessionState:(SentryEnvelope *)envelope
+                    andStartNewSession:(BOOL)startNewSession;
```
brustolin, line 66: l:
```suggestion
- (SentryEnvelope *)updateSessionState:(SentryEnvelope *)envelope
                    startNewSession:(BOOL)startNewSession;
```

brustolin (commented):
It looks good. Just some nitpicks on top of Philipp already commented

Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
krystofwoldrich: Because this is the interface hybrid SDKs use. 

The underlying structure can change as long as session is handled as expected.

brustolin (approved):
LGTM

PR #4071, opened 2024-06-14
Swizzling of view controllers `loadView` that don`t implement `loadView`

brustolin:
## :scroll: Description

Stop swizzling `loadView` function of UIViewControllers that doesn't implement it.

## :bulb: Motivation and Context

close #4070

## :green_heart: How did you test it?

In the sample.
Current ui tests already tackles this. 

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I reviewed the submitted code.
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

Issue #4070, opened 2024-06-14
Crash when initializing a UITableViewController from IB


gabriellanata:
### Platform

iOS

### Environment

Production, Develop, TestFlight

### Installed

Manually

### Version

8.26.0

### Xcode Version

15.4

### Did it work on previous versions?

Unknown, but probably not

### Steps to Reproduce

1. Create a UITableViewController subclass that uses a XIB file
2. Add a view to the xib file and connect it using an @IBOutlet
3. Add some code in the `viewDidLoad` method that checks if the @IBOutlet is nil or not
4. Ensure `enableSwizzling` and `enableUIViewControllerTracing` are enabled as options in Sentry
5. Run the app and navigate to that screen

### Expected Result

@IBOutlet is not nil and everything worked as expected

### Actual Result

@IBOutlet is nil and it caused a crash if the IBOutlet uses force unwrap

### Are you willing to submit a PR?

https://github.com/getsentry/sentry-cocoa/pull/4069

brustolin:
Hello @gabriellanata, thanks for bringing this up.
We will provide a fix soon.

gabriellanata:
Thanks @brustolin, verified the crash is fixed on our side after updating the library

PR #4074, opened 2024-06-14
chore: Changed TableViewController to have a Nib file

brustolin:
TableViewController is now initialized from a Nib file with an extra view. 
This is an additional test for the changes introduced in #4071.

_#skip-changelog_


armcknight:
I don't quite get why we stopped swizzling `loadView` in #4071... Testing on this branch, I set a symbolic breakpoint for `loadView` and then navigated to this new table controller based on the XIB, and `loadView` is still called even though there's no override for it in the implementation. 

I also added it back to the list of methods to swizzle and there was no crash. But this test doesn't follow the repro steps outlined in https://github.com/getsentry/sentry-cocoa/issues/4070, nor does it seem to follow the description of the comment [here](https://github.com/getsentry/sentry-cocoa/pull/4071/files?diff=split#diff-91eb99a8ef5359300fa41d03957b1f6c6b509e85cabfc2917da6d2deb5d54c10R351-R357). Is that intentional?

brustolin:
> and loadView is still called even though there's no override for it in the implementation

This is the UITableViewController `loadView` implementation that was swizzled, and not or own `TableViewController`, that has no `loadView` implementation.

> I also added it back to the list of methods to swizzle and there was no crash.

Im not sure what do you mean. `loadView` was not removed from any list of functions we swizzle.

The fix is at `SentryUIViewControllerSwizzling.m` line 359. To test it, you may change it back to:
```objc
IMP viewControllerImp = class_getMethodImplementation([UIViewController class], selector);
```

philipphofmann (approved):
LGTM

armcknight:
I see what you mean now, I missed that UIAssert.swift is part of the sample apps and not the SDK.
image

PR #4076, opened 2024-06-14
fix: guard against null thread handle for profiler sampling thread

armcknight:
Addresses the conversation starting here: https://github.com/getsentry/sentry-cocoa/pull/3520#discussion_r1439254785

#skip-changelog


Sources/Sentry/SentryBacktrace.cpp
```diff
@@ -63,6 +63,7 @@ namespace profiling {
             }
         }
         std::uintptr_t current;
+        // ???: is this to avoid ever gathering a backtrace for the profiler's sampling thread?
```
armcknight, line 66: @indragiek I was wondering if we should just move all these types of checks up and out of here, to earlier callsites.

Sources/Sentry/SentrySamplingProfiler.cpp
```diff
@@ -64,7 +64,19 @@ namespace profiling {
...
 
                 params->numSamples.fetch_add(1, std::memory_order_relaxed);
-                enumerateBacktracesForAllThreads(params->callback, params->cache);
+
+                const auto current = ThreadHandle::current();
+                if (current == nullptr) {
+                    SENTRY_PROF_LOG_WARN(
+                        "Current thread (the profiler's sampling thread) handle returned as null, "
+                        "will not attempt to gather further backtraces.");
+                    break; // ???: i originally had this as continue, but if this thread's handle is
```
armcknight, line 73: @indragiek wdyt?

philipphofmann, line 73: Thanks for the comment.

Sources/Sentry/SentryBacktrace.cpp
```diff
@@ -103,16 +104,29 @@ namespace profiling {
...
             // This one is probably safe to call while the thread is suspended, but
             // being conservative here in case the platform time functions take any
             // locks that we're not aware of.
             bt.absoluteTimestamp = getAbsoluteTime();
 
+            if (thread == nullptr) {
+                SENTRY_PROF_LOG_WARN(
+                    "Thread handle was null, will not attempt to gather its backtrace.");
+                continue;
+            }
```
philipphofmann, line 128: `l`: Any reason we can't put this at the beginning of the loop / before `Backtrace bt;` to avoid allocating the `Backtrace` struct?

philipphofmann (approved):
LGTM, but I think we should still wait for @indragiek before merging this.

Sources/Sentry/SentryBacktrace.cpp
armcknight: yep, i didn't look up enough. that's better for sure.

Sources/Sentry/SentrySamplingProfiler.cpp
indragiek: Yeah I think this makes sense

PR #4085, opened 2024-06-18
feat: Restart replay session with mobile session

brustolin:
## :scroll: Description

Starts a new replay session when the mobile session is restarted. Usually when the user stays away from the app for more than 60 seconds.

Unit tests

## :pencil: Checklist


Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -186,6 +198,7 @@ - (SentryIntegrationOption)integrationOptions
 
 - (void)uninstall
 {
+    [SentrySDK.currentHub unregisterSessionListener:self];
```
philipphofmann, line 201: `m`: Maybe it's good to call `uninstall` also from `dealloc` so we ensure to clear this up when this class gets deallocated.

Sources/Sentry/SentryHub.m
```diff
@@ -44,6 +44,7 @@
 @property (nonatomic, strong) SentryCrashWrapper *crashWrapper;
 @property (nonatomic, strong) NSMutableSet<NSString *> *installedIntegrationNames;
 @property (nonatomic) NSUInteger errorsBeforeSession;
+@property (nonatomic) NSMutableArray<id<SentrySessionListener>> * sessionListeners;
```
philipphofmann, line 47: `m`: Will there be multiple listeners in the foreseeable future? If not, maybe we can replace this with a single weak reference cause it simplifies things. If you want to keep the array, maybe it's smarter to replace this with weak references via `[NSHashTable weakObjectsHashTable]` to avoid potential retain cycles. You can check `SentryANRTracker` and `SentryFramesTracker` for reference.

Sources/Sentry/SentryHub.m
```diff
@@ -129,10 +132,15 @@ - (void)startSession
 
         [self storeCurrentSession:_session];
         [self captureSession:_session];
+        listeners = [_sessionListeners copy];
```
philipphofmann, line 135: 💯 for keeping the synchronized block small.

Sources/Swift/Protocol/SentrySessionListener.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+@_implementationOnly import _SentryPrivate
+
+@objc
+protocol SentrySessionListener : NSObjectProtocol
```
philipphofmann, line 5: 🥇 for making this in Swift.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -1,5 +1,4 @@
 import Foundation
-import Nimble
```
philipphofmann, line 2: Thanks for getting rid of Nimble 👍

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -32,28 +31,38 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
         clearTestState()
     }
     
-    func startSDK(sessionSampleRate: Float, errorSampleRate: Float, enableSwizzling: Bool = true) {
+    private func getSut() -> SentrySessionReplayIntegration? {
+        guard let sut = SentrySDK.currentHub().installedIntegrations().first as? SentrySessionReplayIntegration else {
+            XCTFail("Did not installed replay integration")
+            return nil
+        }
+        return sut
```
philipphofmann, line 39: `l`: You could also use `XCTUnwrap` to avoid returning `nil`, but it will change this method to a `throws`.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -68,67 +77,90 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
...
+    
+    func testRestartReplayWithNewSession() {
+        startSDK(sessionSampleRate: 1, errorSampleRate: 0)
+        
+        guard let sut = getSut() else { return }
+        XCTAssertNotNil(sut.sessionReplay)
+        SentrySDK.currentHub().endSession()
+        XCTAssertNil(sut.sessionReplay)
+        SentrySDK.currentHub().startSession()
+        XCTAssertNotNil(sut.sessionReplay)
```
philipphofmann, line 162: `l`: Does it make sense to do some asserts here to validate the `sessionReplay`?

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -98,58 +108,49 @@ - (void)startWithOptions:(SentryReplayOptions *)replayOptions
...
+               touchTracker:_touchTracker
+               dateProvider:SentryDependencyContainer.sharedInstance.dateProvider
+                     random:SentryDependencyContainer.sharedInstance.random
+         displayLinkWrapper:[[SentryDisplayLinkWrapper alloc] init]];
+
+    [self.sessionReplay
+              start:SentryDependencyContainer.sharedInstance.application.windows.firstObject
+        fullSession:[self shouldReplayFullSession:replayOptions.sessionSampleRate]];
+
+    [NSNotificationCenter.defaultCenter addObserver:self
```
philipphofmann, line 145: `m`: Some other code in this class already uses the `NotificationCenterWrapper`. It would be great to stick to one of the two.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -162,6 +163,17 @@ - (void)resume
     [self.sessionReplay resume];
 }
 
+- (void)sentrySessionEnded:(SentrySession *)session {
+    [self stop];
+    [NSNotificationCenter.defaultCenter removeObserver:self];
```
philipphofmann, line 168: `m`: Please only use `removeObserver self` in `dealloc`. Instead, remove it with the most specific detail possible.

See: https://developer.apple.com/documentation/foundation/nsnotificationcenter/1413994-removeobserver#discussion
> When removing an observer, remove it with the most specific detail possible. For example, if you used a name and object to register the observer, use [removeObserver:name:object:](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1407263-removeobserver) with the name and object.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -44,38 +44,48 @@ - (BOOL)installWithOptions:(nonnull SentryOptions *)options
...
+             selector:@selector(newSceneActivate)
+             name:UISceneDidActivateNotification];
         }
-
-        return YES;
-    } else {
-        return NO;
     }
+    
+    return;
```
philipphofmann, line 88: `l`: I guess we can ditch that.
```suggestion
```

philipphofmann (commented):
I added a few comments.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
brustolin: true. I mostly forget this. I think its a good idea because it will increase coverage.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
brustolin: I don't think so. if it was initialized is all good, SR behaviour has its own tests.

philipphofmann (approved):
Thanks for including all the feedback, LGTM 🚀

Issue #4020, opened 2024-05-28
Logging improvements


armcknight:
### Description

As we continue developing the SDK, we might find the amount of logs emitted (esp in debug mode) grows unwieldy.

There are a couple things to consider implementing:

1. component-based logging (and/or log levels). the way i've seen this done in the past is to allow enabling either all logs, or to disable all logs and then be able to selectively enable logs only in particular code files. this is easier with macros in objc, not sure how it'd work with swift. maybe we would want something more at the "module" level: e.g. profiling has more than one file we'd want logs from.
2. another more verbose log level like "trace" to put things that either output at high frequency or output large amounts of data per message, like dumping all configuration options or the entire contents of each envelope.

With (2), we'd eventually want: 

3. a way to avoid evaluating the expressions in a log statement if it is below the currently configured log level. No point allocating and building a huge string just to throw it away. This would have a positive perf impact today as-is.

philipphofmann:
I fixed point 3 with https://github.com/getsentry/sentry-cocoa/pull/4028.

armcknight:
I just thought of one more point:

6: provide other log sinks besides just NSLog. For two reasons:

- NSLog is not async-safe; it causes crashes if I try to log from the profiler's thread, eg. Using `printf()` and `write()` are both async-safe, so separate sinks for those would be nice.
- in the case of writing to a file instead of the console, we can debug a wider variety of cases, like crash reports (see point (4) above, the current logging interface is kind of clunky) or retrieving logs from customers.

armcknight:
The chain of PRs starting with #4101 helps with point (4)

armcknight:
I think this hasn't really been an issue in practice, especially with Xcode's modern log console filtering options. Will close this as unplanned, maybe we'll want a specific "log sinks" task in the future.

PR #4088, opened 2024-06-18
ref(logging): provide alternative log output sinks

armcknight:
I need alternative destinations for certain logging tasks besides just NSLog, like printf or files. This refactors the current logging API to allow for providing a combination of those three, defaulting to file and NSLog.

It converts SentryLog's API from class-bound methods to instance-bound, which simplifies a couple things in the tests and the codebase, and will allow for having more than one logger instance, so that the other places that need an alternative can set up their own logger.

The default logging setup is still provided via a singleton, which we already did implicitly, so this is just made more explicit.

Helps with #4020; #skip-changelog

TODO:
- [ ] expose C functions for writing to `printf()` and files (using `write()`) so that contexts that require async-safety can call them directly, bypassing the objc apparatus. Need to figure out how to make this play nicely with configured log levels.


philipphofmann (commented):
Is it possible to write the new classes in Swift?

armcknight:
> Is it possible to write the new classes in Swift?

The next step was going to be factoring out the implementation into C functions that can be called outside of the objc path so we can use them from async contexts, so I'm not sure.

armcknight:
I'm going to close this and work on something slightly different related to logging, because this is getting too deep into file-based logging issues like log size/rotation etc that I didn't initially plan to take on. 

What I really wanted was to be able to use an async-safe logger from more of the profiling code, since there's still some stuff there that winds up using NSLog, which is unsafe from the logging thread.

What I'm going to do there is repurpose the crash reporter logger and reuse it there.

PR #4089, opened 2024-06-19
chore: cleanup replay exports for hybrid SDKs

vaind:
## :scroll: Description

This exposes a define `SENTRY_TARGET_REPLAY_SUPPORTED` (there's already `SENTRY_TARGET_PROFILING_SUPPORTED` so keeping the name format) for hybrid SDKs so they don't have to copy the specific check `SENTRY_HAS_UIKIT && !TARGET_OS_VISION` which may change in the future, e.g. when replay for macOS is made available. Also, while adding this, I've noticed that some replay functions were exported & then logged a warning if replay wasn't available, while some were not exported at all (see link below for the RN PR). I've unified this so that no functions are exported for hybrid SDKs if replay is not available.

Came up in a compilation fail for RN https://github.com/getsentry/sentry-react-native/pull/3846#issuecomment-2177952419

All the original code & tests should pass. Except for tests that tried if the interfaces didn't throw when replay isn't available, obviously, since those are removed now.

## :pencil: Checklist

## :crystal_ball: Next steps

Use this new define in RN and Flutter.

#skip-changelog because this change isn't for normal SDK consumers.

### DIFF
--- a/Sources/Sentry/PrivateSentrySDKOnly.mm
+++ b/Sources/Sentry/PrivateSentrySDKOnly.mm
@@ -3,7 +3,6 @@
 #import "SentryBreadcrumb+Private.h"
 #import "SentryClient.h"
 #import "SentryDebugImageProvider.h"
-#import "SentryDefines.h"
 #import "SentryExtraContextProvider.h"
 #import "SentryHub+Private.h"
 #import "SentryInstallation.h"
--- a/Sources/Sentry/PrivateSentrySDKOnly.mm
+++ b/Sources/Sentry/PrivateSentrySDKOnly.mm
@@ -3,6 +3,7 @@
 #import "SentryBreadcrumb+Private.h"
 #import "SentryClient.h"
 #import "SentryDebugImageProvider.h"
+#import "SentryDefines.h"
 #import "SentryExtraContextProvider.h"
 #import "SentryHub+Private.h"
 #import "SentryInstallation.h"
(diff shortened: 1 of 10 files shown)


Sources/Sentry/PrivateSentrySDKOnly.mm
```diff
@@ -320,31 +321,19 @@ + (nullable SentrySessionReplayIntegration *)getReplayIntegration
...
 
 + (void)captureReplay
 {
-#if SENTRY_HAS_UIKIT && !TARGET_OS_VISION
     [[PrivateSentrySDKOnly getReplayIntegration] captureReplay];
-#else
-    SENTRY_LOG_DEBUG(
-        @"SentrySessionReplayIntegration only works with UIKit enabled and target is "
-        @"not visionOS. Ensure you're using the right configuration of Sentry that links UIKit.");
-#endif
```
vaind, line 333: I've removed these because they couldn't get ever triggered anyway - the header doesn't define these functions when replay is not available.

Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
```diff
@@ -163,7 +163,8 @@ typedef void (^SentryOnAppStartMeasurementAvailable)(
 
 #endif // SENTRY_UIKIT_AVAILABLE
 
-#if SENTRY_HAS_UIKIT && !TARGET_OS_VISION
+#if SENTRY_TARGET_REPLAY_SUPPORTED
++ (SentrySessionReplayIntegration *__nullable)getReplayIntegration;
```
brustolin, line 167: m: Why do you need the integration? The `configureSessionReplay` is not enough?

brustolin (approved):
Thanks!! Easier to maintain this way.

Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
vaind: we probably don't, removing from the header.
```suggestion
```

krystofwoldrich (approved):
Looks good, will be much easier in the future, but we should still log out when the replay is not available in RN code. 

Thank you.

philipphofmann (approved):
Thank you @vaind 💯

PR #4096, opened 2024-06-20
fix: take copy of SentryOptions on SDK start

armcknight:

I've seen questions over time about what happens when a customer changes a value in the options object after starting the SDK. The response is always that it's not supported, yet it is possible in the code, allowing misconfiguration to occur. This will help prevent surprising effects (even though it might itself be surprising that changes are reflected–this will need to be documented better).

Old tests pass and modified where needed. New tests written to ensure that any time a new option is added, that it is also copied correctly.

### DIFF
--- a/Sentry.xcodeproj/xcshareddata/xcschemes/Sentry.xcscheme
+++ b/Sentry.xcodeproj/xcshareddata/xcschemes/Sentry.xcscheme
@@ -104,6 +104,9 @@
                   Identifier = "SentrySubClassFinderTests/testActOnSubclassesOfViewController()">
                </Test>
                <Test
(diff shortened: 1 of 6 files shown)


philipphofmann (commented):
Some customers may rely on this functionality, so we have to make this a breaking change, IMO. We recently discussed whether mobile SDKs should support changing options on the fly for remote config and starting native SDKs before hybrid SDKs without a manual initialization of the native SDKs (cc @krystofwoldrich). Therefore, I'm unsure if we should change this right now.

armcknight:
I understand your point about the breakage. If and when the time comes, we can architect everything you've mentioned while still keeping options immutable, if that's what we think is the best design choice.

PR #4099, opened 2024-06-21
Chore: Exposing SR to hybrid SDK

brustolin:
Private Swift classes cannot be exposed to Hybrid SDK.

Creating a interface to intermediate the usage

_#skip-changelog_


Sentry.xcodeproj/project.pbxproj
brustolin, line 1: This building settings changes are related to `SentryTestUtils` target.

Sources/Sentry/include/HybridPublic/SentrySessionReplayIntegration-Hybrid.h
```diff
@@ -16,4 +18,24 @@
 
 @end
 
+@interface
+SentrySessionReplayIntegration ()
+
++ (id<SentryRRWebEvent>)createBreadcrumbwithTimestamp:(NSDate *)timestamp
+                                             category:(NSString *)category
+                                              message:(NSString *)message
```
vaind, line 26: message should be nullable here, same as in SentryRRWebBreadcrumbEvent initWithTimestmap

Sources/Swift/Integrations/SessionReplay/SentrySRDefaultBreadcrumbConverter.swift
```diff
@@ -3,7 +3,8 @@ import Foundation
 
 @objc
 protocol SentryReplayBreadcrumbConverter: NSObjectProtocol {
-    func convert(breadcrumbs: [Breadcrumb], from: Date, until: Date) -> [SentryRRWebEvent]
+    func convert(breadcrumbs: [Breadcrumb], from: Date, until: Date) -> [SentryRRWebEventProtocol]
+    func convert(from breadcrumb: Breadcrumb) -> SentryRRWebEventProtocol?
```
vaind, line 7: Why do we need to define both now? Did you mean to remove the original method and replace it with a single-breadcrumb one? I'd be in favor of that. 
Otherwise, this change makes no sense on the consumer side because `convertFromBreadcrumb` would never be called anyway.

vaind (changes requested):

Sources/Swift/Integrations/SessionReplay/SentrySRDefaultBreadcrumbConverter.swift
brustolin: Yes. I know this is weird.
For `func convert(breadcrumbs: [Breadcrumb], from: Date, until: Date)` just copy the implementation from `SentrySRDefaultBreadcrumbConverter`

I will fix all of this when I convert `SentrySessionReplay` to Swift. I don't want to deal with this things in Objc right now.

vaind: in case it helps, here's the objc implementation of the filter:

```objective-c

- (NSArray<SentryRRWebEvent *> *_Nonnull)
    convertWithBreadcrumbs:(NSArray<SentryBreadcrumb *> *_Nonnull)breadcrumbs
                      from:(NSDate *_Nonnull)from
                     until:(NSDate *_Nonnull)until {
  NSMutableArray<SentryRRWebEvent *> *outBreadcrumbs = [NSMutableArray array];
rrwebBreadcrumb;
  for (SentryBreadcrumb *breadcrumb in breadcrumbs) {
    // - (NSComparisonResult)compare:(NSDate *)other;
    // If:
    // The receiver and `other` are exactly equal to each other, NSOrderedSame
    // The receiver is later in time than `other`, NSOrderedDescending
    // The receiver is earlier in time than `other`, NSOrderedAscending.
    if (!breadcrumb.timestamp ||
        [breadcrumb.timestamp compare:from] == NSOrderedAscending ||
        [breadcrumb.timestamp compare:until] == NSOrderedDescending) {
      continue;
    }

    SentryRRWebEvent * rrwebBreadcrumb = [selfconvertFrom:breadcrumb];
    if (rrwebBreadcrumb) {
      [outBreadcrumbs addObject:rrwebBreadcrumb];
    }
  }
  return outBreadcrumbs;
}

brustolin: You won't let this one go, right? 🤣

vaind: thx :)

vaind (approved):
lgtm

krystofwoldrich:
@brustolin @vaind  Is this PR working with https://github.com/getsentry/sentry-react-native/pull/3846 ?

I've just tried it but I'm not getting any replays.

brustolin:
> @brustolin @vaind Is this PR working with https://github.com/getsentry/sentry-react-native/pull/3846 ?

No, this PR changes the way RN will convert breadcrumbs

PR #4101, opened 2024-06-21
ref(logging): refactor SentryCrashLog into SentryAsyncSafeLog

armcknight:
Picks up more in the direction I wanted to go in vs https://github.com/getsentry/sentry-cocoa/pull/4088

This prepares the implementation currently from the KSCrash fork for more general async-safe use cases, by renaming things. This is also required by some of the profiler stuff, for example. There should be no changes to any actual implementation logic.

Followed by #4102, which fixes the build error in this branch related to it being unable to find one of the ObjC log methods: 4102 removes the objc portion of the old crash logger entirely, replacing its usages with SentryLog.

#skip-changelog


brustolin (commented):
I have reviewed all the other PRs, and they all seem good.

I would like to take another look when everything is merged back into this one and the tests are fixed.

armcknight:
@brustolin Please see https://github.com/getsentry/sentry-cocoa/pull/4113 which fixes the test breakages, then i'll merge that into this, and merge forward into all the other branches to make sure their checks all pass afterwards

brustolin (approved):
This is awesome!

PR #4106, opened 2024-06-22
ref(logging): simplify async logger configuration

armcknight:
following #4105, followed by #4107

This one simplifies how are the async logger is configured. It used to allow configuration from the build setting in GCC_PREPROCESSOR_MACROS, as well as a separate "local" level per file that used the logger. Now only allow configuring one log level from the logger's source code file.

#skip-changelog


brustolin (approved):
Im all up for reducing unnecessary complexity!! 🚀

PR #4108, opened 2024-06-22
ref(logging): initialize the async logger earlier

armcknight:
following #4107, change how old the async logger is initialized. It used to be done from the crash reporter initialization, but now that it can be used from other contexts, like launch profiling, we need to initialize it along with those parts of the SDK. Now we do it when setting up the regular Objective-C logger. #skip-changelog


Sources/Sentry/SentryAsyncSafeLog.c
```diff
@@ -132,7 +129,7 @@ setLogFD(int fd)
     g_fd = fd;
 }
 
-bool
+int
```
armcknight, line 132: This wasn't playing nicely with the macro to log errno values

Sources/Sentry/SentryAsyncSafeLog.h
```diff
@@ -69,13 +71,10 @@ void sentry_asyncLogC(
...
  *
  * @param filename The file to write to (NULL = write to stdout).
- *
  * @param overwrite If true, overwrite the log file.
+ * @return 0 if successful, 1 otherwise.
  */
-bool sentry_asyncLogSetFileName(const char *filename, bool overwrite);
-
-/** Clear the log file. */
-bool sentry_asyncLogClearLogFile(void);
```
armcknight, line 78: This was unused even before any of this chain of refactors.

Sources/Sentry/include/SentryLog.h
```diff
@@ -43,7 +43,7 @@ NS_ASSUME_NONNULL_END
 #define SENTRY_LOG_ERRNO(statement)                                                                \
     ({                                                                                             \
         errno = 0;                                                                                 \
-        const auto __log_rv = (statement);                                                         \
```
armcknight, line 46: This was the other thing that wasn't working correctly. This macro used to only be called from C++ contexts, but now that it's called from the Objective-C logger to initialize the async clogger, we can't use the auto keyword.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitorContext.h
```diff
@@ -220,9 +220,6 @@ typedef struct SentryCrash_MonitorContext {
         const char *reason;
     } ZombieException;
 
-    /** Full path to the console log, if any. */
-    const char *consoleLogPath;
```
armcknight, line 224: This is only ever used in one place, and at one time in a process run. I just moved the access to be done directly if that spot.

Sources/SentryCrash/Recording/SentryCrashReport.c
```diff
@@ -1598,10 +1598,7 @@ writeDebugInfo(const SentryCrashReportWriter *const writer, const char *const ke
 {
     writer->beginObject(writer, key);
     {
-        if (monitorContext->consoleLogPath != NULL) {
-            addTextLinesFromFile(
-                writer, SentryCrashField_ConsoleLog, monitorContext->consoleLogPath);
-        }
+        addTextLinesFromFile(writer, SentryCrashField_ConsoleLog, g_logFilename);
```
armcknight, line 1601: Here's the direct access of the async log path.

brustolin (approved):

PR #4113, opened 2024-06-24
ref(logging): fold SentryCrashSystemCapabilities.h into SentryInternalCDefines.h

armcknight:
fixes the test failures in #4101, i had started rolling SentryCrashSystemCapabilities.h into SentryInternalCDefines.h, but stopped halfway, so a bunch of things weren't compiling right. I just finished that job and cleaned up some defines that were never used

#skip-changelog


brustolin (approved):
Awesome. Reducing complexity and improving readability is always good.

My only concern when renaming the macros is that it will probably make it harder to apply any updates from the original KSCrash.

armcknight:
I understand the concern, but what I gathered from asking about this recently, we haven't directly applied patches from KSCrash for a long time now. If we really want that kind of process, I'm open to it, but I would rather do it with a real fork, managed as a dependency of Sentry using package manager specs, instead of it being copied into our codebase, so that we could merge/rebase our own patches along with new ones from upstream. 

Barring that, I think it's fair to assume it's in a pretty stable state, and that if we do really need to apply some emergency bugfix, even if we have to do something more manual, it should still be straightforward.

PR #4120, opened 2024-06-25
test: remove Nimble

armcknight:
This was added as a team experiment, but recently we decided not to move forward with it.

Lately, it has been causing me pain on every branch switch. Because SPM still has so many problems in Xcode, I would have to manually re-resolve packages to be able to run tests again. No network connection? Well, now you can't resolve the packages, so local testing ¯\\_(ツ)_/¯ This is the only thing we use SPM for in our SDK project, so will be a marked improvement to jettison.

There are several other reasons we decided not to move forward with Nimble, but I'm not going to repeat them here at this time.

Contains a couple other small fixes, like converting some more IUOs to `try XCTunwrap`, `XCTAssertTrue`->`XCTAssert`, `XCTAssertEqual(..., nil)`->`XCTAssertNil` etc

#skip-changelog


brustolin (approved):
Thanks!!

philipphofmann:
Thanks a lot @armcknight 🥇

PR #1784, opened 2022-04-28
feat: UI event transactions for clicks

philipphofmann:
## :scroll: Description

Creates transactions when the user clicks an UIButton, UISegmentedControl, UIBarButtonItem or UIPageControl. For more details, see https://github.com/getsentry/team-mobile/issues/4.

## :bulb: Motivation and Context

https://github.com/getsentry/team-mobile/issues/4

## :green_heart: How did you test it?
Unit tests, simulator, and real devices.

## :pencil: Checklist

- [x] I reviewed the submitted code
- [x] I added tests to verify the changes
- [ ] I updated the docs if needed
- [x] Review from the native team if needed
- [x] No breaking changes

PR #4123, opened 2024-06-26
ref: remove unused constant

armcknight:
I noticed this hanging around at some point.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryOptions.m
+++ b/Sources/Sentry/SentryOptions.m
@@ -18,6 +18,7 @@
 #import "SentrySessionReplayIntegration.h"
 #import "SentrySwift.h"
 #import "SentrySwiftAsyncIntegration.h"
+#import "SentryTracer.h"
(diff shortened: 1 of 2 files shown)


brustolin (approved):
Just one question. 
The const is not being used, but should we use it?

armcknight:
> Just one question. The const is not being used, but should we use it?

So looking at the git blame for the definition of the constant: https://github.com/getsentry/sentry-cocoa/blob/038edae2a4e56af76056699cf6e8102bbf6f7f5a/Sources/Sentry/include/SentryTracer.h#L19

and the hardcoded value of this option's default: https://github.com/getsentry/sentry-cocoa/blob/038edae2a4e56af76056699cf6e8102bbf6f7f5a/Sources/Sentry/SentryOptions.m#L117

they both came from #1784. I guess it was just forgotten to actually assign the constant, so this PR now does that.

PR #4125, opened 2024-06-27
test: convert some test code from crashable to failable

### LINKED ISSUE #3576: Remove force unwraps from test code

armcknight:
Helps fix #3576 

I hit one of these today so just went through and removed some easy ones. There are still almost 603 uses of the force unwrap operator and 154 unchecked array indexes in the test code after these changes.

image
image

#skip-changelog

### DIFF
--- a/Tests/SentryProfilerTests/SentryTraceProfilerTests.swift
+++ b/Tests/SentryProfilerTests/SentryTraceProfilerTests.swift
@@ -428,10 +428,8 @@ private extension SentryTraceProfilerTests {
         }
     }
 
-    func printTimestamps(entries: [[String: Any]]) throws -> [NSString] {
-        try entries.reduce(into: [NSString](), { partialResult, entry in
-            partialResult.append(try XCTUnwrap(entry["elapsed_since_start_ns"] as? NSString))
-        })
+    func printTimestamps(entries: [[String: Any]]) -> [NSString] {
+        entries.compactMap({ $0["elapsed_since_start_ns"] as? NSString })
     }
 
     func assertMetricEntries(measurements: [String: Any], key: String, expectedEntries: [[String: Any]], transaction: Transaction) throws {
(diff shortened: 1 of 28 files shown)


Tests/SentryProfilerTests/SentryTraceProfilerTests.swift
```diff
@@ -428,9 +428,9 @@ private extension SentryTraceProfilerTests {
         }
     }
 
-    func printTimestamps(entries: [[String: Any]]) -> [NSString] {
-        entries.reduce(into: [NSString](), { partialResult, entry in
-            partialResult.append(entry["elapsed_since_start_ns"] as! NSString)
+    func printTimestamps(entries: [[String: Any]]) throws -> [NSString] {
+        try entries.reduce(into: [NSString](), { partialResult, entry in
+            partialResult.append(try XCTUnwrap(entry["elapsed_since_start_ns"] as? NSString))
```
brustolin, line 433: Does this really need to throw? Or you add it to add it just to comply with XCTUnwrap?

Can we use this?
```Swift
    func printTimestamps(entries: [[String: Any]]) -> [String] {
        return entries.compactMap { $0["elapsed_since_start_ns"] as? String }
    }
```

brustolin (approved):
Thanks. We should keep using this approach, it makes safer tests.

Tests/SentryProfilerTests/SentryTraceProfilerTests.swift
armcknight: oh good point, i like that. then an xctunwrap will work on the result later 👍🏻

PR #4132, opened 2024-06-29
test: convert remaining unchecked accesses to failable versions

armcknight:
Along with #4125, convert some test code that can crash to code that just fails the tests if an assumption is broken, this time targeting the remaining unchecked array indexes.

Helps with #3576.

Two test cases called through to a function that appeared to make assertions, but because they were in an if let block that never was entered, the assertions were not even needed. One of those tests was even incomplete, it should have had an inverted test expectation to ensure a block was never executed.

### DIFF
--- a/Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
+++ b/Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
@@ -196,8 +196,8 @@ private extension SentryContinuousProfilerTests {
 
         let frames = try XCTUnwrap(sampledProfile["frames"] as? [[String: Any]])
         XCTAssertFalse(frames.isEmpty)
-        XCTAssertFalse(try XCTUnwrap(frames[0]["instruction_addr"] as? String).isEmpty)
-        XCTAssertFalse(try XCTUnwrap(frames[0]["function"] as? String).isEmpty)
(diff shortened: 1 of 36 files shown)


Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
```diff
@@ -178,7 +178,7 @@ private extension SentryContinuousProfilerTests {
         let debugMeta = try XCTUnwrap(profile["debug_meta"] as? [String: Any])
         let images = try XCTUnwrap(debugMeta["images"] as? [[String: Any]])
         XCTAssertFalse(images.isEmpty)
-        let firstImage = images[0]
+        let firstImage = try XCTUnwrap(images.first)
```
brustolin, line 181: l: Does this change really matter? Both implementations will throw an error if there is no item in the list.

brustolin (approved):

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
armcknight: If there's no item in the list, this results in an array out of bounds exception, and the test process stops. By throwing an error from a `throws` test case, the test is marked as failed, and the test process can go on to run subsequent test cases.

PR #4133, opened 2024-06-29
ref: reimplement uuid init without string subscript extensions

armcknight:
I noticed this at first by discovering the tests for the string subscripts.

First, I think subscripts are generally bad to use as a custom extension, because discoverability is nonexistent. You can't ⌘ click on them to jump to their def. You can't ⌘ O to quick open to them.

Second, we never even had a test for the one thing they were being used for: creating a UUID from a string. (I added such a test)

Third, they were used in one spot. I rewrote that spot to use a `reduce`. It turns out to be faster, almost twice as fast:

before:

relative standard deviation: 147.699%, values: [0.000473, 0.000064, 0.000047, 0.000044, 0.000042, 0.000041, 0.000041, 0.000040, 0.000040, 0.000040]

after:

relative standard deviation: 179.733%, values: [0.000387, 0.000036, 0.000027, 0.000024, 0.000023, 0.000023, 0.000022, 0.000022, 0.000021, 0.000021]

#skip-changelog


brustolin:
> First, I think subscripts are generally bad to use as a custom extension, because discoverability is nonexistent. You can't ⌘ click on them to jump to their def. You can't ⌘ O to quick open to them.

This is a matter of culture building. After we get used to it, there will be no doubt about how it works.
It's also not possible to "⌘ O" the original string subscript, and that is not a problem.

This particular extension is very useful.

But, I don't mind changing this piece of code. If it works, it works.

Sources/Swift/Extensions/StringExtensions.swift
```diff
@@ -1,43 +1,6 @@
...
-        let start = index(startIndex, offsetBy: range.lowerBound)
-        return self[start..<endIndex]
-    }
-
-    /// Retrieve a substring up to a specified index, excluding that index.
-    subscript (range: PartialRangeUpTo<Int>) -> Substring {
-        let end = index(startIndex, offsetBy: range.upperBound)
-        return self[startIndex..<end]
-    }
-
```
brustolin, line 40: l: I believe the more we use Swift, the more useful this will be.

Tests/SentryTests/Protocol/SentryIdTests.swift
```diff
@@ -22,6 +22,14 @@ class SentryIdTests: XCTestCase {
     func testInit() {
         XCTAssertNotEqual(SentryId(), SentryId())
     }
+    
+    func testInitFromString() {
+        measure {
+            let string = "abcdefabcdefabcdefabcdefabcdefab"
+            let sentryId = SentryId(uuidString: "abcdefabcdefabcdefabcdefabcdefab")
+            XCTAssertEqual(sentryId.sentryIdString, string)
+        }
```
brustolin, line 31: m: `measure` don't work for CI.
```suggestion
        let string = "abcdefabcdefabcdefabcdefabcdefab"
        let sentryId = SentryId(uuidString: "abcdefabcdefabcdefabcdefabcdefab")
        XCTAssertEqual(sentryId.sentryIdString, string)
```

brustolin (approved):

Tests/SentryTests/Protocol/SentryIdTests.swift
armcknight: ah yeah, forgot to remove it after running the experiment

armcknight:
> This is a matter of culture building

Absolutely. I'm trying to build a culture that results in a codebase that is simpler to understand and maintain. Swift is a very flexible language, and thus it's very easy to create an incomprehensible mess of code. I've watched people do it over and over again. This lesson should have been learned with C++ operator overloading, but here we are. "...so preoccupied with whether or not they could that they didn't stop to think if they should..." and I'm saying we shouldn't.

> It's also not possible to "⌘ O" the original string subscript, and that is not a problem

This is a matter of opinion. In my opinion, it is a problem. In my opinion, Swift's substring APIs are horrible. Have fun writing a regex to grep through a codebase for a usage of a subscript, but only the one used for a substring access. It would be much easier to grep for `.substring(from:to:)`.

Sources/Swift/Extensions/StringExtensions.swift
armcknight: If you want a Swift substring wrapper in the future, please create one as a named function like ` func substring(from: Int, to: Int)`, not as a subscript.

PR #4136, opened 2024-06-29
fix(logging): don't unconditionally log to console

armcknight:
As part of https://github.com/getsentry/sentry-cocoa/pull/4103 I made it so that the console logging was optional and is compiled out by default. But, there was already another line of code that would unconditionally log to stdout if any crash logging was enabled. Remove that.

Also, clean up some of the log level strings with extra space, and don't worry about printing __PRETTY_FUNCTION__. It just takes up extra space, which is limited. Filename/lineno are enough.

#skip-changelog


brustolin:
Is there a Dif problem going on?  
I remember seem some of this in #4101.

armcknight:
@brustolin it needed a merge from main, it's g2g now

brustolin (approved):

Sources/Sentry/SentryAsyncSafeLog.c
```diff
@@ -91,7 +91,6 @@ writeToLog(const char *const str)
             pos += bytesWritten;
         }
     }
-    write(STDOUT_FILENO, str, strlen(str));
```
armcknight, line 94: This was the line that did the unconditional log to console in file-only mode and double logged to console in debug mode.

Issue #4086, opened 2024-06-18
Separately named modules for Sentry with and without UIKit


mike-at-home:
### Problem Statement

I'm having some difficulty migrating our build to use both of the binary released sentry-cocoa frameworks (Sentry and Sentry-WithoutUIKitOrAppKit)

The framework without UIKit has the same module name as the one with. I'm not entirely sure if it's possible to get it working with our build system, but I've spent enough time to say that diagnosing why it's not working is difficult enough without the duplicate names.

Would it be possible to get a "no-UIKit" release with a distinct module name or separate out the UIKit in a different way? 
Thanks!

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2209) by [Unito](https://www.unito.io)

brustolin:
We're looking into it, no ETA though.

armcknight:
Hi @mike-at-home , thanks for reporting. We built this special compilation mode specifically for FileProvider extensions which cannot link UIKit. So I guess that if you are delivering such an extension, that your total build process is building regular Sentry (with UIKit) for your app, and then Sentry-without-UIKit for your extension, and because they wind up with the same name, the collision breaks things.

When we set up this new config, I tried creating a test FileProvider extension, but never was able to get it to work. Would you by any chance be able to provide a test project that tests such an integration?

In any case, I think that if we just change the module name setting from this: 
<img width="662" alt="image" src="https://github.com/getsentry/sentry-cocoa/assets/3241469/4138c1fd-5404-43d1-99f9-9b66d393a717">
to this:
<img width="663" alt="image" src="https://github.com/getsentry/sentry-cocoa/assets/3241469/cb28f1de-7a15-45fa-b3d6-09bbfa04a91b">
that should get us most of the way there. We'd just need to validate it doesn't break other things, like package management.

mike-at-home:
Something along those lines should resolve the issue. If the resulting zips can be extracted side by side (ie don't have conflicting Sentry.framework folders) then I think we're good.

Yes, I'm working on Dropbox's codebase which has a file provider. Currently, since we're building both Sentry targets from source, the modules were able to be named separately. But now I'm trying to replace that with the released binaries, so the same flexibility isn't there.

I've never made a file provider project. It's my understanding that it simply cannot link or access UI things. I think you can approximate the same effect by making a command line Mac project. 

[...]

Issue #3219, opened 2023-08-11
Split SDK into feature-based modules


armcknight:
### Description

Customers should be able to selectively integrate separate SDK modules depending on the features they'd like to use. This would optimize the SDK footprint in each app, as instead of still shipping binary for unused features, they wouldn't be present at all.

This has other knock on benefits, like improved build times and enforcing cleaner APIs.

bruno-garcia:
How does this affect installation steps?

armcknight:
We can keep the current package as one that includes everything, and then have subpackages like sentry-core (which would do error/crash reporting), sentry-performance and sentry-UIKit, among other things, that could all be automatically included in the top-level Sentry package. So it shouldn't break anybody.

bruno-garcia:
That makes sense. We did that on Android.
`sentry-android` is a meta-package that includes `sentry-android-core` (actualy has the Android specific core), `sentry-android-ndk` (native stuff) and `sentry-java` the core Java-only which is used also by backend Java/Kotlin/JVM.

So users who don't want the additional 1MB or whatever NDK support, could go straight to `sentry-android-core` but that's rare tbqh. We did need to have two packages for Android and Java (reusable) which is the reason we ended up with this design. All while allowing a simple installation method: Add `sentry-android`.

armcknight:
It's soundung more like this is going to be needed for the long term health of the package. We more recently looked at SDK size, and splitting it up can really reduce the footprint for the average customer who may not want every single feature we offer. It would also help with the ongoing challenges around a UIKitless offering.

philprime:
cc @noahsmartin @philipphofmann @itaybre based on our most recent discussion regarding a project restructuring, should we close this issue as "Not planned"? IIRC you brought up that you advocated that we should not do this.

philippsuess:
Yes, let's close this and open up new issues once we know exactly what we plan on doing. Please reopen, if you disagree.

PR #4140, opened 2024-07-02
fix: define separate module names for UIKitless configurations

armcknight:
Define a different module name for the configurations that build without UIKit enabled. Ideally we split out a separate submodule for UIKit stuff a la https://github.com/getsentry/sentry-cocoa/issues/3219, but this is a workaround in the meantime.

Had to fix an issue with the test targets' configuration. They included Sentry.xcconfig and so inherited the module names. But, they never should have inherited from that config. It looks like they overrode everything anyways, except for the workaround using HEADER_SEARCH_PATHS.

fixes #https://github.com/getsentry/sentry-cocoa/issues/4086


aaronsky:
Try this @armcknight 

[asky-1.patch](https://github.com/user-attachments/files/16067072/asky-1.patch)

Sources/Swift/Integrations/SessionReplay/SentryReplayVideoMaker.swift
```diff
@@ -1,4 +1,4 @@
-#if canImport(UIKit)
```
armcknight, line 1: This reintroduced implicit UIKit linkage in https://github.com/getsentry/sentry-cocoa/pull/3877

brustolin: Don't we have tests to ensure we don't reintroduced implicit UIKit linkage?

armcknight: It wasn't checking for linkage with libswiftUIKit, which I added in https://github.com/getsentry/sentry-cocoa/pull/4140/commits/0cb91fbd63ca0483941f6c7febf10eefe1f7cd1d. 

This wasn't needed when I originally shipped the UIKitless stuff because you all weren't yet writing core SDK code in Swift. At the time you decided to do so, it should've been added.

Sources/Sentry/Public/Sentry.h
```diff
@@ -1,46 +1,49 @@
-#import <Foundation/Foundation.h>
+#if !__has_include(<SentryWithoutUIKit/Sentry.h>)
```
aaronsky, line 1: why this and not `__has_include(<Sentry/Sentry.h>)`?

Sources/Sentry/Public/SentryWithoutUIKit.h
brustolin, line 1: l: I would merge this with `Sentry.h`, so the user can `#import <SentryWithoutUIKit/Sentry.h>` instead of `#import <SentryWithoutUIKit/SentryWithoutUIKit.h>`

At the same time, most user use Swift, and this does not matter.

aaronsky: From my own investigation, this cannot be merged due to the mixed-language module's need for an umbrella header matching the name of the module.

brustolin (approved):
Thanks for this @armcknight !!

All looks good.

Sources/Sentry/Public/Sentry.h
armcknight: No specific reason. Yours is nicer because it avoids the double negative 😄 I'll change it and test, and if all looks good, we'll get this merged.

armcknight: Everything seems to work 👍🏻

PR #4141, opened 2024-07-02
fix(replay): SR Breadcrumbs include level name as a string

krystofwoldrich:
## :scroll: Description

This PR fixes SR Breadcrumb Level serialization. The level is now included as a string, which is used in the UI. Android and Browser work the same.

Before the levels were send as numbers.

unit test, sample app

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/include/SentryLevelHelper.h
+++ b/Sources/Sentry/include/SentryLevelHelper.h
@@ -9,7 +9,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface SentryLevelHelper : NSObject
 
-+ (NSString *_Nonnull)breadcrumbLevel:(SentryBreadcrumb *)breadcrumb;
++ (NSUInteger)breadcrumbLevel:(SentryBreadcrumb *)breadcrumb;
 
 @end
 
--- a/Sources/Sentry/include/SentryLevelHelper.h
+++ b/Sources/Sentry/include/SentryLevelHelper.h
@@ -9,7 +9,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface SentryLevelHelper : NSObject
 
-+ (NSUInteger)breadcrumbLevel:(SentryBreadcrumb *)breadcrumb;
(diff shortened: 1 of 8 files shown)


Sources/Sentry/include/SentryLevelHelper.h
```diff
@@ -9,7 +9,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface SentryLevelHelper : NSObject
 
-+ (NSUInteger)breadcrumbLevel:(SentryBreadcrumb *)breadcrumb;
++ (NSString *_Nonnull)breadcrumbLevel:(SentryBreadcrumb *)breadcrumb;
```
bruno-garcia, line 12: This is a breaking change as it's a public API.

I think it's best we add a new method. And use that internally instead. And we can annotate the old one as obsolete

bruno-garcia (changes requested):
We need to keep the current public API unchanged

Sources/Sentry/include/SentryLevelHelper.h
krystofwoldrich: Make sense, I'll update it.

armcknight: I believe you're correct, it's not public. Either way, I would prefer if instead of modifying the signature, nameForSentryLevel were used inside SentryRRWebBreadcrumbEvent.init.

krystofwoldrich: Thanks for the suggestion, I've reverted the `breadcrumbLevel` changes and added the conversion to `SentryRRWebBreadcrumbEvent`.

[`c7e4835` (#4141)](https://github.com/getsentry/sentry-cocoa/pull/4141/commits/c7e483514e1d0e467e6c810de39f4a6bf75efe1d)

bruno-garcia (approved):

bruno-garcia:
Hm looks like requires code owner approval. Andrew is off the due to 4th July so either I force merge or we wait until monday

PR #4170, opened 2024-07-16
chore: Double check lock extension for NSLock

brustolin:
Writing double check lock in swift is weird, and it's not possible to use the `.synchronized` extension, so I added a little NSLock extension that can be used with a `guard` statement. 

_#skip-changelog_

### DIFF
--- a/Sources/Swift/Extensions/NSLock.swift
+++ b/Sources/Swift/Extensions/NSLock.swift
@@ -20,7 +20,7 @@ extension NSLock {
     /// Only runs the closure if the flag is false, then updates the flag to true.
     /// - Parameter toRun: The closure to run.
     ///
-    /// - Returns: The value the closure generated or `nil` if the flag was enabled already.
+    /// - Returns: `true` if the closure was invoked, otherwise `false`.
     func setFlag(_ flag: inout Bool, toRun closure: () throws -> Void) rethrows -> Bool {
         if flag { return false }
         self.lock()
@@ -38,7 +38,7 @@ extension NSLock {
     /// Only runs the closure if the flag is true, then updates the flag to false.
     /// - Parameter toRun: The closure to run.
     ///
-    /// - Returns: The value the closure generated or `nil` if the flag was enabled already.
+    /// - Returns: `true` if the closure was invoked, otherwise `false`.
     func unsetFlag(_ flag: inout Bool, toRun closure: () throws -> Void) rethrows -> Bool {
         if !flag { return false }
         self.lock()
--- a/Sources/Swift/Extensions/NSLock.swift
+++ b/Sources/Swift/Extensions/NSLock.swift
@@ -12,4 +12,23 @@ extension NSLock {
         self.lock()
         return try closure()
     }
+    
+    /// Executes the closure if the flag is false while acquiring the lock.
+    /// Updates the flag to true. This can be used as double-check lock.
+    ///
+    /// - Parameter flag: A flag that will be double check to make sure closure should be invoked.
+    /// Only runs the closure if the flag is false, then updates the flag to indicate the closure was called.
+    /// - Parameter toRun: The closure to run.
+    ///
+    /// - Returns: The value the closure generated or `nil` if the flag was enabled already.
+    func checkFlag(flag: inout Bool, toRun closure: () throws -> Void) rethrows -> Bool {
+        if flag { return false }
+        self.lock()
+        defer { self.unlock() }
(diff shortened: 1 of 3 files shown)


armcknight (changes requested):
I've noticed this pattern often of us checking some value before taking a lock, then checking that same value after taking the lock. I agree it's annoying.

In the case you used to demonstrate in this PR, I think rather than going forward creating _more abstractions_ to support this, we should _remove state_ we're maintaining that requires the abstraction in the first place, when there's already other state we could check to answer the same question.

Here's a PR that removes `isRunning` as a writable property: https://github.com/getsentry/sentry-cocoa/pull/4192

If you have other places you'd want to use this, please list them, but for now I recommend we don't move forward with this. It involves a complex topic and makes the callsites harder to reason about. IMO it's a big liability.

brustolin:
Double-lock check is a pattern that was suggested to me since I joined Sentry because it has better performance. We have many examples of this across the SDK, though not so many with Swift because we're just starting to write code in Swift. Usually, I prefer to prepare in advance and have this function available instead of having to create it again when we need it, or, even worse, not using it and ending up with hard-to-read code.

armcknight:
You didn't address the core concern I present. You're focusing on double-checked locking on a state variable that I'm saying doesn't even need to exist.

You're also prematurely introducing this abstraction "in case we need it". YAGNI

brustolin:
> YAGNI

We have this approach all around Objc, Why do you think we won't use it with Swift?

philipphofmann:
We only use the NSLock in `SentrySessionReplay.start`, `SentrySessionReplay.stop` and `SentrySessionReplay.resume`. If I'm not mistaken only the `SentrySessionReplayIntegration` calls these methods, so I think we could even remove the NSLock completely from `SentrySessionReplay`. A double-checked lock is an overkill for the `SentrySessionReplay`. It only makes sense to use it when locks are acquired frequently, such as in the SentryHttpTransport or SentryTracer. Therefore, I think we must not move forward with this PR.

armcknight:
> We have this approach all around Objc

Those cases could be wrong as well. If there's a variable that has to be locked in order to be read/written, then reading it before acquiring the lock means you aren't always reading a valid state. You're just adding data races. I would say those other places are probably wrong and should be removed as well.

> we could even remove the NSLock completely from SentrySessionReplay

I went to remove the rest of the usages of that lock, but there's one place where it's used that looks like it would benefit from a dispatch queue instead, around taking screenshots.

Thats the kind of thing that would probably help improve code where we're currently using these "double checked locks". There's even the set of `dispatch_barrier_async_...` functions that act as a synchronization point on a concurrent private queue.

> It only makes sense to use it when locks are acquired frequently, such as in the SentryHttpTransport or SentryTracer

Do we have some data on the performance tradeoff? But again, see above discussing about the validity of checking the value of a volatile variable outside of a critical sector in the first place.

Is this a case of premature optimization or did these get put in place due to an observed issue?

philipphofmann:
> I went to remove the rest of the usages of that lock in https://github.com/getsentry/sentry-cocoa/pull/4192, but there's one place where it's used that looks like it would benefit from a dispatch queue instead, around taking screenshots.

Okay, then we need some synchronization. Thanks for double-checking. A serial dispatch queue could do the trick, yes.

> Do we have some data on the performance tradeoff? But again, see above discussing about the validity of checking the value of a volatile variable outside of a critical sector in the first place.
> Is this a case of premature optimization or did these get put in place due to an observed issue?


I don't have actual data, but I noticed a difference when writing performance tests calling the lock versus a double-checked lock from multiple threads in a tight loop in some scenarios in the SentryHttpTransport, but we removed the code already. I just checked the SentryHttpTransport and figured out we can safely remove the double-checked lock for flushing (https://github.com/getsentry/sentry-cocoa/pull/4198) cause it's very unlikely you call that in a tight loop. I can't recall why I added it there.

The one in SentryTracer for the app start measurement makes sense to me, though: https://github.com/getsentry/sentry-cocoa/blob/061c982c8616fa6ca7f0762998f39a8f4d8d2fc2/Sources/Sentry/SentryTracer.m#L768-L776

The app start measurement is put on the first UI load transaction. Every following transaction would need to acquire a lock to check if it should add the app start measurement. With the double-checked lock we prevent that. 

We also use this pattern in the `SentryDependencyContainer` where it makes sense, in my opinion. Otherwise, every time you resolve a dependency, you acquire a lock. Do you need numbers to prove that no lock is better in that case?

PR #4191, opened 2024-07-23
ci: dependency review action

mdtro:
Add dependency review action. Any new dependencies added with known high or critical vulnerabilities will fail CI.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,19 @@
+name: 'Dependency Review'
+on:
+  pull_request:
+    branches: ['master']
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Dependency Review
+        uses: actions/dependency-review-action@5a2ce3f5b92ee19cbb1541a4984c76d921601d7c # v4.3.4
+        with:
+          # Possible values: "critical", "high", "moderate", "low"
+          fail-on-severity: high


armcknight:
Thanks @mdtro , this sounds like a good idea. Regarding the pinned versions, will a github bot automatically submit PRs for us to upgrade this tool? Or will we have to monitor that tool to see if there are updates? I don't want to set it and forget it and let it slowly rot.

mdtro:
> Thanks @mdtro , this sounds like a good idea. Regarding the pinned versions, will a github bot automatically submit PRs for us to upgrade this tool? Or will we have to monitor that tool to see if there are updates? I don't want to set it and forget it and let it slowly rot.

Dependabot's current configuration on this repo won't update these since we have the SHAs pinned. We would need to enable version updates (instead of just security updates). I can do that in this PR if you'd like. :)

mdtro:
> > Thanks @mdtro , this sounds like a good idea. Regarding the pinned versions, will a github bot automatically submit PRs for us to upgrade this tool? Or will we have to monitor that tool to see if there are updates? I don't want to set it and forget it and let it slowly rot.
> 
> Dependabot's current configuration on this repo won't update these since we have the SHAs pinned. We would need to enable version updates (instead of just security updates). I can do that in this PR if you'd like. :)

@armcknight Scratch that, it looks like Dependabot is already configured here to do version updates on GitHub Actions. So yes, it should be auto-updated. 🎉

armcknight (approved):
Thanks @mdtro ❤️ !

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Dependency Review'
+on:
+  pull_request:
+    branches: ['master']
```
kahest, line 4: @mdtro should this be `main`?

kahest:
@mdtro should we also make sure this is part of the [required status checks](https://github.com/actions/dependency-review-action#using-dependency-review-action-to-block-a-pull-request-from-being-merged) for merging to main?

PR #4192, opened 2024-07-23
ref: remove isRunning state var and inspect displayLink directly

armcknight:
#skip-changelog

We unnecessarily maintain a state variable, when we could look at the underlying state to answer the question it is meant to answer.


Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -61,17 +64,7 @@ class SentrySessionReplay: NSObject {
...
 
     func start(rootView: UIView, fullSession: Bool) {
-        guard !isRunning else { return }
-        
-        lock.lock()
-        guard !isRunning else {
-            lock.unlock()
-            return
-        }
         displayLink.link(withTarget: self, selector: #selector(newFrame(_:)))
```
brustolin, line 67: h: Won't this be a problem? 
You're not invalidating the previous displayLink. 
Which it is really not necessary, we can keep using the same link.
And between `start` and `stop` I believe we should still lock the manipulation of displayLink

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -61,17 +64,7 @@ class SentrySessionReplay: NSObject {
     }
 
     func start(rootView: UIView, fullSession: Bool) {
-        guard !isRunning else { return }
```
brustolin, line 64: m: I believe we should keep this, it will avoid problems described in the next comment.

brustolin (commented):
A few comments.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
armcknight: I can see how we'd want to keep this to essentially debounce multiple calls to `start`.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
armcknight: I'm not sure, since the code never invalidated before my changes. Also, I'm not 100% on this, but I believe a CADisplayLink instance should not be reused after calling `invalidate`. If we wish to "stop"/"start" the same instance, we should be using its `paused` property instead. I'm not sure it's necessary to lock while working directly with the display link property of the SentrySessionReplay instance, because `invalidate`/`paused` are documented as thread safe.

brustolin: It did not invalidate before because if `start` was called when it was already running, the code did nothing and returned early. Now, there is no check anymore, and the code is adding a second callback to displayLink. I'm not sure if displayLink has protection against this.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
armcknight: I added this check back in, so I think both of your comments are addressed by that? LMK if there's anything else needed.

brustolin (approved):

Issue #4149, opened 2024-07-09
add `shellcheck` precommit hook


armcknight:
### Description

to lint shell scripts

brustolin:
Our scripts are not part of the product but tools to automate things. 
Adding lint to them is just an extra unnecessary headache

philipphofmann:
If ShellCheck improves the quality of our shell scripts our developer workflow should benefit and therefore make us more productive. I think we can give this a shot both as a pre commit hook and CI and see how it goes. We can always ditch it if it's a headache.

PR #4200, opened 2024-07-26
chore: add shellcheck precommit hook

armcknight:
for #4149

#skip-changelog


philipphofmann (commented):
It would be great to also run shellcheck in CI to enforce consistency not just by precommit in case somebody doesn't set it up correctly. Apart from that LGTM. Thanks for doing this.

philipphofmann (approved):
LGTM, thanks @armcknight.

PR #4201, opened 2024-07-29
ref: Log for Swift

### LINKED ISSUE #4020: Logging improvements

brustolin:
## :scroll: Description

Standard log mechanism for swift.

Unit tests

## :pencil: Checklist

_#skip-changelog_


Sources/Swift/Tools/SentryLog.swift
```diff
@@ -0,0 +1,76 @@
...
+     * @c NO if not.
+     */
+    @objc
+    static func willLog(atLevel level: SentryLevel) -> Bool {
+        return isDebug && level != .none && level.rawValue >= diagnosticLevel.rawValue
+    }
+ 
+    #if TEST || TESTCI
+    
+    @objc
```
brustolin, line 38:
```suggestion
```

Sources/Swift/Tools/SentryLog.swift
```diff
@@ -0,0 +1,75 @@
...
+    static func getOutput() -> SentryLogOutput {
+        return logOutput
+    }
+    
+    #endif
+}
+
+extension SentryLog {
+    private static func log(level: SentryLevel, message: String, file: String, line: Int) {
+        let path = file as NSString
```
philipphofmann, line 51: `h`: We should also guard this with an early exit to avoid the extra work of allocating a string. I added this with https://github.com/getsentry/sentry-cocoa/pull/4028 and allocating extra strings in log messages caused problems previously, see https://github.com/getsentry/sentry-cocoa/pull/4028.
```suggestion
        guard willLog(atLevel: level) else { return }
        let path = file as NSString
```

Sources/Swift/Tools/SentryLogOutput.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+@objcMembers
+class SentryLogOutput: NSObject {
```
philipphofmann, line 4: `m`: I believe we only use this in Swift, so we can make it Swift only.
```suggestion
class SentryLogOutput {
```

Tests/SentryTests/Integrations/SentryBaseIntegrationTests.swift
```diff
@@ -15,10 +15,10 @@ class SentryBaseIntegrationTests: XCTestCase {
 
     override func setUp() {
         super.setUp()
-        oldDebug = SentryLog.isDebug()
-        oldLevel = SentryLog.diagnosticLevel()
-        oldOutput = SentryLog.logOutput()
+        oldDebug = SentryLog.isDebug
+        oldLevel = SentryLog.diagnosticLevel
```
philipphofmann, line 19: Thanks for changing these to properties 👍

Tests/SentryTests/SentryLevelTests.swift
```diff
@@ -0,0 +1,22 @@
...
+
+class SentryLevelTests: XCTestCase {
+    
+    func testNames() {
+        XCTAssertEqual(SentryLevel.none, SentryLevel.fromName("none"))
+        XCTAssertEqual(SentryLevel.debug, SentryLevel.fromName("debug"))
+        XCTAssertEqual(SentryLevel.error, SentryLevel.fromName("error"))
+        XCTAssertEqual(SentryLevel.info, SentryLevel.fromName("info"))
+        XCTAssertEqual(SentryLevel.fatal, SentryLevel.fromName("fatal"))
+        XCTAssertEqual(SentryLevel.warning, SentryLevel.fromName("warning"))
```
philipphofmann, line 13: `l`: I would add an assert for a garbage name to validate it maps to error.

philipphofmann (approved):
`l`: I would prefix the PR title with `ref` cause it's a refactoring. I found a few minor issues to address. Thanks for doing this 🥇 

LGTM.

PR #4206, opened 2024-07-30
ref: Convert SentryFileContents to Swift

philipphofmann:
I converted this to Swift cause I had a couple of minutes in between meetings. It's a small class and was easy to do.

#skip-changelog


brustolin (approved):
I'm all up for converting things to Swift. 
I'm just wondering, do we really need SentryFileContents? 
I didn't even know we had this class.

Removing code is even better than converting to Swift.

Other than that, all good.

philipphofmann:
Yes, the SentryHttpTransport uses it. I would keep it for now.

PR #4214, opened 2024-07-31
fix(continuous profiling): full chunk stoppage

armcknight:
The profiling team discussed some issues with sending partial chunks at the end of a profiling session. This change lets the last chunk run to the full length and transmits that before shutting down the profiler.

#skip-changelog

### DIFF
--- a/SentryTestUtils/ClearTestState.swift
+++ b/SentryTestUtils/ClearTestState.swift
@@ -43,9 +43,22 @@ class TestCleanup: NSObject {
         _sentry_threadUnsafe_traceProfileTimeoutTimer = nil
         SentryTraceProfiler.getCurrentProfiler()?.stop(for: SentryProfilerTruncationReason.normal)
         SentryTraceProfiler.resetConcurrencyTracking()
-        SentryContinuousProfiler.stop()
         removeAppLaunchProfilingConfigFile()
         sentry_stopAndDiscardLaunchProfileTracer()
+        
+        if SentryContinuousProfiler.isCurrentlyProfiling() {
+//            SentryContinuousProfiler.stop()
+//            // need to advance the time and fire the timer for the profiler to actually stop
+//            if let testDateProvider = SentryDependencyContainer.sharedInstance().dateProvider as? TestCurrentDateProvider {
+//                testDateProvider.advance(by: kSentryProfilerChunkExpirationInterval)
+//            }
+//            if let testTimerFactory = SentryDependencyContainer.sharedInstance().timerFactory as? TestSentryNSTimerFactory {
+//                testTimerFactory.fire()
+//            }
+            
+            // or, call an exposed internal method that just stops it
+            SentryContinuousProfiler.stopTimerAndCleanup()
+        }
 #endif // os(iOS) || os(macOS) || targetEnvironment(macCatalyst)
 
         #if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
--- a/SentryTestUtils/ClearTestState.swift
+++ b/SentryTestUtils/ClearTestState.swift
@@ -47,16 +47,6 @@ class TestCleanup: NSObject {
         sentry_stopAndDiscardLaunchProfileTracer()
         
         if SentryContinuousProfiler.isCurrentlyProfiling() {
-//            SentryContinuousProfiler.stop()
-//            // need to advance the time and fire the timer for the profiler to actually stop
-//            if let testDateProvider = SentryDependencyContainer.sharedInstance().dateProvider as? TestCurrentDateProvider {
-//                testDateProvider.advance(by: kSentryProfilerChunkExpirationInterval)
-//            }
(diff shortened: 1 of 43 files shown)


SentryTestUtils/ClearTestState.swift
```diff
@@ -43,9 +43,22 @@ class TestCleanup: NSObject {
...
+//            SentryContinuousProfiler.stop()
+//            // need to advance the time and fire the timer for the profiler to actually stop
+//            if let testDateProvider = SentryDependencyContainer.sharedInstance().dateProvider as? TestCurrentDateProvider {
+//                testDateProvider.advance(by: kSentryProfilerChunkExpirationInterval)
+//            }
+//            if let testTimerFactory = SentryDependencyContainer.sharedInstance().timerFactory as? TestSentryNSTimerFactory {
+//                testTimerFactory.fire()
+//            }
+            
+            // or, call an exposed internal method that just stops it
```
brustolin, line 59: m: Was this left in here by mistake?

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
```diff
@@ -159,14 +164,32 @@ + (void)scheduleTimer
...
+        }
     }
 
-    _sentry_threadUnsafe_transmitChunkEnvelope();
+    [self stopTimerAndCleanup];
+}
+
++ (void)stopTimerAndCleanup
+{
+    std::lock_guard<std::mutex> l(_threadUnsafe_gContinuousProfilerLock);
```
philipphofmann, line 187: `h`: Is it okay to leave the lock here? Shouldn't we instead move the code of `stopTimerAndCleanup` to `imerExpired` and keep it in the lock?

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
```diff
@@ -159,14 +164,32 @@ + (void)scheduleTimer
...
+
+        _sentry_threadUnsafe_transmitChunkEnvelope();
+
+        if (!_stopCalled) {
+            return;
+        }
     }
 
-    _sentry_threadUnsafe_transmitChunkEnvelope();
+    [self stopTimerAndCleanup];
```
philipphofmann, line 182: `m`: Is it intended to first transmit the chunk envelope and then stop the profiler? Doesn't it make more sense to stop the profiler first and then transmit the chunk envelope?

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
```diff
@@ -110,6 +110,32 @@ final class SentryContinuousProfilerTests: XCTestCase {
         manualSpan.finish()
         automaticSpan.finish()
     }
+
+    // test that when stop is called, the profiler runs to the end of the last
+    // chunk and transmits that before stopping
+    func testStoppingProfilerTransmitsLastFullChunk() throws {
```
philipphofmann, line 116: Thanks for the comment. We need to properly document this behavior for the `SentrySDK.stopProfiler` method once we make it public cause this could confuse users quite a bit.

SentryTestUtils/ClearTestState.swift
armcknight: Yep, oops...

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
armcknight: I factored this method out to make stopping the continuous profiler cleaner in clearTestState, and didn't want that to cause any threading issues, so factored it out in a way that this will acquire the lock separately from the one acquired in the normal codepath that calls stopTimerAndCleanup in SentryContinuousProfiler.stop

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
armcknight: Ah yeah I'll add that to this PR.

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
armcknight: I think there's such a short amount of time between the two it doesn't much matter. And even less so if/when we shuffle the serialization/envelope capture off to a separate thread. The ordering is functionally the same as it was before this change: 
image

philipphofmann (approved):
LGTM

armcknight:
@philipphofmann had to rebase this onto #4234 because a test started failing related to what happens when a bg notification was received, since the change in behavior in this PR combined with that to create an unexpected result. Now, when a bg notification is received, the in-flight continuous profile chunk will be discarded and the profiler stopped, since there's no guarantee that the chunk would be able to finish and be transmitted in time before the app goes to the bg.

The diff should look mostly the same, with the addition of the changes starting with https://github.com/getsentry/sentry-cocoa/pull/4214/commits/db969891878ba18d0814cb7c8086a6074d9763ed

philipphofmann (approved):
Again, LGTM, when CI is green.

PR #4219, opened 2024-08-01
impr: Reduce memory usage of storing envelopes

philipphofmann:
## :scroll: Description

Reduce the memory footprint when storing envelopes to disk by writing
the envelopes in chunks by using an NSFileHandle. This addresses one
part of https://github.com/getsentry/sentry-cocoa/issues/3630.

This addresses one
part of https://github.com/getsentry/sentry-cocoa/issues/3630.

I tested the functionality with unit tests and for the improved memory footprint by storing an envelope with two attachments, each around 10MB, and checked the physical memory peak with XCTest.
```
self.measure(metrics: [XCTMemoryMetric()]) {
    sut.store(envelope)
}
```
The improved version shows a peak of `42794.662 kB` and the existing version `67832.691 kB`. The improved version has around 20 MB smaller physical memory peak, which is roughly the size of the two added attachments. This is because the improved version doesn't duplicate the NSData of the attachment before writing it to disk, but instead directly writes it to disk.

You can check the full test case here, which I didn't commit because CI can't validate the memory peaks as it requires a baseline profile which doesn't work in CI.

<details><summary>Full Test Case</summary>
<p>

```
func testStoreLargeEnvelope() throws {
    let event = Event()
    let header = SentryEnvelopeHeader(id: event.eventId)
    let eventItem = SentryEnvelopeItem(event: event)
    
    let maxAttachmentSize : UInt = 20 * 1024 * 1024
    
    let path = try XCTUnwrap(Bundle(for: type(of: self)).path(forResource: "Resources/image", ofType: "png"))
    
    let attachment1 = Attachment(path: path)
    let attachment2 = Attachment(path: path)
    
    let attachmentItem1 = try XCTUnwrap(SentryEnvelopeItem(attachment: attachment1, maxAttachmentSize: maxAttachmentSize))
    let attachmentItem2 = try XCTUnwrap(SentryEnvelopeItem(attachment: attachment2, maxAttachmentSize: maxAttachmentSize))
    
    let envelope = SentryEnvelope(header: header, items: [eventItem, attachmentItem1, attachmentItem2])
    
    self.measure(metrics: [XCTMemoryMetric()]) {
        sut.store(envelope)
    }
}
```
</p>
</details> 

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -310,7 +310,7 @@ SentryFileManager ()
     }
 }
 
-- (NSString *)storeEnvelope:(SentryEnvelope *)envelope
+- (void)storeEnvelope:(SentryEnvelope *)envelope
 {
     @synchronized(self) {
         NSString *path =
@@ -324,7 +324,7 @@ SentryFileManager ()
 
         if (!fileHandle) {
             SENTRY_LOG_ERROR(@"Couldn't get NSFileHandle for path: %@", path);
-            return path;
+            return;
         }
 
         @try {
@@ -344,7 +344,6 @@ SentryFileManager ()
         }
 
         [self handleEnvelopesLimit];
-        return path;
     }
 }
 
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -23,6 +23,7 @@ NSString *const EnvelopesPathComponent = @"envelopes";
 BOOL
 createDirectoryIfNotExists(NSString *path, NSError **error)
 {
+
     if (![[NSFileManager defaultManager] createDirectoryAtPath:path
                                    withIntermediateDirectories:YES
                                                     attributes:nil
@@ -312,14 +313,10 @@ SentryFileManager ()
 
 - (NSString *)storeEnvelope:(SentryEnvelope *)envelope
 {
-    NSError *error;
     @synchronized(self) {
         NSString *path =
             [self.envelopesPath stringByAppendingPathComponent:[self uniqueAscendingJsonName]];
 
-        if (!createDirectoryIfNotExists(self.envelopesPath, &error)) {
-            return nil;
-        }
         SENTRY_LOG_DEBUG(@"Writing envelope to path: %@", path);
 
         [[NSFileManager defaultManager] createFileAtPath:path contents:nil attributes:nil];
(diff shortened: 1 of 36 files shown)


brustolin (commented):
Thanks for this.
This will help a little bit.

For a next step we should serialize the items, headers and envelope direct to file instead of data and then write the data to file.

philipphofmann:
> For a next step we should serialize the items, headers and envelope direct to file instead of data and then write the data to file.

Yes, exactly.

Sources/Sentry/SentryFileManager.m
```diff
@@ -312,15 +312,35 @@ - (void)removeFileAtPath:(NSString *)path
...
 
-        if (![self writeData:envelopeData toPath:path]) {
-            SENTRY_LOG_WARN(@"Failed to store envelope.");
+        [[NSFileManager defaultManager] createFileAtPath:path contents:nil attributes:nil];
+
+        NSFileHandle *fileHandle = [NSFileHandle fileHandleForWritingAtPath:path];
+
+        if (!fileHandle) {
+            SENTRY_LOG_ERROR(@"Couldn't get NSFileHandle for path: %@", path);
+            return path;
```
armcknight, line 327: Why return `path` here instead of something like `nil` with an inout error parameter filled? I also looked and nothing in the SDK actually uses this return value, only one test case. Would make more sense for this method to accept a path at white to store the envelope, then that test case could retrieve it that way; and this method could instead return a BOOL for success/fail if it would actually help anywhere up the callstack, or just void.

I'm also guessing here but it looks like we only need the `@synchronized` on line 315 because of that call to `uniqueAscendingJsonName`? If so, moving that out reduces the scope of this critical sector and should increase throughput.

armcknight (approved):
Great change! One question.

Sources/Sentry/SentryFileManager.m
philipphofmann: > I'm also guessing here but it looks like we only need the @synchronized on line 315 because of that call to uniqueAscendingJsonName? If so, moving that out reduces the scope of this critical sector and should increase throughput.

Sadly, that's not possible, because we `handleEnvelopesLimit` at the end of the method needs to be also in the synchronized block. If the max number of elements is reached, the method iterates over all envelopes and migrates the sessions init. There is a bunch of IO happening there.

philipphofmann: I made method void, cause as you pointed out we only use the return value for testing and we can get the latest stored envelope in tests via `sut.getAllEnvelopes().last?.path`. Thanks for pointing that out.

PR #4225, opened 2024-08-01
fix: Long-lasting TTID/TTFD spans

philipphofmann:
## :scroll: Description

Presenting a UIViewController when going to the background could lead to long-lasting TTID/TTFD spans. This is fixed now by not creating TTID/TTFD spans when the FramesTracker isn't running.

Docs PR: https://github.com/getsentry/sentry-docs/pull/10960.

A customer reported to get long lasting TTID/TTFD spans for UIViewControllers presented when the `didEnterBackgroundNotification` is posted. They do this due to privacy reasons.

Added a small sample to present an UIViewController when the `didEnterBackgroundNotification` is posted and could sometimes reproduce the problem. This PR fixes that problem. I didn't add the sample as a UITest, because it isn't reliable and this PR addresses an edge case. Furthermore, I added unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
+++ b/Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
@@ -147,11 +147,14 @@ SentryUIViewControllerPerformanceTracker ()
                                            waitForFullDisplay:self.enableWaitForFullDisplay
                                          dispatchQueueWrapper:_dispatchQueueWrapper];
 
-    objc_setAssociatedObject(controller, &SENTRY_UI_PERFORMANCE_TRACKER_TTD_TRACKER, ttdTracker,
-        OBJC_ASSOCIATION_ASSIGN);
-    [ttdTracker startForTracer:(SentryTracer *)vcSpan];
+    if ([ttdTracker startForTracer:(SentryTracer *)vcSpan]) {
+        objc_setAssociatedObject(controller, &SENTRY_UI_PERFORMANCE_TRACKER_TTD_TRACKER, ttdTracker,
+            OBJC_ASSOCIATION_ASSIGN);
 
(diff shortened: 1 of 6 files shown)


armcknight (approved):

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -147,11 +147,14 @@ - (void)createTimeToDisplay:(UIViewController *)controller
                                            waitForFullDisplay:self.enableWaitForFullDisplay
                                          dispatchQueueWrapper:_dispatchQueueWrapper];
 
-    objc_setAssociatedObject(controller, &SENTRY_UI_PERFORMANCE_TRACKER_TTD_TRACKER, ttdTracker,
-        OBJC_ASSOCIATION_ASSIGN);
-    [ttdTracker startForTracer:(SentryTracer *)vcSpan];
+    if ([ttdTracker startForTracer:(SentryTracer *)vcSpan]) {
```
brustolin, line 150: l: Wondering if we should check if the FrameTracker is active  is this function  and don't even create a new displayTracker instance.

brustolin (approved):
Pretty straight forward. I believe this will fix the issue.

Just one suggestion but all looks good.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
philipphofmann: I also considered this approach, but the downside is that we would have the logic of the frames tracker now in here, instead of encapsulated in the `SentryTimeToDisplayTracker`. Furthermore, the `SentryUIViewControllerPerformanceTracker` doesn't use the frames tracker and I would like to keep it that way.

Issue #4227, opened 2024-08-02
Crashes miss handled property


philipphofmann:
### Platform

iOS, tvOS, iPadOS, macOS, watchOS, visionOS

### Environment

Develop

### Installed

Manually

### Version

8.33.0

### Xcode Version

15.4

### Did it work on previous versions?

Yes, but unsure which one.

### Steps to Reproduce

Trigger this code to trigger a NSInvalidArgumentException in the Objective-C sample app or click the crash button in the iOS-Swift sample app.
```
NSObject *object = [[NSObject alloc] init];
[object performSelector:@selector(nonExistentMethod)];
```

### Expected Result

The event in Sentry should have the handled property set to no

![Image](https://github.com/user-attachments/assets/8d4adcce-115e-45d5-8604-054542a46e1b)

### Actual Result

The crash event doesn't have the handled property

![Image](https://github.com/user-attachments/assets/89481ead-fce2-414f-9d6a-b666a2416112)

### Are you willing to submit a PR?

yes

PR #4230, opened 2024-08-05
fix: Missing mach info for crash reports

philipphofmann:
## :scroll: Description

The SentryInternalCDefines didn't define the SENTRY_HOST_APPLE, which led to missing mach info for crash reports, as the SentryCrashReport relies on the SENTRY_HOST_APPLE to write that info. For reference, the SentryCrashCRASH_HOST_APPLE was changed to SENTRY_HOST_APPLE in GH-4101. The problem is fixed now by defining SENTRY_HOST_APPLE in SentryInternalCDefines.

Fixes GH-4227

Unit tests and checking crash reports of a simulator.

| Before | After |
|--------|--------|
| CleanShot 2024-08-05 at 09 42 52@2x | CleanShot 2024-08-05 at 09 43 23@2x |
| CleanShot 2024-08-05 at 09 43 44@2x | CleanShot 2024-08-05 at 09 44 08@2x | 

## :pencil: Checklist

### DIFF
--- a/Tests/SentryTests/Integrations/SentryCrash/SentryCrashReportTests.swift
+++ b/Tests/SentryTests/Integrations/SentryCrash/SentryCrashReportTests.swift
@@ -189,8 +189,8 @@ class SentryCrashReportTests: XCTestCase {
     // We parse JSON so it's fine to disable identifier_name
     // swiftlint:disable identifier_name
     struct CrashReport: Decodable {
-        let user: CrashReportUserInfo
-        let sentry_sdk_scope: CrashReportUserInfo
+        let user: CrashReportUserInfo?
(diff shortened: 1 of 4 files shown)


brustolin (approved):

Tests/SentryTests/Integrations/SentryCrash/SentryCrashReportTests.swift
```diff
@@ -187,14 +230,6 @@ class SentryCrashReportTests: XCTestCase {
         let breadcrumbs: [CrashReportCrumb]?
     }
 
-    struct CrashReportUser: Decodable, Equatable {
```
philipphofmann, line 190: I removed that because it caused problems in tests, but we don't use that code.

PR #4234, opened 2024-08-05
chore(SentryNSNotificationCenterWrapper): add block observation

armcknight:
I need the block observation API for a change in #4214

#skip-changelog


philipphofmann (approved):
Thanks for doing this in an extra PR. When CI is green LGTM.

PR #4237, opened 2024-08-06
Fix: Build crash after merging #4228

brustolin:
#4228 broke framework building on main. I guess it was due to some merging not being taken into account.

_#skip-changelog_


philipphofmann (approved):
LGTM

PR #4241, opened 2024-08-07
test: Disable logs for verbose tracer tests

philipphofmann:
Disable logs for SentryTracerTests
* testAddingSpansOnDifferentThread_WhileFinishing_DoesNotCrash
* testFinish_ShouldIgnoreWaitForChildrenCallback_DoesNotCrash
* testFinishAsync

because they can log around 10K log messages.

#skip-changelog

### DIFF
--- a/Sources/Swift/Tools/UIRedactBuilder.swift
+++ b/Sources/Swift/Tools/UIRedactBuilder.swift
@@ -73,7 +73,7 @@ class UIRedactBuilder {
 #else
         ignoreClassesIdentifiers = []
 #endif
-        redactClassesIdentifiers = Set(redactClasses.map( { ObjectIdentifier($0) }))
+        redactClassesIdentifiers = Set(redactClasses.map({ ObjectIdentifier($0) }))
     }
     
     func containsIgnoreClass(_ ignoreClass: AnyClass) -> Bool {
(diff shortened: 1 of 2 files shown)


Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -73,7 +73,7 @@ class UIRedactBuilder {
 #else
         ignoreClassesIdentifiers = []
 #endif
-        redactClassesIdentifiers = Set(redactClasses.map( { ObjectIdentifier($0) }))
+        redactClassesIdentifiers = Set(redactClasses.map({ ObjectIdentifier($0) }))
```
philipphofmann, line 76: SwiftLint changes this. It's a tiny change, and I think we can keep it.

Tests/SentryTests/Transaction/SentryTracerTests.swift
```diff
@@ -1152,44 +1152,47 @@ class SentryTracerTests: XCTestCase {
...
-        
-        func addChildrenAsync() {
-            for _ in 0 ..< 100 {
-                group.enter()
-                queue.async {
-                    let child = sut.startChild(operation: self.fixture.transactionOperation)
-                    Dynamic(child).frames = [] as [Frame]
-                    child.finish()
-                    group.leave()
+        try SentryLog.withOutLogs {
```
philipphofmann, line 1155: I only added this line here. The test is the same. I don't understand why the diff doesn't detect that correctly.

Tests/SentryTests/Transaction/SentryTracerTests.swift
```diff
@@ -207,38 +207,39 @@ class SentryTracerTests: XCTestCase {
...
-            }
-            
-            for _ in 0..<1_000 {
-                let child = sut.startChild(operation: self.fixture.transactionOperation)
-                child.finish()
-            }
-            
-            dispatchQueue.async {
-                for _ in 0..<10 {
+        SentryLog.withOutLogs {
```
philipphofmann, line 210: I only added this line here. The test is the same. I don't understand why the diff doesn't detect that correctly.

brustolin (approved):

PR #4242, opened 2024-08-07
fix: EXC_BAD_ACCESS in SentryMetricProfiler

philipphofmann:
## :scroll: Description

Add missing synchronize block around the clear method.

I detected this problem when running the unit tests locally. We also see a couple of occurrences in our [internal SDK crash detection](https://sentry.sentry.io/issues/?project=4505469596663808&query=is%3Aunresolved+sentrymetricprofiler&referrer=issue-list&sort=user&statsPeriod=14d).

Running SentrySpanTests repeatedly. Previously the `SentrySpanTests` sometimes crashed with `EXC_BAD_ACCESS` in `SentryMetricProfiler`. Now they don't.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryMetricProfiler.mm
+++ b/Sources/Sentry/SentryMetricProfiler.mm
@@ -113,7 +113,7 @@ SentrySerializedMetricEntry *_Nullable serializeContinuousProfileMetricReadings(
 - (instancetype)initWithMode:(SentryProfilerMode)mode
 {
     if (self = [super init]) {
-        [self clear];
+        [self clearNotThreadSafe];
         _mode = mode;
     }
     return self;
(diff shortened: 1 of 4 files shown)


Sources/Sentry/SentryMetricProfiler.mm
```diff
@@ -113,7 +113,7 @@ @implementation SentryMetricProfiler {
 - (instancetype)initWithMode:(SentryProfilerMode)mode
 {
     if (self = [super init]) {
-        [self clear];
+        [self clearNotThreadSafe];
```
brustolin, line 116: l: Why we need a notThreadSafe function in here?

brustolin (approved):

Sources/Sentry/SentryMetricProfiler.mm
philipphofmann: It doesn't make sense to have a synchronized block in the initializer. That's why.

PR #4256, opened 2024-08-08
ci: Remove unit tests Xcode 14 for tvOS and macOS

philipphofmann:
Both tvOS and macOS were using Xcode 14.3 on macos-13, which hasn't been available for a while now. So, the tests were actually running on Xcode 15.2, which uses tvOS 17.2 instead of tvOS 16 and macOS 14.2 instead of macOS 13. We never noticed this. Furthermore, I think it's not worth the computation time to run tests on these OS versions compared to the small risk of finding bugs. We run the tests on tvOS 15 and macOS 12, which should be sufficient.

#skip-changelog

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -104,11 +104,8 @@ jobs:
             xcode: "13.4.1"
             test-destination-os: "latest"
 
-          # macOS 13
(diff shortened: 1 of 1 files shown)


brustolin (approved):
Yes, Thanks!!

Issue #4261, opened 2024-08-08
[Apple] App Hangs improvements


philipphofmann:
### Description

The implementation of AppHangs has a few drawbacks, see https://github.com/getsentry/sentry-cocoa/discussions/2715. The main problem is that non-fully blocking app hangs have random stacktraces, which aren't actionable. Ideally, we plan on linking such app hangs to profiles, but we haven't decided how to achieve this in the product. Until we decide, we can give the users an option to ignore non-fully blocking app hangs. Furthermore, we plan on adding support for the app hang rate, but for that we need to capture fatal app hangs first, which we currently don't do.

Still, we can do some improvements for AppHangs

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4724
- [ ] https://github.com/getsentry/sentry-cocoa/issues/3492
- [ ] https://github.com/getsentry/sentry-cocoa/issues/2216
- [ ] https://github.com/getsentry/sentry-cocoa/issues/4260
```

philipphofmann:
With SR GA, and the upcoming mobile workshop I'm going to host, I see a low chance that I will deliver this by the end of this quarter. I'm sorry, we have to move this, @kahest.

philipphofmann:
@kahest, again, I'm sorry, but the deserializing of events is going a bit slower than expected. I would like to have enough time for Phil for proper onboarding. I would like to push this by another two weeks.

philipphofmann:
I watched my sick family quite a bit in the last week. So this will be delayed for roughly another week.

philipphofmann:
Extracted app hangs rate into its own milestone: https://github.com/getsentry/sentry-cocoa/issues/3319, as discussed with @kahest. All improvements are finished.

Issue #3492, opened 2023-12-13
Add the option to ignore non fully blocking app hangs


philipphofmann:
### Description

Our app hang logic detects app hangs that don't fully block the main thread, which leads to inaccurate stack traces for such app hang events because the app hang logic might fire a bit too late to capture the proper stack trace responsible for the app hang. For some customers non-fully-blocking app hangs are noise, and they often disable the whole feature. Instead, we could provide an option only to report fully blocking app hangs. The option could be an enum something like reportAppHangLevel with (we have to come up with better names)

1. FullyBlocking
2. NonBlocking
3. None

### Implementation

To achieve this, we have to change our implementation a bit and might have to combine it with the CADisplayLink, which we already use for the SentryFramesTracker, but we have to be careful cause this can increase the CPU overhead, as pointed out here https://github.com/getsentry/sentry-cocoa/blob/f25febbb81e40893076ff878f59f146ddb948ab4/Sources/Sentry/include/SentryANRTracker.h#L13-L20

Related GH discussion: https://github.com/getsentry/sentry-cocoa/discussions/2715
Related GH issue for linking app hangs to profiles https://github.com/getsentry/sentry-cocoa/issues/3493.

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4262
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4265
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4282
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4285
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4286
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4356
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4360
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4379
- [ ] https://github.com/getsentry/sentry-docs/pull/11458
```

PR #4262, opened 2024-08-08
chore: Duplicate ANRTracker classes

philipphofmann:
This is the first step for https://github.com/getsentry/sentry-cocoa/issues/3492, which is part of the EPIC AppHang improvements https://github.com/getsentry/sentry-cocoa/issues/4261. 

The changes for adding the option to ignore non-fully blocking app hangs are bloating up, and the PR would be huge. Instead, we first duplicate the AppHangTracker classes, and then we make smaller digestible PRs. Once we're happy with the AppHang changes, we delete the old V1 version and replace it with the new one. We could also run a beta with the V2 first before changing the algorithm for everyone.

#skip-changelog


brustolin (commented):
Looks good. There's not much to review other than the duplication. However, I would keep this PR open and merge the other changes related to ANR into this one, so we don't make a release with this duplication.

Something that I believe fits this PR is a protection to avoid enabling both integrations at the same time. Users should either use V1 or V2

philipphofmann:
> However, I would keep this PR open and merge the other changes related to ANR into this one, so we don't make a release with this duplication.

The downside with that approach is that we could run into possible merge conflicts, and I need to merge main multiple times into this branch. I wouldn't worry too much about the duplication. It's around 500 lines of code.

> Something that I believe fits this PR is a protection to avoid enabling both integrations at the same time. Users should either use V1 or V2

The option V2 is currently in SentryOptions+Private.h, so nobody can enable it. I will ensure that this can't happen in an upcoming PR. Then it's also easier to review.

brustolin (approved):
I would prefer that some changes be made to the classes already. I will leave this to your discretion.

PR #4264, opened 2024-08-08
feat: Pause replay in session mode when offline

brustolin:
## :scroll: Description

If the device has no connection we stop trying to send segments.
We do capture replay for errors,

Unit tests

## :pencil: Checklist


Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayTests.swift
```diff
@@ -251,6 +251,69 @@ class SentrySessionReplayTests: XCTestCase {
...
+        
+        sut.resume()
+        
+        fixture.dateProvider.advance(by: 1)
+        Dynamic(sut).newFrame(nil)
+        XCTAssertNotNil(fixture.screenshotProvider.lastImageCall)
+        
+        fixture.dateProvider.advance(by: 5)
+        Dynamic(sut).newFrame(nil)
+        XCTAssertNotNil(fixture.replayMaker.lastCallToCreateVideo)
```
philipphofmann, line 282: `m`: Wouldn't it make sense to assert some actual values of `lastImageCall` and `lastCallToCreateVideo` to ensure everything works correctly? Only asserting it's not nil doesn't say much.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -74,6 +76,7 @@ - (BOOL)installWithOptions:(nonnull SentryOptions *)options
             return event;
         }];
 
+    [SentryDependencyContainer.sharedInstance.reachability addObserver:self];
```
philipphofmann, line 79: `m`: I know we already use it like this in the `SentryHttpTransport`, but I saw cases in the test logs of CI showing that the simulator became online or offline during tests, triggering the `connectivityChanged` callback causing tests to flake. I would prefer if we could find a way to use test implementation per default for all tests in the `SentrySessionReplayIntegrationTests` to avoid that.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -93,12 +96,29 @@ class SentrySessionReplay: NSObject {
         delegate?.sessionReplayStarted(replayId: sessionReplayId)
     }
 
+    func pause() {
+        isSessionPaused = true
+        videoSegmentStart = nil
+    }
+    
+    func unpause() {
```
philipphofmann, line 104: `m`: Unpause is not used anywhere. I think we can remove it.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -93,12 +96,29 @@ class SentrySessionReplay: NSObject {
         delegate?.sessionReplayStarted(replayId: sessionReplayId)
     }
 
+    func pause() {
+        isSessionPaused = true
+        videoSegmentStart = nil
```
philipphofmann, line 101: `m`: I'm a bit confused about why we need a lock for `videoSegmentStart` in `resume` and here we don't. Is that intended?

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -464,6 +467,17 @@ - (nullable NSString *)currentScreenNameForSessionReplay
...
 }
 
+#    pragma mark - SentryReachabilityObserver
+
+- (void)connectivityChanged:(BOOL)connected typeDescription:(nonnull NSString *)typeDescription
+{
+    if (connected) {
+        [_sessionReplay resume];
+    } else {
+        [_sessionReplay pause];
```
philipphofmann, line 477: `m`: That code runs on a background thread. I wonder if we need to make `resume` and `pause` thread-safe, as they might be called from other threads as well.

https://github.com/getsentry/sentry-cocoa/blob/e0abb7e49c3133b01a2ad0307415661289c2d99e/Sources/Sentry/SentryReachability.m#L200-L201

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
brustolin: the lock in `resume` is for the `displayLink` not the `videosegmentStart`

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayTests.swift
brustolin: I just need to know that nothing was called/not called. The right values are asserted in other tests.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
brustolin: Ow, yes.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: Good idea

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: I will use the lock for pause and resume

philipphofmann (approved):
LGTM 🙏

PR #4266, opened 2024-08-08
chore: debug logging

armcknight:
Some logging I've added while debugging various things, figured I'd put them up to share, and so I don't have to rewrite them yet again in the future (some of these I've done multiple times now)

#skip-changelog


philipphofmann (approved):
LGTM

Issue #4280, opened 2024-08-13
Crash during SDK initialization, presumably due to corrupted envelope data from a previous run


MrMage:
### Platform

macOS

### Environment

Production

### Installed

CocoaPods

### Version

8.33.0

### Xcode Version

15.2

### Steps to Reproduce

No repeatable reproduction steps just yet, but I hope that inspecting the code around `SentrySerialization.m:225` is sufficient to guard against this issue in the future.

### Expected Result

No crash, even when trying to process corrupted envelopes.

### Actual Result

I have received a report of my app crashing at launch from a user. No reports of this crash could be found in Sentry, however.

Manual inspection of the crash report revealed the reason for that: The crash appears to be caused by a crash in Sentry itself while it attempts to send cached envelopes, possibly for a previous crash. My suspicion is that the previous envelope is corrupted, and the code responsible for sending the cached envelopes is not prepared for that.

I can't share a full crash report at this time (it wouldn't be very useful without my debug symbols), anyway. However, I have manually desymbolicated the stack trace, and the following call sites are involved:

```
sentrycrashfu_lastPathEntry (in XXX) (SentryCrashFileUtils.c:190)
-[SentryCrashExceptionApplication reportException:] (in XXX) (SentryCrashExceptionApplication.m:19)
+[SentrySerialization envelopeWithData:] (in XXX) (SentrySerialization.m:225)
+[SentrySerialization envelopeWithData:] (in XXX) (SentrySerialization.m:225)
-[SentryHttpTransport sendAllCachedEnvelopes] (in XXX) (SentryHttpTransport.m:291)
-[SentryHttpTransport initWithOptions:cachedEnvelopeSendDelay:fileManager:requestManager:requestBuilder:rateLimits:envelopeRateLimit:dispatchQueueWrapper:] (in XXX) (SentryHttpTransport.m:97)
+[SentryTransportFactory initTransports:sentryFileManager:currentDateProvider:] (in XXX) (SentryTransportFactory.m:69)
-[SentryClient initWithOptions:fileManager:deleteOldEnvelopeItems:] (in XXX) (SentryClient.m:103)
-[SentryClient initWithOptions:dispatchQueue:deleteOldEnvelopeItems:] (in XXX) (SentryClient.m:93)
-[SentryClient initWithOptions:] (in XXX) (SentryClient.m:75)
+[SentrySDK startWithOptions:] (in XXX) (SentrySDK.m:215)
[...]

PR #4281, opened 2024-08-13
fix: Crash deserializing empty envelope length>0

kahest:
## :scroll: Description

Deserializing empty envelope with length != 0 crashes SDK during init. Unsure what the root cause (creating this kind of invalid payload) is. 

However, it seems we already had [code for that exact case](https://github.com/getsentry/sentry-cocoa/pull/4281/files?diff=split#diff-b1a03b0a05f5b53fc2dcfa9bdeebe0d6285541a37b2eb9b93733b1c2a7e5af58L199) in place but it didn't catch it because the iterator was incremented at the beginning of the loop, so the if checking for whether end of envelope is reached didn't work.

Least intrusive solution is to move incrementing to right before we use the iterator to avoid side effects.

fix GH-4280

Unit test

## :pencil: Checklist

- [x] ~~I updated the docs if needed.~~
- [x] ~~Review from the native team if needed.~~
- [x] ~~No breaking change or entry added to the changelog.~~
- [x] ~~No breaking change for hybrid SDKs or communicated to hybrid SDKs.~~

### DIFF
--- a/Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
+++ b/Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
@@ -114,6 +114,11 @@ class PrivateSentrySDKOnlyTests: XCTestCase {
         let itemData = "{}\n{\"length\":0,\"type\":\"attachment\"}\n".data(using: .utf8)!
         XCTAssertNotNil(PrivateSentrySDKOnly.envelope(with: itemData))
     }
+    
+    func testEnvelopeWithDataLengthGtZero() throws {
+        let itemData = "{}\n{\"length\":1,\"type\":\"attachment\"}\n".data(using: .utf8)!
+        XCTAssertNil(PrivateSentrySDKOnly.envelope(with: itemData))
+    }
 
     func testGetDebugImages() {
         let images = PrivateSentrySDKOnly.getDebugImages()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 ### Fixes
 
 - Session replay not redacting buttons and other non UILabel texts (#4277)
+- Crash deserializing empty envelope length>0 (#4281]
+
 
 ## 8.33.0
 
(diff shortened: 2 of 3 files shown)


Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
```diff
@@ -114,6 +114,11 @@ class PrivateSentrySDKOnlyTests: XCTestCase {
         let itemData = "{}\n{\"length\":0,\"type\":\"attachment\"}\n".data(using: .utf8)!
         XCTAssertNotNil(PrivateSentrySDKOnly.envelope(with: itemData))
     }
+    
+    func testEnvelopeWithDataLengthGtZero() throws {
+        let itemData = "{}\n{\"length\":1,\"type\":\"attachment\"}\n".data(using: .utf8)!
+        XCTAssertNil(PrivateSentrySDKOnly.envelope(with: itemData))
```
kahest, line 120: I expect this to be `nil`, unsure why the `length: 1` test above expects not nil

philipphofmann (approved):
Thank you @kahest 💯

Tests/SentryTests/PrivateSentrySDKOnlyTests.swift
philipphofmann: Because the first states the attachment is empty, which is a valid envelope. Here, we state that the attachment has a length of 1, but there are no bytes for it. So it's an invalid envelope

philipphofmann:
We see two occurrences of this in our [internal SDK crash detection](https://sentry.sentry.io/issues/?project=4505469596663808&query=is%3Aunresolved+envelopeWithData&referrer=issue-list&sort=user&statsPeriod=14d). So this bug occurs rarely.

MrMage:
Thank you for merging this so quickly! However, I would have recommended to also perform a greater-than-or-equals check. Otherwise, if an attachment ever gets written _partially_, a crash could still occur, because currently you are checking only for no attachment being present at all.

I would also recommend investigating how this can occur in the first place — maybe the attachment file is being written incrementally rather than atomically, so a crash or power outage after writing the header but before writing the attachment itself could lead to this condition?

philipphofmann:
@MrMage, yes, that definitely makes sense. This was just a quick fix; we will investigate this problem further, as `SentrySerialization.dataWithEnvelope` does a couple of calls to `NSMakeRange` that can easily lead to a crash if not handled carefully. I reopened the issue.

https://github.com/getsentry/sentry-cocoa/blob/464117d76c03363a8a5699b3edd07c57398c9518/Sources/Sentry/SentrySerialization.m#L95

philipphofmann:
This is related to https://github.com/getsentry/sentry-cocoa/issues/4338.

Issue #4269, opened 2024-08-09
User Feedback integration


armcknight:
### Description

Create the high level integration API to inject a widget that will present a modal form to capture end user feedback.

armcknight:
We'll offer a variety of ways for the form to be presented. See this for inspiration: https://docs.instabug.com/docs/ios-invocation

options:
- widget (#4270)
- some predefined gestures (shake, the other gestures offered by instabug)
- a UIGestureRecognizer they create and hand to us
- triggering from a button they create and offer for us to inject the action
- manually calling a method to display

They will also be able to use a fully custom method to gather the required data and simply call a method like we already have today with `SentrySDK.submitUserFeedback` to submit the envelope.

armcknight:
To update on the above comment after going through the review process in the linked PR, we decided to not accept a custom gesture recognizer in the API for now.

PR #4284, opened 2024-08-15
feat(user feedback): initial API

armcknight:
An initial draft of the API we'll offer to customers for the new User Feedback widget and managed form, for https://github.com/getsentry/sentry-cocoa/issues/4269, including an example of how it would look to configure it, in iOS-Swift.

Docs PR: https://github.com/getsentry/sentry-docs/pull/11191

#skip-changelog


Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
```diff
@@ -154,6 +154,33 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 }
                 return scope
             }
+            
+            options.configureUserFeedback = {
```
bruno-garcia, line 158: Could we split the options that are optional from what's required, some comment blocks would be enough.
Just would love to see what is required first, to get an idea of minimal required setting (we could also do this on a different sample, and have this one show all settings)

armcknight: I think you're suggesting we have a couple different examples of a configuration, one completely default, one minimal, one highly customized?

I'm not sure I follow required vs. optional, as I currently think about all the config settings as optional. You can set the integration, define no options, and get a completely stock experience.

armcknight: Looking at how I'm going to do the docs, I think a collection of examples is better served there. The goal of this snippet in the sample app, for the time being anyways, is to make sure that all the sub-areas of configurations settings function correctly (the main config class, and the auxiliary classes, since i have them split up into general/widget/form/theme settings).

Sources/Swift/SentryUserFeedbackConfiguration.swift
```diff
@@ -0,0 +1,84 @@
...
+ * The settings to use for how the user feedback form is presented, what data is required and how
+ * it's submitted, and some auxiliary hooks to customize the workflow.
+ */
+@available(iOSApplicationExtension 13.0, *)
+@objc public class SentryUserFeedbackConfiguration: NSObject {
+    /**
+     * Configuration settings specific to the managed widget that displays the UI form.
+     * - note: Default: `nil` to use the default widget settings.
+     */
+    public var widget: ((SentryUserFeedbackWidgetConfiguration) -> Void)?
```
brustolin, line 16: l: This name is a little misleading (as is uiForm). Starting the Sentry experience from the options would lead me to believe that this is the widget itself.

Suggestions:
- widgetConfigurator
- configureWidget

Or any name that makes it clear this will be an action rather than an object.

Sources/Swift/SentryUserFeedbackFormConfiguration.swift
```diff
@@ -0,0 +1,144 @@
+import Foundation
+#if (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT
+@_implementationOnly import _SentryPrivate
+import UIKit
+
+/**
+ * Settings to control the behavior and appearance of the UI form.
+ */
+@available(iOSApplicationExtension 13.0, *)
+@objc public class SentryUserFeedbackFormConfiguration: NSObject {
```
brustolin, line 10: l: I believe this should extend `SentryUserFeedbackThemeConfiguration`

brustolin (commented):
Looking good, some comments.

Please, move the Swift files to "Swift/Integrations/FeedbackUser", also in the project organization.

Sources/Swift/SentryUserFeedbackFormConfiguration.swift
armcknight: I'm going for composition, don't want `SentryUserFeedbackFormConfiguration` to duplicate `SentryUserFeedbackThemeConfiguration`'s properties

Sources/Swift/SentryUserFeedbackConfiguration.swift
```diff
@@ -0,0 +1,84 @@
...
+    
+    /**
+     * A custom gesture a user can perform to display the user feedback form.
+     * - note: Setting this property does not disable the widget. In order to do so, you must set `SentryUserFeedbackWidgetConfiguration.autoInject` to `false` using the `SentryUserFeedbackConfiguration.widget` config builder. 
+     */
+    public var gesture: UIGestureRecognizer?
```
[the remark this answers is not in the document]
armcknight: It does make sense that they could call `SentrySDK.gatherUserFeedback()` from their own gesture recognizer delegate. I'm good to drop this, we can document that as a way they can set it up, and if people ask for it in the future, can always revisit this.

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
```diff
@@ -155,6 +155,39 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 }
                 return scope
             }
+            
+            options.configureUserFeedback = { config in
```
bruno-garcia, line 159: SHould we call this: `configureUserFeedbackWidget` ? might be too verobse but we're only configuring the widget behavior here, right? Not `captureUserFeedback` behavior?

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -707,6 +707,20 @@ NS_SWIFT_NAME(Options)
  */
 @property (nullable, nonatomic, copy) SentryBeforeEmitMetricCallback beforeEmitMetric;
 
+/**
+ * Enables the User Feedback integration. See `configureUserFeedback` for options that customize the
+ * experience.
+ * @warning This is an experimental feature and may still have bugs.
+ * @note Default value is @c NO .
+ */
+@property (nonatomic, assign) BOOL enableUserFeedbackIntegration;
```
bruno-garcia, line 716: Is this enabling the widget? if so, should we call it:
```suggestion
@property (nonatomic, assign) BOOL enableUserFeedbackWidget;
```

bruno-garcia (commented):
Maybe my comments dont' make sense if we're consider "user feedback" as the behavior of popping the modal on shake and all the modal settings. But since we've had an API (`captureUserFeedback`) for many years, I find that it might be confusing now with the new capabilities. 

I wonder if we can find a better way to differentiate between the APIs to capture feedback (new `captureFeedback since we're deprecating `captureUserFeedback`) and any automatic behavior like poping up a modal on crash or on shake.

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
armcknight: It's more than just the widget, there is also the form open/close hooks, success/error hooks, whether to use the configured sentry user from scope... we could call it `configureUserFeedbackIntegration`?

Sources/Sentry/Public/SentryOptions.h
armcknight: The integration manages the widget, but looking at the other properties in options that manage enabling integrations, we don't actually put "Integration" in the name, so I'd _would_ be ok just naming it enableUserFeedback, as long as we don't think there's confusion with the preexisting API as you mentioned above (which is why I tacked on the "Integration" here)

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackWidgetConfiguration.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+#if (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT
+@_implementationOnly import _SentryPrivate
+import UIKit
+
+/**
+ * Settings for whether to show the widget and how it should appear.
+ */
+@objc public class SentryUserFeedbackWidgetConfiguration: NSObject {
```
brustolin, line 9: m: Maybe we should add options to configure the color/format/image of the widget?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegration.swift
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * An integration managing a workflow for end users to report feedback via Sentry.
+ * - note: The default method to show the feedback form is via a floating widget placed in the bottom trailing corner of the screen. See the configuration classes for alternative options.
+ */
+class SentryUserFeedbackIntegration {
+    let configuration: SentryUserFeedbackConfiguration
+    private var widgetConfig: SentryUserFeedbackWidgetConfiguration?
+    private var formConfig: SentryUserFeedbackFormConfiguration?
+    private var lightThemeOverrides: SentryUserFeedbackThemeConfiguration?
+    private var darkThemeOverrides: SentryUserFeedbackThemeConfiguration?
```
brustolin, line 15: m: What will this light and dark theme be used for, since formConfig also have it?

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
bruno-garcia: tbqh I don't see the value in appending `Integration` to the name. We do have the `Integration` properties on options but even Integration 'types' I don't think needs appending `Integration`.
In that case `configureUserFeedback` is a better choice imo. 

Since we're renaming the func from `captureUserFeedback` to `captureFeedback`, how about `configureFeedback` ? Or `configureFeedbackUI` if options under that are all related to the behavior of the UI components.

We need to be clear that the success/failure hooks won't be called if you simply called `captureFeedback`, right? That's just when using the modal. At least on JS that's the way afaik

Sources/Sentry/Public/SentryOptions.h
armcknight: That's fair

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackWidgetConfiguration.swift
armcknight: Maybe! @bruno-garcia wdyt? It's not in any of the other SDKs, we just provide the option to use a fully custom-built button. But theming the stock button makes sense IMO too.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegration.swift
armcknight: The properties on the form config object are the builder closures, these are the actual objects that are passed to the builders for modification and then stored until needed. See https://github.com/getsentry/sentry-cocoa/blob/d1dd7999f98b2366ab390c665d90b86148ca5cd7/Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegration.swift#L30-L39

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
```diff
@@ -0,0 +1,161 @@
+import Foundation
+#if (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT
+@_implementationOnly import _SentryPrivate
+import UIKit
+
+/**
+ * Settings to control the behavior and appearance of the UI form.
+ */
+@objc public class SentryUserFeedbackFormConfiguration: NSObject {
```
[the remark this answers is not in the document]
armcknight: Nice tip 👍🏻

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
```diff
@@ -0,0 +1,161 @@
...
+     * Builder for light mode theme overrides.
+     * - note: Default: `nil`
+     */
+    @objc public var lightThemeOverrides: ((SentryUserFeedbackThemeConfiguration) -> Void)?
+    
+    /**
+     * Builder for dark mode theme overrides.
+     * - note: Default: `nil`
+     */
+    @objc public var darkThemeOverrides: ((SentryUserFeedbackThemeConfiguration) -> Void)?
```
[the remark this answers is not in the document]
armcknight: Ah you're right, we do simply need `theme`/`themeOverrides`, at least for iOS versions older than those that support dark mode, and it can default to light mode. 

For those iOS versions that do support dark mode, we should check which mode the OS is currently in, and use the light or dark config.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackThemeConfiguration.swift
```diff
@@ -0,0 +1,111 @@
...
+     * - note: Default light mode: `rgb(255, 255, 255)`; dark mode: `rgb(41, 35, 47)`
+     */
+    @objc public var background: UIColor = {
+        let lightModeDefault = UIColor.white
+        if #available(iOSApplicationExtension 12.0, *) {
+            return UIScreen.main.traitCollection.userInterfaceStyle == .dark ? UIColor(red: 41 / 255, green: 35 / 255, blue: 47 / 255, alpha: 1) : lightModeDefault
+        } else {
+            return lightModeDefault
+        }
+    }()
```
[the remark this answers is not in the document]
armcknight: I'll let @bruno-garcia and the design team respond to this.

Sources/Sentry/Public/SentryOptions.h
armcknight: LOL I can tell it was useless because I hadn't even set it in the sample apps 😆

Sources/Sentry/Public/SentrySDK.h
```diff
@@ -238,6 +238,15 @@ SENTRY_NO_INIT
...
     NS_SWIFT_NAME(capture(userFeedback:));
 
+/**
+ * Display a form to gather information from an end user in the app to send to Sentry as a user
+ * feedback event.
+ * @see @c SentryOptions.enableUserFeedbackIntegration and @c SentryOptions.configureUserFeedback to
+ * enable the functionality and customize the experience.
+ * @note If @c SentryOptions.enableUserFeedbackIntegration is @c NO, this method is a no-op.
+ */
++ (void)gatherUserFeedback;
```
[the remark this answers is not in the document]
armcknight: I went with `showUserFeedbackForm`

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
armcknight: > Since we're renaming the func from `captureUserFeedback` to `captureFeedback`

I forget where we had this conversation, but I'm not sure it's actually necessary. For now, I've added notes in the headerdoc of `SentrySDK.captureUserFeedback` (the current one) and `SentrySDK.showUserFeedbackForm` (the new one that immediately pops up the new form) as well as `SentryOptions.configureUserFeedback` itself. See if you think those are sufficient, and if not, we can brainstorm more on a better name for `configureUserFeedback`.

Yes, the hooks are not called with the preexisting function. Only with our UI form. I made a note of it in the headerdocs for `onSubmitSuccess`/`onSubmitError`.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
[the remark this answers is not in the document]
armcknight: Hmm, actually, what do you think about just having two: `themeOverrides` and `darkThemeOverrides`?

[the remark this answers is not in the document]
brustolin: Yeah, Im fine with `themeOverrides` and `darkThemeOverrides`

bruno-garcia:
Also adding @c298lee and @billyvg from the Replay team (they are familiar with the Feedback SDK in JS)

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
philipphofmann: If we plan to rename `captureUserFeedback` to `captureFeedback`, I think we should remove `user` from the naming here to be consistent. 

> It's more than just the widget, there is also the form open/close hooks, success/error hooks, whether to use the configured sentry user from scope

But still, all that functionality could be part of the widget. For me, `configureFeedbackWidget` sounds like a reasonable name. We could rename the `SentryUserFeedbackWidgetConfiguration` to `SentryFeedbackTriggerButtonConfiguration` or something similar to avoid the naming collision. The `SentryUserFeedbackWidgetConfiguration` is only a button to trigger showing the widget.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackWidgetConfiguration.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+#if (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT
+@_implementationOnly import _SentryPrivate
+import UIKit
+
+/**
+ * Settings for whether to show the widget and how it should appear.
+ */
+@objcMembers public class SentryUserFeedbackWidgetConfiguration: NSObject {
```
philipphofmann, line 9: `m`: Isn't this more of a button that opens the feedback form / widget. We could rename this to something like `
SentryFeedbackTriggerConfiguration` and then we could rename `triggerLabel` to just `label` and `triggerAccessibilityLabel` to just `accessibilityLabel`.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackWidgetConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+     * The location for positioning the widget.
+     * - note: Default: `[.bottom, .right]`
+     */
+    public var location: UIRectEdge = [.bottom, .right]
+    
+    /**
+     * The distance to use from the widget button to the superview's `layoutMarginsGuide`.
+     * - note: Default: `UIOffset.zero`
+     */
+    public var layoutOffset: UIOffset = UIOffset.zero
```
philipphofmann, line 46: `l`: Maybe that's a better name to clearly reflect he type in the naming.
```suggestion
    public var layoutUIOffset: UIOffset = UIOffset.zero
```

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
```diff
@@ -0,0 +1,164 @@
...
+     * The label of the submit button used in the feedback form.
+     * - note: Default: `"Send Bug Report"`
+     */
+    public var submitButtonLabel: String = "Send Bug Report"
+    
+    /**
+     * The label of cancel buttons used in the feedback form.
+     * - note: Default: `"Cancel"`
+     */
+    public var cancelButtonLabel: String = "Cancel"
```
philipphofmann, line 138: `m`: Does it make sense to add accessibility labels for all the buttons here?

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
```diff
@@ -0,0 +1,87 @@
...
+     * - note: Default: `false`
+     */
+    public var draggable: Bool = false
+    
+    /**
+     * Use a shake gesture to display the form.
+     * - note: Default: `false`
+     * - note: Setting this to true does not disable the widget. In order to do so, you must set `SentryUserFeedbackWidgetConfiguration.autoInject` to `false` using the `SentryUserFeedbackConfiguration.configureWidget` config builder.
+     */
+    public var useShakeGesture: Bool = false
```
philipphofmann, line 28: `m`: We don't have these two features on the web, I believe. While they add value, we could skip them in the first iteration if we want to narrow the scope a bit.

philipphofmann (commented):
This looks great, the biggest questions are how do we name the root config, do we remove user from the naming, what is the widget exactly and what is a form. We can also discuss this on a call cause these discussions could take forever on GH. The rest LGTM.

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
armcknight: > The SentryUserFeedbackWidgetConfiguration is only a button to trigger showing the widget

This confuses what is a widget and what is the form. 

The button that triggers the form to display is the widget.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackWidgetConfiguration.swift
armcknight: Form and widget are two different things, see https://github.com/getsentry/sentry-cocoa/pull/4284#discussion_r1765741874

That being said we can certainly rename things to remove the `trigger` prefix from widget config properties. These were mostly copied wholesale from the javascript SDK's implementation/docs.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
armcknight: Yes, I think that would be nice.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
armcknight: You are correct. I think the shake gesture would be trivial to add, but we can defer the widget draggability feature.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackWidgetConfiguration.swift
armcknight: > to just accessibilityLabel

`accessibilityLabel` is actually a property already on `NSObject` so we can't call it this. I went with `widgetAccessibilityLabel`
image

armcknight: Also went with `labelText` instead of just `label`. Could also choose `title` if people like that more.

armcknight:
> automatic behavior like poping up a modal on crash or on shake

@bruno-garcia I do have the shake trigger in here already, but popping a dialog on crash detection is a good one too 👍🏻

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
philipphofmann: I leave it up to you to make the best choice then for the API. I guess we will do an experimental release first, so we can still change it if we get some feedback. Also, this isn't an essential API of our SDK. You most likely set it up once and then leave it. Still, it shouldn't be confusing.

philipphofmann (approved):
The user-facing API, LGTM.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackWidgetConfiguration.swift
armcknight: Will leave this out for now, can always add it later. Noted for future consideration.

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
billyvg: On the web SDK we vaguely refer to the button + form as the widget. For the button, we call it `actor` in some parts of the code, and publicly [here](https://docs.sentry.io/platforms/javascript/user-feedback/configuration/) we call it `trigger`

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackThemeConfiguration.swift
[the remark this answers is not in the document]
armcknight: For clarification @brustolin , are you saying we should not supply our own color palette and simply use unthemed UI elements? I don't think this is what is done for the javascript SDK, for instance.

armcknight:
Pushed a commit to remove this from public access while we merge intermediate PRs

Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
armcknight: `actor`, `widget`, `button`, `form`, `modal` and `trigger` all have specific meanings in the Apple Human Interface Guidelines and Developer documentation that our customers will already be familiar with. We should avoid overloading or blurring the lines between them, and instead use the appropriate terms to convey what each piece of this feature will do.

- [`trigger`](https://developer.apple.com/documentation/swiftui/primitivebuttonstyleconfiguration/trigger()): performs [a] button’s action ([also](https://developer.apple.com/documentation/realitykit/creating-a-trigger): Define when and how a behavior fires) 
- [`actor`](https://developer.apple.com/documentation/swift/actor): actors execute tasks on a shared global concurrency thread pool
- [`button`](https://developer.apple.com/design/human-interface-guidelines/buttons): initiates an instantaneous action ([also](https://developer.apple.com/documentation/swiftui/button): A control that initiates an action)
- [`widget`](https://developer.apple.com/design/human-interface-guidelines/widgets): People tap or click a widget to launch its corresponding app (in our case, we aren't launching an app as we're an SDK inside an app–we're displaying a modal screen inside that app)
- [`form`](https://developer.apple.com/documentation/swiftui/form): A container for grouping controls used for data entry
- [`modal`](https://developer.apple.com/design/Human-Interface-Guidelines/modality): Modality is a design technique that presents content in a separate, dedicated mode that prevents interaction with the parent view and requires an explicit action to dismiss

The way I would describe this feature:
> The User Feedback **integration** embeds a **widget** in your application that, when pressed, displays a **modal** **form** for your end user to fill in details of their experience that will help you become aware of and fix issues, and to generally improve your product.

Implicitly, the **widget** is (or at least contains) a **button** (although UIButton has always been able to contain both a textual and image component, which is all our widget UI has), and that button's **trigger** is opening the **form**.

I would completely avoid using the term "actor" when describing this feature. I also don't see that term at all on the docs page. For comparison, here is the first description of how the feature works from our [Javascript doc](https://docs.sentry.io/platforms/javascript/user-feedback/#user-feedback-widget): 
> The embeddable JavaScript widget allows users to submit feedback from anywhere inside your application. The [Crash-Report Modal](https://docs.sentry.io/#crash-report-modal) collects reactive feedback tied to an error event

On the topic of not wanting to append "integration" onto the end of this options property to call it `configureFeedbackIntegration`, note that's essentially what the [Javascript SDK](https://docs.sentry.io/platforms/javascript/user-feedback/configuration/#general) does:
> The following options can be configured for the integration in feedbackIntegration({})

PR #4297, opened 2024-08-26
fix: Crash when reading corrupted envelope

brustolin:
## :scroll: Description

Added a safe guard to prevent crashing when reading corrupted envelope

This was mentioned in #4280

Unit tests

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentrySerialization.m
+++ b/Sources/Sentry/SentrySerialization.m
@@ -212,7 +212,7 @@ NS_ASSUME_NONNULL_BEGIN
             if (bodyLength > 0 && data.length < (i + 1 + bodyLength)) {
                 SENTRY_LOG_ERROR(@"Envelope is corrupted or has invalid data. Trying to read %li "
                                  @"bytes by skiping %li from a buffer of %li bytes.",
-                                 (unsigned long)data.length, bodyLength, i + 1);
+                    (unsigned long)data.length, bodyLength, i + 1);
                 return nil;
             }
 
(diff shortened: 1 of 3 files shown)


Sources/Sentry/SentrySerialization.m
```diff
@@ -208,6 +208,12 @@ + (SentryEnvelope *_Nullable)envelopeWithData:(NSData *)data
             if (endOfEnvelope == i) {
                 i++; // 0 byte attachment
             }
+
+            if (data.length < (i + 1 + bodyLength)) {
+                SENTRY_LOG_DEBUG(@"Envelope is corrupted or has invalid data.")
```
kahest, line 213: `m`: info on lengths (data.length, i, bodyLength) would probably be helpful in the log

kahest: maybe also higher log level, similar to https://github.com/getsentry/sentry-cocoa/pull/4297/files#diff-b1a03b0a05f5b53fc2dcfa9bdeebe0d6285541a37b2eb9b93733b1c2a7e5af58R183

Tests/SentryTests/Helper/SentrySerializationTests.swift
```diff
@@ -267,6 +267,17 @@ class SentrySerializationTests: XCTestCase {
         XCTAssertNil(actual)
     }
     
+    func testReturnNilForCorruptedEnvelope() throws {
+        let envelope = SentryEnvelope(event: Event(error: NSError(domain: "test", code: -1, userInfo: nil)))
+        let data = try XCTUnwrap(SentrySerialization.data(with: envelope))
+        
+        let corruptedData = data[0..<data.count - 1]
+        
+        let unserialized = SentrySerialization.envelope(with: corruptedData)
```
kahest, line 276: `m`: maybe add a happy path version of the same test where we just deserialize the data object like `SentrySerialization.envelope(with: data)` (didn't find a test that does exactly this)

brustolin: We have other tests that serialize and deserialize the envelope, like `testSentryEnvelopeSerializer_WithSingleEvent`

PR #4303, opened 2024-08-27
Fix Resumes replay when the app becomes active

brustolin:
Changing the SR resume from "app enter foreground" to "app became active" to align with session tracker.

This will also prevent the problem happening in [here](https://github.com/getsentry/sentry-react-native/issues/4047).


krystofwoldrich:
Nice, such a small change.

Is it possible to tests that the new session is created before the replay starts, I assume that was the root issue? That the replay start before the new session so it was appended to the previous session?

brustolin:
> Is it possible to tests that the new session is created before the replay starts, I assume that was the root issue? That the replay start before the new session so it was appended to the previous session?

We have tests to ensure that both session tracking and session replay end with the same type of notification. Therefore, the order in which they are called does not matter; both actions will occur before the next frame, which prevents a new session segment from being sent.

krystofwoldrich (approved):
Looks good! 🚀

PR #4305, opened 2024-08-29
Track thermal state

gabriellanata:
## :scroll: Description

Adds tracking of device thermal state

Thermal state is an important measurement because it can explain device throttling that commonly cause app hangs.

Added to existing tests and ran tests locally

## :pencil: Checklist


scripts/xcode-test.sh
```diff
@@ -14,7 +14,7 @@ REF_NAME="${3-HEAD}"
 IS_LOCAL_BUILD="${4:-ci}"
 COMMAND="${5:-test}"
 DEVICE=${6:-iPhone 14}
-CONFIGURATION_OVERRIDE="${7}"
+CONFIGURATION_OVERRIDE="${7:-}"
```
brustolin, line 17: m: Why do we need this change?

gabriellanata: This is technically unrelated, but right now calling `make test` doesn't work because it doesn't find this environment property. I'm guessing it's mostly being run from CI which sets the property and it was an oversight to not include this as optional for local runs

Sources/Sentry/SentryExtraContextProvider.m
```diff
@@ -44,6 +44,23 @@ - (NSDictionary *)getExtraDeviceContext
...
+        break;
+    case NSProcessInfoThermalStateSerious:
+        extraDeviceContext[@"thermal_state"] = @"serious";
+        break;
+    case NSProcessInfoThermalStateCritical:
+        extraDeviceContext[@"thermal_state"] = @"critical";
+        break;
+    default:
+        break;
+    }
```
brustolin, line 62: h: The only thing that I would change is to extract this naming from here.
Something like what we have in [here](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentryDataCategoryMapper.m).

brustolin: And to align with future android plans we need to change the names:

NSProcessInfoThermalStateNominal = "light"
NSProcessInfoThermalStateFair = "moderate"
NSProcessInfoThermalStateSerious = "severe"
NSProcessInfoThermalStateCritical = "critical
"

kahest: reading the [iOS](https://developer.apple.com/documentation/foundation/nsprocessinfothermalstate) and [Android](https://developer.android.com/reference/android/os/PowerManager#THERMAL_STATUS_NONE) docs, I think these don't correlate. on iOS it's more about the thermal state, whereas on Android it's about how severely system is affected by thermal state. e.g. `moderate` on Android means "Thermal status code: Moderate throttling where UX is not largely impacted." which means it's already elevated and impacts UX at least to some extent.

TLDR: I wouldn't try to match/correlate and instead just report as-is

brustolin: Ok. Lets keep the name as is then.

gabriellanata: @brustolin updated to extract naming to constants at the top. Do you also want a separate file that does the mapping or are the extracted strings enough? I based it on the phone orientation property right below it

brustolin (approved):
LGTM

Sources/Sentry/SentryExtraContextProvider.m
brustolin: All good, thanks for the help.

brustolin (changes requested):
Some of the new tests are failing.

gabriellanata:
@brustolin Seems like the failures are unrelated to my changes: `Error cloning certificates git repo, please make sure you have access to the repository - see instructions above` Is there something i should do about that?

armcknight:
Yeah those permission errors happen whenever a PR comes in from a fork, because the fork doesn't have access to the secrets in our CI settings.

But there are some related failures, like https://github.com/getsentry/sentry-cocoa/actions/runs/10649320635/job/29561450624?pr=4305#step:11:38

gabriellanata:
Ok i think the test is fixed

Sources/Sentry/SentryExtraContextProvider.m
```diff
@@ -44,6 +49,23 @@ - (NSDictionary *)getExtraDeviceContext
...
+        extraDeviceContext[@"thermal_state"] = kSentryProcessInfoThermalStateFair;
+        break;
+    case NSProcessInfoThermalStateSerious:
+        extraDeviceContext[@"thermal_state"] = kSentryProcessInfoThermalStateSerious;
+        break;
+    case NSProcessInfoThermalStateCritical:
+        extraDeviceContext[@"thermal_state"] = kSentryProcessInfoThermalStateCritical;
+        break;
+    default:
+        break;
```
armcknight: I'm pushing a commit with a log here

Tests/SentryTests/Helper/SentryExtraContextProviderTests.swift
```diff
@@ -64,11 +64,14 @@ final class SentryExtraContextProviderTests: XCTestCase {
...
         let sut = fixture.getSut()
         fixture.processWrapper.overrides.processorCount = 12
-        
+        fixture.processWrapper.overrides.thermalState = .critical
+
         let actualContext = sut.getExtraContext()
         let device = actualContext["device"] as? [String: Any]
         
         XCTAssertEqual(device?["processor_count"] as? UInt, fixture.processWrapper.overrides.processorCount)
+        XCTAssertEqual(device?["thermal_state"] as? ProcessInfo.ThermalState,
```
armcknight: Needed to extract this as a String, not a thermal state enum value.
```suggestion
        XCTAssertEqual(device?["thermal_state"] as? ProcessInfo.ThermalState,
```

scripts/xcode-test.sh
armcknight: thanks for catching and correcting this 🙏🏻

armcknight (approved):
I pushed a couple quick commits, but everything else here looks great, thanks for adding this. Confirmed it shows up in a test issue: 
image

gabriellanata:
cc @brustolin

brustolin (approved):

brustolin:
Thanks very much @gabriellanata

PR #4307, opened 2024-09-03
chore: Adds SentryFileManager and SentrySerialization to dangerous files

kahest:

Adds SentryFileManager and SentrySerialization to list of dangerous files. A recent (GH-4219) change in these files is the likely root cause of invalid envelopes that subsequently caused crashes.

#skip-changelog

n/a

## :pencil: Checklist

- [ ] ~~I added tests to verify the changes.~~
- [ ] ~~I updated the docs if needed.~~
- [ ] ~~Review from the native team if needed.~~


brustolin (approved):
LGTM

We should discuss the procedure we should take when some of these files are changed. The alert itself would not prevent the previous problem since the changes to these files were the target of the PR. In these cases, the alert is ignored.

PR #4311, opened 2024-09-03
fix: remove object_literal from swiftlint rules

armcknight:
I am unable to make actual use of `#colorLiteral` as it just shows a broken image (and Xcode support for them in general is very poor) 
image

and we don't have a need for image literals, we should just remove this rule

#skip-changelog


brustolin (approved):
Yeah, `object_literal` is a bad rule, it does not even make sense in an app project

PR #4312, opened 2024-09-04
chore: Added UI Test to check invalid envelope is handled correct

brustolin:
This is a test that will detect the problem introduced in v8.33.0

Running the same test in v8.33 would result in the following: 

Screenshot 2024-09-04 at 16 23 14

_#skip-changelog_


Samples/iOS-Swift/iOS-Swift-UITests/LaunchUITests.swift
```diff
@@ -92,6 +92,15 @@ class LaunchUITests: BaseUITest {
         XCTAssertTrue(0.5 > frozenFramesPercentage, "Too many frozen frames.")
     }
     
+    func testCorruptEnvelope() {
```
kahest, line 95: pls add a comment with a link to GH-4280 and describing on a high level how the test works 🙏

Samples/iOS-Swift/iOS-Swift-UITests/LaunchUITests.swift
```diff
@@ -92,6 +92,20 @@ class LaunchUITests: BaseUITest {
...
+        // This is to prevent https://github.com/getsentry/sentry-cocoa/issues/4280
+        // Tapping "Corrupt Envelope" will try to capture an envelope but it closes the app
+        // in the middle of the process. For 8.33.0 this would generate a corrupted envelope.
+        // 8.35.0 does not write the envelope to disk at all.
```
kahest, line 99:
```suggestion
        // 8.35.0+ reverts to writing envelopes atomically, so that in this scenario no envelope is written, as is expected.
```

armcknight (approved):
Looks good once the last suggestion from @kahest goes in 👍🏻

PR #4314, opened 2024-09-05
fix: Session replay for crash not created because of a race condition

brustolin:
## :scroll: Description

This fix an issue where the SR is not sent for crashes that mostly happens in RN.

Unit tests

## :pencil: Checklist


krystofwoldrich (approved):
I don't see why the new replay was started before the previous was processed. 

But the fix looks good. 👍

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -308,13 +308,20 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
     }
     
     func createLastSessionReplay(writeSessionInfo: Bool = true, errorSampleRate: Double = 1) throws {
-        let replayFolder = SentryDependencyContainer.sharedInstance().fileManager.sentryPath + "/replay"
-        let jsonPath = replayFolder + "/lastreplay"
+        let options = Options()
+        options.dsn = "https://user@test.com/test"
+        options.cacheDirectoryPath = FileManager.default.temporaryDirectory.path
```
armcknight, line 313: Doesn't have to hold up this pr, but consider that using a wrapper for this in SentryFileManager and using a predictable, persistent location would allow for better debugging, especially from CI

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -169,6 +172,10 @@ - (void)resumePreviousSessionReplay:(SentryEvent *)event
     eventContext[@"replay"] =
         [NSDictionary dictionaryWithObjectsAndKeys:replayId.sentryIdString, @"replay_id", nil];
     event.context = eventContext;
+
+    if ([NSFileManager.defaultManager removeItemAtURL:lastReplayURL error:&error] == NO) {
+        SENTRY_LOG_ERROR(@"Can`t delete '%@': %@", SENTRY_LAST_REPLAY, error);
+    }
```
armcknight, line 178: We should expose `SentryFilemanager._non_thread_safe_removeFileAtPath` and reuse it here, it has even better logging for different kinds of errors

armcknight (approved):
Looks fine. I would prefer to see SentryFileManager instead of naked NSFileManager usages. Even if we don't use an instance of SentryFileManager, we should be breaking out the logic into static functions so we can reuse what we've already built, instead of always rewriting the same error handling.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
brustolin: I tried that, SentryFileManager has a weird initialisation, it was easier to hard coded it, which is also predictable I guess.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: Thanks for the tip, It makes sense, I will create a note for refactoring later.

PR #4315, opened 2024-09-05
Use `Sentry.framework` instead of `Sentry.xcframework`

denrase:
#skip-changelog

## :scroll: Description

Replaced `Sentry.xcframework` with `Sentry.framework` in the watchOS sample, so it is in line with the other apps.

Relates to [#4300](https://github.com/getsentry/sentry-cocoa/issues/4300)

## :pencil: Checklist


brustolin:
> The watchOS sample was not building

I think this is by design, we have a script that make it work. I dont know the reason tho.
If nothing is breaking I guess we could change that.
@philipphofmann is back in a week if I'm not mistaken, we can double check with him.

armcknight:
Yeah I'm not sure why the projects weren't set up the same way. I am pushing a small commit to try to fix the build > watchOS sample CI failure. I made the watchOS app's project and linkages mirror iOS-Swift's. 

Before (on this branch as of my review):

image

After my commit:

image

armcknight:
So one more thing I saw when trying to reproduce that last failure locally, is that if you build the watchOS App scheme, it automatically builds the sentry sdk for iphone, and then can't find the build product because it's not in the right derived data tree. But if you build the _extension_, it gets it right. Pushing a commit to change that.

image

Sources/Configuration/SDK.xcconfig
```diff
@@ -17,7 +17,6 @@ CLANG_ENABLE_OBJC_ARC = YES
 CLANG_ENABLE_MODULES = YES
 //OTHER_CFLAGS = -Wall -Wextra -Wpedantic -Wno-gnu-conditional-omitted-operand
 
-ONLY_ACTIVE_ARCH[config=Debug] = YES
```
armcknight, line 20: This did nothing so I just removed it as part of cleaning things up
image

armcknight:
I'm having trouble finding the right invocation for the watchOS sample app build to succeed in CI 😕

denrase:
@armcknight Thx for the info, i will check if i can make it run on ci!

denrase:
@armcknight Doing the same in CI as for the macOS and iOS seems to work.

@philipphofmann I removed the code to generate the xcframework just for the watchOS sample project.

denrase:
@armcknight Do we still need the `watchOS-Swift WatchKit Extension.xcscheme` or can it be removed again?

armcknight:
@denrase I'm not 100% sure. But if you got it working without it, then chuck it! We can always add it back and figure it out if we need it again for some reason.

armcknight (approved):
Thanks @denrase , so much nicer to have this work consistently like the other sample apps.

PR #4316, opened 2024-09-05
Fix: testCorruptEnvelope Test

brustolin:
## :scroll: Description

Fixing flakiness of the test.

_#skip-changelog_

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift-UITests/LaunchUITests.swift
+++ b/Samples/iOS-Swift/iOS-Swift-UITests/LaunchUITests.swift
@@ -103,6 +103,7 @@ class LaunchUITests: BaseUITest {
         
         app.launch()
         app.tabBars.firstMatch.waitForExistence("App did not open again")
+        XCTAssertEqual(app.state, .runningForeground)
     }
     
     func testCheckTotalFrames() {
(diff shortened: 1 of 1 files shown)


Samples/iOS-Swift/iOS-Swift-UITests/LaunchUITests.swift
```diff
@@ -102,7 +102,6 @@ class LaunchUITests: BaseUITest {
         XCTAssertEqual(app.state, .notRunning)
         
         app.launch()
-        XCTAssertEqual(app.state, .runningForeground)
```
brustolin, line 105: Removing this because checking for the app state right after closing the app can be flake and fail in CI.
`waitForExistence` in the next line has a 10 seconds timeout to wait for the app to open again.

kahest: `h`: please test this against 8.33.0 and ensure the test still fails

brustolin: Yes it does fail.

kahest: does it work and make sense to switch the lines, i.e. first `waitForExistence` and then the assertion that app is running

brustolin: It does work. Not sure if it makes a different since waitForExistence does need the app to be open, but I also dont think it will harm.

kahest: then let's keep it pls

kahest (approved):

PR #4325, opened 2024-09-11
fix: Don't redact clipped views

brustolin:
## :scroll: Description

A label that is outside of bounds of a view with clipBounds true is no longer being redacted.

|Screen|Before|After|
|-|-|-|
|image|747730062 107031 copy||

Unit tests

## :pencil: Checklist


armcknight (approved):
Thanks for the extra explanation in the comments 🙏🏻 Seems fine

PR #4326, opened 2024-09-11
ref:  Moving session replay public API and default quality

brustolin:
## :scroll: Description

Moved session replay API to `SentrySDK.replay` and changed default quality to `medium`

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/Public/SentryReplay.h
+++ b/Sources/Sentry/Public/SentryReplay.h
@@ -1,9 +1,9 @@
 #import <Foundation/Foundation.h>
 
 #if __has_include(<Sentry/SentryDefines.h>)
-#import <Sentry/SentryDefines.h>
+#    import <Sentry/SentryDefines.h>
 #else
-#import <SentryWithoutUIKit/SentryDefines.h>
+#    import <SentryWithoutUIKit/SentryDefines.h>
 #endif
 
 #if SENTRY_TARGET_REPLAY_SUPPORTED
--- a/Sources/Sentry/Public/SentryReplay.h
+++ b/Sources/Sentry/Public/SentryReplay.h
@@ -1,5 +1,10 @@
 #import <Foundation/Foundation.h>
+
+#if __has_include(<Sentry/SentryDefines.h>)
 #import <Sentry/SentryDefines.h>
+#else
+#import <SentryWithoutUIKit/SentryDefines.h>
+#endif
 
 #if SENTRY_TARGET_REPLAY_SUPPORTED
 
--- a/Sources/Sentry/Public/SentryReplay.h
+++ b/Sources/Sentry/Public/SentryReplay.h
@@ -1,0 +1,45 @@
+#import <Foundation/Foundation.h>
+#import <Sentry/SentryDefines.h>
+
+#if SENTRY_TARGET_REPLAY_SUPPORTED
+
+@class UIView;
+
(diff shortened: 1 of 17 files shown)


Sources/Sentry/Public/SentryReplay.h
```diff
@@ -0,0 +1,50 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentryReplay : NSObject
+
+/**
+ * Marks this view to be redacted during replays.
+ *
+ * @warning This is an experimental feature and may still have bugs.
+ */
+- (void)replayRedactView:(UIView *)view;
```
romtsn, line 22: nit: I think we can drop the `replay` part now, and just have `redactView` since it's under the SentryReplay namespace already

Sources/Sentry/Public/SentryReplay.h
```diff
@@ -0,0 +1,50 @@
...
+#    import <SentryWithoutUIKit/SentryDefines.h>
+#endif
+
+#if SENTRY_TARGET_REPLAY_SUPPORTED
+
+@class UIView;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentryReplay : NSObject
```
romtsn, line 15: also, dunno if you want to be consistent with naming, but metrics is called `SentryMetricsAPI`. For me it's fine either way though

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -96,15 +96,15 @@ class SentrySessionReplay: NSObject {
         delegate?.sessionReplayStarted(replayId: sessionReplayId)
     }
 
-    func pause() {
+    func pauseSessionMode() {
```
romtsn, line 99: just curious: would it be possible to drop this method entirely, and just call `pause()` when there's no internet connection? We'll stop getting display link callbacks anyway and not capture new frames, right?

romtsn: just thinking that we have `isSessionPaused` variable, and then we set it to `false` in the `pause` method which might look confusing

romtsn (approved):

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
brustolin: - The pause function stops everything.
- `pauseSessionMode` keeps taking screenshots when replay is in buffer mode to at least capture the first error.

I dont think we can remove `pauseSessionMode`, but I need to reevaluate if we should disable `isSessionPaused` during pause. This needs to be in another PR tho.

romtsn: > pauseSessionMode keeps taking screenshots when replay is in buffer mode to at least capture the first error.

does it keep taking screenshots or am I misinterpreting this line? https://github.com/getsentry/sentry-cocoa/blob/e1c38196160220f9d7580685802a1b2a5b580c29/Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift#L181

brustolin: Yes, it keeps taking screenshots. But only for buffer mode

PR #4327, opened 2024-09-12
ref: Changing the redact logic

brustolin:
## :scroll: Description

Before this PR, `redactAllText` was wrongly used as a "redact is enabled" flag.

Now we're using `redactAllText` and `redactAllImages` as an indicator to how start the redact list:

```swift
if options.redactAllText {
       redactClasses += [ UILabel.self, UITextView.self, UITextField.self ]
}
        
if options.redactAllImages {
    //this classes are used by SwiftUI to display images.
    redactClasses += ["_TtCOCV7SwiftUI11DisplayList11ViewUpdater8Platform13CGDrawingView",
             "_TtC7SwiftUIP33_A34643117F00277B93DEBAB70EC0697122_UIShapeHitTestingView",
             "SwiftUI._UIGraphicsView", "SwiftUI.ImageLayer"
    ].compactMap { NSClassFromString($0) }
            
    redactClasses.append(UIImageView.self)
}
   ```

An exception is UIWebView and WKWebView that are always added to redact view.

Also, we're making a change to the ignoreRedact behaviour, that will not be inherited by its subviews.

Unit tests

## :pencil: Checklist

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -15,10 +15,12 @@
 
 - Resumes replay when the app becomes active (#4303)
 - Session replay redact view with transformation (#4308)
+- Don't redact clipped views (#4325)
```
brustolin, line 18:
```suggestion
```

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -44,23 +44,58 @@ class UIRedactBuilder {
...
-    init() {
         
-        var redactClasses = [ UILabel.self, UITextView.self, UITextField.self ] +
-        //this classes are used by SwiftUI to display images.
-        ["_TtCOCV7SwiftUI11DisplayList11ViewUpdater8Platform13CGDrawingView",
-         "_TtC7SwiftUIP33_A34643117F00277B93DEBAB70EC0697122_UIShapeHitTestingView",
-         "SwiftUI._UIGraphicsView", "SwiftUI.ImageLayer", "UIWebView"
-        ].compactMap { NSClassFromString($0) }
+    /**
+     Initializes a new instance of the redaction process with specified options.
```
philipphofmann, line 49: `l`:
```suggestion
     Initializes a new instance of the redaction process with the specified options.
```

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -44,23 +44,58 @@ class UIRedactBuilder {
...
+     This initializer configures which `UIView` subclasses should be redacted from screenshots and which should be ignored during the redaction process.
+
+     - parameter options: A `SentryRedactOptions` object that specifies the configuration for the redaction process.
+     
+     - If `options.redactAllText` is `true`, common text-related views such as `UILabel`, `UITextView`, and `UITextField` are redacted.
+     - If `options.redactAllImages` is `true`, common image-related views such as `UIImageView` and various internal `SwiftUI` image views are redacted.
+     - The `options.ignoreRedactViewTypes` allows specifying custom view types to be ignored during the redaction process.
+     - The `options.redactViewTypes` allows specifying additional custom view types to be redacted.
+
+     - note: On iOS, views such as `WKWebView` and `UIWebView` are automatically redacted, and controls like `UISlider` and `UISwitch` are ignored.
```
philipphofmann, line 60: `l`: Wouldn't it make sense to have that info at `SentryRedactOptions` and not here?

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -44,23 +44,58 @@ class UIRedactBuilder {
...
+        if options.redactAllText {
+            redactClasses += [ UILabel.self, UITextView.self, UITextField.self ]
+        }
+        
+        if options.redactAllImages {
+            //this classes are used by SwiftUI to display images.
+            redactClasses += ["_TtCOCV7SwiftUI11DisplayList11ViewUpdater8Platform13CGDrawingView",
+             "_TtC7SwiftUIP33_A34643117F00277B93DEBAB70EC0697122_UIShapeHitTestingView",
+             "SwiftUI._UIGraphicsView", "SwiftUI.ImageLayer"
+            ].compactMap(NSClassFromString(_:))
```
philipphofmann, line 74: `m`: I know this was already here before, but how did we come up with these class names? Doesn't this change from time to time? Can we add an explanation or a link so we can update it in the future?

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -44,23 +44,58 @@ class UIRedactBuilder {
...
+            
+            redactClasses.append(UIImageView.self)
+        }
         
 #if os(iOS)
         redactClasses += [ WKWebView.self ]
+
+        //If we try to use 'UIWebView.self' it will not compile for macCatalyst, but the class does exists.
```
philipphofmann, line 82: 💯 for the comment.

philipphofmann (approved):
LGTM, with a few suggestions 👍

Sources/Swift/Tools/UIRedactBuilder.swift
brustolin: Probably both.

Sources/Swift/Tools/UIRedactBuilder.swift
brustolin: They should not change from time to time for the same class name. So this will be updated only if Apple start using other classes

Issue #4334, opened 2024-09-16
Exclude SDK's working/cache/temp files from backup


kahest:
### Description

The SDK stores various files temporarily or persistently in the app container's file system. 

✅  Currently, the SDK [creates the working dir for most data in a cache dir](https://github.com/getsentry/sentry-cocoa/blob/24e07440d85b8d07ae4a98885d0f8f440ede339e/Sources/Sentry/SentryOptions.m#L193) which is not backed up.

:warning: However, some data is stored under the [application support dir](https://github.com/getsentry/sentry-cocoa/blob/24e07440d85b8d07ae4a98885d0f8f440ede339e/Sources/Sentry/SentryFileManager.m#L735) which is included in backups, see [Apple docs](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1).

To improve backup sizes and not expose data unnecessarily in potentially unencrypted backups, these files (ideally the working dir as a whole) should be excluded from backups as well.

This can be done in two (or possibly more) ways:
- move app support files into cache (or temp) dirs
  - :warning: these can get deleted/purged by the OS without warning, so data might get lost and additional safeguards to ensure working dirs and subdirs are present would be required
  - since we already use cache dir for some data and temp is more volatile, cache likely makes more sense
- set `NSURLIsExcludedFromBackupKey`:
  - see docs: https://developer.apple.com/documentation/foundation/nsurlisexcludedfrombackupkey 
  - this works for files and directories, recursively (see e.g. here on [SO](https://stackoverflow.com/questions/14376883/is-nsurlisexcludedfrombackupkey-recursive) - not explicitly documented by Apple, but implied because it can be set to generic file sytem resources)

PR #4335, opened 2024-09-16
fix: discontinue NSApplicationSupportDirectory usage

armcknight:
Update: we will no longer use NSApplicationSupportDirectory and just move the few usages of it to use NSCachesDirectory.

Original description:

We received a question about what information we store to disk. Almost everything is stored in NSCachesDirectory, which is not included in device backups: https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html

image

But, we started adding things to NSApplicationSupportDirectory, which is included in backups by default. Currently, no PII is written there, but this PR will mark the entire directory as excluded from backups as it's not necessary to back it up, and this will mitigate any accidental PII leaks in the future if we start writing more things there.

Fixes https://github.com/getsentry/sentry-cocoa/issues/4334


Sources/Sentry/SentryFileManager.m
```diff
@@ -746,6 +746,15 @@ - (void)createPathsWithOptions:(SentryOptions *)options
         }
         sentryApplicationSupportPath =
             [applicationSupportDirectory stringByAppendingPathComponent:@"io.sentry"];
+
+        // exclude our app support directory from backups
+        NSURL *url = [NSURL URLWithString:sentryApplicationSupportPath];
+        NSError *error;
+        if (![url setResourceValue:@YES forKey:NSURLIsExcludedFromBackupKey error:&error]) {
```
armcknight, line 753: This brings up the question of what to do if this line fails... maybe we should set `sentryApplicationSupportPath` to `nil` so that anything that wants to write there in the SDK after this cannot do so.

brustolin: I really doubt this will ever fail. 
If it fail its not our fault, we did what was asked of us.

philipphofmann: What about completely ditching `sentryApplicationSupportPath`? We only use it for `sentry_initializeAsyncLogFile`, which could also write to the cache I guess and for `enableAppLaunchProfiling`, which is still experimental. We only need to ensure that we clean up existing files in`sentryApplicationSupportPath`. As `enableAppLaunchProfiling` is still experimental we don't need to migrate it, IMO, and it's OK to delete the file for `sentry_initializeAsyncLogFile`, I guess @brustolin?

Then, we don't need to deal with this edge case. Also, the docs for [`NSURLIsExcludedFromBackupKey`](https://developer.apple.com/documentation/foundation/nsurlisexcludedfrombackupkey) sate:

> Use this property to exclude cache and other app support files that aren’t necessary in a backup. Set this property each time you save a file because some common file operations cause this property to reset to false.

It will be complicated to ensure to set this property for all files saved in `sentryApplicationSupportPath`. Even if we ensure to do that now, we can easily break this in the future. So I think it's better to ditch writing files to `sentryApplicationSupportPath`.

philipphofmann (commented):
I think it's better to ditch using `sentryApplicationSupportPath`. See my comment.

brustolin:
I also prefer to ditch sentryApplicationSupportPath all together.

Sources/Sentry/SentryFileManager.m
armcknight: > I really doubt this will ever fail.

Same, but if an API gives a way to report errors, you gotta handle the errors. Murphy's law.

> If it fail its not our fault, we did what was asked of us.

I don't see it this way. We're ultiimately responsible for our usage of any API we choose to use, even if we don't cover an edge case correctly.

Ugh, good catch with that doc @philipphofmann. I'm fine with changing this to NSCachesDirectory for now. 

As for how it can affect profiling, it's possible we lose the ability to profile anything other than a warm or prewarmed launch, because presumably a cold launch will only happen after something like an app update or device backup restore where caches may not already be available. I don't have a sense for how frequently NSCachesDirectory is wiped otherwise. We could maybe build telemetry for that, might even make a nice SDK feature. But I guess this is probably not used too much yet so probably wouldn't impact customers much, and we can always respond if they do ask for us to revert the behavior. I think ultimately we can get the NSApplicationSupportDirectory usage right if we wrap it correctly with SentryFileManager, should the need arise.

brustolin (approved):

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Don't redact clipped views (#4325)
 - Session replay for crash not created because of a race condition (#4314)
 - Double-quoted include, expected angle-bracketed instead (#4298)
+- Exclude the directory we use in NSApplicationSupportDirectory from device backups (#4335)
```
kahest, line 22: `m`: I think this needs to be updated now

Sources/Sentry/SentryFileManager.m
philipphofmann: > As for how it can affect profiling, it's possible we lose the ability to profile anything other than a warm or prewarmed launch, because presumably a cold launch will only happen after something like an app update or device backup restore where caches may not already be available.

I mean we can also keep it if required, but we need to ensure it's hard to shoot ourselves in the foot by correctly wrapping the usage. We shouldn't return the path but instead offer methods to read and write to `NSApplicationSupportDirectory` or something similar.

philipphofmann (approved):
Thank you @armcknight 💯

CHANGELOG.md
armcknight: Thanks!

PR #4336, opened 2024-09-17
fix: Safe guard strncpy usage

brustolin:
## :scroll: Description

We use strncpy in some cases, but this function does not guarantee a null-terminated string if the source is larger than the destination. 

To safeguard this, a wrapper function was created that guarantees null-termination.

Unit tests

## :pencil: Checklist


brustolin:
Follow-up discussion:

We have some code that uses `strncpy` to compose file paths. So far, we’ve used a large enough buffer that accommodates all the use cases we have.

Should we start addressing this? I don’t think we’ve encountered any issues related to file paths so far.

JoshuaMoelans (approved):
LGTM as well

PR #4340, opened 2024-09-17
chore: Formating code with clang-format 19

brustolin:
`clang-format` had updated and we need a PR to use its new format

_#skip-changelog_


philipphofmann (approved):
Thanks 🚀

PR #4343, opened 2024-09-18
ref: For envelopeWithData serialization

philipphofmann:
Small refactorings for SentrySerialization.envelopeWithData.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/4342.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentrySerialization.m
+++ b/Sources/Sentry/SentrySerialization.m
@@ -135,6 +135,7 @@ NS_ASSUME_NONNULL_BEGIN
         return nil;
     }
 
+    NSAssert(envelopeHeaderIndex > 0, @"EnvelopeHeader was parsed, its index is expected.");
     if (envelopeHeaderIndex == 0) {
         SENTRY_LOG_ERROR(@"EnvelopeHeader was parsed, its index is expected.");
         return nil;
--- a/Sources/Sentry/SentrySerialization.m
+++ b/Sources/Sentry/SentrySerialization.m
@@ -82,10 +82,9 @@ NS_ASSUME_NONNULL_BEGIN
 {
     SentryEnvelopeHeader *envelopeHeader = nil;
     const unsigned char *bytes = [data bytes];
-    NSUInteger endOfEnvelope = data.length - 1;
     int envelopeHeaderIndex = 0;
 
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentrySerialization.m
```diff
@@ -100,32 +101,33 @@ + (SentryEnvelope *_Nullable)envelopeWithData:(NSData *)data
...
-
-                envelopeHeader = [[SentryEnvelopeHeader alloc] initWithId:eventId
-                                                                  sdkInfo:sdkInfo
-                                                             traceContext:traceContext];
-
-                if (headerDictionary[@"sent_at"] != nil) {
-                    envelopeHeader.sentAt = sentry_fromIso8601String(headerDictionary[@"sent_at"]);
-                }
+                break;
+            }
```
philipphofmann, line 105: I added the break here to reduce the code indentation. No changes to the code itself.

Sources/Sentry/SentrySerialization.m
```diff
@@ -100,32 +101,33 @@ + (SentryEnvelope *_Nullable)envelopeWithData:(NSData *)data
...
+
+            envelopeHeader = [[SentryEnvelopeHeader alloc] initWithId:eventId
+                                                              sdkInfo:sdkInfo
+                                                         traceContext:traceContext];
+
+            if (headerDictionary[@"sent_at"] != nil) {
+                envelopeHeader.sentAt = sentry_fromIso8601String(headerDictionary[@"sent_at"]);
+            }
+
             break;
```
brustolin, line 131: l: What is this break at the end of the loop?

brustolin (approved):

Sources/Sentry/SentrySerialization.m
philipphofmann: It's not at the end of the loop. It's at the end of the if statement. It also confused me a bit when I looked at it the first time 😁

brustolin: But after the If the loops ends, what I'm missing here?

brustolin: Forget about it. I got it now. I was thinking about `continue`

PR #4348, opened 2024-09-19
test: Switch testCorruptEnvelope to a unit test

philipphofmann:
Change the testCorruptEnvelope to a unit test instead of using a UITest. Tapping the corruptEnvelope button after v 8.33.0 doesn't even create an envelope because the writing of the envelope is atomic. The app terminates without writing an envelope. As the UITest was flaky, we instead switched to a unit test validating the underlying problem of deserializing a corrupted envelope. Furthermore, we have to ensure that the corrupted envelopes don't crash when deserializing them because we have no guarantee that our users don't modify the envelope files.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Helper/SentrySerializationTests.swift
+++ b/Tests/SentryTests/Helper/SentrySerializationTests.swift
@@ -298,20 +298,6 @@ class SentrySerializationTests: XCTestCase {
         XCTAssertEqual(payloadAsString.data(using: .utf8), item.data)
     }
     
-    /// This is to prevent https://github.com/getsentry/sentry-cocoa/issues/4280
-    /// With 8.33.0, writing an envelope could fail in the middle of the process, which the envelope
-    /// payload below simulates. The JSON stems from writing an envelope to disk with vast data
-    /// that leads to an OOM termination on v 8.33.0.
-    /// Running this test on v 8.33.0 leads to a crash.
-    func testEnvelopeWithDataCorruptEnvelope() throws {
-        let itemData = """
-                       {"event_id":"1990b5bc31904b7395fd07feb72daf1c","sdk":{"name":"sentry.cocoa","version":"8.33.0"}}
-                       {"type":"test","length":50}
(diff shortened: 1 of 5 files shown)


Tests/SentryTests/Helper/SentrySerializationTests.swift
```diff
@@ -298,6 +298,20 @@ class SentrySerializationTests: XCTestCase {
...
+    /// that leads to an OOM termination on v 8.33.0.
+    /// Running this test on v 8.33.0 leads to a crash.
+    func testEnvelopeWithDataCorruptEnvelope() throws {
+        let itemData = """
+                       {"event_id":"1990b5bc31904b7395fd07feb72daf1c","sdk":{"name":"sentry.cocoa","version":"8.33.0"}}
+                       {"type":"test","length":50}
+                       """.data(using: .utf8)!
+        
+        XCTAssertNil(SentrySerialization.envelope(with: itemData))
+    }
```
brustolin, line 313: h: This is a duplicate of `testReturnNilForCorruptedEnvelope`. Does it make sense to have both?

philipphofmann: `testReturnNilForCorruptedEnvelope` is slightly different. The contents of the envelope contain only an event but it's truncated by 1 byte. This one here has no attachment although there should be one. I think it adds value.

brustolin: From a code flow perspective, it’s exactly the same. The code will attempt to read the next 50 bytes and will not succeed. However, having two tests doesn’t cause any harm other than extending CI time, so I’m fine with it.

brustolin (approved):

Tests/SentryTests/Helper/SentrySerializationTests.swift
philipphofmann: I moved it now to SentrySDKTests to start the SDK with a corrupted envelope. Thanks for pointing it out.

PR #4349, opened 2024-09-19
test: Add tests for envelopeWithData

philipphofmann:
Add more test cases for corrupted envelopes for
SentrySerialization.envelopeWithData.

#skip-changelog


brustolin (approved):
I tend to dislike these envelopes from a string because if the envelope format ever changes, we will have to update all the tests. The best approach would be to create data from an envelope object and then corrupt it.

But it works, so its up to you.

PR #4355, opened 2024-09-20
chore: Fix broken link

brustolin:
Fixing a broken link that explains a piece of code.

_#skip-changelog_

### DIFF
--- a/Sources/SentryCrash/Recording/Tools/SentryCrashSysCtl.c
+++ b/Sources/SentryCrash/Recording/Tools/SentryCrashSysCtl.c
@@ -233,7 +233,7 @@ bool
 sentrycrashsysctl_getMacAddress(const char *const name, char *const macAddressBuffer)
 {
     // Based off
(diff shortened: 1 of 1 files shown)


Sources/SentryCrash/Recording/Tools/SentryCrashSysCtl.c
```diff
@@ -233,7 +233,7 @@ bool
 sentrycrashsysctl_getMacAddress(const char *const name, char *const macAddressBuffer)
 {
     // Based off
-    // http://iphonedevelopertips.com/device/determine-mac-address.html
+    // http://web.archive.org/web/20111109013928/http://iphonedevelopertips.com/device/determine-mac-address.html
```
brustolin, line 236: The original link is broken now, but I found a version in the web archive. 
Also not sure if should keep this around.

philipphofmann (approved):
I think it's fine to keep it. At least we know then where it comes from.

PR #4356, opened 2024-09-20
ref: Unify ANRTrackerDelegate

philipphofmann:
Merge SentryANRTrackerDelegate and SentryANRTrackerDelegateV2 into one implementation to reduce complexity.

This is the first step in making swapping between the ANRTrackerV1 and ANRTrackerV2 easier. 

#skip-changelog


brustolin (approved):
Thanks!

Issue #4043, opened 2024-06-10
Add `breadrumb.origin`


krystofwoldrich:
### Description

Add new runtime property to `SentryBreadcrumb` called `origin`, this can be used to identify source of the breadcrumb.

For example hybrid SDKs can identify native breadcrumbs from JS or Flutter.

```[tasklist]
### Required for
- [ ] https://github.com/getsentry/sentry-capacitor/pull/629
- [ ] https://github.com/getsentry/sentry-react-native/issues/2146 
``` 

### Related issues
- https://github.com/getsentry/sentry-java/issues/3470

krystofwoldrich:
@lucas-zimerman When you open a PR , you can link this issue.

philipphofmann:
Please update the [develop docs](https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/) before closing this issue.

antonis:
Hey @philipphofmann 👋 
thank you for the heads up 🙇 
Since [the new field is private for internal use](https://github.com/getsentry/sentry-cocoa/pull/4358#discussion_r1773227798) my understanding is we don't need to add it in the [developer docs](https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/). Let me know if this is not the case 🙏

philipphofmann:
@antonis, it's still part of the protocol. The developer docs are for SDK maintainers. You don't have to do that, but either @krystofwoldrich, who opened the issue, or one of the Sentry folks should do it.

antonis:
Makes sense. Thank you for the feedback Philipp 🙇 
I'm reopening the issue till this is tackled.

PR #4358, opened 2024-09-20
feat: Adds breadcrumb.origin field

antonis:
## :scroll: Description

Adds a new `SentryBreadcrumb` optional private string property `origin`

Fixes https://github.com/getsentry/sentry-cocoa/issues/4043
The discussion in https://github.com/getsentry/sentry-capacitor/pull/629#discussion_r1617321563 adds some context on the need of a new field.

Unit tests

## :pencil: Checklist


Sources/Sentry/Public/SentryBreadcrumb.h
```diff
@@ -34,6 +34,12 @@ NS_SWIFT_NAME(Breadcrumb)
  */
 @property (nonatomic, copy) NSString *_Nullable type;
 
+/**
+ * Origin of the breadcrumb that is used to identify source of the breadcrumb
+ * For example hybrid SDKs can identify native breadcrumbs from JS or Flutter
+ */
+@property (nonatomic, copy) NSString *_Nullable origin;
```
brustolin, line 41: l: Wondering if we should make this public, or just for internal use?
Do we expect the user to also add "App" as origin of custom breadcrumbs?

brustolin: l: This is a nitpick, but its prettier to use like this. We can update the rest of `_Nullable` usage in this old file

```suggestion
/**
 * Origin of the breadcrumb that is used to identify source of the breadcrumb
 * For example hybrid SDKs can identify native breadcrumbs from JS or Flutter
 */
@property (nonatomic, copy, nullable) NSString * origin;
```

brustolin (commented):
First pass, it looks good.

Just a few comments.

Sources/Sentry/Public/SentryBreadcrumb.h
antonis: > l: This is a nitpick, but its prettier to use like this. We can update the rest of _Nullable usage in this old file

Great suggestion @brustolin 👍 I've updated the code with 77b6372e7186c762c0418246d16536c3d576168d and c2f49e080b3e1a4470540dd5e79b8ab9df8cb40d

antonis: > l: Wondering if we should make this public, or just for internal use?
Do we expect the user to also add "App" as origin of custom breadcrumbs?

Good point @brustolin. My understanding from https://github.com/getsentry/sentry-cocoa/issues/4043 and the previous discussion is that this [will only be used internally to solve a technical problem for now](https://github.com/getsentry/sentry-capacitor/pull/629#discussion_r1617377778).

antonis (commented):
Thank you for reviewing @brustolin and the feedback 🙇 
I applied your suggestions and it is ready for another pass.

Note that the `Test / Unit iOS - Xcode 15.4 - OS 17.2` and `Benchmarking / Collect app metrics` CI checks are failing for me. I think they are flaky and I'll try to rerun. Let me know if you have more context since I'm new to the repo 🙏

Sources/Sentry/Public/SentryBreadcrumb.h
brustolin: > Good point @brustolin. My understanding from https://github.com/getsentry/sentry-cocoa/issues/4043 and the previous discussion is that this https://github.com/getsentry/sentry-capacitor/pull/629#discussion_r1617377778.

I think this would be the best approach right now, in that case we need to make this property private. 
You can take a look at `SentryOptions+Private.h` to see how we do this.

Sources/Sentry/include/HybridPublic/SentryBreadcrumb+Private.h
```diff
@@ -2,10 +2,16 @@
...
+ */
+@property (nonatomic, copy, nullable) NSString *origin;
+
 /**
  * Initializes a SentryBreadcrumb from a JSON object.
  * @param dictionary The dictionary containing breadcrumb data.
  * @return The SentryBreadcrumb.
  */
-- (instancetype)initWithDictionary:(NSDictionary *)dictionary;
+- (instancetype _Nonnull)initWithDictionary:(NSDictionary *_Nonnull)dictionary;
```
antonis, line 16: Added this to handle a `missing a nullability type specifier` compile issue.

Sources/Sentry/Public/SentryBreadcrumb.h
antonis: Great suggestion @brustolin 🙇 
I've updated the code with 5da364f4fa26229299a64950e0cf637fd5fea464

brustolin (approved):
Thanks for this. Looks good!

antonis:
Thank you for the review and the feedback 🙇 
Given that the `Benchmarking / Collect app metrics` is also failing on `main` at this point I'll proceed with merging.

krystofwoldrich:
Thank you for adding the field. @antonis

PR #4360, opened 2024-09-23
ref: Add ANRTracker protocol

philipphofmann:
Add a protocol for the ANRTracker in preparation for switching between V1 and V2. Furthermore, remove the ANRTrackerV2 integration as it would contain mostly duplicated code from ANTTrackerV1.

#skip-changelog


philipphofmann:
SwiftLint fix is here https://github.com/getsentry/sentry-cocoa/pull/4361.

Sources/Sentry/include/HybridPublic/SentryDependencyContainer.h
```diff
@@ -90,7 +90,7 @@ SENTRY_NO_INIT
 @property (nonatomic, strong) SentryReachability *reachability;
 #endif // !TARGET_OS_WATCH
 
-- (SentryANRTracker *)getANRTracker:(NSTimeInterval)timeout;
+- (SentryANRTrackerV1 *)getANRTrackerV1:(NSTimeInterval)timeout;
 #if SENTRY_UIKIT_AVAILABLE
 - (SentryANRTrackerV2 *)getANRTrackerV2:(NSTimeInterval)timeout;
```
brustolin, line 95: l: We could make this functions also return the protocol.
So we can remove the `@class` reference from this file, and the `#include "SentryANRTrackerV1.h"` (and v2) from other files like `SentryANRTrackingIntegration.h`

brustolin (approved):
LGTM

Sources/Sentry/include/HybridPublic/SentryDependencyContainer.h
philipphofmann: I can do this in the next PR. I'm still unsure if the dependency container should be the place to decide which version we use or if it should stay dump. I will cover this in the next PR.

Issue #4270, opened 2024-08-09
User Feedback UI widget


armcknight:
### Description

Build the interface widget an end-user would tap to present the feedback capture form

brustolin:
IMO, a UI widget is overkill. I believe a public API to open the feedback form is enough. The user can be responsible for creating the button, link, or action that opens the form, as each app has its own visual identity.

philipphofmann:
@brustolin, I think that's the main point of this new feature. If I'm not mistaken, people on the web love it, but I'm also unsure if folks on native mobile apps would use it. @bruno-garcia or @armcknight, maybe you can elaborate on why it's worth the effort to build a UI widget.

armcknight:
There will be a way for customers to BYOB (bring your own button) and hook it into the integration. They would be completely responsible for displaying that button, we just use the touch event to display our form.

I hadn't yet considered an option where we take a button from them, but manage displaying it. That could be another nice option, so they get their branding, but don't have to worry about the mechanics.

In any case, it's not a requirement to use the widget. It's another option, there may be people out there that feel differently than you about this. It doesn't make sense to withhold the option from them just because someone here doesn't like it. Ideally we can attach analytics to it and we can see how many people use each method in practice.

philipphofmann:
@armcknight, yes, that makes sense. Thanks for explaining 😀.

armcknight:
Current state of the widget:

using an all-default config (except positioning, which I used to nudge it out of the way of the tab bar):
![Image](https://github.com/user-attachments/assets/07c53108-27a1-400b-bd37-f7fb78f89dc5)

And with some [extreme] overrides:![Image](https://github.com/user-attachments/assets/489bda53-d098-49a8-8934-e5a4c672b8b5)

So there's a little more work to do to correctly size/align the megaphone with edge case-y font sizes (which will be a concern with accessibility settings).

armcknight:
Working to get the icon and text perfectly vertically centered in the button, but it's tough because different font faces have different amounts of space inside the text bounding box, so i need to find the right combination of constraints that work generally across them all. See for instance `ChalkBoardSE-Regular`: ![Image](https://github.com/user-attachments/assets/65692fc8-580b-4d6e-92b9-0637f0106e8f)
vs `HelveticaNeue`:![Image](https://github.com/user-attachments/assets/67299339-c8b1-4fb8-8ae6-3baa438f81a9)

armcknight:
[...]

PR #4364, opened 2024-09-24
feat(user feedback): display a button in a window that presents a vc

armcknight:
#skip-changelog; for #4270

Deliver a complete styled button like we have in the javascript SDK docs that opens a modal view where the form will go in a subsequent PR.
image

Check the linked issue for updates on what it looks like with various config settings and device settings like for accessibility and dark mode.


Sources/Sentry/Public/SentryOptions.h
```diff
@@ -164,7 +164,7 @@ NS_SWIFT_NAME(Options)
 /**
  * Array of integrations to install.
  */
-@property (nullable, nonatomic, copy) NSArray<NSString *> *integrations;
+@property (nullable, nonatomic, readonly) NSArray<NSString *> *integrations;
```
brustolin, line 167: h: Why did we changed this in here?
Integrations can be manipulated by users to remove something they don’t like.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
```diff
@@ -7,19 +7,22 @@ import UIKit
  * An integration managing a workflow for end users to report feedback via Sentry.
  * - note: The default method to show the feedback form is via a floating widget placed in the bottom trailing corner of the screen. See the configuration classes for alternative options.
  */
-class SentryUserFeedbackIntegration {
+@objcMembers public class SentryUserFeedbackIntegrationDriver: NSObject {
```
brustolin, line 10: h: By making it public, this will be exposed to the user. Is that the intention? If so, how should the user interact with it directly?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegration.m
```diff
@@ -0,0 +1,14 @@
+#import "SentryUserFeedbackIntegration.h"
+#import "SentryOptions+Private.h"
+#import "SentrySwift.h"
+
+@implementation SentryUserFeedbackIntegration {
```
brustolin, line 5: l: Do you plan to add more features to this class?
Why not use `SentryUserFeedbackIntegrationDriver` as the integration like before?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
```diff
@@ -0,0 +1,92 @@
...
+            fatalError("init(coder:) has not been implemented")
+        }
+        
+        override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+            guard let result = super.hitTest(point, with: event) else {
+                return nil
+            }
+            guard result.isKind(of: UIButton.self) else {
+                return nil
+            }
```
brustolin, line 86: l: ```suggestion
            guard let result = super.hitTest(point, with: event), !result.isKind(of: UIButton.self) else {
                return nil
            }
```

**Comment by brustolin** [comment]
I think we should detect TabBar to arrange the button according.

Simulator Screenshot - iPhone 16 Pro - 2024-09-24 at 09 15 13

**Comment by brustolin** [comment]
h: We can debate it later, but currently we're organising Swift code under Swift folder, so

|This|Should be here|
|-|-|
|Screenshot 2024-09-24 at 09 28 55|Screenshot 2024-09-24 at 09 29 06|

File: Sources/Sentry/Public/SentryOptions.h (hunk 1 above)
Comment (armcknight) [reply Sources/Sentry/Public/SentryOptions.h:167]: I wanted to be able to mutate the list in the implementation.

I misunderstood `copy`, thinking it _hands out copies on access_, but it actually _makes copies on assignment_... that's ok, because I should be able to make these readwrite/copy and it should still work for my purposes.

File: Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegration.m (hunk 3 above)
Comment (armcknight) [reply Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegration.m:5]: Originally I had it like that, but it wasn't compiling with a swift integration class, and we don't have any others like that. I essentially tried to copy what you have for replays.

File: Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift (hunk 4 above)
Comment (armcknight) [reply Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift:86]: Yes, I'll have to add a boolean property there to allow hits to the form while it's displayed. That'll come in a separate PR for implementing the form (and I have implementations of things I've done this for in the past that I can reference)

Comment (armcknight) [reply Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift:86]: I usually prefer leaving separate conditions in separate guards instead of using comma-delimited lists, helps debugging and logging later if need be

File: Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift [hunk 5]
```diff
@@ -0,0 +1,92 @@
...
+                label.text = "Hi, I'm a user feedback form!"
+                formDialog.view.addSubview(label)
+                label.translatesAutoresizingMaskIntoConstraints = false
+                label.textAlignment = .center
+                NSLayoutConstraint.activate([
+                    label.leadingAnchor.constraint(equalTo: formDialog.view.leadingAnchor),
+                    label.trailingAnchor.constraint(equalTo: formDialog.view.trailingAnchor),
+                    label.centerYAnchor.constraint(equalTo: formDialog.view.centerYAnchor),
+                ])
+                present(formDialog, animated: true)
```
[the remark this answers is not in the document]
armcknight: Yep I added the animations idea last minute and haven't completed it yet.

armcknight:
> I think we should detect TabBar to arrange the button according.
> 
> Simulator Screenshot - iPhone 16 Pro - 2024-09-24 at 09 15 13

This should be on the consumer to use the `layoutUIOffset` and later, enabling draggability of the button will obviate this. I don't think trying to predict all the possible combinations of view hierarchy, or spelunking through them, is wise. That being said I'm not sure which sample app or device you're testing, because I thought I already fixed this with the sample apps' usage of said `layoutUIOffset` property 🤔

Sources/Sentry/Public/SentryOptions.h
armcknight: Hm, sadly I cannot make it a private readwrite NSMutableArray and public copy nonmutable NSArray

armcknight: I found another way to do it

brustolin:
> That being said I'm not sure which sample app or device you're testing, because I thought I already fixed this with the sample apps' usage of said layoutUIOffset property 🤔

I just downloaded this branch.

armcknight:
> > That being said I'm not sure which sample app or device you're testing, because I thought I already fixed this with the sample apps' usage of said layoutUIOffset property 🤔
> 
> I just downloaded this branch.

> which sample app or device

brustolin:
> which sample app or device

iOS-Swift sample
iPhone 16 Pro

armcknight:
The diff is currently a little over 500 lines added, note that half of that is just the UIBezier drawing the megaphone that I got from our SVG design asset :D

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationShell.h
```diff
@@ -0,0 +1,9 @@
+#import "SentryBaseIntegration.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentryUserFeedbackIntegrationShell : SentryBaseIntegration
```
armcknight, line 5: So, in the previous PR, I made `SentryUserFeedbackIntegration` a swift file, before I knew that it isn't currently possible to implement integrations in Swift (AFAICT anyway). But, renaming that file/class so that this objc class could have that name made the diff of that file messy, so for now I'm making this a placeholder name, and I'll do the renames in a separate PR.

brustolin:
For default values, we should respect iOS assessibility values:

Simulator Screenshot - iPhone 16 Pro - 2024-10-16 at 08 42 45

Simulator Screenshot - iPhone 16 Pro - 2024-10-16 at 08 44 10

This:
image 

Should be:
image

In this configuration

brustolin (changes requested):
[This](https://github.com/getsentry/sentry-cocoa/pull/4364#issuecomment-2370418184) need to be addressed before merging the PR.

brustolin (commented):
I didn't review the code yet, just how to use it and the button itself. 

A couple of extra things:
- Need right-to-left support
- I think configureTheme, should be inside configureWidget, or at least the properties that configure the widget, it took me an extra second to find this out, when it would be much easier to set the widget theme inside the configureWidget callback

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidgetButtonMegaphoneIconView.swift
```diff
@@ -0,0 +1,174 @@
...
+            if UIScreen.main.traitCollection.userInterfaceStyle == .dark {
+                svgLayer.fillColor = config.darkTheme.foreground.cgColor
+            } else {
+                configureLightTheme()
+            }
+        } else {
+            configureLightTheme()
+        }
+        
+        svgLayer.fillRule = .evenOdd
```
brustolin, line 23: l: This is just a suggestion to use alongside the megaphoneShape suggestion.
Feel free to ignore it 

```suggestion
        svgLayer.path = megaphoneShape
        svgLayer.lineWidth = 1.5
        svgLayer.lineJoin = .round
        svgLayer.fillColor = UIColor.clear.cgColor
        
        func configureLightTheme() {
            svgLayer.strokeColor = config.theme.foreground.cgColor
        }
        if #available(iOS 12.0, *) {
            if UIScreen.main.traitCollection.userInterfaceStyle == .dark {
                svgLayer.fillColor = config.darkTheme.foreground.cgColor
            } else {
                configureLightTheme()
            }
        } else {
            configureLightTheme()
        }
        
```

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidgetButtonMegaphoneIconView.swift
```diff
@@ -0,0 +1,174 @@
...
+            layer.transform = CATransform3DMakeScale(config.scaleFactor, config.scaleFactor, 0)
+        }
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    //swiftlint:disable function_body_length
+    lazy var megaphoneShape: CGPath = {
```
brustolin, line 38: l: I love the idea of drawing like this.

You can achieve a very similar result with the following code:

```Swift
lazy var megaphoneShape: CGPath = {
        let path = CGMutablePath()
        
        path.move(to: CGPoint(x: 1, y: 3))
        path.addLine(to: CGPoint(x: 7, y: 3))
        path.addLine(to: CGPoint(x: 10, y: 1))
        path.addLine(to: CGPoint(x: 12, y: 1))
        path.addLine(to: CGPoint(x: 12, y: 11))
        path.addLine(to: CGPoint(x: 10, y: 11))
        path.addLine(to: CGPoint(x: 7, y: 9))
        path.addLine(to: CGPoint(x: 1, y: 9))
        path.closeSubpath()
        
        path.addRect(CGRect(x: 2, y: 9, width: 3.5, height: 6))
        
        path.move(to: CGPoint(x: 12, y: 6))
        path.addRelativeArc(center: CGPoint(x: 12, y: 6), radius: 3, startAngle: -(.pi/2), delta: .pi)
        
        return path
    }
```

brustolin (commented):
Just a suggestion for the megaphone drawing

armcknight:
@brustolin accessibility sizes and right-to-left languages are known tasks: https://github.com/getsentry/sentry-cocoa/issues/4270#issuecomment-2408210459

I already have right-to-left support going after i linked the PR last night, see the latest update in the issue: https://github.com/getsentry/sentry-cocoa/issues/4270#issuecomment-2417077633 with the implementation work in https://github.com/getsentry/sentry-cocoa/pull/4364/commits/abc51191c823c86b6cdf09317f14746e728b3e9d

armcknight:
> I think configureTheme, should be inside configureWidget, or at least the properties that configure the widget, it took me an extra second to find this out, when it would be much easier to set the widget theme inside the configureWidget callback

This can be debated separately from this PR if people want to. I'll just note here that theme options will also be needed in the form, so I tried creating a composition hierarchy that indicates they are used by more than just the widget (button):
Screenshot 2024-10-16 at 11 07 56 AM

Since the widget and form will use a lot of the same options I don't think it makes sense to make separate theme objects for each to own separately.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidgetButtonMegaphoneIconView.swift
armcknight: I will let @Jesse-Box or @vuluongj20 comment on this.

armcknight:
> [This](https://github.com/getsentry/sentry-cocoa/pull/4364#issuecomment-2370418184) need to be addressed before merging the PR.

@brustolin Would you mind if we do it in a separate PR along with https://github.com/getsentry/sentry-cocoa/pull/4364#discussion_r1787018390? I just don't want to mess the diff up here. Nothing is being shipped yet so there's no danger of a file not being delivered where it should be.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidgetButtonMegaphoneIconView.swift
armcknight: I think i'm just gonna go with the simpler version and if anyone doesn't like it, we can always revert. Thanks @brustolin

brustolin:
@brustolin yes, then plan is to merge to main for each PR. The option is currently in SentryOptions+Private so they should not be able to get a widget showing in their app.

Ok, good enough for me. Thanks!

Sources/Sentry/SentryOptions.m
```diff
@@ -39,35 +39,33 @@ @implementation SentryOptions {
     BOOL _enableTracingManual;
 }
 
-+ (NSArray<NSString *> *)defaultIntegrations
+NSMutableArray<NSString *> *
+sentry_defaultIntegrations(void)
```
brustolin, line 43: l: Why was this changed?
Just curious.

brustolin (approved):
I have my reservations about the composition of theme configuration, but this is not a blocker. 

LGTM

Sources/Sentry/SentryOptions.m
armcknight: I changed it to return a mutable array so that I can conditionally add the user feedback integration before installing all integrations, if the options property to configure its options is set (which implicitly enables the feature)

PR #4366, opened 2024-09-24
impr: Speed up SentryNetworkTracker

philipphofmann:
## :scroll: Description

Speed up SentryNetworkTracker for multiple HTTP requests in parallel by removing unnecessary locks.

@brustolin, you added these locks in the past. Can you recall if there was a good reason for that, except a race condition? I just want to double-check, that we don't break anything.

This surfaced while investigating a potential slow down for a specific UIViewController when enabling the Sentry Cocoa SDK.

Running unit performance tests locally. Sadly, they don't work in CI, so adding them to the repository adds no value.

## :pencil: Checklist

### DIFF
--- a/Sources/Resources/ThreadSanitizer.sup
+++ b/Sources/Resources/ThreadSanitizer.sup
@@ -3,7 +3,6 @@
 
 # Races to fix
 race:returnResponse
-race:enableNetworkTracking
 race:enableNetworkBreadcrumbs
 race:disable
 race:URLSessionDataTaskMock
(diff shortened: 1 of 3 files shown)


Sources/Resources/ThreadSanitizer.sup
```diff
@@ -3,7 +3,6 @@
 
 # Races to fix
 race:returnResponse
-race:enableNetworkTracking
```
philipphofmann, line 6: I replaced this with `SENTRY_DISABLE_THREAD_SANITIZER`.

brustolin (approved):

Issue #4359, opened 2024-09-20
Fix beta release bundle info


brustolin:
### Description

It's not possible to release an app that links to a Beta release.

Here is the reason why:
![Image](https://github.com/user-attachments/assets/3271920c-0ee1-49f6-ba58-8b7f78266261)

philipphofmann:
How did you fix the issue so you could release an app version?

brustolin:
> How did you fix the issue so you could release an app version?

By using the static framework directly without embedding it.

PR #4368, opened 2024-09-25
Chore: Fix the versioning to support app release with Beta versions

brustolin:
## :scroll: Description

Xcode does not allow an app to be publish with a framework that does not follow the versioning standard.

For more information see #4359

_#skip-changelog_


Tests/SentryTests/SentryTests.m
```diff
@@ -30,21 +30,6 @@ - (void)setUp
...
-    if ([info[@"CFBundleIdentifier"] isEqualToString:@"io.sentry.Sentry"]) {
-        // This test is running on a bundle that is not the SDK
-        // (code was loaded inside an app for example)
-        // in this case, we don't care about asserting our hard coded value matches
-        // since this will be the app version instead of our SDK version.
-        XCTAssert([version isEqualToString:SentryMeta.versionString],
-            @"Version of bundle:%@ not equal to version of SentryMeta:%@", version,
-            SentryMeta.versionString);
-    }
-}
```
brustolin, line 46: This test will fail for beta versions, because the `CFBundleShortVersionString` will not have the beta information.
Technically this test Is really not necessary since we dont use `CFBundleShortVersionString` from the SDK bundle for anything, since the SDK version comes from SentryMeta.

philipphofmann: Thanks for the comment.

philipphofmann (approved):
LGTM

philipphofmann:
Doesn't this deserve an entry in the changelog?

PR #4369, opened 2024-09-25
Chore: Update size difference benchmark

brustolin:
## :scroll: Description

CI is failing because the SDK size has increased.

_#skip-changelog_

### DIFF
--- a/Tests/Perf/metrics-test.yml
+++ b/Tests/Perf/metrics-test.yml
@@ -11,4 +11,4 @@ startupTimeTest:
 
 binarySizeTest:
   diffMin: 200 KiB
-  diffMax: 900 KiB
+  diffMax: 800 KiB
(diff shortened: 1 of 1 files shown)


Tests/Perf/metrics-test.yml
```diff
@@ -11,4 +11,4 @@ startupTimeTest:
 
 binarySizeTest:
   diffMin: 200 KiB
-  diffMax: 700 KiB
+  diffMax: 900 KiB
```
philipphofmann, line 14: I would go for 800 so we get reminded again to do something against it.

philipphofmann (approved):
LGTM

Tests/Perf/metrics-test.yml
brustolin:
```suggestion
  diffMax: 800 KiB
```

PR #4370, opened 2024-09-25
impr: Slightly speed up SentryInAppLogic

philipphofmann:
## :scroll: Description

Slightly speed up the SentryInAppLogic by avoiding calls to lastPathComponent and lowercaseString in loops. Both properties require some extra work, such as iterating over the contents of the string. This is noticeable for larger apps with many images and inAppIncludes and inAppExcludes. For small apps, this doesn't make a difference.

This came up while investigating the Sentry Cocoa SDK potentially slowing down a specific UIViewController of a customer.

Unit tests still pass, and by running a performance test locally, which doesn't work in CI, so I'm not adding it.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryInAppLogic.m
+++ b/Sources/Sentry/SentryInAppLogic.m
@@ -66,7 +66,7 @@ NS_ASSUME_NONNULL_BEGIN
 {
     return [SentryInAppLogic
         isImageNameLastPathComponentInApp:imageName.lastPathComponent.lowercaseString
-                             inAppInclude:inAppInclude];
+                             inAppInclude:inAppInclude.lowercaseString];
 }
 
 + (BOOL)isImageNameLastPathComponentInApp:(NSString *)imageNameLastPathComponent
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentryInAppLogic.m
```diff
@@ -48,7 +64,15 @@ - (BOOL)isClassInApp:(Class)targetClass
...
-    return [imageName.lastPathComponent.lowercaseString hasPrefix:inAppInclude.lowercaseString];
+    return [SentryInAppLogic
+        isImageNameLastPathComponentInApp:imageName.lastPathComponent.lowercaseString
+                             inAppInclude:inAppInclude];
+}
+
++ (BOOL)isImageNameLastPathComponentInApp:(NSString *)imageNameLastPathComponent
+                             inAppInclude:(NSString *)inAppInclude
+{
+    return [imageNameLastPathComponent hasPrefix:inAppInclude];
```
brustolin, line 75: l: Having this function split seems like a bit of overkill.

brustolin (approved):

Sources/Sentry/SentryInAppLogic.m
philipphofmann: I fully agree, but the other one is public.

PR #4374, opened 2024-09-26
chore: add first pre-commit hook to check tool versions

armcknight:
Check the installed version of clang-format and swiftlint and fail fast from the pre-commit run if they don't match what's in the homebrew lockfile. This approach could be extended to e.g. Gemfile.lock as well if desired.

We previously wanted to go the other way on this in https://github.com/getsentry/sentry-cocoa/pull/3739. But I still ran into local issues with mismatched lint/format rules dirtying unmodified files, so that didn't fix the root issue. I think this is the better solution than just not tracking the lockfiles.

#skip-changelog


Makefile
```diff
@@ -5,12 +5,25 @@ init: setup-git
...
+
+BREW_CLANG_FORMAT_VERSION=$(jq '.entries.brew."clang-format".version' Brewfile.lock.json | sed s/'"'//g)
+BREW_SWIFTLINT_VERSION=$(jq '.entries.brew.swiftlint.version' Brewfile.lock.json | sed s/'"'//g)
+LOCAL_CLANG_FORMAT_VERSION=$(clang-format --version | awk '{print $$3}')
+LOCAL_SWIFTLINT_VERSION=$(swiftlint version)
+.PHONY: check-versions
+check-versions:
+	@if [ "$(LOCAL_CLANG_FORMAT_VERSION)" != "$(BREW_CLANG_FORMAT_VERSION)" ]; then \
+		echo "clang-format version mismatch!"; \
+		echo "Expected: $(BREW_CLANG_FORMAT_VERSION), but found: $(LOCAL_CLANG_FORMAT_VERSION)"; \
```
philipphofmann, line 18: `m`: I think it would be great to explain here how to fix this. Especially contributors might be confused.

philipphofmann (approved):
Thanks @armcknight 💯

philipphofmann (approved):
Thanks for the resolution message.

PR #4377, opened 2024-09-30
impr: Capture trace based profiles on a BG Thread

philipphofmann:
## :scroll: Description

Serializing and capturing trace based profiles can run on the main thread and block it for a couple of milliseconds. Therefore, we move this to a background thread to avoid potentially blocking the main thread.

This came up while investigating profiles for potentially slowing down UIViewControllers.

Unit test and checking if the profile is still correctly attached to UIViewController transactions with our iOS-Swift sample app.

## :pencil: Checklist


Sources/Sentry/SentryTracer.m
```diff
@@ -631,19 +631,31 @@ - (void)finishInternal
 - (void)captureTransactionWithProfile:(SentryTransaction *)transaction
                        startTimestamp:(NSDate *)startTimestamp
 {
-    SentryEnvelopeItem *profileEnvelopeItem
-        = sentry_traceProfileEnvelopeItem(transaction, startTimestamp);
+    // This code can run on the main thread, and the profile serialization can take a couple of
+    // milliseconds. Therefore, we move this to a background thread to avoid potentially blocking
+    // the main thread.
+    __weak SentryTracer *weakSelf = self;
```
brustolin, line 637: h: I don’t think this is desirable, as it makes it possible (and actually easy) to lose the profiling. I believe it’s okay to hold a reference to self until the capture finishes, or at least hold a reference to the hub, which is the only thing we need to use inside the block.

brustolin (commented):
Thanks for this. There is one thing to discuss.

Sources/Sentry/SentryTracer.m
philipphofmann: That makes sense. Good catch thanks 👍

brustolin (approved):
LGTM!

philipphofmann:
I'm still waiting for @armcknight to approve this so we can avoid shooting ourselves in the foot.

armcknight (commented):
Could you be more specific about how you tested this? I just tried looking at a manual transaction with trace-based profiling on main vs this branch and the profiles look different. I can't get a launch profile to show up when running from this branch.

I tried comparing the differences between main and this branch with continuous profiling but I'm not even seeing what I expect on main so i need to ask the team what's going on there.

At a minimum for anything profiling right now you have to check:
- trace-based
    - regular transaction
    - launch
- continuous
    - regular transaction
    - launch

Trace-based regular txn on main (https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/cdc0737ef92245a0b14dcdb40c5d68c3/flamegraph/?colorCoding=by%20system%20vs%20application%20frame&fov=0%2C0%2C3822472960%2C19&query=&sorting=call%20order&tid=259&view=top%20down):

image

and on this branch (https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/28b4f9cc46ea4c2aa94cdfc400d11497/flamegraph/?colorCoding=by%20system%20vs%20application%20frame&fov=0%2C0%2C5127519744%2C19&query=&sorting=call%20order&tid=259&view=top%20down):

image

armcknight:
Also could you please share the results of the investigation that concluded this change was needed?

philipphofmann:
Hmm, weird, I didn't test it in great detail as I didn't expect that when all the tests are green, such a small change can have such a big impact. @armcknight, do you have a clue about what could cause this? I will look into this tomorrow.

armcknight:
@philipphofmann my guess is that you aren't dispatching the right call to the bg queue, and so it captures a set of profiling data much earlier than it marks the end of the transaction. You may need to look for a more granular point in that process, or create one by refactoring.

philipphofmann:
I found the problem. This PR only impacts trace based profiles, as only the SentryTracer uses `sentry_traceProfileEnvelopeItem`. Continuous profiling doesn't use this method.

I tested this for manual transactions, UIViewController transactions and app start transactions.







| | Main Branch | This feature Branch |
|--------|--------|--------|
| Manual Transaction | CleanShot 2024-10-02 at 09 54 05@2x  [Profile in Sentry](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/a2125fe0a8674f9ab0adb190d8a0501f/flamegraph/?colorCoding=by%20library&fov=0%2C2%2C1552376960%2C15&query=&sorting=call%20order&tid=259&view=top%20down) | CleanShot 2024-10-02 at 09 55 03@2x [Profile in Sentry](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/6c31f49a098a4fa6a62330842bde66b1/flamegraph/?colorCoding=by%20library&fov=0%2C0%2C1565427072%2C15&query=&sorting=call%20order&tid=259&view=top%20down) |
| TransactionsViewController | CleanShot 2024-10-02 at 09 56 06@2x [Profile in Sentry](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/44d87f70946445adb4a25d546be80744/flamegraph/?colorCoding=by%20library&fov=0%2C0%2C154258016%2C10&query=&sorting=call%20order&tid=259&view=top%20down) | CleanShot 2024-10-02 at 09 56 32@2x [Profile in Sentry](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/de85758a7a3e471880dc32f685730fa8/flamegraph/?colorCoding=by%20library&query=&sorting=call%20order&tid=259&view=top%20down) |
| App Start Transaction |  CleanShot 2024-10-02 at 09 58 00@2x [Profile in Sentry](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/60bec6f2dc2f446d831c817817787dee/flamegraph/?colorCoding=by%20library&fov=512648928%2C0%2C190049904%2C13&query=&sorting=call%20order&tid=259&view=top%20down) | CleanShot 2024-10-02 at 09 58 56@2x [Profile in Sentry](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/ff2ea6ebf7824f60b6e19997d8473b23/flamegraph/?colorCoding=by%20library&fov=375289216%2C0%2C455652480%2C13&query=&sorting=call%20order&tid=259&view=top%20down) |

Sources/Sentry/Profiling/SentryCaptureTransactionWithProfile.mm
```diff
@@ -0,0 +1,50 @@
...
+sentry_captureTransactionWithProfile(SentryHub *hub, SentryDispatchQueueWrapper *dispatchQueue,
+    SentryTransaction *transaction, NSDate *startTimestamp)
+{
+    const auto profiler = sentry_profilerForFinishedTracer(transaction.trace.internalID);
+    if (!profiler) {
+        [hub captureTransaction:transaction withScope:hub.scope];
+        return;
+    }
+
+    const auto profilingData = [profiler.state copyProfilingData];
```
philipphofmann, line 28: @armcknight, when I move this into the async block the profiles the SDK shortens the profiles uploaded to Sentry, but I can't replicate this behavior in the `testCaptureTransactionWithProfile_SetsCorrectAmountOfSamples`. I would like to have a failing test for this to ensure we don't break this by mistake. Maybe the tests with the mocks behave differently than the actual profiling code, or I just don't get it.

armcknight (approved):
Really nice test report! ✨ Looking at the actual profile durations in them, it seems the same, and it's the transaction that is actually more variable. I think that variability is not due to this change but is coming from preexisting behavior in the SDK so i'm not worried about it.

A comment on the tests you added, but otherwise things look good here, so when you get that ironed out, feel free to ship it.

Also regarding my earlier comment on continuous profiling, it looks like there's some delay with data coming in, being aggregated, or becoming accessible. 

If and when you get around to bg'ing the continuous serialization function (maybe in a separate PR?), test it by looking at https://sentry-sdks.sentry.io/profiling/?project=5428557&statsPeriod=1m&tab=flamegraph:

image

Tests/SentryProfilerTests/SentryTraceProfilerTests.swift
```diff
@@ -31,6 +31,48 @@ class SentryTraceProfilerTests: XCTestCase {
...
+        
+        self.fixture.dispatchQueueWrapper.dispatchAsyncExecutesBlock = true
+        
+        try addMockSamples()
+        fixture.currentDateProvider.advanceBy(interval: 0.1)
+        
+        self.fixture.dispatchQueueWrapper.invokeLastDispatchAsync()
+        
+        try self.assertMetricsPayload()
+        try self.assertValidTraceProfileData(expectedSampleCount: expectedSampleCount)
```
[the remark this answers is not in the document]
philipphofmann: @armcknight, I actually removed the test. It doesn't really make sense. All we need to validate is that the logic stops the profiler on the calling thread. Everything else is validated here https://github.com/getsentry/sentry-cocoa/blob/0f8dc8883f038344b211921f5b59249623f9a752/Tests/SentryProfilerTests/SentryTraceProfilerTests.swift#L490-L615

I thought that the change pointed out in https://github.com/getsentry/sentry-cocoa/pull/4377/files#r1784119054 makes a difference. I wanted to add a test for it, but the test doesn't fail when I move `[profiler.state copyProfilingData]` into the async block. It doesn't make a difference then. Can you confirm that this change shouldn't make a difference?

philipphofmann:
> If and when you get around to bg'ing the continuous serialization function (maybe in a separate PR?)

Ups, sorry, that's not the goal of this PR. I only want to capture trace-based profiles on a BG thread. I updated the description and title.

Tests/SentryProfilerTests/SentryTraceProfilerTests.swift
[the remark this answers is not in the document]
armcknight: I explained in reply to that comment, but no, I think it shouldn't make a difference.

Sources/Sentry/Profiling/SentryCaptureTransactionWithProfile.mm
philipphofmann: OK, then it was a false alarm. I noticed that some app start transactions with profiles also sometimes have a shorter profile on the main branch: see [profile](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/2613a9cccfad44feaf8faccff7afb90c/flamegraph/?colorCoding=by%20library&query=&sorting=call%20order&tid=259&view=top%20down)
CleanShot 2024-10-09 at 10 43 19@2x

I guess that's a different problem then. Are you aware of such a problem, @armcknight?

PR #4397, opened 2024-10-03
chore: Bumping Xcode and SauceLab CLI for benchmarking

brustolin:
Our benchmarking workflows still uses Xcode 14 that does not work with mergeable-libraries.

_#skip-changelog_


philipphofmann (approved):
LGTM, but all the benchmark CI jobs are failing.

PR #4414, opened 2024-10-08
feat: Start/Stop session replay

brustolin:
## :scroll: Description

Added Api to manually control session replay start and stop

Unit tests

## :pencil: Checklist


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- feat: Start/Stop session replay (#4414)
```
philipphofmann, line 7: `l`: Maybe a bit cleaner.
```suggestion
- feat: API to Start/Stop Session Replay (#4414)
```

Sources/Sentry/SentryReplayApi.m
```diff
@@ -22,18 +23,49 @@ - (void)unmaskView:(UIView *)view
...
     [replayIntegration resume];
 }
 
+- (void)start
+{
+    SentrySessionReplayIntegration *replayIntegration
+        = (SentrySessionReplayIntegration *)[SentrySDK.currentHub
+            getInstalledIntegration:SentrySessionReplayIntegration.class];
+
+    if (replayIntegration == nil) {
```
philipphofmann, line 46: `h`: Is it really intended to install the integration the user didn't decide to enable it in the first place? IMO, this should be a NoOP, if SessionReplay isn't installed.

Sources/Sentry/SentryReplayApi.m
```diff
@@ -22,18 +23,49 @@ - (void)unmaskView:(UIView *)view
...
+        replayIntegration = [[SentrySessionReplayIntegration alloc] init];
+
+        SentryOptions *options = [[SentryOptions alloc] init];
+        options.enableSwizzling = SentrySDK.currentHub.client.options.enableSwizzling;
+        options.experimental.sessionReplay.sessionSampleRate = 1;
+        __unused BOOL installed = [replayIntegration installWithOptions:options];
```
philipphofmann, line 52: `h`: This can break easily. What if parts of the replay integration rely on other options? I think we must not create a new instance of the options. Instead, we should use the existing options.

Sources/Sentry/SentryReplayApi.m
```diff
@@ -22,18 +23,49 @@ - (void)unmaskView:(UIView *)view
...
+    SentrySessionReplayIntegration *replayIntegration
+        = (SentrySessionReplayIntegration *)[SentrySDK.currentHub
+            getInstalledIntegration:SentrySessionReplayIntegration.class];
+
+    if (replayIntegration == nil) {
+        replayIntegration = [[SentrySessionReplayIntegration alloc] init];
+
+        SentryOptions *options = [[SentryOptions alloc] init];
+        options.enableSwizzling = SentrySDK.currentHub.client.options.enableSwizzling;
+        options.experimental.sessionReplay.sessionSampleRate = 1;
```
philipphofmann, line 51: `h`: What if you want to sample your replays when using the start/stop API? I expect that the SDK uses the sample rate of the options even with the manual API.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -351,6 +351,35 @@ - (void)resume
...
+            [_notificationCenter addObserver:self
+                                    selector:@selector(newSceneActivate)
+                                        name:UISceneDidActivateNotification];
+        }
+    }
+}
+
+- (void)stop
+{
+    [self.sessionReplay pause];
```
philipphofmann, line 379: `l`: It's a bit confusing to me that this method here is named `pause`. It would be great to be consistent with the naming.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -351,6 +351,35 @@ - (void)resume
...
+        // If a window its already available start replay right away
+        [self startWithOptions:_replayOptions fullSession:YES];
+    } else {
+        // Wait for a scene to be available to started the replay
+        if (@available(iOS 13.0, tvOS 13.0, *)) {
+            [_notificationCenter addObserver:self
+                                    selector:@selector(newSceneActivate)
+                                        name:UISceneDidActivateNotification];
+        }
+    }
```
philipphofmann, line 374: `m`: That's duplicate code of `startSession`. Maybe we can reuse some code.

https://github.com/getsentry/sentry-cocoa/blob/913fab76543f3893b2e1f620546aa2e0636c3102/Sources/Sentry/SentrySessionReplayIntegration.m#L215-L225

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -351,6 +351,35 @@ - (void)resume
...
+    _startedAsFullSession = YES;
+    if (SentryDependencyContainer.sharedInstance.application.windows.count > 0) {
+        // If a window its already available start replay right away
+        [self startWithOptions:_replayOptions fullSession:YES];
+    } else {
+        // Wait for a scene to be available to started the replay
+        if (@available(iOS 13.0, tvOS 13.0, *)) {
+            [_notificationCenter addObserver:self
+                                    selector:@selector(newSceneActivate)
+                                        name:UISceneDidActivateNotification];
```
philipphofmann, line 372: `m`: I don't see that we remove `self` from `UISceneDidActivateNotification`. We should remove observers with the most detail possible; see https://developer.apple.com/documentation/foundation/nsnotificationcenter/1413994-removeobserver

> When removing an observer, remove it with the most specific detail possible. For example, if you used a name and object to register the observer, use [removeObserver:name:object:](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1407263-removeobserver) with the name and object.

philipphofmann (commented):
I'm a bit confused with the install integration if it's nil approach.

CHANGELOG.md
armcknight: or even
```suggestion
- feat: API to manually start/stop Session Replay (#4414)
```

Sources/Sentry/SentryReplayApi.m
armcknight: as a counterpoint, the continuous profiler does not respect any sample rate when using the start/stop calls, and we document that if customers want sampling, they need to handle it (arguable on whether we should offer it, but that's just how it is right now. maybe sample rates only make sense for the automatic integration stuff)

Sources/Sentry/SentryReplayApi.m
armcknight: yeah, i imagine the whole integration pattern as our way of letting users set things that will go on to work automatically after that. in this kind of API, start/stop should be alternative API to the integration, not ways of controlling an integration.

Sources/Sentry/SentrySessionReplayIntegration.m
armcknight: Is this a typo? I saw that there was a new function `stop` added to SentrySessionReplay: https://github.com/getsentry/sentry-cocoa/pull/4414/files#diff-7e90bc4012112a489d229dd525a4bb5c21ee3478152aae0f1c0b75b34aa85947R134

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -130,6 +130,16 @@ class SentrySessionReplay: NSObject {
...
         displayLink.link(withTarget: self, selector: #selector(newFrame(_:)))
     }
+    
+    func stop() {
+        lock.lock()
+        defer { lock.unlock() }
+        displayLink.invalidate()
+        if isFullSession {
+            prepareSegmentUntil(date: dateProvider.date())
+        }
```
armcknight, line 140: This is essentially `self.pause()` except for the last line in pause that sets `isSessionPaused = false`, could we reuse it? Although I would expect `pause()` to set `isSessionPaused = true`, I don't understand what's going on there.

Sources/Sentry/SentryReplayApi.m
brustolin: The start function is a way of manually control SR even when the integration was not started, so we can't use a NoOP.
SentryReplayApi is just a dummy interface to bridge commands over the session replay, we need a place to hold reference of session replay where we can use if later.
A late installed integration is as good as any other approach, and already have all the code to control the replay.

Sources/Sentry/SentryReplayApi.m
brustolin: > What if you want to sample your replays when using the start/stop API? I expect that the SDK uses the sample rate of the options even with the manual API.

Not an option, if the user call start it means they want to replay running for sure.

Sources/Sentry/SentryReplayApi.m
brustolin: A manually started replay integration should be controlled by its own set of options.
But there are options missing in here.

Maybe we could offer a parameter in the start function

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: > I saw that there was a new function `stop` 

I need to remove the SentrySessionReplay.stop function, it's unnecessary. 

> It's a bit confusing to me that this method here is named pause. It would be great to be consistent with the naming.

If I change the pause name to stop, in the integration pause function I would have to call replay stop.
stop function pauses it and also clean it in the next line.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
brustolin: I forget the stop function in here. I will remove it.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: We remove it at `newSceneActivate`

philipphofmann: Sorry, I had a typo in the message above. Yes, we remove it, but instead we should remove observers with the most detail possible ⬆️

Sources/Sentry/SentryReplayApi.m
philipphofmann: If that's intended, then all good. Just wanted to double-ckeck.

Sources/Sentry/SentryReplayApi.m
philipphofmann: Okay again another pain point of our integrations. It's confusing, but out of scope of this PR.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: Fixed

Sources/Sentry/SentryReplayApi.m
```diff
@@ -22,18 +23,46 @@ - (void)unmaskView:(UIView *)view
...
 {
-    SentrySessionReplayIntegration *replayIntegration =
-        [SentrySDK.currentHub getInstalledIntegration:SentrySessionReplayIntegration.class];
+    SentrySessionReplayIntegration *replayIntegration
+        = (SentrySessionReplayIntegration *)[SentrySDK.currentHub
+            getInstalledIntegration:SentrySessionReplayIntegration.class];
     [replayIntegration resume];
 }
 
+- (void)start
```
philipphofmann, line 40: `m`: Adding some tests for this here would be nice, as Codecov is complaining or is it broken?

philipphofmann (approved):
LGTM, maybe we miss some tests, but that's not blocking.

Sources/Sentry/SentryReplayApi.m
brustolin: We have 3 tests for this function alone:
```
func testStartWithNoSessionReplay() 
func testStartWithSessionReplayRunning() 
func testStartWithBufferSessionReplay() 
```

philipphofmann: Yeah I thought so. Go home Codecov, you're drunk.

Issue #4383, opened 2024-10-01
TTID is wrong when viewWillAppear isn't called


philipphofmann:
### Description

When the OS doesn't call `viewWillAppear` the TTID span gets finished when it's tracer finishes:

https://github.com/getsentry/sentry-cocoa/blob/8c38fb1263a5f545f452b8445045102c4eb6235d/Sources/Sentry/SentryTimeToDisplayTracker.m#L97-L101

This is because the `reportInitialDisplay` only gets called from the `SentryUIViewControllerPerformanceTracker` when `viewWillAppear` gets called:
https://github.com/getsentry/sentry-cocoa/blob/8c38fb1263a5f545f452b8445045102c4eb6235d/Sources/Sentry/SentryUIViewControllerPerformanceTracker.m#L191

There are scenarios for which the OS doesn't call viewWillAppear; see [Apple docs](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621510-viewwillappear). We could call `reportInitialDisplay` also in subsequent UIViewController methods that the OS calls before rendering the first frame of the UIViewController, such as `viewIsAppearing` or `viewWillLayoutSubView,` etc.

Especially when having `performanceV2` enabled, this leads to confusion because the initial frame render of the app start differs from the TTID. The app start metric subscribes to the very first frame being rendered https://github.com/getsentry/sentry-cocoa/blob/8c38fb1263a5f545f452b8445045102c4eb6235d/Sources/Sentry/SentryAppStartTracker.m#L68-L71

Maybe we need to align the initial frame render of the app start time with the TTID when fixing this bug.

PR #4417, opened 2024-10-09
fix: Finish TTID when viewWillAppear is skipped

philipphofmann:
## :scroll: Description

According to the Apple docs, viewWillAppear should be called before the UIViewController is added to the view hierarchy. There are some edge cases, though, when this doesn't happen, and we saw customers' transactions also proofing this. Therefore, we must also report the initial display here, as the customers' transactions had spans for viewWillLayoutSubViews.

Fixes GH-4383

Unit tests

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
+++ b/Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
@@ -198,10 +198,7 @@
         };
 
         [self.tracker activateSpan:spanId duringBlock:duringBlock];
-
-        SentryTimeToDisplayTracker *ttdTracker
-            = objc_getAssociatedObject(controller, &SENTRY_UI_PERFORMANCE_TRACKER_TTD_TRACKER);
-        [ttdTracker reportInitialDisplay];
+        [self reportInitialDisplay:controller];
     };
 
     [self limitOverride:@"viewWillAppear"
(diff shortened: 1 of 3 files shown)


Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -320,6 +317,15 @@ - (void)viewControllerViewWillLayoutSubViews:(UIViewController *)controller
...
         [self.tracker activateSpan:spanId duringBlock:duringBlock];
+
+        // According to the Apple docs
+        // (https://developer.apple.com/documentation/uikit/uiviewcontroller/1621510-viewwillappear),
+        // viewWillAppear should be called for before the UIViewController is added to the view
+        // hierarchy. There are some edge cases, though, when this doesn't happen, and we saw
+        // customers' transactions also proofing this. Therefore, we must also report the initial
+        // display here, as the customers' transactions had spans for `viewWillLayoutSubviews`.
+
+        [self reportInitialDisplay:controller];
```
brustolin, line 328: m: What about `viewDidAppear` does it also have edge cases where it does not get called?
I would prefer it over `viewWillLayoutSubViews`

brustolin (approved):
I have one suggestion, but it looks good to me

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
philipphofmann: The problem is that `viewDidAppear` could already be after a frame is rendered I guess.

brustolin: Actually many frames. It only trigger after the animation.

PR #4419, opened 2024-10-09
ref: Session replay performance for SwiftUI

brustolin:
## :scroll: Description

Improved Performance for SwiftUI session replay

Samples

## :pencil: Checklist


Sources/Swift/Tools/SentryViewPhotographer.swift
```diff
@@ -45,18 +45,24 @@ class SentryViewPhotographer: NSObject, SentryViewScreenshotProvider {
...
                 
+                var latestRegion: RedactRegion?
                 for region in redact {
                     let rect = CGRect(origin: CGPoint.zero, size: region.size)
                     var transform = region.transform
                     let path = CGPath(rect: rect, transform: &transform)
                     
+                    defer { latestRegion = region }
+                    
+                    guard region != latestRegion && imageRect.contains(path.boundingBoxOfPath) else { continue }
```
philipphofmann, line 64: `m`: It seems like this speeds up the performance significantly. Maybe it's worth adding a comment for this, cause we don't have a test to validate that we do this.

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -39,6 +39,10 @@ struct RedactRegion {
         self.type = type
         self.color = color
     }
+    
+    static func == (lhs: RedactRegion, rhs: RedactRegion) -> Bool {
+        lhs.size == rhs.size && lhs.transform == rhs.transform && lhs.type == rhs.type
```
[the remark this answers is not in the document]
brustolin: Yeah, we can do that.

Sources/Swift/Tools/SentryViewPhotographer.swift
brustolin: We do have a test to validate this.
Its the only test added to this PR

philipphofmann (approved):
LGTM

Issue #3056, opened 2023-05-24
Make SentryDebugImageProvider only available for HybridSDKs


philipphofmann:
### Description

The `SentryDebugImageProvider` is reserved for Hybrid SKDs and has not need to be public.

https://github.com/getsentry/sentry-cocoa/blob/4053ee90b75473db21ccfc2e1f361f22e2d0544c/Sources/Sentry/Public/SentryDebugImageProvider.h#L8-L11

We can move the header file to `Sources/Sentry/include/HybridPublic/`
https://github.com/getsentry/sentry-cocoa/blob/4053ee90b75473db21ccfc2e1f361f22e2d0544c/Sentry.podspec#L44

philprime:
I believe with the migration of `SentryDebugImageProvider.h` to `SentryDebugImageProviderInternal.h` in #6454 it is now exposed as a public Swift file via `SentryDebugImageProvider.swift`. Therefore it's not further actionable

Issue #4399, opened 2024-10-03
sentrycrashdl_getBinaryImageForHeader causes AppHangs


philipphofmann:
### Description

A customer reported around 10k app hangs pointing to `sentrycrashdl_getBinaryImageForHeader` when being called from the SentryTracer.

```
sentrycrashdl_getBinaryImageForHeader
-[SentryDebugImageProvider getDebugImagesCrashed:]
-[SentryDebugImageProvider getDebugImagesCrashed:]
-[SentryDebugImageProvider getDebugImagesForAddresses:isCrash:]
-[SentryDebugImageProvider getDebugImagesForFrames:isCrash:]
-[SentryTracer toTransaction]
-[SentryTracer finishInternal]
```

PR #4425, opened 2024-10-10
fix: Make SentryDebugImageProvider internal

philipphofmann:
## :scroll: Description

Move SentryDebugImageProvider.h to HybridPublic, as it's not supposed to be public. Strictly speaking, this is a breaking change, but nobody should have to use this API except the hybrid SDKs. We can always make it public again if we get complaints. Furthermore, remove a method not used by the SDK or hybrid SDKs.

It will make it easier to speed up `getDebugImagesForFrames` by using the `SentryBinaryImageCache` for GH-4399, because the `SentryBinaryImageCache` can't retrieve crash info.

Fixes GH-3056

CI.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 via the option `swizzleClassNameExclude`.
 - Add TTID/TTFD spans when loadView gets skipped (#4415)
 - Swizzling RootUIViewController if ignored by `swizzleClassNameExclude` (#4407)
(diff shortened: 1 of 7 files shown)


brustolin (commented):
This was made public because some users were creating their own Error event, and they wanted to add the debug image information.

Are we sure that we don’t want users to use this?

philipphofmann:
> This was made public because some users were creating their own Error event, and they wanted to add the debug image information.
> 
> Are we sure that we don’t want users to use this?

Ah, I didn't know that. Then we keep it for now.

Issue #4351, opened 2024-09-19
EXC_BAD_MEMORY for TestCurrentDateProvider.systemDate() on Xcode 16


philipphofmann:
### Description

Unit tests using `TestCurrentDateProvider`, such as SentryFramesTrackerTests or `SentryANRTrackerV2Tests`, fail with `EXC_BAD_MEMORY` when accessing `TestCurrentDateProvider.systemDate()`.
![Image](https://github.com/user-attachments/assets/ed3cef38-e89b-47f3-9623-ae019b9b0c09)

This is a known issue and some people report that a similar looking crash was fixed in Xcode Beta versions, see:
* https://github.com/aws-amplify/amplify-swift/issues/3806
* https://github.com/RevenueCat/purchases-ios/issues/3949

On this repository, the issue persists: https://github.com/RevenueCat/purchases-ios/issues/4177.

On Xcode 15.4, the unit tests run without failure.

kahest:
IIUC this is blocked by fix in Xcode and we wanna keep this here for everybody's information

philipphofmann:
Fixed with https://github.com/getsentry/sentry-cocoa/pull/4433.

PR #4433, opened 2024-10-14
chore: Fix tests for Xcode 16

brustolin:
Our tests are broken when running with Xcode 16.

Our setup is a little different:

- We have the SDK
- Two test targets
- A library used to share code in the tests

We use `@testable import Sentry` inside the library to create mock classes that we want to use in the tests. Essentially, we are trying to access an SDK private class from a test target by making it public through the library.

This used to work with Xcode 15, but it no longer works in Xcode 16. Maybe the Swift compiler changed something. I don’t think Apple will release a fix for this (I’m not even sure if this is a bug).

_#skip-changelog_


Sources/Swift/Helper/SentryCurrentDateProvider.swift
```diff
@@ -1,7 +1,28 @@
 @_implementationOnly import _SentryPrivate
 import Foundation
 
-@objcMembers class SentryCurrentDateProvider: NSObject {
+#if TEST
```
brustolin, line 4: It's an ugly solution for our unusual project setup.

Im also fine im making this class public, I dont think it would be a problem.

philipphofmann:
I guess this PR should fix this issue https://github.com/getsentry/sentry-cocoa/issues/4351.

Sources/Swift/Helper/SentryCurrentDateProvider.swift
philipphofmann: `m`: Wouldn't defining a protocol also work? The advantage would be that it's easier to notice when the two get out of sync, cause it will throw a compiler error. Also, it's easy add different implementations of `SentryCurrentDateProvider` for the tests.

```swift
@objc protocol SentryCurrentDateProvider {
    func date() -> Date
    func timezoneOffset() -> Int
    func systemTime() -> UInt64
    func systemUptime() -> TimeInterval
}

@objcMembers
class SentryDefaultCurrentDateProvider: NSObject, SentryCurrentDateProvider {
    
    func date() -> Date {
        return Date()
    }

    func timezoneOffset() -> Int {
        return TimeZone.current.secondsFromGMT()
    }

    func systemTime() -> UInt64 {
        getAbsoluteTime()
    }
    
    func systemUptime() -> TimeInterval {
        ProcessInfo.processInfo.systemUptime
    }
}

**Review by philipphofmann — COMMENTED** [review]
Maybe a protocol is the better choice here?

File: Sources/Swift/Helper/SentryCurrentDateProvider.swift (hunk 1 above)
Comment (brustolin) [reply Sources/Swift/Helper/SentryCurrentDateProvider.swift]: It seems that the same problem does not occur for a private Protocol
Good!!

Its a bigger refactoring tho.

File: Sources/Swift/Helper/SentryCurrentDateProvider.swift [hunk 2]
```diff
@@ -1,16 +1,24 @@
...
 
-@objcMembers class SentryCurrentDateProvider: NSObject {
-    
+@objc
+protocol SentryCurrentDateProvider {
+    func date() -> Date
+    func timezoneOffset() -> Int
+    func systemTime() -> UInt64
+    func systemUptime() -> TimeInterval
+}
```
philipphofmann, line 10: 💯 way better than the other solution and the refactoring isn't that big either:
CleanShot 2024-10-15 at 14 00 15@2x

Sources/Swift/Helper/SentryCurrentDateProvider.swift
```diff
@@ -1,16 +1,24 @@
 @_implementationOnly import _SentryPrivate
 import Foundation
 
-@objcMembers class SentryCurrentDateProvider: NSObject {
-    
+@objc
+protocol SentryCurrentDateProvider {
```
philipphofmann, line 5: `m`: Maybe explain why we need the protocol here. I tend to forget these things and then maybe refactor it away in a couple of weeks and then wonder why the tests now start to fail again 😅

philipphofmann (approved):
Thanks.

Sources/Swift/Helper/SentryCurrentDateProvider.swift
```diff
@@ -1,16 +1,24 @@
 @_implementationOnly import _SentryPrivate
 import Foundation
 
-@objcMembers class SentryCurrentDateProvider: NSObject {
-    
+@objc
```
brustolin, line 4:
```suggestion
/**
 * We need a protocol to expose SentryCurrentDateProvider to tests.
 * Mocking the previous private class from `SentryTestUtils` stopped working in Xcode 16.
*/
@objc
```

Sources/Swift/Helper/SentryCurrentDateProvider.swift
```diff
@@ -1,16 +1,28 @@
 @_implementationOnly import _SentryPrivate
 import Foundation
 
-@objcMembers class SentryCurrentDateProvider: NSObject {
-    
+/**
+ * We need a protocol to expose SentryCurrentDateProvider to tests.
+ * Mocking the previous private class from `SentryTestUtils` stopped working in Xcode 16.
```
philipphofmann, line 6: 👍

philipphofmann (approved):
LGTM.

Issue #4427, opened 2024-10-10
Cancel non used view controllers transactions


brustolin:
### Description

Its possible to create a view controller and programmatically interact with its view without adding it to the view hierarchy:

```swift
var downloadScreen: DownloadViewController?
func prepareDownloadScreen() {
    downloadScreen = DownloadViewController()
    downloadScreen?.view.backgroundColor = .white //This will trigger loadView
}
```

Since loadView was called, we start a new transaction. However, it’s possible that the view controller was initialized in advance to speed things up, even though it was never added to the view hierarchy. Although this approach is not recommended, it is possible.

The transaction associated with this view controller will eventually time out, which is ultimately useless information to know. Additionally, its TTID will be as long as the transaction itself, since no other lifecycle functions were called.

PR #4437, opened 2024-10-14
Fix: Dont create transaction for unused ViewControllers

brustolin:
## :scroll: Description

Its possible to create a view controller but never add it to the view hierarchy.
This will create a transaction with data that is not helpful

Unit tests

## :pencil: Checklist


Sources/Sentry/SentryTracer.m
```diff
@@ -672,7 +681,11 @@ - (SentryTransaction *)toTransaction
     NSMutableArray<id<SentrySpan>> *spans = [[NSMutableArray alloc] initWithCapacity:capacity];
 
     @synchronized(_children) {
-        [spans addObjectsFromArray:_children];
+        for (SentrySpan *child in _children) {
+            if (child.shouldIgnore == NO) {
+                [spans addObject:child];
+            }
```
philipphofmann, line 687: `h`: If I'm not mistaken, this approach won't work for nested UIViewControllers, as this would only remove the spans for `loadView` and `viewDidLoad` and not the other spans, such as TTID/TTFD.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerPerformanceTrackerTests.swift
```diff
@@ -132,6 +132,7 @@ class SentryUIViewControllerPerformanceTrackerTests: XCTestCase {
             callbackExpectation.fulfill()
         }
         let tracer = try XCTUnwrap(transactionSpan as? SentryTracer)
+        XCTAssertTrue(tracer.shouldIgnore)
```
philipphofmann, line 135: `m`: It seems to me that these test additions don't fully cover the edge case described in GH-4427. I would love to see a test that emulates these edge cases.

philipphofmann (commented):
I'm unsure if this approach will work for nested UIViewControllers. I'm not a big fan of the `shouldIgnore` flag approach. The logic for covering the edge case of unused view controllers is spread across multiple classes. Flipping some flags on the spans looks like something that can break easily. 

What about a transaction validation/span filtering logic? We could have a method looking all the spans with some metadata and validating that it is a correct transaction. In our case, the logic checks if it's an auto-generated transaction with a ui.load operation. It then validates that the transaction has a `loadView` or a `viewDidLoad` spans, and also spans for `viewWillDisappear` or `viewDidAppear`. The logic would discard the transaction or filter spans for nested view controllers that never get loaded.
The max transaction duration logic should also be in that validation logic
https://github.com/getsentry/sentry-cocoa/blob/79eb67d4633ce0f8346e7a91791e1023b0138eef/Sources/Sentry/SentryTracer.m#L560-L570

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerPerformanceTrackerTests.swift
brustolin: This addition are meant to check whether `SentryUIViewControllerPerformanceTrackerTests.swift` is properly setting this values.

The test to check whether we create or not the transaction belongs to `SentryTracerTests.swift`

Sources/Sentry/SentryTracer.m
philipphofmann: Sorry, TTID/TTFD was a bad example then. What about other spans, such as file IO, DB queries, manually added ones, or other future spans that we aren't yet aware of?

brustolin: Ok, I can see whether this could happen.
What do you think we should do to them?

armcknight: The validation idea mentioned in the top-level comment sounds promising 👍🏻

brustolin: We probably need to talk about this in person, maybe I didn't get it but your suggestion in the PR review comment does not solve the problem raised in here.

Also, I would not like to have a "scenario specific" logic inside the generic SentryTracer class, it feels more hacky than the one flag solution that indicates whether the span was validated by its creator or not.

brustolin:
> I'm not a big fan of the shouldIgnore flag approach. The logic for covering the edge case of unused view controllers is spread across multiple classes

I think you’re looking at this from the wrong perspective. Don’t think about `shouldIgnore` as a solution for unused view controller transactions. Instead, consider it a broader solution: a flag indicating that a particular span was not ‘approved’ by the mechanism that created it.

philipphofmann:
The `SentryUIViewControllerPerformanceTracker` calls finish on the tracer when the UIViewController is added to the view hierarchy / appears on the screen. Only when this happens can we keep the transaction. So, instead of putting a flag on the span, @brustolin and I discussed adding a new config option to the tracer. This flag will be called something `finishMustBeCalled`. When this flag is set to true, the transaction must be finished before the deadline timer fires. Otherwise, the tracer discards the transaction.

PR #4440, opened 2024-10-15
Expose `HybridSDK` from `Sentry` framework

denrase:
## :scroll: Description

We want to support SPM in our sentry flutter framework. In order to do this, we need the `SentryHybridSDK` classes, which we expose though Cocoapods, to also be available in our SPM package.

This PR adds an explicit module with the hybrid headers, except "SentryEnvelope.h".

Unblocks https://github.com/getsentry/sentry-dart/pull/2280
## :pencil: Checklist


denrase:
@philipphofmann Maybe it's enough to mark the headers in question as package private, wdyt?

Sources/Resources/Sentry.modulemap
```diff
@@ -3,4 +3,24 @@ framework module Sentry {
...
+        header "SentryDependencyContainer.h"
+//        header "SentryEnvelope.h"
+        header "SentryEnvelopeItemType.h"
+        header "SentryFormatter.h"
+        header "SentryFramesTracker.h"
+        header "SentryOptions+HybridSDKs.h"
+        header "SentryScreenFrames.h"
+        header "SentrySessionReplayIntegration-Hybrid.h"
+        header "SentrySwizzle.h"
+        header "SentryUser+Private.h"
```
brustolin, line 23: h: I think we should create an umbrella header for private headers. 
It gives us more control if necessary later.

Sentry.xcodeproj/project.pbxproj
```diff
@@ -5306,6 +5308,7 @@
 				GCC_WARN_SHADOW = YES;
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				LD_DYLIB_INSTALL_NAME = "$(DYLIB_INSTALL_NAME_BASE:standardizepath)/$(EXECUTABLE_PATH)";
+				MODULEMAP_PRIVATE_FILE = "";
```
brustolin, line 5311: l: Please remove all this empty `MODULEMAP_PRIVATE_FILE` entries

brustolin:
We will need a test workflow of sample using `PrivateSentrySDKOnly` from Swift, so we make sure to not break it.

Sources/Resources/Sentry.modulemap
```diff
@@ -3,4 +3,24 @@ framework module Sentry {
 
     export *
     module * { export * }
+    
+    explicit module HybridSDK {
```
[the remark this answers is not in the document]
philipphofmann: > I think we can rename it to _HybridSDK, its a hack way to prevent autocomplete to suggest it.

Sounds reasonable.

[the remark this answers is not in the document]
denrase: Ok, i just tried this in one of the sample apps. Adding the headers through the `explicit model` makes them exposed, even if just importing Sentry, independent from the module name. Assuming we want to avoid this, this approach is not working either. Do you have another suggestion how to address this?

Bildschirmfoto 2024-10-21 um 14 15 59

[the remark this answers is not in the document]
denrase: It's because the headers for the hybrid SDK moved from `project` to `private` visibility, so they can be exported by the explicit module. But as i understand this, this also exposes them in `Sentry`, making the `explicit module` redundant. I'm sure i'm missing something here.

denrase:
Closed in favor of #4486

denrase:
Closed in favor of https://github.com/getsentry/sentry-cocoa/pull/4486

Sources/Resources/Sentry.modulemap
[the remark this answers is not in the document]
denrase: Turns out this was just an issue in the sample apps. When using the created framework in Flutter, the explicit header has to be imported.

denrase:
@brustolin @philipphofmann Turns our we do need the explicit module after all to import the needed files in Flutter when integrating through SPM. Therefore i reopened this PR.

Sources/Resources/Sentry.modulemap
```diff
@@ -3,4 +3,32 @@ framework module Sentry {
...
+        header "SentryScreenFrames.h"
+        header "SentrySwizzle.h"
+        header "SentryUser+Private.h"
+
+        header "SentryBaseIntegration.h"
+        header "SentrySessionReplayIntegration.h"
+        header "SentrySessionReplayIntegration-Hybrid.h"
+        
+        header "SentrySdkInfo.h"
+        header "SentryInternalSerializable.h"
```
brustolin, line 30: l: Is it possible to have one header that has all this other headers as reference? The same way we do with `Sentry.h`.

Just because a header can be more versatile.

brustolin (commented):
We need to implement some CI tests to prevent regression, as it’s easy to break since this is only used by hybrid SDKs.

Maybe trying to access `Sentry._Hybrid` from the test would be enough.

Sources/Resources/Sentry.modulemap
denrase: Had some issues with this approach, but will try again.

denrase: I always get an error about non-modular headers when i try to move the header imports into one header.

Bildschirmfoto 2024-11-13 um 14 10 00

brustolin: Ok, no problem. We can continue without it.

PR #4439, opened 2024-10-15
Fix: Thread running at User-initiated quality-of-service for Session replay

brustolin:
## :scroll: Description

AVFoundation video maker runs on a lower QoS than user-initiated processes, and the thread sanitizer complains about waiting for it on a higher-level thread. We are changing the QoS of our call to create video.

## :bulb: Motivation and Context

Reported in here: https://github.com/getsentry/sentry/discussions/74322#discussioncomment-10934816

## :green_heart: How did you test it?

Unit test

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I reviewed the submitted code.
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] Review from the native team if needed.
- [x]  No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #4442, opened 2024-10-15
impr: Align SDK dispatch queue names

philipphofmann:
## :scroll: Description

All dispatch queues now start with the prefix `io.sentry`.

Came up in https://github.com/getsentry/sentry-cocoa/pull/4439#discussion_r1801008595.

CI still green.

## :pencil: Checklist


brustolin (approved):

philipphofmann:
> IMO we should not allow usage of `[[SentryDispatchQueueWrapper alloc] init]` at all because that lets the SDK create multiple queue instances with the same label. Each should be unique.

What would you propose to fix this?

PR #4446, opened 2024-10-15
fix: Delete old session replay files

brustolin:
## :scroll: Description

Clean up for old session replay files

Unit tests

## :pencil: Checklist


Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -359,6 +367,43 @@ - (void)moveCurrentReplay
...
 }
 
+- (void)cleanUp
+{
+    NSURL *replayDir = [self replayDirectory];
+    NSDictionary<NSString *, id> *lastReplayInfo = [self lastReplayInfo];
+    NSString *lastReplayFolder = lastReplayInfo[@"path"];
+
+    // Mapping replay folder here and not in dispatched queue to prevent a race condition between
+    // listing files and creating a new replay session.
```
armcknight, line 377: The dispatch queue in the dependency manager is a serial queue, so there shouldn't be a danger of race condition: https://github.com/getsentry/sentry-cocoa/blob/2d2068d4247d91a9143c43c4681805caa06d0443/Sources/Sentry/SentryDispatchQueueWrapper.m#L13-L14

In fact, you might even be creating a race condition by not placing this on the serial queue, depending on what other replay operations are or aren't also performed on the queue. Something that was previously deferred might be interleaved with this line and the next block enqueued below in various combinations. 

For instance, what happens if multiple calls to `SentryReplayApi.start` are received in quick succession?

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -359,6 +367,43 @@ - (void)moveCurrentReplay
...
+            }
+
+            // Check if the file exists and is a directory before deleting it.
+            BOOL isDirectory = NO;
+            if ([fileManager fileExistsAtPath:filePath isDirectory:&isDirectory] && isDirectory) {
+                NSError *error = nil;
+                if (![fileManager removeItemAtPath:filePath error:&error]) {
+                    SENTRY_LOG_ERROR(@"Error deleting file at path: %@, error: %@", filePath,
+                        error.localizedDescription);
+                }
```
armcknight, line 401: we already have a better implementation of this logic at https://github.com/getsentry/sentry-cocoa/blob/2d2068d4247d91a9143c43c4681805caa06d0443/Sources/Sentry/SentryFileManager.m#L43-L58

armcknight: Also, what happens if `isDirectory` turns out to be false?

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -359,6 +367,43 @@ - (void)moveCurrentReplay
...
+
+    [SentryDependencyContainer.sharedInstance.dispatchQueueWrapper dispatchAsyncWithBlock:^{
+        NSFileManager *fileManager = [NSFileManager defaultManager];
+        for (NSString *file in replayFiles) {
+            NSString *filePath = [replayDir.path stringByAppendingPathComponent:file];
+
+            // Skip the last replay folder.
+            if ([file isEqualToString:lastReplayFolder]) {
+                continue;
+            }
```
philipphofmann, line 392: `l`: Putting the check on top avoids creating the filePath when nor required.
```suggestion
            // Skip the last replay folder.
            if ([file isEqualToString:lastReplayFolder]) {
                continue;
            }
            
            NSString *filePath = [replayDir.path stringByAppendingPathComponent:file];
```

Sources/Sentry/SentrySessionReplayIntegration.m
philipphofmann: `h`: I would prefer using the `SentryFileManager.removeFileAtPath`, which uses `_non_thread_safe_removeFileAtPath` with a synchronize. We often use `NSFileManager.defaultManager` in this class. Moving the `removeItemAtPath` to a BG thread will cause race conditions as mentioned by @armcknight above https://github.com/getsentry/sentry-cocoa/pull/4446/files#r1801517947. I think we actually need to replace all calls of `NSFileManager.defaultManager` with calls to `SentryFileManager` which already has a working locking strategy in place. It's a bit conservative by mainly using `synchronized(self)` but it works. Or maybe you have good reasons to use `NSFileManager.defaultManager`, that I'm unware of.

brustolin: > we already have a better implementation of this logic at

`_non_thread_safe_removeFileAtPath` is unnecessary noise, it's private, and the public function that calls it synchronizes on self, which is not required here

> Also, what happens if isDirectory turns out to be false?

We dont remove it. We just want to remove replay sessions, the other files in this folders should not be touched.

brustolin: I think using `synchronized(self)` its an unnecessary bottle neck that may slow down other integrations startup. `SentrySessionReplayIntegration` is the only place session replays are managed, and the setup of the integration itself will only happen once. Furthermore, if you ask FileManager to delete the same file twice (in a race condition) what actually happen is one of those calls will return an error that will be printed in the log (file not found or something).

> Also, what happens if isDirectory turns out to be false?

We dont delete it. It means that the file it's one of the info files and not a left over replay session.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: > The dispatch queue in the dependency manager is a serial queue

But the `SentrySessionReplay` class that creates a new replay session has it own queue.

> what happens if multiple calls to SentryReplayApi.start

The Integration is initialized only once and the cleanUp will run only once.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: > SentryReplayAPI.start, a public API, calls setupWith, which calls cleanUp.

Actually, `SentryReplayAPI.start` initialize the integration that uses those methods.
An integration should be unique, nor two of the same integration should be installed at the same time, and this should be atomic, if not, we should fix it.

And then, consider `moveCurrentReplay` and `cleanUp` part of the initialization process that are split in different functions for convenience.

But no problem, I will syncronize all of it because actually the performance penalty is not worth going back and forth in a discussion.

philipphofmann: Thanks

Sources/Sentry/SentrySessionReplayIntegration.m
armcknight: Hm, the only reason I mentioned the dependency manager's queue is because that's what's being used in this file. Should we use one queue for all replay activity? Why the dependency manager's here and the private queue in SentrySessionReplay.swift, wouldn't that open up the possibility of more race conditions?

brustolin: No, I dont think there is a need for shared dispatch queue, the integration do some setup, and SessionReplay do the rest.

Sources/Sentry/SentrySessionReplayIntegration.m
philipphofmann: @brustolin, I'm a bit confused. You stated above you are going to add sync to all of it, but I don't see any of that or am I missing something?

brustolin: I did remove the use of NSFileManager in favor of SentryFileManager.
Am I missing something?

philipphofmann: We just chatted about this on a call. There aren't any race conditions. It was a misunderstanding from my side.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -365,6 +373,37 @@ - (void)moveCurrentReplay
...
+    NSURL *replayDir = [self replayDirectory];
+    NSDictionary<NSString *, id> *lastReplayInfo = [self lastReplayInfo];
+    NSString *lastReplayFolder = lastReplayInfo[@"path"];
+
+    SentryFileManager *fileManager = SentryDependencyContainer.sharedInstance.fileManager;
+    // Mapping replay folder here and not in dispatched queue to prevent a race condition between
+    // listing files and creating a new replay session.
+    NSArray *replayFiles = [fileManager allFilesInFolder:replayDir.path];
+    if (replayFiles.count == 0) {
+        return;
```
philipphofmann, line 387: `l`: Maybe also add a test for this.

philipphofmann (approved):
LGTM

PR #4448, opened 2024-10-16
Fix: Dont create transaction for unused ViewControllers

brustolin:
## :scroll: Description

Its possible to create a view controller but never add it to the view hierarchy.
This will create a transaction with data that is not helpful

Unit tests

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryTracer.m
+++ b/Sources/Sentry/SentryTracer.m
@@ -96,6 +96,7 @@ static const NSTimeInterval SENTRY_AUTO_TRANSACTION_DEADLINE = 30.0;
     dispatch_block_t _idleTimeoutBlock;
     NSMutableArray<id<SentrySpan>> *_children;
     BOOL _startTimeChanged;
+    BOOL _timeout;
     NSObject *_idleTimeoutLock;
 
 #if SENTRY_HAS_UIKIT
@@ -148,6 +149,8 @@ static BOOL appStartMeasurementRead;
     self.wasFinishCalled = NO;
     _measurements = [[NSMutableDictionary alloc] init];
     self.finishStatus = kSentrySpanStatusUndefined;
+    self.finishMustBeCalled = NO;
+    _timeout = YES;
 
     if (_configuration.timerFactory == nil) {
         _configuration.timerFactory = [[SentryNSTimerFactory alloc] init];
@@ -289,6 +292,8 @@ static BOOL appStartMeasurementRead;
 - (void)deadlineTimerFired
 {
     SENTRY_LOG_DEBUG(@"Sentry tracer deadline fired");
+    _timeout = YES;
+
     @synchronized(self) {
         // This try to minimize a race condition with a proper call to `finishInternal`.
         if (self.isFinished) {
@@ -303,7 +308,8 @@ static BOOL appStartMeasurementRead;
         }
     }
 
-    [self finishWithStatus:kSentrySpanStatusDeadlineExceeded];
+    _finishStatus = kSentrySpanStatusDeadlineExceeded;
+    [self finishInternal];
 }
 
 - (void)cancelDeadlineTimer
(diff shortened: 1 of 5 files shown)


Sources/Sentry/SentryTracer.m
```diff
@@ -96,6 +96,7 @@ @implementation SentryTracer {
     dispatch_block_t _idleTimeoutBlock;
     NSMutableArray<id<SentrySpan>> *_children;
     BOOL _startTimeChanged;
+    BOOL _timeout;
```
philipphofmann, line 99: `m`: I'm confused. We only set this to YES, and never use it. Or am I missing something?

Sources/Sentry/SentryTracer.m
```diff
@@ -303,7 +308,8 @@ - (void)deadlineTimerFired
         }
     }
 
-    [self finishWithStatus:kSentrySpanStatusDeadlineExceeded];
+    _finishStatus = kSentrySpanStatusDeadlineExceeded;
+    [self finishInternal];
```
philipphofmann, line 312: The logic in the tracer is so complicated 🤦. I'm only complaining. We should simplify this at some point, but not in this PR.

Sources/Sentry/SentryTracer.m
```diff
@@ -581,6 +587,12 @@ - (void)finishInternal
         }
     }];
 
+    if (self.finishMustBeCalled && !self.wasFinishCalled) {
+        SENTRY_LOG_DEBUG(
+            @"Not capturing transaction because finish was not called before timing out.");
+        return;
+    }
```
philipphofmann, line 594: `m`: Wouldn't it be cleaner to move this right below ⬇️ , so we avoid extra work?
https://github.com/getsentry/sentry-cocoa/blob/b695b615f3b6fe1c5eacc4b298572524446b83dc/Sources/Sentry/SentryTracer.m#L524-L527

Sources/Sentry/include/SentryTracer.h
```diff
@@ -43,6 +43,13 @@ static const NSTimeInterval SENTRY_AUTO_TRANSACTION_MAX_DURATION = 500.0;
 
 @property (nullable, nonatomic, copy) BOOL (^shouldIgnoreWaitForChildrenCallback)(id<SentrySpan>);
 
+/**
+ * This flag indicates whether the trace should be captured when the timeout triggers.
+ * If Yes, this tracer will be discarced in case the timeout triggers.
+ * Default @c NO
+ */
+@property (nonatomic) BOOL finishMustBeCalled;
```
philipphofmann, line 51: `m`: Any good reason why this can't be part of `SentryTracerConfiguration`? Logically it would make sense to me to add it there.

Tests/SentryTests/Transaction/SentryTracerTests.swift
```diff
@@ -1292,6 +1292,15 @@ class SentryTracerTests: XCTestCase {
     }
 #endif
     
+    func testFinishShouldBeCalled_Timeout_NotCaptured() {
```
philipphofmann, line 1295: 💯 Such a neat test 👏

Sources/Sentry/include/SentryTracer.h
philipphofmann: Ah, I guess because we need to set the hub to nil and call all the callback, right?

Sources/Sentry/SentryTracer.m
brustolin: True. This was a first idea that I abandoned, after learning more about our messy tracer flow.
I will remove it

Sources/Sentry/SentryTracer.m
brustolin: We need to remove the transaction from the scope if that is the case.

Sources/Sentry/include/SentryTracer.h
brustolin: > Any good reason why this can't be part of SentryTracerConfiguration? Logically it would make sense to me to add it there

Actually because, `SentryPerformanceTracker` is the one creating the tracer, but `SentryUIViewControllerPerformanceTracker` is the class that knows those tracers need that flag.
I guess we could add an extra argument in `SentryPerformanceTracker`.

PR #4455, opened 2024-10-17
ref: Make `SentrySDK.replay.start()` thread safe

brustolin:
## :scroll: Description

Making `SentrySDK.replay.start()` thread safe to avoid multiple initialization

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryReplayApi.m
+++ b/Sources/Sentry/SentryReplayApi.m
@@ -39,20 +39,25 @@
 
 - (void)start
 {
-    SentrySessionReplayIntegration *replayIntegration
+    // Start could be misused and called multiple times, causing it to
+    // be initialized more than once before being installed.
+    // Synchronizing it will prevent this problem.
+    @synchronized (self) {
+        SentrySessionReplayIntegration *replayIntegration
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentryReplayApi.m
```diff
@@ -39,20 +39,26 @@ - (void)resume
 
 - (void)start
 {
-    SentrySessionReplayIntegration *replayIntegration
-        = (SentrySessionReplayIntegration *)[SentrySDK.currentHub
-            getInstalledIntegration:SentrySessionReplayIntegration.class];
+    // Start could be misused and called multiple times, causing it to
+    // be initialized more than once before being installed.
+    // Synchronizing it will prevent this problem.
+    @synchronized(self) {
```
philipphofmann, line 45: `m`: I guess a double-checked lock would make sense here to avoid acquiring the lock in 99.99% of the cases after the first start call. We use it in the dependency container. Please use `SENTRY_DISABLE_THREAD_SANITIZER` to disable the thread sanitizer warning.

philipphofmann (approved):
LGTM

philipphofmann (approved):
Thanks, LGTM.

PR #4459, opened 2024-10-21
test: Add macOS SwiftUI sample app

philipphofmann:
Add a sample app for macOS with SwiftUI using the
NSApplicationDelegateAdaptor.

This is required because a customer has this setup and would like to make uncaught NSExceptions work correctly. With this sample app, we can test their setup.

#skip-changelog


brustolin (approved):
Just one suggestion, I would prefer to have less assets (in this case the icons). 
With cocoapod or SPM, our users download all the samples too, the less bits the better.

philipphofmann:
> Just one suggestion, I would prefer to have less assets (in this case the icons). With cocoapod or SPM, our users download all the samples too the less bits the better.

Excellent point. I will update the PR after merging this one: https://github.com/getsentry/sentry-cocoa/pull/4461.

PR #4461, opened 2024-10-21
test: Add shared samples assets

philipphofmann:
Remove duplicated AppIcon assets by using one shared samples assets.

#skip-changelog


brustolin (approved):
This is perfect. Thanks!!

PR #4480, opened 2024-10-29
fix: Too many navigation breadcrumbs for Session Replay

brustolin:
## :scroll: Description

We create a navigation breadcrumb for every view controller that appears, but some view controllers are child view controllers and they dont represent a navigation.

Filtering out this child breadcrumbs in the session replay

Fixes https://github.com/getsentry/team-replay/issues/467

Unit tests

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
+++ b/Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
@@ -269,13 +269,25 @@ class SentrySessionReplay: NSObject {
             SentryLog.debug("Could not delete replay segment from disk: \(error.localizedDescription)")
         }
     }
-
+    
     private func convertBreadcrumbs(breadcrumbs: [Breadcrumb], from: Date, until: Date) -> [any SentryRRWebEventProtocol] {
-        return breadcrumbs.filter {
-            guard let time = $0.timestamp, time >= from && time < until else { return false }
-            return true
+        var filteredResult: [Breadcrumb] = []
+        var lastNavigationTime: Date = from.addingTimeInterval(-1)
+        
+        for breadcrumb in breadcrumbs {
+            guard let time = breadcrumb.timestamp, time >= from && time < until else { continue }
+            
+            // If it's a "navigation" breadcrumb, check the timestamp difference from the previous breadcrumb.
+            // Skip any breadcrumbs that have occurred within 50ms of the last one,
+            // as these represent child view controllers that don’t need their own navigation breadcrumb.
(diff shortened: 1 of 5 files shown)


Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -269,13 +269,25 @@ class SentrySessionReplay: NSObject {
...
-
+    
     private func convertBreadcrumbs(breadcrumbs: [Breadcrumb], from: Date, until: Date) -> [any SentryRRWebEventProtocol] {
-        return breadcrumbs.filter {
-            guard let time = $0.timestamp, time >= from && time < until else { return false }
-            return true
+        var filteredResult: [Breadcrumb] = []
+        var lastNavigationTime: Date = from.addingTimeInterval(-1)
+        
+        for breadcrumb in breadcrumbs {
```
philipphofmann, line 277: `h`: If I'm not mistaken, we grab the breadcrumbs from the scope. So, if there are too many breadcrumbs on the scope, shouldn't we instead fix it there so events and transactions also benefit from this improvement?

brustolin: This would be some kind of break change. 
Also, this is actually not a bug; it's the way it was designed for Cocoa and view controllers, but not desirable for the session replay player. 

We could discuss if it makes sense to also apply this change to general navigation breadcrumbs. But I would do it in another PR and then remove this change.

brustolin: This is a team decision. The session replay player timeline has too many of the same breadcrumbs in the same place. Breadcrumbs in session replay are used to navigate the replay, but having multiple items at the same point in time doesn’t add any value.

This is a different product with a different presentation. Nobody had complained about navigation breadcrumbs for issues or transactions, let’s not touch it.

brustolin: By the way, perhaps you’re thinking that this PR is removing something that shouldn’t be there or fixing an issue with the source of breadcrumbs. But, that’s not the case. Each individual breadcrumb was created by a distinct ViewController, as a view controller can have multiple child view controllers. This information does make sense for debugging purposes (in issues and transactions).

philipphofmann: Thanks for explaining. I still don't agree, but I accept the decision.

philipphofmann (approved):
LGTM

PR #4482, opened 2024-10-29
test(critical): Remove iOS 17

krystofwoldrich:
It's flaky, as the only test from iOS 16, 17 and 18, the 17 keeps failing on XCUIServer not found.

#skip-changelog


philipphofmann (approved):
LGTM, when you add a comment.

PR #4495, opened 2024-10-31
internal(hybrid): Add Replay Mask options to dict init

krystofwoldrich:
## :scroll: Description

Add `quality` option for hybrid sdks using native replay.

Needed to expose the option in RN.

unit tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 - Log a warning when dropping envelopes due to rate-limiting (#4463)
 - Expose `SentrySessionReplayIntegration-Hybrid.h` as `private` (#4486)
 - Add `maskedViewClasses` and `unmaskedViewClasses` to SentryReplayOptions init via dict (#4492)
-- Add `quality` to SentryReplayOptions init via dict (#4494)
+- Add `quality` to SentryReplayOptions init via dict (#4495)
 
 ## 8.39.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add option to report uncaught NSExceptions on macOS (#4471)
 - Build visionOS project with static Sentry SDK (#4462)
+- Too many navigation breadcrumbs for Session Replay (#4480)
 - Time-of-check time-of-use filesystem race condition (#4473)
 - Capture all touches with session replay (#4477)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -393,7 +393,7 @@ to receive SIGTERM events, set the option `enableSigtermReporting = true`.
 
 ### Features
 
-The following two features, disabled by default, were mistakenly added to the release. We usually only add features in minor releases. 
+The following two features, disabled by default, were mistakenly added to the release. We usually only add features in minor releases.
 
 - Add option to use own NSURLSession for transport (#3811)
 - Support sending GraphQL operation names in HTTP breadcrumbs (#3931)
@@ -414,9 +414,9 @@ The following two features, disabled by default, were mistakenly added to the re
 
 - Remove not needed lock for logging (#3934)
 - Session replay Improvements (#3877)
-  - Use image average color and text font color to redact session replay 
+  - Use image average color and text font color to redact session replay
   - Removed iOS 16 restriction from session replay
-  - Performance improvement 
+  - Performance improvement
 
 ## 8.25.0
 
@@ -428,7 +428,7 @@ The following two features, disabled by default, were mistakenly added to the re
 ### Fixes
 
 - Crash due to a background call to -[UIApplication applicationState] (#3855)
-- Save framework without UIKit/AppKit as Github Asset for releases (#3858) 
+- Save framework without UIKit/AppKit as Github Asset for releases (#3858)
 - Fix crash associated with runtime collision in global C function names (#3862)
 - Remove wrong error log in SentryCoreDataTracker (#3894)
 - Don't transmit device boot time in envelopes enriched with crash data (#3912, #3916)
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - Log a warning when dropping envelopes due to rate-limiting (#4463)
 - Expose `SentrySessionReplayIntegration-Hybrid.h` as `private` (#4486)
 - Add `maskedViewClasses` and `unmaskedViewClasses` to SentryReplayOptions init via dict (#4492)
+- Add `quality` to SentryReplayOptions init via dict (#4495)
```
armcknight, line 19: Why does this need a changelog entry?

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -2,31 +2,35 @@ import Foundation
...
     public enum SentryReplayQuality: Int {
+        static let names = [ "low", "medium", "high"]
+
         /**
          * Video Scale: 80%
          * Bit Rate: 20.000
          */
-        case low
-        
+        case low = 0
```
armcknight, line 17: Because this is already an `: Int` enum, these same values are already implied.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -2,31 +2,35 @@ import Foundation
...
 @objcMembers
 public class SentryReplayOptions: NSObject, SentryRedactOptions {
-    
+
     /**
      * Enum to define the quality of the session replay.
      */
     @objc
     public enum SentryReplayQuality: Int {
+        static let names = [ "low", "medium", "high"]
```
armcknight, line 11: Instead of doing this where we have to maintain the indexing to the int values, let's add a conformance to `CustomStringConvertible` on the enum. That will be much more type safe, and logically safe as we would use a switch and those are compiler-checked for completeness. 

I see that's actually already added [here](https://github.com/getsentry/sentry-cocoa/pull/4495/files#diff-2659ecdf6eb3bd69676310e24f3344f77e2fa86765c77c6b6dd92ca22544f315R172), so I'd recommend just moving that to here, and using a switch instead of array indexing.

If for some reason you still need the array of all cases, you can also have it conform to `CaseIterable`.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -158,5 +162,20 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
         self.unmaskedViewClasses = ((dictionary["unmaskedViewClasses"] as? NSArray) ?? []).compactMap({ element in
             NSClassFromString((element as? String) ?? "")
         })
+        if let quality = (dictionary["quality"] as? String) {
+            self.quality = SentryReplayQuality.fromName(quality)
+        }
```
brustolin, line 167: m: This is only for hybrid SDKs, we dont need the string to int conversion. 
Let's also use int in the dictionary. Use a Int "enum" in the RN side too and all good.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -2,31 +2,35 @@ import Foundation
...
         /**
          * Video Scale: 100%
          * Bit Rate: 60.000
          */
-        case high
+        case high = 2
     }
-    
+
+    static let defaultQuality: SentryReplayQuality = .medium
```
brustolin, line 32: I dont think we need to expose this.
```suggestion
    private static let defaultQuality: SentryReplayQuality = .medium
```

CHANGELOG.md
krystofwoldrich: It's internal, only for hybrids, so yes, we could omit it. But it's a bit easier to see when this was added for the hybrids SDKs dev.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
krystofwoldrich: Thanks, I've copied `SentryLevel` which uses the array, I'll update it.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
krystofwoldrich: We serialize the data to pass them to native and we can't use the enum in JS/Dart...

And the number without the enum is not clear.

brustolin: > And the number without the enum is not clear

Does it need to be clear? Should this dictionary be readable or it's just a way to pass information from RN to native?

What is the data type of this property in the RN options? Can you create a converter from that data type to int and pass the int?

krystofwoldrich: It will be a string `low`, `medium` and `high`. 

Technically the dict can be a number, it doesn't have to be readable when passing thru layers, only serializable.

---

By having the conversion here all the hybrids could use it.

krystofwoldrich: I can split the dict and the string to enum int conversion.

https://github.com/getsentry/sentry-cocoa/pull/4495/commits/6df0f60f0f9d2b5c651e1897fae1823c9a879ec2

brustolin: Using enum (and by extend numbers) is less prone to errors. Also there is the performance aspect of conversions.

I would suggest you to not use string in the RN property, JS dont have enum but you could use some const int values to minimize user mistakes. 

But I will leave it up to you. We can keep this as it is.

brustolin (approved):
LGTM

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
krystofwoldrich: Yeah, sadly JS/TS enums are bad. Can const int is a solution, but hard to discover for JS devs, but that's out of scope for this PR. 

I was mainly interested in the `why numbers`. Thank you for the details.

PR #4496, opened 2024-10-31
ref: Stops session replay if rate limiting is activated

brustolin:
## :scroll: Description

Stop session replay if the app is being rate limited.

https://github.com/getsentry/team-replay/issues/496

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryReachability.m
+++ b/Sources/Sentry/SentryReachability.m
@@ -135,7 +135,7 @@ SentryConnectivityActualCallback(
 {
     SENTRY_LOG_DEBUG(
         @"SentryConnectivityCallback called with target: %@; flags: %u", target, flags);
-#    if defined(TEST) || defined(TESTCI) || defined(DEBUG)
+#    if defined(TEST) || defined(TESTCI) | wai | defined(DEBUG)
     if (sentry_reachability_ignore_actual_callback) {
         SENTRY_LOG_DEBUG(@"Ignoring actual callback.");
         return;
--- a/Sources/Sentry/SentryReachability.m
+++ b/Sources/Sentry/SentryReachability.m
@@ -135,7 +135,7 @@ SentryConnectivityActualCallback(
 {
     SENTRY_LOG_DEBUG(
         @"SentryConnectivityCallback called with target: %@; flags: %u", target, flags);
-#    if defined(TEST) || defined(TESTCI) | wai | defined(DEBUG)
+#    if defined(TEST) || defined(TESTCI) || defined(DEBUG)
     if (sentry_reachability_ignore_actual_callback) {
         SENTRY_LOG_DEBUG(@"Ignoring actual callback.");
         return;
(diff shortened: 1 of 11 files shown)


Sources/Sentry/SentryReachability.m
```diff
@@ -135,7 +135,7 @@
 {
     SENTRY_LOG_DEBUG(
         @"SentryConnectivityCallback called with target: %@; flags: %u", target, flags);
-#    if defined(TEST) || defined(TESTCI) || defined(DEBUG)
+#    if defined(TEST) || defined(TESTCI) | wai | defined(DEBUG)
```
philipphofmann, line 138: `m`: `wai`? A leftover that should be ditched?

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -46,6 +47,8 @@ @implementation SentrySessionReplayIntegration {
     SentryReplayOptions *_replayOptions;
     SentryNSNotificationCenterWrapper *_notificationCenter;
     SentryOnDemandReplay *_resumeReplayMaker;
+    id<SentryRateLimits> _rateLimits;
+    BOOL _rateLimited;
```
philipphofmann, line 51: `h`: What stops us from checking the rate limit for every session replay we start? Checking the rate active limits is cheap. This flag can quickly get out of sync with the rate limit, leading to potentially dropping data even if we don't need to.

philipphofmann (commented):
I would rather check the active rate limits than having a flag.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: > This flag can quickly get out of sync with the rate limit

This is the intention. Once the app was rate limited, we should not start the session replay again for the entire session.

Sources/Sentry/SentryReachability.m
brustolin:
```suggestion
#    if defined(TEST) || defined(TESTCI) || defined(DEBUG)
```

Sources/Sentry/SentrySessionReplayIntegration.m
philipphofmann: If we always limit this to an entire session, we lose flexibility. Why can't the server simply send a long enough timeout? Why does it have to be an entire session?

brustolin: Right now, this is the easiest way to ensure segment 0 will always reach the server, because session replay absolutely needs segment 0 to make replay work. 

We could add more steps to check whether the envelope carrying the segment 0 was delivered, but it seems an overkill for an edge case (rate limiting).

philipphofmann: Please add this essential info as a code comment somewhere, because the future me will forget and get confused again.

philipphofmann (approved):
LGTM, when adding a comment why we rate limit until the next session.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin, line 51:
```suggestion
   // We need to use this variable to identify whether rate limiting was ever activated for session replay in this session, instead of always looking for the rate status in `SentryRateLimits`
   // This is the easiest way to ensure segment 0 will always reach the server, because session replay absolutely needs segment 0 to make replay work.
    BOOL _rateLimited;
```

PR #4512, opened 2024-11-07
fix: Concurrency crash for Swift 6

### LINKED ISSUE #4509: Crash caused by inclusion of `beforeSend` option.
### Platform

iOS

### Environment

TestFlight, Production, Develop

### Installed

Swift Package Manager

### Version

8.36.0

### Xcode Version

16.1

No

### Steps to Reproduce

1. Include a `beforeSend` block in SentrySDK initialisation.
2. Invoke a crash in the application.
3. Attempt to relaunch the application

It's worth calling out that this crash started appearing with Xcode 16.0 & Xcode 16.1 builds. Builds with Xcode 15 haven't shown this issue.
I've also tried updating Sentry to v8.39.0, but this doesn't resolve the issue.

Sample project for reproduction attached. 
This crash occurs intermittently in the supplied sample project, but ever time in our main project. 

[BeforeSendCrash.zip](https://github.com/user-attachments/files/17635843/BeforeSendCrash.zip)

### Expected Result

Application relaunches as usual and report is sent to Sentry.

### Actual Result

Application crashes on first attempted relaunch.
Application launches successfully on second attempt, and sends two crash reports to Sentry - the original crash and the SentrySDK crash.

Image
Image

https://github.com/user-attachments/assets/465e3f5f-065b-4566-b626-cceda34cfbd4

No

brustolin:
## :scroll: Description

Swift 6 introduce some concurrency checks to avoid race conditions, but objc does not have this features.
We need to annotate of callbacks so swift knows how it will be used.

Close #4509 

Samples and UITest

## :pencil: Checklist


brustolin:
>  This looks great. I'm only waiting for CI to be green. Thanks for adding a sample

Ci is drunk today.

Samples/iOS-Swift/iOS-Swift6/ViewController.swift
```diff
@@ -0,0 +1,10 @@
+import Sentry
+import UIKit
+
+class ViewController: UIViewController {
```
philipphofmann, line 4: Can we change the sample to SwiftUI? I guess most people using Swift 6 will also use SwiftUI.

Samples/iOS-Swift/iOS-Swift6UITests/UITests.swift
```diff
@@ -0,0 +1,15 @@
+import XCTest
+
+final class UITests: XCTestCase {
```
philipphofmann, line 3: `m`: As this can lead to crashes we could add this to UI tests critical and migrate this test to maestro.

philipphofmann (approved):
Thanks for adding a sample. Maybe it makes sense to use SwiftUI; see comment. LGTM.

Samples/iOS-Swift/iOS-Swift6UITests/UITests.swift
brustolin: We're using Mastro for tests that require the app to crash and be reopened.
This is not the case. If the app closes in this test, it's because something went wrong.
I prefer to avoid an external tool (Maestro in this case) as much as possible.

Samples/iOS-Swift/iOS-Swift6/ViewController.swift
philipphofmann: SwiftUI would have been less code 😄

Samples/iOS-Swift/iOS-Swift6UITests/UITests.swift
philipphofmann: Also fine.

PR #4518, opened 2024-11-07
fix: Keep PropagationContext when cloning scope

philipphofmann:
## :scroll: Description

When cloning the scope with initWithScope, the SDK wrongly created a new propagation context instead of cloning it. This is fixe now.

I found this bug in https://github.com/getsentry/sentry-cocoa/pull/4504. This aligns with the behavior from Java: https://github.com/getsentry/sentry-java/blob/e039872d80a9a352d10e642ff630a3d905ff89d0/sentry/src/main/java/io/sentry/Scope.java#L154

Unit test.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryScope.m
+++ b/Sources/Sentry/SentryScope.m
@@ -97,7 +97,7 @@ NS_ASSUME_NONNULL_BEGIN
         [_fingerprintArray addObjectsFromArray:[scope fingerprints]];
         [_attachmentArray addObjectsFromArray:[scope attachments]];
 
-        self.propagationContext = [[SentryPropagationContext alloc] init];
+        self.propagationContext = scope.propagationContext;
         self.maxBreadcrumbs = scope.maxBreadcrumbs;
         self.userObject = scope.userObject.copy;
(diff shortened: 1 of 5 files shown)


Sources/Sentry/SentryScope.m
```diff
@@ -97,7 +97,7 @@ - (instancetype)initWithScope:(SentryScope *)scope
         [_fingerprintArray addObjectsFromArray:[scope fingerprints]];
         [_attachmentArray addObjectsFromArray:[scope attachments]];
 
-        self.propagationContext = [[SentryPropagationContext alloc] init];
+        self.propagationContext = scope.propagationContext;
```
brustolin, line 100: l: Wondering if we should keep the same pointer reference or also copy `propagationContext`.

brustolin (approved):
LGTM

Sources/Sentry/SentryScope.m
philipphofmann: Good point, I made SentryPropagationContext readonly instead, so it doesn't matter.

Issue #4513, opened 2024-11-07
Finish TTFD span when new transaction bound to scope


philipphofmann:
### Description

When a user doesn't call `reportFullyDisplayed` for UIViewControllerA and the SDK binds a new auto-generated transaction to the scope for UIViewControllerB, the transaction for UIViewControllerA times out with 30 seconds. Instead, the SDK should finish the transaction for UIVIewControllerA when a new transaction is bound to the scope, because `reportFullyDisplayed` will only impact UIViewControllerB and not anymore UIViewControllerA.

Current behaviour

1. Open ViewControllerA. Scope: Transaction A
2. Navigate to ViewControllerB. Scope:Transaction A

Expected behaviour 

1. Open ViewControllerA. Scope: Transaction A
2. Navigate to ViewControllerB. Scope:Transaction B

PR #4519, opened 2024-11-08
fix: Make Scope.span fully thread safe

philipphofmann:
## :scroll: Description

Setting the span on the scope and useSpan were thread safe, but not setSpan. This is fixed now.

This came up while investigating #4513.

Unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryScope.m
+++ b/Sources/Sentry/SentryScope.m
@@ -59,6 +59,8 @@ NS_ASSUME_NONNULL_BEGIN
     NSObject *_spanLock;
 }
 
+@synthesize span = _span;
+
 #pragma mark Initializer
 
 - (instancetype)initWithMaxBreadcrumbs:(NSInteger)maxBreadcrumbs
@@ -146,12 +148,16 @@ NS_ASSUME_NONNULL_BEGIN
     }
 }
 
+- (nullable id<SentrySpan>)span
+{
+    @synchronized(_spanLock) {
+        return _span;
+    }
+}
+
 - (void)useSpan:(SentrySpanCallback)callback
 {
-    id<SentrySpan> localSpan = nil;
-    @synchronized(_spanLock) {
-        localSpan = _span;
-    }
+    id<SentrySpan> localSpan = [self span];
     callback(localSpan);
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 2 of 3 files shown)


Sources/Sentry/SentryScope.m
```diff
@@ -59,6 +59,8 @@ @implementation SentryScope {
     NSObject *_spanLock;
 }
 
+@synthesize span = _span;
```
brustolin, line 62: m: Actually, we don’t need any of the changes about span in the `.m` file, just need to change the span property declaration from `nonatomic` to `atomic`, same result.

eg: `@property (nullable, atomic, strong) id<SentrySpan> span;`

P.S.: You don’t need this `@synthesize` if you are using the standard name declaration (_ in front of the name) like this one.

brustolin (approved):
LGTM

Sources/Sentry/SentryScope.m
brustolin: Not totally true, we could remove the existing `setSpan` function.

brustolin:
Also I dont think we need useSpan anymore, we should mark it as deprecated.

philipphofmann:
> Also I dont think we need useSpan anymore, we should mark it as deprecated.

We still use `useSpan` in multiple places in the SDK. We can consider removing it in the next version.

Sources/Sentry/SentryScope.m
philipphofmann: When I remove the `synthesize` I get:  `Use of undeclared identifier '_span'; did you mean 'span'?` I'm a bit confused.

`atomic` is a good point, but I would like to include the scope observer call in the same lock as setting the span to ensure the observers and the scope are in sync. That doesn't work correctly when using atomic.
```objc
@synchronized(_spanLock) {
    _span = span;

    for (id<SentryScopeObserver> observer in self.observers) {
        [observer setTraceContext:[self buildTraceContext:span]];
    }
}
```

brustolin: > When I remove the synthesize I get:  Use of undeclared identifier '_span'; did you mean 'span'? I'm a bit confused.

I mean the entire line.
```@synthesize span = _span;```

> but I would like to include the scope observer call in the same lock as setting the span

True. Thanks!!

philipphofmann: > I mean the entire line.
> @synthesize span = _span;

Yes, I removed the entire line.

PR #4521, opened 2024-11-08
Chore: Fix certificate for iOS-Swift sample

brustolin:
Fixing the certificate for the sample that we use for benchmark and TestFlight.

_#skip-changelog_


Samples/iOS-Swift/iOS-Swift.xcodeproj/project.pbxproj
```diff
@@ -1430,6 +1432,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = "io.sentry.sample.iOS-Swift";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE_SPECIFIER = "match Development io.sentry.sample.iOS-Swift";
+				"PROVISIONING_PROFILE_SPECIFIER[sdk=iphoneos*]" = "match AppStore io.sentry.sample.iOS-Swift";
```
krystofwoldrich, line 1435: Are these extra values needed with the new cert and provisioning profile?

brustolin: It is the way Xcode 16 handled it. Better than manually changing the .xcodeproj file

philipphofmann (approved):
Thanks 🚀

PR #4522, opened 2024-11-11
internal(replay): Add IgnoreWrapper class for hybrid SDKs

krystofwoldrich:
## :scroll: Description

This PR adds `IgnoreWrapper` functionality to Replay Redaction logic.

Why is this needed? In RN JS we don't have access to the native view instances directly. So we create a native wrapper class which we can use from JS.

- https://github.com/getsentry/sentry-react-native/pull/4224

rn sample app, unit and integration tests

## :pencil: Checklist

#skip-changelog

### DIFF
--- a/Sources/Swift/Tools/UIRedactBuilder.swift
+++ b/Sources/Swift/Tools/UIRedactBuilder.swift
@@ -46,7 +46,8 @@ struct RedactRegion {
 }
 
 class UIRedactBuilder {
-    
+    ///This is a wrapper which marks it's direct children to be redacted
+    private var ignoreWrapperClassIdentifier: ObjectIdentifier?
     ///This is a list of UIView subclasses that will be ignored during redact process
     private var ignoreClassesIdentifiers: Set<ObjectIdentifier>
     ///This is a list of UIView subclasses that need to be redacted from screenshot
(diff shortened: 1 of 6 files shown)


brustolin (commented):
Just a few code improvement suggestions: In Swift, it’s better to use `guard` for simple early return, since the compiler will enforce the early return.

Apart from that, it’s all good.

krystofwoldrich:
@brustolin Ready for a second pass. I fixed the guards and added RedactContainer. 

To fix the issue where RedactContainer would be overwritten by IgnoreContainer in the following setup IgnoreContainer -> RedactContainer -> 'This should be redacted'

Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -178,9 +202,28 @@ class UIRedactBuilder {
...
     }
-    
+
+    private func shouldIgnoreParentContainer(_ view: UIView) -> Bool {
+        if isRedactContainerClass(type(of: view)) {
+            return false
+        }
+
+        guard let parent = view.superview  else { return false }
+        return isIgnoreContainerClass(type(of: parent))
```
brustolin, line 214: m: Try to use `guard` as much as possible for early return.

```suggestion
        guard !isRedactContainerClass(type(of: view)),  let parent = view.superview  else { return false }
        return isIgnoreContainerClass(type(of: parent))
```

brustolin (approved):
Thanks for renaming the API

PR #4526, opened 2024-11-12
fix: Finish TTFD when binding transaction to scope

philipphofmann:
## :scroll: Description

Finish the TTFD span before creating a new transaction and binding it to the scope in SentryUIViewControllerPerformanceTracker when the user doesn't call reportFullyDisplayed.

Fixes GH-4513

Unit tests and simulator.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryTimeToDisplayTracker.m
+++ b/Sources/Sentry/SentryTimeToDisplayTracker.m
@@ -146,7 +146,8 @@
     if (self.fullDisplaySpan.isFinished == NO) {
         SENTRY_LOG_WARN(
             @"You didn't call SentrySDK.reportFullyDisplayed(). Finishing full display span with "
-            @"status deadline exceeded.");
+            @"status: %@.",
+            nameForSentrySpanStatus(kSentrySpanStatusDeadlineExceeded));
 
         [self.fullDisplaySpan finishWithStatus:kSentrySpanStatusDeadlineExceeded];
     }
--- a/Sources/Sentry/SentryTimeToDisplayTracker.m
+++ b/Sources/Sentry/SentryTimeToDisplayTracker.m
@@ -149,7 +149,6 @@
             @"status deadline exceeded.");
 
         [self.fullDisplaySpan finishWithStatus:kSentrySpanStatusDeadlineExceeded];
-        return;
     }
 }
 
--- a/Sources/Sentry/SentryTimeToDisplayTracker.m
+++ b/Sources/Sentry/SentryTimeToDisplayTracker.m
@@ -144,10 +144,9 @@
     }
 
     if (self.fullDisplaySpan.isFinished == NO) {
-        SENTRY_LOG_WARN(
-            @"You didn't call SentrySDK.reportFullyDisplayed(). Finishing full display span with "
-            @"status: %@.",
-            nameForSentrySpanStatus(kSentrySpanStatusDeadlineExceeded));
+        SENTRY_LOG_WARN(@"You didn't call SentrySDK.reportFullyDisplayed() for UIViewController: "
+                        @"%@. Finishing full display span with status: %@.",
+            _controllerName, nameForSentrySpanStatus(kSentrySpanStatusDeadlineExceeded));
 
         [self.fullDisplaySpan finishWithStatus:kSentrySpanStatusDeadlineExceeded];
     }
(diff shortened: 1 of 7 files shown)


brustolin (approved):
LGTM

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -137,6 +137,22 @@ - (void)reportFullyDisplayed
...
+    if (self.initialDisplaySpan.isFinished == NO) {
+        [self.initialDisplaySpan finish];
+    }
+
+    if (self.fullDisplaySpan.isFinished == NO) {
+        SENTRY_LOG_WARN(
+            @"You didn't call SentrySDK.reportFullyDisplayed(). Finishing full display span with "
+            @"status deadline exceeded.");
+
+        [self.fullDisplaySpan finishWithStatus:kSentrySpanStatusDeadlineExceeded];
```
brustolin, line 151: m: I just dont know if this is a good status for this case.

philipphofmann: I also thought about that. It's the same as when you forget to call reportFullyDisplayed, and the 30s timeout hits. I wanted to keep it consistend.

brustolin:
BTW, I just realize a problem that goes like this:

- ViewControllerA opens, start a http request request, user clicks a button and open ViewControllerB.
- ViewControllerB starts a http request
- ViewControllerA http request ends and `SentrySDK.reportFullyDisplayed` is called


This will finish TTFD for ViewControllerB

philipphofmann:
> BTW, I just realized a problem that goes like this:
> 
> * ViewControllerA opens, starts a HTTP request, user clicks a button and opens ViewControllerB.
> * ViewControllerB starts an HTTP request.
> * ViewControllerA’s HTTP request ends and `SentrySDK.reportFullyDisplayed` is called.
> 
> This will finish TTFD for ViewControllerB.

Do you think this PR adds this problem or that the problem still exists and needs to be addressed @brustolin?

brustolin:
>Do you think this PR adds this problem or that the problem still exists and needs to be addressed @brustolin?

I think this PR adds this problem. Of course, by fixing another problem, we need to choose.

> Why do you think the ViewControllerA’s HTTP request ending does call SentrySDK.reportFullyDisplayed? Do you mean the user calls it? 

The user would call it in the code. 

> Before this PR calling SentrySDK.reportFullyDisplayed when being on ViewControllerB ends finishes TTFD of ViewControllerB. The same applies to this PR, but the difference now is that the TTFD of UIViewControllerA doesn't timeout any more.

🤔 I need to check If I can reproduce what Im claiming.

philipphofmann:
I'm not sure if this PR adds the problem. Yes, please try if you can reproduce it, ideally with a test.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerPerformanceTrackerTests.swift
```diff
@@ -766,6 +766,137 @@ class SentryUIViewControllerPerformanceTrackerTests: XCTestCase {
...
+        }
+        
+        XCTAssertEqual(secondTracer?.traceId, SentrySDK.span?.traceId)
+        
+        XCTAssertTrue(firstFullDisplaySpan.isFinished)
+        XCTAssertEqual(.deadlineExceeded, firstFullDisplaySpan.status)
+        
+        let secondFullDisplaySpan = try XCTUnwrap(secondTracer?.children.first { $0.operation == "ui.load.full_display" })
+        
+        XCTAssertFalse(secondFullDisplaySpan.isFinished)
```
brustolin, line 814: h: I dont get this test, how could this `true` even before this PR?

brustolin:
@philipphofmann I wrote the test to reproduce what Im talking about in [here](https://github.com/getsentry/sentry-cocoa/pull/4531)

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerPerformanceTrackerTests.swift
brustolin: > the secondFullDisplaySpan didn't even exist

Ok, this is the problem. Got it.

brustolin (approved):
LGTM

Issue #4272, opened 2024-08-09
User Feedback envelope (Due: Dec 13)


armcknight:
### Description

Package up the inputs from the form into the envelope format for transmission to the Sentry backend.

Old format ~Develop docs on user feedback envelope items: https://develop.sentry.dev/sdk/data-model/envelope-items/#user-feedback~
New format: https://develop.sentry.dev/application/feedback-architecture/#feedback-events

Reference implementation: https://github.com/getsentry/sentry-javascript/blob/be9edf161f72bb0b9ccf38d70297b798054b3ce3/packages/feedback/src/core/sendFeedback.ts#L77-L116

Schema of the event envelope:
 ```
     event[”contexts”][”feedback”] = {
     "name": <user-provided>,
     "contact_email": <user-provided>,
     "message": <user-provided>,
     "url": <referring web page>,
     "source": <developer-provided, ex: "widget">,
     "associated_event_id": <developer-provided, should be a valid error event in the same project>
 ```
A more complete example from the javascript reference implementation:
 ```
   {
     "type": "feedback",
     "event_id": "d2132d31b39445f1938d7e21b6bf0ec4",
     "timestamp": 1597977777.6189718,
     "dist": "1.12",
     "platform": "javascript",
     "environment": "production",
     "release": 42,
     "tags": {"transaction": "/organizations/:orgId/performance/:eventSlug/"},
     "sdk": {"name": "name", "version": "version"},
     "user": {
         "id": "123",
         "username": "user",
         "email": "user@site.com",
         "ip_address": "192.168.11.12",
     },
     "request": {
         "url": None,
         "headers": {
             "user-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.5 Safari/605.1.15"
         },
     },
     "contexts": {
         "feedback": {
             "message": "test message",
             "contact_email": "test@example.com",
             "type": "feedback",
         },
         "trace": {
             "trace_id": "4C79F60C11214EB38604F4AE0781BFB2",
             "span_id": "FA90FDEAD5F74052",
             "type": "trace",
         },
         "replay": {
             "replay_id": "e2d42047b1c5431c8cba85ee2a8ab25d",
         },
     },
   }
 ```

[...]

PR #4535, opened 2024-11-15
feat(feedback): capture envelopes

armcknight:
for https://github.com/getsentry/sentry-cocoa/issues/4272; #skip-changelog

capture the user feedback in an envelope. currently, an error is required to have been captured, with its event ID supplied in the user feedback object. so, a dummy error is currently created to satisfy this requirement to be able to see them in the dashboard (see linked issue for progress/results)

TODOs
- [x] (ETA: working!) this currently captures feedback using the old captureFeedback function. need to implement the new feedback envelope format and then capture the envelope that way: https://develop.sentry.dev/application/feedback-architecture/#feedback-events


Sources/Sentry/Public/SentryUserFeedback.h
```diff
@@ -15,14 +15,24 @@ NS_ASSUME_NONNULL_BEGIN
...
  * Initializes SentryUserFeedback and sets the required eventId.
  * @param eventId The eventId of the event to which the user feedback is associated.
+ * @note Uses the old envelope format descried at
+ * https://develop.sentry.dev/sdk/data-model/envelope-items/#user-feedback and will be deprecated in
+ * the future.
  */
 - (instancetype)initWithEventId:(SentryId *)eventId;
 
+/**
+ * Initializes a new `SentryUserFeedback` as its own event, instead of one attached to a transaction
```
philipphofmann, line 29: `l`: I think in the docs, we could refer to user feedback V0 and V1 or V1 and V2 instead of old and new. What if we again add a new version, is it then new new?

Sources/Sentry/include/SentryClient+Private.h
```diff
@@ -2,8 +2,15 @@
...
-    SentryReplayEvent, SentryReplayRecording, SentryEnvelope;
+@class SentryAttachment;
+@class SentryEnvelope;
+@class SentryEnvelopeItem;
+@class SentryFeedback;
+@class SentryId;
+@class SentryReplayEvent;
+@class SentryReplayRecording;
+@class SentrySession;
+@class SentryThreadInspector;
```
philipphofmann, line 13: Thanks 👍

Sources/Sentry/SentryEnvelope.m
```diff
@@ -129,6 +129,24 @@ - (instancetype)initWithUserFeedback:(SentryUserFeedback *)userFeedback
                            data:json];
 }
 
+- (instancetype)initWithFeedback:(SentryFeedback *)feedback
```
philipphofmann, line 132: `l`: This method is almost identical to `initWithUserFeedback`. Maybe it makes sense to extract a common method for them.

Sources/Sentry/SentryUserFeedback.m
```diff
@@ -2,7 +2,26 @@
 #import "SentrySwift.h"
 #import <Foundation/Foundation.h>
 
-@implementation SentryUserFeedback
+typedef enum : NSUInteger {
```
philipphofmann, line 5: `m`: I'm a bit confused as to why we need this as we have the Swift class SentryFeedback, or do you plan on removing this?

Sources/Sentry/include/HybridPublic/SentryEnvelope.h
```diff
@@ -7,13 +7,14 @@
...
-@class SentryId;
-@class SentryUserFeedback;
 @class SentryAttachment;
 @class SentryEnvelopeItemHeader;
+@class SentryEvent;
+@class SentryFeedback;
+@class SentryId;
+@class SentrySession;
 @class SentryTraceContext;
+@class SentryUserFeedback;
```
philipphofmann, line 17: Again thanks 💯

Sources/Sentry/include/HybridPublic/SentryEnvelopeItemType.h
```diff
@@ -3,6 +3,7 @@
 static NSString *const SentryEnvelopeItemTypeEvent = @"event";
 static NSString *const SentryEnvelopeItemTypeSession = @"session";
 static NSString *const SentryEnvelopeItemTypeUserFeedback = @"user_report";
+static NSString *const SentryEnvelopeItemTypeFeedback = @"feedback";
```
philipphofmann, line 6: `h`: You most likely also have to add a `SentryDataCategory` and an entry to `sentryDataCategoryForEnvelopItemType` for client reports to work correctly.

Sources/Sentry/include/SentrySDK+Private.h
```diff
@@ -51,6 +55,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 + (void)captureEnvelope:(SentryEnvelope *)envelope;
 
++ (void)captureFeedback:(SentryFeedback *)feedback;
```
philipphofmann, line 58: `m`: I'm a bit confused why we have this here in `SentrySDK+Private.h` and in `SentrySDK.h`.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
```diff
@@ -0,0 +1,57 @@
...
+    var email: String?
+    var message: String
+    var hints: [String: Any]?
+    var source: Source
+    let eventId: SentryId
+    
+    /// The event id that this feedback is associated with, like a crash report.
+    var associatedEventId: String?
+
+    init(name: String?, email: String?, message: String, hints: [String: Any]? = nil, source: Source = .widget, associatedEventId: String? = nil) {
```
philipphofmann, line 27: `l`: I would put the required message first and set the others to = nil per default, so the API is a bit cleaner.
```suggestion
    init(message: String, name: String? = nil, email: String? = nil, hints: [String: Any]? = nil, source: Source = .widget, associatedEventId: String? = nil) {
```

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
```diff
@@ -0,0 +1,57 @@
...
+        self.name = name
+        self.email = email
+        self.message = message
+        self.hints = hints
+        self.source = source
+        self.associatedEventId = associatedEventId
+        super.init()
+    }
+    
+    func serialize() -> [String: Any] {
```
philipphofmann, line 38: `m`: This still misses a few properties if I'm not mistaken.

philipphofmann (commented):
I know this is a draft, but I already gave you early feedback.

Sources/Sentry/SentryUserFeedback.m
armcknight: Ah, this was an earlier attempt to try to reuse the existing class. I've switched to just duplicating everything so the old stuff can be removed after the deprecation period.

Sources/Sentry/include/HybridPublic/SentryEnvelopeItemType.h
armcknight: Thanks for pointing it out 👍🏻 I'm not intimately familiar with all this machinery.

Sources/Sentry/include/SentrySDK+Private.h
armcknight: probably just me shuffling things around to get them working. will make sure to clean this up 👍🏻

Sources/Sentry/include/HybridPublic/SentryEnvelopeItemType.h
philipphofmann: That's what I'm here for 😅

Sources/Sentry/SentryEnvelope.m
armcknight: Same as https://github.com/getsentry/sentry-cocoa/pull/4535#discussion_r1847524583, at this point I just want to keep things separate so it's easier to remove the other thing later.

Sources/Sentry/Public/SentryUserFeedback.h
armcknight: A couple thoughts here. First, I agree from our earlier slack convo that using V1/V2 etc is bad practice, I've always disliked it, like with the earlier PerformanceV2 stuff. So I don't know why we'd want to do it here now.

Second, if we are already rolling out a new new user feedback paradigm (V3) on top of the old new one (V2) before even finishing deprecating/removing the original (V1), we need to step back and take a breath 😆 

We are going to roll the manual calls to capture feedback directly and/or from a custom built UI into this new paradigm, by extending the current functionality (report feedback with an attached event ID) to not _require_ the event ID, and also be able to attach screenshots. And then provide the managed UI to capture the inputs that go into the same callpath.

Do we really anticipate a completely different thing coming any time soon?

What we can do here is simply remove "old" and "new" and add a deprecation message at the appropriate time. They don't have to be "old" vs "new", they can just be "different", and "deprecated" vs "supported".

armcknight: Oh and also, I'm going to put back SentryUserFeedback and just use the new type SentryFeedback that I introduced for this separate pathway.

Sources/Sentry/include/HybridPublic/SentryEnvelopeItemType.h
armcknight: A question, while we're on the subject: what is the actual difference between e.g. `SentryEnvelopeItemTypeUserFeedback` and `kSentryDataCategoryNameUserFeedback`? It looks like we pass `SentryEnvelopeItemTypeUserFeedback` into `sentryDataCategoryForEnvelopItemType`  in at least one place: https://github.com/getsentry/sentry-cocoa/blob/37fd7a27e43b84eddf942bce1a0066d5cbc050a0/Sources/Sentry/SentryEnvelopeRateLimit.m#L57-L58 which then later spits out `kSentryDataCategoryNameUserFeedback` in the call to `nameForSentryDataCategory` here: https://github.com/getsentry/sentry-cocoa/blob/37fd7a27e43b84eddf942bce1a0066d5cbc050a0/Sources/Sentry/SentryHttpTransport.m#L167. Couldn't we deduplicate the string defs?

philipphofmann: No sadly not. Envelope item types and data categories look similar, but they have slight differences. I'm going to open a PR to add a comment. This also confused me quite a bit in the past.
CleanShot 2024-11-20 at 16 58 29@2x

philipphofmann: PR to clarify that: https://github.com/getsentry/sentry-cocoa/pull/4556

armcknight: Thanks, I tried to look for the differences but I guess I was only comparing members that were common between the two sets and missed the ones that only existed in one or the other!

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
armcknight: so the event ID is used to build the envelope header, and i'm not actually sure if/how we are going to use hints for this. I can't remember why hints were even originally included in the API. I'm just going to remove it for now.

Sources/Sentry/SentryClient.m
```diff
@@ -585,6 +584,50 @@ - (void)captureUserFeedback:(SentryUserFeedback *)userFeedback
...
+    SentryEvent *preparedEvent = [self prepareEvent:feedbackEvent
+                                          withScope:scope
+                             alwaysAttachStacktrace:NO];
+    SentryTraceContext *traceContext = [self getTraceStateWithEvent:preparedEvent withScope:scope];
+    NSArray *attachments = [[self attachmentsForEvent:preparedEvent scope:scope]
+        arrayByAddingObjectsFromArray:[feedback attachments]];
+    [self.transportAdapter sendEvent:preparedEvent
+                        traceContext:traceContext
+                         attachments:attachments
+             additionalEnvelopeItems:@[]];
```
armcknight, line 628: This is basically copied from the implementation for `-[sendEvent:withScope:alwaysAttachStacktrace:isCrashEvent:additionalEnvelopeItems:]`. we can't simply call through to that because the feedback screenshot must be send as an envelope attachment, but there are no other methods in this class that allow specifying attachments that aren't already on the scope, and i didn't want to attach the screenshot to the scope in case another event would also pick it up. the reference implementation does show a trace context so i thought it best to keep it here too https://github.com/getsentry/sentry-javascript/blob/be9edf161f72bb0b9ccf38d70297b798054b3ce3/packages/feedback/src/core/sendFeedback.ts#L107. 

it also shows a replay context, which i'm not sure how to retrieve yet in this sdk, any thoughts on that @brustolin ?

armcknight:
The lint failures are bubbling up from an upstream PR: https://github.com/getsentry/sentry-cocoa/pull/4630#issuecomment-2540323633

Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+@testable import Sentry
+import XCTest
+
+class SentryFeedbackTests: XCTestCase {
+    func testSerialize() throws {
```
armcknight, line 6: So I have this test, but I'm looking for the best place to write another test (an examples of anything that might have previously been done similarly) that will inspect the actual envelope contents when calling `SentrySDK.captureFeedback`. I see a lot of tests that use `Invocations`, but that is not what I'm talking about. I want to make assertions on the actual contents of the envelope JSON. Anything like that?

Sources/Sentry/SentryClient.m
brustolin: current replay can be retrieved from `scope.replayId`.

What about using SentryClient.attachmentProcessors in the user feedback integration in a similar way that Screenshot integration does?

Sources/Sentry/SentryClient.m
```diff
@@ -585,6 +584,30 @@ - (void)captureUserFeedback:(SentryUserFeedback *)userFeedback
...
+{
+    SentryEvent *feedbackEvent = [[SentryEvent alloc] init];
+    feedbackEvent.eventId = feedback.eventId;
+    feedbackEvent.type = SentryEnvelopeItemTypeFeedback;
+
+    NSDictionary *serializedFeedback = [feedback serialize];
+
+    NSMutableDictionary *context = [NSMutableDictionary dictionary];
+    context[@"feedback"] = serializedFeedback;
+    feedbackEvent.context = context;
```
philipphofmann, line 597: `l`: A bit shorter and requires memory if I'm not mistaken.
```suggestion
    feedbackEvent.context = @{ @"feedback": serializedFeedback };
```

Sources/Sentry/SentryClient.m
```diff
@@ -585,6 +584,30 @@ - (void)captureUserFeedback:(SentryUserFeedback *)userFeedback
     [self.transportAdapter sendUserFeedback:userFeedback];
 }
 
+- (void)captureFeedback:(SentryFeedback *)feedback withScope:(SentryScope *)scope
```
philipphofmann, line 587: `m`: It would be great to add some test(s) to `SentryClientTests` for this new method.

Sources/Sentry/SentryDataCategoryMapper.m
```diff
@@ -18,6 +18,7 @@
 NSString *const kSentryDataCategoryNameReplay = @"replay";
 NSString *const kSentryDataCategoryNameMetricBucket = @"metric_bucket";
 NSString *const kSentryDataCategoryNameSpan = @"span";
+NSString *const kSentryDataCategoryNameFeedback = @"feedback";
```
philipphofmann, line 21: `m`: Please add or update the tests for this new data category.

Sources/Sentry/SentryEnvelope.m
```diff
@@ -95,6 +95,8 @@ - (instancetype)initWithEvent:(SentryEvent *)event
     // default. In any case in the envelope type it should be event. Except for transactions
     NSString *envelopeType = [event.type isEqualToString:SentryEnvelopeItemTypeTransaction]
         ? SentryEnvelopeItemTypeTransaction
+        : [event.type isEqualToString:SentryEnvelopeItemTypeFeedback]
```
philipphofmann, line 98: `m`: Again a test would be nice.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
```diff
@@ -0,0 +1,63 @@
...
+        }
+        dict["source"] = source.rawValue
+        
+        return dict
+    }
+    
+    /**
+     * - note: Currently there is only a single attachment possible, for the screenshot, of which there can be only one.
+     */
+    func attachments() -> [Attachment] {
```
philipphofmann, line 56: `l`: I think renaming this to something like `convertScreenshotToAttachment` would communicate the info in the comment a bit better. Now, when only looking at the code in the SentryClient I could assume that the SentryFeedback has multiple attachments. Only when I read this comment or jump to this code I understand what's going on.

philipphofmann (commented):
A few comments.

Sources/Sentry/SentryEnvelope.m
armcknight: yeah this is part of why i want to validate the envelope contents

Sources/Sentry/SentryClient.m
armcknight: What is the value of all the tests that only validate whether a method was called, using Invocations? if i write a test that validates the contents of the envelope, it implicitly validates this took place. I assume they have provided unique value to you in the past, I just can't imagine what it is so would appreciate any explanation.

Sources/Sentry/SentryClient.m
armcknight: This will still have to allocate memory, but, I can use `-[NSMutableDictionary dictionaryWithCapacity:]` to prevent reallocation and moving. I also as a rule will never use dictionary literals in production implementation again due to past experience with crashes (tests are fine).

Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
armcknight: Yes, this sounds like a good path. Thanks!

Sources/Sentry/SentryClient.m
armcknight: beforeSend is a good candidate for this. I think envelope content validation can check for stacktraces, screenshots, and viewHierarchy in addition to the core things like the feedback message, name and email.

Sources/Sentry/SentryClient.m
armcknight: I don't love having to store something in the global scope just to access it from the client later. The client has the feedback object right here that it can grab the attachments from.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -132,12 +133,14 @@ class SentryUserFeedbackForm: UIViewController {
             present(alert, animated: config.animations)
             return
         }
-        
-        delegate?.confirmed()
+
+        let feedback = SentryFeedback(message: messageTextView.text, name: fullNameTextField.text, email: emailTextField.text, screenshot: screenshotImageView.image?.pngData())
+        SentrySDK.capture(feedback: feedback)
+        delegate?.finished()
```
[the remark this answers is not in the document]
armcknight: Yes, I do want to do this, along the same lines of the envelope inspection and what I did for profiling.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
armcknight: What difference does it make what kind of attachments it appends?

Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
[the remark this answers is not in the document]
philipphofmann: I'm a bit confused. This comment is marked as resolved, but I don't see an extra test for this or a comment explaining why the above described test doesn't make sense, @armcknight.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
philipphofmann: As I mentioned above, when reading the SentryClient code, I was a bit confused that the feedback can have multiple attachments, so I jumped here and saw that it's one attachment. Renaming the method to something I proposed above, could make this a bit easier to understand. But we can ignore this. That's nitpicking. Let's focus on the more important things. It's not even worth the time I write this comment for discussing such small things 😅

Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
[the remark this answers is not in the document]
armcknight: I haven't pushed the commit yet.

armcknight: I am currently taking the request manager approach, which has a double benefit of not spamming the backend with every envelope sent when running these tests.

As a point of clarification, I would not call these end-to-end tests, but at most, integration tests. An E2E test IMO would exercise a mobile app with the SDK, and then validate on our web app that the right thing showed up as a customer would see it, by inspecting the DOM.

armcknight: see https://github.com/getsentry/sentry-cocoa/pull/4652

Sources/Sentry/SentryClient.m
armcknight: the hooks actually still aren't implemented, so i will defer testing those to a later PR.

armcknight:
the remaining failures with the static pod lint steps are fixed in https://github.com/getsentry/sentry-cocoa/pull/4659

Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
```diff
@@ -134,6 +137,21 @@ class UserFeedbackUITests: BaseUITest {
         XCTAssertEqual(try XCTUnwrap(emailField.value as? String), "your.email@example.org")
         
         XCTAssertEqual(try XCTUnwrap(messageTextView.value as? String), "", "The UITextView shouldn't have any initial text functioning as a placeholder; as UITextView has no placeholder property, the \"placeholder\" is a label on top of it.")
+        
+        cancelButton.tap()
+        
+        app.buttons["Extra"].tap()
+        app.buttons["io.sentry.ui-test.button.get-latest-envelope"].tap()
+        let marshaledDataBase64 = try XCTUnwrap(app.textFields["io.sentry.ui-test.text-field.data-marshaling.latest-envelope"].value as? String)
```
philipphofmann, line 145: Thanks 👍

Sources/Sentry/SentryQueueableRequestManager.m
```diff
@@ -27,6 +27,20 @@ - (instancetype)initWithSession:(NSURLSession *)session
 
 - (BOOL)isReady
 {
+#if TEST || TESTCI
```
philipphofmann, line 30: `m`: I would prefer a different type of implementation of the protocol `SentryRequestManager` that implements the behavior we need for testing. I think this class shouldn't know about this test functionality.

Sources/Sentry/SentryEnvelope.m
```diff
@@ -161,7 +161,17 @@ - (_Nullable instancetype)initWithAttachment:(SentryAttachment *)attachment
             return nil;
         }
 
+#if DEBUG || TEST || TESTCI
+        if ([NSProcessInfo.processInfo.arguments
+                containsObject:@"--io.sentry.base64-attachment-data"]) {
+            data = [[attachment.data base64EncodedStringWithOptions:0]
+                dataUsingEncoding:NSUTF8StringEncoding];
```
philipphofmann, line 168: `m`: It feels quite strange to modify the envelope behavior based on process info arguments for testing. I don't like it when you flip a flag somewhere, and it changes the behavior deep down in the code. This is a sign that something is wrong with the SDK setup. It should be easier to change the behavior required for testing. I don't have a proper solution at hand right now, but maybe I could come up with something after looking at this PR more closely.

philipphofmann (approved):
LGTM, but the testing approach seems quite strange to me. I don't want this to block this PR though.

Sources/Sentry/SentryEnvelope.m
armcknight: Definitely open to other strategies. The only other ways I know of to change SDK behavior are using SentryOptions, so then we'd have to add some new properties to its private interface, and then we'd still have to use a launch arg to tell the app to use that, although I agree it'd be better to change behavior in the app than in the sdk using an app launch arg.

Another idea is to have a network integration API similar to the "feature integration" API we already have. This would by default have our current request manager stack, and then we could provide spotlight as another prepackaged option that can be used in addition or instead of the default. Then we could start writing test stacks that could be inserted.

Honestly that would be my preference. Then we could modularize out the spotlight thing if we ever get to splitting up the SDK into nice little packages. But the goal with the experimental phase of the user feedback dev is to get something working quick and dirty so we can show it to customers to get feedback on it, and refactor the implementation later.

philipphofmann: Making the transport public, as other SDKs have it, would do the trick. Yep, we can refactor that later, as it's not the feature's fault that we don't have this in the SDK yet.

PR #4538, opened 2024-11-15
Chore: Move SentryUserFeedbackIntegration initialization.

brustolin:
Integrations should be initialized in the `defaultIntegrations` property of the options.

Is there any reason to handle this differently for UserFeedback, @armcknight?

_#skip-changelog_


armcknight:
Hmm, the reason I put this here is that I saw `defaultIntegrations` (which calls `defaultIntegrationClasses`) called in SentryOptions init, and assumed that was the only time that list would be accessed, which is too early to determine if user feedback is configured or not. But I see now that `defaultIntegrationClasses` is called again in SentrySDK.startWithOptions, which will have the options configured by then.

In what case would the version of `defaultIntegrationClasses` be used that is accessed here instead of the later result? Is it always overwritten, or just in some cases? 

https://github.com/getsentry/sentry-cocoa/blob/522f8da1aef040a58ae580345895e2ec1c215d36/Sources/Sentry/SentryOptions.m#L100

vs 

https://github.com/getsentry/sentry-cocoa/blob/9e94954fac5b4dfcc19f721c3a44d83e643f0475/Sources/Sentry/SentrySDK.m#L467

philipphofmann (approved):
LGTM, as this aligns with how we enable all the other integrations. I know it's confusing, but this PR makes it consistent. I'm just a bit surprise we need to changes in the tests for that.

PR #4540, opened 2024-11-15
ci: Fix failing macOS-SwiftUI build

philipphofmann:
Use macos-15 for Xcode 16, cause Xcode 16 is not available on macos-14.

#skip-changelog


brustolin (approved):

lucas-zimerman:
oh so thats why sentry cordova was failing to build

PR #4541, opened 2024-11-15
Fix: Load integration from same binary

brustolin:
## :scroll: Description

We use the class name to load integrations, and this could lead to a wrong integration being loaded if sentry is duplicated inside a project.

## :pencil: Checklist


armcknight (commented):
Thanks for removing the string-based logic, that has bugged me for a long time. I have a comment on how the method is being declared.

Sources/Sentry/include/SentryOptions+Private.h
```diff
@@ -32,6 +32,8 @@ FOUNDATION_EXPORT NSString *const kSentryDefaultEnvironment;
 @property (nonatomic, copy, nullable)
     SentryUserFeedbackConfigurationBlock configureUserFeedback API_AVAILABLE(ios(13.0));
 
+@property (nonatomic, readonly, class) NSArray<Class> *defaultIntegrationClasses;
```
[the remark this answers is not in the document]
brustolin: Thanks for the suggestions. This boils down to preference. I prefer dot notation over square brackets calls.

I know that a C function has better performance than an ObjC function, but the difference is negligible and ultimately we had an ObjC wrapping the C function, so it really does not matter.

Sources/Sentry/SentrySDK.m
```diff
@@ -464,9 +464,19 @@ + (void)installIntegrations
     NSMutableArray<NSString *> *integrationNames =
         [SentrySDK.currentHub getClient].options.integrations.mutableCopy;
 
+    NSArray<Class> *defaultIntegrations = SentryOptions.defaultIntegrationClasses;
+    NSMutableDictionary<NSString *, Class> *integrationDictionary =
+        [[NSMutableDictionary alloc] init];
+
+    for (Class integrationClass in defaultIntegrations) {
+        integrationDictionary[NSStringFromClass(integrationClass)] = integrationClass;
```
philipphofmann, line 472: `h`: I don't get it. The code only writes to the dictionary but never reads from it. How does it solve the underlying problem?

Sources/Sentry/SentrySDK.m
```diff
@@ -464,9 +464,19 @@ + (void)installIntegrations
     NSMutableArray<NSString *> *integrationNames =
         [SentrySDK.currentHub getClient].options.integrations.mutableCopy;
 
+    NSArray<Class> *defaultIntegrations = SentryOptions.defaultIntegrationClasses;
```
philipphofmann, line 467: `m`: It would be good to add a comment somewhere on why we interact with the classes directly and require this dictionary. That isn't obvious.

philipphofmann (commented):
It would be great to have a test failing if we change the code in the SentrySDK.m back to the old implementation.

Sources/Sentry/SentrySDK.m
brustolin: 🤔
Strange, that is something missing here

brustolin:
> It would be great to have a test failing if we change the code in the SentrySDK.m back to the old implementation.

I considered it, but I’m not sure how to enforce it. It seems like a sketch test.

Sources/Sentry/SentrySDK.m
brustolin: Super weird stuff, somehow I hade reverted the change I made.

philipphofmann:
> > It would be great to have a test failing if we change the code in the SentrySDK.m back to the old implementation.
> 
> I considered it, but I’m not sure how to enforce it. It seems like a sketch test.

What about using the sample app you created and adding a UI or integration test?

.github/workflows/ui-tests.yml
```diff
@@ -129,3 +129,23 @@ jobs:
...
             ./fastlane/test_output/**
+            
+  duplication-tests:
+    name: UI Tests for project with Sentry duplicated
+    runs-on: macos-15
+
+    steps:
+      - uses: actions/checkout@v4
+      - run: ./scripts/ci-select-xcode.sh "16.0"
+      - run: ./scripts/build-xcframework.sh gameOnly
```
philipphofmann, line 140: `m`: `gameOnly` is a bit confusing to me. Ideally this would use the artifact built here https://github.com/getsentry/sentry-cocoa/blob/166d67abd7e722717fe4cf460c20fcc28412ac1b/.github/workflows/build.yml#L86-L87

but if I'm not mistaken this workflow would need to wait until the build workflow is finished, and that would slow down all UI tests. We could add another workflow called something ui-tests-with-xcframework, and move the duplication test there.

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTestApp.swift
```diff
@@ -0,0 +1,50 @@
+import ObjectiveC.runtime
+import Sentry
+import SwiftUI
+
+@main
+struct DuplicatedSDKTestApp: App {
+    
+    init () {
+        SentrySDK.start { options in
+            options.dsn = "https://a@o.sentry.io/1"
```
philipphofmann, line 10: `m`: What drove you to not put the DSN of the other sample apps here?

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTest-Bridging-Header.h
```diff
@@ -0,0 +1,14 @@
+
+#import <Sentry/Sentry.h>
```
philipphofmann, line 2: `m`: I would prefer moving this project to `TestSamples/`.

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTestApp.swift
```diff
@@ -0,0 +1,50 @@
...
+        guard let integrations = SentrySDK.currentHub().installedIntegrations else {
+            return false
+        }
+        
+        guard let sentryImageC = class_getImageName(SentrySDK.self) else {
+            return false
+        }
+        let sentryImage = String(cString: sentryImageC)
+        
+        return integrations.allSatisfy({ element in
```
philipphofmann, line 41: TIL: `allSatisfy`

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTestApp.swift
```diff
@@ -0,0 +1,50 @@
...
+    var body: some View {
+        VStack {
+            Text("Test SentrySDK Integrations")
+            Text("\(checkIntegrations())")
+                .accessibilityIdentifier("TEST_RESULT")
+        }
+        .padding()
+    }
+    
+    func checkIntegrations() -> Bool {
```
philipphofmann, line 31: `l`: I think a bit more a descriptive method name as `areSDKIntegrationsInstalledInSameImageAsSentrySDK` or adding a comment what this does would improve the readability of the code.

Tests/DuplicatedSDKTest/DuplicatedSDKTest/UITest.swift
```diff
@@ -0,0 +1,11 @@
+import XCTest
+
+final class UITestDuplicatedSDK: XCTestCase {
+
+    @MainActor
+    func testExample() throws {
+        let app = XCUIApplication()
+        app.launch()
+        XCTAssertEqual(app.staticTexts["TEST_RESULT"].label, "true")
```
philipphofmann, line 9: `m`: It would be great if the test would fail with something more descriptive than label text isn't true, because when this fails the next time it could be a bit hard to understand what is going on.

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTestApp.swift
```diff
@@ -0,0 +1,50 @@
+import ObjectiveC.runtime
+import Sentry
+import SwiftUI
+
+@main
+struct DuplicatedSDKTestApp: App {
```
philipphofmann, line 6: `m`: A quick comment explaining why we have this sample app would be great cause it's not that obvious.

philipphofmann (commented):
Thanks a lot for adding the sample to validate that this doesn't happen again anymore. I added a few comments mostly for improving the readability of the code.

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTest-Bridging-Header.h
brustolin: I would prefer to move TestSamples to be a sub folder Tests. Actually Im thinking about doing that in a another PR.

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTestApp.swift
brustolin: Actually it was copilot auto complete that wrote that. And since it works, I see no reason to use a valid one.

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTest-Bridging-Header.h
philipphofmann: Also fine. I don't want two different location for the same thing 👍

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTestApp.swift
philipphofmann: If somebody once debugs it manually, you could expect that the project also sends the events to sentry-cocoa in the sentry-sdks org, as every other sample project does. If it doesn't, it could take a while to figure this out.

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTestApp.swift
brustolin, line 6:
```suggestion

// We will use this sample for the only purpose of testing 
// whether the SDK will work if its duplicated in an App,
// by loading it in two different frameworks used by the app.
struct DuplicatedSDKTestApp: App {
```

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTestApp.swift
brustolin: You may be overthinking.  That is not the purpose of this test project, but it is also fine; it does not really matter.
```suggestion
            options.dsn = "https://6cc9bae94def43cab8444a99e0031c28@o447951.ingest.sentry.io/5428557"
```

Tests/DuplicatedSDKTest/DuplicatedSDKTest/UITest.swift
brustolin:
```suggestion
        XCTAssertEqual(app.staticTexts["TEST_RESULT"].label, "true", "SentrySDK integrations are not being loaded from the same binary. This will lead to undefined behavior.")
```

Tests/DuplicatedSDKTest/DuplicatedSDKTest/UITest.swift
```diff
@@ -0,0 +1,11 @@
+import XCTest
+
+final class UITestDuplicatedSDK: XCTestCase {
+
+    @MainActor
+    func testExample() throws {
```
brustolin, line 6:
```suggestion
    func testLoadIntegrations() throws {
```

Sources/Sentry/SentrySDK.m
```diff
@@ -465,6 +465,11 @@ + (void)installIntegrations
         [SentrySDK.currentHub getClient].options.integrations.mutableCopy;
 
     NSArray<Class> *defaultIntegrations = SentryOptions.defaultIntegrationClasses;
+
+    // We are going to use a dictionary that links the classes with their names
+    // so we can quickly check whether that class is in the option integrations collection.
+    // We cannot load the class itself with NSClassFromString because doing so may load a class
+    // that was duplicated in another module, leading to undefined behavior.
```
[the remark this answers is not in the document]
brustolin: > // We cannot load the class itself with NSClassFromString because doing so may load a class
    // that was duplicated in another module, leading to undefined behavior.
    
IMO this to be descriptive enough.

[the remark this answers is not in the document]
brustolin: Feel free to add a suggestion  ` ```suggestion ` for an alternative

.github/workflows/ui-tests.yml
brustolin: Let's wait to see if this will become a recurring thing, which I dont think so.
For now having the SDK compiled for the test itself works fine.

Tests/DuplicatedSDKTest/DuplicatedSDKTest/DuplicatedSDKTestApp.swift
```diff
@@ -0,0 +1,54 @@
...
+@main
+
+// We will use this sample for the only purpose of testing 
+// whether the SDK will work if its duplicated in an App,
+// by loading it in two different frameworks used by the app.
+struct DuplicatedSDKTestApp: App {
+    
+    init () {
+        SentrySDK.start { options in
+            options.dsn = "https://6cc9bae94def43cab8444a99e0031c28@o447951.ingest.sentry.io/5428557"
```
philipphofmann, line 14: Thanks 👍

Tests/DuplicatedSDKTest/DuplicatedSDKTest/UITest.swift
```diff
@@ -0,0 +1,11 @@
+import XCTest
+
+final class UITestDuplicatedSDK: XCTestCase {
+
+    @MainActor
+    func testLoadIntegrations() throws {
+        let app = XCUIApplication()
+        app.launch()
+        XCTAssertEqual(app.staticTexts["TEST_RESULT"].label, "true", "SentrySDK integrations are not being loaded from the same binary. This will lead to undefined behavior.")
```
philipphofmann, line 9: Thanks 🥇

Sources/Sentry/SentrySDK.m
```diff
@@ -464,16 +464,32 @@ + (void)installIntegrations
     NSMutableArray<NSString *> *integrationNames =
         [SentrySDK.currentHub getClient].options.integrations.mutableCopy;
 
+    NSArray<Class> *defaultIntegrations = SentryOptions.defaultIntegrationClasses;
+
+    // We are going to use a dictionary that links the classes with their names
+    // so we can quickly check whether that class is in the option integrations collection.
+    // We cannot load the class itself with NSClassFromString because doing so may load a class
+    // that was duplicated in another module, leading to undefined behavior.
```
philipphofmann, line 472: Here you go:

```suggestion
    // Since 8.22.0, we use a precompiled XCFramework for SPM, which can lead to Sentry's 
    // definition getting duplicated in the app with a warning “SentrySDK is defined in both
    // ModuleA and ModuleB”. This doesn't happen when users use Sentry-Dynamic and
    // when compiling Sentry from source via SPM. Due to the duplication, some users didn't
    // see any crashes reported to Sentry cause the SentryCrashReportSink couldn't find
    // a hub bound to the SentrySDK, and it dropped the crash events. This problem
    // is fixed now by using a dictionary that links the classes with their names
    // so we can quickly check whether that class is in the option integrations collection.
    // We cannot load the class itself with NSClassFromString because doing so may load a class
    // that was duplicated in another module, leading to undefined behavior.
```

philipphofmann (approved):
LGTM, when including the more detailed comment.

PR #4542, opened 2024-11-15
chore: Bump OS versions for unit tests

brustolin:
Testing our SDK in the latest iOS.

_#skip-changelog_


philprime:
I believe we should not bump the version from iOS 17 to iOS 18, but instead keep test both versions due to changes in FileIO (#4546).

In addition I would propose we also add macOS 15 unit tests to this PR.

philprime:
I just noticed the same applies for tvOS 17.5 and 18.1

philprime:
@philipphofmann @armcknight I propose we use this PR to test compatibility with iOS 18, macOS 15 and tvOS 18 by updating the branch to `main` and merge the PR as soon as all checks pass. When we notice something is broken, like file I/O tracking in #4546, we fix it in other PRs instead. This should lead to multiple commits with clearer intentions.

philipphofmann:
We know that file IO instrumentation is broken. I would skip the failing tests for file IO instrumentation on iOS 18, merge this PR, so we run all other tests on iOS 18, and then enable the tests again once we fix the file IO instrumentation. It could also be that the file IO instrumentation won't work anymore on iOS 18, so we can't enable the failing tests anymore.

philprime:
As I am about to finish fixing FileIO in #4605, there might not be much of a point in disabling and then re-enabling it, except for having CI running on the introduced OS versions too.

But there is actually another test failing with the newer OS versions, we need to look into/disable:

```
Failing tests:
	-[SentryFileIOTrackingIntegrationObjCTests test_NSFileManagerCreateFile]
	-[SentryFileIOTrackingIntegrationObjCTests test_NSFileManagerCreateFile]
	-[SentryFileIOTrackingIntegrationObjCTests test_NSFileManagerCreateFile]
	-[SentryFileIOTrackingIntegrationObjCTests test_NSFileManagerCreateFile]
	-[SentryFileIOTrackingIntegrationObjCTests test_NSFileManagerCreateFile]
	SentryFileIOTrackingIntegrationTests.test_ReadingURL_Tracking()
	SentryFileIOTrackingIntegrationTests.test_ReadingURLWithOption_Tracking()
	SentryFileIOTrackingIntegrationTests.test_Writing_Tracking()
	SentryFileIOTrackingIntegrationTests.test_WritingWithOption_Tracking()

	SentryStacktraceBuilderTests.testFramesOrder() <---
```

philipphofmann:
It's also fine to disable all failing tests for now and create follow-up issues. Most tests running on iOS 18 is long overdue.

philprime:
The broken tests are now skipped, so I believe this can be merged.

Tests/SentryTests/Integrations/Performance/IO/SentryFileIOTrackingIntegrationObjCTests.m
```diff
@@ -177,6 +177,9 @@ - (void)test_NSFileManagerContentAtPath
 
 - (void)test_NSFileManagerCreateFile
 {
+    if (@available(iOS 18, macOS 15, tvOS 15, *)) {
+        XCTSkip("File IO tracking for Swift.Data is disabled for this OS version");
```
philipphofmann: `l`: A bit more specific:
```suggestion
        XCTSkip("File IO tracking for Swift.Data is not working for this OS version. Therefore, we disable this test until we fix it: https://github.com/getsentry/sentry-cocoa/issues/4546");
```

philipphofmann (approved):
LGTM.

Issue #4271, opened 2024-08-09
User Feedback form (due: Dec 6)


armcknight:
### Description

Build the configurable UI form that captures the end user's feedback

armcknight:
Current progress: have the form built, displaying/dismissing correctly with hiding/reshowing widget button. 

Needs fixing: 
- custom text field borders don't have same dimensions as default
- svg logo drawing: svg arc definitions weren't translated to cgpath arcs correctly by cursor, will have to calculate and do by hand, or potentially use PaintCodeApp if possible

|Default|Customized|
|---|---|
|![Image](https://github.com/user-attachments/assets/8941a489-b08f-45ac-b885-ffe609b2e01a)|![Image](https://github.com/user-attachments/assets/b479c783-e222-4ffd-bb3c-149d764a8759)|

armcknight:
Update, got the logo fixed, and got the field/button heights a little taller to match the screenshot in the docs for the web version. Did some work to get the the UITextField and UITextView inner padding equal, as by default they are annoyingly different: ![Image](https://github.com/user-attachments/assets/aa031ab0-b60d-442a-b8e0-3173c611ffba)

armcknight:
This is what it will look like when there is a screenshot selected: ![Image](https://github.com/user-attachments/assets/b50f80d2-2601-49e0-9460-ac3eb12105e3)

armcknight:
There's an issue where the keyboard covers elements on smaller screens. There are a couple options to mitigate this, and we might take a combination of approaches:
- change the scrollview's content size and offsets so that the portion of the screen above the keyboard is effectively the entire visible viewport for its content
- automatically dismiss the keyboard when scrolling down, similar to how UITableView works (we _could_ implement the whole view as a UITableView to get this for free, but I think layout would actually be more of a pain that way; cost/benefit doesn't seem worth it, unless for some reason implementing this behavior is a huge stretch for UIScrollView, which I doubt)
- add an accessory view to the keyboard with buttons to dismiss the keyboard (and also traverse the different form elements; I do this in several of my own apps, it's one of my favorite approaches, but is a screen real estate tradeoff, especially if the predictive text ribbon is also visible, which it will be for most people)
- make the keyboard's "return" key dismiss the keyboard. should work fine in all but the text view, where we'd have to decide that it's ok to not allow users to write in multiple paragraphs

armcknight:
[...]

PR #4544, opened 2024-11-15
feat(user-feedback): iteration on form

armcknight:
for https://github.com/getsentry/sentry-cocoa/issues/4271; #skip-changelog

fleshes out some of the outstanding things from https://github.com/getsentry/sentry-cocoa/pull/4530


Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -109,7 +109,28 @@ class SentryUserFeedbackForm: UIViewController {
...
+        
+        if !messageTextView.hasText {
+            missing.append("description")
+        }
+        
+        guard missing.isEmpty else {
+            let list = missing.count == 1 ? missing[0] : missing[0 ..< missing.count - 1].joined(separator: ", ") + " and " + missing[missing.count - 1]
+            let alert = UIAlertController(title: "Error", message: "You must provide all required information. Please check the following fields: \(list).", preferredStyle: .alert)
+            alert.addAction(UIAlertAction(title: "OK", style: .default))
+            present(alert, animated: config.widgetConfig.animations)
```
philipphofmann, line 130: `l`: I'm curious. Do we need localization for this at some point?

Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
```diff
@@ -82,7 +80,9 @@ class UserFeedbackUITests: BaseUITest {
         
         app.staticTexts["Send Bug Report"].tap()
         
-        // TODO: need to implement some kind of error dialog and then assert for it
+        XCTAssert(app.staticTexts["Error"].exists)
```
philipphofmann, line 83: `m`: I think it would be great to validate the actual error message in the tests.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -109,7 +109,28 @@ class SentryUserFeedbackForm: UIViewController {
...
     
     func submitFeedbackButtonTapped() {
-        // TODO: validate and package entries
+        var missing = [String]()
+        
+        if config.formConfig.isNameRequired && fullNameTextField.hasText {
+            missing.append("name")
+        }
+        
+        if config.formConfig.isEmailRequired && emailTextField.hasText {
```
philipphofmann, line 118: `m`: Still in progess, I know, but I think it should be like that:
```suggestion
        if config.formConfig.isNameRequired && !fullNameTextField.hasText {
            missing.append("name")
        }
        
        if config.formConfig.isEmailRequired && !emailTextField.hasText {
```

philipphofmann (commented):
Early feedback

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: I just mentioned this the other day but it was never officially discussed yet, and I don't know what is currently being done on the Javascript side. Presumably we would want it one day. wdyt @bruno-garcia ?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: Oops, you're right, I changed these from `.text.isEmpty` without adding negation 🤦🏻

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
philipphofmann: Step by step. We do it once people ask for it.

armcknight:
The remaining broken test is fixed in an downstream PR: https://github.com/getsentry/sentry-cocoa/pull/4594

Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
armcknight: Disagree; this is set with standard API, and we are already validating that the alert is showing. Testing that kind of detail is testing the Apple framework, not anything we have control over. https://github.com/getsentry/sentry-cocoa/blob/e61b1088ff17cfa6432913f43115647a0cab83eb/Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift#L124-L126

philipphofmann: Then all good

armcknight: Actually I reconsidered this and added many more assertions for various things like this in https://github.com/getsentry/sentry-cocoa/pull/4600

PR #4548, opened 2024-11-18
Fix: Session replay touch tracking race condition

brustolin:
## :scroll: Description

Touch tracker adds events on the main thread, but reads it from a background thread, this could lead to a race condition.

Unit tests

## :pencil: Checklist


Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -116,7 +120,11 @@ class SentryTouchTracker: NSObject {
         
         var result = [SentryRRWebEvent]()
         
-        for info in trackedTouches.values {
+        trackedTouchesLock.lock()
+        let touches = trackedTouches.values
+        trackedTouchesLock.unlock()
```
philipphofmann, line 125: `l`: 
```suggestion
        let touches = trackedTouchesLock.synchronized {
            return trackedTouches.values
        }
```

Tests/SentryTests/Integrations/SessionReplay/SentryTouchTrackerTests.swift
```diff
@@ -300,5 +300,38 @@ class SentryTouchTrackerTests: XCTestCase {
...
+        }
+        
+        DispatchQueue.global().async {
+            group.enter()
+            for _ in 0..<50_000 {
+                _ = sut.replayEvents(from: self.referenceDate, until: self.referenceDate.addingTimeInterval(10_000.0))
+            }
+            group.leave()
+        }
+        group.wait()
```
philipphofmann, line 334: `m`: Please replace this with using [XCTestExpectation](https://developer.apple.com/documentation/xctest/xctestexpectation), as they let the tests fail if they don't get called and this here can run forever.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -45,28 +46,29 @@ class SentryTouchTracker: NSObject {
...
-                info.endEvent = newEvent
-            case .moved:
-                // If the distance between two points is smaller than 10 points, we don't record the second movement.
-                // iOS event polling is fast and will capture any movement; we don't need this granularity for replay.
-                if let last = info.moveEvents.last, touchesDelta(last.point, position) < 10 { continue }
-                info.moveEvents.append(newEvent)
-                debounceEvents(in: info)
-            default:
-                continue
+        trackedTouchesLock.synchronized {
```
philipphofmann, line 49: `h`: This gets called from the main thread and is a huge risk for slowing it down. We already accept race conditions in the frames tracker when getting frames statistics from a background thread because, anyway, the statistics won't be 100% accurate, and it's better to have 99% accuracy over blocking the main thread. Instead, we should ensure we don't crash when reading tracked touches from a background thread and adding touches on the main thread.

philipphofmann (commented):
There is a risk of blocking the main thread.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
brustolin: But this race condition can lead to crashes. By removing this synchronization, the test just not fail, it crashes.

This is not a matter of "we accept the consequences of race conditions". This is actually necessary.

brustolin: Also, the other two methods that also synchronize in this object are super fast locks. I don’t think we need to worry about this.

Tests/SentryTests/Integrations/SessionReplay/SentryTouchTrackerTests.swift
brustolin: Ow, true, I forgot the timeout

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philipphofmann: OK, so `flushFinishedEvents` and `replayEvents` read `trackedTouches`, which get called on a bg thread. Therefore it doesn't matter if we also do this on a bg thread. We only need to ensure updating the values runs before reading them. So, we can move this logic here in the lock to a bg thread. Ideally, it uses the same sync queue as the code calling `flushFinishedEvents` and `replayEvents`, and then you wouldn't even need locks.

philipphofmann: I would prefer not blocking the main thread for `sendEvent:`. That gets called frequently when you for example scroll on a page.

brustolin: Yeah, we can do that. But Im keeping the lock and not managing dispatch queues in here

brustolin: Okay, I’m not keeping the lock, but I need to synchronize the access to `trackedTouches` for `flushFinishedEvents` and `replayEvents`. Otherwise, the test fails because the responsibility of calling these functions in the same dispatcher as `trackedTouches` would fall on the caller, which poses a significant future risk.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -34,39 +35,48 @@ class SentryTouchTracker: NSObject {
...
-    init(dateProvider: SentryCurrentDateProvider, scale: Float) {
+    init(dateProvider: SentryCurrentDateProvider, scale: Float, dispatchQueue: SentryDispatchQueueWrapper) {
         self.dateProvider = dateProvider
         self.scale = CGAffineTransform(scaleX: CGFloat(scale), y: CGFloat(scale))
+        self.dispatchQueue = dispatchQueue
+    }
+    
+    convenience init(dateProvider: SentryCurrentDateProvider, scale: Float) {
+        self.init(dateProvider: dateProvider, scale: scale, dispatchQueue: SentryDispatchQueueWrapper())
     }
```
philipphofmann, line 51: `m`: https://github.com/getsentry/sentry-cocoa/blob/da7a50e80c96a369ab8e53bfa59ea4efb0b57f90/Sources/Sentry/SentrySessionReplayIntegration.m#L90-L95 now uses the convenience init which creates an extra `SentryDispatchQueueWrapper`. Does this class need it's own instance? If not, I would recommend deleting the convenience init and using the `SentryDispatchQueueWrapper` from the DependencyContainer.

philipphofmann (commented):
Almost, LGTM

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
brustolin: SentryDispatchQueueWrapper default dispatchQueue uses a serial queue shared with every other code that uses it. I really dont think this is the way to go only to save a few bytes in memory, specially in an operation that will use its dispatch queue super often.

philipphofmann: > I really don’t think this is the way to go only to save a few bytes in memory, especially in an operation that will use its dispatch queue super often.

I prefer saving a few bytes in memory when it's easy possible when we don't need the extra queue. Do you think we need an extra one and if so why?

brustolin: Yes, to avoid adding another bottleneck to session replay, which is already slow.

philipphofmann: > Yes, to avoid adding another bottleneck to session replay, which is already slow.

Please add this info as a comment for the convenience init.

> What is the problem with SentryTouchTracker having its own dispatch queue?

A few bytes in memory, but if you want to have your own dispatch queue to ensure other stuff doesn't slow down SR, I'm fine with the tradeoff.

philipphofmann (approved):
LGTM

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -34,39 +35,48 @@ class SentryTouchTracker: NSObject {
...
     
-    init(dateProvider: SentryCurrentDateProvider, scale: Float) {
+    init(dateProvider: SentryCurrentDateProvider, scale: Float, dispatchQueue: SentryDispatchQueueWrapper) {
         self.dateProvider = dateProvider
         self.scale = CGAffineTransform(scaleX: CGFloat(scale), y: CGFloat(scale))
+        self.dispatchQueue = dispatchQueue
+    }
+    
+    convenience init(dateProvider: SentryCurrentDateProvider, scale: Float) {
+        self.init(dateProvider: dateProvider, scale: scale, dispatchQueue: SentryDispatchQueueWrapper())
```
brustolin, line 50:
```suggestion
        // SentryTouchTracker has it own dispatch queue instead of using the one
        // from Dependency container to avoid the bottleneck of sharing the same
        // queue with the rest of the SDK.
        self.init(dateProvider: dateProvider, scale: scale, dispatchQueue: SentryDispatchQueueWrapper())
```

PR #4553, opened 2024-11-19
ref: Make SentryId Sendable

### LINKED ISSUE #4545: Conform `SentryId` to `Sendable`
### Problem Statement

Hello!

I kindly request adding `Sendable` conformance to `SentryId`, seeing as it was refactored to Swift in #3708. Ideally this type would be a `struct` and this would be trivial, but as most of the SDK is still written in ObjC I understand why it is not.

Adding `Sendable` conformance should be really straightforward:
- make it a `final` class
- make the global `empty` a constant

Concrete use case I have for this is passing the `eventId` out of `onCrashedLastRun` in a Swift 6 app using strict concurrency. Right now I am falling back to passing the `sentryIdString` instead and then reconstructing the `SentryId` from that string once user feedback was collected.

Thank you!

### Solution Brainstorm

_No response_

Yes

brustolin:
## :scroll: Description

Transform SentryId into a Sendable to easily accommodate for strict concurrency

close #4545 

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Improvements
+
+- Make SentryId Sendable
+
 ## 8.41.0-beta.1
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Improvements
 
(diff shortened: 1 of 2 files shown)


philipphofmann (approved):
LGTM

PR #4554, opened 2024-11-19
ref: Enable strict concurrency check

brustolin:
## :scroll: Description

Enabled Strict Concurrency Check so we can have this extra layer of protection against race conditions, and prepare for customer Swift projects also changing to Swift 6.


brustolin:
Unfortunately this will not work for Xcode 15, which is still supported by Apple. So we will need to postpone it.

PR #4562, opened 2024-11-22
chore: Explain how to promote a beta release

philipphofmann:
Explain how to promote a beta release in Contributing.md.

#skip-changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,8 +128,8 @@ We frequently release a beta version of our SDK and dogfood it with internal app
 2. Duplicate the Changelog.md entry of the beta release and change header of the version number to unreleased.
 3. Commit and push the changes.
 4. Trigger the release workflow with use workflow from the `publish/x.x.x` branch and set the target branch to merge into to `publish/x.x.x`, cause per default craft will merge into the main branch and this could lead to merge conflicts in the changelog.
-  a. After the successful release, craft merges the changes back into `publish/x.x.x` branch and deletes the release branch.
-5. Manually open a PR from the `publish/x.x.x` branch into the main branch and merge it.
+5. After the successful release, validate that craft merged the changes back into `publish/x.x.x` branch and deleted the release branch.
+6. Manually open a PR from the `publish/x.x.x` branch into the main branch and merge it.
 
 ## Final Notes
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -127,7 +127,8 @@ We frequently release a beta version of our SDK and dogfood it with internal app
 1. Checkout a new branch from the GH tag of the beta release: `git checkout -b publish/x.x.x x.x.x-beta.1`. You can't use `release/x.x.x` or `x.x.x` as the branch name as craft will fail, as it creates a `release/x.x.x` branch for updating the changelog and it will create a tag `x.x.x` for the release.
 2. Duplicate the Changelog.md entry of the beta release and change header of the version number to unreleased.
 3. Commit and push the changes.
-4. Trigger the release workflow with use workflow from the `publish/x.x.x` branch.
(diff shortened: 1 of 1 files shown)


brustolin (approved):

CONTRIBUTING.md
```diff
@@ -120,6 +120,17 @@ You can run samples in a real device without changing certificates and provision
...
 
+## Promoting a beta release to a normal release
+
+We frequently release a beta version of our SDK and dogfood it with internal apps to increase our SDK stability. We continue to merge PRs to the main branch, so we can't promote a beta release by publishing it from the main branch. Instead, we create a branch from the GH tag of the beta release and promote it from there. To do this, follow these steps:
+
+1. Checkout a new branch from the GH tag of the beta release: `git checkout -b publish/x.x.x x.x.x-beta.1`. You can't use `release/x.x.x` or `x.x.x` as the branch name as craft will fail, as it creates a `release/x.x.x` branch for updating the changelog and it will create a tag `x.x.x` for the release.
+2. Duplicate the Changelog.md entry of the beta release and change header of the version number to unreleased.
+3. Commit and push the changes.
+4. Trigger the release workflow with use workflow from the `publish/x.x.x` branch.
```
brustolin, line 130: m: During the release workflow, we can choose the same branch for Craft to merge it back, so Craft does not fail.
Also,  we need to manually update the changelog in the main branch.

philipphofmann: But that sucks pretty much. Ideally, we would merge the branch back to main. Does it then fail if we do that?

brustolin: It fails because more likely than not, the main will be "ahead" of the beta tag.

philipphofmann: OK, but then I would say let crash merge back into the publish branch and then merge the publish branch manually back to main?

brustolin: Yes, we can do that after merging main into the publish branch. Good idea

krystofwoldrich: I'm also for setting the target as `publish/x.x.x` for the `x.x.x` release of `x.x.x-beta.x`. 

And then merging manually, if merged directly to mainl, it will fail in better case, in worse case it will break changelog or code.

Not ideal but safe. 

Note: The JS team solved this by using `develop` branch as default and main for releases. Maybe we could adopt similar workflow in the future.

philipphofmann: I updated the description. @brustolin and @krystofwoldrich please have another look.

CONTRIBUTING.md
```diff
@@ -120,6 +120,18 @@ You can run samples in a real device without changing certificates and provision
...
+
+1. Checkout a new branch from the GH tag of the beta release: `git checkout -b publish/x.x.x x.x.x-beta.1`. You can't use `release/x.x.x` or `x.x.x` as the branch name as craft will fail, as it creates a `release/x.x.x` branch for updating the changelog and it will create a tag `x.x.x` for the release.
+2. Duplicate the Changelog.md entry of the beta release and change header of the version number to unreleased.
+3. Commit and push the changes.
+4. Trigger the release workflow with use workflow from the `publish/x.x.x` branch and set the target branch to merge into to `publish/x.x.x`, cause per default craft will merge into the main branch and this could lead to merge conflicts in the changelog.
+5. Manually open a PR from the `publish/x.x.x` branch into the main branch and merge it.
+
+After the successful release, craft merges the `publish/x.x.x` branch into the main branch and deletes the release branch.
```
brustolin, line 133:
```suggestion
4. Trigger the release workflow with use workflow from the `publish/x.x.x` branch and set the target branch to merge into to `publish/x.x.x`, cause per default craft will merge into the main branch and this could lead to merge conflicts in the changelog.
4.1. After the successful release, craft merges the changes back into `publish/x.x.x` branch and deletes the release branch.
5. Manually open a PR from the `publish/x.x.x` branch into the main branch and merge it.
```

Or  something like this.

PR #4567, opened 2024-11-22
fix: Invalid context on GraphQL request errors

### LINKED ISSUE #4560: enableGraphQLOperationTracking attaches invalid context to network errors
### Platform

iOS

### Environment

Production

### Installed

Swift Package Manager

### Version

8.40.1

### Xcode Version

16.1

Unknown

### Steps to Reproduce

1. Set `enableGraphQLOperationTracking = true` when configuring the Sentry SDK.
2. Ensure `enableNetworkTracking` is enabled
3. In a running app, make a GraphQL request that returns a 500

### Expected Result

An event should appear in Sentry with the `graphql_operation_name` in the context. Probably a "GraphQL" context with an "Operation Name" field.

(As a bonus, I'd love it for `graphql_operation_name` to show up in the tags so that it can be used to differentiate between different request errors. Right now all of our network errors to our API get lumped together into a single Sentry issue.)

### Actual Result

The Sentry event includes a warning indicating that the `graphql` context is invalid because it's a String instead of an object.

Image

From the event:

```json
"_meta": {
  "contexts": {
    "graphql_operation_name": {
      "": {
        "err": [
          [
            "invalid_data",
            {
              "reason": "expected an object"
            }
          ]
        ],
        "val": "Settings"
      }
    }
  }
}
```

Sure

jmccance:
## :scroll: Description

Fixes SentryNetworkTracker to add an object for GraphQL context instead of a string.

Fixes #4560.

- Added a unit test to SentryNetworkTracker to verify that the context object has a `graphql` key that looks the way we expect it to look.
- Tested the change locally in our app and verified that the warning documented in #4560 doesn't appear after this change and the new `graphql` context section does.

## :pencil: Checklist

  - I don't _think_ there are any docs for this that need updating.
  - This change shouldn't affect project setup
  - I don't think so, but will wait for someone to tell me.

## :crystal_ball: Next steps

Ideally I'd like to include the GraphQL operation name in the client tags. This would help us identify which actual calls are causing the most errors — currently we have one issue with tens of thousands of events in it and no good way to figure out which calls are the problem.

But for now I'm just making sure we're able to see the operation correctly on network errors.


Brewfile.lock.json
```diff
@@ -2,40 +2,40 @@
   "entries": {
     "brew": {
       "clang-format": {
-        "version": "19.1.3",
+        "version": "19.1.4",
```
jmccance, line 5: Following instructions from `check-tooling-versions.sh`:

> clang-format version mismatch, expected: 19.1.3, but found: 19.1.4. Please run `make init` to update your local dev tools. This may actually upgrade to a newer version than what is currently recorded in the lockfile; if that happens, please commit the update to the lockfile as well.

Let me know if you don't want this change in this PR, but it was preventing me from committing with the pre-commit hook in place.

Sources/Sentry/SentryNetworkTracker.m
```diff
@@ -456,8 +456,9 @@ - (void)captureFailedRequests:(NSURLSessionTask *)sessionTask
     context[@"response"] = response;
 
     if (self.isGraphQLOperationTrackingEnabled) {
-        context[@"graphql_operation_name"] =
-            [URLSessionTaskHelper getGraphQLOperationNameFrom:sessionTask];
+        NSMutableDictionary<NSString *, id> *graphql = [[NSMutableDictionary alloc] init];
+        context[@"graphql"] = graphql;
+        graphql[@"operation_name"] = [URLSessionTaskHelper getGraphQLOperationNameFrom:sessionTask];
```
jmccance, line 461: Not sure if there's anything special needed to get prettier names in the Sentry dashboard, or if that's necessary. Here's how this looks in a test event I generated with our app:

image

`GraphQL` and `Operation Name` would look more aligned with the `Operating System` context, but my hunch is these may be getting special treatment server-side since I don't see these strings in the `sentry-cocoa` project anywhere.

philipphofmann:
Thanks a lot for the PR. I didn't get to this today, @jmccance. I will review this tomorrow.

philipphofmann (commented):
Thanks, @jmccance. Almost LGTM. Please also add an entry to the Changelog.md.

jmccance:
Thanks for the feedback, @philipphofmann! Let me know if there's anything else I can do.

philipphofmann (approved):
LGTM, thank you, @jmccance, for the PR.

PR #4574, opened 2024-11-26
fix: Masking for fast animations

brustolin:
## :scroll: Description

Speeding up mask calculation 

Samples Only

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Tools/UIRedactBuilder.swift
+++ b/Sources/Swift/Tools/UIRedactBuilder.swift
@@ -181,7 +181,7 @@ class UIRedactBuilder {
     func redactRegionsFor(view: UIView) -> [RedactRegion] {
         var redactingRegions = [RedactRegion]()
         
-        self.mapRedactRegion(fromView: view,
+        self.mapRedactRegion(fromLayer: view.layer.presentation() ?? view.layer,
                              relativeTo: nil,
                              redacting: &redactingRegions,
                              rootFrame: view.frame,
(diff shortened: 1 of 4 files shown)


Sources/Swift/Tools/UIRedactBuilder.swift
```diff
@@ -272,8 +270,8 @@ class UIRedactBuilder {
             /// The beginning will be added after all the subviews have been mapped.
             redacting.append(RedactRegion(size: layer.bounds.size, transform: newTransform, type: .clipEnd))
         }
-        for subview in view.subviews.sorted(by: { $0.layer.zPosition < $1.layer.zPosition }) {
-            mapRedactRegion(fromView: subview, relativeTo: layer, redacting: &redacting, rootFrame: rootFrame, transform: newTransform, forceRedact: enforceRedact)
+        for subLayer in (layer.sublayers ?? []).sorted(by: { $0.zPosition < $1.zPosition }) {
+            mapRedactRegion(fromLayer: subLayer, relativeTo: layer, redacting: &redacting, rootFrame: rootFrame, transform: newTransform, forceRedact: enforceRedact)
         }
```
philipphofmann, line 275: `m`: What stopped you from writing a test for this in `UIRedactBuilderTests`?

philipphofmann (approved):
LGTM, if we really can't write a test for this.

Sources/Swift/Tools/UIRedactBuilder.swift
brustolin: There is nothing different to test.
Instead of iterating over view.subviews we're iterating over the view layer which is a lower level api, but navigating the same hierarchy pretty much.

But all the UIRedactBuilder tests in all the ways we could use it are there.

brustolin: There is nothing different to test.
Instead of iterating over view.subviews we're iterating over the view layer which is a lower level api, but navigating the same hierarchy pretty much.

But all the UIRedactBuilder tests in all the ways we could use it are there.
What do you have in mind?

philipphofmann: A test that fails with the old code that validates that masking for fast animation works. I have no clue how to do this 😅. If you say it's not worth the effort, merge the PR.

PR #4585, opened 2024-11-29
Fix: Easy use of Swift classes in Objective-C projects

### LINKED ISSUE #4543: Replay options not visible from Objective-C without @import
### Platform

iOS

### Environment

Production

### Installed

CocoaPods

### Version

Latest React Native one

### Xcode Version

not sure

_No response_

### Steps to Reproduce

A customer on React Native, tried to enable replay through the native layer directly (they are initializign Sentry native directly).
But they said they couldn't call `options.experimental` from Objective-C

_No response_

brustolin:
## :scroll: Description

In order to access Swift classes in an Objective-C project, developers need to either use `@import Sentry;` which is not always possible because the project needs to be compatible with `-fmodules` and `-fcxx-modules`, or they need to import both `#import <Sentry/Sentry.h>` and `#import <Sentry/Sentry-Swift.h>`, which is not intuitive and they are always coming to git hub to understand what's going on.

Closes #4543 

Samples and CI

## :pencil: Checklist


CHANGELOG.md
```diff
@@ -13,7 +13,7 @@
 - Use `options.reportAccessibilityIdentifier` for Breadcrumbs and UIEvents (#4569) 
 - Session replay transformed view masking (#4529)
 - Load integration from same binary (#4541)
-
+- Easy use of Swift classes in Objective-C projects ()
```
brustolin, line 16:
```suggestion
- Easy use of Swift classes in Objective-C projects (#4585)
```

Sources/Sentry/Public/Sentry.h
```diff
@@ -42,11 +42,13 @@ FOUNDATION_EXPORT const unsigned char SentryVersionString[];
 #    import <Sentry/SentrySpanProtocol.h>
 #    import <Sentry/SentrySpanStatus.h>
 #    import <Sentry/SentryStacktrace.h>
+#    import <Sentry/SentrySwift.h>
```
philipphofmann, line 45: `h`: Doesn't this make all Swift classes public?

I can now access `NSArray<NSString *> * features = [SentryEnabledFeaturesBuilder getEnabledFeaturesWithOptions:options];` in the AppDelegate.m from the iOS-Objective-C sample app.

brustolin: Yes, it does for Objc project.

If you think this is a problem we close the PR, I tried other ways to solve this without success, its this or nothing.
I dont think users will start to use things they dont know how to use. It's already hard to make them use the public APIs that have documentation.

Also, this will be a big problem for some users with special projects, like this one:  #4543

philipphofmann: > I dont think users will start to use things they dont know how to use.

Murphy's law: "Anything that can go wrong will go wrong." So someone will use these classes. They always do. If I'm not mistaken, there is no difference between an actual public class in ObjC and an accidentally public class as `SentryEnabledFeaturesBuilder`. So, how can I tell the difference between the two in ObjC code? I didn't look closely into this, but maybe a stupid question: Can't we split up the classes into two headers? We can also discuss this on a call, if this turns out the be a longer discussion.

philipphofmann: Also, didn't we want to move the SR code into its own module / SPM / Cocoa package? Maybe this would help here as well.

brustolin: > Also, didn't we want to move the SR code into its own module / SPM / Cocoa package? Maybe this would help here.

Yes, this would be a solution.

> So, how can I tell the difference between the two in ObjC code?

You can't, both are public.

> Can't we split up the classes into two headers? 

No, its an auto generated header by Xcode

philipphofmann: I would appreciate it if we didn't make all the Swift classes public to fix this issue, especially if we plan on converting more code to Swift. So even though moving the SR code to its extra package seems like a lot of work to fix this, it's the right thing to do, IMO. It would also reduce the size of the sentry-cocoa lib.

brustolin:
We dont want to expose all the Swift classes to objc. We will have to find another solution.

philipphofmann:
@brustolin, agreed thanks for the comment.

PR #4536, opened 2024-11-15
feat(feedback): form ui iteration and tests

armcknight:
for #4271; #skip-changelog

- Adds UI tests to validate the functionality of the widget button and form functionality
- Iterates on the form UI to fix up some interaction and layout issues.

PR #4590, opened 2024-12-04
fix(feedback): keyboard avoidance

armcknight:
for #4271; #skip-changelog

fixes the issue described in https://github.com/getsentry/sentry-cocoa/pull/4536#issuecomment-2515934691; the contents of the form were already in a scrollview, so this just adjusts its insets to regain visibility of all elements while the keyboard is occupying screen space

before:

https://github.com/user-attachments/assets/e2655496-f13e-4888-a293-01295260b0a8

after:

https://github.com/user-attachments/assets/dd9fa542-650d-4b0b-89fd-f481840f5b01


Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -31,6 +33,17 @@ class SentryUserFeedbackForm: UIViewController {
...
+        
+        NotificationCenter.default.addObserver(forName: UIResponder.keyboardDidShowNotification, object: nil, queue: nil) { note in
+            guard let keyboardValue = note.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue else { return }
+            let keyboardViewEndFrame = self.view.convert(keyboardValue.cgRectValue, from: self.view.window)
+            self.setScrollViewBottomInset(keyboardViewEndFrame.height - self.view.safeAreaInsets.bottom)
+
+        }
+        NotificationCenter.default.addObserver(forName: UIResponder.keyboardWillHideNotification, object: nil, queue: nil) { _ in
+            self.setScrollViewBottomInset(0)
+        }
```
brustolin, line 45: h: We need to remove the observer when the view controller disappears.

Also, I would prefer to use a function reference instead of a closure to improve readability.

armcknight: You're right 👍🏻  but we'll only need one or the other. Function-based automatically removes observation.

brustolin: > but we'll only need one or the other

Yes.

PR #4596, opened 2024-12-04
feat: SwiftUI time for initial display and time for full display

brustolin:
## :scroll: Description

Track time for initial display and time for full display in SwiftUI

UI Test and sample

## :pencil: Checklist


Samples/iOS-SwiftUI/iOS-SwiftUI-UITests/SentryRedactModifierTests.swift
```diff
@@ -0,0 +1,23 @@
+#if canImport(UIKit) && canImport(SwiftUI)
+@testable import Sentry
+@testable import SentrySwiftUI
+import SwiftUI
+import XCTest
+
+class SentryRedactModifierTests: XCTestCase {
```
armcknight, line 7: What is the relationship between redaction and TTID/TTFD?

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -36,15 +36,15 @@ @implementation SentryTimeToDisplayTracker {
     BOOL _waitForFullDisplay;
     BOOL _initialDisplayReported;
     BOOL _fullyDisplayedReported;
-    NSString *_controllerName;
+    NSString *_name;
```
armcknight, line 39: I expected this to be related to something with TracingView.name, but I only saw "UIViewController" being passed into the init parameter for this. Did I miss anything?

brustolin: Now with SwiftUI, we also pass the view name.

Samples/iOS-SwiftUI/iOS-SwiftUI-UITests/SentryRedactModifierTests.swift
brustolin: This test was moved from `SentryTest` to `iOS-SwiftUI-UITests` because with all the changes to SentrySwiftUI, we cant add it as a SentryTest dependency. It complains about duplicate declarations.

Sources/SentrySwiftUI/TracingView.swift
```diff
@@ -0,0 +1,58 @@
...
+            }
+            
+            super.init(frame: CGRect(origin: .zero, size: CGSize(width: 1, height: 1)))
+        }
+        
+        required init?(coder: NSCoder) {
+            fatalError("init(coder:) has not been implemented")
+        }
+      
+        override func didMoveToWindow() {
```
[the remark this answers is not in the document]
brustolin: `didMoveToSuperView` can happen outside the view hierarchy.
We actually dont care about the parent, we just need to know that the view was added to the view hierarchy and is about to be rendered.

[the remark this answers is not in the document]
philprime: Can you please further elaborate how we know that `didMoveToSuperView` is for sure called when the view is rendered? On iOS it is possible to have multiple windows, even on iOS (I have done that before), calling `UIWindow.makeKeyAndVisible()` for the main window, while others are stacked by deriving instances from a `UIWindowScene` and setting `isHidden` to `false`.

In most cases this will probably be good enough, just want to point out that there might be unexpected edge cases.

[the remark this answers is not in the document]
philprime: Ah yes, I ment to write didMoveToWindow. Thanks for elaborating

Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -39,61 +37,112 @@ import SwiftUI
...
+            }
         }
+        
+        return content()
+            .onAppear { viewAppeared = true }
+#if canImport(SwiftUI) && canImport(UIKit) && os(iOS) || os(tvOS)
+            // We need to add a UIView to the view hierarchy to be able to
+            // monitor ui life cycles. We will use the background modifier
+            // to add this tracking view behind the content.
+            .background(TracingView(name: self.name, waitForFullDisplay: self.waitforFullDisplay, tracer: trace))
```
philipphofmann, line 104: `h`: Maybe that's a dump question 😅: I don't understand why we can't use `.onAppear` and we have to use a UIView. Why do we need to monitor UI life cycles? The `SentryTimeToDisplayTracker` waits for the next frame to be drawn anyway. Please elaborate, or we can discuss this on a call.

philipphofmann (commented):
I have one important comment to address that could impact a rework of this PR. Therefore, I wait until we resolve it before giving this a more detailed review.

Sources/SentrySwiftUI/SentryTracedView.swift
brustolin: It's not a dump question at all. 
And actually I had something else in mind when I started the PR that was not possible. .onAppear may be good enough for this purpose.

philipphofmann: Excellent. This would be great because the code won't depend on UIKit directly.

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -53,6 +53,13 @@ - (instancetype)initForController:(UIViewController *)controller
     return self;
 }
 
+- (instancetype)initWithName:(NSString *)name waitForFullDisplay:(BOOL)waitForFullDisplay
+{
+    return [self initWithName:name
+           waitForFullDisplay:waitForFullDisplay
+         dispatchQueueWrapper:SentryDependencyContainer.sharedInstance.dispatchQueueWrapper];
```
philipphofmann, line 60: `m`: Please stick to the pattern of passing in the `dispatchQueueWrapper` as a dependency.

Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -38,62 +40,122 @@ import SwiftUI
...
-            }
+        var trace: SentryTracer?
+        var spanId: SpanId?
+        
+        if !viewAppeared {
+            trace = ensureTransactionExists()
+            spanId = createAndPushBodySpan(transactionCreated: trace != nil)
+#if canImport(SwiftUI) && canImport(UIKit) && os(iOS) || os(tvOS)
+            if let trace = trace { startTTDTraker(for: trace) }
+#endif
```
philipphofmann, line 97: `m`: I would prefer keeping the logic for creating the transaction or span and keeping track of the TTDTracker in the `SentryUIViewControllerPerformanceTracker`. Now, we have two different places that can easily get out of sync.

Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -38,62 +40,122 @@ import SwiftUI
...
+            operation: "ui.load",
+            origin: traceOrigin
+        )
+        SentryPerformanceTracker.shared.pushActiveSpan(spanId)
+        return spanId
+    }
+    
+    private func finishSpan(_ spanId: SpanId) {
+        SentryPerformanceTracker.shared.popActiveSpan()
+        SentryPerformanceTracker.shared.finishSpan(spanId)
```
philipphofmann, line 158: `m`:  There is so much logic in this code in here. I would appreciate it if we could extract the logic into an extra testable component and/or write tests for all the different types of edge cases. Currently, I don't know how I can trust this code without having a good test suite covering multiple edge cases.

Sources/Sentry/SentryTimeToDisplayTracker.m
brustolin: This is a convenient init so we dont have to re-define SentryDispatchQueue for SentrySwiftUI, which is a different module, it does not have access to everything inside of Sentry.

Sources/SentrySwiftUI/SentryTracedView.swift
brustolin: This specific piece of logic is particular to the SwiftUI lifecycle. I don’t think it makes sense to move it to SentryUIViewControllerPerformanceTracker, but I did move other parts related to TTID to SentryUIViewControllerPerformanceTracker to reduce the responsibilities of this class.

Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -38,62 +109,56 @@ import SwiftUI
...
+    /// - Parameter waitForFullDisplay: Indicates whether this view transaction should wait for `SentrySDK.reportFullyDisplayed()`
+    /// in case you need to track some asyncronous task. This is ignored for any `SentryTracedView` that is child of another `SentryTracedView`
+    /// - Parameter content: The content that you want to track the performance
+    public init(_ viewName: String? = nil, waitForFullDisplay: Bool? = nil, @ViewBuilder content: @escaping () -> Content) {
         self.content = content
-        self.name = viewName ?? SentryTracedView.extractName(content: Content.self)
-        self.nameSource = viewName == nil ? .component : .custom
+        let name = viewName ?? SentryTracedView.extractName(content: Content.self)
+        let nameSource = viewName == nil ? SentryTransactionNameSource.component : SentryTransactionNameSource.custom
+        let waitforFullDisplay = waitForFullDisplay ?? SentrySDK.options?.enableTimeToFullDisplayTracing ?? false
```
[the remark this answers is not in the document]
brustolin: True. The tests related to TTID are missing. Going to add it

Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -1,14 +1,85 @@
...
+    }
+    
+    func viewDidAppear() {
+        guard !viewAppeared else { return }
+        viewAppeared = true
+        tracker?.reportInitialDisplay()
+        
+        if let transactionId = transactionId {
+            self.finishSpan(transactionId)
+            SentryPerformanceTracker.shared.popActiveSpan()
```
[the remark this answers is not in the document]
brustolin: We need to pop the root transaction, but not always. There is a bug in here 🤦🏻

scripts/xcode-test.sh
```diff
@@ -15,6 +15,7 @@ COMMAND="${4:-test}"
 DEVICE=${5:-iPhone 14}
 CONFIGURATION_OVERRIDE="${6:-}"
 DERIVED_DATA_PATH="${7:-}"
+TEST_SCHEME="${8:Sentry}"
```
brustolin, line 18:
```suggestion
TEST_SCHEME="${8:-Sentry}"
```

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -163,25 +163,51 @@ - (void)startRootSpanFor:(UIViewController *)controller
...
+            [self.currentTTDTracker reportFullyDisplayed];
+        } else {
+            SENTRY_LOG_WARN(@"Transaction is not waiting for full display report. You can enable "
+                            @"`enableTimeToFullDisplay` option, or use the waitForFullDisplay "
+                            @"property in our `SentryTracedView` view for SwiftUI.");
+        }
+    } else {
+        SENTRY_LOG_DEBUG(@"No screen transaction being tracked right now.")
+    }
+}
```
philprime, line 191: `l:` this method could be rewritten using Early Return pattern, to reduce nesting

```suggestion
    SentryTimeToDisplayTracker *tracker = self.currentTTDTracker;
    if (tracker == nil) {
        SENTRY_LOG_DEBUG(@"No screen transaction being tracked right now.")
        return;
    }
    if (!tracker.waitForFullDisplay) {
        SENTRY_LOG_WARN(@"Transaction is not waiting for full display report. You can enable "
                        @"`enableTimeToFullDisplay` option, or use the waitForFullDisplay "
                        @"property in our `SentryTracedView` view for SwiftUI.");
        return;
    }
    [self.currentTTDTracker reportFullyDisplayed];
```

Sources/SentrySwiftUI/SentryInternal/SentryInternal.h
```diff
@@ -56,4 +71,46 @@ typedef NS_ENUM(NSUInteger, SentrySpanStatus);
 
 @end
 
+@interface SentryTimeToDisplayTracker : NSObject
```
philprime, line 74: `l:` This header is growing to a point where you should consider moving the class interfaces into own header files which are imported and exported, turning this into an umbrella header file

Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -38,62 +109,56 @@ import SwiftUI
...
-            DispatchQueue.main.async {
-                SentryPerformanceTracker.shared.popActiveSpan()
-                SentryPerformanceTracker.shared.finishSpan(transactionId)
-            }
-        }
-
-        let id = SentryPerformanceTracker.shared.startSpan(withName: transactionCreated ? "\(self.name) - body" : self.name, nameSource: nameSource, operation: "ui.load", origin: self.traceOrigin)
-
-        SentryPerformanceTracker.shared.pushActiveSpan(id)
+        let spanId = viewModel.startSpan()
```
philprime, line 153: `m:` Would this create a new span on every single re-calculation of the SwiftUI view hierarchy, instead of only the first time?
Do we know if this has a performance impact?

philprime (changes requested):

philprime (approved):
Approved with comments so the PR is not blocked by me

Sources/SentrySwiftUI/SentryTracedView.swift
brustolin: No, it is only created at the first time. The `startSpan()` function checks for that.

Sources/SentrySwiftUI/SentryInternal/SentryInternal.h
brustolin: Let's do it in the next iteration, in case we need to add more things in here. 😉

Sources/SentrySwiftUI/SentryTracedView.swift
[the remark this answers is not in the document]
brustolin: Actually you're right, just need to remove this pop duplication

Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -38,62 +109,56 @@ import SwiftUI
...
-    let nameSource: SentryTransactionNameSource
     
-    let traceOrigin = "auto.ui.swift_ui"
-
-    public init(_ viewName: String? = nil, content: @escaping () -> Content) {
+#if canImport(SwiftUI) && canImport(UIKit) && os(iOS) || os(tvOS)
+    /// Creates a view that measures the performance of its `content`.
+    ///
+    /// - Parameter viewName: The name that will be used for the span, if nil we try to get the name of the content class.
+    /// - Parameter waitForFullDisplay: Indicates whether this view transaction should wait for `SentrySDK.reportFullyDisplayed()`
```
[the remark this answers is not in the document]
brustolin: I added more info in the parameter comment

PR #4597, opened 2024-12-04
chore(ci): add direct branch cloning for fastlane match

philprime:
## :scroll: Description

Changes `sync_code_signing` in fastlane lanes to use direct branch cloning.

Workflow runs were occasionally failing due to the following error:

```
Cloning remote git repo...
If cloning the repo takes too long, you can use the `clone_branch_directly` option in match.
Checking out branch master...
fatal: a branch named 'master' already exists
```

Example: https://github.com/getsentry/sentry-cocoa/actions/runs/12154028573/job/33893118719?pr=4593#step:5:96

By directly defining the branch to clone, the branch checkout step should be skipped instead.

Created this PR to run workflows on CI.

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -83,7 +83,9 @@ platform :ios do
     sync_code_signing(
       type: "appstore",
       readonly: true,
-      app_identifier: ["io.sentry.sample.iOS-Swift",  "io.sentry.sample.iOS-Swift.Clip"]
+      app_identifier: ["io.sentry.sample.iOS-Swift",  "io.sentry.sample.iOS-Swift.Clip"],
+      git_branch: "master",
+      clone_branch_directly: true
     )
 
     build_app(
@@ -107,7 +109,9 @@ platform :ios do
     sync_code_signing(
       type: "development",
       readonly: true,
-      app_identifier: ["io.sentry.sample.iOS-Swift", "io.sentry.sample.iOS-Swift.Clip"]
+      app_identifier: ["io.sentry.sample.iOS-Swift", "io.sentry.sample.iOS-Swift.Clip"],
+      git_branch: "master",
+      clone_branch_directly: true
     )
 
     build_app(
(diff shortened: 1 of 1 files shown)


brustolin (approved):
LGTM

fastlane/Fastfile
```diff
@@ -83,7 +83,9 @@ platform :ios do
     sync_code_signing(
       type: "appstore",
       readonly: true,
-      app_identifier: ["io.sentry.sample.iOS-Swift",  "io.sentry.sample.iOS-Swift.Clip"]
+      app_identifier: ["io.sentry.sample.iOS-Swift",  "io.sentry.sample.iOS-Swift.Clip"],
+      git_branch: "master",
+      clone_branch_directly: true
```
philipphofmann, line 88: `l`: Please comment on why we have to use this. Basically it fixes the rare fatal: a branch named 'master' already exists in GH actions.

philipphofmann (approved):
LGTM. 

Worth a shot if it doesn't improve we can always roll back. I noticed that checkout in GH actions sets the default branch to master
```
/usr/local/bin/git init /Users/runner/work/sentry-cocoa/sentry-cocoa
  hint: Using 'master' as the name for the initial branch. This default branch name
  hint: is subject to change. To configure the initial branch name to use in all
  hint: of your new repositories, which will suppress this warning, call:
  hint:
  hint: 	git config --global init.defaultBranch <name>
  hint:
  hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
  hint: 'development'. The just-created branch can be renamed via this command:
  hint:
  hint: 	git branch -m <name>
  Initialized empty Git repository in /Users/runner/work/sentry-cocoa/sentry-cocoa/.git/
  ```

But that's the same for release build swift succeeding or failing. If we see the same error happening again in the future, maybe we could look into this.

fastlane/Fastfile
philipphofmann: No, I think that's not required. We can always check what the flag does. The change fixes a problem with CI that's not obvious when you look at the code. I'm thinking of the future me that sees that in 1 year, has no clue why it's there, removes it, and then wonders why CI starts to fail again.

philprime: Added comments

philipphofmann (approved):
LGTM

PR #4598, opened 2024-12-04
chore(internal): Add `packages` set to SentrySdkInfo

krystofwoldrich:
## :scroll: Description

This PR adds `packages` to Sentry Envelope Header SDK Info and to Sentry Event SDK Info.

Before this PR `packages` were not set in the Event and in Header the Cocoa package was hardcoded.

This will be used by hybrid SDKs to add itself as package, later we can use this information during debugging SR and other events. 

By sending events from Sentry SwiftUI sample and unit tests.

## :pencil: Checklist


Sources/Sentry/SentrySdkInfo.m
```diff
@@ -72,43 +56,41 @@ - (instancetype)initWithDictInternal:(NSDictionary *)dict orDefaults:(SentrySdkI
...
-        @"version" : self.version,
-    }
-                                   .mutableCopy;
-    if (self.packageManager != SentryPackageManagerUnkown) {
-        NSString *format = [self getPackageName:self.packageManager];
-        if (format != nil) {
-            sdk[@"packages"] = @{
-                @"name" : [NSString stringWithFormat:format, self.name],
-                @"version" : self.version
-            };
```
krystofwoldrich, line 107: Based on the development docs this should be an array, I don't know if this value was used anywhere, so I just updated it to the array.

If it's not used I would consider removing it completely and only create packages in SentryClient (same as features, integrations...).

krystofwoldrich:
This seems like a lot of code for a small change, please when review let me know, if I've overlooked something and there is better place to save the packages to.

Sources/Sentry/SentryEnvelope.m
```diff
@@ -31,7 +31,8 @@ - (instancetype)initWithId:(nullable SentryId *)eventId
               traceContext:(nullable SentryTraceContext *)traceContext
 {
     SentrySdkInfo *sdkInfo = [[SentrySdkInfo alloc] initWithName:SentryMeta.sdkName
-                                                      andVersion:SentryMeta.versionString];
+                                                      andVersion:SentryMeta.versionString
+                                                     andPackages:[SentryMeta getSdkPackages]];
```
philipphofmann, line 35: `h`: The SDKInfo of the envelope header is the same as the one from the event payload: see

> This can carry the same payload as the [sdk interface](https://develop.sentry.dev/sdk/data-model/event-payloads/sdk/)
in the event payload but can be carried for all events. This means that SDK
information can be carried for minidumps, session data and other submissions. - https://develop.sentry.dev/sdk/data-model/envelopes/#envelope-headers

Now that we touch both SDKInfo for the envelope header and the event payload, I think it would be great to merge the different implementations. To make this easier we could do this in an extra PR before adding the new packages implementation.

Sources/Sentry/include/SentryMeta.h
```diff
@@ -14,6 +15,28 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, class, copy) NSString *sdkName;
 
+/**
+ * Return an array of Serialized SDK packages
+ */
++ (NSArray<NSDictionary<NSString *, NSString *> *> *)getSdkPackagesSerialized;
```
philipphofmann, line 21: `m`: I don't think the SDK packages really belong here. The SentryMeta is only about version and SDK name. `getSdkPackagesSerialized` is a sign for me that it doesn't belong here. Instead, I would create an extra class just for the SDK packages. You can even write it in Swift.

Sources/Sentry/SentrySdkPackage.m
```diff
@@ -0,0 +1,120 @@
+#import "SentrySdkPackage.h"
+#import "SentryMeta.h"
+#import <Foundation/Foundation.h>
+
+typedef NS_ENUM(NSUInteger, SentryPackageManagerOption) {
```
philipphofmann, line 5: `m`: Can we convert this code to Swift, please?

Sources/Sentry/SentrySdkPackage+Private.h
```diff
@@ -0,0 +1,11 @@
+#if __has_include(<Sentry/SentryDefines.h>)
+#    import <Sentry/SentrySdkPackage.h>
+#else
+#    import "SentrySdkPackage.h"
+#endif
+
+@interface SentrySdkPackage ()
```
philipphofmann, line 7: `m`: SentrySDKPackage is internal. Why did you have to add a `+Private.h` header?

Sources/Sentry/SentrySdkPackage+Equality.h
```diff
@@ -0,0 +1,13 @@
+#import "SentrySdkPackage.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentrySdkPackage (Equality)
```
philipphofmann, line 5: `m`: If we use a Swift struct for just the data of SentrySDKPackage you get isEqual for free.

philipphofmann (commented):
I found a view high-level issues. We need to address them before I give this a more detailed review. @krystofwoldrich we can also discuss this on a call.

Issue #4426, opened 2024-10-10
Investigate why `sentry-cocoa` requires `APPLICATION_EXTENSION_API_ONLY=YES`


krystofwoldrich:
### Description

- Related to https://github.com/getsentry/sentry-react-native/issues/3908#issuecomment-2265703635 and https://github.com/getsentry/sentry-react-native/issues/3883

Large number of RN users are directly and indirectly overwrite this configuration to `NO`. Likely due to soma other library. But we didn't get that confirmed. 

Do we know why `sentry-cocoa` requires this? Can we investigate it why we deed it to document the reasons or if possible (and would help RN) create a `sentry-cocoa` version which doesn't require this build flag?

antonis:
Added some notes on the [RN issue](https://github.com/getsentry/sentry-react-native/issues/3908#issuecomment-2438136419).

krystofwoldrich:
The reason is captured in https://github.com/getsentry/sentry-cocoa/pull/4602

PR #4599, opened 2024-12-04
Allow compiling with APPLICATION_EXTENSION_API_ONLY=NO

antonis:
## :scroll: Description

⚠️ **Exposes internal methods** to allow compiling with `APPLICATION_EXTENSION_API_ONLY=NO`.

I was expecting that setting `APPLICATION_EXTENSION_API_ONLY=NO` would cause no issues since [it is supposed to be less restricting allowing the usage of forbidden extension API](https://github.com/getsentry/sentry-react-native/issues/3908#issuecomment-2438136419). 
In reality as soon as you enable `Allow app extension API only` the generated header will contain both public and internal classes ([reference](https://stackoverflow.com/questions/42319892/checking-allow-app-extension-api-only-in-ios-framework-project-exposes-interna), [reference 2](https://forums.developer.apple.com/forums/thread/72768)). This means that disabling the flag hides many of the internal declarations that the project needs to compile.

This PR exposes those classes allowing for the project to compile with `APPLICATION_EXTENSION_API_ONLY` both set to `YES` and `NO`.

Fixes https://github.com/getsentry/sentry-cocoa/issues/4426 and https://github.com/getsentry/sentry-react-native/issues/3908

1. Compile and run `iOS-Swift` project on the `main` branch
2. Select the `Sentry` target and set `APPLICATION_EXTENSION_API_ONLY` to `NO`
3. Notice that compilation fails with random errors ❌ 
4. Switch to this branch
5. Compile and run `iOS-Swift` project with `APPLICATION_EXTENSION_API_ONLY` both set to `YES` and `NO` ✅ 

## :pencil: Checklist

## :crystal_ball: Next steps

If this is is a viable solution:
- fix failing tests
- clean up the code
- document the changes
- add CI checks compiling the project with `APPLICATION_EXTENSION_API_ONLY` both set to `YES` and `NO`

### DIFF
--- a/Sources/Sentry/Public/SentryDebugImageProvider.h
+++ b/Sources/Sentry/Public/SentryDebugImageProvider.h
@@ -1,6 +1,8 @@
+#import "SentryDebugMeta.h"
+#import "SentryThread.h"
 #import <Foundation/Foundation.h>
 
-@class SentryDebugMeta, SentryThread, SentryFrame;
+@class SentryFrame;
 
 NS_ASSUME_NONNULL_BEGIN
 
--- a/Sources/Sentry/SentrySessionReplayIntegration.m
+++ b/Sources/Sentry/SentrySessionReplayIntegration.m
@@ -304,16 +304,16 @@ static SentryTouchTracker *_touchTracker;
                                               attributes:attributes];
 
     self.sessionReplay = [[SentrySessionReplay alloc]
-        initWithReplayOptions:replayOptions
-             replayFolderPath:docs
-           screenshotProvider:screenshotProvider
-                  replayMaker:replayMaker
-          breadcrumbConverter:breadcrumbConverter
-                 touchTracker:_touchTracker
-                 dateProvider:SentryDependencyContainer.sharedInstance.dateProvider
-                     delegate:self
-                dispatchQueue:dispatchQueue
-           displayLinkWrapper:[[SentryDisplayLinkWrapper alloc] init]];
+            initWithReplayOptions:replayOptions
+                 replayFolderPath:docs
+               screenshotProvider:screenshotProvider
+                      replayMaker:replayMaker
+              breadcrumbConverter:breadcrumbConverter
+                     touchTracker:_touchTracker
+                     dateProvider:SentryDependencyContainer.sharedInstance.dateProvider
+                         delegate:self
+             dispatchQueueGeneric:dispatchQueue
+        displayLinkWrapperGeneric:[[SentryDisplayLinkWrapper alloc] init]];
 
     [self.sessionReplay
         startWithRootView:SentryDependencyContainer.sharedInstance.application.windows.firstObject
@@ -447,7 +447,7 @@ static SentryTouchTracker *_touchTracker;
     self.sessionReplay = nil;
 }
 
-- (void)sentrySessionEnded:(SentrySession *)session
+- (void)sentrySessionEnded
 {
     [self pause];
     [_notificationCenter removeObserver:self
@@ -459,7 +459,7 @@ static SentryTouchTracker *_touchTracker;
     _sessionReplay = nil;
 }
 
-- (void)sentrySessionStarted:(SentrySession *)session
+- (void)sentrySessionStarted
 {
     _rateLimited = NO;
     [self startSession];
--- a/Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
+++ b/Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
@@ -9,7 +9,7 @@ enum SessionReplayError: Error {
 }
 
 @objc
-protocol SentrySessionReplayDelegate: NSObjectProtocol {
+public protocol SentrySessionReplayDelegate: NSObjectProtocol {
     func sessionReplayShouldCaptureReplayForError() -> Bool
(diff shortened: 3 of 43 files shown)


antonis:
Hey @krystofwoldrich, @brustolin 👋 
Though this PR is still draft I'd like you feedback on if this is a viable solution since it exposes a lot of our internal API 🙇

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -459,7 +459,7 @@ - (void)sentrySessionEnded:(SentrySession *)session
     _sessionReplay = nil;
 }
 
-- (void)sentrySessionStarted:(SentrySession *)session
```
antonis, line 462: Removed the parameter to avoid exposing it since it was not really used in the implementation

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -72,9 +73,33 @@ class SentrySessionReplay: NSObject {
...
+        self.init(replayOptions: replayOptions,
+                  replayFolderPath: replayFolderPath,
+                  screenshotProvider: screenshotProvider,
+                  replayMaker: replayMaker,
+                  breadcrumbConverter: breadcrumbConverter,
+                  touchTracker: touchTracker,
+                  dateProvider: dateProvider,
+                  delegate: delegate,
+                  dispatchQueue: dispatchQueueGeneric as! SentryDispatchQueueWrapper,
+                  displayLinkWrapper: displayLinkWrapperGeneric as! SentryDisplayLinkWrapper)
```
antonis, line 96: This convenience initialiser is more of a trick to avoid exposing the `SentryDispatchQueueWrapper` and `SentryDisplayLinkWrapper` types since this created more complications. I will revisit this for a better solution without a forced cast if needed.

Sources/Sentry/Public/SentryDebugImageProvider.h
```diff
@@ -1,6 +1,8 @@
+#import "SentryDebugMeta.h"
+#import "SentryThread.h"
 #import <Foundation/Foundation.h>
 
-@class SentryDebugMeta, SentryThread, SentryFrame;
```
antonis, line 3: Forward declarations also didn't work in this and similar cases and were replaced with imports.

brustolin (commented):
We have discussed recently that we don’t want to expose all the classes that were made public in this PR. Do we have any alternative approach?

krystofwoldrich:
> making private classes public, which would lead to users misusing it and more problems for us.

Yes, that make sense.

> I dont think so.

Got it, I've mixed up being in `Sentry-Swift.h` and `public`.

krystofwoldrich:
I just added the side effect explanation in separate PR as that information is important regardless the way forward we choose.

https://github.com/getsentry/sentry-cocoa/pull/4602

antonis:
Thank you all for the feedback 🙇 

> I believe the right thing to do is to remove @objc from as many Swift classes as possible, and refactor around its usage so we only have Swift public classes that are meant to be used by the user. 

This sounds like a more proper approach 👍 

<details><summary>The list of swift files turned public with this PR may provide an initial scope for the refactoring.</summary>
<p>

```
Helper/Log/SentryLevel.swift
Helper/SentryBaggageSerialization.swift
Helper/SentryCurrentDateProvider.swift
Helper/SentryEnabledFeaturesBuilder.swift
Helper/SentryFileContents.swift
Integrations/ANR/SentryANRTracker.swift
Integrations/ANR/SentryANRTrackerV2Delegate.swift
Integrations/ANR/SentryANRType.swift
Integrations/FramesTracking/SentryFramesDelayResult.swift
Integrations/Performance/SwizzleClassNameExclude.swift
Integrations/SessionReplay/RRWeb/SentryRRWebBreadcrumbEvent.swift
Integrations/SessionReplay/RRWeb/SentryRRWebCustomEvent.swift
Integrations/SessionReplay/RRWeb/SentryRRWebEvent.swift
Integrations/SessionReplay/RRWeb/SentryRRWebSpanEvent.swift
Integrations/SessionReplay/SentryOnDemandReplay.swift
Integrations/SessionReplay/SentryReplayEvent.swift
Integrations/SessionReplay/SentryReplayOptions.swift
Integrations/SessionReplay/SentryReplayRecording.swift
Integrations/SessionReplay/SentryReplayType.swift
Integrations/SessionReplay/SentryReplayVideoMaker.swift
Integrations/SessionReplay/SentrySRDefaultBreadcrumbConverter.swift
Integrations/SessionReplay/SentrySessionReplay.swift
Integrations/SessionReplay/SentryTouchTracker.swift
Integrations/SessionReplay/SentryVideoInfo.swift
Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
MetricKit/SentryMXCallStackTree.swift
MetricKit/SentryMXManager.swift
Protocol/SentryRedactOptions.swift
Protocol/SentrySessionListener.swift
SentryExperimentalOptions.swift
SwiftDescriptor.swift
Tools/HTTPHeaderSanitizer.swift
Tools/SentryLog.swift
Tools/SentryViewPhotographer.swift
Tools/SentryViewScreenshotProvider.swift
Tools/UIRedactBuilder.swift
Tools/URLSessionTaskHelper.swift
Tools/UrlSanitized.swift
```
</p>
</details> 

> But this will require some work that I don't think can be solved in one PR.

Makes sense. I'm closing this PR since this we won't move forward with this approach.

PR #4601, opened 2024-12-05
fix: Don't start the SDK inside Xcode preview

brustolin:
## :scroll: Description

The SDK was running inside Xcode preview for SwiftUI and slowing things down for development.

Sample and Unit test

## :pencil: Checklist


Sources/Sentry/SentrySDK.m
```diff
@@ -39,15 +39,20 @@
 #    import "SentryProfiler+Private.h"
 #endif // SENTRY_TARGET_PROFILING_SUPPORTED
 
+#define XCODE_PREVIEW_ENVIRONMENT_KEY @"XCODE_RUNNING_FOR_PREVIEWS"
+
 @interface SentrySDK ()
 
 @property (class) SentryHub *currentHub;
 
+#if TEST
```
armcknight, line 48: Does this need to be `TEST || TESTCI`? I'm guessing it's why tests are failing in CI.

Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -79,12 +79,21 @@ public struct SentryTracedView<Content: View>: View {
         
         return result
     }
+
+    private func shouldTrace() -> Bool {
+#if DEBUG
+        if  ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"] == "1" {
```
armcknight, line 85: If you define that string value as an exported string constant instead of a macro, you should be able to use it here too

Sources/Sentry/SentrySDK.m
```diff
@@ -39,15 +39,20 @@
 #    import "SentryProfiler+Private.h"
 #endif // SENTRY_TARGET_PROFILING_SUPPORTED
 
+#define XCODE_PREVIEW_ENVIRONMENT_KEY @"XCODE_RUNNING_FOR_PREVIEWS"
```
armcknight, line 42: If we keep this as a macro we should prefix the name with `SENTRY_`

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -194,6 +194,17 @@ class SentrySDKTests: XCTestCase {
         XCTAssertIdentical(scope, SentrySDK.currentHub().scope)
     }
     
+    func testDontStartInsideXcodePreview() {
+        SentrySDK.processInfoEnvironment = ["XCODE_RUNNING_FOR_PREVIEWS": "1"]
+        
+        SentrySDK.start { options in
+            options.debug = true
+            options.removeAllIntegrations()
```
armcknight, line 202: Is there some significance to calling this method?

Sources/Sentry/SentrySDK.m
brustolin: If im not mistaken, always running the test should have the TEST macro enabled. I will check if this is what's causing the fail, but we can add TESTCI

Tests/SentryTests/SentrySDKTests.swift
brustolin: Not really no. Ctrl+c Ctrl+v from another test.

brustolin:
```suggestion
```

Sources/Sentry/SentrySDK.m
brustolin:
```suggestion
#if TEST || TESTCI
```

philipphofmann (commented):
I'm not sure about the whole approach of the PR. What exactly is causing the slow-down? Can we do something to improve this? Maybe we should only turn off certain parts of the SDK. It's not obvious that we do this, and some users might get confused.

brustolin:
> I'm not sure about the whole approach of the PR. What exactly is causing the slow-down? Can we do something to improve this? Maybe we should only turn off certain parts of the SDK. It's not obvious that we do this, and some users might get confused.

Do you have any reason why should Sentry run inside Xcode preview?

brustolin:
> Some users might expect it to and have reasons and get confused that we deactivate it. 

I disagree; No user expects their monitoring solution to be running inside of Xcode preview, that is meant to help with creation of screens. There is no relevant information that could be used out of it, and the SDK would be initialized and dealloc many times in sequence, we don’t even know what the expected behavior of such usage is.

> Again, I would like to understand what's causing the slowdown before deactivating it.

The entire SDK. In order to see my SwiftUI preview the SDK needs to initialize all its integrations, needs to create a new thread to monitor Apphangs, needs to swizzle a bunch of  functions and make an analysis of the startup time, and none of this will be used.

Again, what's the user case to run the SDK inside Xcode preview?

Sources/Sentry/SentrySDK.m
```diff
@@ -201,6 +206,18 @@ + (void)setStartTimestamp:(NSDate *)value
 
 + (void)startWithOptions:(SentryOptions *)options
 {
+#if TEST
```
brustolin, line 209:
```suggestion
#if TEST || TESTCI
```

Sources/Sentry/SentrySDK.m
```diff
@@ -613,6 +630,21 @@ + (void)stopProfiler
 }
 #endif // SENTRY_TARGET_PROFILING_SUPPORTED
 
+#if TEST
```
brustolin, line 633:
```suggestion
#if TEST || TESTCI
```

philprime:
I agree with @brustolin on this one. 

Xcode preview supports a static Preview Mode and an interactive Live Mode ([Docs](https://developer.apple.com/documentation/xcode/previewing-your-apps-interface-in-xcode)). As the preview can be used for individual UI components, it is most likely used in some kind of container framework/application (which then runs on the simulator in live mode), therefore not the full application as if it would run normally.

Letting users think they can use the SDK in this environment, might actually cause confusion down the road.

Regarding the performance, not sure how heavy the impact is, but generally the `body` method of any view should be as quickly as possible, especially when redraw happens many times like in preview.

philipphofmann:
It seems wrong that we have to add this logic to our SDK. Why is the SDK even running in the Preview? Isn't this something our users should ensure doesn't happen? The preview shouldn't start Sentry. That should be in the `@main` init of the App. It seems to me we're fixing problems that our users should take care of. Again, I didn't see this in any other SDK yet.

armcknight:
I agree that _most_ things in Sentry should not run in Xcode previews. (Maybe once we launch User Feedback, we may want that to be previewable, but that can also be something we handle when the time comes, after merging something like this). I found lots of other dev tools on GitHub that are looking at this environment variable to decide whether or not to do things:
- https://github.com/search?q=repo%3Agithub/CopilotForXcode%20isPreview&type=code
- https://github.com/mongodb/docs-realm/blob/289e25607ac6d043fef6a71a6b8ad48461f8b277/source/sdk/swift/swiftui/swiftui-previews.txt#L90
- https://github.com/DataDog/dd-sdk-ios/blob/a4c6a1cb49a22eec9f2d5dada8be160082533d35/E2ETests/Runner/Scenarios/Scenario.swift#L43
- https://github.com/pointfreeco/swift-dependencies/blob/a0f3ffd32a670d48a2e7fe96ffd67919acf3d448/Sources/Dependencies/DependencyValues.swift#L404-L405
- https://github.com/EmergeTools/SnapshotPreviews/blob/e29969072e600518867af25d4d2acd0d1d2ffd5f/Examples/DemoApp/DemoModule/ConversationMessageView.swift#L78-L82
- https://github.com/pointfreeco/swift-issue-reporting/blob/a3f634d1a409c7979cabc0a71b3f26ffa9fc8af1/Sources/IssueReporting/IssueReporters/RuntimeWarningReporter.swift#L69-L87
- https://github.com/Dashlane/apple-apps/blob/e66b4b162ac898c751b9ca0403dcb792cf70a566/Packages/Core/CoreUserTracking/Sources/CoreUserTracking/UI/View%2BreportPageAppearance.swift#L33-L38
- https://github.com/pointfreeco/swift-issue-reporting/blob/a3f634d1a409c7979cabc0a71b3f26ffa9fc8af1/Sources/IssueReporting/IssueReporters/RuntimeWarningReporter.swift#L69-L87

I think we should ultimately adopt a similar pattern to CopilotForXcode, Dashlane or PointFreeCo/swift-dependencieas, so that integrations and other basic components (like launch profiling which runs independently of SentrySDK.start, or if we ever implement remote config) can be selectively disabled in previews, instead of the entire SDK.

Sources/Sentry/SentrySDK.m
```diff
@@ -39,15 +39,20 @@
...
 #endif // SENTRY_TARGET_PROFILING_SUPPORTED
 
+NSString *const SENTRY_XCODE_PREVIEW_ENVIRONMENT_KEY = @"XCODE_RUNNING_FOR_PREVIEWS";
+
 @interface SentrySDK ()
 
 @property (class) SentryHub *currentHub;
 
+#if TEST || TESTCI
+@property (nonatomic, strong, class) NSDictionary<NSString *, NSString *> *processInfoEnvironment;
```
armcknight, line 49: can we instead make this part of `TestSentryNSProcessInfoWrapper`? I think if we do that, then we don't have to do the little dance at https://github.com/getsentry/sentry-cocoa/pull/4601/files#diff-1b8a7e8fc89dbafd6be9e731ae59c015f849bb96967cca64c665873b6fdecc1cR209-R219 and https://github.com/getsentry/sentry-cocoa/pull/4601/files#diff-1b8a7e8fc89dbafd6be9e731ae59c015f849bb96967cca64c665873b6fdecc1cR633-R646

armcknight (approved):
A question about using TestSentryNSProcessInfoWrapper, but otherwise I agree we should move forward with this.

Sources/Sentry/SentrySDK.m
brustolin: Yes. It makes sense.

Sources/Sentry/SentrySDK.m
```diff
@@ -613,6 +620,21 @@ + (void)stopProfiler
...
+{
+    _processInfoEnvironment = dictionary;
+}
+
++ (NSDictionary<NSString *, NSString *> *)processInfoEnvironment
+{
+    return _processInfoEnvironment;
+}
+
+#endif
```
philipphofmann, line 636: `m`: That code isn't used.

Sources/Sentry/SentrySDK.m
```diff
@@ -201,6 +203,11 @@ + (void)setStartTimestamp:(NSDate *)value
 
 + (void)startWithOptions:(SentryOptions *)options
 {
+    if ([SentryDependencyContainer.sharedInstance.processInfoWrapper
+                .environment[SENTRY_XCODE_PREVIEW_ENVIRONMENT_KEY] isEqualToString:@"1"]) {
+        return;
```
philipphofmann, line 208: `m`: Please add a WARN log message to communicate why we don't start the SDK.

Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -79,12 +79,21 @@ public struct SentryTracedView<Content: View>: View {
         
         return result
     }
+
+    private func shouldTrace() -> Bool {
+#if DEBUG
+        if  ProcessInfo.processInfo.environment[SENTRY_XCODE_PREVIEW_ENVIRONMENT_KEY] == "1" {
+            return false
```
philipphofmann, line 86: `m`: Please add a log message explaining why we don't trace.

philipphofmann (approved):
You all convinced me. LGTM when adding some log messages.

Sources/Sentry/SentryFramesTracker.m
```diff
@@ -62,6 +61,7 @@ @implementation SentryFramesTracker {
     unsigned int _totalFrames;
     unsigned int _slowFrames;
     unsigned int _frozenFrames;
+    NSString *test;
```
brustolin, line 64:
```suggestion
```

Sources/Sentry/SentryFramesTracker.m
```diff
@@ -223,8 +223,8 @@ - (void)displayLinkCallback
         && frameDuration <= SentryFrozenFrameThreshold) {
         _slowFrames++;
 #    if SENTRY_TARGET_PROFILING_SUPPORTED
-        SENTRY_LOG_DEBUG(
-            @"Detected slow frame starting at %@ (frame tracker: %@).", profilingTimestamp, self);
+        SENTRY_LOG_DEBUG(@"Detected slow frame starting at %@ (frame tracker: %@). %@",
+            profilingTimestamp, self, test);
```
brustolin, line 227:
```suggestion
        SENTRY_LOG_DEBUG(@"Detected slow frame starting at %@ (frame tracker: %@).",
            profilingTimestamp, self);
```

philipphofmann:
@brustolin, is anything stopping us from merging this?

brustolin:
> @brustolin, is anything stopping us from merging this? 

I’m just waiting to finish the SwiftUI TTID PR because this code was created on top of that one, and I don’t want to add it to the other PR and have it reviewed again.

Samples/iOS-SwiftUI/iOS-SwiftUI/ContentView.swift
```diff
@@ -239,7 +239,7 @@ struct ContentView: View {
                     Text(TTDInfo)
                         .accessibilityIdentifier("TTDInfo")
                 }
-            }
+        }
```
brustolin, line 242:
```suggestion
            }
```

philprime (approved):
LGTM

PR #4603, opened 2024-12-05
chore(build): Add an error message when build with missing header due to APPLICATION_EXTENSION_API_ONLY=NO

krystofwoldrich:
## :scroll: Description

Quick improvement before proper fix of https://github.com/getsentry/sentry-cocoa/pull/4599

To allow building of Sentry Cococa with `APPLICATION_EXTENSION_API_ONLY=NO` will require change of structure of the project which is a large change, which needs to be planned.

This PR largely improves the situation as users will see error message pointing them to the solution of the failed build.

Build with and without APPLICATION_EXTENSION_API_ONLY.

Screenshot 2024-12-05 at 20 00 26

## :pencil: Checklist


antonis (approved):
Thank you for this PR @krystofwoldrich 🙇 
The changes LGTM and the compilation fails with the expected error when APPLICATION_EXTENSION_API_ONLY=NO 🎉

krystofwoldrich:
Thank @brustolin the actual error message looks better. 

Screenshot 2024-12-06 at 10 40 07

Sources/Sentry/Public/SentryDefines.h
```diff
@@ -1,5 +1,11 @@
 #import <Foundation/Foundation.h>
 
+// SentryDefines.h is a key header and will be checked early,
+// ensuring this error appears first during the compile process.
+#if APPLICATION_EXTENSION_API_ONLY_NO
+#    error Set APPLICATION_EXTENSION_API_ONLY to YES in the Sentry build settings.
+#endif
```
philipphofmann, line 7: `h`: I'm a bit confused. When we set the `APPLICATION_EXTENSION_API_ONLY` to yes in our Xcode project the compiler ensures we're only using APIs available to app extension. Flipping this flag for normal projects in the Sentry build settings shouldn't make a different. If we now add this check, the compilation will fail for users setting it to NO, even though it used to work for them. Or am I missing something?

brustolin: @philipphofmann, you can check the full explanation in here: https://github.com/getsentry/sentry-cocoa/pull/4599#issuecomment-2520738894 

Summary: Flipping this flag to `NO` makes our Swift internal code not available in `Sentry-Swift.h` breaking the entire project.

brustolin (approved):

Sources/Sentry/Public/SentryDefines.h
philipphofmann: Does this only happen for RN projects or also for iOS projects? I'm afraid we could break existing users setting this flag to NO on purpose.

brustolin: Any project.

> I'm afraid we could break existing users setting this flag to NO on purpose.

That's the thing. We cant set this to NO, it will break Sentry.

philipphofmann: Oh, that's super weird. Now, I get it thanks for explaining.

Sources/Sentry/Public/SentryDefines.h
```diff
@@ -1,5 +1,11 @@
 #import <Foundation/Foundation.h>
 
+// SentryDefines.h is a key header and will be checked early,
+// ensuring this error appears first during the compile process.
+#if APPLICATION_EXTENSION_API_ONLY_NO
+#    error Set APPLICATION_EXTENSION_API_ONLY to YES in the Sentry build settings.
```
philipphofmann, line 6: `m`: Please add a more detailed explanation of why this is required and maybe add a link to the troubleshooting guide. You can do this like that
```
#    error "Set APPLICATION_EXTENSION_API_ONLY to YES in the Sentry build settings. \
#      Some comment \
#      For more information, visit https://docs.sentry.io/platforms/apple/"
```

Sources/Sentry/Public/SentryDefines.h
```diff
@@ -1,5 +1,11 @@
 #import <Foundation/Foundation.h>
 
+// SentryDefines.h is a key header and will be checked early,
+// ensuring this error appears first during the compile process.
```
philipphofmann, line 4: `m`: Please explain the detailed background why this is required including the weird side effect of 
 APPLICATION_EXTENSION_API_ONLY_NO and maybe even some links to the original issues, cause this can be quite confusing.

krystofwoldrich: Added in https://github.com/getsentry/sentry-cocoa/pull/4603/commits/5e14a9d18a78cbca8527d08ae237a005a4434c99

Sources/Sentry/Public/SentryDefines.h
krystofwoldrich: Added in https://github.com/getsentry/sentry-cocoa/pull/4603/commits/3164dd6faee1522cae147f7e69794ab02f54e54e

krystofwoldrich:
Screenshot 2024-12-06 at 16 26 24

Sources/Sentry/Public/SentryDefines.h
philipphofmann: Maybe also briefly mention the problems it caused on RN with a link to the issue.

philipphofmann (approved):
LGTM

brustolin:
> Screenshot 2024-12-06 at 16 26 24

Personally, I don’t think we should say, “Setting the flag to YES has a side effect of including all Swift classes in the `Sentry-Swift.h` header.” In my opinion, this sounds bad and provides too much information.

Simply saying, “Sentry needs this flag set to YES in order to work,” is sufficient for the user.

However, I’ll leave the final decision up to you.

brustolin (approved):
Merge it!! 🚀

PR #4605, opened 2024-12-06
feat: Add manual file IO tracking for Swift.Data and Swift.FileManager

### LINKED ISSUE #4546: FIle IO integration not working for iOS 18
### Description

Some file IO operations not being tracked for iOS 18.

See https://github.com/getsentry/sentry-cocoa/actions/runs/11858258351/job/33048500905?pr=4542

philprime:
Open Sub Tasks:
- [x] Decide if refactoring `SentryNSDataTracker` to `SentryFileIOTracker`
- [x] Decide if iOS/macOS availability check is necessary
- [x] Remove changes in GitHub Actions before merging, as the changes are in #4542 instead
- [x] Fix unit tests in `SentryFileIOTrackingIntegrationTests.swift`

Out-of-scope for this PR:
- [x] Add checks to detect cascading spans created by swizzled methods calling other swizzled methods

Closes #4546 

Separated into multiple PRs:

- [x] #4859 
- [x] #4860 
- [x] #4861
- [x] #4862 
- [ ] #4863


philprime:
We could consider to refactor the class `SentryNSDataTracker` to a broader name like `SentryFileIOTracker` as we are adding methods not directly related to `NSData`

brustolin:
Thanks for this investigation. 

Since you enable testing for iOS 18.2, we have other I/O tests failing. We could solve them in a different PR that will merge to this one, or disable those tests before merging this to Main.

Sources/Swift/Helper/SentryDataWrapper.swift
```diff
@@ -0,0 +1,605 @@
...
+    /// Write the contents of the `SentryDataWrapper` to a location.
+    ///
+    /// - parameter url: The location to write the data into.
+    /// - parameter options: Options for writing the data. Default value is `[]`.
+    /// - throws: An error in the Cocoa domain, if there is an error writing to the `URL`.
+    public func write(to url: URL, options: Data.WritingOptions = []) throws {
+        // TODO: begin file.write via static tracker
+        try self.data.write(to: url, options: options)
+        // TODO: end file.write via static tracker
+    }
```
philprime, line 353: @philipphofmann @brustolin @armcknight 

This is my draft for a wrapper for `Swift.Data`, as it does not bridge to NSData anymore.

I created a Swift struct that can mostly be used as a direct drop-in replacement for `Swift.Data` (it's a `@frozen struct` so I don't think we can "subclass"/swizzle/monkey-patch). 
I also considered creating this wrapper as an Objective-C file, if I am not mistaken this would require to use `NSData`, because AFAIK we can not use the `Swift.Data` struct from Objective-C classes.

Is this the right approach? Right now I am not sure how to access the hub from here and create the `file.read/write` spans. If we want to mimic the swizzling in `SentryNSDataSwizzling.m`, we would need access to the `SentryNSDataSwizzling.shared.dataTracker` or similar

philprime: Closing this comment, further discussion in #4634

brustolin: Imitating Sentry.Data does sound like a good solution, but it requires a lot of work.

Since we only track the atomic reads and writes of NSData, Im wondering if we could of a wrapper and extensions for those functions alone and then return a Sentry.Data

Let's say SentryData.dataForUrl(:) (and similar) returns a Data object. 
And a some Data extensions like .trackingWrite(path:).

But this is not a drop-in replacement and could discourage usage, so if you think maintaining a complete copy of it won't be a problem I believe we can go with it.

Sources/Swift/Helper/SentryDataWrapper.swift
```diff
@@ -0,0 +1,353 @@
...
+//
+//  Created by Philip Niedertscheider on 12.12.24.
+//  Copyright © 2024 Sentry. All rights reserved.
+//
+
+/// A drop-in replacement for the standard ``Swift.Data`` but with automatic tracking for file I/O operations.
+///
+/// This structure is intended to resemble the same method signatures as of ``Swift.Data``.
+@available(iOS 18, macOS 15, tvOS 18, *)
+@frozen public struct SentryDataWrapper: Equatable, Hashable, RandomAccessCollection, MutableCollection, RangeReplaceableCollection, MutableDataProtocol, ContiguousBytes, Sendable {
```
brustolin, line 13: l: Name too long IMO.

Suggestions
- SentryData
- TrackedData
- TracedData

Sources/Swift/Helper/SentryDataWrapper.swift
```diff
@@ -0,0 +1,353 @@
...
+    @available(swift, deprecated: 5, message: "use `init(_:)` instead")
+    public init<S>(bytes elements: S) where S: Sequence, S.Element == UInt8 {
+        self.data = Data(bytes: elements)
+    }
+
+    /// See `Data.init(contentsOf:options:)`
+    public init(contentsOf url: URL, options: Data.ReadingOptions = []) throws {
+        // start file.read via static tracker
+        self.data = try Data(contentsOf: url, options: options)
+        // end file.read via static tracker
```
philprime, line 79: Add tracking

Sources/Swift/Helper/SentryDataWrapper.swift
```diff
@@ -0,0 +1,353 @@
...
+    /// See `Data.subdata(in:)`
+    public func subdata(in range: Range<Data.Index>) -> Data {
+        return self.data.subdata(in: range)
+    }
+
+    /// See `Data.write(to:options:)`
+    public func write(to url: URL, options: Data.WritingOptions = []) throws {
+        // begin file.write via static tracker
+        try self.data.write(to: url, options: options)
+        // end file.write via static tracker
```
philprime, line 203: Add tracking

Sources/Swift/Helper/SentryDataWrapper.swift
```diff
@@ -0,0 +1,348 @@
+//
+//  SentryDataWrapper.swift
+//  Sentry
+//
+//  Created by Philip Niedertscheider on 12.12.24.
+//  Copyright © 2024 Sentry. All rights reserved.
+//
+
```
brustolin, line 8: h: We dont use file headers
```suggestion
```

Sources/Swift/Helper/SentryDataWrapper.swift
philipphofmann: Yes, I'm also wondering if we can do something better than a copy. Maybe we should instead come up with a better API to add custom spans to transactions bound to the scope in general. Or maybe extension methods. I'm not sure yet.

Sources/Swift/Helper/SentryDataWrapper.swift
armcknight: "Clarity is more important than brevity" so I don't take issue with a difference of a couple characters. 

We currently have `SentryNSDataTracker`, I would suggest a parallel convention of `SentrySwiftDataTracker`, with the _annoying exception_ of there now being a SwiftData product offering from Apple that would create a naming overload ಠ_ಠ Maybe just `SentryDataTracker`. Even though the namespacing isn't required in Swift, I still think it's a good idea to keep `Sentry...` tacked onto our types.

philprime: `SentryNSDataTracker` will be renamed to `SentryFileIOTracker` in #4634, rather than a helper class to deal with spans, so it might be used by `SentryDataWrapper`, but it does not serve the same purpose.

I like `Sentry` in the name, but dislike the keyword `Wrapper`, but not due to the length. Therefore I would also agree with `SentryTrackedData` or similar

Sources/Swift/Helper/SentryDataWrapper.swift
philprime: Changed the API according to our internal decision

Sources/Swift/Helper/SentryDataWrapper.swift
philprime: Changed the API and naming according to our internal decision

Sources/Swift/Integrations/Performance/IO/Data+SentryTracing.swift
```diff
@@ -0,0 +1,46 @@
+@_implementationOnly import _SentryPrivate
+
+/// A `Data` extension that automatically tracks read and write operations with Sentry.
+///
+/// - Note: Methods provided by this extension reflect the same functionality as the original `Data` methods,
+///         but they automatically track the operation with Sentry.
+@available(iOS 18, macOS 15, tvOS 18, *)
```
philipphofmann, line 7: `h`: This will make using the API a bit complicated, because our users need to do this
```
let data = "SOME DATA".data(using: .utf8) ?? Data()
try data.write(to: fixture.fileURL)

if #available(iOS 18, macOS 15, tvOS 18, *)  {
    try data.writeWithSentryTracing(to: fixture.fileURL)
} else {
    try data.write(to: fixture.fileURL)
}
```

Instead, I think when they decide using `writeWithSentryTracing`, we need to ensure that we're not creating duplicated spans for the file IO operations on older OS versions. It should be this:
```
let data = "SOME DATA".data(using: .utf8) ?? Data()
try data.write(to: fixture.fileURL)

try data.writeWithSentryTracing(to: fixture.fileURL)
```

Otherwise, all users have to deal with this complexity.

philipphofmann (commented):
I didn't give this a full pass yet cause I found one potential major issue that could require some significant changes to the PR, which could make plenty of comments obsolete.

Sources/Sentry/include/SentryTraceOrigins.h
```diff
@@ -1,11 +1,13 @@
 #import <Foundation/Foundation.h>
 
+// Note: Consider adding new operations to the `SentrySpanOperation` enum in
+// `SentrySpanOperations.swift` instead of adding them here.
```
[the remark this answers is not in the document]
philprime: Will be handled in #4718

Sources/Swift/Integrations/Performance/IO/SentryFileIOTracker+SwiftHelpers.swift
```diff
@@ -0,0 +1,35 @@
...
+        origin: String,
+        method: (_ data: Data, _ url: URL, _ options: Data.WritingOptions) throws -> Void
+    ) throws {
+        guard url.scheme == NSURLFileScheme else {
+            return try method(data, url, options)
+        }
+        let span = self.startTrackingWriting(data, filePath: url.path, origin: origin)
+        try method(data, url, options)
+        if let span = span {
+            self.finishTrackingNSData(data, span: span)
```
philipphofmann, line 32: `h`: This must be in a defer, because otherwise if the method throws an error we never finish the span.

Sources/Swift/Integrations/Performance/IO/Data+SentryTracing.swift
philipphofmann: I tried this in the LoremIpsumViewController in the iOS-Swift sample app, ran it on an simulator with iOS 17, and I got duplicated spans. It don't get duplicates on iOS 18 though.

```
do {
    let data = "hello".data(using: .utf8)
    let tempURL = FileManager.default.temporaryDirectory
    try data?.writeWithSentryTracing(to:tempURL)
} catch {
    SentrySDK.capture(error: error)
}
```

[Simulator iOS 17 transaction](https://sentry-sdks.sentry.io/performance/trace/6ec7c9c9223d43acb9557fc4367c521a/?dataset=transactions&field=title&field=project&field=user.display&field=timestamp&field=replayId&fov=0%2C579&id=18214&name=&node=txn-267192f9c3164028b079df079b14033a&project=5428557&query=&queryDataset=transaction-like&sort=-timestamp&source=discover&statsPeri
CleanShot 2025-01-16 at 14 35 17@2x
od=7d&timestamp=1737034343&topEvents=5&yAxis=count%28%29)

[Simulator iOS 18 transaction](https://sentry-sdks.sentry.io/performance/trace/c27bafae1a434ac4afa1db683769bd57/?dataset=transactions&field=title&field=project&field=user.display&field=timestamp&field=replayId&id=18214&name=&node=span-647270e6ba584ad6&node=txn-07b2fa4487e74e1aafcab583e3060b7e&project=5428557&query=&queryDataset=transaction-like&sort=-timestamp&source=discover&statsPeriod=7d&timestamp=1737034283&topEvents=5&yAxis=count%28%29)
CleanShot 2025-01-16 at 14 36 05@2x

Sources/Swift/Integrations/Performance/IO/SentryFileIOTracker+SwiftHelpers.swift
philprime: In this case it is possible, because `data` is available.

Does the same apply to `measureReadingData`? The `finishTrackingNSData` requires a non-nil `Data`, which is not available if reading fails.

The Objective-C implementation calls [self finishTrackingNSData:data span:span]; in every case, even in an error where no `NSData`. It seems to me there is no test case to handle the case when an error is thrown.

Based on the [documentation](https://developer.apple.com/documentation/foundation/nsdata/1411145-initwithcontentsoffile) I assume it returns a zero bytes data, therefore I could use empty data as a nil fallback

Sources/Swift/Integrations/Performance/IO/Data+SentryTracing.swift
philprime: Thanks for looking into this, going to investigate this.

Sources/Swift/Integrations/Performance/IO/SentryFileIOTracker+SwiftHelpers.swift
philipphofmann: > Does the same apply to measureReadingData?

Yes, it should. You call try in that method, so it could exit before ending the span.

Sources/Swift/Integrations/Performance/IO/Data+SentryTracing.swift
philprime: NSData swizzling was interfering, I added unit tests to make sure only one span is created.

Sources/Swift/Integrations/Performance/IO/SentryFileIOTracker+SwiftHelpers.swift
philprime: Added a defer with `file.size` set to `0`. If you prefer to not set the `file.size` instead, let me know, I would leave it like that for now.

philprime:
I've now retried the failing workflow 7 times and keeps failing at different places, but they seem unrelated to my changes (i.e. ANR or timeouts).

@philipphofmann @brustolin would appreciate your assistance in debugging the CI failures.

philipphofmann:
About CI failures: We see timeouts failing all the time for GH actions. When running those tests locally, they are usually rock solid. This is giving us headaches already for years. The weird thing often is that sometimes we don't see these failures for a while and then they come back even when we don't change the code or the tests. I think we could consider defining a constant for test expectations timeouts which we set to 10 or 30 seconds just for CI. If you have a better idea please let me know. Another idea is to simply disable all the flaky tests, because it's better to have fewer tests we can trust than flaky tests.

philprime:
So @philipphofmann should we keep this PR open until we have stabilized CI?

Open PRs could be useful to test CI improvements, but it also keeps us from merging this.

philipphofmann:
> So @philipphofmann should we keep this PR open until we have stabilized CI?
> 
> Open PRs could be useful to test CI improvements, but it also keeps us from merging this.

Flaky CI shouldn't stop us from merging this PR. We should address these issues separately.

Sources/Swift/Integrations/Performance/IO/Data+SentryTracing.swift
```diff
@@ -0,0 +1,55 @@
...
+    /// Initialize a `Data` with the contents of a `URL`, automatically tracking the operation with Sentry.
+    ///
+    /// - parameter url: The `URL` to read.
+    /// - parameter options: Options for the read operation. Default value is `[]`.
+    /// - throws: An error in the Cocoa domain, if `url` cannot be read.
+    init(contentsOfUrlWithSentryTracing url: URL, options: Data.ReadingOptions = []) throws {
+        let tracker = SentryFileIOTracker.sharedInstance()
+        // Using the bridging of `Data` to `NSData` caused issues on older versions of macOS.
+        // Therefore we do not use the `measureNSData` method.
+        if #available(iOS 18, macOS 15, tvOS 18, *) {
```
philipphofmann, line 18: `h`: I'm unsure about this approach. It still could occur that some OS versions switch back to bridging NSData, and then users would get duplicated spans. This also has the downside that people can't use this on older iOS versions when they, for example, disable swizzling. Instead, it would be great if we could implement some deduplication logic for duplicated spans. I still need to figure out how we can do this properly, and will get back to you.

philipphofmann (commented):
After we decide on the deduplication logic on spans, I will give this another review cause this can have a significant impact on the PR.

Sources/Swift/Integrations/Performance/IO/Data+SentryTracing.swift
philipphofmann: We agreed in Slack to skip a deduplication logic because it's complicated to get it right, it's a lot of work, and most people won't need it. Instead, we should allow the use of manual API before iOS 18. If users want file IO spans, they need to use the manual API anyway, and they can turn off auto file IO instrumentation in general. With this, they avoid having duplicates before iOS 18. We need to mention that people should consider turning off auto file IO instrumentation when using the manual API.

philprime:
@philipphofmann I added an experimental option to disable swizzling for `NSData` and added manual methods for `Data` and `FileManager`. I believe this PR is now ready for review.

philprime:
We've decided in Slack that this PR is rather large and should be split into multiple ones (as far as possible).

philipphofmann:
@philprime does it still make sense to keep this PR open? Can't we close this?

philprime:
I want this to stay open so I can verify after everything is merged that I did not miss anything

philprime:
Superseded by PRs in description

PR #4619, opened 2024-12-11
chore: Comment on released issues

philipphofmann:
When a release is done the Changelog will be parsed for issue numbers. Then a comment will be added to those issues that a fix has been released.

I took this from the JS: https://github.com/getsentry/sentry-javascript/blob/develop/.github/workflows/release-comment-issues.yml 

Python also added it here https://github.com/getsentry/sentry-python/pull/3866

#skip-changelog


philprime (approved):
LGTM

PR #4621, opened 2024-12-11
chore: Removing unused argument

brustolin:
SentryViewPhotographer had an unused argument for  SentryViewPhotographer.image()

_#skip-changelog_


brustolin:
@vaind I think this breaks session replay for Flutter. Can you take a look?

PR #4634, opened 2024-12-13
fix: Add swizzling for NSFileManager.createFileAtPath

philprime:
## :scroll: Description

This PR fixes the automatic tracking of `file.write` spans when using `NSFileManager. createFileAtPath`. 

The changes are split of from #4605.

Before macOS 15, iOS 18 and tvOS 18 the implementation would use the already swizzled methods of `NSData`. As the internal implementation has changed, the `NSFileManager` must be swizzled directly.

See #4546 for full discussion.

## :pencil: Checklist

## :crystal_ball: Next steps

- [x] Merge #4647 as unswizzling is required


.github/file-filters.yml
```diff
@@ -5,6 +5,7 @@ high_risk_code: &high_risk_code
   - 'Sources/Sentry/SentryNetworkTracker.m'
   - 'Sources/Sentry/SentryUIViewControllerSwizzling.m'
   - 'Sources/Sentry/SentryNSDataSwizzling.m'
+  - 'Sources/Sentry/SentryNSFileManagerSwizzling.m'
```
philipphofmann, line 8: 🥇

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixes
 
+- Fix span recording for `NSFileManager.createFileAtPath` starting with iOS 18, macOS 15 and tvOS 18 (#4634)
```
philipphofmann, line 11: `m`: I think we should mention that you have to enable a feature for this, which is still experimental and will be enabled by default in the future.

Tests/SentryTests/Integrations/Performance/IO/SentryFileIOTrackerTests.swift
```diff
@@ -275,6 +275,25 @@ class SentryNSDataTrackerTests: XCTestCase {
...
+            FileAttributeKey.size: 123
+        ], method: { path, data, attributes in
+            methodPath = path
+            methodData = data
+            methodAttributes = attributes
+            return true
+        })
+        XCTAssertEqual(methodPath, fixture.filePath)
+        XCTAssertEqual(methodData, fixture.data)
+        XCTAssertEqual(methodAttributes?[FileAttributeKey.size] as? Int, 123)
```
philipphofmann, line 294: `m`: This test needs to validate the created span.

Sources/Sentry/SentryNSFileManagerSwizzling.m
```diff
@@ -0,0 +1,84 @@
...
+#import "SentrySwizzle.h"
+#import "SentryThreadInspector.h"
+#import <SentryLog.h>
+#import <objc/runtime.h>
+
+@interface SentryNSFileManagerSwizzling ()
+
+@property (nonatomic, strong) SentryFileIOTracker *dataTracker;
+
+@end
```
philipphofmann, line 20: `m`: What drove you to create an extra class for `SentryNSFileManagerSwizzling`? Wouldn't it be better to have one class for swizzling file IO operations for tracing? Now, we have two different instances of the `dataTracker` and some duplicated logic.

Sources/Sentry/SentryNSFileManagerSwizzling.m
```diff
@@ -0,0 +1,84 @@
...
+
+// SentrySwizzleInstanceMethod declaration shadows a local variable. The swizzling is working
+// fine and we accept this warning.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshadow"
++ (void)swizzleNSFileManager
+{
+    // Before iOS 18.0 and macOS 15.0 the NSFileManager used NSData.writeToFile internally,
+    // which was tracked using swizzling of NSData. This behaviour changed, therefore the
+    // file manager needs to swizzled for later versions.
```
philipphofmann, line 59: `l`: Maybe we should even link the specific PR of Swift here? https://github.com/swiftlang/swift-foundation/pull/410?

Sources/Sentry/SentryNSFileManagerSwizzling.m
```diff
@@ -0,0 +1,84 @@
...
+    if (@available(iOS 18, macOS 15, *)) {
+        SEL createFileAtPathContentsAttributes
```
philipphofmann, line 61: `m`: We need to write a unit test that validates that this is a NoOp before iOS 18.

Sources/Sentry/SentryNSFileManagerSwizzling.m
```diff
@@ -0,0 +1,84 @@
...
+    self.dataTracker = [[SentryFileIOTracker alloc]
+        initWithThreadInspector:[[SentryThreadInspector alloc] initWithOptions:options]
+             processInfoWrapper:[SentryDependencyContainer.sharedInstance processInfoWrapper]];
+    [self.dataTracker enable];
+
+    if (!options.experimental.enableFileManagerSwizzling) {
+        SENTRY_LOG_DEBUG(@"Experimental auto-tracking of FileManager is disabled")
+        return;
+    }
+    [SentryNSFileManagerSwizzling swizzleNSFileManager];
```
philipphofmann, line 43: `h`: We must only swizzle the file manager on iOS 18, macOS 15 and above, as swizzling can always break something.

Sources/Sentry/SentryNSFileManagerSwizzling.m
```diff
@@ -0,0 +1,84 @@
...
+// SentrySwizzleInstanceMethod declaration shadows a local variable. The swizzling is working
+// fine and we accept this warning.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshadow"
++ (void)swizzleNSFileManager
+{
+    // Before iOS 18.0 and macOS 15.0 the NSFileManager used NSData.writeToFile internally,
+    // which was tracked using swizzling of NSData. This behaviour changed, therefore the
+    // file manager needs to swizzled for later versions.
+    if (@available(iOS 18, macOS 15, *)) {
```
philipphofmann, line 60: `m`: What about tvOS, visionOS and watchOS? Can we also specify the exact versions for these?

philipphofmann (commented):
I added a few comments, and we should be able to enable the previously disabled failing tests again in this PR.

CHANGELOG.md
philprime: Added context how to enable it

Tests/SentryTests/Integrations/Performance/IO/SentryFileIOTrackerTests.swift
philprime: Added span assertion

Sources/Sentry/SentryNSFileManagerSwizzling.m
philprime: The main reason was future-proofing. NSFileManager offers many other methods which could be useful to auto-trace further down the road, i.e.:

- [contentsOfDirectoryAtPath:error:](https://developer.apple.com/documentation/foundation/nsfilemanager/1414584-contentsofdirectoryatpath) → `file.read`?
- [createDirectoryAtPath:withIntermediateDirectories:attributes:error:](https://developer.apple.com/documentation/foundation/nsfilemanager/1407884-createdirectoryatpath) → `file.write`?
- [removeItemAtURL:error:](https://developer.apple.com/documentation/foundation/nsfilemanager/1413590-removeitematurl) → `file.delete`
- [copyItemAtURL:toURL:error:](https://developer.apple.com/documentation/foundation/nsfilemanager/1412957-copyitematurl) → `file.rename`

That's why I believe its good to have separate swizzled classes into individual files.

But I do agree that the setup of multiple data trackers are not ideal. Two options:

1. Setup a shared data tracker in in `SentryFileIOTrackingIntegration.m` and pass it as an argument to `SentryNSDataSwizzling` and `SentryNSFileManagerSwizzling`.
2. Add the data tracker to `SentryDependencyContainer.m`

Sources/Sentry/SentryNSFileManagerSwizzling.m
philprime: As mentioned in the other PR, there might be additional swizzling coming up in future OS updates. Therefore the availability checks are inside `swizzleNSFileManager` and not in `startWithOptions`

Sources/Sentry/SentryNSFileManagerSwizzling.m
philprime: Thanks for pointing that out. Missed adding it here.

I don't believe there is a difference between i.e. `iOS 18.0.0` and `iOS 18` so I didn't declare them in such detail.

brustolin (approved):
LGTM

Sources/Sentry/SentryNSFileManagerSwizzling.m
philprime: I added unit tests for `SentryNSFileManagerSwizzling` to test the behaviour on older OS and newer OS, as well testing the experimental flag

Sources/Sentry/SentryNSFileManagerSwizzling.m
philipphofmann: I would prefer option 1.

Sources/Sentry/SentryNSFileManagerSwizzling.m
philipphofmann: Ah I missed that check. Thanks 👍

philipphofmann (approved):
LGTM, thanks.

philprime:
So right now after adding the unit tests for enabling and disabling the swizzling via the experimental feature flag, the tests in `SentryFileIOTrackingIntegrationObjCTests.m` are causing the tests in `SentryNSFileManagerSwizzlingTests.swift` to partially fail.

I currently believe we need to unswizzle methods in the `uninstall` of integrations, so I'll look into that to get this fixed.

philipphofmann:
> So right now after adding the unit tests for enabling and disabling the swizzling via the experimental feature flag, the tests in `SentryFileIOTrackingIntegrationObjCTests.m` are causing the tests in `SentryNSFileManagerSwizzlingTests.swift` to partially fail.
> 
> I currently believe we need to unswizzle methods in the `uninstall` of integrations, so I'll look into that to get this fixed.

That sounds like side effects which you won't be able to solve with unswizzling. Let's have a chat about this, as this sounds like something that will bite us later.

philprime:
@philipphofmann I already narrowed the issue down to the swizzling. I went ahead and created a local Xcode scheme only running tests in these files:

Screenshot 2024-12-19 at 16 39 34

Without unswizzling, if I run the tests individually, they all pass. If I run all the tests, they fail. I added breakpoints and manually tested, and in test cases where swizzling *should not* happen, the methods were still swizzled from previous tests.

Maybe the whole unswizzling/uninstall should only be implemented in the tests, but that's probably better to be discussed in #4647

PR #4636, opened 2024-12-13
feat: rename strncpy to strlcpy; added tests

### LINKED ISSUE #2783: Replace strncpy with strlcpy
### Description

Our codebase uses [strncpy](https://github.com/search?q=repo%3Agetsentry%2Fsentry-cocoa+strncpy&type=code), which is unsafe according to the [Apple Secure Coding Guide](https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Articles/BufferOverflows.html#//apple_ref/doc/uid/TP40002577-SW1). We should replace it with `strlcpy`.  Worth investigating if KSCrash has already fixed this.

If we can simply replace strncpy with strlcpy, cause the code has test coverage we should do this quickly. If replacing is a bit complicated, needs refactoring to make things testable, we should reconsider the priority.

Related security issue: [CWE-676](https://cwe.mitre.org/data/definitions/676.html).

philprime:
## :scroll: Description

Renames all occurences of `strncpy` with `strlcpy` as explained in #2783.

I looked all the occurences and tried to create additional unit tests where applicable. Some of the cases using `strlcpy` have larger buffer sizes set, i.e. `SentryCrashFU_MAX_PATH_LENGTH` is set to `500`, but I was not able to file with a longer name because the OS throws NSPOSIXErrorDomain, code 63, indicating that the file system does not support such long files.

Closes #2783

## :pencil: Checklist

## :crystal_ball: Next steps

- [ ] Validate if additional tests are necessary


armcknight:
One thing I'd recommend here is to wrap whatever function we actually want to use in a function we control, to make it easier to test and potentially change later. It could be the case that `strncpy` is ok in some places, but not others where we'd then want `strlcpy`. (I wrote this paragraph before taking a deeper look at the code, and it looks like we used to have such a function, and we do have places in the code that emulate strlcpy and others that don't. So, we would need to validate the behavior after making this change).

My understanding after reading the docs is that `strncpy` is fine to use for copying from string buffers to characters arrays, but doesn't guarantee null termination (although does fill in the remainder of the array if the size of the source buffer is lesser than the length of the destination char array), which, I don't know why that's actually needed in a character array aside from copying it back to a buffer, which I don't think we do with any of these.

from https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/strncpy.3.html#//apple_ref/doc/man/3/strncpy:

> The strncpy() function copies at most n characters from s2 into s1.  If s2 is less than n characters long, the remainder of s1 is filled with `\0' characters.  Otherwise, s1 is not terminated.

`strlcpy` is guaranteed to write at least one null terminator, and so that must be accounted for in the length parameter passed into it. I think straight replacement will result in one less character of actual data being written to accommodate that, so we might have to add a `+1` to each place here? 

from https://linux.die.net/man/3/strlcpy:

> The strlcpy() function copies up to size - 1 characters from the NUL-terminated string src to dst, NUL-terminating the result.

I guess this is what this line of code is trying to replicate: https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-de0cc4487a50688dff672b8cd13b3901fdbf09abe93afa34b54bd561d1796fe8L1260 and I also just saw we actually already had another function to do this: https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-898d2165aedeeac3e2023d294e4bffee79b96b5c04d9d9d5be7e7593d4eb9aeeL21-L27

This is why a wrapper function is my preference, we can hide all these details and keep things consistent. If we have places that move data from char arrays back to buffers, we should also have a wrapper function for that as well.

philprime:
Thanks @armcknight for reviewing the PR.
I haven't had a chance yet to add more written context, therefore the PR is marked as draft.

As requested in the issue and mentioned by you, we need to make sure the changes do not break the SDK. That's why I went ahead and started adding additional unit tests.

> strlcpy is guaranteed to write at least one null terminator, and so that must be accounted for in the length parameter passed into it. I think straight replacement will result in one less character of actual data being written to accommodate that, so we might have to add a +1 to each place here?

I was thinking of the same, and yes we might need to extend the buffers at some places by one character, if it actually uses the full buffer length.

I also found the function `strncpy_safe`, but looking at the implementation it seems to work mimic the behaviour of `strlcpy` so I don't see a point in maintaining a duplicate implementation.

https://github.com/getsentry/sentry-cocoa/blob/419f1d49be4b60e498f43f85e8ed9b50d7f975dd/Sources/SentryCrash/Reporting/Filters/Tools/SentryStringUtils.h#L21-L27


**TL;DR: I added todos to the description what is left to do.**

In total I found 7 uses of `strlcpy` and `strncpy`, so I'll quickly add my thoughts for each so far:

1. [`int sentry_asyncLogSetFileName(const char *filename, bool overwrite)`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-e3adbeedfe33eb95c1a435565bd543cd1f9340c411270e831453b3648eaf036c):

- Called by `SentryAsyncLogWrapper` with the full path of the `async.log` file, which is placed in the result of `NSSearchPathForDirectoriesInDomains`. Therefore the length of the path is controlled by the OS.
- Currently no maximum file path length is enforced → therefore the current implementation using `strncpy` will write up to 1024 characters leaving it unterminated, with `strlcpy` up to 1023 characters but definitely terminated.
- Copies the given filename to the `g_logFileName`, which is a 1024 char buffer.
- The `g_logFileName` is used by `addTextLinesFromFile(..)` in `SentryCrashReport.c` which then uses the path with `open` to create a file descriptor.
- I don't think the `g_logFileName` should ever be unterminated, so we should definitely replace it.

2. [`static void CPPExceptionTerminate(void)`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-4e41fb589178ce0a39a00f90969cb8a6318a2cd7d48778521aff3d06c2dff068):

- Uses the `strncpy_safe` to copy the exception description into a 1000 char long `descriptionBuff`.
- The description should be null terminated.
- As mentioned above, the `strncpy_safe` method tries to do the same as `strlcpy`, so I believe we should replace this.
- I will try to create a unit test with a description longer than 1000 chars, to verify the expected behaviour.

3. [`static void onCrash(struct SentryCrash_MonitorContext *monitorContext)`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-dbb8caac8f4cb74b4c4883bef2b2f5a4f2edd245cab062d2488d90db0b50488c):

- Generates a crash report file path using `sentrycrashcrs_getNextCrashReportPath`, then writes it to a local variable `crashReportFilePath` with a size of `SentryCrashFU_MAX_PATH_LENGTH = 500` characters.
- `sentrycrashcrs_getNextCrashReportPath` uses `getCrashReportPathByID` to generate the path
- `getCrashReportPathByID` uses `snprintf` which also has `SentryCrashCRS_MAX_PATH_LENGTH` defined as the max length.
- The generated path will be null terminated, from https://linux.die.net/man/3/snprintf: 

> The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.

- Therefore `crashReportFilePath` will always be null terminated and can safely be copied to `g_lastCrashReportFilePath` without changing the buffer size.
- I added tests in [`SentryCrashCTests.swift`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-9a142f88cf8a70de508a88b109206abe82c66b08771486012110d1969e7c65d0)

4. [`void sentrycrashreport_writeRecrashReport(..., const *path, ...)`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-1623dc49c0b0bba9eba2448be0c9887850ebac388800b9c4d0d85da3a182038a):

- This one I still need to verify/test, because it uses `strncpy` to copy `path` to a `tempPath` which has a length limit of `SentryCrashFU_MAX_PATH_LENGTH`, but then replaces the last characters with a `.old` extension
- I tried to write unit tests in [`SentryCrashCTests.swift`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-9a142f88cf8a70de508a88b109206abe82c66b08771486012110d1969e7c65d0) but kind-off failed because it tries to create a monitor context from thread data, which I do not fully understand yet.

https://github.com/getsentry/sentry-cocoa/blob/419f1d49be4b60e498f43f85e8ed9b50d7f975dd/Sources/SentryCrash/Recording/SentryCrashReport.c#L1474-L1479

5. [`static bool increaseDepth(FixupDepth *context, const char *name)`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-8f48191cce8b970b9434bc86de445c680c4a2b1f8fd1d1d6039e466c17917836):

- Copies the given `name` to the `context->objectPath[context->currentDepth]`
- If `name` is NULL it sets the `context->objectPath[context->currentDepth]` to `\0`, therefore an empty but terminated string
- Based on that, I believe it is safe to assume that the `context->objectPath[context->currentDepth]` should always be terminated, and it is safe to use `strlcpy` instead of `strncpy`.
- The object path length is `MAX_NAME_LENGTH = 100` so we might have to increase the buffer size by one, as there could be names having exactly that length.

https://github.com/getsentry/sentry-cocoa/blob/419f1d49be4b60e498f43f85e8ed9b50d7f975dd/Sources/SentryCrash/Recording/SentryCrashReportFixer.c#L54-L68

6. [`static bool deletePathContents(const char *path, bool deleteTopLevelPathAlso)` in `SentryCrashFileUtils.c`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-40caed0a92c62c4bc7ffbddcb3b34ce951b0d5a9868b16c35605371e76153e86):

- The given `path` is passed to `snprintf` which truncates it to `SentryCrashFU_MAX_PATH_LENGTH - 1 - 1` characters, then appending `/` and `\0` and written to `pathBuffer`.
- The address of the end of `pathBuffer` is set to `pathPtr` to calculate the remaining buffer length and set it to `pathRemainingLength`.
- `strncpy` will use the the `pathRemainingLength` as the limit of characters copied to `pathPtr` and therefore to `pathBuffer`. 
- The `pathBuffer` is then used to recursively call `deletePathContents` which uses `snprintf` again to terminate the string, therefore we can use `strlcpy` instead of `strncpy` because it will be terminated anyways.

https://github.com/getsentry/sentry-cocoa/blob/419f1d49be4b60e498f43f85e8ed9b50d7f975dd/Sources/SentryCrash/Recording/Tools/SentryCrashFileUtils.c#L155-L167

- I added additional unit tests in [`SentryCrashFileUtils_Tests.m`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-3aa6dc1aea37941079b6bdee18cffae367124c0b25be71e99c3ca2123801fedb)
- ⚠️ The current implementation might actually not work if the path is longer than `SentryCrashFU_MAX_PATH_LENGTH`. I will add another unit test to verify.

7. [`decodeElement()` in `SentryCrashJSONCodec.c`](https://github.com/getsentry/sentry-cocoa/pull/4636/files#diff-de0cc4487a50688dff672b8cd13b3901fdbf09abe93afa34b54bd561d1796fe8):

- The current implementation performs manual null termination while using `strncpy`, so I believe it to be safe to replace it with `strlcpy` instead:

https://github.com/getsentry/sentry-cocoa/blob/419f1d49be4b60e498f43f85e8ed9b50d7f975dd/Sources/SentryCrash/Recording/Tools/SentryCrashJSONCodec.c#L1259-L1260

CHANGELOG.md
```diff
@@ -10,19 +16,6 @@
...
-### Improvements
-
-- Improve compiler error message for missing Swift declarations due to APPLICATION_EXTENSION_API_ONLY (#4603)
-- Mask screenshots for errors (#4623)
-- Slightly speed up serializing scope (#4661)
-
-### Internal
-
-- Remove loading `integrations` names from `event.extra` (#4627)
-- Add Hybrid SDKs API to add extra SDK packages (#4637)
```
armcknight, line 25: I think I resolved this conflict incorrectly. Did we want these items from the beta release's changes repeated in the 8.43.0 RC?

Tests/SentryTests/Recording/SentryCrashCTests.swift
```diff
@@ -0,0 +1,282 @@
...
+        // Calling the handle exception will trigger the onCrash handler
+        sentrycrashcm_handleException(&monitorContext)
+
+        // -- Assert --
+        let reportUrls = try FileManager.default
+            .contentsOfDirectory(atPath: expectedReportsDir.path)
+        XCTAssertEqual(reportUrls.count, 1)
+    }
+
+    func testOnCrash_notCrashedDuringCrashHandling_installFilePathTooLong_shouldNotWriteToDisk() {
```
armcknight, line 39: Good on you for encoding behavior in a test like this, but IMO this situation is something we should fix ASAP if it's happening in the wild. We should always expect to be able to deliver crash reports, and guard against situations that would prevent that and work around them whenever possible. At the very least a warning message should be available for end users to also potentially work around it, like if they provide a custom cache path that is too long. The only other thing I can see that would cause this is a bundle id that is too long (we should also only use the bundle id for mac apps; ios/tvos/visionos are all self-contained files trees per app)

Tests/SentryTests/Recording/SentryCrashCTests.swift
```diff
@@ -0,0 +1,282 @@
...
+        var installPath = installDir
+            .path
+            .cString(using: .utf8)!
+        let expectedReportsDir = installDir
+            .appendingPathComponent("Reports")
+
+        // Smoke test the existence of the directory
+        XCTAssertFalse(FileManager.default.fileExists(
+            atPath: expectedReportsDir.path
+        ))
```
armcknight, line 20: Gratuitous use of vertical space. I'd change all such code:
```suggestion
        var appName = "SentryCrashCTests".cString(using: .utf8)!
        let installDir = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent("SentryCrashCTests-\(UUID().uuidString)")
        var installPath = installDir.path.cString(using: .utf8)!
        let expectedReportsDir = installDir.appendingPathComponent("Reports")

        // Smoke test the existence of the directory
        XCTAssertFalse(FileManager.default.fileExists(atPath: expectedReportsDir.path))
```

armcknight (approved):
Good work to make the changes and write the tests that cover a lot of implementation code.

We use a common pattern of `sut` (system under test) which you might be able to use to refactor some of the tests. I know there's a tension between DRY and self-contained test logic. Figured I'd just offer it up as food for thought.

Overall I think the changes are fine so this has my approval. I merely offer some suggestions and comments along the way.

CHANGELOG.md
philprime: You are correct, there have been a couple of releases since I opened up this PR. I'll fix this when getting ready to merge.

Sources/SentryCrash/Recording/Tools/SentryCrashJSONCodec.c
```diff
@@ -1256,9 +1256,18 @@ decodeElement(const char *const name, SentryCrashJSONDecodeContext *context)
...
             return SentryCrashJSON_ERROR_DATA_TOO_LONG;
         }
+        // Must use strncpy instead of strlcpy, because of the following reason:
+        //
+        //   Also note that strlcpy() and strlcat() only operate on true ''C'' strings.
+        //   This means that for strlcpy() src must be NUL-terminated and for strlcat()
+        //   both src and dst must be NUL-terminated.
+        //
+        // Source: https://linux.die.net/man/3/strlcpy
         strncpy(context->stringBuffer, start, len);
```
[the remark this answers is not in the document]
philprime: During testing I went ahead an replaced it with `strlcpy` first, without changing any of the unit tests. It kept crashing with `EXC_BAD_ACCESS`, even though `strncpy` and even `memcpy` did not. I also considered to replace the existing implementation with `memcpy` but decided to stick with it, simply to not change a running system without exact reason.

Thanks for the cleanup of `strlcat`, I left it in there so that it's actually a quote, but I also see the missing value of additional text.

Tests/SentryTests/Recording/SentryCrashCTests.swift
philprime: I had a quick chat with @philipphofmann yesterday, I still need to validate if a too long file path can actually be passed down from the SentrySDK options using `cacheDirectoryPath` to the `SentryCrashC`, or if it causes issues before that.

He also mentioned a warning should be sufficient, but at this time I already created this test case to actually test the behavior with invalid paths.

In general I still believe we should also add tests to pin down expected negative behavior (if possible), so even if an edge case can hardly ever happen, we do have documentation that it can happen. If it then turns out to be an actual issue, we can change test-case, and fix the implementation until it passes again.

Tests/SentryTests/SentryCrash/SentryCrashFileUtils_Tests.m
```diff
@@ -622,4 +622,149 @@ - (void)testReadLineFromFD
     XCTAssertTrue(bytesRead == 0, @"");
 }
 
+- (void)testDeleteContentsOfPath_canNotDeletePath_shouldNotDeleteTopLevelPath
```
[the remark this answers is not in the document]
philprime: In all honesty I don't see much of a point in having a wrapper around `strlcpy` as we should test it's usage, rather than it's implementation.

For this PR I tried to add test cases for all places where I changed from `strncpy` to `strlcpy`, therefore also `sentrycrashfu_deleteContentsOfPath` defined in `SentryCrashFileUtils.h`. As we want to have high test coverage anyways, I decided to add test cases for all code paths.

Tests/SentryTests/SentryCrash/SentryCrashMonitor_CppException_Tests.mm
```diff
@@ -24,29 +27,135 @@ - (void)setUp
...
 }
 
 - (void)testCallTerminationHandler_Enabled
 {
-
+    // -- Arrange --
     std::set_terminate(&testTerminationHandler);
 
-    SentryCrashMonitorAPI *api = sentrycrashcm_cppexception_getAPI();
+    api = sentrycrashcm_cppexception_getAPI();
```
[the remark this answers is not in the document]
philprime: You are right, for some reason I thought this had to be after `std::set_terminate` which is different per test case

Tests/SentryTests/SentryCrash/SentryCrashMonitor_CppException_Tests.mm
```diff
@@ -24,29 +27,135 @@ - (void)setUp
...
+    if (api != NULL) {
+        api->setEnabled(false);
+    }
+    sentrycrashcm_setEventCallback(NULL);
+    capturedExceptionContextCrashReason = NULL;
+}
 
+- (void)testCallTerminationHandler_NotEnabled
+{
+    // -- Arrange --
```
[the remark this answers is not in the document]
philprime: High chance that article actually inspired me to use that pattern. Macros etc. could give this even more structure, but there are always cases where that could make it more complicated/harder to maintain.

Tests/SentryTests/SentryCrash/SentryCrashFileUtils_Tests.m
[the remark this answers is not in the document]
armcknight: Yep totally fair 👍🏻

Tests/SentryTests/SentryCrash/SentryCrashMonitor_CppException_Tests.mm
```diff
@@ -24,29 +27,135 @@ - (void)setUp
...
 
+    // -- Act --
     sentrycrashcm_cppexception_callOriginalTerminationHandler();
 
+    // -- Assert
     XCTAssertTrue(terminateCalled);
+}
+
+void
+testHandleExceptionHandler(struct SentryCrash_MonitorContext *context)
```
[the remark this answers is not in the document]
philprime: I agree, renamed the newly introduced and the existing handler to mock...

philipphofmann (commented):
I have nothing to add as @armcknight knowledge on C is way more extensive than mine.

philipphofmann (commented):
I think a few smoke tests, if the crash reports end up successfully in Sentry via the sample app via Testflight, would reduce the chance of bugs. We should test this with the following buttons in the iOS-Swift sample app: Throw FatalError, Crash the app, force unwrap optional, and an unhandled CPP exception, which isn't in the sample app I believe. Finally, we should do a beta release with our dogfooding app before releasing this.

philprime:
These also show up in the Testflight crashes as expected:

- Throw FatalError ✅
- Force unwrap optional ✅
- Crash the app ✅

LGTM?

Tests/SentryTests/Recording/SentryCrashCTests.swift
armcknight: Definitely 👍🏻

.github/workflows/testflight.yml
```diff
@@ -14,6 +14,7 @@ on:
   pull_request:
     paths:
       - '.github/workflows/testflight.yml'
+  workflow_dispatch:
```
armcknight, line 17: Is this a vestige of some test/experimentation?
```suggestion
```

armcknight (approved):
LGTM, nice work @philprime !

.github/workflows/testflight.yml
philprime: Not really, it allows us to trigger the uploading to Testflight directly from the GitHub web UI as soon as the line is available on main.

PR #4637, opened 2024-12-15
feat: Add extra SdkInfo packages

krystofwoldrich:
## :scroll: Description

- This PR adds ability to add extra packages to SdkInfo.

- This PR exposes API for hybrid SDKs to add packages.

- This PR replaces https://github.com/getsentry/sentry-cocoa/pull/4598

tests

## :pencil: Checklist


SentryTestUtils/ClearTestState.swift
```diff
@@ -67,5 +67,8 @@ class TestCleanup: NSObject {
         #endif // os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
 
         sentrycrash_scopesync_reset()
+
+        SentrySdkInfo.resetPackageManager()
+        SentrySdkInfo.clearExtraPackages()
```
philipphofmann, line 72: `m`: Do we really need this here? Although I added this some time ago, I'm not a fan of this global reset, and I would prefer that we use it sparingly. I would prefer if we could tear down the global state of the SDKInfo in the tests using it.

Sources/Sentry/include/SentrySdkInfo.h
```diff
@@ -24,6 +24,9 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SentrySdkInfo : NSObject <SentryInternalSerializable>
 SENTRY_NO_INIT
 
++ (void)addPackageName:(NSString *)name version:(NSString *)version;
++ (NSMutableSet<NSDictionary<NSString *, NSString *> *> *)getExtraPackages;
```
philipphofmann, line 28: `m`: Nobody uses this method.
```suggestion
```

Sources/Sentry/SentrySdkInfo.m
```diff
@@ -29,16 +29,95 @@ typedef NS_ENUM(NSUInteger, SentryPackageManagerOption) {
...
+            return [extraPackages mutableCopy];
+        }
+    }
+}
+
+#if TEST || TESTCI
++ (void)clearExtraPackages
+{
+    extraPackages = [[NSSet alloc] init];
+}
```
philipphofmann, line 73: `m`: It would seems a bit cleaner to me if we extract the extra packages code into its own class. We could even write it in Swift, similar to `SentryEnabledFeaturesBuilder`.

philipphofmann (commented):
A few comments. Not far from LGTM.

SentryTestUtils/ClearTestState.swift
krystofwoldrich: Since we have this method now I think it should, if possible, correctly reset the state, otherwise it's very confusing.

Sources/Sentry/SentrySdkInfo.m
krystofwoldrich: We can't write it in Swift because the test suit (data comparisons) will break on the non-deterministing dict entries order after serialization.

SentryTestUtils/ClearTestState.swift
krystofwoldrich: If it's already broken, than I agree it doesn't make sense to add to it. But if till now it was correct, we should add there the packages reset.

Sources/Sentry/include/SentrySdkInfo.h
krystofwoldrich: Thanks, yes, it's only used in tests, I've updated the test and removed it from the interface.

krystofwoldrich: [`792ba11` (#4637)](https://github.com/getsentry/sentry-cocoa/pull/4637/commits/792ba118471a33b309ff31da0a9ad190f36a806e)

Sources/Sentry/SentrySdkInfo.m
philipphofmann: > non-deterministing dict entries order after serialization

Wait what? Can't we fix the data comparison? That sounds stupid 😢

krystofwoldrich: We can, but it doesn't make sense to block this PR by refactoring the tests.

krystofwoldrich:
@philipphofmann I've updated the PR, it's ready for another pass.

Sources/Sentry/include/SentryExtraPackages.h
```diff
@@ -0,0 +1,23 @@
...
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentryExtraPackages : NSObject
+SENTRY_NO_INIT
+
++ (void)addPackageName:(NSString *)name version:(NSString *)version;
++ (NSMutableSet<NSDictionary<NSString *, NSString *> *> *)getPackages;
```
philipphofmann, line 15: `l`: I guess this can be a set.
```suggestion
+ (NSSet<NSDictionary<NSString *, NSString *> *> *)getPackages;
```

Sources/Sentry/SentryExtraPackages.m
```diff
@@ -0,0 +1,47 @@
...
+    if (name == nil || version == nil) {
+        return;
+    }
+
+    @synchronized(extraPackages) {
+        NSDictionary<NSString *, NSString *> *newPackage =
+            @{ @"name" : name, @"version" : version };
+        if (extraPackages == nil) {
+            extraPackages = [[NSSet alloc] initWithObjects:newPackage, nil];
+        } else {
```
philipphofmann, line 21: `m`: Instead of checking for nil here, you can add a static init method, as we have it here:

https://github.com/getsentry/sentry-cocoa/blob/716a5b0ea1d64474b2285f5799a81ff0faac918f/Sources/Sentry/SentrySDK.m#L71-L80

Sources/Sentry/SentryExtraPackages.m
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
+
++ (NSMutableSet<NSDictionary<NSString *, NSString *> *> *)getPackages
+{
+    @synchronized(extraPackages) {
+        if (extraPackages == nil) {
+            return [[NSMutableSet alloc] init];
+        } else {
+            return [extraPackages mutableCopy];
```
philipphofmann, line 33: `l`: If you change the `NSMutableSet` to a `NSSet`.
```suggestion
            return [extraPackages copy];
```

philipphofmann (approved):
LGTM

Sources/Sentry/include/SentryExtraPackages.h
krystofwoldrich: I've done this on purpose to avoid one extra copy. We create a copy when getting the packages and then we need to add the SDK Package (https://github.com/getsentry/sentry-cocoa/pull/4637/files#diff-def4a3cae1e4126909fdbd84870ad3fb7b9922755d490980cc1e6f1a756a5ad9R43) so by making this mutable we avoid making another copy.

PR #4639, opened 2024-12-16
feat: Show session replay options as replay tags

brustolin:
## :scroll: Description

Send the options used for session replay in the replay tag event.

image

Unit test

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Improve compiler error message for missing Swift declarations due to APPLICATION_EXTENSION_API_ONLY (#4603)
 
+### Features
+
+- Show session replay options as replay tags ()
+
 ### Fixes
 
 - `SentrySdkInfo.packages` should be an array (#4626)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Features
 
-- Show session replay options as replay tags ()
+- Show session replay options as replay tags (#4639)
 
 ### Fixes
 
--- a/Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebOptionsEvent.swift
+++ b/Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebOptionsEvent.swift
@@ -10,6 +10,7 @@ import Foundation
                         "errorSampleRate": options.onErrorSampleRate,
                         "maskAllText": options.maskAllText,
                         "maskAllImages": options.maskAllImages,
+                        "quality": String(describing: options.quality),
                         "maskedViewClasses": options.maskedViewClasses.map(String.init(describing: )).joined(separator: ", "),
                         "unmaskedViewClasses": options.unmaskedViewClasses.map(String.init(describing: )).joined(separator: ", ")
                     ]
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Improve compiler error message for missing Swift declarations due to APPLICATION_EXTENSION_API_ONLY (#4603)
 
+### Features
+
+- Show session replay options as replay tags ()
```
brustolin, line 11:
```suggestion
- Show session replay options as replay tags (#4639)
```

Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebOptionsEvent.swift
```diff
@@ -0,0 +1,18 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+@objc class SentryRRWebOptionsEvent: SentryRRWebCustomEvent {
```
armcknight, line 4: Calling these `...Event` is a big confusing. It looks like it's being appended to an array of breadcrumbs, can we do some renaming for this?

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -258,7 +258,11 @@ class SentrySessionReplay: NSObject {
             events.append(contentsOf: touchTracker.replayEvents(from: videoSegmentStart ?? video.start, until: video.end))
             touchTracker.flushFinishedEvents()
         }
-
+        
+        if segment == 0 {
+            events.append(SentryRRWebOptionsEvent(timestamp: video.start, options: self.replayOptions))
```
armcknight, line 263: `events` is an array of breadcrumbs AFAICT, can we rename that as well as the types like `SentryRRWebOptionsEvent` -> `SentryRRWebOptionsBreadcrumb` and `SentryRRWebCustomEvent` -> `SentryRRWebCustomEventBreadcrumb` (or s/breadcrumb/info, etc)

armcknight (approved):
LGTM for functionality and tests! I would just prefer different names.

Also you just need some `#if` sectors to conditionally compile this out of tvOS builds, see the unit test build failures.

Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebOptionsEvent.swift
brustolin: We're following RRWeb naming. This is called an event for RRWeb. for android we're doing the same.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
brustolin: Events is not an array o breadcrumbs. `Events` is and array or RRWebEvents, among them we have breadcrumbs, video info, and custom events.

philipphofmann (approved):
LGTM

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
armcknight: OK, `events` is really just an array of SentrySerializable-conforming objects and is a mixture of both breadcrumbs and RRWebEvents, it seems. I only looked at the return type for `convertBreadcrumbs` and didn't check out `replayEvents(from:until:)`. Polymorphism between ObjC and Swift is fun :D I still think

armcknight: Not that it really matters for this PR, but here's how I would've named that function so that it's clearer at the callsite what's happening: 
image

PR #4643, opened 2024-12-17
ref: One class forward declaration per line

philipphofmann:
Change all class forward declarations to one per line.

#skip-changelog


armcknight (approved):
Excellent 🥇 I took a quick look to see if we could set some format/lint rule for this but didn't come up with anything.

philipphofmann:
> Excellent 🥇 I took a quick look to see if we could set some format/lint rule for this but didn't come up with anything.

Same for me, that's why I opened this PR 😄

Issue #4642, opened 2024-12-17
Allow to use custom UIViewController name when tracking breadcumbs


m1entus:
### Problem Statement

I would love to be able to report custom controller name when sending breadcumbs. Unfortunately we are using modules and each of our feature module UIViewControleller is `ViewController` as a class `Feature.ViewController` and seems it is always showing `ViewController` as a name in the panel. Ideally instead of `NSString *name = [SwiftDescriptor getObjectClassName:controller];` we could use some like `if ([controller conformsToProtocol:@protocol(ViewControllerBreadcrumbTracking)]` then use custom name instead of `getObjectClassName`.

```swift
protocol ViewControllerBreadcrumbTracking {
    var screenName: String { get }
}
```

```objective-c
#if SENTRY_HAS_UIKIT
- (void)swizzleViewDidAppear
{

    // SentrySwizzleInstanceMethod declaration shadows a local variable. The swizzling is working
    // fine and we accept this warning.
#    pragma clang diagnostic push
#    pragma clang diagnostic ignored "-Wshadow"

    static const void *swizzleViewDidAppearKey = &swizzleViewDidAppearKey;
    SEL selector = NSSelectorFromString(@"viewDidAppear:");
    SentryBreadcrumbTracker *__weak weakSelf = self;

    SentrySwizzleMode mode = SentrySwizzleModeOncePerClassAndSuperclasses;

#    if defined(TEST) || defined(TESTCI)
    // some tests need to swizzle multiple times, once for each test case. but since they're in the
    // same process, if they set something other than "always", subsequent swizzles fail. override
    // it here for tests
    mode = SentrySwizzleModeAlways;
#    endif // defined(TEST) || defined(TESTCI)

    SentrySwizzleInstanceMethod(UIViewController.class, selector, SentrySWReturnType(void),
        SentrySWArguments(BOOL animated), SentrySWReplacement({
            SentryBreadcrumb *crumb = [[SentryBreadcrumb alloc] initWithLevel:kSentryLevelInfo
                                                                     category:@"ui.lifecycle"];
            crumb.type = @"navigation";
            crumb.data = [SentryBreadcrumbTracker fetchInfoAboutViewController:self];

            [weakSelf.delegate addBreadcrumb:crumb];

            SentrySWCallOriginal(animated);
        }),
        mode, swizzleViewDidAppearKey);
#    pragma clang diagnostic pop
}
```

```objective-c
[...]

PR #4646, opened 2024-12-18
feat: add Sentry screenName tracking

m1entus:
## :scroll: Description

Added UIViewController custom screenName tracking ref: https://github.com/getsentry/sentry-cocoa/issues/4642 

I used

```swift
protocol SentryViewControllerBreadcrumbTracking {
    var screenName: String { get }
}
```

If viewController responds to `SentryViewControllerBreadcrumbTracking ` then use `screenName` instead of `[SwiftDescriptor getObjectClassName:controller]`

## :pencil: Checklist


Sources/Swift/Protocol/SentryViewControllerBreadcrumbTracking.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+@objc
+public protocol SentryViewControllerBreadcrumbTracking: NSObjectProtocol {
+
+    var screenName: String { get }
+}
```
brustolin, line 7: Thanks @m1entus for this PR.

Since this is a new public protocol, im just missing a documentation in here.

```suggestion
///
/// Use this protocol to customize the name used in the automatic
/// UIViewController performance tracker.
///
@objc
public protocol SentryViewControllerBreadcrumbTracking: NSObjectProtocol {

    /// The custom name of the UIViewController
    /// that will be used for the transaction name.
    var screenName: String { get }
}
```

Something like this.

brustolin (approved):
Thanks for the contribution @m1entus.

LGTM

Sources/Swift/Protocol/SentryViewControllerBreadcrumbTracking.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+///
+/// Use this protocol to customize the name used in the automatic
+/// UIViewController performance tracker.
+///
+@objc
+public protocol SentryViewControllerBreadcrumbTracking: NSObjectProtocol {
```
philipphofmann, line 8: `h`: We use `SwiftDescriptor getObjectClassName` in multiple locations in our code base for getting the UIViewControllerName; see https://github.com/search?q=repo%3Agetsentry%2Fsentry-cocoa%20SwiftDescriptor%20getObjectClassName&type=code

I think we should use `SentryViewControllerBreadcrumbTracking` in all of these scenarios for consistency. Therefore, we should rename this to `SentryUIViewControllerName` and update the code comment above accordingly.

Sources/Swift/Protocol/SentryViewControllerBreadcrumbTracking.swift
```diff
@@ -0,0 +1,13 @@
...
+
+    /// The custom name of the UIViewController
+    /// that will be used for the transaction name.
+    var screenName: String { get }
```
philipphofmann, line 12: `m`: I think we should prefix this with Sentry to avoid name collisions.
```suggestion
    var sentryName: String { get }
```

Sources/Swift/Protocol/SentryViewControllerBreadcrumbTracking.swift
m1entus: Ok i will change, i was thinking about some generic protocol in case in future you would like to add some additional parameters which might be send with breadcrumb. If we name it UIViewControllerName and in future we would like to add additional parameters which might, propably we would have to rename it?

philipphofmann:
@m1entus, I'm sorry I haven't gotten to this yet. I still have some catchup to do after the Christmas holidays. I will definitely review this by the end of the week.

m1entus:
@philipphofmann Cool thanks, no worries - take your time

Sources/Swift/Protocol/SentryViewControllerBreadcrumbTracking.swift
philipphofmann: > If we name it UIViewControllerName and in future we would like to add additional parameters which might, propably we would have to rename it?

Maybe yes, or we add a different protocol. I would stick to something like `SentryUIViewControllerDescriptor` and use it everywhere. Otherwise, you now have a different name for the UIViewController in the breadcrumbs than for UIViewController transactions. If using it everywhere is too much for you to do in this PR, we can also take it from here and finish the PR.

m1entus: @philipphofmann i updated PR, created separate method in `SwiftDescriptor ` for viewController description

Sources/Swift/Protocol/SentryViewControllerBreadcrumbTracking.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+///
+/// Use this protocol to customize the name used in the automatic
+/// UIViewController performance tracker.
```
philipphofmann, line 5:
```suggestion
/// Use this protocol to customize the name used in the automatic
/// UIViewController performance tracker, view hierarchy, and breadcrumbs.
```

Sources/Swift/Protocol/SentryViewControllerBreadcrumbTracking.swift
```diff
@@ -0,0 +1,13 @@
...
+///
+@objc
+public protocol SentryUIViewControllerDescriptor: NSObjectProtocol {
+
+    /// The custom name of the UIViewController
+    /// that will be used for the transaction name.
+    var sentryName: String { get }
```
philipphofmann, line 12:
```suggestion
    /// The custom name of the UIViewController
    /// that the Sentry SDK uses for transaction names, breadcrumbs, and
    /// view hierarchy.
    var sentryName: String { get }
```

Sources/Swift/SwiftDescriptor.swift
```diff
@@ -7,7 +8,15 @@ class SwiftDescriptor: NSObject {
...
         return String(describing: type(of: object))
     }
-    
+
+    @objc
+    static func getViewControllerClassName(_ object: UIViewController) -> String {
+        if let object = object as? SentryUIViewControllerDescriptor {
+            return object.sentryName
+        }
+        return String(describing: type(of: object))
```
philipphofmann, line 17: `m`: Less duplication:
```suggestion
        return getObjectClassName(object)
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Features
 
 - Show session replay options as replay tags (#4639)
+- Add UIViewController custom screenName for breadcrumb tracking (#4642)
```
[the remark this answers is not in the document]
philipphofmann: Please update the PR number

philipphofmann (commented):
Thanks for the quick update. Almost, LGTM.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Features
 
 - Show session replay options as replay tags (#4639)
+- Add UIViewController custom screenName for breadcrumb tracking (#4646)
```
philipphofmann, line 12:
```suggestion
- Add protocol for custom screenName for UIViewControllers (#4646) 
```

philipphofmann (approved):
One minor Changelog entry, which won't block the PR. I need to wait for CI to be green and then we can merge this. Thanks @m1entus.

philipphofmann:
@m1entus, CI is complaining. You can ignore Benchmarking and Build / Release, but the other errors need fixing.

m1entus:
> @m1entus, CI is complaining. You can ignore Benchmarking and Build / Release, but the other errors need fixing.

@philipphofmann Could you tell me what should be fixes? I see that some tests are failing because error cloning repo - so not fully sure what should be fixed ?

PR #4647, opened 2024-12-19
feat: add method unswizzling

philprime:
## :scroll: Description

I adapted the method swizzling helper to keep a reference to the original implementation, so that the `unswizzle` methods can revert swizzling.

We are introducing a new experimental flag in #4634 to conditionally swizzle Objective-C methods. Therefore some unit tests (e.g `testFoo`) swizzle methods, while others (e.g. `testBar`) do not. If we do not offer a functionality to unswizzle methods, they are leaking isolation, i.e. `testFoo` runs before `testBar` therefore methods are swizzled, even tough we don't want that.

This is required for #4634 to pass tests.

I added a unit test to test the unswizzling.

## :pencil: Checklist

## :crystal_ball: Next steps

- [x] Change unswizzling to be wrapped by `#if TEST` to not be used in production


philipphofmann (commented):
FYI, we already tried that in the past and it always blew up the tests. We instead chose a different approach: We don't unswizzle, but we make sure the operations are NoOps when closing the SDK. I mean we can try again, but we need some good reasons. What drove you to open this PR?

philprime:
@philipphofmann in #4634 I added unit tests initializing the SentrySDK with an experimental feature flag enabled or disabled. If the SentrySDK is enabled, and swizzling happens, the swizzling remains even after closing and re-enabling the SentrySDK without swizzling.

We could also change this PR to only unswizzle in unit tests, i.e. using `#if TEST`

Tests/SentryTests/Helper/SentrySwizzleTests.m
```diff
@@ -353,4 +357,23 @@ - (void)testSwizzleDontCallOriginalImplementation
...
+    SentrySwizzleTestClass_A *object = [SentrySwizzleTestClass_B new];
+    swizzleVoidMethod(
+        [SentrySwizzleTestClass_A class], methodForUnswizzling, ^{ SentryTestsLog(@"A"); },
+        SentrySwizzleModeAlways, (void *)methodForUnswizzling);
+    [object methodForUnswizzling];
+    ASSERT_LOG_IS(@"A");
+
+    [SentryTestsLog clear];
+
+    [SentrySwizzle unswizzleInstanceMethod:methodForUnswizzling
```
armcknight, line 373: What happens if something swizzles after us, and then we call this method? Do we break the call chain? I'd be interested to see a test case for that. 

- some object A has a method `foo`
- swizzle `foo` and replace it with `bar`, which calls through to `foo`
- swizzle `foo` again and replace it with `baz`, which calls through to `bar`
- run unswizzle. does that replace `baz` with `foo`? then whatever was relying on `baz` is broken

brustolin:
> We could also change this PR to only unswizzle in unit tests, i.e. using #if TEST

I agree with this!!

Tests/SentryTests/Helper/SentrySwizzleTests.m
philprime: Good catch!

I will have to further look into that and add handling for cascading swizzling. But as the main point of this PR is isolation of unit tests, I would focus on fixing that.

Sources/Sentry/SentrySwizzle.m
```diff
@@ -106,10 +146,40 @@ @implementation SentrySwizzle
...
+static void
+unswizzle(Class classToUnswizzle, SEL selector, const void *key)
+{
+    Method method = class_getInstanceMethod(classToUnswizzle, selector);
+
+    NSCAssert(NULL != method, @"Selector %@ not found in %@ methods of class %@.",
+        NSStringFromSelector(selector),
+        class_isMetaClass(classToUnswizzle) ? @"class" : @"instance", classToUnswizzle);
+
+    static pthread_mutex_t gLock = PTHREAD_MUTEX_INITIALIZER;
```
philipphofmann, line 168: `m`: If I'm not mistaken, this is a different mutex from the one in Swizzle. Wouldn't it be better to share one mutex as they manipulate the same dict.

Sources/Sentry/SentrySwizzle.m
```diff
@@ -40,8 +40,48 @@ - (SentrySwizzleOriginalIMP)getOriginalImplementation
...
+}
+
+static IMP
+getRefToOriginalImplementation(const void *key)
+{
+    NSMutableDictionary<NSValue *, NSValue *> *refsToOriginalImplementations
+        = refsToOriginalImplementationsDictionary();
+    NSValue *keyValue = [NSValue valueWithPointer:key];
+    NSValue *originalImplementationValue = [refsToOriginalImplementations objectForKey:keyValue];
+    return (IMP)[originalImplementationValue pointerValue];
```
philipphofmann, line 78: `m`: What happens if the dict doesn't contain a value for the key?

Sources/Sentry/SentrySwizzle.m
```diff
@@ -174,6 +244,31 @@ + (BOOL)swizzleInstanceMethod:(SEL)selector
     return YES;
 }
 
+#if TEST || TESTCI
++ (BOOL)unswizzleInstanceMethod:(SEL)selector inClass:(Class)classToUnswizzle key:(const void *)key
+{
+    NSAssert(key != NULL, @"Key may not be NULL.");
+
+    if (key == NULL) {
+        NSLog(@"Key may not be NULL.");
```
philipphofmann, line 253: `l`: I would prefer using SentryLog in here instead of NSLog.

Sources/Sentry/include/HybridPublic/SentrySwizzle.h
```diff
@@ -83,6 +83,21 @@
         _SentrySWWrapArg(SentrySWArguments), _SentrySWWrapArg(SentrySWReplacement),                \
         SentrySwizzleMode, key)
 
+#if TEST || TESTCI
+/**
+ * Unswizzles the instance method of the class.
```
philipphofmann, line 88: `m`: Please also add a comment that we don't want to use this in production and only for tests explaining that this isn't safe to use in production.

philipphofmann (commented):
I added a few comments. I'm fine to give this a shot for the tests, but definitely not in production cause swizzling can have drastic unwanted side effects.

Tests/SentryTests/Helper/SentrySwizzleTests.m
philipphofmann, line 373: `m`: I would like to have some test cases for calling unswizzleInstanceMethod for non swizzled methods and calling unswizzleInstanceMethod twice.

Sources/Sentry/SentrySwizzle.m
philprime: I agree, all the swizzle calls share the same mutex, moved it to the class level.

Sources/Sentry/SentrySwizzle.m
philprime: I replaced all NSLog with SentryLog, even the existing ones in the swizzling logic

CHANGELOG.md
```diff
@@ -2,44 +2,10 @@
 
 ## Unreleased
 
-### Internal
-
-- Update to Xcode 16.2 in workflows (#4673)
-
-## 8.43.0
```
philipphofmann, line 9: `m`: This looks like something went wrong when merging main into this branch.

Sources/Sentry/SentrySwizzle.m
```diff
@@ -40,17 +41,104 @@ - (SentrySwizzleOriginalIMP)getOriginalImplementation
...
+}
+
+/**
+ * Adds a reference to the original implementation to the dictionary.
+ *
+ * If the key is NULL, it will log an error and NOT store the reference.
+ *
+ * - Parameter key: The key for which to store the reference to the original implementation.
+ * - Parameter implementation: Reference to the original implementation to store.
+ */
```
philipphofmann, line 68: `l`: Thanks for the code docs, but I don't think you have to add these for internal methods. We can keep them, but you can safe some extra work next time if you want.

Sources/Sentry/include/HybridPublic/SentrySwizzle.h
```diff
@@ -83,6 +83,23 @@
...
+ * To reduce the risk of breaking functionality with unswizzling, this method is not considered
```
philipphofmann, line 89: `l`:
```suggestion
 * @warning To reduce the risk of breaking functionality with unswizzling, this method is not considered
```

philipphofmann (approved):
LGTM, thanks.

Sources/Sentry/SentrySwizzle.m
philprime: I added the comments initially to explain the pre-condition that key must not be NULL, but then decided to add null-handling instead.

philprime:
FYI @philipphofmann I had to remove an `NSCAssert` from the `swizzle` method, because it is actually valid to call it with `key` set to `NULL`. Does not change the implementation, only affects logging.

PR #4651, opened 2024-12-19
fix: Session Redact wrong clipping order

brustolin:
## :scroll: Description

Fix an error where SR clipping happens in the wrong order.

Unit test

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - `SentrySdkInfo.packages` should be an array (#4626)
 - Use the same SdkInfo for envelope header and event (#4629)
 - Fixes Session replay screenshot provider crash (#4649)
+- Session Redact wrong clipping order ()
 
 ### Internal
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - `SentrySdkInfo.packages` should be an array (#4626)
 - Use the same SdkInfo for envelope header and event (#4629)
 - Fixes Session replay screenshot provider crash (#4649)
+- Session Redact wrong clipping order ()
```
brustolin, line 19:
```suggestion
- Session Redact wrong clipping order (#4651)
```

armcknight:
Is there an issue tracking this? Could we see a before/after shot from actual results in the dashboard?

brustolin:
> Is there an issue tracking this? 
There is no issue for this, we were investigating a crash and we found the crash was caused by wrongly skipping some clips

Tests/SentryTests/SentryViewPhotographerTests.swift
```diff
@@ -268,6 +268,32 @@ class SentryViewPhotographerTests: XCTestCase {
         ])
     }
     
+    func testLabelRedactedStackedHierarchy() throws {
```
brustolin, line 271: This test is pretty much to avoid regression. It fails for 8.42.0

philprime (approved):
LGTM

PR #4652, opened 2024-12-20
test(feedback): validate envelopes

armcknight:
building off of #4535, tests for the new envelope we create for the feedback event; for #4272 

#skip-changelog


armcknight:
the remaining failures with the static pod lint steps are fixed in https://github.com/getsentry/sentry-cocoa/pull/4659

PR #4645, opened 2024-12-18
chore(ci): replace xcpretty with xcbeautify

philprime:
## :scroll: Description

Replaces the Ruby Gem `xcpretty` with the CLI `xcbeautify`.

## :bulb: Motivation and Context

- `xcpretty` just broke our CI
- `fastlane` even recommends `xcbeautify` → https://docs.fastlane.tools/best-practices/xcodebuild-formatters/
- Looking at `xcpretty` [commit history](https://github.com/xcpretty/xcpretty/commits/master/), there has not been much activity in the last 2 years, with a lot of open issues

## :green_heart: How did you test it?

- Ran commands locally

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
- [X] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [X] I updated the docs if needed.
- [X] I updated the wizard if needed.
- [X] Review from the native team if needed.
- [X] No breaking change or entry added to the changelog.
- [X] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

#skip-changelog

PR #4653, opened 2024-12-20
chore(ci): make xcbeautify quieter

armcknight:
xcbeautify by default prints every passing test, which results in huge log output making it hard to find failing tests. use options to only print failing tests. 

also try the github action renderer mode. 

#skip-changelog

### DIFF
--- a/scripts/xcode-test.sh
+++ b/scripts/xcode-test.sh
@@ -87,7 +87,7 @@ if [ $RUN_BUILD == true ]; then
         -quiet \
         build 2>&1 |
         tee raw-build-output.log |
-        xcbeautify
+        xcbeautify --quiet --is-ci --renderer github-actions
 fi
 
 if [ $RUN_BUILD_FOR_TESTING == true ]; then
@@ -99,7 +99,7 @@ if [ $RUN_BUILD_FOR_TESTING == true ]; then
         -quiet \
         build-for-testing 2>&1 |
         tee raw-build-for-testing-output.log |
-        xcbeautify
+        xcbeautify --quiet --is-ci --renderer github-actions
 fi
 
 if [ $RUN_TEST_WITHOUT_BUILDING == true ]; then
@@ -110,6 +110,6 @@ if [ $RUN_TEST_WITHOUT_BUILDING == true ]; then
         -destination "$DESTINATION" \
         test-without-building 2>&1 |
         tee raw-test-output.log |
-        xcbeautify &&
+        xcbeautify --quiet --is-ci --renderer github-actions &&
         slather coverage --configuration "$CONFIGURATION"
 fi


philipphofmann:
Yes please.

philprime:
As I introduced xcbeautify, I'll also look into how we can reduce the output to only display failing tests

philipphofmann:
@philprime, I mostly care about seeing failing tests quickly in CI. I don't want to download an artifact to find out which tests were failing. What we had before merging https://github.com/getsentry/sentry-cocoa/pull/4645 was sufficient for me.

philprime:
TBH I am not happy that xcbeautify kind of made things worse, while trying to be better than `xcpretty`. I did consider to propose we revert to xcpretty, but 0.4.0 is [still broken](https://github.com/xcpretty/xcpretty/issues/391), even though [fastlane is also using that version](https://github.com/fastlane/fastlane/blob/master/Gemfile.lock#L44).

I created a prototype for a simple script without external tools to print a test summary in PR #4671. It should be the first thing shown when opening up the GitHub Actions log, like this one:
https://github.com/getsentry/sentry-cocoa/actions/runs/12595175843/job/35103995300

philipphofmann:
@armcknight, I hope it's not a problem that I used this draft PR to play around with some different possibilities. Feel free to revert my changes if you want. 

@philprime, I think all we need to do is to remove `--is-ci` and use `quieter`. With quieter the test log output is OK for me. It's easy to spot the failing test with the failed assert message, see https://github.com/getsentry/sentry-cocoa/actions/runs/12596343110/job/35107314976. While the custom script (https://github.com/getsentry/sentry-cocoa/pull/4671) looks great, it's plenty of code. We might have to maintain and change it. I would prefer sticking to xcbeautify to reduce complexity.

I think we can keep the `--renderer github-actions`, as it looks like a great feature. The noise stems from errors in the tests. We should turn off the logging for these specific tests or fix the errors if possible IMO.

philprime:
I agree with you on the size of the custom script, more code means more maintenance. I created it to replicate the log output of the RSpec style/xcpretty log, but we could also remove parts we don't need, like the code snippet lookup.

For me personally having a clean summary at the end could have the pro of showing actual failing tests and not error output like this:

```
SentryCrashReportConverterTests
Error: Hello fatal\n";
            }
        );
Error: Hello fatal\n";
            }
        );
Error: Hello fatal\n";
            }
        );
Error: Hello fatal\n";
            }
        );
```

Therefore I would propose we add at least a list of failing tests at the end. That could also be extracted from the end of the build log file, as it is there, but xcbeautify does not print it with `--quieter`.

philipphofmann:
I don't mind the error output too much as it's not too much noise at the moment. 

> Therefore I would propose we add at least a list of failing tests at the end. 

Yes, using quieter with a simple grep to only print lines starting with `error: ` would be sufficient for me right now.

armcknight:
No problems on my end to mess with this, I shifted to other stuff after trying a few things and nothing seemed to work. Thanks for finding a solution @philipphofmann

philipphofmann (approved):
Just adding the quieter drastically improves the situation. We can add the grep later.

philprime (approved):
LGTM

PR #4656, opened 2024-12-20
Chore: Persist SR Screenshot provider and breadcrumbs converter

brustolin:
This is a fix for hybrid SDK, where the screenshot provider and the breadcrumb converter was reverting to the default after starting a new app session, usually when the app stays too long in the background.

_#skip-changelog_

### DIFF
--- a/Sources/Sentry/include/HybridPublic/SentrySessionReplayIntegration-Hybrid.h
+++ b/Sources/Sentry/include/HybridPublic/SentrySessionReplayIntegration-Hybrid.h
@@ -22,11 +22,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface SentrySessionReplayIntegration ()
 
-- (void)startWithOptions:(SentryReplayOptions *)replayOptions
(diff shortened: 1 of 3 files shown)


Sources/Sentry/include/HybridPublic/SentrySessionReplayIntegration-Hybrid.h
```diff
@@ -22,11 +22,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface SentrySessionReplayIntegration ()
 
-- (void)startWithOptions:(SentryReplayOptions *)replayOptions
-      screenshotProvider:(id<SentryViewScreenshotProvider>)screenshotProvider
-     breadcrumbConverter:(id<SentryReplayBreadcrumbConverter>)breadcrumbConverter
-             fullSession:(BOOL)shouldReplayFullSession;
-
```
brustolin, line 29: I removed this because I didn't find anything using it.

vaind (approved):

PR #4658, opened 2024-12-21
feat(feedback): fix static pod lint

armcknight:
trying to fix the last build issue in #4652; #skip-changelog; for #4272


armcknight:
looks like the correct solution is https://github.com/getsentry/sentry-cocoa/pull/4659

PR #4659, opened 2024-12-21
fix(feedback): bubble up feedback capture to objc layer

armcknight:
fixes the last build issue in #4652; #skip-changelog; for https://github.com/getsentry/sentry-cocoa/issues/4272

### DIFF
--- a/Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
+++ b/Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
@@ -32,7 +32,7 @@ class SentryFeedback: NSObject, SentrySerializable {
         super.init()
     }
     
-    public func serialize() -> [String: Any] {
+    func serialize() -> [String: Any] {
         let numberOfOptionalItems = (name == nil ? 0 : 1) + (email == nil ? 0 : 1) + (associatedEventId == nil ? 0 : 1)
         var dict = [String: Any](minimumCapacity: 2 + numberOfOptionalItems)
         dict["message"] = message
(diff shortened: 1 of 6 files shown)


Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
```diff
@@ -32,7 +32,7 @@ class SentryFeedback: NSObject, SentrySerializable {
         super.init()
     }
     
-    func serialize() -> [String: Any] {
+    public func serialize() -> [String: Any] {
```
brustolin, line 35: Curious why do you set this function public if the class is not public?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
```diff
@@ -7,6 +7,10 @@ import UIKit
 
 var displayingForm = false
 
+protocol SentryUserFeedbackWidgetDelegate: NSObjectProtocol {
```
brustolin, line 10: l: If this is a `NSObjectProtocol` should it also have `@objc`?

brustolin (approved):
A couple of comments but LGTM.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
armcknight: oh, leftover from when i thought the class needed to be public, didn't remove this one again 👍🏻

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
armcknight: correct, it doesn't need it, it's implied

PR #4675, opened 2025-01-03
feat(feedback): hooks

armcknight:
Implement the various hooks that can be set in the form config.

for #4271; #skip-changelog


armcknight:
Need to investigate why the UI tests are now failing with the new assertions while they pass locally.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
```diff
@@ -50,6 +50,22 @@ class SentryFeedback: NSObject, SentrySerializable {
         return dict
     }
     
+    func dataDictionary() -> [String: Any] {
+        var dict: [String: Any] = [
+            "message": message
```
philipphofmann, line 55: `m`: No UI test is failing when I modify this to the following
```suggestion
            "messag": message
```

I don't think we should add a UI test for this, but instead some unit tests to validate the functionality.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -130,12 +130,19 @@ class SentryUserFeedbackForm: UIViewController {
             let list = missing.count == 1 ? missing[0] : missing[0 ..< missing.count - 1].joined(separator: ", ") + " and " + missing[missing.count - 1]
             let alert = UIAlertController(title: "Error", message: "You must provide all required information. Please check the following field\(missing.count > 1 ? "s" : ""): \(list).", preferredStyle: .alert)
             alert.addAction(UIAlertAction(title: "OK", style: .default))
-            present(alert, animated: config.animations)
+            present(alert, animated: config.animations) {
+                if let block = self.config.onSubmitError {
+                    block(NSError(domain: "io.sentry.error", code: 1, userInfo: ["missing_fields": missing, NSLocalizedDescriptionKey: "The user did not complete the feedback form."]))
```
philipphofmann, line 135: `m`: Wouldn't be [Swift errors](https://developer.apple.com/documentation/swift/error) be more suitable than an NSError? Basically using an enum or a struct.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
```diff
@@ -50,6 +50,22 @@ class SentryFeedback: NSObject, SentrySerializable {
         return dict
     }
     
+    func dataDictionary() -> [String: Any] {
```
philipphofmann, line 53: `m`: Why is this different from serialize? Can't this be the same?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
philipphofmann, line 135: `m`: When I change this to the following, no test fails. It would be great to have a test asserting the contents of NSError.
```suggestion
                    block(NSError(domain: "io.sentry", code: 1, userInfo: ["missing_fields": missing, NSLocalizedDescriptionKey: "The user did not complete the feedback form."]))
```

Also a test should fail when the code doesn't invoice the block, which isn't the case right now. The same applies to `onSubmitSuccess`.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
armcknight: It's a different schema, from the docs. There's extra info in the serialization stuff that isn't even available at this point.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: When I comment out the invocation of `onSubmitSuccess`, `testSubmitWithOnlyRequiredFieldsFilled`, `testSubmitFullyFilledForm` and `testSubmissionErrorThenSuccessAfterFixingIssues` fail.

For `onSubmitError`, the following fail: 
testSubmissionErrorThenSuccessAfterFixingIssues()
testSubmitOnlyWithOptionalFieldsFilled()
testSubmitWithNoFieldsFilledAllRequired()
testSubmitWithNoFieldsFilledAllRequiredCustomLabels()
testSubmitWithNoFieldsFilledDefault()
testSubmitWithNoFieldsFilledEmailAndMessageRequired()

What did you try that you didn't see the failures?

armcknight:
Agree that the contents of the parameters in onSubmitError should be validated. I think it'd take a lot of work to refactor things to be able to unit test it, I will first look at doing it with the UI test. I'd even thought about it but forgot to write that one, so thanks for reminding me!

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
philipphofmann: Maybe I got it wrong. I'm sorry, I can't recall. If you're confident tests are failing, I trust you. Thanks for double checking.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
philipphofmann: Please add the info as a comment.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
```diff
@@ -49,6 +51,26 @@ class SentryFeedback: NSObject, SentrySerializable {
         
         return dict
     }
+}
+ 
+// MARK: Public
+extension SentryFeedback {
+    /// - note: This dictionary is to pass to the block `SentryUserFeedbackConfiguration.onSubmitSuccess`, describing the contents submitted. This is different from the serialized form of the feedback for envelope transmission, because there are some internal details in that serialization that are irrelevant to the consumer and are not available at the time `onSubmitSuccess` is called.
```
philipphofmann, line 58: Thanks 👍

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
philipphofmann: It would be great to add this info as a comment, cause I see myself asking the question again in 2 months when I look at this code again 😅.

philipphofmann (approved):
LGTM

armcknight:
I'm just working on an issue with the tests where the dictionary contents in the success/error block parameters aren't consistently ordered.

PR #4677, opened 2025-01-04
ref(iOS-Swift): extract sdk config out of app delegate

armcknight:
This was violating the file length lint rule, so I just moved it out. Would make it easier to share between other sample apps later, as a bonus.

#skip-changelog


philprime (approved):

armcknight:
For some reason iOS13-Swift can't be built because it can't see the declaration of `configureUserFeedback` from SentryOptions+Private.h, whereas iOS-Swift can. I tried looking at how the two app targets might differ but can't find anything yet. Is there something simple I could check out?

armcknight:
Ugh, didn't realize that i set this to merge to my branch and not main 🤦🏻

armcknight:
@philipphofmann try again at #4684

PR #4682, opened 2025-01-08
fix(profiling): implicit autorelease pool retain cycle

### LINKED ISSUE #4679: Memory use balooning
### Platform

macOS

### Environment

Production

### Installed

Manually

### Version

8.43.0

### Xcode Version

16.0

_No response_

### Steps to Reproduce

Integrated Sentry
Profilling with "Instruments" in "Leaks" mode
While app is idle, persistant memory allocations keep going up in a rate of a megabyte per 10 seconds
Seeing strings and arrays allocated in [SentryProfilerState appendBacktrace]
Disabled Sentry and ran again, problem gone
Also tried to disable various options for performance testing (which I'm not interested in anyway, I just need it for crash reporting)Image

### Expected Result

Memory consumption should not go up

### Actual Result

Memory consumption goes up

_No response_

armcknight:
fixes #4679 

when i reproduced the issue using continuous profiling, i saw that even though on each successive serialization of profile data there was a relatively constant amount of data being processed, like ~6k SentrySample instances, and then the in-memory caches would be reset each time, something was still holding references to all the previous SentrySamples, which I confirmed by looking at the memory debugger which at one point showed 90k instances still in memory. 

i modified the logic so that only one sample would be retained each chuck transmission, so i'd be able to navigate the list of all of them in memory. after the first chunk serialized and we reset the data, i was able to compare the memory graph of the now two instances in memory:
image
 
after trying some modifications to eliminate block-based retain cycles, i tried the current solution, which resolved the issue in the memory graph: now only the currently serializing sample instances were still in memory, and the previous ones were deallocated. validated with a new instruments run that memory growth remains flat across many profiling chunk transmissions and in-memory data structure resets: 
image

i tested with the fix and without to make sure the profiling data looks the same. it appears this change doesn't have any adverse effects on profile data collection. tested with the cpu workload feature in the sample app: 
|`main`|with change|
|---|---|
|image|image|
|[profile](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/flamegraph/?colorCoding=by%20library&end=2025-01-08T06%3A34%3A25.551Z&eventId=cc4dd4e213c645e78e68d3e5bed7945d&profilerId=9e716a5a42bf43ca91e75f95c9ec77f7&query=&sorting=call%20order&start=2025-01-08T06%3A34%3A20.318Z&tid=130819&traceId=a2fe142df5cc425a9156eb99607aa2df&view=top%20down)|[profile](https://sentry-sdks.sentry.io/profiling/profile/sentry-cocoa/flamegraph/?colorCoding=by%20library&end=2025-01-08T01%3A49%3A01.784Z&eventId=2f760854e989476fa2e2150453402653&profilerId=433f0dcffae545eb8e7dcbd97c970c40&query=&sorting=call%20order&start=2025-01-08T01%3A48%3A56.768Z&tid=130307&traceId=35eb4bea92934984bf637fc5729aa65e&view=top%20down)|


Sources/Sentry/Profiling/SentryProfilerState.mm
```diff
@@ -77,7 +77,9 @@ - (void)mutate:(void (^)(SentryProfilerMutableState *))block
 {
     NSParameterAssert(block);
     std::lock_guard<std::mutex> l(_lock);
-    block(_mutableState);
+    @autoreleasepool {
+        block(_mutableState);
+    }
```
philipphofmann, line 82: `m`: We only use the `mutate` method in `appendBacktrace` and in one test. Instead, we could remove the method and put the logic in `appendBacktrace`. Then, we don't have to mess around with blocks not being deallocated. Because now, we still allocate plenty of blocks and we have to use an autoreleasepool to fix the problem. I would prefer getting rid of the `autoreleasepool` by removing the need for blocks.

philipphofmann (commented):
Wow, weird memory growth issue 🤔

Sources/Sentry/Profiling/SentryProfilerState.mm
indragiek: I don't believe the block itself has anything to do with the issue. Whether the objects are allocated within the block, or outside of the block (within `appendBacktrace` directly) we would have the same problem. The actual cause is further up the chain: https://github.com/getsentry/sentry-cocoa/blob/24052e9d7159b1c2c6ca915c201a8dfe812770e9/Sources/Sentry/SentryProfiler.mm#L173

That lambda is called by the sampling thread within `samplingThreadMain`: https://github.com/getsentry/sentry-cocoa/blob/24052e9d7159b1c2c6ca915c201a8dfe812770e9/Sources/Sentry/SentrySamplingProfiler.cpp#L43

The sampling thread has no autorelease pool because it is a `pthread` that is created directly by the C++ code (not `NSThread`, which implicitly creates an autorelease pool), and that's the root cause here. 

Wrapping the implementation of `samplingThreadMain` in `@autoreleasepool` would be the correct fix.

indragiek (changes requested):

armcknight:
After moving the autoreleasepool, confirmed memory growth remains flat: 
image

armcknight:
Confirmed that the growth issue remains resolved after moving the location of the autoreleasepool: 
image

indragiek (approved):

Sources/Sentry/Profiling/SentryProfilerState.mm
philipphofmann: @indragiek, thanks for jumping in here. It seemed a bit strange to me that blocks would cause such a huge difference, and I should have investigated this further. 

@indragiek and @armcknight: It seems to me that we don't even need to create a backtrace copy here: https://github.com/getsentry/sentry-cocoa/blob/d7cec05f5812b86ecdc40e424baceb0d7f040c97/Sources/Sentry/SentryProfiler.mm#L170-L172

because we create a new backtrace struct in the for loop here:

https://github.com/getsentry/sentry-cocoa/blob/d7cec05f5812b86ecdc40e424baceb0d7f040c97/Sources/Sentry/SentryBacktrace.cpp#L112-L118

So, it could be that we simply need to remove the copy of the backtrace and then we can remove the `@autoreleasepool`. Then we would avoid allocating numerous structs in a tight for loop.

Ideally, it would be great to move the allocation of the struct out of the for loop in `SentryBacktrace` and avoid creating copies of it. The for loop could overwrite the values of the same struct and pass it to the callback. UI frameworks also frequently reuse objects to avoid allocations in a tight loop and reduce memory footprint. That's where I got the idea from. Maybe I'm missing something here, so please let me know if I do.

armcknight: We do need to make the copy of the backtrace, but it's a struct so isn't part of the equation with the autoreleasepool causing retained objects to hang around.

We have to make that copy because this is the earliest time we can attach the modified timestamp, and we can't directly assign that property on the backtrace struct instance that's passed into the lambda block.

philipphofmann: Thanks for clarifying. I don't have enough knowledge on that part of the code and trust you to make a proper judgment. It seems legit.

PR #4684, opened 2025-01-08
ref(iOS-Swift): extract sdk config out of app delegate

armcknight:
Accidentally merged https://github.com/getsentry/sentry-cocoa/pull/4677 into my work branch instead of main. Let's try it again.

#skip-changelog

PR #4323, opened 2024-09-09
test: add tags with git branch/commit information to sample apps

armcknight:
For https://github.com/getsentry/sentry-cocoa/issues/4321

Add commit and branch info. Will attach `-dirtied` to commit hash if there are changes to the working index. If working from a detached head state, the branch is reported as `HEAD`.

Dirty index on a branch: https://sentry-sdks.sentry.io/issues/5485500008/events/latest/?project=5428557&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=latest-event&statsPeriod=1h&stream_index=0

image

Detached HEAD with clean index: https://sentry-sdks.sentry.io/issues/5485405120/events/latest/?project=5428557&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=latest-event&statsPeriod=1m&stream_index=0

image

#skip-changelog

PR #2025, opened 2022-07-29
meta: migrate to XcodeGen and xcconfigs

armcknight:
See https://github.com/getsentry/sentry-cocoa/discussions/2024 for discussion on this task.

# Points of interest

- all xcodeproj, xcscheme and xcworkspace files have been removed from git tracking and are now .gitignore'd
- all targets from all previous projects will now be in a single project together (except SPM/Carthage based projects: Carthage-Validation, macOS-SPM-CommandLine, test-server)
- the previously multiplatform targets Sentry and SentryTests now are one target per platform
- the project spec itself: `//project.yml`
- code signing is now set to automatic by default, and is only swapped for match-managed creds in CI (see `Fastfile`)
- new makefile task to generate the project from the XcodeGen spec
- new xcconfigs in `//xcconfig/`

# TODO

- [x] Some files are committed to this branch to help keep track of intermediate results and checks when rebasing to keep up with `master`; they must be removed before merging this PR:
    - `//xcodebuild...` directories
    - //scripts/output-all-schemes-showbuildsettings.sh
- [x] update `//develop-docs/README.md`
- [x] update `//CONTRIBUTING.md`

#skip-changelog

PR #4691, opened 2025-01-09
ref: create new shared xcodeproj for sample apps

armcknight:
Between the sample apps we currently have, I see a couple organizational problems:

- duplicated code/assets
- most test apps lack the ability to test most SDK features

By moving as much as possible into a common library, we can easily share more sample/test implementation between more apps. The apps themselves should really just be the minimum interface possible to get the sample implementation onto a particular platform/version. That's more xcconfig/pbxproj territory; most objc/swift is not germane to the matter (aside from maybe UIApplicationDelegate/UISceneDelegate or SwiftUI main, or Swift 5 vs 6 sources, but even those I bet could be abstracted into the lib in different targets). For instance, after this we can share SentrySDKWrapper.swift, I extracted in #4684, to all sample apps (with the exception of iOS-ObjectiveC; for which we'd probably want to create an objc sentry wrapper, then that could be shared to other objc sample apps, along with essentially an entire parallel objc implementation of all our swift sample app code, if necessary).

This will be an ongoing refactor with many small PRs over time, which will be cleaner than trying to do it all at once. This PR serves to kick off the effort by creating the lib project, moving GitInjections.swift (from #4323) into it, and linking it to the sample apps that all used to directly compile that file from a shared location.

Longer term, I think we should reconsider my previous proposal to use XcodeGen to manage the project structures instead of xcodeproj/xcworkspace files. This would eliminate duplication of settings (which can be somewhat mitigated with xcconfigs) and build phases. (See #2025)

#skip-changelog


philprime:
I agree that sharing the same code base in multiple sample apps can reduce the maintenance workload and also increase the functionality available/tested in the different cases.

On the other hand we should consider if there are side-effects when creating an Objective-C app, using a shared Swift framework, to then use the Objective-C implementation of Sentry.

I believe there is a valid point in having an Objective-C sample app which is not using Swift by itself at all.

On the idea of introducing xcodegen again: I've been heavily using Xcodegen for my recent projects, therefore I see its value, especially when reducing merge conflicts. We could maybe even add a pre-commit hook or an GitHub Action workflow to make sure the generated projects is correctly generated.

But Xcodegen has its limitations too, especially when setting values which Xcode reports as "reason to migrate the Xcode project". Luckily we can always get rid of Xcodegen again, and just use the previously generated Xcode project manually.

philipphofmann:
I don't have a strong opinion here. For me multiple Xcode project has never been an issue. As it's only for the sample apps the potential damage is that we are annoyed, which we can revert easily. Once CI is green, I would approve the PR.

armcknight:
> On the other hand we should consider if there are side-effects when creating an Objective-C app, using a shared Swift framework, to then use the Objective-C implementation of Sentry.

@philprime Definitely need to consider that. We could create a shared objc target in SampleShared, so there's be no objc(app)->swift(shared sample lib)->objc(sdk) switching going on. See the end of my longest (admittedly very wordy) paragraph, I touched on this.

On the other hand, there are probably customers that actually do this (objc or objc/swift app using swift library wrapping sentry), so it might actually be beneficial to have the ability to test this setup.

armcknight:
So, I can't roll back the objectVersion of the xcode project far enough back to get it compatible with the older project versions of our sample app projects. and i can't move those projects up, because we still need to test them on older versions of xcode. See my various attempts in #4692.

This kind of gridlock that prevents us from moving forward with sensible architectural work is exactly the kind of thing that I want to avoid by using XcodeGen and removing the actual pbxproj files from source control.

The only thing I can think to do at this point is to delete all the sample projects and the new lib's project, and recreate them all with Xcode 14 on my machine, to get them all to use a common, compatible objectVersion. I don't really have the appetite for that right now, it would be better to just start using XcodeGen. So for now i'm just going to close this PR.

PR #4693, opened 2025-01-09
fix: add logging for invalid cache directory path validation

philprime:
## :scroll: Description

Adds additional validation and error logging to the SDK option `cacheDirectoryPath`.

While testing edge cases of #4636 I noticed that paths are truncated to `SentryCrashFU_MAX_PATH_LENGTH`, which can cause file operations to use incomplete paths if they are longer than the given limit:

https://github.com/getsentry/sentry-cocoa/blob/52c33ede189798e1332ad172e813067ad61105e6/Sources/SentryCrash/Recording/Tools/SentryCrashFileUtils.h#L41

https://github.com/getsentry/sentry-cocoa/blob/52c33ede189798e1332ad172e813067ad61105e6/Sources/SentryCrash/Recording/SentryCrashReport.c#L1476-L1484

Currently the SDK will just silently fail and stop working. SDK users should be informed that their cache directory path is invalid, so they can proactively fix it.

This issue might not apply to many users, as they most likely will set shorter paths or use OS-specific defaults.

- Configure Sentry in the `SentrySDKWrapper.swift.configureSentryOptions` using these test examples:

  **Valid path:**
  ```
  options.cacheDirectoryPath = "/tmp/sentry"
  ```

  **A path that is too long for the system:**
  ```swift
  options.cacheDirectoryPath = "/tmp/sentry" + (0..<10)
    .map { i in "\(i)abcdefghijklmnopqrstuvwxyz" }
      .joined(separator: "-")
  ```

  **A path where a path component is too long:**
  ```swift
  options.cacheDirectoryPath = "/tmp/sentry/" + Array(repeating: "A", count: 256).joined() + "/B/"
  ```
- Set a breakpoint in `_non_thread_safe_removeFileAtPath(NSString *path)` in line 52 with the `SENTRY_LOG_ERROR` to view the error object or just check the log output for `"File name too long"`

https://github.com/getsentry/sentry-cocoa/blob/52c33ede189798e1332ad172e813067ad61105e6/Sources/Sentry/SentryFileManager.m#L43-L58

- With the changes of this PR, the following statement will be in the log:

```
The configured cache directory path looks invalid, the SDK might not be able to write reports to disk: /tmp/sentry/AAA...AAA/B/
```

## :pencil: Checklist

## :crystal_ball: Next steps

There are a couple of caveats to consider:

- [x] Decide if logging via `SentryLog` is sufficient, because this is a critical configuration error and `SentryLog` only logs errors if `options.debug` is `true`
- [x] Decide if the options validation must be at every point of usage of the `options.cacheDirectoryPath` or if using it once during SDK startup is enough.


Sources/Sentry/SentryOptions.m
```diff
@@ -360,7 +360,7 @@ - (BOOL)validateOptions:(NSDictionary<NSString *, id> *)options
         self.maxCacheItems = [options[@"maxCacheItems"] unsignedIntValue];
     }
 
-    if ([options[@"cacheDirectoryPath"] isKindOfClass:[NSString class]]) {
+    if ([SentryOptionsValidator isCacheDirectoryPathValidWithPath:options[@"cacheDirectoryPath"]]) {
```
philipphofmann, line 363: `h`: What drove you to duplicate this logic all over the SDK? Wouldn't one check in a custom setter of `cacheDirectoryPath` be sufficient?

Sources/Swift/SentryOptionsValidator.swift
```diff
@@ -0,0 +1,37 @@
...
+        if !Self.isCacheDirectoryPathValid(path: options.cacheDirectoryPath) {
+            SentryLog.fatal("The configured cache directory path looks invalid, the SDK might not be able to write reports to disk: \(options.cacheDirectoryPath)")
+        }
+    }
+
+    @objc
+    static func isCacheDirectoryPathValid(path: Any) -> Bool {
+        guard let path = path as? String else {
+            return false
+        }
```
philipphofmann, line 15: `m`: It seems odd to me that the `path` is of type `Any`. I would change this to a String and do the casting where it's not a String.

Sources/Swift/SentryOptionsValidator.swift
```diff
@@ -0,0 +1,37 @@
...
+        let fileUrl = URL(fileURLWithPath: path)
+
+        // The cache directory path is used as a base path in the SDK,
+        // therefore actual paths can be even longer and it is necessary
+        // to include some reserved space for appendix.
+        //
+        // The following length is assumed based on paths used in the SDK.
+        let reservedInternalLength: Int32 = 256
```
philipphofmann, line 23: `h`: If I'm not mistaken, the limit exists because of `SentryCrashFU_MAX_PATH_LENGTH.` If yes, the logic here should be tied to root cause of the issue here. 

https://github.com/getsentry/sentry-cocoa/blob/52c33ede189798e1332ad172e813067ad61105e6/Sources/SentryCrash/Recording/SentryCrashReport.c#L1476-L1484

Sources/Swift/SentryOptionsValidator.swift
```diff
@@ -0,0 +1,37 @@
+@objcMembers
+class SentryOptionsValidator: NSObject {
+    @objc
+    static func validate(options: Options) {
+        SentryLog.debug("Validating Sentry SDK options")
+        if !Self.isCacheDirectoryPathValid(path: options.cacheDirectoryPath) {
+            SentryLog.fatal("The configured cache directory path looks invalid, the SDK might not be able to write reports to disk: \(options.cacheDirectoryPath)")
```
philipphofmann, line 7: `m`: The log message should indicate how you can fix the error. Now, it only states it's not working, but instead,d it should communicate that the path is too long and how to fix it.

Sources/Swift/SentryOptionsValidator.swift
philprime: The type is `Any` because this method is also used from `SentryOptions.validateOptions:` which is used by Hybrid SDKs to pass type-erased options.

philipphofmann: What about only doing the type check there?

Sources/Sentry/SentryOptions.m
philprime: This was the only validation logic I could find regarding options validation, and it only validates casting of the type-erased value from the Hybrid SDKs.

As this was my starting point to implement a validator class, I moved the logic away from here.

I'm open for discussion if this should not be changed.

Regarding the duplicating all over the SDK, there is a TODO in this PR to decide where the appropriate place to add validation is, therefore I'll add a comment directly to the PR instead.

Sources/Swift/SentryOptionsValidator.swift
philprime: There are multiple sources of issues here:
- The operating system does not support paths longer than `PATH_MAX`
- The operating system does not support path names (file/directory names) longer than `NAME_MAX`
- The buffer with length `SentryCrashFU_MAX_PATH_LENGTH` is shorter than the actual path

The `reservedInternalLength` is not related to any other field, but as the comment states, it is assumed to be long enough to fit sub paths we are appending to the `cacheDirectoryPath`, i.e. paths to crash reports.

Maybe extending the `SentryCrashFU_MAX_PATH_LENGTH` to `PATH_MAX` should also be considered

Sources/Swift/SentryOptionsValidator.swift
philprime: I agree, I'll change that to provide actual feedback how to solve the issue

philprime:
Regarding why I added the validation to multiple locations in the SDK:

The main indicator was the `options.cacheDirectoryPath` to select the positions. I agree that it is probably not necessary to add validation to all of the positions, but we need to look at all of them.

For example the `PrivateSentrySDKOnly.installationID` is used from the Hybrid SDKs, but not in the SDK itself. 

Originally I considered only adding it to `SentrySDK.startWithOptions`, but then I noticed that also dependencies such as `crashReporter` in the `SentryDependencyInjection` is accessing the static options, which (if i am not mistaken) could be accessed before `startWithOptions` is even triggered.

Please leave comments during review which duplicates of the implementation you consider to be obsolete.

Sources/Swift/SentryOptionsValidator.swift
philipphofmann: OK, then, let's take a step back. Something close to 99% of the users don't use the `cacheDirectoryPath`. This is an edge case. If you pass in a custom `cacheDirectoryPath` you better ensure it's working correctly, but we should notify you with a log message when it doesn't work. Instead of checking it upfront, I think it's acceptable to print error messages whenever we try to write a file to disk and we can't. Because we only know that we can't write a file once we try it. The current validation logic is based on some logic we came up with. If the file writing logic changes, we might miss changing the file path validation logic. For writing a crash report, this is different, because we can't wait to print an error message when we write the crash report because the user won't see that. To fix this, we could use the same logic for creating the crash report in `sentrycrashreport_writeRecrashReport` to do a check when SentryCrash initializes, if we can write a crash report when the app crashes. If we can't, I would log a fatal error message. This is a long write-up. If you have questions, I think it's better to discuss this on a call.

philprime:
Blocked until #4707 is merged

Sources/Sentry/SentryFileManager.m
```diff
@@ -20,19 +20,44 @@
...
+    NSError *underlyingError = NULL;
+    if (@available(macOS 11.3, iOS 14.5, watchOS 7.4, tvOS 14.5, *)) {
+        underlyingError = error.underlyingErrors.firstObject;
+    }
+    if (underlyingError == NULL) {
+        id errorInUserInfo = [error.userInfo valueForKey:NSUnderlyingErrorKey];
+        if ([errorInUserInfo isKindOfClass:[NSError class]]) {
+            underlyingError = errorInUserInfo;
+        } else {
+            SENTRY_LOG_FATAL(@"Failed to get underlying error from error: %@", error);
```
philipphofmann, line 35: `m`: So for OS versions below `(macOS 11.3, iOS 14.5, watchOS 7.4, tvOS 14.5, *)` we always log this fatal error message here, which can be confusing. Instead, we could simply return NO here and just skip logging a fatal error. It's very unlikely that this happens. To simplify the code we could do an early return above if we can't get the underlyingErrror.

philipphofmann (approved):
LGTM, with one comment.

Sources/Sentry/SentryFileManager.m
philprime: Ah yes, missed that one. I will change the code to return early if no underlying error is available

Sources/Sentry/SentryFileManager.m
```diff
@@ -20,19 +20,45 @@
 
 NSString *const EnvelopesPathComponent = @"envelopes";
 
+BOOL
+isErrorPathTooLong(NSError *error)
+{
+    NSError *underlyingError = NULL;
```
philprime, line 26: @philipphofmann this should now correctly unwrap the underlying error, otherwise it will fall back to the error itself

brustolin (approved):
LGTM

philipphofmann:
> Therefore this PR does not directly affect that, but it could happen that writing a crash fails due to e.g. out of space. I can create another issue for that case, as we might have to discuss how to test the behavior.

I'm not worried out of space, but any other issue. TLDR; the SDK should validate during start if it could write a crash report. If it can't, it should log a fatal error.

Sources/Sentry/SentryFileManager.m
philipphofmann: Hmm there is so much logic in this logging code that I would feel more comfortable adding tests for this. We can test the log output, but here I'm mostly worried about test coverage. We don't need to validate the log output, but having 100% test coverage ensuring we're not crashing would be good. We had issues in the past in logging code that led to crashes because we didn't add tests for it. Then, when an edge case occurred, the log message crashed the app.

philprime: I agree, I'll add tests

philprime:
Not sure if validation during startup is actual worth it. My example with the out-of-space, is something that can occur later on during app execution, rather than at the beginning, therefore validating the write capabilities at the start will not help it.

Not sure what you want me to implement now

philipphofmann:
> Not sure if validation during startup is actual worth it. My example with the out-of-space, is something that can occur later on during app execution, rather than at the beginning, therefore validating the write capabilities at the start will not help it.
> 
> Not sure what you want me to implement now

Nothing, just add the tests and let's merge it.

Sources/Sentry/include/SentryFileManager.h
```diff
@@ -15,6 +15,12 @@ NS_ASSUME_NONNULL_BEGIN
 @class SentryOptions;
 @class SentrySession;
 
+#if SENTRY_TEST || SENTRY_TEST_CI
+BOOL isErrorPathTooLong(NSError *error);
+
+BOOL createDirectoryIfNotExists(NSString *path, NSError **error);
+#endif // SENTRY_TEST || SENTRY_TEST_CI
+
```
philipphofmann, line 23: `l`: You could also move this to `SentryFileManager+Test.h` to keep making things visible for testing approach consistent for this class.

Sources/Sentry/SentryFileManager.m
```diff
@@ -20,19 +20,45 @@
...
+                                             withIntermediateDirectories:YES
+                                                              attributes:nil
+                                                                   error:error];
+    if (success) {
+        return YES;
     }
-    return YES;
+
+    if (isErrorPathTooLong(*error)) {
+        SENTRY_LOG_FATAL(@"Failed to create directory, path is too long: %@", path);
```
philipphofmann, line 57: `m`: Thanks for all the tests, it would be great to add one test that actually triggers logging this error event, cause here is the biggest risk for me to cause a crash or something.

philipphofmann (approved):
LGTM, I think we should still add one more test. Thanks

SentryTestUtils/SentryFileManager+Test.h
```diff
@@ -2,6 +2,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+BOOL isErrorPathTooLong(NSError *error);
+BOOL createDirectoryIfNotExists(NSString *path, NSError **error);
```
philipphofmann, line 6: Thanks 🚀

philipphofmann (approved):
Thanks for adding the extra tests, LTRM :shipit:

philprime:
Had to catch another error on older OS versions.
I believe the test cases are fine and this is ready to be merged.

PR #4694, opened 2025-01-09
chore(ci): add retry logic to benchmark runs

### LINKED ISSUE #4689: Benchmarks run on Sauce Labs fail with internal error 500
### GitHub action Run Link

Error "Error while extracting response for type" / 500:
- https://github.com/getsentry/sentry-cocoa/actions/runs/12673254561/job/35319654909
- https://github.com/getsentry/sentry-cocoa/actions/runs/12680382104/job/35342258694
- https://github.com/getsentry/sentry-cocoa/actions/runs/12680787913/job/35343387518

Error "failed to execute run command error="unexpected EOF"":
- https://github.com/getsentry/sentry-cocoa/actions/runs/12669582201/job/35307809819
- https://github.com/getsentry/sentry-cocoa/actions/runs/12669582201/job/35307810243
- https://github.com/getsentry/sentry-cocoa/actions/runs/12691690229/job/35375663666

Error "Suite timed out.":
- https://github.com/getsentry/sentry-cocoa/actions/runs/12467375278/job/34796712717
- https://github.com/getsentry/sentry-cocoa/actions/runs/12467375278/job/34796712937
- https://github.com/getsentry/sentry-cocoa/actions/runs/12467375278/job/34796713155

### Disabling PR

none

### Description

Internal errors of SauceLab are causing our CI checks to fail.
Manual re-triggering of the GitHub action workflows fixes the issue.

We should catch internal errors and retry multiple times, before failing our checks.

Example:
https://app.saucelabs.com/tests/3c4f1d75134a46aca99263390370d768

Image
Image

philprime:
## :scroll: Description

Adds retry mechanism for flaky uploading and testing of benchmarks running on Saucelabs.

Closes #4689

Internal errors of SauceLab are causing our CI checks to fail.
Often the retriggering helps to pass the test.

This approach uses the generated log to identify the test case id, to check if the test case even exists. If the lookup of the test fails, we can assume that SauceLabs failed to run the test and that we should trigger it again.

I implemented the check as a bash script in a workflow job. As it turned out to be quite long, I added a single retry instead of multiple ones.
In case we want multiple retries, I can refactor the script to a file and add additional workflow steps.

An example of a simulated failure can be found here:

https://github.com/getsentry/sentry-cocoa/actions/runs/12710298356/job/35431329854

CI checks of this PR

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

Closes #4689


.github/workflows/benchmarking.yml
```diff
@@ -94,10 +94,31 @@ jobs:
...
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
-        run: saucectl run --select-suite "${{matrix.suite}}" --config .sauce/benchmarking-config.yml --tags benchmark
+        run: saucectl run --select-suite "${{matrix.suite}}" --config .sauce/benchmarking-config.yml --tags benchmark --retries 3
+      - name: Run Benchmarks in SauceLab - Retry 1
+        id: run-benchmarks-in-sauce-lab-retry-1
+        if: ${{ steps.run-benchmarks-in-sauce-lab.outcome == 'failure' }}
+        env:
+          SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
+        run: saucectl run --select-suite "${{matrix.suite}}" --config .sauce/benchmarking-config.yml --tags benchmark --retries 3
```
philipphofmann, line 108: `m`: I would only retry if we get the specific 500 error. Otherwise, we also retry when UI tests are flaky, which will lead to flaky tests staying unnoticed for some time and will increase CI time.

philprime: I agree, but I don't see a reliable way of detecting the error cause other than trying to extract information from the error log - which also has no clear indicator

philprime: In case we want to go for that approach, I addded additional workflow jobs that failed to #4689 and grouped them by their log message

philipphofmann: What about always retrying 3 times unless there is a failing test? We should be able get the info if a test failed in Saucelabs. If it did we don't retry.

philprime: That could work, I'll look into this, but we should also retry when tests time out

philipphofmann: Yes, please.

philipphofmann: But you can do timeouts later. Step by step.

philprime: https://github.com/getsentry/sentry-cocoa/actions/runs/12710298356/job/35431329854?pr=4694

I added a simulated failure.

Using `continue-on-error` following jobs are executed even when the workflow step fails. I had to extract the test id from the build log to get the test result from `saucectl`.

Not happy with this approach, even though it works, extraction of an ID from the log is unstable.

philprime:
I just noticed we are using [actions/github-scripts](https://github.com/actions/github-script).

If it helps maintainability I can refactor the PR to use it instead of plain bash script

philipphofmann:
> I just noticed we are using [actions/github-scripts](https://github.com/actions/github-script).
> 
> If it helps maintainability I can refactor the PR to use it instead of plain bash script

Yes, please.

philprime:
Refactored to use action/github-script

Here are simulated examples of the recovery:

1. Test ID was not found in log - should retry
https://github.com/getsentry/sentry-cocoa/actions/runs/12768860847/job/35590530258

2. Test ID found in log with simulation as "testing failed" - should not retry
https://github.com/getsentry/sentry-cocoa/actions/runs/12767996209/job/35587804661

.github/workflows/benchmarking.yml
```diff
@@ -94,10 +94,69 @@ jobs:
...
+            const { execSync } = require('child_process');
+            
+            console.log("Extracting test ID from output log");
+            const outputLog = fs.readFileSync('output.log', 'utf8');
+            const match = outputLog.match(/https:\/\/app\.saucelabs\.com\/tests\/([^\s]+)/);
+            const testId = match?.[1] ?? '';
+            
+            if (!testId) {
+              console.log("No test ID found in the output, retrying...");
+              core.setOutput('RETRY_TEST', 'true');
```
philipphofmann, line 129: `m`: What happens if the log output of saucelabs changes and we can't find a match for every test run anymore? Then we would retry for every run, if I'm not mistaken. If this is true, do you have any ideas on how we can mitigate this problem? Otherwise, it could be that we retry for a long time until we realize that this changed.

philipphofmann (approved):
Thanks, lets give this a shot. If it's working correctly, we could consider adding this to the other mobile SDKs as well.

philprime:
Example where the tests where run but timed out themselves:

https://github.com/getsentry/sentry-cocoa/actions/runs/12769779159/job/35593612580?pr=4694

The CI workflow is failing as expected for this case, as we only retry on infrastructure issues.

.github/workflows/benchmarking.yml
philprime: This is a fair feedback point I do not have a solution for right now.
We could add a GitHub job annotation warning that the retry mechanism was triggered, so at least we have a prominent alert.

philipphofmann: Then I think adding this downside as a comment and modifying the log message to hint that the log output could have changed if it always retries could mitigate this a bit.

philprime: I added a warning log statement:

https://github.com/getsentry/sentry-cocoa/actions/runs/12785908532/job/35642172190#step:6:49

It also adds an annotation to the log:

https://github.com/getsentry/sentry-cocoa/actions/runs/12785908532

Does this match your request?

philprime: As discussed we will leave it like this for now

Issue #4401, opened 2024-10-03
Accessibility audit


armcknight:
### Description

Implement and check the standard accessibility stuff for a U, both for the widget and formI:
- [ ] voiceover navigation and readouts
- [ ] huge text size

armcknight:
@bruno-garcia One question is what, if anything, voiceover should read out for the logo? Currently I have it as "provided by Sentry", which, combined with the default form title, would read out as "Report a bug, provided by Sentry".

armcknight:
Pushing this until after end-to-end functionality via #4272

armcknight:
Audit findings:
Widget: 
- tapping anywhere in the feedback button doesn't get voiceover readout–only tapping on the actual label
	- fix: need to make the entire widget button view a single accessibility element reading out the button's title text
 
Form: 
- sentry logo doesn't get voiceover readout
	- fix: because we use a UIView with a CAShapeLayer, it's not an accessibility element by default since it's not a control element. need to set `isAccessibilityElement` to true. from the docs:
		> The default value for this property is false unless the element is a standard UIKit control, in which case, the value is true.
- screenshot/image preview doesn't get voiceover readout
	- fix: same as above, as a UIImageView is also not a UIKit control
- (ETA: this is not an issue, see note below) having trouble dismissing keyboard while editing the UITextView for the feedback message/description, which covers the submit button so totally breaks the UX. 
	- Current planned workaround is to change the name/email fields' `returnKeyType` to `UIReturnKeyType.done` and then implement a handler for that to dismiss the keyboard.
	- **Note:** VoiceOver has a three-finger swipe up gesture for "go to next page" and performing this scrolls the UIScrollView down to reveal the remainder of the form (from https://support.apple.com/guide/iphone/use-voiceover-gestures-iph3e2e2281/ios)
- the UITextView and the "placeholder" label we float over top of it both get read out by VoiceOver. 
	- fix: Set the labe's accessibility element property to `false` so that it is ignored by VoiceOver

armcknight:
I also think we should:
- add `accessibilityHint` to the "add screenshot" button to explain that it will place them in the standard
- and to the submit button to read back a description of everything that is about to be sent

armcknight:
There are still a couple issues with huge text sizes:
- the message text view's floating placeholder label is aligned too far down, and this persists even when switching back to smaller text size again: 

[...]

PR #4698, opened 2025-01-10
fix(feedback): accessibility audit action items

armcknight:
for #4401; #skip-changelog


Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
```diff
@@ -103,7 +104,7 @@ public class SentryUserFeedbackFormConfiguration: NSObject {
     lazy var nameLabelContents = fullLabelText(labelText: nameLabel, required: isNameRequired)
     
     /**
-     * The placeholder for the name input field.
+     * The placeholder for the name input field.sc
```
philprime, line 107: Seems like a typo

```suggestion
     * The placeholder for the name input field.
```

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -97,11 +95,35 @@ class SentryUserFeedbackForm: UIViewController {
...
+            let imagePickerController = UIImagePickerController()
+            imagePickerController.delegate = self
+            imagePickerController.sourceType = .photoLibrary
+            imagePickerController.allowsEditing = true
+            DispatchQueue.main.async {
+                self.present(imagePickerController, animated: self.config.animations)
+            }
+        }
+        
+        let status = PHPhotoLibrary.authorizationStatus()
```
philprime, line 109: `m`: `authorizationStatus()` is [deprecated](https://developer.apple.com/documentation/photos/phphotolibrary/authorizationstatus()), we should start using [authorizationStatus(for:)](https://developer.apple.com/documentation/photos/phphotolibrary/authorizationstatus(for:)) starting with `iOS 14.0`.

Also

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -97,11 +95,35 @@ class SentryUserFeedbackForm: UIViewController {
...
+        switch status {
+        case .notDetermined:
```
philprime, line 111: `l`: We should also add more detailed logging for different [status](https://developer.apple.com/documentation/photos/phauthorizationstatus#Status-Types), i.e. add a debug log when the user access is restricted, explaining SDK users that the app users have to change the setting in the system settings to enable it again.

Maybe we can also add the authorization status as context to the feedback report, so that when a developer receives a report via Sentry without an image, they can tell if the user has not selected one or if the SDK simply couldn't ask for one.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -128,15 +150,22 @@ class SentryUserFeedbackForm: UIViewController {
         
         guard missing.isEmpty else {
             let list = missing.count == 1 ? missing[0] : missing[0 ..< missing.count - 1].joined(separator: ", ") + " and " + missing[missing.count - 1]
-            let alert = UIAlertController(title: "Error", message: "You must provide all required information. Please check the following field\(missing.count > 1 ? "s" : ""): \(list).", preferredStyle: .alert)
+            return "You must provide all required information. Please check the following field\(missing.count > 1 ? "s" : ""): \(list)."
```
philprime, line 153: `l`: should we support localizations?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -318,14 +351,19 @@ class SentryUserFeedbackForm: UIViewController {
         return textView
     }()
     
-    lazy var screenshotImageView = UIImageView()
+    lazy var screenshotImageView = {
```
philprime, line 354: `l`: Explicitly defining the type of screenshotImageView could improve compile time.

```suggestion
    lazy var screenshotImageView: UIImageView = {
```

Maybe possible to measure using these `OTHER_SWIFT_FLAGS`: 
```
-Xfrontend -warn-long-function-bodies=50 -Xfrontend -warn-long-expression-type-checking=50
```

Maybe helpful: https://github.com/fastred/Optimizing-Swift-Build-Times

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -318,14 +351,19 @@ class SentryUserFeedbackForm: UIViewController {
...
+        return iv
+    }()
     
     lazy var addScreenshotButton = {
         let button = UIButton(frame: .zero)
         button.setTitle(config.formConfig.addScreenshotButtonLabel, for: .normal)
         button.accessibilityLabel = config.formConfig.addScreenshotButtonAccessibilityLabel
         button.addTarget(self, action: #selector(addScreenshotButtonTapped), for: .touchUpInside)
         button.accessibilityIdentifier = "io.sentry.feedback.form.add-screenshot"
+        button.accessibilityHint = "Will present the iOS photo picker for you to choose an image to attach to the feedback report."
```
philprime, line 366: `l`: Should we add localization?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -432,32 +510,51 @@ class SentryUserFeedbackForm: UIViewController {
...
         addedScreenshot(image: photo)
-        dismiss(animated: config.animations)
+         
+        guard let asset = info[.phAsset] as? PHAsset else {
+            // TODO: handle error
+            return
+        }
+        
+        guard let date = asset.creationDate else {
+            // TODO: handle error
```
philprime, line 554: `l`: How did this `TODO:` pass the pre-commit hooks? Shouldn't it be detected by SwiftLint?

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
```diff
@@ -56,6 +56,7 @@ public class SentryUserFeedbackFormConfiguration: NSObject {
     /**
      * Allows the user to send a screenshot attachment with their feedback.
      * - note: Default: `true`
+     * - warning: Your app must provide a value in its Info plist for the `NSPhotoLibraryUsageDescription` key, or the app will crash upon requesting the necessary `PHPhotoLibrary` authorization.
```
philipphofmann, line 59: `h`: That sounds severe. Could we mitigate this problem with more than a comment? Can't we disable the photo library feature if the user doesn't provide this specific key? We could also log an error message when the user enables this feature and the key isn't in the info.plist file.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
philipphofmann: If I'm not mistaken, the plan is to add this later. @armcknight can confirm.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
```diff
@@ -346,9 +384,49 @@ class SentryUserFeedbackForm: UIViewController {
         button.setTitleColor(config.theme.submitForeground, for: .normal)
         button.addTarget(self, action: #selector(submitFeedbackButtonTapped), for: .touchUpInside)
         button.accessibilityIdentifier = "io.sentry.feedback.form.submit"
+        button.accessibilityHint = submitAccessibilityHint
         return button
     }()
     
+    lazy var submitAccessibilityHint = {
```
philipphofmann, line 391: `m`: I would prefer extracting this logic into an extra method or class and maybe detach it from UIKit dependencies and write plain Swift unit tests for it. I'm also fine with keeping dependencies to the textFields and imagView if we can easily write tests for it. I think this is a simple logic that can easily introduce bugs and testing all the edge cases with UI tests is too complicated IMO. Unit tests run wait faster and speed up development. Still having a few UI tests checking that the main use cases work fine, is even more important.

philipphofmann (commented):
I added a few comments.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: Yes, it would be a future thing but i'm not even sure where that is on the roadmap. I'm not familiar with any kind of localization for any of our products.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: I have that swiftlint rule disabled for some of these implementations

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
armcknight: yes, I think we could check for the existence in the main bundle's user info dictionary before trying to call the APIs.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: I can see some benefit to it, although it will come with its own risk that the interface between the core controller and interface driver can contain its own kinds of bugs. 

I think we should at least wait until we've got everything working on a happy path we want to settle on, which it seems we're just about there, and maybe even not until we actually have some feedback and reported bugs to start writing unit tests for this. In any case a separate PR dedicated solely to that refactor I think would be best.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: i'll take care of this one as it should probably just be another log, but I also added a checklist item to remove the overrides before shipping: #4709

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
philipphofmann: > although it will come with its own risk that the interface between the core controller and interface driver can contain its own kinds of bugs.

That's why I think a combination of UI tests and unit tests is best. UI tests test the most critical happy path. Unit tests validate all the edge cases and maybe even the happy path.

> I think we should at least wait until we've got everything working on a happy path we want to settle on

The problem with that is that we potentially build code on untestable code, and quite often we forget refactoring it later. I would prefer fixing this right away. It's simply extracting the logic into an extra internal method. That should do the trick. But that's up to you.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
philipphofmann: > I'm not familiar with any kind of localization for any of our products.

Me neither, because we mostly build for devs and they usually speak English 😅. We'll figure that out later.

philipphofmann: @philprime, I know, that's why we most likely will need localization.

armcknight: Correct @philprime, we can look to the existing offering for user feedback in the javascript sdk: https://github.com/getsentry/sentry-javascript/blob/develop/packages/feedback/src/constants/index.ts

I don't think there are other places to look for translated user-facing strings in that implementation. I'll ask in the user feedback project's channel what the thinking is for i18n.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: I can add this back in. I originally had it but it was crashing because I hadn't added the infoplist property. Although before I realized that was the cause, I thought there was something going on with this API 😆

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
armcknight: Thanks!

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: I used to recommend this more but haven't noticed much difference in modern times. Just to be sure, I checked the build time analysis with no explicit type annotations (so, the file as it is today) and with _everything_ explicitly annotated:

|before|after|
|---|---|
|image|image|

So about 113 additions of types only saved about 100 ms. I'm gonna pass on this one!

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
armcknight: I'm going to do this in a separate PR because I've found some other things that need to be fixed in the photo picker UX.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: I'll also do this in the separate photo picker PR

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: Also doing the logging in the separate PR for photos improvement: #4723 

I'm not sure how much value there'd be in sending the photos authorization status to sentry, if there's a structured way to do that, or indeed, if it's even wise to do so from a privacy perspective.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackForm.swift
armcknight: I'm tackling this in https://github.com/getsentry/sentry-cocoa/pull/4726

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackFormConfiguration.swift
```diff
@@ -56,6 +56,10 @@ public class SentryUserFeedbackFormConfiguration: NSObject {
     /**
      * Allows the user to send a screenshot attachment with their feedback.
      * - note: Default: `true`
+     * - warning: Your app must provide a value in its Info plist for the
```
philprime, line 59: `l`: do we still need this warning?

Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
```diff
@@ -30,4 +55,107 @@ class SentryFeedbackTests: XCTestCase {
         let attachments = sut.attachments()
         XCTAssertEqual(attachments.count, 0)
     }
+
```
philprime, line 58: `l`: instead of creating a single test which loops over all input combinations, you could create individual tests for each input combination, but each one calls a helper function. The helper function would then pass `...file: StaticString = #file, line: UInt = #line)` as parameters down to the `XCTAssertEqual(..., file: file, line: line)`, therefore propagating the error up to the calling function.

That would then result in individual tests failing if the input combination is not working, rather than failing the whole

scripts/.clang-format-version
```diff
@@ -1 +1 @@
-19.1.6
+19.1.7
```
philprime, line 1: `l`: maybe move this and the swiftlint-version to its own PR

philprime (approved):
LGTM, smaller comments to decide at your discretion

philipphofmann (approved):
Nothing more to add than @philprime. LGTM

Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
armcknight: i did consider dynamically generating a test function per case, if this gets painful enough i will spend more time on it later

scripts/.clang-format-version
armcknight: I will open automerging PRs for this in the future

Issue #4700, opened 2025-01-11
feat(feedback): use success/error color


armcknight:
### Description

we have configuration options at SentryUserFeedbackThemeConfiguration.successColor and errorColor, but they are currently unused. Either use them or remove them.

Issue #4699, opened 2025-01-11
feat(feedback): Set configured tags


armcknight:
### Description

I think this was forgotten in #4535 . Need to add `SentryUserFeedbackConfiguration.tags`

PR #4701, opened 2025-01-11
fix(feedback): use SentryUser stored in scope by default

armcknight:
As I was updating https://github.com/getsentry/sentry-docs/pull/11191, I noticed a couple things that will need to be fixed. Filed a few as issues (#4699, #4700), and fixed the `useSentryUser` default here.

#skip-changelog


philprime (approved):

philprime:
Maybe we should include this change in behaviour in the changelog

philipphofmann (approved):
Thanks, LGTM.

philipphofmann (commented):
Does this maybe deserve a changelog entry? Or was the feature not public yet?

armcknight:
The feature is not yet public @philipphofmann @philprime, thanks for double checking!

armcknight:
I added a change to the test that should have broken with this change. It was then broken with the prior version of the implementation, and the current version with the change in this PR makes it pass again 👍🏻

philipphofmann (approved):
LGTM again, but CI complains a bit. No clue if it's related.

armcknight:
It seems like it was related, after I merged the hooks PR #4675 which also had some testing updates. I wound up needing to fix quite a few more tests here, where they were inputting user info where it wasn't needed any longer, and added a new test to go over the happy path when not using defaults (since user info is now prefilled by default). also needed to update the launch arg.

PR #4707, opened 2025-01-13
feat: add threshold to always log fatal logs

### LINKED ISSUE #4467: Change default log level to warning
### Description

The default SDK log level is debug. When you turn on `option.debug = true` the SDK logs too much information, so warnings and errors get lost. Instead the SDK should log only warnings and above when `option.debug = true`. Users can manually set the `diagnosticLevel` to debug if required. When we detect a debug build, we turn on debug = true, so you get only the warnings and errors. We still need to discuss if we're going to handle this as a breaking change or not, and we should create a meta issue for this to align across all SDKs.

This is also required to surface severe SDK misconfiguration or other errors clearly to the user such as: https://github.com/getsentry/sentry-cocoa/issues/4566

```[tasklist]
### 🛑 Blocked by
- [ ] https://github.com/getsentry/sentry-docs/pull/12100
```

philprime:
## :scroll: Description

Introduces a new constant in `SentryLog` to configure a log level that will always be logged.

Closes #4467

I added a statement `SENTRY_LOG_FATAL(@"test")` in the SDK install code and checked if it gets printed.

## :pencil: Checklist

## :crystal_ball: Next steps

_none_

### DIFF
--- a/Sources/Swift/Tools/SentryLog.swift
+++ b/Sources/Swift/Tools/SentryLog.swift
@@ -8,7 +8,7 @@ class SentryLog: NSObject {
     static private(set) var diagnosticLevel = SentryLevel.error
 
     /**
-     * The log level that will always be logged, regardless of the current configuration
+     * Threshold log level to always log, regardless of the current configuration
      */
     static let alwaysLevel = SentryLevel.fatal
     private static var logOutput = SentryLogOutput()
--- a/Sources/Swift/Tools/SentryLog.swift
+++ b/Sources/Swift/Tools/SentryLog.swift
@@ -6,6 +6,11 @@ class SentryLog: NSObject {
     
     static private(set) var isDebug = true
     static private(set) var diagnosticLevel = SentryLevel.error
+
+    /**
+     * The log level that will always be logged, regardless of the current configuration
+     */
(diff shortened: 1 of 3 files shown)


Sources/Swift/Tools/SentryLog.swift
```diff
@@ -30,7 +35,13 @@ class SentryLog: NSObject {
      */
     @objc
     static func willLog(atLevel level: SentryLevel) -> Bool {
-        return isDebug && level != .none && level.rawValue >= diagnosticLevel.rawValue
+        if level == .none {
+            return false
+        }
+        if level.rawValue >= alwaysLevel.rawValue {
+            return true
+        }
```
brustolin, line 43: l: Can we change this to `guard` to ensure early return in compile time?

philprime: As much as I like the `guard` statement, I am not sure if the compromise in readability is worth it:

```suggestion
        guard level != .none else {
            return false
        }
        guard level.rawValue < alwaysLevel.rawValue else {
            return true
        }
```

brustolin (approved):
LGTM

I will leave the `guard` thing to your best judgement.

PR #4717, opened 2025-01-15
refactor: Convert constants SentryTraceOrigins to Swift

philprime:
## :scroll: Description

Moves the global constants defined in `SentryTraceOrigins.h` to static constants in a Swift class

- We want to migrate the code base to Swift over time
- We want the constants to be available in Swift without complicated setup
- It uses a `@objcMembers` class to make the static properties accessible from Objective-C.
- The current pattern uses the same syntax as Swift's `enum` which is not available in Objective-C. Eventually it will only be used from Swift files, and can be replaced with an enum without further code changes.

Unit tests

## :pencil: Checklist


armcknight (approved):
Also nice 👍🏻 Looks like this one will need a few fixes for tests, tho

Sources/Swift/Transactions/SentryTraceOrigin.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+@objcMembers
+public class SentryTraceOrigin: NSObject {
```
philipphofmann, line 4: `h`: Again I would like to keep this internal, same as here: https://github.com/getsentry/sentry-cocoa/pull/4718#discussion_r1918222526

philipphofmann (commented):
I would like to keep the constants internal, please.

Sources/Swift/Transactions/SentryTraceOrigin.swift
philprime: Not possible because they are used by SentrySwiftUI

philprime:
@brustolin can you please take a look at the failing `SentryTimeToDisplayTrackerTest. testTracerFinishesBeforeReportInitialDisplay_FinishesInitialDisplaySpan`.

I can't really tell why my changes break this test, and the test setup is quite complicated.

brustolin (approved):
LGTM

PR #4718, opened 2025-01-15
refactor: Convert constants SentrySpanOperations to Swift

philprime:
## :scroll: Description

Collects all global constants defined in SentrySpanOperations.h and other various variables into static constants in a Swift class

- We want to migrate the code base to Swift over time
- We want the constants to be available in Swift without complicated setup
- Span operation names are defined in the [Sentry specification](https://develop.sentry.dev/sdk/telemetry/traces/span-operations/) and should be managed in one spot to match the spec
- It uses a @objcMembers class to make the static properties accessible from Objective-C.
- The current pattern uses the same syntax as Swift's enum which is not available in Objective-C. Eventually it will only be used from Swift files, and can be replaced with an enum without further code changes.

CI tests

## :pencil: Checklist


armcknight (approved):
Nice 👍🏻

Sources/Swift/Transactions/SentrySpanOperation.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+@objcMembers
+public class SentrySpanOperation: NSObject {
```
philipphofmann, line 4: `h`: Does this need to be public? I would prefer to keep this internal. If the only reason is
```
/// Needs to be public to be accessible from `SentrySwiftUI`
public static let uiLoad = "ui.load"
```
I would prefer having some duplication over everything using the same constant.

philipphofmann (commented):
I would like to keep the constants internal if possible.

Sources/Swift/Transactions/SentrySpanOperation.swift
philprime: I understand your point in not having it publicly available, but I disagree with preferring duplication.

The constants are defined in the [Span Operations Spec](https://develop.sentry.dev/sdk/telemetry/traces/span-operations/), which IMHO should be represented in a single location in code, allowing to edit and append as new spans are introduced.

philipphofmann: We can't change things until the next major when making them public. Making things public is not worth it to avoid one duplication of a string constant that hardly ever changes. The SentrySwiftUI code already uses `SentryInternal.h` to access nonpublic classes from the SDK. Let's use this or a similar approach instead. Then, we get the best out of both worlds. Would that work for you, @philprime?

philipphofmann: You can create an internal Objective-C wrapper class for the constants in Swift. That could work. I am just so sensitive to making things public because I've been bitten by this multiple times. For example, SentrySerializable doesn't need to be public, but it is. If it would be internal, we could just change.

philprime: The compiler generates a file `Sentry-Swift.h` which contains a interface of the `SentrySpanOperation` to make it accessible from Objective-C.

I did not succeed to import the `Sentry-Swift.h` into `SentryInternal.h` which makes internal types accessible to `SentrySwiftUI`.

I tried to re-declare the `SentrySpanOperation` interface directly in `SentryInternal.h` and that did work under the condition that it matches exactly the header defined in `Sentry-Swift.h`:

```
@interface SentrySpanOperation : NSObject
@property (nonatomic, class, readonly, copy) NSString * _Nonnull appLifecycle;
+ (NSString * _Nonnull)appLifecycle;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull coredataFetchOperation;
+ (NSString * _Nonnull)coredataFetchOperation;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull coredataSaveOperation;
+ (NSString * _Nonnull)coredataSaveOperation;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull fileRead;
+ (NSString * _Nonnull)fileRead;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull fileWrite;
+ (NSString * _Nonnull)fileWrite;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull networkRequestOperation;
+ (NSString * _Nonnull)networkRequestOperation;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull uiAction;
+ (NSString * _Nonnull)uiAction;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull uiActionClick;
+ (NSString * _Nonnull)uiActionClick;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull uiLoad;
+ (NSString * _Nonnull)uiLoad;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull uiLoadInitialDisplay;
+ (NSString * _Nonnull)uiLoadInitialDisplay;
@property (nonatomic, class, readonly, copy) NSString * _Nonnull uiLoadFullDisplay;
+ (NSString * _Nonnull)uiLoadFullDisplay;
 - (nonnull instancetype)init __attribute__((objc_designated_initializer));
@end
```

If I only include the relevant `uiLoad` the compiler fails:

```
'SentrySpanOperation' has different definitions in different modules; first difference is definition in module 'SentryInternal' found property 'uiLoad'
```

Defining a huge interface like this just for a single value seems overkill.
Therefore I have two options to offer:

1. Double-down and leave the value as-is, not having a class which contains all the constants
2. Duplicate the `SentrySpanOperation.swift` into `SentrySwiftUI`, removing all values not required and add a comment to edit the `SentrySpanOperation.swift` in the `Sentry` instead.
3. Create an internal module `SentryCore` which is imported by `Sentry` and `SentrySwiftUI`. The values will be public, but that module should not be used by developers anyways.
4. Include the `SentrySpanOperation.swift` in `Sentry` and `SentrySwiftUI` in the compile process

I personally prefer 2 alternatively 1

philipphofmann: I also prefer option 2.

brustolin: 2 Is fine.

3 sounds very appealing as way to move forward to modularization.

Sources/SentrySwiftUI/SentrySpanOperation.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+// Discarded Approach 1:
+//   - Define `@interface SentrySpanOperation` in `SentryInternal.h`
+//   - Swift class is exposed to Objective-C in auto-generated `Sentry-Swift.h`
+//   - Conflict: Duplicate interface definition
+//
+// Discarded Approach 2:
+//   - Declare Swift class `SentrySpanOperation` in main file as `public`.
+//   - Auto-generated `Sentry-Swift.h` is manually imported to `SentrySwift.h`
+//   - Issue: Internal class is public for SDK users, which is not desired
```
philipphofmann, line 17: Thanks 🚀

philipphofmann (approved):
LGTM

Issue #4722, opened 2025-01-16
feedback: photo picker improvements


armcknight:
### Description

- [ ] Check for ability to request authorization before calling the API, which can crash the app if the infoplist key is missing
- [ ] make sure it is shown in a popover on ipad, as mentioned at https://developer.apple.com/documentation/uikit/uiimagepickercontroller:
<img width="535" alt="Image" src="https://github.com/user-attachments/assets/af7dd05c-4e38-4b83-8902-5ff30aa8c815" />

armcknight:
We will need to defer this until we can deliver a separate sdk submodule so that only people integrating that will need the required photos access plist key. See https://github.com/getsentry/sentry-cocoa/issues/4768

philprime:
Closing — the photo picker was removed from the feedback form entirely. The form now accepts a pre-captured `screenshot: UIImage?` parameter instead of presenting an interactive picker. All three requested improvements (Info.plist check, PHPickerViewController, iPad popover) are no longer applicable.

PR #4723, opened 2025-01-16
impr(feedback): iteration on photo picker

armcknight:
Addressing some feedback in #4698 as a separate piece of work (#4722); #skip-changelog


Samples/iOS-Swift/iOS-Swift/SentrySDKWrapper.swift
```diff
@@ -222,7 +222,14 @@ extension SentrySDKWrapper {
...
             UIApplication.shared.delegate?.window??.rootViewController?.present(alert, animated: true)
-            let jsonData = (try? JSONSerialization.data(withJSONObject: info, options: .sortedKeys)) ?? Data()
+            
+            // if there's a screenshot's Data in this dictionary, JSONSerialization crashes _even though_ there's a `try?`, so we'll write the base64 encoding of it
+            var infoToWriteToFile = info
+            if let attachments = info["attachments"] as? [Any], let screenshot = attachments.first as? Data {
+                infoToWriteToFile["attachments"] = [screenshot.base64EncodedString()]
+            }
+            
+            let jsonData = (try? JSONSerialization.data(withJSONObject: infoToWriteToFile, options: .sortedKeys)) ?? Data()
```
philprime, line 232: `l`: How about replacing JSONSerialization with JSONEncoder and Codable? Might be more complex with arbitrary data, but could improve type safety

philprime (approved):

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackFormViewModel.swift
```diff
@@ -19,6 +19,11 @@ class SentryUserFeedbackFormViewModel: NSObject {
     unowned let controller: SentryUserFeedbackFormController
     weak var delegate: SentryUserFeedbackFormViewModelDelegate?
     
+    /// Checks to make sure the app provides the necessary Info plist key to request authorization. If the key is not present, trying to interact with certain Photos APIs will crash the app.
+    var canRequestAuthorizationToAttachPhotos = {
+        Bundle.main.infoDictionary?["NSPhotoLibraryUsageDescription"] != nil
+    }()
```
philipphofmann, line 25: `m`: Wouldn't it make sense to check that once the user configures `formConfig.enableScreenshot = true`? Then the user would immediately see the error message. I would even change the log level to error. Now, we wait until we try to attach a photo.

philipphofmann (approved):
LGTM, maybe it makes sense to print an error message for the missing `NSPhotoLibraryUsageDescription` earlier, but that shouldn't block the PR.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackFormViewModel.swift
armcknight: This will prevent the button from even appearing in the form for now, but I will also make a check at the point the configuration is set in the call to SentrySDK.startWithOptions to emit an earlier log.

armcknight: This will now also log a warning at the time the integration is installed:

image

armcknight: shoot, forgot to push the commit before merging this. i pushed it to https://github.com/getsentry/sentry-cocoa/pull/4698/commits/d6aecd5952d318160021cf7bc18e95cbd1f99075

PR #4732, opened 2025-01-21
feat: Deserialize User

philipphofmann:

Adds the Decodable implementation for SentryUser, including decoding arbitrary data.


Sources/Swift/Protocol/Codable/SentryUserCodable.swift
```diff
@@ -0,0 +1,34 @@
...
+        case email
+        case username
+        case ipAddress = "ip_address"
+        case segment
+        case name
+        case geo
+        case data
+    }
+    
+    @available(*, deprecated, message: "Segment is deprecated, but we still need decode it.")
```
philprime, line 17: `l`: I am not sure if I understand the warning, why do we implement a decoding method which is already deprecated?

philprime (approved):

Sources/Swift/Protocol/Codable/SentryUserCodable.swift
philipphofmann: I'm sorry, I didn't update the message accordingly. This is a bit complicated. Is it clearer now, @philprime? If you have a better suggestion on how to suppress the deprecation error, please let me know.

philprime: I get it now, for future reference how it looks like without the deprecation availability annotation

image

philipphofmann: Yes exactly.

PR #4735, opened 2025-01-21
test: Increase timeout for ANRV1 tests

philipphofmann:
Increase the timeout for ANRTrackerV1 tests from 1.0 to 2.0, cause they timed out here https://github.com/getsentry/sentry-cocoa/actions/runs/12886410167/job/35926973742.

2.0 is still OK. If they still are flaky we need to disable them.

#skip-changelog


philprime (approved):

philipphofmann:
Thanks @vaind, but at least that's not a timeout 😅. It's a different problem.

PR #4737, opened 2025-01-21
feat: Session replay masking preview for SwiftUI

### LINKED ISSUE #4633: [Feature Request] View masking preview for SwiftUI views
### Problem Statement

I am integrating Session Replay in a SwiftUI based app with some custom masking rules. Each time I need to check how the view is masked, I have to run the app, wait for the session to appear on sentry, scroll the video to find my target screen and finally check the masking. Repeat until I've got it right.

This is very tedious to do and is error-prone.

### Solution Brainstorm

Add a SwiftUI view modifier `sentryMaskPreview` that can be used to render a masked view in previews.

Maybe

brustolin:
## :scroll: Description

Allows the user to preview how its app session replay will be masked during development.

image

closes #4633 

Sample

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentrySDK.m
+++ b/Sources/Sentry/SentrySDK.m
@@ -203,6 +203,9 @@ static NSDate *_Nullable startTimestamp = nil;
 
 + (void)startWithOptions:(SentryOptions *)options
 {
+    // We save the options before checking for Xcode preview because
+    // we will use this options in the preview
+    startOption = options;
     if ([SentryDependencyContainer.sharedInstance.processInfoWrapper
                 .environment[SENTRY_XCODE_PREVIEW_ENVIRONMENT_KEY] isEqualToString:@"1"]) {
         // Using NSLog because SentryLog was not initialized yet.
@@ -210,7 +213,6 @@ static NSDate *_Nullable startTimestamp = nil;
         return;
     }
 
-    startOption = options;
     [SentryLog configure:options.debug diagnosticLevel:options.diagnosticLevel];
 
     // We accept the tradeoff that the SDK might not be fully initialized directly after
--- a/Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
+++ b/Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
@@ -34,7 +34,7 @@ struct SentryReplayPreviewView: UIViewRepresentable {
     }
     
     func updateUIView(_ uiView: SentryReplayMaskPreviewUIView, context: Context) {
-        uiView.opacity = opacity
+        uiView.opacity = CGFloat(opacity)
     }
 }
 
--- a/Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
+++ b/Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
@@ -1,0 +1,41 @@
+#if canImport(SwiftUI) && canImport(UIKit) && os(iOS) || os(tvOS)
+import Sentry
+import SwiftUI
+import UIKit
+
+#if CARTHAGE || SWIFT_PACKAGE
+@_implementationOnly import SentryInternal
+#endif
+
+@available(iOS 13, macOS 10.15, tvOS 13, *)
+struct SentryReplayMaskPreview: ViewModifier {
+    let redactOptions: SentryRedactOptions
+    let opacity: Float
+    func body(content: Content) -> some View {
+        content.overlay(SentryReplayPreviewView(redactOptions: redactOptions, opacity: opacity).disabled(true))
+    }
+}
+
+@available(iOS 13, macOS 10.15, tvOS 13, *)
+public extension View {
+    func sentryReplayPreviewMask(redactOptions: SentryRedactOptions? = nil, opacity: Float = 1) -> some View {
+        let options = redactOptions ?? SentrySDK.options?.sessionReplay ?? PreviewRedactOptions()
+        return modifier(SentryReplayMaskPreview(redactOptions: options, opacity: opacity))
+    }
+}
+
+@available(iOS 13, macOS 10.15, tvOS 13, *)
+struct SentryReplayPreviewView: UIViewRepresentable {
+    let redactOptions: SentryRedactOptions
+    let opacity: Float
+    
+    func makeUIView(context: Context) -> SentryReplayMaskPreviewUIView {
+        return SentryReplayMaskPreviewUIView(redactOptions: redactOptions)
+    }
+    
+    func updateUIView(_ uiView: SentryReplayMaskPreviewUIView, context: Context) {
+        uiView.opacity = opacity
+    }
+}
+
+#endif
--- a/Sources/SentrySwiftUI/Preview/SentryReplayMaskPreviewUIView.swift
+++ b/Sources/SentrySwiftUI/Preview/SentryReplayMaskPreviewUIView.swift
@@ -1,0 +1,84 @@
+#if canImport(SwiftUI) && canImport(UIKit) && os(iOS) || os(tvOS)
+import Sentry
+import UIKit
+
+#if CARTHAGE || SWIFT_PACKAGE
+@_implementationOnly import SentryInternal
+#endif
+
+class PreviewImageView: UIImageView {
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        return nil
+    }
+}
+
+class SentryReplayMaskPreviewUIView: UIView {
+    private let photographer: SentryViewPhotographer
+    private var displayLink: CADisplayLink?
+    private var imageView = PreviewImageView()
+    
+    var opacity: Float {
+        get { return Float(imageView.alpha) }
+        set { imageView.alpha = CGFloat(newValue)}
+    }
+    
+    init(redactOptions: SentryRedactOptions) {
+        self.photographer = SentryViewPhotographer(renderer: PreviewRederer(), redactOptions: redactOptions)
+        super.init(frame: .zero)
+        self.isUserInteractionEnabled = false
+        imageView.isUserInteractionEnabled = false
+        imageView.sentryReplayUnmask()
+    }
+    
+    deinit {
+        displayLink?.invalidate()
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func didMoveToSuperview() {
+        if ProcessInfo.processInfo.environment[SENTRY_XCODE_PREVIEW_ENVIRONMENT_KEY] == "1" {
+            displayLink = CADisplayLink(target: self, selector: #selector(update))
+            displayLink?.add(to: .main, forMode: .common)
+        } else {
+            print("[SENTRY] [WARNING] SentryReplayMaskPreview is not meant to be used in your app, only with SwiftUI Previews.")
+        }
+    }
+    
+    @objc
+    private func update() {
+        guard let window = self.window else { return }
+        self.photographer.image(view: window) { image in
+            DispatchQueue.main.async {
+                self.showImage(image: image)
+            }
+        }
+    }
+    
+    private func showImage(image: UIImage) {
+        guard let window = super.window else { return }
+        if imageView.superview != window {
+            window.addSubview(imageView)
+        }
+        imageView.image = image
+        imageView.frame = window.bounds
+    }
+    
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        return nil
+    }
+}
+
+class PreviewRederer: ViewRenderer {
+    func render(view: UIView) -> UIImage {
+        return UIGraphicsImageRenderer(size: view.frame.size, format: .init(for: .init(displayScale: 1))).image { _ in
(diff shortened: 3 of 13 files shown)


Samples/iOS-SwiftUI/iOS-SwiftUI/SwiftUIApp.swift
```diff
@@ -10,7 +10,7 @@ struct SwiftUIApp: App {
             options.debug = true
             options.tracesSampleRate = 1.0
             options.profilesSampleRate = 1.0
-            options.sessionReplay.sessionSampleRate = 1.0
+            options.sessionReplay.sessionSampleRate = 0.0
```
brustolin, line 13:
```suggestion
            options.sessionReplay.sessionSampleRate = 1.0
```

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
```diff
@@ -0,0 +1,43 @@
...
+    func makeUIView(context: Context) -> UIView {
+        let view = SentryReplayMaskPreviewUIView(redactOptions: redactOptions)
+        view.isUserInteractionEnabled = false
+        return view
+    }
+    
+    func updateUIView(_ uiView: UIView, context: Context) {
+        (uiView as? SentryReplayMaskPreviewUIView)?.opacity = opacity
+    }
+}
```
philprime, line 41: `l`: Correct me if I am wrong, but you can remove the type casting by defining the return type directly:

```suggestion
    func makeUIView(context: Context) -> SentryReplayMaskPreviewUIView {
        let view = SentryReplayMaskPreviewUIView(redactOptions: redactOptions)
        view.isUserInteractionEnabled = false
        return view
    }
    
    func updateUIView(_ uiView: SentryReplayMaskPreviewUIView, context: Context) {
        uiView.opacity = opacity
    }
}
```

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
```diff
@@ -0,0 +1,43 @@
...
+
+@available(iOS 13, macOS 10.15, tvOS 13, *)
+struct SentryReplayPreviewView: UIViewRepresentable {
+    let redactOptions: SentryRedactOptions
+    let opacity: Float
+    
+    func makeUIView(context: Context) -> UIView {
+        let view = SentryReplayMaskPreviewUIView(redactOptions: redactOptions)
+        view.isUserInteractionEnabled = false
```
philprime, line 34: `m`: Did you test this with older iOS versions, that it does not block interactive elements under the overlay? We might have to implement a hit-test proxy view.

Sources/Swift/Tools/SentryViewPhotographer.swift
```diff
@@ -20,24 +20,24 @@ class DefaultViewRenderer: ViewRenderer {
 }
 
 @objcMembers
-class SentryViewPhotographer: NSObject, SentryViewScreenshotProvider {
+public class SentryViewPhotographer: NSObject, SentryViewScreenshotProvider {
```
philprime, line 23: @philipphofmann will probably not agree with changing this to public, we should at least somehow mark this class as "internal use only"

brustolin: We should stop using hacky solutions and add a lot of "code smells" into the code.
If the user try to use an undocumented class, that is not part of the main product, let them try.

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
brustolin: So weird. I have being using `isUserInteractionEnabled` since forever to ignore hits. It was never a problem to me, but it looks like its a problem for SwiftUI.

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreviewUIView.swift
```diff
@@ -0,0 +1,71 @@
+#if canImport(SwiftUI) && canImport(UIKit) && os(iOS) || os(tvOS)
+import Sentry
+import SwiftUI
+import UIKit
+
+#if CARTHAGE || SWIFT_PACKAGE
+@_implementationOnly import SentryInternal
+#endif
+
+class SentryReplayMaskPreviewUIView: UIView {
```
philipphofmann, line 10: `h`: We can move this to the main Sentry package as it has no dependencies on SwiftUI. @brustolin I think we could also use this for non SwiftUI apps. We could port this so people can use this with an option for any iOS app while debugging. With that we could also keep the `SentryViewPhotographer` internal.

philipphofmann (commented):
I see potential, that we can use the solution for debugging any iOS app. If I'm not mistaken we don't need to limit this for SwiftUI Previews. Or am I missing something, @brustolin?

brustolin:
> I see potential, that we can use the solution for debugging any iOS app

Yes, technically we could also use it while running in debug.
Step-by-step

Sources/Sentry/SentrySDK.m
```diff
@@ -203,14 +203,16 @@ + (void)setStartTimestamp:(NSDate *)value
 
 + (void)startWithOptions:(SentryOptions *)options
 {
+    // We save the options before checking for xcode preview because
+    // we will use this options in the preview
+    startOption = options;
```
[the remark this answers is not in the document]
brustolin: https://github.com/getsentry/sentry-cocoa/blob/main/Tests/SentryTests/SentrySDKTests.swift#L560-L563

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
```diff
@@ -0,0 +1,43 @@
...
+    let opacity: Float
+    func body(content: Content) -> some View {
+        content.overlay(SentryReplayPreviewView(redactOptions: redactOptions, opacity: opacity))
+    }
+}
+
+@available(iOS 13, macOS 10.15, tvOS 13, *)
+public extension View {
+    func sentryReplayPreviewMask(redactOptions: SentryRedactOptions? = nil, opacity: Float = 1) -> some View {
+        let options = redactOptions ?? SentrySDK.options?.sessionReplay ?? PreviewRedactOptions()
```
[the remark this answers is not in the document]
brustolin: It’s up to the user to decide how they will use this. They might try different configurations directly from the preview without changing the project options.

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreviewUIView.swift
brustolin: We could, not sure if we should. It think session replay should have its own lib.
Moving this to main just change what needs to be public, because this would need to be public in order for SentrySwiftUI to use it.

Lets go step-by-step.
After finishing this PR for SwiftUI we can debate debugging it for UIKit that has no preview.

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreviewUIView.swift
```diff
@@ -0,0 +1,71 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func didMoveToSuperview() {
+        if ProcessInfo.processInfo.environment[SENTRY_XCODE_PREVIEW_ENVIRONMENT_KEY] == "1" {
+            displayLink = CADisplayLink(target: self, selector: #selector(update))
+            displayLink?.add(to: .main, forMode: .common)
+        } else {
+            print("[SENTRY] [WARNING] SentryReplayMaskPreview is not meant to be used in your app, only with SwiftUI Previews.")
+        }
```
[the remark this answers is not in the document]
brustolin: > If somebody wants to use it...

This could be said for so many things. 
I choose to make it harder for the user to forget this on the app by mistake.

We can discuss this approach, but so far we had zero request for it.

Sources/Swift/Protocol/SentryRedactOptions.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 @objc
-protocol SentryRedactOptions {
+public protocol SentryRedactOptions {
```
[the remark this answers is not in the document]
brustolin: IMO protocol  is the best approach, it gives a lot of flexibility and separation of concerns.

We can make SentryReplayOptions implement SentryRedactOptionsSentry without the need to have another level in the replay options.

And we can have a simpler object for screenshot redact, and then another one for preview redact, maybe with more options in there specific for preview.

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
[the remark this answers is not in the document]
philipphofmann: OK that makes sense. Thanks.

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreviewUIView.swift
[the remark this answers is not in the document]
philipphofmann: What about allowing it and logging a warning when we see it's used elsewhere?

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreviewUIView.swift
brustolin: This is just a matter of deciding what we want to make public. SentryReplayMaskPreviewUIView should not be directly used by the user. There is a correct way to use it, but most users will likely find it uncomfortable or inconvenient.

This arbitrary decision about what should be public and what should not is only making our lives harder without providing any real benefit. Users only care about having code snippets they can copy and paste to make Sentry work.

brustolin: This is just a matter of deciding what we want to make public. SentryReplayMaskPreviewUIView should not be directly used by the user either. There is a correct way to use it and most users will not know how to correct use it.

I believe we're overthinking this "what should be public and what should not" thing, and this only makes our lives harder without providing any real benefit. Users only care about having code snippets they can copy and paste to make Sentry work.

philipphofmann: > I believe we're overthinking this "what should be public and what should not" thing, and this only makes our lives harder without providing any real benefit. Users only care about having code snippets they can copy and paste to make Sentry work.

If users can use it, they will, and then we will break them, which I would like to avoid.

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -561,6 +558,13 @@ class SentrySDKTests: XCTestCase {
         SentrySDK.start(options: fixture.options)
         XCTAssertEqual(SentrySDK.options, fixture.options)
     }
+    
+    func testGlobalOptionsForPreview() {
```
philipphofmann, line 562: Thanks 👍

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreviewUIView.swift
brustolin: > If users can use it, they will

That's the part I believe we're overthinking.

I will open another PR to start the UIKit debug feature and reorganize the public code.

Sources/Swift/Protocol/SentryRedactOptions.swift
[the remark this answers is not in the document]
philipphofmann: Do we need this flexibility? This flexibility increases the complexity. We then have to maintain different classes, and users could get confused by the various options. Users are used to one big option class for everything. I'm unsure if this is the right place to deviate from that approach. Furthermore, I think most users will simply have one configuration of the `SentryReplayOptions`. Why should you have a different one in the preview? I have troubles coming up with good use cases.

Sources/Swift/Tools/SentryViewPhotographer.swift
philipphofmann: I disagree with simply allowing users to use undocumented classes. If you allow it, people will use it, and then you might break them. I'm OK with making something public that shouldn't be use, if we prefix with something like InternalUseOnly_, so it's 100% clear. We already used that approach with `PrivateSentrySDKOnly` and never got complaints.

philipphofmann (commented):
I would prefix the ViewPhotographer class with something like `InternalUseOnly_` and keep the `SentryReplayMaskPreviewUIView` internal. With that approach, we could still move `SentryReplayMaskPreviewUIView` to the core package and add use it for UIKit later on. Furthermore, we could make the `ViewPhotographer` internal again, if we decided to.

Sources/Swift/Protocol/SentryRedactOptions.swift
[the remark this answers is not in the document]
brustolin: We're already using this flexibility. `SentryReplayOptions` implements `SentryRedactOptions` and so does `SentryRedactDefaultOptions` that is used for screenshot masking.

Also, depending on how the SDK is initialized for SwiftUI, the global options might not be available for preview, hence the need for a redactOptions argument in the preview modifier.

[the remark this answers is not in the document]
philipphofmann: Ah OK. Thanks for explaining. All good 👍

Issue #4687, opened 2025-01-08
Implement Sampling Seed Propagation


kahest:
### Description

See https://github.com/getsentry/projects/issues/316

To increase the chance of capturing complete traces when users return a new sample rate tracesSampler in backend services, we propagate the random value used by the SDK for computing the sampling decision instead of creating a new random value in every service. Therefore, across a trace every SDK uses the same random value.

[Develop: **Spec**](https://develop.sentry.dev/sdk/telemetry/traces/#propagated-random-value)

### See also

Sample implemeantion in PHP https://github.com/getsentry/sentry-php/pull/1793.

philprime:
I understand that we neeed to add the `sample_rate` header to the baggage, but do we also need to implement this "[..] the sampling decision (trailing `-1` or `-0` from the `sentry-trace` header) are present in the incoming trace [..]`?

We already send a header `sampled` in the baggage.

philprime:
Update from my side: Looking at the encoding logic in `SentryTraceHeader.m`, it seems like the value `sampled` is already appended to the header string here:

https://github.com/getsentry/sentry-cocoa/blob/ba3679be555bbd8fb5c0a6744e30da5d758cc36f/Sources/Sentry/SentryTraceHeader.m#L25-L31

PR #4751, opened 2025-01-24
feat: add sample_rand / fix sample_rate in baggage

philprime:
## :scroll: Description

- Add `sample_rand` to the baggage
- Fixed `sample_rate` not added to baggage
- Add additional initializer tests for public interfaces 

See #4687

- I kept the original tests to make sure the behaviour doesn't change
- I added a lot of additional tests to ensure a stable API for the public headers
- Used guidance here: https://github.com/getsentry/projects/issues/316#issuecomment-2612117275
- CI
- Checked the envelope header using a proxy debugging tool

## :pencil: Checklist

## :crystal_ball: Next steps

- [x] Perform integrations tests


Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -39,7 +39,8 @@
     SentryTracerConfiguration *config = [SentryTracerConfiguration defaultConfiguration];
     config.profilesSamplerDecision =
         [[SentrySamplerDecision alloc] initWithDecision:kSentrySampleDecisionYes
-                                          forSampleRate:profilesRate];
+                                          forSampleRate:profilesRate
+                                         withSampleRand:@0.9999];
```
philprime, line 43: Not sure if this is valid, the decision is hard-coded to `YES`, so I would go with `1.0` sample rand, but the docs mention the range to be `[0,1)` excluding `1.0`

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -153,9 +155,9 @@
     SENTRY_LOG_INFO(@"Starting app launch trace profile at %llu.", getAbsoluteTime());
     sentry_isTracingAppLaunch = YES;
     sentry_launchTracer =
-        [[SentryTracer alloc] initWithTransactionContext:sentry_context(tracesRate)
+        [[SentryTracer alloc] initWithTransactionContext:sentry_context(tracesRate, @1.0)
```
philprime, line 158: Is setting the random value to `1.0` the correct way? Should we set it to something < 1.0?

Sources/Sentry/Public/SentryBaggage.h
```diff
@@ -67,6 +74,17 @@ NS_SWIFT_NAME(Baggage)
                         sampled:(nullable NSString *)sampled
                        replayId:(nullable NSString *)replayId;
 
+- (instancetype)initWithTraceId:(SentryId *)traceId
```
philprime, line 77: As the `SentryBaggage.h` is public adding the `sampleRand` parameter could break other SDKs using sentry-cocoa.

I currently see 3 options:

1. Change the public API and potentially cause crashes, keep only one method --> cleaner
2. Add the second method with the additional parameter --> SDK users accessing this method must make sure to use the one with `sampleRand`
3. Add the second method with the additional parameter and deprecate the original one --> SDK users would get a warning, but we are treating deprecation as hard errors in Swift, so tests won't compile

Sources/Sentry/Public/SentrySpanContext.h
```diff
@@ -78,6 +88,19 @@ SENTRY_NO_INIT
  */
 - (instancetype)initWithOperation:(NSString *)operation sampled:(SentrySampleDecision)sampled;
 
+/**
```
philprime, line 91: Same as in https://github.com/getsentry/sentry-cocoa/pull/4751#discussion_r1930846919

Sources/Sentry/Public/SentrySpanContext.h
```diff
@@ -91,6 +114,23 @@ SENTRY_NO_INIT
                       operation:(NSString *)operation
                         sampled:(SentrySampleDecision)sampled;
 
+/**
```
philprime, line 117: Ssame as in https://github.com/getsentry/sentry-cocoa/pull/4751#discussion_r1930846919

Sources/Sentry/Public/SentryTraceContext.h
```diff
@@ -75,6 +80,20 @@ NS_SWIFT_NAME(TraceContext)
                         sampled:(nullable NSString *)sampled
                        replayId:(nullable NSString *)replayId;
 
+/**
+ * Initializes a SentryTraceContext with given properties.
+ */
+- (instancetype)initWithTraceId:(SentryId *)traceId
```
philprime, line 86: Same as in https://github.com/getsentry/sentry-cocoa/pull/4751#discussion_r1930846919

Sources/Sentry/Public/SentryTransactionContext.h
```diff
@@ -56,6 +61,17 @@ SENTRY_NO_INIT
                    operation:(NSString *)operation
                      sampled:(SentrySampleDecision)sampled;
 
+/**
```
philprime, line 64: Same as in https://github.com/getsentry/sentry-cocoa/pull/4751#discussion_r1930846919

Sources/Sentry/Public/SentryTransactionContext.h
```diff
@@ -71,6 +87,23 @@ SENTRY_NO_INIT
                 parentSpanId:(nullable SentrySpanId *)parentSpanId
                parentSampled:(SentrySampleDecision)parentSampled;
 
+/**
```
philprime, line 90: Same as in https://github.com/getsentry/sentry-cocoa/pull/4751#discussion_r1930846919

philprime:
Back to draft because I noticed missing test cases

Sources/Sentry/Public/SentryBaggage.h
buenaflor: > could break other SDKs using sentry-cocoa

do you know which one's could break? e.g flutter has its own implementation for distributed tracing and react native I think inherits it from javascript, dont think this change would affect those

brustolin (commented):
I do have one comment that I think would short this PR a lot.

Sources/Sentry/Public/SentryBaggage.h
philprime: Not particularly, but as the header is also in the `public` folder I am not sure who is actually able to access it. Maybe even SDK user? cc @philipphofmann

philipphofmann: I would go with option 3. We don't need to worry much about the deprecation because I'm sure most users don't use this class.

Public means it's public. Every user can access it.

Tests/SentryTests/Transaction/SentryBaggageTests.swift
```diff
@@ -15,4 +17,157 @@ class SentryBaggageTests: XCTestCase {
         
         XCTAssertEqual(header, "sentry-public_key=publicKey,sentry-trace_id=00000000000000000000000000000000")
     }
+
+    // MARK: - Tests with sampleRand
```
philipphofmann, line 21: Thanks for the extensive tests 🚀

Sources/Sentry/Public/SentrySpanProtocol.h
```diff
@@ -39,6 +39,16 @@ NS_SWIFT_NAME(Span)
...
 
+/**
+ * Rate of sampling
+ */
+@property (nullable, nonatomic, strong) NSNumber *sampleRate;
+
+/**
+ * Random value used to determine if the span is sampled.
+ */
+@property (nullable, nonatomic, strong) NSNumber *sampleRand;
```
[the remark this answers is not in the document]
philprime: Furthermore, I assumed the value to be relevant because the spans already have a field `sampled`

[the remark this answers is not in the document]
philipphofmann: I agre with @brustolin: [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and [KISS](https://en.wikipedia.org/wiki/KISS_principle). Please let's not add things we don't need yet. Things change so quickly at Sentry. Let's keep spans out of this. I also don't see any changes for spans here https://github.com/getsentry/sentry-php/pull/1793 and the develop docs don't mention spans for this feature.

philipphofmann (commented):
I will give this a full review once we remove the span changes, as this will significantly reduce the scope of the PR.

Sources/Sentry/Public/SentryBaggage.h
philprime: @philipphofmann I tried option 3 during development and calling the initializers from from Swift failed to compile because deprecation warnings are treated as compile-time errors.

I'll try to reproduce, maybe we need to remove the tests for the deprecated methods, or maybe I can somehow ignore it.

philipphofmann: Oh no 🤦. Maybe we should disable this warning if it gets too complicated.

Sources/Sentry/Public/SentrySpanProtocol.h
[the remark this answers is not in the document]
philprime: Removed it from this PR

Makefile
```diff
@@ -48,7 +48,12 @@ test:
 run-test-server:
 	cd ./test-server && swift build
 	cd ./test-server && swift run &
-.PHONY: run-test-server
+
+run-test-server-sync:
+	cd ./test-server && swift build
+	cd ./test-server && swift run
+
+.PHONY: run-test-server run-test-server-sync
```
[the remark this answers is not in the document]
philprime: I don't want to have a daemon test server running on my machine in the background, which I can not control anymore without opening Activity Monitor or using `kill`, especially when it relates to the currently checked out commit.

I considered this a small enough change to be part of any PR, but I am fine with adding it in another PR.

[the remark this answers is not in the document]
philipphofmann: It's OK to keep it.

Tests/SentryTests/Protocol/SentryEnvelopeTests.swift
```diff
@@ -162,6 +162,24 @@ class SentryEnvelopeTests: XCTestCase {
         XCTAssertEqual(eventId, envelopeHeader.eventId)
         XCTAssertEqual(traceContext, envelopeHeader.traceContext)
     }
+
+    func testInitSentryEnvelopeHeader_withSampleRand_SetIdAndTraceState() {
```
philipphofmann, line 166: `m`: What extra value does this test add? There aren't any changes in `SentryEnvelopeHeader`. We already have `testInitSentryEnvelopeHeader_SetIdAndTraceState` which validates that the init method of `SentryEnvelopeHeader` assigns the traceContext. Here, we only use a different initializer of the `TraceContext`. Am I missing something?

Tests/SentryTests/SentryHubTests.swift
```diff
@@ -370,28 +370,68 @@ class SentryHubTests: XCTestCase {
         XCTAssertEqual(span.sampled, .no)
     }
     
+    @available(*, deprecated) // The test is marked as deprecated to silence the deprecation warning of the initializer
```
philipphofmann, line 373: `l`: Maybe a bit cleaner.
```suggestion
    @available(*, deprecated, message: "The test is marked as deprecated to silence the deprecation warning of the initializer")
```

philipphofmann (approved):
LGTM, thanks for all the extra tests

Sources/Sentry/Public/SentryTraceContext.h
```diff
@@ -52,6 +52,11 @@ NS_SWIFT_NAME(TraceContext)
  */
 @property (nullable, nonatomic, readonly) NSString *sampleRate;
 
+/**
+ * Random value used to determine if the trace is sampled.
+ */
+@property (nullable, nonatomic, readonly) NSString *sampleRand;
```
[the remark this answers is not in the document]
philprime: Do we know the reason? I can just asume that the SentryTraceContext properties are already serialized

Tests/SentryTests/Protocol/SentryEnvelopeTests.swift
philprime: You are correct, thanks for pointing this out. I'll remove the added test

Sources/Sentry/Public/SentryTraceContext.h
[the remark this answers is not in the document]
philprime: Ok yes, it looks like the SentryTraceContext is a structure holding serialized data. I'll add a comment to the docs.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
philipphofmann: We store the `SentryLaunchProfileConfig` to disk

https://github.com/getsentry/sentry-cocoa/blob/70c49ca151ccd1a5763b1e1d316448cd5ebad948/Sources/Sentry/Profiling/SentryLaunchProfiling.m#L48-L53

We need to read the sampleRand from the stored `SentryLaunchProfileConfig`

https://github.com/getsentry/sentry-cocoa/blob/70c49ca151ccd1a5763b1e1d316448cd5ebad948/Sources/Sentry/Profiling/SentryLaunchProfiling.m#L133-L137

The FileManager should already store that info to the disk and load it to the dict here
https://github.com/getsentry/sentry-cocoa/blob/70c49ca151ccd1a5763b1e1d316448cd5ebad948/Sources/Sentry/SentryFileManager.m#L803-L816

philipphofmann: Sorry, I missed that you need to add new key similar to kSentryLaunchProfileConfigKeyTracesSampleRate and update the method here

https://github.com/getsentry/sentry-cocoa/blob/70c49ca151ccd1a5763b1e1d316448cd5ebad948/Sources/Sentry/Profiling/SentryLaunchProfiling.m#L165-L187

philprime: Added it.

philipphofmann (approved):
LGTM thanks.

PR #4761, opened 2025-01-28
feat:  Add `showMaskPreview` to `SentrySDK.replay` api to debug replay masking

brustolin:
## :scroll: Description

Add `showMaskPreview` to `SentrySDK.replay` api to debug replay masking

Samples

## :pencil: Checklist

- [] I updated the docs if needed.


Sources/Sentry/include/SentryCrashWrapper.h
```diff
@@ -9,7 +9,6 @@ NS_ASSUME_NONNULL_BEGIN
 /** A wrapper around SentryCrash for testability.
  */
 @interface SentryCrashWrapper : NSObject
-SENTRY_NO_INIT
```
brustolin, line 12: I removed this in order to be able to mock SentryCrashWrapper for tests.

Sources/Swift/Integrations/SessionReplay/Preview/SentryMaskingPreviewView.swift
```diff
@@ -0,0 +1,67 @@
...
+    private var imageView = UIImageView()
+    
+    var opacity: Float {
+        get { return Float(imageView.alpha) }
+        set { imageView.alpha = CGFloat(newValue)}
+    }
+    
+    init(redactOptions: SentryRedactOptions) {
+        self.photographer = SentryViewPhotographer(renderer: PreviewRederer(), redactOptions: redactOptions)
+        super.init(frame: CGRect(x: 0, y: 0, width: 100, height: 100))
```
armcknight, line 28: Could we use `CGRect.zero` here, since we'll set the frame to the window bounds in `-[SentrySessionReplayIntegration showMaskPreview:]`, and to the superview bounds in `didMoveToSuperview`?

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -21,6 +21,18 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
         }
     }
     
+    private class TestCrashWrapper: SentryCrashWrapper {
+        var traced = true
```
armcknight, line 25: This will always be set to true by default in the init, and will never be modified otherwise, so we could declare this thusly:
```suggestion
        let traced: Bool
```

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -21,6 +21,18 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
...
+        
+        init(traced: Bool = true) {
+            self.traced = traced
+        }
```
armcknight, line 29: This is an NSObject subclass, so I think we'd need to call `super.init()`, although I guess you don't actually want to start all the crash reporting machinery here, you just want the answer to whether the process is being debugged. Sounds like something we should refactor out of the crash wrapper as some point. 

If you really are avoiding `super.init()` for this reason, could we comment that?
```suggestion
        init(traced: Bool = true) {
            self.traced = traced
            // not calling super.init() here as we don't actually want to install crash reporter machinery
        }
```

Sources/Swift/Tools/UIImageHelper.swift
```diff
@@ -25,9 +25,8 @@ final class UIImageHelper {
         let blue = CGFloat(data[0]) / 255.0
         let green = CGFloat(data[1]) / 255.0
         let red = CGFloat(data[2]) / 255.0
-        let alpha = CGFloat(data[3]) / 255.0
```
armcknight, line 28: Does this mean that we could change line 23 to the following? I was able to run the full unit test suite without it crashing, anyways.
```
let data = pixelBuffer.bindMemory(to: UInt8.self, capacity: 3)
```

armcknight (approved):
Looks good generally! Just small stuff. i'd propose changing before merging.

Sources/Swift/Integrations/SessionReplay/Preview/SentryMaskingPreviewView.swift
brustolin: Yes. I will change. Thanks!!

Sources/Swift/Tools/UIImageHelper.swift
brustolin: Oh yes. We dont need to read all 4 bytes.

Sources/Sentry/Public/SentryReplayApi.h
```diff
@@ -44,6 +44,42 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)stop;
 
+/**
+ * Shows an overlay on the app to debug session replay masking.
```
philipphofmann, line 48: `h`: If I'm not mistaken, this doesn't work for SwiftUI and only for UIKit? If yes, please mention this in the comment to clarify this.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -615,6 +617,33 @@ - (nullable NSString *)currentScreenNameForSessionReplay
                .firstObject;
 }
 
+- (void)showMaskPreview:(CGFloat)opacity
+{
+    if ([SentryDependencyContainer.sharedInstance.crashWrapper isBeingTraced] == NO) {
+        return;
+    }
```
philipphofmann, line 624: `m`: What drove you to limit this to only when the debugger is attached?

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -615,6 +617,33 @@ - (nullable NSString *)currentScreenNameForSessionReplay
...
+
+    if (_previewView == nil) {
+        _previewView = [[SentryMaskingPreviewView alloc] initWithRedactOptions:_replayOptions];
+    }
+
+    _previewView.opacity = opacity;
+    [_previewView setFrame:window.bounds];
+    [window addSubview:_previewView];
```
philipphofmann, line 638: `h`: What happens if a user calls this method multiple times in a row? If I'm not mistaken, the code adds multiple `_previewView`.

philipphofmann, line 638: `m`: This will crash when somebody calls `SentrySDK.replay.showMaskPreview(0.5)` from a BG thread. I think we can ignore this scenario by clearly pointing out in the code docs that this method must be called from the main thread.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -615,6 +617,33 @@ - (nullable NSString *)currentScreenNameForSessionReplay
                .firstObject;
 }
 
+- (void)showMaskPreview:(CGFloat)opacity
```
philipphofmann, line 620: `m`: It bugs me a bit, that we don't have tests for that code. What can we do to increase the test coverage?

philipphofmann (commented):
I found a few potential issue, but the feature looks promising.

Sources/Sentry/Public/SentryReplayApi.h
brustolin: It does work for SwiftUI if used in runtime.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: No releasing it by mistake

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: We do have tests for showMaskPreview. With and without the debugger attached.

Sources/Sentry/SentrySessionReplayIntegration.m
brustolin: Since it's the same instance (see lines 632 - 634), it just reads the same view to the main window. The window may have changed if its an iPad app.

brustolin: It will not crash, users get a warning from UIKIt that they are handling ui in a background thread.

Issue #4768, opened 2025-01-29
feedback: Split into separate module


armcknight:
### Description

We currently check for the NSPhotoLibraryUsageDescription info plist key before including Photos UX. If the developer doesn't include that key, we won't take any codepaths that lead to requesting photos access or showing the photo picker. 

However, app store submissions would still include that code in the binary. We may need to split the user feedback implementation, or at least the photos portion of it, into a separate module that can be excluded from  builds if customers don't want to include that info plist key for whatever reason.

philprime:
A workaround we could use (with it's own downsides) could be obfuscating the string `NSPhotoLibraryUsageDescription`

armcknight:
I think the best course of action now, after sleeping on it, will be to only allow instantaneous screenshot addition, by opening a form in response to taking one, as an alternate trigger to tapping the widget button.

armcknight:
Will close in favor of #4402 since that includes the screenshot trigger.

armcknight:
Turns out that the screenshot notification doesn't provide the screenshot; it'd still need to be grabbed via Photos. So, will reopen this.

armcknight:
So, we will observe the screenshot notification, and then capture our own rendering of the view hierarchy and attach that to the triggered form display.

PR #4769, opened 2025-01-29
fix: temporarily remove Photos code

armcknight:
In service of fixing the underlying issue in https://github.com/getsentry/sentry-cocoa/issues/4768

#skip-changelog


philprime:
How about instead of commenting it out, wrapping it with a compiler flag for now?

armcknight:
I think I'm actually going to change it from using Photos to only responding to a screenshot capture for now, as an alternate trigger to show the dialog. So the form will change to only having the "remove screenshot" button, and will never have an "add screenshot" button if it wasn't opened in response to one.

armcknight:
So we cannot access the screenshot without photos access. Reopening this so we can temporarily remove screenshot functionality until we can modularize the feature.

PR #4781, opened 2025-01-31
impr: Add uptime to SDK log messages

philipphofmann:
## :scroll: Description

Add the system uptime to the SDK log messages to know precisely when the logger logs. This can help when investigating issues by reading the SDK log messages.

For investigating the raw logs of flaky tests this can be highly useful to understand exactly where a test might have hanged.

Unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Tools/SentryLog.swift
+++ b/Sources/Swift/Tools/SentryLog.swift
@@ -13,6 +13,7 @@ class SentryLog: NSObject {
     static let alwaysLevel = SentryLevel.fatal
     private static var logOutput = SentryLogOutput()
     private static var logConfigureLock = NSLock()
+    private static var dateProvider: SentryCurrentDateProvider = SentryDefaultCurrentDateProvider()
 
     @objc
     static func configure(_ isDebug: Bool, diagnosticLevel: SentryLevel) {
@@ -26,7 +27,14 @@ class SentryLog: NSObject {
     @objc
     static func log(message: String, andLevel level: SentryLevel) {
         guard willLog(atLevel: level) else { return }
-        logOutput.log("[Sentry] [\(level)] \(message)")
+        
+        // We use the uptime because it's guaranteed to be linear even when the
+        // date time changes. Furthermore, initializing date objects and using the
+        // date format adds some overhead. The uptime shows the time difference
+        // between the logs. If we ever need the actual date and time, we can
+        // reconsider.
+        let uptime = nanosecondsToTimeInterval(self.dateProvider.systemTime())
+        logOutput.log("[Sentry] [\(level)] [uptime:\(uptime)] \(message)")
     }
 
     /**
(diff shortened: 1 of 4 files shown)


armcknight (changes requested):
We can't send this data off-device, see https://developer.apple.com/documentation/bundleresources/app-privacy-configuration/nsprivacyaccessedapitypes/nsprivacyaccessedapitype?language=objc (search for `NSPrivacyAccessedAPICategorySystemBootTime`).

The only exception is derived data that calculates a duration. Otherwise, we can't send the raw value or other derivations off-device.

philipphofmann:
> We can't send this data off-device, see https://developer.apple.com/documentation/bundleresources/app-privacy-configuration/nsprivacyaccessedapitypes/nsprivacyaccessedapitype?language=objc (search for `NSPrivacyAccessedAPICategorySystemBootTime`).
> 
> The only exception is derived data that calculates a duration. Otherwise, we can't send the raw value or other derivations off-device.

@armcknight, we don't send this data off the device. We use it for only log messages in this PR. You only get access to this data if you have access to the device logs. If I'm not mistaken, this is acceptable. But still, it is a risk of accidentally sending it by mistake. I will change it to timeSinceReferenceDate or something. Thanks for pointing it out.

philprime (approved):
LGTM

Sources/Swift/Tools/SentryLog.swift
```diff
@@ -26,7 +27,13 @@ class SentryLog: NSObject {
     @objc
     static func log(message: String, andLevel level: SentryLevel) {
         guard willLog(atLevel: level) else { return }
-        logOutput.log("[Sentry] [\(level)] \(message)")
+        
+        // We use the timeIntervalSinceReferenceDate because date format is
+        // expensive and we only care about the time difference between the
+        // log messages. We don't use system uptime because of privacy concerns
+        // see: NSPrivacyAccessedAPICategorySystemBootTime.
+        let time = self.dateProvider.date().timeIntervalSinceReferenceDate
```
philprime, line 35: No change required, any reason you chose `timeIntervalSinceReferenceDate` over `timeIntervalSince1970`?

philipphofmann: No, no specific reason. Would you prefer `timeIntervalSince1970`, if yes why?

philprime: No preferred reason, coming from devOPs perspective, most logging systems use epoch time as absolute time, so relative time to 1970, while `timeIntervalSinceReferenceDate` is relative to 2001.

As we do not submit the logs anywhere else, where the time could be interpreted wrongly, I don't see an issue with using `timeIntervalSinceReferenceDate`.

philipphofmann: Ahh, I'm working too long with the Apple ecosystem. timeIntervalSinceReferenceDate doesn't seem confusing to me anymore. Will switch to timeIntervalSince1970.

philipphofmann:
@armcknight, I need your approval for merging because you requested changes. Please review again.

armcknight (approved):
Sorry about that, completely misread where these time values were going.

philipphofmann:
> Sorry about that, completely misread where these time values were going.

No worries, it's better to be safe than sorry 😃, especially for privacy-related topics that could lead to users getting their apps rejected from the app store because we messed up.

PR #4785, opened 2025-02-03
feat: Deserialize SentryBreadcrumb

philipphofmann:
Add Decodable/Deserializing of SentryBreadcrumb.

According to @denrase he added unknown in https://github.com/getsentry/sentry-cocoa/pull/2820 because we have this pattern also in Java and Dart. We don't have this pattern anywhere else in Cocoa. So, IMO we should either remove it for SentryBreadcrumb or also add it for all other classes. This decision shouldn't block this PR at the moment. I added an item to #4724, so we don't forget about it.

#skip-changelog


philprime (approved):
LGTM, even though I do not fully understand the "unknown" field

philipphofmann:
> LGTM, even though I do not fully understand the "unknown" field

When you call initWithDict on the breadcrumb class we keep unknown properties and serialize them. This gives you extra flexibility for hybrid SDKs. They could add a breadcrumb property without having the Cocoa SDK actually adding the properly. I don't see any strong use cases for this and as other classes don't have it on Cocoa, I would actually vote for removing it for simplicity.

philprime:
Did you align this with the Hybrid SDK maintainers? We should not break expected behavior.

philipphofmann:
> Did you align this with the Hybrid SDK maintainers? We should not break expected behavior.

I will do before merging https://github.com/getsentry/sentry-cocoa/pull/4724.

PR #4792, opened 2025-02-05
feat: Deserialize SentryEvent

philipphofmann:
Add Decodable/Deserializing of SentryEvent.

#skip-changelog

### DIFF
--- a/Sources/Sentry/Public/SentryEvent.h
+++ b/Sources/Sentry/Public/SentryEvent.h
@@ -196,4 +196,24 @@ NS_SWIFT_NAME(Event)
 
 @end
 
+/**
+ * Subclass of SentryEvent so we can add the Decodable implementation via a Swift extension. We need
+ * this due to our mixed use of public Swift and ObjC classes. We could avoid this class by
+ * converting SentryReplayEvent back to ObjC, but we rather accept this tradeoff as we want to
+ * convert all public classes to Swift in the future. This class needs to be public as we can't add
+ * the Decodable extension implementation to a class that is not public.
+ *
+ * @note: We can’t add the extension for Decodable directly on SentryEvent, because we get an error
+ * in SentryReplayEvent: 'required' initializer 'init(from:)' must be provided by subclass of
+ * 'Event' Once we add the initializer with required convenience public init(from decoder: any
+ * Decoder) throws { fatalError("init(from:) has not been implemented")
+ * }
+ * we get the error initializer 'init(from:)' is declared in extension of 'Event' and cannot be
+ * overridden. Therefore, we add the Decodable implementation not on the Event, but to a subclass of
+ * the event.
+ */
+@interface SentryEventDecodable : SentryEvent
+
+@end
+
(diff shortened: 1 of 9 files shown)


Sources/Sentry/Public/SentryEvent.h
```diff
@@ -196,4 +196,24 @@ NS_SWIFT_NAME(Event)
...
+ * @note: We can’t add the extension for Decodable directly on SentryEvent, because we get an error
+ * in SentryReplayEvent: 'required' initializer 'init(from:)' must be provided by subclass of
+ * 'Event' Once we add the initializer with required convenience public init(from decoder: any
+ * Decoder) throws { fatalError("init(from:) has not been implemented")
+ * }
+ * we get the error initializer 'init(from:)' is declared in extension of 'Event' and cannot be
+ * overridden. Therefore, we add the Decodable implementation not on the Event, but to a subclass of
+ * the event.
+ */
+@interface SentryEventDecodable : SentryEvent
```
philipphofmann, line 215: Please let me know your thoughts about this somewhat ugly but practical workaround. I'm open to suggestions on how we can avoid this. As already pointed out, I don't want to convert the SentryReplayEvent back to ObjC, but maybe it's a better approach than this one.

Sources/Swift/Protocol/Codable/SentryEventCodable.swift
```diff
@@ -0,0 +1,107 @@
...
+        
+        if let rawDebugMeta = try container.decodeIfPresent([String: [DebugMeta]].self, forKey: .debugMeta) {
+            self.debugMeta = rawDebugMeta["images"]
+        }
+        
+        self.breadcrumbs = try container.decodeIfPresent([Breadcrumb].self, forKey: .breadcrumbs)
+        self.request = try container.decodeIfPresent(SentryRequest.self, forKey: .request)
+    }
+    
+    private func decodeTimeIntervalSince1970(
```
philprime, line 98: `l`: Should this be a decoding helper outside of this type?

Sources/Sentry/Public/SentryEvent.h
philprime: I agree with not converting back to Objective-C.

If this approach works, It seems fine to me.

philprime (approved):

Sources/Swift/Protocol/Codable/SentryEventCodable.swift
philipphofmann: We're currently only using it in this class. Once we use it somewhere else, I would move it. Is that OK for you?

philipphofmann: I'm going to follow up with another PR on this comment if required.

philipphofmann: If fixed this problem with https://github.com/getsentry/sentry-cocoa/pull/4802.

Issue #4402, opened 2024-10-04
Other triggers


armcknight:
### Description

Besides the widget button, implement other triggers that will open the form:
- [ ] shake gesture
- [ ] screenshot
- [ ] launch after a crash

armcknight:
@bruno-garcia please see my two questions underneath the screenshot bullet point in the description here.

bruno-garcia:
> Should there be a configuration option to either immediately show the form, or to first prompt the user that they could open a prefilled feedback form?

Not sure I follow exactly the options, but I'd suggest going with the simplest/most practical to start, and avoiding more options since things can get complicated and overwhelming to the user when they are faced with them.

> also, a configuration option to redact or not? or, should this also be a user-facing option? maybe tapping the screenshot preview opens a full-screen preview, where there is the option to remove the redactions?

I don't think we should redact anything if the user is providing the screenshot. It keeps things simple (most important) but also it's the user' intent to send us that file, masking will only get in the way

codr-tw:
> manual function call (eg SentrySDK.showUserFeedbackForm): https://github.com/getsentry/sentry-cocoa/pull/5236

This is checked marked but the referencing PR is actually a **toggle to show the "widget"** not the user feedback form.

philippsuess:
Thanks for reporting this @codr-tw. That's correct. The SDK offers ` SentrySDK.feedback.showWidget()`,` SentrySDK.feedback.showWidget()`, and `SentrySDK.feedback.hideWidget()`, but no API to show the user feedback form. 

On [Android](https://docs.sentry.io/platforms/android/user-feedback/), you can use the following code to show the form
```kotlin
Sentry.showUserFeedbackDialog(sentryId, options -> {
  // The options set here will be applied to the current form only
  options.setFormTitle("We want to hear from you!");
});
```

@bruno-garcia, I'm unsure about the other options mentioned in the issue description: shake gesture and launch after crash. As far as I know, we don't have these on Android or other platforms. My proposal is to close this issue and create a new one for adding `showUserFeedbackDialog` to Cocoa. Do you agree, @bruno-garcia?

bruno-garcia:
[...]

PR #4801, opened 2025-02-06
fix(feedback): change photos picker to screenshot trigger

armcknight:
for https://github.com/getsentry/sentry-cocoa/issues/4402

We decided we can't ship Photos usage because then everyone using the sdk would have to include the plist key for it, or risk app store rejection. We may add it back in as a separate module in the future. For now, we will detect screenshots and append a screenshot with our own capture of the view hierarchy. See the linked issue for ideas on how we may need to refine the UX.

#skip-changelog


philipphofmann (commented):
The overall approach, LGTM. Thanks @armcknight.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackFormViewModel.swift
```diff
@@ -245,11 +242,27 @@ class SentryUserFeedbackFormViewModel: NSObject {
...
     lazy var submitButtonHeightConstraint = submitButton.heightAnchor.constraint(equalToConstant: formElementHeight * config.scaleFactor)
     lazy var cancelButtonHeightConstraint = cancelButton.heightAnchor.constraint(equalToConstant: formElementHeight * config.scaleFactor)
-    lazy var screenshotImageAspectRatioConstraint = screenshotImageView.widthAnchor.constraint(equalTo: screenshotImageView.heightAnchor)
+    lazy var screenshotImageAspectRatioConstraint = {
+        let aspectRatio: CGFloat
+        if let screenshot = self.screenshot {
+            if screenshot.size.height == 0 {
+#if !SENTRY_TEST
+                SentryLog.warning("Image had 0 height, won't be able to set a reasonable aspect ratio. Defaulting to 1:1.")
+#endif // !SENTRY_TEST
```
philipphofmann, line 254: `m`: Why do we want to disable this warning log in the tests? If that's spamming our test logs we could consider disabling the logs for this specific test. You could use 
```
SentryLog.withOutLogs {
...
}
```

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackFormViewModel.swift
```diff
@@ -245,11 +242,27 @@ class SentryUserFeedbackFormViewModel: NSObject {
...
     lazy var emailTextFieldHeightConstraint = emailTextField.heightAnchor.constraint(equalToConstant: formElementHeight * config.scaleFactor)
-    lazy var addScreenshotButtonHeightConstraint = addScreenshotButton.heightAnchor.constraint(equalToConstant: formElementHeight * config.scaleFactor)
     lazy var removeScreenshotButtonHeightConstraint = removeScreenshotButton.heightAnchor.constraint(equalToConstant: formElementHeight * config.scaleFactor)
     lazy var submitButtonHeightConstraint = submitButton.heightAnchor.constraint(equalToConstant: formElementHeight * config.scaleFactor)
     lazy var cancelButtonHeightConstraint = cancelButton.heightAnchor.constraint(equalToConstant: formElementHeight * config.scaleFactor)
-    lazy var screenshotImageAspectRatioConstraint = screenshotImageView.widthAnchor.constraint(equalTo: screenshotImageView.heightAnchor)
+    lazy var screenshotImageAspectRatioConstraint = {
+        let aspectRatio: CGFloat
+        if let screenshot = self.screenshot {
+            if screenshot.size.height == 0 {
```
philipphofmann, line 251: `m`: Maybe it's safer to check for both height and width not being 0.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
```diff
@@ -50,6 +47,8 @@ struct SentryUserFeedbackWidget {
                     constraints.append(button.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor, constant: config.widgetConfig.layoutUIOffset.horizontal))
                 }
                 NSLayoutConstraint.activate(constraints)
+                
+                NotificationCenter.default.addObserver(self, selector: #selector(userCapturedScreenshot), name: UIApplication.userDidTakeScreenshotNotification, object: nil)
```
philipphofmann, line 51: `h`: I'm a bit confused. This fires when the user takes a screenshot, but we don't use the screenshot instead we take our own screenshot. Why do we need to subscribe to this?

philipphofmann, line 51: `h`: If we need to subscribe to this, we also must remove the observer at some point.

armcknight: We can't access the real screenshot without pulling it from the photos library, which brings us to the same problem of plist key requirement that we're working around by doing this

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackFormViewModel.swift
armcknight: This is just to avoid a div0 crash. If we have an image with 0 width, presentation is going to be borked regardless of whether or aspect ratio is 0 or 1

philipphofmann (approved):
LGTM, even though we have one unresolved `h` comment. I let you decide how to handle this @armcknight.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
armcknight: There are two flows as of now: 
1. the user opens the feedback form with the widget button
2. the user takes a screenshot, which we respond to by opening the form and affixing our version of the screenshot
I'll leave it to @bruno-garcia and @jas-kas to decide what they want to do with the UX, we can leave this to a later PR, but it shouldn't block this as this is more of a workaround to fix a potential issue with app store rejections.

PR #4802, opened 2025-02-06
impr: Add date encoding strategy

philipphofmann:
Add a custom date encoding strategy, which supports time intervals since 1970 and ISO 8601 formatted strings.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Protocol/SentryEventTests.swift
+++ b/Tests/SentryTests/Protocol/SentryEventTests.swift
@@ -109,6 +109,7 @@ class SentryEventTests: XCTestCase {
 
     func testDecode_WithAllProperties() throws {
         // Arrange
+        SentryDependencyContainer.sharedInstance().dateProvider = TestCurrentDateProvider()
         let event = TestData.event
         // Start timestamp is only serialized if event type is transaction
         event.type = "transaction"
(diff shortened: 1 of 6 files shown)


Tests/SentryTests/Protocol/SentryEventTests.swift
```diff
@@ -109,7 +109,6 @@ class SentryEventTests: XCTestCase {
 
     func testDecode_WithAllProperties() throws {
         // Arrange
-        SentryDependencyContainer.sharedInstance().dateProvider = TestCurrentDateProvider()
```
philprime, line 112: Why is the date provider not needed anymore?

philprime (approved):
LGTM but tests are failing

Tests/SentryTests/Protocol/SentryEventTests.swift
philipphofmann: The event uses the current date per default. You need to set the TestCurrentDateProvider so it uses a deterministic date. Before this PR the tests sometimes failed because of the missing milliseconds precision. I added it back to surface such potential issues again. But the major downside is that it will lead to not deterministic tests. As this PR fixes the underlying issue, I think it's better to go back to deterministic tests. So I added it again.

PR #4803, opened 2025-02-06
fix: Remove internal unknown dict for Breadcrumbs

philipphofmann:
## :scroll: Description

Remove unknown dict when initializing the crumbs with a dict added in #2820. The Java SDK, for example, implements this pattern to keep unknown JSON properties when deserializing and serializing classes. This allows more flexibility when a hybrid SDK or the native SDK serializes an event to disk, and the Java SDK deserializes this event. With the unknown dict, the Java SDK doesn't need to implement all the properties other SDKs have. We only have this for crumbs on the Cocoa SDK, and it doesn't make sense only to keep it for one class. Instead, we removed it to simplify the code because we would need to implement some complex deserialization logic with Swift Decodable for deserializing events. We can always reconsider this approach, but then we would need to add the unknown property concept for all serializable/decodable classes.

Came up in https://github.com/getsentry/sentry-cocoa/pull/4785

Unit tests.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@
 
 - Fix missing `sample_rate` in baggage (#4751)
 
+### Internal
+
+- Remove internal unknown dict for Breadcrumbs (#4803) This potentially only impacts hybrid SDKs.
+
+
 ## 8.44.0
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


philprime (approved):
LGTM

krystofwoldrich:
[RN](https://github.com/getsentry/sentry-react-native/blob/fd54b38f817bdc5b64f3884644ebde307fba68ec/packages/core/ios/RNSentryBreadcrumb.m#L6) doesn't use any unknown fields.

PR #4808, opened 2025-02-06
chore(ci): refactor xcode-test from positional to named arguments

philprime:
Changes the `scripts/xcode-test.sh` to use named arguments instead of positional.

This should improve usability of the script, and reduce accidental usage errors when changing the script.

Furthermore it surfaced that the 7th argument `${{ matrix.scheme }}` is actually used as the `DERIVED_DATA` path, instead of the 8th argument

https://github.com/getsentry/sentry-cocoa/blob/bdd896e6bef9469b6d43891fed3df6d6bd019cb9/.github/workflows/test.yml#L188

https://github.com/getsentry/sentry-cocoa/blob/bdd896e6bef9469b6d43891fed3df6d6bd019cb9/scripts/xcode-test.sh#L11-L18

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -102,8 +102,8 @@ jobs:
           run-id: ${{ env.FRAMEWORK_RUN_ID }}
 
       - name: Check XCFramework
-        #We dont compile the framework during releases to not change the artefact SHA value
-        #instead we use the one archive as an artefact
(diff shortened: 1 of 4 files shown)


philipphofmann (approved):
Thanks, LGTM.

philprime:
Blocked by #4809

armcknight (approved):
Long overdue, thanks!

PR #4814, opened 2025-02-08
fix(ci): revert back to xcpretty from xcbeautify

armcknight:
#skip-changelog


Makefile
```diff
@@ -3,15 +3,15 @@ init:
...
+	rbenv exec bundle install
 	clang-format --version | awk '{print $$3}' > scripts/.clang-format-version
 	swiftlint version > scripts/.swiftlint-version
 	
 # installs the tools needed to test various CI tasks locally
 init-ci: init
 	brew bundle --file Brewfile-ci
-	rbenv install --skip-existing
-	rbenv exec gem update bundler
-	rbenv exec bundle install
```
armcknight, line 14: This is the only part that wasn't strictly part of reverting the previous changes. I noticed this while looking at the makefile: rbenv stuff should only be done locally, not on ci (see xcode-test.sh and IS_LOCAL_BUILD)

scripts/xcode-test.sh
```diff
@@ -79,38 +89,35 @@ case $COMMAND in
...
-        tee raw-build-for-testing-output.log |
-        xcbeautify
+        -destination "$DESTINATION" -quiet \
+        build-for-testing \
+        | xcpretty
 fi
 
 if [ $RUN_TEST_WITHOUT_BUILDING == true ]; then
-    set -o pipefail && NSUnbufferedIO=YES xcodebuild \
+    set -o pipefail && env NSUnbufferedIO=YES && set -o pipefail && xcodebuild \
```
armcknight, line 114:
```suggestion
    set -o pipefail && env NSUnbufferedIO=YES xcodebuild \
```

armcknight:
The last CI failures are due to not having the right invocation of `xcpretty`, which I think should be fixed in https://github.com/getsentry/sentry-cocoa/pull/4815, or at least by applying the same strategy in there

Makefile
philprime: Why should `rbenv` stuff not be done in CI?

philipphofmann:
As discussed on [Slack](https://sentry.slack.com/archives/C040TTUKQT1/p1738975604257829), it might be better to stick to xcbeautify because it's maintained, and we rather want to use a maintained tool.

Furthermore, we can fix not finding failing tests with a simple grep and we can disable the verbose GH actions annotations by removing the github-actions renderer of xcbeautify.

cpisciotta:
Hey all 👋 I'm Charles – the maintainer of xcbeautify. Just came across this PR, so I was naturally inclined to understand what's motivating a potential reversion to xcpretty. Can you share what pain points you're facing with xcbeautify? I'd be happy to help where I can.

philipphofmann:
@cpisciotta, we are currently trying to make it work with https://github.com/getsentry/sentry-cocoa/pull/4828. If we have more problems, we will let you know. Thanks a lot for chiming in.

PR #4815, opened 2025-02-08
impr(ci): automatically detect ci/local environments

armcknight:
I realized after making the changes in #4814 that there has to be a better way to differentiate between ci and local runs. Instead of relying on a parameter to the script, just check one of the environment variables set by github actions.

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -212,7 +212,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build for Debug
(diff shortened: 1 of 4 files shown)


philprime (approved):
LGTM, the `CI` flag is in general a good indicator

armcknight:
after rebasing onto `main`, there are currently no other places this would be used. there are still places where fastlane is called without rbenv, i will revisit this and fix those up in a future pr if i hit a need to run fastlane locally (which i haven't done in a while)

PR #4816, opened 2025-02-08
impr(ci): use brew bundle to install dependencies, and split test/dep…

armcknight:
I noticed that the brewfile for ci tooling was never actually used in CI as I should've done in #4593 

also further differentiates the ci tooling by test and deploy, to avoid brew installing carthage on CI where it isn't even needed (esp. since it comes preinstalled in ci)

#skip-changelog


philprime:
Why is installing carthage again, if it is already installed an issue worth bringing in two Brewfiles?

philprime (commented):
LGTM, see my comment on multiple Brewfiles

armcknight:
> Why is installing carthage again, if it is already installed an issue worth bringing in two Brewfiles?

Eventually the version available on one of the macOS runners will lag behind what's available via homebrew, and it'll update it on the CI machine, which will take a long time

philprime (approved):
Okay, looks good to me.

philipphofmann (approved):
That makes sense. Thanks, LGTM

PR #4809, opened 2025-02-06
tests: fixed execution and compilation of SentrySwiftUI tests

philprime:
The CI is not running tests for scheme `SentrySwiftUI`.

#skip-changelog

PR #4817, opened 2025-02-08
ref(ci): rename xcode-test.sh to sentry-xcodebuild.sh

armcknight:
Renaming this file in it's own PR so that changes I make to the file later diff more cleanly in #4818 

#skip-changelog


philprime (approved):
LGTM, we can merge this before #4808 because that one is still blocked by #4809

PR #4818, opened 2025-02-08
ref(ci): use sentry-xcodebuild.sh to wrap more xcodebuild usages

armcknight:
We can simplify some places where there's a long complicated xcodebuild invocation, and instead place that in the existing wrapper script we have for it, where it can benefit from some other tricks we've learned over time.

There's still a few clunky invocations of the wrapper script itself that should improve after #4808

I stared at the one in the fastfile for a while but i'm not immediately sure what to do with it. I kinda feel like that step just shouldn't be done with fastlane, but i'll save that for another PR.

#skip-changelog


scripts/sentry-xcodebuild.sh
```diff
@@ -57,63 +63,122 @@ fi
...
+    RUN_ANALYZE=false
     RUN_BUILD=false
     RUN_BUILD_FOR_TESTING=true
     RUN_TEST_WITHOUT_BUILDING=true
+    RUN_TEST_WITH_TSAN=false
     ;;
 esac
 
-if [ "$CI" == true ]; then
+if [ "${CI-false}" == true ]; then
```
philprime, line 109: `l`: why the fallback to false? isn't `"" == true` also falsy?

scripts/sentry-xcodebuild.sh
```diff
@@ -57,63 +63,122 @@ fi
...
     RUBY_ENV_ARGS=""
 else
     RUBY_ENV_ARGS="rbenv exec bundle exec"
 fi
 
 if [ $RUN_BUILD == true ]; then
-    set -o pipefail && env NSUnbufferedIO=YES xcodebuild \
+    set -o pipefail && env NSUnbufferedIO=YES CODE_SIGN_IDENTITY=\"\" CODE_SIGNING_REQUIRED=NO xcodebuild \
```
philprime, line 116: I am not sure if we can add these env values here, as we are also building signed versions of the sample apps.

scripts/sentry-xcodebuild.sh
```diff
@@ -57,63 +63,122 @@ fi
...
-        test-without-building \
-        | tee raw-test-output.log \
-        | $RUBY_ENV_ARGS xcpretty -t \
-        && slather coverage --configuration "$CONFIGURATION"
+        -skip-testing:"SentryProfilerTests" \
+        test \
+        | tee thread-sanitizer.log \
+        | xcpretty -t
+
+    testStatus=$?
```
philprime, line 155: `m`: Correct me if I am wrong, but I think this will not be executed because the entire script will exit if the xcodebuild command fails

scripts/sentry-xcodebuild.sh
armcknight: the script fails here if you run it locally if your machine doesn't define the `CI` environment variable at all

scripts/sentry-xcodebuild.sh
armcknight: IDK, i moved this from https://github.com/getsentry/sentry-cocoa/pull/4818/files#diff-b3aa5b62d9aab3e854e3fe47cc6aaacd2e4bd86fd6bb7484fb51d698623bf6a6L8-L19

scripts/sentry-xcodebuild.sh
philprime: Why does it fail when a variable is not set? shouldn't it just evaluate to an empty string?

armcknight: hm, after trying another minimal example it does. i was getting an unbound variable error at some point. i'll fix this.

scripts/sentry-xcodebuild.sh
armcknight: Could it be that it used to work, but then the addition of `&& set -o pipefail` at the beginning of the pipeline would cause it to no longer propagate the `$?` the same way? we also already had that at the top of the script, but i think that wouldn't have applied to the pipeline, which is why it was added there?

armcknight:
honestly every time i make an edit to sentry-xcodebuild.sh at this point, it feels wrong to continue with it as a bash script. rewriting it in swift would be a huge improvement.

armcknight:
This feels low priority and i haven't touched it in a while. It's working so I'm just going to leave it for now.

PR #4819, opened 2025-02-08
chore(ci): update ruby version to match github actions

armcknight:
due to https://github.com/actions/runner-images/issues/11345

I also tried adding it to the tooling check precommit hook script, lmk what you think of that.

#skip-changelog


philprime (approved):
LGTM, this way we also get a full list of missing tools

philipphofmann (approved):
LGTM, thanks.

PR #4820, opened 2025-02-08
ref: use a json file to define the large unit test matrix

armcknight:
I got tired of scrolling past a huge test matrix definition 🙃 

#skip-changelog

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -45,7 +45,7 @@ jobs:
           path: |
             ~/Library/Logs/scan/*.log
             ./fastlane/test_output/**
-
+            
   # SwiftUI only supports iOS 14+ so we run it in a separate matrix here
   ui-tests-swift-ui:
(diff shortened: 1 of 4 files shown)


.github/workflows/ui-tests.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           path: |
             ~/Library/Logs/scan/*.log
             ./fastlane/test_output/**
-
+
```
armcknight, line 48:
```suggestion

```

philprime:
While I see your point in the large test matix, move it to a JSON file breaks local validation and code completion with the [VSCode/Cursor extension for GitHub Actions](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions).

It's not a deal breaker, but this adds complexity in maintenance just for readability.

philipphofmann:
I agree with @philprime. Furthermore, this PR removes the code comments in the matrix that provide helpful information. I never see the test matrix as a problem. I would like to stick to the normal exiting approach.

PR #4823, opened 2025-02-10
chore(ci): upgrade Xcode Runner image version to macOS-15

philprime:
Changes all the runner base images to macOS-15 (where possible without dependency/simulator constraints)

#skip-changelog


philprime:
@philipphofmann this is just a draft PR for me to understand where we could update to newer runner versions and where not. I am not expecting this to be merged like this.

PR #4807, opened 2025-02-06
chore(ci): add multiple segmented unit test schemes

philprime:
- Splits unit tests into multiple schemes to run more isolated tests.
- I added the jobs in addition to existing tests, so we can compare the stability.
- Migrates the schemes to use test plans for more control
- I added a test plan "Base.xctestplan" where all new tests are added automatically (basically a catch-all), configured as `skippedTests`.
- For each segment I created an additional test plan which does **not** add new tests automatically, configured as `selectedTests`.

**Further details in this comment:**

https://github.com/getsentry/sentry-cocoa/pull/4825#issuecomment-2650263237

Todo:
- [ ] Reduce build-time using build once for testing, use for multiple tests

#skip-changelog

PR #4825, opened 2025-02-10
chore(ci): add multiple segmented ui test schemes

philprime:
- Splits ui tests into multiple schemes to run more isolated tests.
- I added the jobs in addition to existing tests, so we can compare the stability.
- Migrates the schemes to use test plans for more control
- I added a test plan `Base.xctestplan` where all new tests are added automatically (basically a catch-all), configured as `skippedTests`.
- For each segment I created an additional test plan which does **not** add new tests automatically, configured as `selectedTests`.

Todo:
- [ ] Reduce build-time using build once for testing, use for multiple tests

#skip-changelog

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -6,14 +6,13 @@ on:
 
   pull_request:
     paths:
-      - 'Sources/**'
-      - 'Tests/**'
-      - '.github/workflows/ui-tests.yml'
-      - 'fastlane/**'
-      - '.sauce/config.yml'
-      - 'scripts/ci-select-xcode.sh'
-      - 'Samples/iOS-Swift/**'
-
+      - "Sources/**"
+      - "Tests/**"
+      - ".github/workflows/ui-tests.yml"
+      - "fastlane/**"
+      - ".sauce/config.yml"
+      - "scripts/ci-select-xcode.sh"
+      - "Samples/iOS-Swift/**"
 
 # https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value
 concurrency:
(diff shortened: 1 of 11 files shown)


armcknight:
Is it possible for you to separate the changes into different PRs, one that creates differentiated schemes, and another that switches them to using test plans?

In that PR, I'd be interested to know more about what these are for:
> I added a test plan Base.xctestplan where all new tests are added automatically (basically a catch-all), configured as skippedTests.
> For each segment I created an additional test plan which does not add new tests automatically, configured as selectedTests.

philprime:
@armcknight this PR is marked as "draft" on purpose, as it is mostly intended to isolate flaky test suites. It might not be necessary to merge it into main, the value of migrating to test plans is not clear yet.

To explain these changes:

> - I added a test plan Base.xctestplan where all new tests are added automatically (basically a catch-all), configured as skippedTests.

If we add new test cases we could easily forget to add them to a test target, leaving them in the project but not executed in CI. `xcscheme` files define which tests should be skipped, therefore automatically testing every new test case added. This test plan is therefore working in the same way. 

> - For each segment I created an additional test plan which does not add new tests automatically, configured as selectedTests.

To reduce duplicate execution of testing, we manually need to skip a test in the base test plan and select in the other test plan. We could add a test case to both, having a "all" test plan, and one to partially test. But, as I believe we are having flaky CI due to broken test isolation (i.e. unit test `foo` affects unit test `bar` even though they are unrelated), the goal of the segments is mostly exploring how flaky tests are when run in segments.

armcknight:
@philprime understood that it's a draft, i noticed that, i just wanted to comment on the strategy only and get some understanding of the motivation. there was no mention of flaky test isolation, so that's good to know now. 

> unit test foo affects unit test bar even though they are unrelated

this kind of problem has plagued us for a long time. that's why you'll notice there is a separate target for profiling tests. they were long blamed for taking down the general test suite, and finally i had enough and separated them. then we went on later to find the true culprits.

even if you don't merge in a change to test plans, i would be very interested to see us move towards not only differentiated test schemes, but targets to back those up, for this very reason. reducing blast radius will get us to fixes faster.

philprime:
@armcknight thanks for the comment. I am about to spin up some additional PRs derived from this one and #4807 and will give you more details on the next steps there.

philprime:
Closing this because experiment is done.
Will reference this in future PRs applying findings.
See #4848.

PR #4826, opened 2025-02-10
impr(profiling): increase buffer length from 10s to 60s

armcknight:
We need to increase the continuous profiling client buffer size to 60 seconds (from https://www.notion.so/sentry/Continuous-UI-Profiling-SDK-API-Spec-17e8b10e4b5d80c59a40c6e114470934?pvs=4)


Sources/Sentry/Profiling/SentryProfilerDefines.h
```diff
@@ -15,7 +15,7 @@ typedef NS_ENUM(NSUInteger, SentryProfilerTruncationReason) {
     SentryProfilerTruncationReasonAppMovedToBackground,
 };
 
-static NSTimeInterval kSentryProfilerChunkExpirationInterval = 10;
+static NSTimeInterval kSentryProfilerChunkExpirationInterval = 60;
```
philipphofmann, line 18: `m`: It would be great to have a test failing if we change this by mistake. Then, you need to mess up the code in two different places.

philipphofmann (approved):
LGTM

SentryTestUtils/TestSentryNSTimerFactory.swift
```diff
@@ -1,52 +1,92 @@
...
 
 // We must not subclass NSTimer, see https://developer.apple.com/documentation/foundation/nstimer#1770465.
 // Therefore we return a NSTimer instance here with TimeInterval.infinity.
 public class TestSentryNSTimerFactory: SentryNSTimerFactory {
     public struct Overrides {
-    	private var _timer: Timer?
-    
-    	public var timer: Timer {
-        	get {
-         	   _timer ?? Timer()
```
armcknight, line 12: The fact that this returned a new timer if one had never been initialized I think was causing false negatives where tests should've been failing.

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
```diff
@@ -174,13 +195,13 @@ private extension SentryContinuousProfilerTests {
...
         XCTAssert(SentryContinuousProfiler.isCurrentlyProfiling())
         SentryContinuousProfiler.stop()
-        assertContinuousProfileStoppage()
+        try assertContinuousProfileStoppage()
     }
     
-    func assertContinuousProfileStoppage() {
+    func assertContinuousProfileStoppage() throws {
         XCTAssert(SentryContinuousProfiler.isCurrentlyProfiling())
-        fixture.currentDateProvider.advance(by: kSentryProfilerTimeoutInterval)
```
armcknight, line 182: Oof, his should've been SentryProfilerChunkExpirationInterval all along. But, I'm now thinking we shouldn't even reuse the constant definitions in the test cases. I don't know... I guess we can still use them and we're just testing all the integration in the middle is hooked up correctly. Maybe I'll switch back from the magic numbers in the tests.

SentryTestUtils/TestSentryNSTimerFactory.swift
```diff
@@ -1,52 +1,92 @@
...
+    }
+    
+    // check the current time against the last fire time and interval, and if enough time has elapsed, execute any block/invocation registered with the timer
+    func check() throws {
+        guard let overrides = overrides else { throw TestTimerError.timerNotInitialized }
+        let currentDate = currentDateProvider.date()
+        if currentDate.timeIntervalSince(overrides.lastFireDate) >= overrides.interval {
+            try fire()
+        }
+    }
```
armcknight, line 80: This function is a stronger assertion than just firing the timer block directly. Now, we're actually testing that the timer fires after the correct interval. I think this lead to some false confidence in tests that would call `fire` and sort of throw away the context of how much time has elapsed or what should happen. See comments in SentryTracerTests below.

Tests/SentryTests/Transaction/SentryTracerTests.swift
```diff
@@ -52,8 +52,6 @@ class SentryTracerTests: XCTestCase {
 #endif // os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
         
         init() {
-            dispatchQueue.blockBeforeMainBlock = { false }
```
armcknight, line 55: I don't know why this was configured this way. It essentially led to test cases never actually initializing timeout timers, so calling TestSentryNSTimerFactory.fire() was a no-op.

Tests/SentryTests/Transaction/SentryTracerTests.swift
```diff
@@ -378,27 +376,27 @@ class SentryTracerTests: XCTestCase {
...
 
-        fixture.timerFactory.fire()
+        try fixture.timerFactory.fire()
 
-        XCTAssertEqual(sut.status, .undefined)
-        XCTAssertEqual(child1.status, .undefined)
-        XCTAssertEqual(child2.status, .undefined)
+        XCTAssertEqual(sut.status, .deadlineExceeded)
+        XCTAssertEqual(child1.status, .deadlineExceeded)
+        XCTAssertEqual(child2.status, .deadlineExceeded)
```
armcknight, line 391: So I am unsure of what is going on here, or what the correct result should be. Previously, the call to fire was a no-op, see my above comment on the dispatch queue wrapper configuration. Now that the deadline timer is actually configured, this is the new result. I traced execution from the call to child3.finish() down into SentryTracer.canBeFinished, and it winds up taking the branch here: https://github.com/getsentry/sentry-cocoa/blob/44332bfdab0ae43dbe4edd86cd2d72de4fab4507/Sources/Sentry/SentryTracer.m#L506-L512, so the remaining spans are still in flight when the deadline timer is fired on the next line of the test.

armcknight (commented):
@philipphofmann I wound up needing to perform a substantial refactoring of TestSentryNSTimerFactory to add a new function, which led to finding some weirdness in other test cases, and iterating through this a few times. Please check out my comments below.

Tests/SentryTests/Transaction/SentryTracerTests.swift
philipphofmann: Good catch, I pushed a commit to address this: https://github.com/getsentry/sentry-cocoa/pull/4826/commits/e36a8ce8a3c052772566884d9e8bc48b02c1214a.

Tests/SentryTests/Transaction/SentryTracerTests.swift
philipphofmann: I think that was a hacky workaround for skipping initializing the timer, we can remove it safely. I also remove the other invocations with: https://github.com/getsentry/sentry-cocoa/pull/4826/commits/3ab75ee75a65c6810ccb94e2f3a82a2c72df4c6b

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
philipphofmann: Yes, exactly. I prefer not using constants in tests because I prefer having to the info duplicate. So I need to change it in two different places. For important constants a test like this is sometimes great:
```
func testSentryProfilerChunkExpirationInterval() {
    XCTAssertEqual(60, kSentryProfilerChunkExpirationInterval)
}
```

philipphofmann (approved):
Thanks for the tests, LGTM

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
armcknight: That's a good idea, I added both that one, plus in SentryTraceProfilerTests:
```
func testSentryProfilerChunkExpirationInterval() {
    XCTAssertEqual(30, kSentryProfilerTimeoutInterval)
}
```

PR #4828, opened 2025-02-11
impr(ci): xcbeautify junit reports for unit tests

armcknight:
If we want to stick with xcbeautify, we need to find a way to make the test results more readable. This uses the `-reports junit` option, and then tries out a few different github marketplace actions that consume a JUnit XML file and prepare reports, add annotations to the PR diff, etc for a failing test.

TODO
~- [ ] get the UI tests generating JUnit XML files as well, unsure if Fastlane's xcbeautify integration supports this, just try it out and see~ will do in another pr, don't want to hold this one up: https://github.com/getsentry/sentry-cocoa/pull/4847

#skip-changelog


armcknight:
Here is my comparison of the three junit parsers i tried. TL;DR I prefer `mikepenz/action-junit-report` so am leaving this PR set to that one.

# test-summary/action

https://github.com/getsentry/sentry-cocoa/actions/runs/13319710405

Simply makes a nicer summary on the main test workflow that lists all the actual test cases that failed. IMO this is better than looking through the annotations we currently have. Those annotations are still there, in case there's other important info we want from them.

image

# mikepenz/action-junit-report

https://github.com/getsentry/sentry-cocoa/actions/runs/13319702196

shows a summary including which tests failed right on the workflow summary page: 
image
 
and you can also see them if you expand the toggle in the job's log in the parser's step: 
image


# dorny/test-reporter

https://github.com/getsentry/sentry-cocoa/actions/runs/13319706591

produces a separate page to view test results in the github actions checks GUI, accessible from a link it places in the test run logs, so requires an additional navigation to get to it. Doesn't bubble failures to the top so still requires scrolling through lots of test names to find the failures. Doesn't show the name of the test case that failed, only the suite.

image
image
image

OK, it does show the failing test case, all the way at the bottom: 

image

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -5,6 +5,10 @@ import XCTest
 // swiftlint:disable file_length
 class SentrySDKTests: XCTestCase {
     
+    func testFailingTest() {
+        XCTFail("Test a failure")
+    }
+
```
armcknight, line 11: remove experiment

```suggestion
```

philipphofmann (approved):
Thanks @armcknight. Let's give this a shot. It looks great at first glance. If we don't like it, we iterate.

Issue #2216, opened 2022-09-23
Measure how long an app hang lasts


philipphofmann:
### Description

The SDK doesn't track the duration of an app hang. The ANRTrackingIntegration immediately reports an app hang when it occurs. To get the exact duration, the SDK would need to write the app hang to disk first because the watchdog could kill the app anytime when it is blocking the main thread. When the ANR stops, which we detect here
https://github.com/getsentry/sentry-cocoa/blob/b86953665ea253dabe1dced60d438f1a11f8934f/Sources/Sentry/SentryANRTrackingIntegration.m#L83-L86
the SDK could calculate the exact duration. When the watchdog kills the process cause of the app hang, the SDK would need to pick up the stored error and send it.

See also information in duplicate issue: https://github.com/getsentry/sentry-cocoa/issues/3556.

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4822
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4829
```

brustolin:
I think we can't calculate the exact duration if the OS kills the app, but then we could report crash because of app hangs, which is also very nice.

To improve the calculation of duration we can look into the last frame from the frame tracker to now when the hang began.

philipphofmann:
Yes, we can't calculate it when the OS kills the app, but we can calculate it if it doesn't.

philipphofmann:
Maybe we could also detect if the user terminates the app because it kept hanging, and report this as a special type of app hang.

PR #4829, opened 2025-02-11
internal: Measure app hang duration

philipphofmann:
Measure the app hang duration for ANRTrackerV2 and report it via the delegate.anrStopped, so we can set the app hang duration in the ANRTrackingIntegration in a future PR.

Required for https://github.com/getsentry/sentry-cocoa/issues/2216.

#skip-changelog


philipphofmann:
Converted this back to a draft, to change the errorMessage to min and max duration for the delegate.

philprime (approved):
LGTM

PR #4837, opened 2025-02-12
docs: apply formatting to markdown files

philprime:
Formatted Markdown documents using prettier to remove unused whitespaces and apply best-practices for character escaping (i.e. `\_` instead of `_` because it is used to format text in `_text_`)

#skip-changelog


philipphofmann (approved):
LGTM, it would be great to use a linter or something like that in the precommit hooks and/or CI to ensure this stays like that. But please in a follow up PR, if you decide to do this.

PR #4846, opened 2025-02-13
chore: add prettier to pre-commit for markdown and json files

philprime:
- Adds linting with prettier to pre-commit
- Adds install, linting and formatting to Makefile
- Adds formatting with prettier to CI
- Adds linting with prettier to CI
- Adds prettier to the recommended VSCode extensions

#skip-changelog


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 
 - Deserializing SentryEvents with Decodable (#4724)
 - Remove internal unknown dict for Breadcrumbs (#4803) This potentially only impacts hybrid SDKs.
+- Add prettier to pre-commit for markdown and json files (#4846)
```
armcknight, line 30: I don't think belongs in the changelog. If anything, write about it in CONTRIBUTING.md or develop-docs/README.md

armcknight (commented):
I'm not seeing the benefit of this. Did you run json and md linting over the whole codebase already? Because the only thing I see changed are asset catalogs in the sample apps. It's a heavyweight change bringing in NPM and doesn't appear to have made any material improvements.

philprime:
@armcknight Yes, I already fixed most of the files in #4837, where you can see that there were quite a lot of formatting errors like indents, double whitespaces etc. I am working with VSCode/Cursor using Prettier as the default formatter. As soon as I save the CHANGELOG.md, it keeps formatting. Of course this is behavior on my machine, and I can disable it, but simply due to the fact that the default configuration of the prettier formatter is applying changes is indicator enough that there's room for improvement.

This PR is a follow-up based on the proposition of @philipphofmann in [his comment](https://github.com/getsentry/sentry-cocoa/pull/4837#pullrequestreview-2611442998).

I do agree with you that NPM is heavy weight, but AFAIK there's no other way to get prettier enabled.
We're already using prettier in sentry-docs, so it's not like this introduces a completely new tool:

https://github.com/getsentry/sentry-docs/blob/master/.pre-commit-config.yaml#L26-L32

Regarding only changing the `Content.json`: I configured prettier to not run on the assets we use in unit tests (see `.prettierignore`), if we remove that - except for the garbage JSON files which are broken on purpose - there are tons of additional files.

We can introduce it as a CI-only dependency or add an additional "Sentry-Contributor" init command to keep it more light weight.

philprime:
On a side-note, with the introduction of test plans in #4848, we are adding more JSON documents over time.

armcknight:
Ah I see, thanks for linking me to that PR. I don't feel strongly about not bringing in npm, and since it's preinstalled in CI, that shouldn't increase CI times significantly.

Although I would say, I would prefer to exclude anything generated by Xcode, like the manifests for asset catalogs or test plans. The test resources you currently have excluded, I think should be linted, since those are hand edited and maintained by us. And like you said, we can exclude the ones that are supposed to be broken in some way (maybe those can be put in a subdirectory to make that easier).

CHANGELOG.md
philprime:
```suggestion
```

philprime:
I'll change the PR to format test resources, and exclude the garbage ones 👍 

Regarding the Xcode files, I don't fully understand why we should exclude them. Even though they are generated by Xcode, I believe we only need to format them once and they are good to go. Do I miss something here?

philprime:
We could split the PR in one configuring everything and one to fix issues. Might make it easier to look into it in the future

philipphofmann:
> We could split the PR in one configuring everything and one to fix issues. Might make it easier to look into it in the future

If you think that's a chicken-egg problem and not worth the effort. If you set up everything, you must disable most of it, only to enable it again in another PR. I'm fine with doing all in one PR, unless you disagree.

PR #4848, opened 2025-02-14
chore: add Xcode test plans

philprime:
This PR is based on findings in #4807 and #4825.

- Converts existing test configuration to [Xcode Test Plans](https://developer.apple.com/documentation/xcode/organizing-tests-to-improve-feedback)
- Does not add any test segmentation or custom plans.
- All plans are configured as "Automatically include new tests", by only defining a `skippedTests` list where applicable.
- I added the suffix `_Base` to all of the test plans, to indicate that this is the default test plan. This is not relevant right now, but I will also create further PRs shortly, to introduce multiple test plans. See the note below.
- Test plan selection is not added to GitHub Action workflows yet, because in this PR it would be redundant.

**Note on test plans:**

To make sure that all new added test cases are executed locally and in CI, we must include base test plans with the "catch-all" approach. Each test plan must be added to an Xcode Scheme, but only one plan can be the default. When using `xcodebuild test` it will execute the default plan only, therefore we need to run tests steps/jobs multiple times with different test plans as the `-testPlan` argument.

#skip-changelog

cc @armcknight I hope this explanation gives enough insights.


Plans/visionOS-Swift_Base.xctestplan
```diff
@@ -0,0 +1,18 @@
+{
+  "configurations" : [
```
philipphofmann, line 2: `m`: Do we really need a testplan for visionOS and watchOS. We could ditch these if possible, because we don't run tests on these platforms.

philipphofmann (approved):
> Note on test plans:

> To make sure that all new added test cases are executed locally and in CI, we must include base test plans with the "catch-all" approach. Each test plan must be added to an Xcode Scheme, but only one plan can be the default. When using xcodebuild test it will execute the default plan only, therefore we need to run tests steps/jobs multiple times with different test plans as the -testPlan argument.

Maybe it makes sense to add this info to our internal develop-docs: https://github.com/getsentry/sentry-cocoa/tree/main/develop-docs#unit-tests-with-thread-sanitizer

Thanks for doing this.

Plans/Sentry_Base.xctestplan
```diff
@@ -0,0 +1,68 @@
...
+      "target" : {
+        "containerPath" : "container:Sentry.xcodeproj",
+        "identifier" : "63AA76641EB8CB2F00D153DE",
+        "name" : "SentryTests"
+      }
+    },
+    {
+      "skippedTests" : [
+        "SentryProfilerTests\/testProfilerMutationDuringSlicing"
+      ],
```
armcknight, line 59: Can we move this to SentryProfilerTests_base.xctestplan?

armcknight (changes requested):
A couple general points with this:
- I checked out the branch and the test experience locally seems good (minus your typical Xcode GUI trash 😭) 👍🏻 
- please write up the findings from the previous PRs that explain the motivation for using test plans. personally i haven't used them, so i just am ignorant of the cost/benefit tradeoff. for this i would expect to see some problem we have that this solves
    - as a subpoint to this, i still don't really understand the base test thing, or the problem it is solving. did we have tests that weren't running? or is there a danger of creating tests that don't run in a test plan, and this is an inherent problem with test plans? (i bet this becomes simpler if we move to basing our project off of a Package.swift or XcodeGen spec, but that's a discussion for another day)
- some of the test plans have  `"defaultOptions" : { "codeCoverage" : false }`, can we leave code coverage generation enabled? i don't see anywhere that's changed for CI runs, but I also use it locally
- the only reason i am submitting this as Requests Changes is that it was already approved by @philipphofmann, but .github/workflows/test.yml needs to be updated so that any diffs to a xctestplan file triggers a test run, so we can see this work in CI (arguably, changes to the xcschemes should also trigger this)
    - on the subject, can we move `//Plans/` to `//Tests/Configuration/Plans/`?

philprime:
@philipphofmann I will add documentation to the develop-docs in this PR

@armcknight I'll try to explain your points in detail again and also try to cover them in the develop docs:


- i would expect to see some problem we have that this solves as a subpoint to this

The main reason why I am introducing this changes is that it allows us to dig deeper into the issue of flaky CI. Without creating more targets or schemes, we can add multiple test plans and configure different sets of tests to be run in each one. 

In addition test plans allow custom features which can be helpful for us, such as automatic repetitions or built-in retry on failures. Maybe [this WWDC21 video](https://developer.apple.com/videos/play/wwdc2021/10296/) can provide further insights.

In addition we should further explore the test plan settings such as `Automatic Screen Capture` which might help us to drill down the issues in the flaky UI tests.

Screenshot 2025-02-19 at 10 16 50

But I do not see a point in explaining the potential in test plans in the develop docs.

- i still don't really understand the base test thing, or the problem it is solving. did we have tests that weren't running? or is there a danger of creating tests that don't run in a test plan, and this is an inherent problem with test plans? 

The main problem is not working cleanly. When developing tests locally you might run them differently than they are run in CI as we have multiple targets and schemes. We even didn't run any SwiftUI tests in CI for a while until I fixed it in #4809. Therefore the whole point of the base plan to catch all tests is 1) convenience of new test files being executed automatically and 2) to make sure we can not forget to execute them.

- (i bet this becomes simpler if we move to basing our project off of a Package.swift or XcodeGen spec, but that's a discussion for another day)

Yes, this is a different topic and out-of-scope for this.

- some of the test plans have "defaultOptions" : { "codeCoverage" : false }, can we leave code coverage generation enabled? i don't see anywhere that's changed for CI runs, but I also use it locally

I am not 100% certain which tests should collect coverage, so I tried to reuse the settings of the xcschemes as much as possible. Here I need yours or @philipphofmann guidance on which code coverage is relevant to us.

- the only reason i am submitting this as Requests Changes is that it was already approved by @philipphofmann 

You could still have left a comment with `h` instead of blocking the PR, because I do respect the notation and now the PR needs to stay open until you have reviewed it again (which is valuable, but timezones delays our process).

- but .github/workflows/test.yml needs to be updated so that any diffs to a xctestplan file triggers a test run, so we can see this work in CI (arguably, changes to the xcschemes should also trigger this, but we never did that, not sure if we ever considered it or why we might've declined)

That is a good catch, I can add that to the workflow file. Thanks for pointing it out. Regarding the xcschemes, I do not know why it wasn't considered before, that's from before I started working on this SDK. I'll do it in another PR.

- on the subject, can we move //Plans/ to //Tests/Configuration/Plans/?

I decided against placing it there because the test plans are used by samples and SDKs, and should therefore be at the nearest shared level which is the repository root. If you insist I will move it, but it is a lot of work to fix all the paths as they are double-linked from plans and schemes.

philprime:
I just figured out how to view the UI test replays.
All you have to do is go into the test results, double click on the test and you can see the UI test interaction history, as well as a playback of the screen.

Screenshot 2025-02-19 at 15 22 05
Screenshot 2025-02-19 at 15 18 06

Plans/Sentry_Base.xctestplan
philprime: We could probably remove the entire SentryProfilerTests from the `Sentry_base.xctestplan`. This base plan is based on `Sentry.xcodeproj/xcshareddata/xcschemes/Sentry.xcscheme`, so we should question why it was there in the first place

armcknight (approved):
Thanks for getting the test workflow triggers sorted!

Issue #4852, opened 2025-02-14
profiling: soft-deprecate trace profiling API


armcknight:
### Description

We need to deprecate the following APIs for eventual removal in a future major revision of the SDK:

- SentryOptions.profilesSampleRate
- SentryOptions.profilesSampler
- SentryOptions.enableAppLaunchProfiling

philprime:
Should we mention this in #2946?

philipphofmann:
Yes we can and/or create issues with the breaking label and add it to the milestone 9.0.0.

armcknight:
Done @philipphofmann @philprime , thanks for reminding me https://github.com/getsentry/sentry-cocoa/discussions/2946#discussioncomment-12543022

Issue #4851, opened 2025-02-14
profiling: updates to continuous profiling API


armcknight:
### Description

We have some API additions/changes planned before shipping continuous profiling as a GA feature, discussed at https://www.notion.so/sentry/Continuous-UI-Profiling-SDK-API-Spec-17e8b10e4b5d80c59a40c6e114470934?pvs=4

PR #4854, opened 2025-02-14
change(profiling): deprecate old trace-and continuous profiling API

armcknight:
For #4852. Deprecate all old API. We'll be adding new API that improve the way continuous profiling works in future PRs (see the epic issue #4851).

#skip-changelog


philipphofmann:
> Because there is [no way to ignore deprecation warnings in Swift](https://forums.swift.org/t/suppressing-deprecated-warnings/53970/3), and we have this Swift implementation that references one of the things I need to deprecate here, I am now faced with a decision to make on what kind of hack to implement to work around this limitation to resolve it, because it's breaking basically every single CI job we run: image
> 
> ETA: i'm temporarily disabling warnings as errors until i can get a workaround in place, so it doesn't slow down further work on the rest of the tasks.

Yeah, that sucks pretty much. You can mark the whole method as deprecated like this:

https://github.com/getsentry/sentry-cocoa/blob/a286774f1f994d1d78f7f959e17dd7fc8b10880a/Sources/Swift/Protocol/Codable/SentryUserCodable.swift#L17-L25

I'm starting to believe that we should disable deprecation warnings or disable change warnings as errors and set every warning to an error level that should be an error.

Sources/Configuration/SDK.xcconfig
```diff
@@ -163,7 +161,8 @@ SWIFT_OPTIMIZATION_LEVEL_Test = -Onone
 SWIFT_OPTIMIZATION_LEVEL_TestCI = -Onone
 SWIFT_OPTIMIZATION_LEVEL = $(SWIFT_OPTIMIZATION_LEVEL_$(CONFIGURATION))
 
-SWIFT_TREAT_WARNINGS_AS_ERRORS = YES
+GCC_TREAT_WARNINGS_AS_ERRORS = NO
+SWIFT_TREAT_WARNINGS_AS_ERRORS = NO
```
philipphofmann, line 165: `h`: Ah 😡 , this is again a point where Swift sucks pretty much compared to ObjC. I didn't find a way to only treat deprecations as warnings or properly ignore deprecation warnings. I thought it's possible but I think it's not. The problem is that we open the door to ignore new compiler warnings, which could cause bugs. I enabled treat warnings as errors some time ago because a warning led to a severe bug, but I can't recall which one. 

If we disable warnings as errors, we need to ensure we're not introducing new warnings by accident. We could have an allowed warning list in a file and then compare it with the Xcode output or something. This should run in CI and fail if there's a new warning that's not on that list, in my opinion. We shouldn't do this in this PR thought. Maybe using the weird deprecated workaround for once more is acdeptable.

armcknight: yes, i'll try the deprecation workaround you showed me. this change turning off warnings as errors isn't meant to be merged.

armcknight: Unfortunately I don't think it's going to be possible to work around the error I pointed out in https://github.com/getsentry/sentry-cocoa/pull/4854#issuecomment-2660347367, because we can't deprecate `SentryEnabledFeaturesBuilder.getEnabledFeatures` :/ 

Sounds like we need a release-time process to evaluate all compiler errors produced to decide if the release is fine to deploy, or if warnings need to be resolved. Even here, only setting treating warnings as errors for the release config won't work, because the build will simply fail, but that's not really what we want, because that would hide other potential things we want to know about while preparing a release, and we don't want to have a death-by-a-thousand-cuts scenario.

Does `craft` have any way to track warnings from each release, and produce a diff of warnings from the last released version and the current release being prepared, to more easily see new ones that have been introduced?

philipphofmann: >Does craft have any way to track warnings from each release, and produce a diff of warnings from the last released version and the current release being prepared, to more easily see new ones that have been introduced?

I would be surprised if it does.

armcknight: yep we can do that 👍🏻

armcknight: Per my last comment in https://github.com/getsentry/sentry-cocoa/pull/4946, I'm simply going to remove the deprecation annotations for now and only soft deprecate these API using docs for now.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Deprecations
+
+- Some profiling API are deprecated in favor of new ways to manage starting and stopping continuous profiling session (#4854)
```
```suggestion
- Some profiling API are deprecated in favor of new ways to manage starting and stopping continuous profiling sessions (#4854)
```

philipphofmann (approved):
LGTM

Issue #4853, opened 2025-02-14
profiling: add new profiling API


armcknight:
### Description

Top level topic branch: https://github.com/getsentry/sentry-cocoa/pull/4952

Deprecating old api: https://github.com/getsentry/sentry-cocoa/pull/4854

These API will supersede the current continuous profiling API and build on it for more fine-grained control of the functionality:

- [x] `start_profile_session: function` https://github.com/getsentry/sentry-cocoa/pull/4855 Starts a profiling session. This function is a no-op if *one of* the following conditions are met:
    - The profiling session is sampled with respect `profile_session_sample_rate` and the profiler is already running
    - The profiling session is **not sampled**
    - `profile_lifecycle` is set to `trace` — in this scenario, the profiler is automatically started and stopped depending on whether there is an active sampled span (see docs below) so it is not permitted to manually start profiling
    - SDKs should log a warning when in `debug` mode if someone calls this API and it no-ops
- [x] `stop_profile_session: function` https://github.com/getsentry/sentry-cocoa/pull/4855 Stops a profiling session. The behavior of this function depends on the setting of `profile_lifecycle` (see docs below)
    - If `profile_lifecycle` is `manual`: Stops the current profiling session. Profiling can be started again by calling `start_profile_session`
    - If `profile_lifecycle` is `trace`: No-op. SDKs should log a warning when in `debug` mode if someone calls this API and it no-ops
- [x] `profile_session_sample_rate` https://github.com/getsentry/sentry-cocoa/pull/4979 **(default: `0`)**: Determines the sampling rate for the entire profiling session. Whether or not the session is sampled is determined once initially when the SDK is configured. The profiling session is defined as follows, depending on the type of profiling:
    - **UI Profiling:** the session starts when a new *user* session begins, and stops when the user session ends
        - `profile_session_sample_rate` can be used to set what % of user sessions have profiling enabled
        - **On mobile:** backgrounding and foregrounding the app starts a new user session and **sampling is re-evaluated**. If there is no active trace when the app is backgrounded, profiling should be stopped before the app backgrounds. If there is an active trace and profiling is in-flight when the app is foregrounded again, the same profiling session should continue until the last root span in that trace finishes — this means that the re-evaluated sample rate does not actually take effect until the profiler is started again.
[...]

PR #4855, opened 2025-02-15
feat(profiling): new start and stop API

armcknight:
stub the new start and stop methods we'll offer for continuous profiling. there are some TODOs for checks that come in later PR branches as those API are implemented and become available.

#skip-changelog; for #4853


Sources/Sentry/Public/SentrySDK.h
```diff
@@ -382,6 +382,19 @@ SENTRY_NO_INIT
...
 
+/**
+ * Start a new continuous profile session.
+ * @note Does nothing if the profiling session is sampled with respect
+ * @c SentryOptions.profileSessionSampleRate and the profiler is already running.
+ * @note Does nothing if the profile session is not sampled.
+ * @note Does nothing if @c SentryOptions.profileLifecycle is set to @c trace . In this scenario,
+ * the profiler is automatically started and stopped depending on whether there is an active sampled
+ * span so it is not permitted to manually start profiling.
+ * @warning Continuous profiling mode is experimental and may still contain bugs.
```
philipphofmann, line 393: `m`: Maybe we should put the warning above all the `@note` so it's clearer this is still experimental.

Sources/Sentry/SentrySDK.m
```diff
@@ -632,6 +632,25 @@ + (void)stopProfiler
...
++ (void)startProfileSession
+{
+    // TODO: log a debug message and bail if the profiling session is sampled
+    // with respect SentryOptions.profileSessionSampleRate and the profiler is already running
+
+    // TODO: log a debug message and bail if the profile session is not sampled
+
+    // TODO: log a debug message and bail if SentryOptions.profileLifecycle is set to trace
+
+    [SentryContinuousProfiler start];
```
philipphofmann, line 645: `m`: It would be great if we could write a simple test validating that this method actually starts the SentryContinuousProfiler.

Sources/Sentry/SentrySDK.m
```diff
@@ -632,6 +632,25 @@ + (void)stopProfiler
 
     [SentryContinuousProfiler stop];
 }
+
++ (void)startProfileSession
```
philipphofmann, line 636: `m`: What happens if the user didn't start the SDK yet? Should we check that and make this a no op? The same applies for stop.

philipphofmann (approved):
As this doesn't get merged to main, the TODOs are OK, LGTM

Sources/Sentry/SentrySDK.m
armcknight: It's a no-op because we try to check `[currentHub.client.options isContinuousProfilingEnabled]` but `currentHub.client.` is `nil` if the SDK is closed. I added a test for that in https://github.com/getsentry/sentry-cocoa/pull/4961

Sources/Sentry/SentrySDK.m
armcknight: If you mean specifically for the internal method `+ [SentryContinuousProfiler start]`, we do have a test for that already: https://github.com/getsentry/sentry-cocoa/blob/c964075f766455bae09be6d409d8823eb75a62ed/Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift#L31-L33 

the actual assertion is made at https://github.com/getsentry/sentry-cocoa/blob/c964075f766455bae09be6d409d8823eb75a62ed/Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift#L176-L179

armcknight: but i will add more tests for the public API entrypoint of SentrySDK.startProfileSession to exercise all the early exits there for the various checks, as i did in https://github.com/getsentry/sentry-cocoa/pull/4961

Sources/Sentry/Public/SentrySDK.h
armcknight: Done!

PR #4856, opened 2025-02-15
feat(profiling): new options class and property; lifecycle option

armcknight:
Implementing the new option for profile session lifecycle. Includes using the new swift namespacing pattern to house these new config options.

#skip-changelog; for https://github.com/getsentry/sentry-cocoa/issues/4853


philipphofmann:
Sorry @armcknight I didn't get to this today. I will review this tomorrow. It would be great if you could already address the failing unit tests .

armcknight:
Still need to check up on the UI tests, reverting this to draft status, will have a look again tomorrow

philipphofmann:
@armcknight the diff seems a bit odd. It shows changes like the refactoring form `event.isCrashEvent` to `event.isFatalEvent`, as this PR is big, it would be great if you could double-check the diff before I give this a review.

armcknight:
Sorry about that @philipphofmann , I think a bad merge happened. It looks like it's fixed now!

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -49,9 +73,46 @@
...
+ * Decrement number of root spans in flight and stop continuous profiler if there are none left.
+ * @warning Must be called from a synchronized context.
+ */
+void
+_unsafe_cleanUpContinuousTraceProfiler()
+{
+    if (SENTRY_CASSERT_RETURN(_gInFlightRootSpans > 0,
+            @"Attempted to decrement count of root spans to less than zero.")) {
+        _gInFlightRootSpans -= 1;
+    }
```
philipphofmann, line 86: `m`: I'm unsure if we shouldn't choose an approach that degrades gracefully here. If we have a bug in our logic, this would make the SDK unusable in debug builds. Wouldn't it be better to only log a error or fatal log and continue? This applies to all SENTRY_CASSERT_RETURN. As this involves span tracking and async code is involved, there is a risk for having such a bug.

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -82,27 +143,46 @@
...
+        if (hub.getClient.options.profiling.lifecycle != SentryProfileLifecycleTrace) {
+            return;
+        }
+        if (!traceSampled) {
+            return;
+        }
+        _unsafe_cleanUpContinuousTraceProfiler();
+    } else {
+        if (internalTraceId == nil) {
+            return;
```
philipphofmann, line 165: `l`: Does this maybe deserve an extra log message, as this seems like an edge case? Maybe I'm also mistaken.

philipphofmann (approved):
I added a few comments, nothing blocking though, LGTM.

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
armcknight: yeah i'm always good with more logging!

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
armcknight: My current thinking on this is that I want to know about such an unbalancing, should it pop up in unit tests. I don't think it currently is possible, and can't think of a way to write a test for it, so I left this here to point it out should it somehow occur in the future. 

I guess I look at this as an inline unit test, but I did want it to degrade gracefully for customers, however I see what you're saying about debug builds. How about I conditionally compile in an assertion only for SENTRY_TEST and SENTRY_TESTCI configs?

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -61,6 +61,13 @@
 sentry_shouldProfileNextLaunch(SentryOptions *options)
 {
     if (options.enableAppLaunchProfiling && [options isContinuousProfilingEnabled]) {
+        if (options.profiling.lifecycle == SentryProfileLifecycleTrace
+            && !SENTRY_CASSERT_RETURN(options.isTracingEnabled,
+                @"Tracing must be enabled in order to configure profiling with trace lifecycle.")) {
```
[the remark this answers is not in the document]
armcknight: You're right about this one, shouldn't be an assert.

Sources/Swift/Integrations/Performance/SentryProfileOptions.swift
```diff
@@ -0,0 +1,36 @@
...
+
+/// An object containing configuration for the Sentry profiler.
+/// - note: Profiling is automatically disabled if a thread sanitizer is attached. 
+@objcMembers
+public class SentryProfileOptions: NSObject {
+    /// Different modes for starting and stopping the profiler.
+    @objc public enum SentryProfileLifecycle: Int {
+        /// Profiling is controlled manually, and is independent of transactions & spans. Developers
+        /// must `SentrySDK.startProfileSession()` and `SentrySDK.stopProfileSession()` to control
+        /// the lifecycle of the profiler. If the session is sampled,
```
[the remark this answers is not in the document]
armcknight: I've taken this feedback to the team to discuss for a final wording, so let's push this to later docs and polish PRs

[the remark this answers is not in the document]
armcknight: however, i think we might still have an issue then with the names of the functions startProfileSession and stopProfileSession, because those don't correspond to the same concept of session i just outlined above. i'm going to punt on this as we're still discussing it, and i have docs prs that will come after this.

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
philipphofmann: You could assert the log output of your tests. So, basically, if the specific log message ever gets logged, let the test fail. But that won't fail then if some other future test ends up in this part of the code.

Using `SENTRY_CASSERT_RETURN` because we don't know how to test a specific part of the code, seems like a code smell to me and maybe we should restructure our code so we can test it. I would prefer having extracting the logic so we can test various scenarios and don't have to rely on `SENTRY_CASSERT_RETURN` to potentially reveil future problems. I can't tell you the exact solution right now. If you want I can take a closer look an open a draft PR with my some ideas, @armcknight.

Sources/Swift/Integrations/Performance/SentryProfileOptions.swift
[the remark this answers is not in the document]
armcknight: explained it with more detail in https://github.com/getsentry/sentry-cocoa/pull/4995/files#diff-e0d7fd53b3625823380c1b343b8be2b6e1febd5532e5fe1f91c66fbfd0179ed2L45-R52

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
armcknight: I see what you're saying, and I'll think a little bit more about how I can exercise the codepaths that traverse these lines. 

But there are reasonable uses for asserts. Unit tests ensure proper operation of the codepaths as they exist along with the tests. Assertions ensure proper operation of a particular part of a codepath _when the code is being changed_–you can't write all the unit tests exercising all the possible future code paths that could ever exist when this code is touched later, by anyone.

philipphofmann: Both approaches have pros and cons, as always 😃. In this specific case, I think it's OK to keep the assertions.

PR #4859, opened 2025-02-18
feat: Add experimental flag to disable swizzling of NSData

philprime:
This is a derived PR from #4605

As we want to allow users to choose between automatic instrumentation and manual file IO tracing, the newly introduced experimental flag allows disabling swizzling of NSData.

I chose to keep the current behavior of NSData swizzling being enabled by default when swizzling is enabled, and offer an option to disable it instead. This is a different approach than the experimental flag to enable swizzling of NSFileManager.


Sources/Sentry/SentryFileIOTrackingIntegration.m
```diff
@@ -32,8 +32,8 @@ - (BOOL)installWithOptions:(SentryOptions *)options
 
 - (SentryIntegrationOption)integrationOptions
 {
-    return kIntegrationOptionEnableSwizzling | kIntegrationOptionIsTracingEnabled
-        | kIntegrationOptionEnableAutoPerformanceTracing | kIntegrationOptionEnableFileIOTracing;
+    return kIntegrationOptionIsTracingEnabled | kIntegrationOptionEnableAutoPerformanceTracing
+        | kIntegrationOptionEnableFileIOTracing;
```
armcknight, line 36: Why are we removing `kIntegrationOptionEnableSwizzling` here?

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -1,7 +1,18 @@
...
+     * - Note: Swizzling is enabled by setting ``SentryOptions.enableSwizzling`` to `true`.
+     *         This option allows you to disable swizzling for `NSData` only, while keeping swizzling enabled for other classes.
+     *         This is useful if you want to use manual tracing for file operations.
+     */
+    public var disableDataSwizzling = false
+
     /**
      * Enables swizzling of`NSFileManager` to automatically track file operations.
+     *
+     * - Requires: Swizzling must be enabled by setting ``SentryOptions.enableSwizzling`` to `true`.
```
armcknight, line 15: I think we'd also need this in the headerdoc for `disableDataSwizzling`?

armcknight (commented):
I think we might have to modify the check in `SentryBaseIntegration.shouldBeEnabledWithOptions`: https://github.com/getsentry/sentry-cocoa/blob/a286774f1f994d1d78f7f959e17dd7fc8b10880a/Sources/Sentry/SentryBaseIntegration.m#L104-L107

(Side note on that FYI: I think our process for checking whether an integration should be enabled is really convoluted, going back and forth between superclass and subclass, with the enum values. I have several local branches that simplify it and put it just into the subclasses, but that work stalled due to other priorities.)

Also, it seems wrong that we are able to change the logic in the swizzling wrappers without requiring changes in tests. WDYT @philipphofmann ?

Sources/Sentry/SentryFileIOTrackingIntegration.m
philprime: Because if it stays there, people can not use the manual methods introduced in #4862 and #4863

Sources/Swift/SentryExperimentalOptions.swift
philprime: Can you please elaborate how the current note not sufficient?

Sources/Sentry/SentryNSDataSwizzling.m
```diff
@@ -23,6 +23,11 @@ - (void)startWithOptions:(SentryOptions *)options tracker:(SentryFileIOTracker *
 {
     self.tracker = tracker;
 
+    if (!options.enableSwizzling || options.experimental.disableDataSwizzling) {
+        SENTRY_LOG_DEBUG(@"Auto-tracking of NSData is disabled")
+        return;
```
philipphofmann, line 28: `m`: Splitting this up into two early returns and printing one log message per option helps to understand which flag exactly was responsible for disabling the NSDataSwizzling.

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -1,7 +1,18 @@
 @objcMembers
 public class SentryExperimentalOptions: NSObject {
+    /**
+     * Disables swizzling of`NSData` to automatically track file operations.
+     *
+     * - Note: Swizzling is enabled by setting ``SentryOptions.enableSwizzling`` to `true`.
+     *         This option allows you to disable swizzling for `NSData` only, while keeping swizzling enabled for other classes.
+     *         This is useful if you want to use manual tracing for file operations.
+     */
+    public var disableDataSwizzling = false
```
philipphofmann, line 10: `h`: We don't have a single option starting with `disable`. Please change it to an enable flag and make the default true.
```suggestion
    public var enableDataSwizzling = true
```

Sources/Swift/SentryExperimentalOptions.swift
philipphofmann: I think @armcknight point is that you should also add this information to the headerdoc of `disableDataSwizzling`, because the same applies for `disableDataSwizzling`.

Sources/Swift/SentryExperimentalOptions.swift
philipphofmann, line 10: `h`: I just saw that I approved the PR for adding the option `enableFileManagerSwizzling` in GH-4634. As our other options related to auto-tracing don't contain swizzling, I suggest renaming `disableDataSwizzling` to `enableDataTracing` and `enableFileManagerSwizzling` to `enableFileManagerTracing`. As `enableFileManagerSwizzling` is experimental I think we don't need to deprecate it. We can simply rename it with pointing it out in the changelog.

philipphofmann (commented):
We need to algin the new option with the other options.

philprime:
I would like to propose that we enable swizzling the NSFileManager by default.

philprime:
@philipphofmann to pick up your comment and re-iterate everything that happened in #4605:

> `h`: I just saw that I approved the PR for adding the option `enableFileManagerSwizzling` in [GH-4634](https://github.com/getsentry/sentry-cocoa/pull/4634). As our other options related to auto-tracing don't contain swizzling, I suggest renaming `disableDataSwizzling` to `enableDataTracing` and `enableFileManagerSwizzling` to `enableFileManagerTracing`. As `enableFileManagerSwizzling` is experimental I think we don't need to deprecate it. We can simply rename it with pointing it out in the changelog.

I believe that we need to decide on the actual purpose of these flags, because they are getting mixed up right now.

In the currently released version we have these options:
- `enableSwizzling` to enable swizzling
- `isTracingEnabled` to enable tracing 
- `enableAutoPerformanceTracing` to enable auto performance tracing
- `enableFileIOTracing` to enable file IO tracing

As the File IO integration only swizzled `NSData` before, these options are sufficient. Then we introduced swizzling of `NSFileManager`, but because swizzling can be dangerous it had to be enabled using `enableFileManagerSwizzling` (opt-in). A user could now enable swizzling of the file-manager to have both, but can also disable the entire integration by setting `enableSwizzling` to `false`.

As the file IO tracking integration only worked with swizzling this was fine. 
Now we want to introduce manual file IO tracking in #4862 and #4863, which are also part of this integration.

The manual tracking does *not* use swizzling, therefore it should also work when `enableSwizzling` is set to false → that's why we remove it from the integration options, as these only include required options.

Now we are in a situation where swizzling is enabled and people use manual tracking which can cause conflicts. Therefore we need to allow them to keep `enableSwizzling` set to `true` because it is used by other integrations, but allow to **explicitly disable swizzling** for `NSData` and `NSFileManager`, that's what the two newly introduced options are for.

 That's why I believe we should keep the naming at `enableFileManagerSwizzling` and `enableDataSwizzling` rather than `enableFileManagerTracing` or `enableDataTracing`, because that would also imply that you enable tracing in the manual methods in #4862 and #4863, which is not the case!

I don't see a point in introducing flags for the enabling the methods in #4862 or #4863, at least not in this PR right here.

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -1,9 +1,22 @@
 @objcMembers
 public class SentryExperimentalOptions: NSObject {
+    /**
+     * Disables swizzling of`NSData` to automatically track file operations.
```
philprime, line 4:
```suggestion
     * Enables swizzling of`NSData` to automatically track file operations.
```

Sources/Sentry/SentryFileIOTrackingIntegration.m
philprime: I added a comment to the PR with details.

Sources/Swift/SentryExperimentalOptions.swift
philprime: I added a comment to the PR so we can find it again in the future. Please check it, leave feedback and review again.

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -1,9 +1,22 @@
...
+
     /**
      * Enables swizzling of`NSFileManager` to automatically track file operations.
+     *
+     * - Note: Swizzling is enabled by setting ``SentryOptions.enableSwizzling`` to `true`.
+     *         This option allows you to disable swizzling for `NSFileManager` only, while keeping swizzling enabled for other classes.
+     *         This is useful if you want to use manual tracing for file operations.
      */
-    public var enableFileManagerSwizzling = false
+    public var enableFileManagerSwizzling = true
```
philipphofmann, line 19: `h`: Why did you decide to turn this experimental feature on by default? Do we have enough data to know that the swizzling of the FileManager doesn't cause any major problems? Am I missing something?

Side note: We can safely set `enableDataSwizzling` to true` because we swizzle NSData already for years without major problems. It's acceptable to deviate the default values here.

philipphofmann (commented):
I'm a bit confused why we turn on the fileManagerSwizzling by default.

Sources/Swift/SentryExperimentalOptions.swift
philprime: I'll revert and create an issue so we eventually enable it.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Improvements
+
+> [!WARNING]
+> This release enables the experimental feature `enableFileManagerSwizzling` by default. You can disable it by setting `experimental.enableFileManagerSwizzling` to `false`.
```
philipphofmann, line 8: `h`: The changelog needs an update to reflect that `enableFileManagerSwizzling` isn't enabled by default.

philipphofmann (approved):
LGTM, when fixing the changelog.

CHANGELOG.md
philprime: As the option is not changed in this PR anymore, I removed it from the changelog all-together

PR #4861, opened 2025-02-18
refactor: Add file IO tracker to dependency injection container

philprime:
This PR is derived from #4605 

It moves the File IO tracker to the dependency injection container because it will be used by automatic/swizzled tracking and manual tracking.

#skip-changelog


Sources/Sentry/include/SentryFileIOTracker.h
```diff
@@ -8,6 +8,15 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SentryFileIOTracker : NSObject
 SENTRY_NO_INIT
 
+/**
+ * Convenience accessor to the shared instance of the tracker in the dependency container.
+ *
+ * @note Can be used from Swift without import the entire dependency container.
```
armcknight, line 14: I get the convenience of this, but I don't think we should start moving backwards to every class managing its own singleton. Currently we're injecting it into Swift implementations from ObjC wrappers of them, where we can access the instance from the dep container. 

I also don't see it being used in this PR, so maybe we can kick the can down the road until we actually need to use it from Swift.

philprime: This PR is derived from #4605 to split up the rather large PR into multiple smaller ones.

The introduced static instanece is used in so #4862 and #4863 so, no, we can not hold back and I need this merged now.

philprime: Furthermore, the class does **not** manage the shared instance, it is simply an accessor to the dependency container which is not available in Swift.

philipphofmann: > Currently we solve this problem by injecting into Swift implementations from ObjC wrappers of them, where we can access the instance from the dep container.

I agree with @armcknight here. We already use the ObjC wrappers. We should only deviate from this approach if we have good reasons. Both solutions aren't ideal, but I would stick to the ObjC for now to keep it consistent. If this causes plenty of merge conflicts, it's acceptable for me to keep this now and fix it later.

philipphofmann (approved):
LGTM.

Sources/Sentry/include/SentryFileIOTracker.h
philprime: Maybe I missed the usual implementation, please point me to an example and the relevant develop docs so I can change my implementation accordingly. I do not intend to introduce a new approach, I just want it to compile and lint as necessary.

philipphofmann: For example the SessionReplayIntegration is written in ObjC mainly for that purpose. It glues together the Swift code and passes in all the dependencies. We don't follow a strict pattern here. Simply write the glue code in ObjC and the logic in Swift. Is that sufficient for you?

philipphofmann: I see you're using it in a Swift extension. I don't think it's worth the effort to create an extra ObjC wrapper class for this. I think the current approach is fine. KISS.

Issue #4162, opened 2024-07-15
TTID/TTFD for SwiftUI


kahest:
### Problem Statement

What it says on the tin. :)

Please upvote if this is important to you.

### Solution Brainstorm

Might/will require introducing new APIs to call manually

### Tasks

- [X] Implementation: [https://github.com/getsentry/sentry-cocoa/pull/4596](https://github.com/getsentry/sentry-cocoa/pull/4596)
- [X] Support for `sentryTrace`: [https://github.com/getsentry/sentry-cocoa/issues/4787](https://github.com/getsentry/sentry-cocoa/issues/4787)
- [X] ~~Mark as experimental in code~~ [https://github.com/getsentry/sentry-cocoa/pull/4869](https://github.com/getsentry/sentry-cocoa/pull/4869)
- [X] Docs (see [https://github.com/getsentry/sentry-cocoa/issues/4162#issuecomment-2663444176](https://github.com/getsentry/sentry-cocoa/issues/4162#issuecomment-2663444176))
- [X] ~~TBD: Add feature usage tracking?~~

kahest:
relevant docs pages:
- https://docs.sentry.io/platforms/apple/guides/ios/features/experimental-features/
- https://docs.sentry.io/platforms/apple/guides/ios/tracing/instrumentation/automatic-instrumentation/#time-to-initial-display
- https://docs.sentry.io/platforms/apple/guides/ios/tracing/instrumentation/automatic-instrumentation/#time-to-full-display
- https://docs.sentry.io/platforms/apple/guides/ios/tracing/instrumentation/swiftui-instrumentation/

philprime:
The docs should clarify the following question:

> If we use the update view modifier with waitForFullDisplay set to true, do we still need to call SentrySDK.reportFullyDisplayed() and from where?

philprime:
Superseded by https://github.com/getsentry/sentry-docs/pull/15638

PR #4869, opened 2025-02-19
docs: add warning for experimental feature

philprime:
Splits up the initializer of `SentryTracedView` into one with the parameter `waitForFullDisplay` and one without. The one with the parameter is marked as experimental.

See #4162

#skip-changelog

### DIFF
--- a/Sources/SentrySwiftUI/SentryTracedView.swift
+++ b/Sources/SentrySwiftUI/SentryTracedView.swift
@@ -121,11 +121,25 @@ public struct SentryTracedView<Content: View>: View {
     /// Creates a view that measures the performance of its `content`.
     ///
     /// - Parameter viewName: The name that will be used for the span, if nil we try to get the name of the content class.
+    /// - Parameter content: The content that you want to track the performance
+    public init(_ viewName: String? = nil, @ViewBuilder content: @escaping () -> Content) {
+        self.content = content
+        let name = viewName ?? SentryTracedView.extractName(content: Content.self)
+        let nameSource = viewName == nil ? SentryTransactionNameSource.component : SentryTransactionNameSource.custom
+        let initialViewModel = SentryTraceViewModel(name: name, nameSource: nameSource, waitForFullDisplay: nil)
+        _viewModel = State(initialValue: initialViewModel)
+    }
+
+    /// Creates a view that measures the performance of its `content`.
+    ///
+    /// - Parameter viewName: The name that will be used for the span, if nil we try to get the name of the content class.
     /// - Parameter waitForFullDisplay: Indicates whether this view transaction should wait for `SentrySDK.reportFullyDisplayed()`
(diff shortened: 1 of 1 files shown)


Sources/SentrySwiftUI/SentryTracedView.swift
```diff
@@ -118,14 +118,28 @@ public struct SentryTracedView<Content: View>: View {
...
     /// Creates a view that measures the performance of its `content`.
     ///
     /// - Parameter viewName: The name that will be used for the span, if nil we try to get the name of the content class.
     /// - Parameter waitForFullDisplay: Indicates whether this view transaction should wait for `SentrySDK.reportFullyDisplayed()`
     /// in case you need to track some asyncronous task. This is ignored for any `SentryTracedView` that is child of another `SentryTracedView`.
     /// If nil, it will use the `enableTimeToFullDisplayTracing` option from the SDK.
     /// - Parameter content: The content that you want to track the performance
-    public init(_ viewName: String? = nil, waitForFullDisplay: Bool? = nil, @ViewBuilder content: @escaping () -> Content) {
+    ///
+    /// - Warning: This initializer is an experimental feature and may still have bugs.
```
kahest, line 141: do we have a standardized way to do this? I see in ObjC .h/.m we use @warning, in Swift we use "/// - Experiment" in [I think 2 places](https://github.com/search?q=repo%3Agetsentry%2Fsentry-cocoa+experimental+language%3ASwift&type=code)

kahest: also I think it'd be sufficient if we keep one function and mark the param as experimental, since I don't think Xcode will automatically surface the warning in either way, right?

philprime: I am fine with `- Experiment` instead of `- Warning` too.

The `/// - Experiment` field is displayed as a bright purple box like this in Xcode. I do not believe it is possible to show this kind of visible box just for one parameter. That was one of the reasons I duplicated the initializer, so that one is marked like this, while the other isn't.

image

If we mark it as `@available(*, unavailable, message: "This is experimental and may change.")`, it will block the use of the initializer:

image

We could mark it as deprecated, but that could lead to issues if people have strict enforcing of linters to not use any deprecated methods.

`@available(*, deprecated, message: "This is experimental and may change.")`

image

kahest: thanks for explaining the different options - then your solution seems the best middle ground 👍

philprime:
@philipphofmann IIRC this should actually have been experimental in the first place.
This kept getting pushed back, @kahest should we even add the experimental note?

kahest:
yeah fair point, it's out for roughly 4 months now. I think we can skip the experiment now, and focus on adding the necessary docs -> https://github.com/getsentry/sentry-cocoa/issues/4162#issuecomment-2663444176

PR #4871, opened 2025-02-19
chore(tests): add capturing and uploading of screen capture of ui tests

philprime:
- Enables automatic video capture of ui tests
- Adds upload of the captured videos on ui test failure

See #4848 for more context and details how this works.

My intention with this PR is to have visual assitance why our UI tests are failing without changes (flakiness).

Things we need to consider:
- Maybe we should change the path of the output xcresult so that it isn't included in the raw test log upload
- We need to decide if we want a video for all results, or only for failing results

#skip-changelog


philprime:
Additional context on the directory behaviour:

https://docs.fastlane.tools/advanced/fastlane/#directory-behavior

philprime:
Example from the [failed UI test](https://github.com/getsentry/sentry-cocoa/actions/runs/13416607222/job/37478946831?pr=4871):

Screenshot 2025-02-19 at 17 27 45

Btw seems like there's a race condition happening

philipphofmann (approved):
Thanks, let's give this a shot 🚀

Issue #4709, opened 2025-01-14
feedback: First shipment to public as early access / beta


armcknight:
### Description

TODO:
- [x] remove all swiftlint `todo` overrides to make sure nothing was missed that was saved for later
- [x] final fixes: https://github.com/getsentry/sentry-cocoa/pull/4873
- [x] move interfaces from SentrySDK+Private and SentryOptions+Private to the public headers: https://github.com/getsentry/sentry-cocoa/pull/4874
- [x] approve and publish release: 
    - [x] https://github.com/getsentry/publish/issues/5137 (retracted, will need to run a new release)
    - [x] https://github.com/getsentry/publish/issues/5144
- [x] docs: https://github.com/getsentry/sentry-docs/pull/11191

PR #4874, opened 2025-02-19
feat(feedback): initial release

armcknight:
Publicly exposes the feature. For https://github.com/getsentry/sentry-cocoa/issues/4709

TODO: 
- [x] deprecate old api
- [ ] Docs PR: https://github.com/getsentry/sentry-docs/pull/11191


Tests/SentryTests/SentryClientTests.swift
```diff
@@ -1464,24 +1478,47 @@ class SentryClientTest: XCTestCase {
...
+        let sut = fixture.getSutDisabledSdk()
+        sut.capture(feedback: fixture.feedback, scope: fixture.scope)
+        assertNothingSent()
+    }
+    
+    @available(*, deprecated, message: "This is only marked as deprecated because assertNothingSent is marked as deprecated, due to it using a deprecated property inside it. When that property usage is removed, this deprecation annotation can go away.")
+    func testCaptureFeedback_WithEmptyEventId() {
+        let sut = fixture.getSut()
+        sut.capture(feedback: fixture.feedback, scope: fixture.scope)
+        // TODO: SentryFeedback will not require event IDs, so we should assert that is _is_ sent here
```
armcknight, line 1520: Just need to implement this check

Tests/SentryTests/SentryClientTests.swift
```diff
@@ -1464,24 +1478,48 @@ class SentryClientTest: XCTestCase {
...
+        sut.capture(feedback: fixture.feedback, scope: fixture.scope)
+        assertNothingSent()
+    }
+    
+    @available(*, deprecated, message: "This is only marked as deprecated because assertNothingSent is marked as deprecated, due to it using a deprecated property inside it. When that property usage is removed, this deprecation annotation can go away.")
+    func testCaptureFeedback_WithEmptyEventId() {
+        let sut = fixture.getSut()
+        XCTAssertTrue(fixture.transportAdapter.sendEventWithTraceStateInvocations.isEmpty)
+        sut.capture(feedback: fixture.feedback, scope: fixture.scope)
+        XCTAssertFalse(fixture.transportAdapter.sendEventWithTraceStateInvocations.isEmpty)
```
armcknight, line 1521: @philipphofmann LMK what you think of this test. I debugged and traced execution to something I could check, which turned out to be the transport adapter, so just used that. I do see other test cases using `sendEventWithTraceStateInvocations` as well.

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -766,6 +766,29 @@ NS_SWIFT_NAME(Options)
...
+ * A block that can be defined that receives a user feedback configuration object to modify.
+ * @warning This is an experimental feature and may still have bugs.
+ * @note This is unrelated to @c SentrySDK.captureUserFeedback which is the old method of submitting
+ * user feedback you've already gathered via your own UI
+ * (see https://docs.sentry.io/platforms/apple/user-feedback/#user-feedback-api). The new method
+ * uses either this block to configure and a widget and UI form to gather feedback, or directly
+ * submits calling the new method @c SentrySDK.captureFeedback (see
+ * https://docs.sentry.io/platforms/apple/user-feedback/configuration/).
+ */
+@property (nonatomic, copy, nullable)
```
philprime, line 788: `l`: Is there any value in not configuring this in the root options, but instead moving it to something like `options.userFeedback.configure { ... }`, implementing the nested options in Swift instead? Similar to the experimental options

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
```diff
@@ -56,7 +56,7 @@ extension SentryFeedback: SentrySerializable {
         if let associatedEventId = associatedEventId {
             dict["associated_event_id"] = associatedEventId.sentryIdString
         }
-        dict["source"] = source.rawValue
+        dict["source"] = source.description
```
philprime, line 59: `l`: Just want to point out that in most cases it is bad practice to use `description` in any serialization, because if the implementation of `description` in line 7 is removed, the Swift compiler could still synthesize the `CustomStringConvertible`, and it's not guaranteed to match the raw string value (could include information like `Optional<String>`)

Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
```diff
@@ -43,6 +43,39 @@ class SentryFeedbackTests: XCTestCase {
         XCTAssertEqual(try XCTUnwrap(attachments.first).contentType, "application/png")
     }
     
+    func testSerializeCustomFeedback() throws {
+        let sut = SentryFeedback(message: "Test feedback message", name: "Test feedback provider", email: "test-feedback-provider@sentry.io", source: .custom, attachments: [Data()])
+        
+        let serialization = sut.serialize()
+        XCTAssertEqual(try XCTUnwrap(serialization["message"] as? String), "Test feedback message")
```
philprime, line 50: `l`: Why the nesting of `XCTUnwrap` and `XCTAssertEqual`? AFAIK the XCTUnwrap is unnecessary, because we can compare `serialization["message"] as? String` directly to `Test feedback message`

philprime (approved):
Left a couple of comments, but other than that LGTM

Samples/iOS-ObjectiveC/iOS-ObjectiveC/ViewController.m
```diff
@@ -78,6 +78,28 @@ - (IBAction)captureUserFeedback:(id)sender
     [SentrySDK captureUserFeedback:userFeedback];
 }
 
+- (IBAction)captureNewUserFeedback:(id)sender
```
philipphofmann, line 81: `l`: Using V2 instead of new for naming is usually better. What if we have user feedback V3? Are we going to call it newnew then?

Sources/Sentry/Public/SentryClient.h
```diff
@@ -95,9 +96,19 @@ SENTRY_NO_INIT
...
  * @param userFeedback The user feedback to send to Sentry.
+ * @deprecated Use @c -[SentryClient @c captureFeedback:withScope:] .
  */
 - (void)captureUserFeedback:(SentryUserFeedback *)userFeedback
-    NS_SWIFT_NAME(capture(userFeedback:));
+    NS_SWIFT_NAME(capture(userFeedback:))
+        DEPRECATED_MSG_ATTRIBUTE("Use -[SentryClient captureFeedback:withScope:].");
+
+/**
+ * Captures a new-style user feedback and sends it to Sentry.
```
philipphofmann, line 106: `m`: Again maybe we should talk about V2 instead of new-style, cause this is a public API.

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -766,6 +766,29 @@ NS_SWIFT_NAME(Options)
...
 @property (nonatomic, readonly) SentryExperimentalOptions *experimental;
 
+#if TARGET_OS_IOS && SENTRY_HAS_UIKIT
+
+/**
+ * Block used to configure the user feedback widget, form, behaviors and submission data.
+ */
+API_AVAILABLE(ios(13.0))
+typedef void (^SentryUserFeedbackConfigurationBlock)(
+    SentryUserFeedbackConfiguration *_Nonnull configuration);
```
philipphofmann, line 776: `l`: We have all the other typedefs for the callbacks in the options in SentryDefines.h. You could move it there for consistency.

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -766,6 +766,29 @@ NS_SWIFT_NAME(Options)
...
+
+/**
+ * A block that can be defined that receives a user feedback configuration object to modify.
+ * @warning This is an experimental feature and may still have bugs.
+ * @note This is unrelated to @c SentrySDK.captureUserFeedback which is the old method of submitting
+ * user feedback you've already gathered via your own UI
```
philipphofmann, line 782: `m`: I would write about deprecated instead of old.
```suggestion
 * @note This is unrelated to @c SentrySDK.captureUserFeedback, which is the deprecated method of submitting
 * user feedback you've already gathered via your own UI
```

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -766,6 +766,29 @@ NS_SWIFT_NAME(Options)
...
+ * (see https://docs.sentry.io/platforms/apple/user-feedback/#user-feedback-api). The new method
+ * uses either this block to configure and a widget and UI form to gather feedback, or directly
```
philipphofmann, line 784: `l`:
```suggestion
 * (see https://docs.sentry.io/platforms/apple/user-feedback/#user-feedback-api). This method
 * uses either this block to configure and a widget and UI form to gather feedback, or directly
```

Tests/SentryTests/SentryClientTests.swift
```diff
@@ -1464,24 +1478,48 @@ class SentryClientTest: XCTestCase {
...
+    }
+    
+    @available(*, deprecated, message: "This is only marked as deprecated because assertNothingSent is marked as deprecated, due to it using a deprecated property inside it. When that property usage is removed, this deprecation annotation can go away.")
+    func testDisabled_FeedbackNotSent() {
+        let sut = fixture.getSutDisabledSdk()
+        sut.capture(feedback: fixture.feedback, scope: fixture.scope)
+        assertNothingSent()
+    }
+    
+    @available(*, deprecated, message: "This is only marked as deprecated because assertNothingSent is marked as deprecated, due to it using a deprecated property inside it. When that property usage is removed, this deprecation annotation can go away.")
```
philipphofmann, line 1516: Oh my god this repetition is very annoying 🤦. We will fix it later.

Tests/SentryTests/SentryClientTests.swift
philipphofmann: I think it's OK to do it like that, and the test should assert some properties of the recorded event. Now it only asserts that there was an event recorded but not that the contents of the event are correct.

philipphofmann (approved):
LGTM, with a couple of things to improve, but nothing blocking.

Samples/iOS-ObjectiveC/iOS-ObjectiveC/ViewController.m
armcknight: oh yeah, we talked about this before with the profiling stuff 🤣 I'll change it to V2

Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
armcknight: Here's a comparison of how they fail:
```
func testFoo() throws {
    let a: String? = nil
    XCTAssertEqual(a, "a")                // 1. XCTAssertEqual failed: ("nil") is not equal to ("Optional("a")")
    XCTAssertEqual(try XCTUnwrap(a), "a") // 2. XCTUnwrap failed: expected non-nil value of type "String"
}
```
I prefer (2) because (1) coerces even the string literal in the expression to an optional type, which is not the intended comparison. I want to both ensure it's not nil, and then compare then concrete value inside it. I prefer it failing at the unwrap before trying to compare.

Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift
armcknight: Good point, i'll rename it and avoid using CustomStringConvertible for this.

armcknight: Hmm, are we sure swiftc will synthesize CustomStringConvertible for an `@objc enum`? When I removed the conformance, here's how compilation broke:
image

armcknight: I think you're thinking of NSObject descendants, since `- [NSObject description]` automatically gets inherited by them all?

armcknight: Out of an abundance of caution I will make the change, but I found this little exploration interesting so just wanted to share.

Sources/Sentry/Public/SentryOptions.h
armcknight: From the perspective of SentryOptions, this is no different, since there is only one point of entry to the user feedback configuration, which is SentryOptions.configureUserFeedback, and everything else is contained in there.

Sources/Sentry/Public/SentryOptions.h
armcknight: i reworded it a little differently because the "the new method" you wanted to change didn't refer to an actual function/method to call, but rather, an entire strategy of how to gather feedback: https://github.com/getsentry/sentry-cocoa/blob/ae42d842ef5367a85ca99fd3564cef4a4d8c8e75/Sources/Sentry/Public/SentryOptions.h#L771-L782

Tests/SentryTests/SentryClientTests.swift
armcknight: Added checks of the envelope contents in https://github.com/getsentry/sentry-cocoa/pull/4874/commits/530757562689af243b5e11762c59eede7428c509

Issue #4887, opened 2025-02-24
Access to SentryTraceOrigin inside useSpan causes crashes


andre-alves:
### Platform

iOS

### Environment

Production

### Installed

CocoaPods

### Version

8.44.0

### Xcode Version

16.2

### Did it work on previous versions?

8.43.0

### Steps to Reproduce

This crash is happening in production and we are unable to reliably reproduce it.

### Expected Result

No crashes.

### Actual Result

After updating from Sentry 8.43.0 to 8.44.0 our app started crashing with the following stack trace:

```
Exception Type: EXC_CRASH (SIGABRT)
Crashed Thread: 12

Application Specific Information:
*** error for object %p: pointer being freed was not allocated
 > iFood > iFood(11093,0x16d777000) malloc: *** error for object 0x301e224c0: pointer being freed was not allocated


Thread 12 Crashed:
0   libsystem_kernel.dylib          0x3b331a2d4         __pthread_kill
1   libsystem_pthread.dylib         0x425fda598         pthread_kill
2   libsystem_c.dylib               0x3200e1b04         abort
3   libsystem_malloc.dylib          0x3310c8b70         malloc_vreport
4   libsystem_malloc.dylib          0x3310c87e4         malloc_report
5   libsystem_malloc.dylib          0x3310c5d58         find_zone_and_free
6   libswiftCore.dylib              0x30d412178         _swift_release_dealloc
7   libswiftCore.dylib              0x30d41328c         swift::RefCounts<T>::doDecrementSlow<T>
8   iFood                           0x202ffd9a0         __50-[SentryFileIOTracker spanForPath:operation:size:]_block_invoke (SentryFileIOTracker.m:182)
9   iFood                           0x20302d8f8         -[SentryScope useSpan:] (SentryScope.m:165)
10  iFood                           0x202ffd61c         -[SentryFileIOTracker spanForPath:operation:size:] (SentryFileIOTracker.m:178)
11  iFood                           0x202ffddc0         -[SentryFileIOTracker startTrackingReadingFilePath:] (SentryFileIOTracker.m:248)
12  iFood                           0x202ffd174         -[SentryFileIOTracker measureNSDataFromFile:method:] (SentryFileIOTracker.m:93)
13  iFood                           0x203018aa8         __32+[SentryNSDataSwizzling swizzle]_block_invoke_2.86 (SentryNSDataSwizzling.m:90)
[...]

PR #4888, opened 2025-02-24
fix: move assignment of file IO span origin outside of block

philprime:
## :scroll: Description

Moves the assignment of the `ioSpan` in the `NSBlock` to the outside, as it might affect memory management. 

Note: This might also not be relevant anymore as the access to Bridged-from-Swift type `SentryTraceOrigin` is not in the block anymore.

See #4887 

Effect can not be tested reliably. Method `spanForPath` is already covered by tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryFileIOTracker.m
+++ b/Sources/Sentry/SentryFileIOTracker.m
@@ -198,7 +198,6 @@ NSString *const SENTRY_TRACKING_COUNTER_KEY = @"SENTRY_TRACKING_COUNTER_KEY";
         ioSpan = [span startChildWithOperation:operation
                                    description:[self transactionDescriptionForFile:path
                                                                           fileSize:size]];
-        ioSpan.origin = origin;
     }];
 
     if (ioSpan == nil) {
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentryFileIOTracker.m
```diff
@@ -194,11 +194,9 @@ - (BOOL)measureNSFileManagerCreateFileAtPath:(NSString *)path
     }
 
     __block id<SentrySpan> ioSpan;
+    NSString *spanDescription = [self transactionDescriptionForFile:path fileSize:size];
     [SentrySDK.currentHub.scope useSpan:^(id<SentrySpan> _Nullable span) {
-        ioSpan = [span startChildWithOperation:operation
-                                   description:[self transactionDescriptionForFile:path
-                                                                          fileSize:size]];
-        ioSpan.origin = origin;
+        ioSpan = [span startChildWithOperation:operation description:spanDescription];
```
philipphofmann, line 199: `m`: Please add a comment to keep the logic here to a minimum, because of the issue we just had and reference the issue.

philipphofmann (approved):
LGTM. Food for thought: What can we do to avoid such an issue in the future? Do you have any ideas, @philprime?

Issue #4260, opened 2024-08-08
Report Fatal App Hangs


philipphofmann:
### Description

When an app hang leads to a watchdog termination, we can report this as a fatal app hang. When the SDK detects a fully blocking app hang, it needs to store the app hang to disk and wait for the app to recover from the app hang. When the app doesn't recover, the SDK must send the app hang as a fatal on the next app launch. When the app recovers from the app hang, then it must send it as a fully blocking app hang. The downside of this approach is that app hangs will be reported a bit delayed, which is acceptable. 

We must align with Android and report sessions as abnormal when a fatal app hang occurs. For reference: https://github.com/getsentry/sentry-java/pull/2475

This issue is a prerequisite to https://github.com/getsentry/sentry-cocoa/issues/3319, because the app hang rate only considers fatal app hangs.

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4889
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4893
```

philipphofmann:
Fixed with https://github.com/getsentry/sentry-cocoa/pull/4889. We will add the app hang rate with https://github.com/getsentry/sentry-cocoa/issues/3319.

PR #4893, opened 2025-02-25
ref: Rename SentryCrashIntegrationSessionHandler

philipphofmann:
Rename from SentrySessionCrashedHandler to
SentryCrashIntegrationSessionHandler, as we're going to so we can add more logic to it, which isn't only related to crashed sessions.

Required for GH-4260.

#skip-changelog


philprime (approved):
LGTM

PR #4894, opened 2025-02-25
refactor: Change type of Swift String constants to NSString

philprime:
## :scroll: Description

Change the type of constants to `NSString`.

We purposely defined constants such as `SentryTraceOrigin`, `SentrySpanOperation` and `SentrySpanDataKey` in Swift for our long-term effort in migrating to Swift.

Stack traces in #4887 indicate that the bridging of Swift String to NSString might be causing memory issues when used from Objective-C blocks (i.e. `useSpan`).

This PR is an attempt to reduce the amount of bridging performed.

Can not be reproduced/tested reliably.


philipphofmann (approved):
LGTM, if the tests are green. As already pointed out we should follow up with a strategy to prevent regressions.

philprime:
It keeps failing with flaky UI tests unrelated to the changes. Merging this now.

PR #4895, opened 2025-02-25
ci: Increase allowed binary size

philipphofmann:
Fix CI by allowing bit of increase for the binary size. At some point, we need to split up the SDK into multiple modules or come up with other strategies to decrease the binary size.


philprime (approved):
LGTM, do we have an issue to track efforts of modularization?

philipphofmann:
> LGTM, do we have an issue to track efforts of modularization?

We have a GH issue https://github.com/getsentry/sentry-cocoa/issues/3219

PR #4896, opened 2025-02-25
refactor: Replace calls useSpan with callback to direct span accessor

philprime:
## :scroll: Description

Removes the callback-based access to the span in scope with a direct accessor

We have noticed issues in #4887 relevant to the `useSpan` method. @philipphofmann mentioned before that `NSBlock` has caused issues in the past. The current implementation of `useSpan` really doesn't do anything other than accessing the `span` then calling the callback, therefore inlining the method into uses could reduce the usage of `NSBlock`.

### DIFF
--- a/Sources/Sentry/Public/SentryScope.h
+++ b/Sources/Sentry/Public/SentryScope.h
@@ -158,7 +158,9 @@ NS_SWIFT_NAME(Scope)
  * Mutates the current transaction atomically.
  * @param callback the SentrySpanCallback.
  */
-- (void)useSpan:(SentrySpanCallback)callback;
+- (void)useSpan:(SentrySpanCallback)callback DEPRECATED_MSG_ATTRIBUTE("use `span` instead");
+
+- (id<SentrySpan> _Nullable)span;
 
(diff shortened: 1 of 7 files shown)


Sources/Sentry/Public/SentryScope.h
```diff
@@ -158,7 +158,12 @@ NS_SWIFT_NAME(Scope)
  * Mutates the current transaction atomically.
  * @param callback the SentrySpanCallback.
  */
-- (void)useSpan:(SentrySpanCallback)callback;
+- (void)useSpan:(SentrySpanCallback)callback DEPRECATED_MSG_ATTRIBUTE("use `span` instead");
```
philipphofmann, line 161: `m`: I think we should mention in the code docs why we deprecated this.

philipphofmann (approved):
LGTM

Sources/Sentry/Public/SentryScope.h
philipphofmann: I think that's sufficient. Thanks

PR #4898, opened 2025-02-25
test: Add minimal test mock library

philprime:
## :scroll: Description

Adds minimal testing framework for mocking dependencies and assert function calling in unit testing. Heavily inspiried by `jest`.

We want to add unit tests which tests files and logic in isolation, as sometimes spinning up the entire SDK can be cumbersome. Using subclasses and protocols we can inject mocked instances. The minimal framework allows creating readable assertion tests.

This has been a timeboxed experiment to evaluate the option of using this instead of a third-party mock SDK which can be bloated and bring in additional unwanted complexity.

## :crystal_ball: Next steps

- [ ] Add testing for framework
- [ ] Add more documentation
- [ ] Support more than 3 arguments


philprime:
Closing this because not needed anymore for #4862

PR #4899, opened 2025-02-25
ci: push iOS-Swift builds to testflight for release processes

armcknight:
It occurred to me that it might be nice to have a build of iOS-Swift to play with as part of the release process.

#skip-changelog


philprime (approved):
I like that. Do we inject the release version / build number into the build so we can identify it in App Store Connect?

philipphofmann (approved):
Yes, we can give this a shot.

PR #4901, opened 2025-02-26
test: invert launch arg so feedback widget is injected by default

armcknight:
Before we generate a release dogfood build of iOS-Sample, we need to change showing the feedback widget from opt-in via launch arg, to opt-out, as that was previously the only way to show it, but we can't set launch args in the dogfood build.

#skip-changelog


philprime (approved):
LGTM

PR #4903, opened 2025-02-26
test: add integration test using Cocoapods

philprime:
## :scroll: Description

Adds an integration test to install the SDK using Cocoapods and running a simple launch test.

We had a support request asking if we support Swift 6 and Cocoapods.
I believe we do and this PR contains the sample I used to test it.
To make sure we are not regressing, I suggest we add this as an "integration test" to the pipelines.

I did not commit the generated `xcworkspace` on purpose, so that the Cocoapods install command actually does some work in the setup.

Used the same commands locally.

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog


.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,51 @@
+name: Integration Tests
```
philipphofmann, line 1: `m`: As this is currently only specific to CocoaPods, I think we should rename this to CocoaPods Tests including the file.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
+  cocoapods:
+    name: Integration Test for ${{ matrix.sample.name }}
+    runs-on: macos-15
+    strategy:
+      matrix:
+        sample:
+          - name: "iOS-Cocoapods-Swift6"
+            path: Samples/iOS-Cocoapods-Swift6
```
philipphofmann, line 29: `m`: Why do we need the matrix? Currently, it only has one dimension.

philipphofmann (approved):
LGTM, thanks a lot.

armcknight (approved):
Excellent, this is great to have and a good starting place to build more later. Thanks for doing it!

PR #4907, opened 2025-02-26
fix: Resolve static assertion failure of const in std::vector

macmade:
## :scroll: Description

The Sentry SDK cannot be built with the upcoming Xcode 16.3 because of changes in the C++ library and compiler.

**1 - Vector of `const` elements in `SentryThreadMetadataCache.cpp`:**

    std::vector<const ThreadHandleMetadataPair> threadMetadataCache_;

Using `const` elements in a vector was disallowed before C++11 and has been relaxed since then.  
However, there's currently no explicit requirement for the elements of a vector in the current C++ standard.

Xcode 16.3 fails with the following error:

    Static assertion failed due to requirement '!is_const<const sentry::profiling::ThreadMetadataCache::ThreadHandleMetadataPair>::value': std::allocator does not support const types

**2 - Use of Clang extension for variable length arrays:**

The macro `SentryCrashMC_NEW_CONTEXT` makes use of variable length arrays.  
This now triggers a compiler error in Xcode 16.3.

The error is now silenced by ignoring `-Wvla-cxx-extension` with a `#pragma clang`.  
Unfortunately, this is a new flag, so it cannot be set in `SDK.xcconfig` as it would break builds on previous Xcode versions.

These changes allow the Sentry SDK to be built with Xcode 16.3 Beta.

The Sentry SDK now compiles with Xcode 16.3.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryThreadInspector.m
+++ b/Sources/Sentry/SentryThreadInspector.m
@@ -12,6 +12,10 @@
 #import "SentryThread.h"
 #include <pthread.h>
 
+#if defined( __clang__ ) && defined( __clang_major__ ) && __clang_major__ >= 17
+#pragma clang diagnostic ignored "-Wvla-cxx-extension"
+#endif
+
 @interface SentryThreadInspector ()
 
 @property (nonatomic, strong) SentryStacktraceBuilder *stacktraceBuilder;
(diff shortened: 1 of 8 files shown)


philipphofmann:
@armcknight, can you maybe have a look, as your expertise in C++ is way broader than mine?

armcknight:
Thanks so much for the contribution @macmade ! I would prefer not to silence the warning, so made what I believe will be the actual fix for the variable-length array issue in https://github.com/getsentry/sentry-cocoa/pull/4917.

And I've pushed a separate PR for the const vector type in https://github.com/getsentry/sentry-cocoa/pull/4918.

I'm going to close this PR in favor of those two, but again I really appreciate the effort you put into generating this PR, please don't hesitate to contribute in the future!

PR #4911, opened 2025-02-27
fix: Deadline timeout crash in SentryTracer

philipphofmann:
## :scroll: Description

Replace the NSTimer with dispatch async, which we already use for the idle timeout and haven't had issues for a long time. Dispatch async also has the advantage that it can be called from any thread.

Fixes GH-3320

Unit tests and simulator.l

## :pencil: Checklist


philprime (approved):
LGTM, while I do not fully understand the usage of the timer (yet), I did not spot any obvious issues with the changes.

Tests are still failing, so we should wait for CI to be green.

PR #4912, opened 2025-02-27
chore: change order of pre-commit hooks

philprime:
Changes the order of pre-commit hooks to fix formatting automatically before linting in the final step.

#skip-changelog


philipphofmann (approved):
LGTM

Issue #4886, opened 2025-02-24
Xcode 16.3 prevents `sentry-cocoa` from compiling due to removal of a non-standard extension


bc-lee:
### Platform

iOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

HEAD (`97989471c267e456b5dd3048b45cbb5c2548e5c9`)

### Xcode Version

16.3 beta (`16E5104o`)

### Did it work on previous versions?

Yes, it compiled successfully in **Xcode 16.2** (`16C5032a`).

### Steps to Reproduce

1. Clone [`sentry-cocoa`](https://github.com/getsentry/sentry-cocoa) and check out commit `HEAD (97989471c267e456b5dd3048b45cbb5c2548e5c9)`.
2. Open the sample project:  
   ```
   Samples/iOS-ObjectiveC/iOS-ObjectiveC.xcodeproj
   ```
3. Build the project.

### Expected Result

The project should build successfully.

### Actual Result

The build **fails** with the following error:

<details>
<summary>Click to expand</summary>

```
CompileC /path/to/DerivedData/iOS-ObjectiveC-dhweowpocucrxnengpfevuapqonb/Build/Intermediates.noindex/Sentry.build/Debug-iphonesimulator/Sentry.build/Objects-normal/arm64/SentryThreadMetadataCache.o /path/to/sentry-cocoa/Sources/Sentry/SentryThreadMetadataCache.cpp normal arm64 c++ com.apple.compilers.llvm.clang.1_0.compiler (in target 'Sentry' from project 'Sentry')
    cd /path/to/sentry-cocoa

    Using response file: /path/to/DerivedData/iOS-ObjectiveC-dhweowpocucrxnengpfevuapqonb/Build/Intermediates.noindex/Sentry.build/Debug-iphonesimulator/Sentry.build/Objects-normal/arm64/82b82416624d2658e5098eb0a28c15c5-common-args.resp

[...]

PR #4917, opened 2025-02-27
fix: use a static const to define array size

armcknight:
As pointed out in https://github.com/getsentry/sentry-cocoa/pull/4907, we have a variable-length static array being allocated, which currently breaks in the latest Xcode beta (16.3).

Instead of hiding the issue with a pragma, this PR attempts to fix the problem. I think it was just a matter of how the code was written, since at the end of the day, it's using something that _should_ be computable at compile time: the size of a struct. We just need to rewrite the code in a way that the compiler can know that. 

I originally tried using `constexpr` on the function definition, but that requires changing lots of C code to C++ which led to a bunch of other issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Move assignment of file IO span origin outside of block (#4888)
 - Deadline timeout crash in SentryTracer (#4911)
 - Improve memory-safety by converting Swift constants to Objective-C (#4910)
+- Fix C++ compilation error due to changes in Xcode 16.3 beta's compiler toolchain (#4917)
 
 ## 8.45.0
 
--- a/Sources/SentryCrash/Recording/Tools/SentryCrashMachineContext.c
+++ b/Sources/SentryCrash/Recording/Tools/SentryCrashMachineContext.c
@@ -91,12 +91,6 @@ getThreadList(SentryCrashMachineContext *context)
     return true;
 }
 
-int
-sentrycrashmc_contextSize(void)
-{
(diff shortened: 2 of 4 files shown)


philprime (approved):
LGTM but do we have any tests to make sure this doesn't break more than it fixes? I know that SentryCrash is vendored and therefore only partially tested

kahest:
related issue: https://github.com/getsentry/sentry-cocoa/issues/4886

philipphofmann (approved):
LGTM, I would at least do a smoke test via TestFligh: trigger a crash in the sample app and check that the crash event still works correctly.

indragiek (approved):

armcknight:
@philipphofmann do you manually push builds to testflight for that? PR checks don't do it normally so there is currently no TF build to check it with. I checked it just by running iOS-Swift from xcode, but we don't currently have it set up to upload symbols so the stack trace is currently all <unknown> symbols, but have the addresses and register contents so I think everything is working normally: https://sentry-sdks.sentry.io/issues/5992437707/?project=5428557&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=5m&stream_index=0

armcknight:
Confirmed things look fine: https://sentry-sdks.sentry.io/issues/6044162827/events/latest/?project=5428557&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=latest-event&sort=date&statsPeriod=5m&stream_index=0
image

PR #4918, opened 2025-02-27
fix: dont use unsupported const vector member type

armcknight:
Remove the `const` qualifier for a vector member type, which is unsupported in mainstream C++, and becomes a build error in the upcoming Xcode version.

As reported in https://github.com/getsentry/sentry-cocoa/issues/4886 and demonstrated fixed in https://github.com/getsentry/sentry-cocoa/pull/4907.


philprime (approved):

philipphofmann (approved):
Thanks, @armcknight

indragiek (approved):

PR #4889, opened 2025-02-24
feat: Report fatal app hangs for V2

philipphofmann:
## :scroll: Description

This PR reports fatal app hangs without session updates, only when the experimental option enableAppHangTrackingV2 is enabled. Updating the session will be done in a follow-up PR.

## :bulb: Motivation and Context

This fixes the first part of GH-4260.

## :green_heart: How did you test it?
Unit tests and an archived build distributed for development cause the watchdog doesn't terminate debug builds or simply takes a very long time to do that.

https://sentry-sdks.sentry.io/issues/6328843559/events/c79567d75edf48299845a55e2063a700/?project=5428557
CleanShot 2025-02-24 at 19 41 49@2x

https://sentry-sdks.sentry.io/issues/6328844382/events/8f9916263d124370b28d434f1647133d/?project=5428557
CleanShot 2025-02-24 at 19 42 17@2x

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

I will update the docs for fatal app hangs after I added the session updates.

Issue #3319, opened 2023-10-02
[Apple] App Hang Rate


kahest:
### Description

Similar to ANR Rates for Android, support App Hangs Rate for iOS. Upvote this issue or leave a comment if this is important for you!

Some required tasks:
- [Correlate App Hangs with sessions](https://develop.sentry.dev/sdk/sessions/#session-update-payload)
- Enable metric extraction/processing
- Support [showing the card on project page](https://docs.sentry.io/platforms/android/performance/instrumentation/performance-metrics#foreground-anr-rate)
- Update docs

Example: Screenshot from Android SDK sample app (ANR Rate on Android ~= App Hang Rate on iOS)

![Image](https://github.com/user-attachments/assets/3615d84f-2197-4806-8bd9-304d9233c41e)

Related PRs for ANR rate:
https://github.com/getsentry/rfcs/pull/37
https://github.com/getsentry/sentry/pull/42255
https://github.com/getsentry/sentry/pull/42460
https://github.com/getsentry/sentry/pull/42785
https://github.com/getsentry/relay/pull/1665
https://github.com/getsentry/sentry-docs/pull/6127

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2022) by [Unito](https://www.unito.io)

ckurt3:
This is important to us, as we need to roll this metric up to leadership. Thank you!

chodges15:
@kahest - we have the same use case as above where our leadership wants to see this metric on iOS, would be grateful if this could be prioritized
cc @yuval-sentry

kahest:
Thanks for everybody leaving comments and upvoting! We've elevated the priority and aim to start implementation soon. We will keep this issue updated.

kahest:
Update: This is currently under development and we're making good progress. The previous timeline changed because we decided to prioritize improvements to App Hang detection and reporting, to ultimately make App Hang rates more valuable and accurate. Thank you all for your patience.

androideveloper:
Thank you for the update @kahest ! Looking forward to it

philipphofmann:
### Beta testers wanted

The app hang rate is available as an experimental feature. To use this feature, you need to 
1. use the experimental [AppHangsV2](https://docs.sentry.io/platforms/apple/guides/ios/configuration/app-hangs/#app-hangs-v2) in the Cocoa SDK, which you can enable via `options.enableAppHangTrackingV2 = true`
[...]

PR #4921, opened 2025-02-28
feat: Send fatal app hang session updates

philipphofmann:
## :scroll: Description

This PR is the first step for adding support for an app hang rate for Apple projects in Sentry. It ends sessions as abnormal and sends the `abnormalMechanism = "anr_foreground"`, which is required for the app hang rate. 

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/4889.

First step of GH-3319 and fixes https://github.com/getsentry/sentry-cocoa/issues/4923.

Unit tests. We still need to enable ANR rates in Sentry for iOS projects so we can correctly test this, but I changed the platform from sentry-cocoa quickly to Android and could successfully see an ANR rate:
CleanShot 2025-03-04 at 08 35 02@2x

As this feature is only enabled when enabling the experimental feature `enableAppHangTrackingV2` it's acceptable to fix issues later if required.

## :pencil: Checklist


Sources/Sentry/SentryHub.m
```diff
@@ -262,6 +262,38 @@ - (void)captureCrashEvent:(SentryEvent *)event withScope:(SentryScope *)scope
...
 }
 
+/**
+ * This method expects an abnormal session already stored to disk. For more info checkout: @c
+ * SentryCrashIntegrationSessionHandler
+ */
+- (void)captureFatalAppHangEvent:(SentryEvent *)event
+{
+    // We tread fatal app hang events similar to crashes.
+    event.isCrashEvent = YES;
```
philipphofmann, line 272: I will refactor this property in an upcoming PR. See https://github.com/getsentry/sentry-cocoa/issues/4935

Sources/Sentry/SentryANRTrackingIntegration.m
```diff
@@ -150,7 +152,7 @@ - (void)anrDetectedWithType:(enum SentryANRType)type
     // We only measure app hang duration for V2.
     // For V1, we directly capture the app hang event.
     if (self.options.enableAppHangTrackingV2) {
-        // We only temporarily store the app hang duration info, so we can change the error message
+        // We only temporarily stoe the app hang duration info, so we can change the error message
```
philprime, line 155:
```suggestion
        // We only temporarily store the app hang duration info, so we can change the error message
```

Sources/Sentry/SentryCrashIntegrationSessionHandler.m
```diff
@@ -36,31 +36,50 @@ - (instancetype)initWithCrashWrapper:(SentryCrashWrapper *)crashWrapper
     return self;
 }
 
-- (void)endCurrentSessionAsCrashedWhenCrashOrWatchdogTermination
+- (void)endCurrentSessionIfRequired
 {
+    SentryFileManager *fileManager = [[[SentrySDK currentHub] getClient] fileManager];
```
philprime, line 41: `l`: Maybe some debug logs could be useful in this method as there are multiple conditions

Sources/Sentry/SentryFileManager.m
```diff
@@ -324,6 +327,21 @@ - (void)deleteCrashedSession
     [self deleteSession:self.crashedSessionFilePath];
 }
 
+- (void)storeAbnormalSession:(SentrySession *)session
```
philprime, line 330: `l`: Should we start marking parameters as `_Nonnull` to reduce nullability issues?

Sources/Sentry/SentryHub.m
```diff
@@ -262,6 +262,42 @@ - (void)captureCrashEvent:(SentryEvent *)event withScope:(SentryScope *)scope
...
+/**
+ * This method expects an abnormal session already stored to disk. For more info checkout: @c
+ * SentryCrashIntegrationSessionHandler
+ */
+- (void)captureFatalAppHangEvent:(SentryEvent *)event
+{
+    // We treat fatal app hang events similar to crashes.
+    event.isCrashEvent = YES;
+
+    SentryClient *client = _client;
```
philprime, line 276: `l`: should we be more explicit with nullability in general and add all `Nonnull` or `Nullable` keywords?

Tests/SentryTests/Helper/SentryFileManagerTests.swift
```diff
@@ -525,6 +526,74 @@ class SentryFileManagerTests: XCTestCase {
...
+    
+    func testAbnormalSessionAsync_DoesNotCrash() {
+        // Arrange
+        let iterations = 1_000
+        let expectation = expectation(description: "complete all abnormal session interactions")
+        expectation.expectedFulfillmentCount = iterations * 3
+        let dispatchQueue = DispatchQueue(label: "testAbnormalSessionAsync_DoesNotCrash", qos: .userInitiated, attributes: [.concurrent])
+        
+        // Act
+        for _ in 0..<iterations {
```
philprime, line 576: `l`: it might be worth to bring in some variance by adding sleep statements (but not with random durations, as test input should not change in between invocations)

Tests/SentryTests/SentryCrash/SentryCrashReportSinkTests.swift
```diff
@@ -40,10 +40,14 @@ class SentryCrashReportSinkTests: SentrySDKIntegrationTestsBase {
...
-                expect.fulfill()
+            do {
+                try self.assertCrashEventWithScope { _, scope in
+                    let data = scope?.serialize()
+                    XCTAssertEqual(data?["environment"] as? String, "testFilterReports_CopyHubScope")
+                    expect.fulfill()
+                }
+            } catch {
+                XCTFail("Failed to assert crash event with scope: \(error)")
             }
```
philprime, line 51: `m`: Please look into using the `rethrows` keyword for the callback, i.e. [here](https://www.avanderlee.com/swift/rethrows/). As the closure is not using `try` inside, it might be possible to drop the `try` outside too, therefore no need to add the do-catch (should also remove the Codecov branch coverage warning)

Tests/SentryTests/SentryCrash/SentryCrashReportSinkTests.swift
```diff
@@ -115,8 +119,12 @@ class SentryCrashReportSinkTests: SentrySDKIntegrationTestsBase {
...
         let report = ["attachments": ["file.png"]]
         fixture.sut.filterReports([report]) { _, _, _ in
-            self.assertCrashEventWithScope { _, scope in
-                XCTAssertEqual(scope?.attachments.count, 1)
+            do {
+                try self.assertCrashEventWithScope { _, scope in
+                    XCTAssertEqual(scope?.attachments.count, 1)
+                }
+            } catch {
+                XCTFail("Failed to assert crash event with scope: \(error)")
```
philprime, line 127: `m`: see above

Tests/SentryTests/SentrySDKIntegrationTestsBase.swift
```diff
@@ -53,15 +53,15 @@ class SentrySDKIntegrationTestsBase: XCTestCase {
         try callback(capture?.event, capture?.scope, capture?.additionalEnvelopeItems)
     }
     
-    func assertCrashEventWithScope(_ callback: (Event?, Scope?) -> Void) {
+    func assertCrashEventWithScope(_ callback: (Event?, Scope?) throws -> Void) throws {
```
philprime, line 56: `m`: See my comment in `Tests/SentryTests/SentryCrash/SentryCrashReportSinkTests.swift` to consider `rethrows` instead of `throws`

philprime (approved):
In general LGTM, but please take a look at my comments and implement at your decision.

Sources/Sentry/SentryFileManager.m
philipphofmann: It's nonnull cause of NS_ASSUME_NONNULL_BEGIN. Thanks for double checking.

Tests/SentryTests/Helper/SentryFileManagerTests.swift
philipphofmann: This test aims to identify potential locking issues that could lead to a crash. While adding sleeps may make the test appear more realistic, it significantly reduces its effectiveness by decreasing the likelihood of read, write, and delete operations overlapping across different threads, thereby making it less likely to uncover concurrency issues.

PR #4930, opened 2025-03-03
chore: Recommendation for test assertions

philipphofmann:
Document how to use test assertions for a more consistent test code base.

I also fixed some headings in this PR so that there is one section for testing. 

#skip-changelog


philprime (approved):
LGTM, thanks for documenting this.

PR #4934, opened 2025-03-04
test: Increase ANR tracker test wait timeout

philipphofmann:
Modify the wait timeout in SentryANRTrackerV2Tests to provide more reliable test execution by extending the timeout from 1.0 to 5.0 seconds.

#skip-changelog


philprime (approved):
LGTM

PR #4938, opened 2025-03-04
refactor(session-replay): add injectable mask and view renderer

philprime:
- Session Replay consists of three phases:
  1. Calculate frames to redact
  2. Take a screenshot of the window using a view renderer
  3. Mask the screenshot at the redact frames
- This PR refactors the view and mask renderer to be individual files and being injectable
- This PR is preparation for a follow-up PR where I'll introduce a more efficient experimental view renderer

#skip-changelog


Sources/Swift/Tools/ViewCapture/SentryViewPhotographer.swift
```diff
@@ -0,0 +1,76 @@
...
+    private let dispatchQueue = SentryDispatchQueueWrapper()
+
+    var renderer: SentryViewRenderer
+
+    init(
+        renderer: SentryViewRenderer,
+        redactOptions: SentryRedactOptions
+    ) {
+        self.renderer = renderer
+        self.maskRenderer = SentryDefaultMaskRenderer()
```
philprime, line 22: The mask renderer uses Swift-only structs, so it can not be used from Objective-C.
Therefore the mask renderer can not be injected but is abstracted using the protocol `SentryMaskRenderer`

philipphofmann (approved):
Thanks for putting this refactoring in a clear purpose PR, so it is very easy to review 💯. LGTM

PR #4939, opened 2025-03-04
fix(ci): Add full path to xcresult for upload artifact

philprime:
The path to the `.xcresult` must be absolute.

#skip-changelog


philipphofmann (approved):
Thanks, LGTM.

Issue #4000, opened 2024-05-21
SessionReplay causes user-visible frame drop


dfed:
### Platform

iOS

### Environment

Production, Develop, TestFlight, Other

### Installed

Swift Package Manager

### Version

8.26.0

### Xcode Version

15.4

### Did it work on previous versions?

8.25.0

### Steps to Reproduce

Set the following options:

```swift
options.experimental.sessionReplay.errorSampleRate = 1.0
options.experimental.sessionReplay.sessionSampleRate = 1.0
options.experimental.sessionReplay.redactAllText = false
options.experimental.sessionReplay.redactAllImages = false
```

### Expected Result

Do not see user-visible frame drops

### Actual Result

See user visible frame drops every second

brustolin:
Thanks @dfed for reaching out. We will investigate this.

dfed:
More info! This issue is happening in both 8.25.1 and 8.25.2, in addition to 8.26.0. 8.25.0 does not have this issue.

brustolin:
I belive I know what cause the change. I dont know whether we can go back. But I will try to find an alternative solution.

kahest:
related: #4235

krystofwoldrich:
- Related to https://github.com/getsentry/sentry-react-native/issues/4499

philprime:
Rough ideas what could be the issue here, needs further investigation:


- `DefaultViewRenderer` uses [`drawHierarchy(in:afterScreenUpdates:)`](https://developer.apple.com/documentation/uikit/uiview/drawhierarchy(in:afterscreenupdates:)) which is slower than [`snapshotview(afterscreenupdates:)`](https://developer.apple.com/documentation/uikit/uiview/snapshotview(afterscreenupdates:)) according to [documentation](https://developer.apple.com/documentation/uikit/uiview/drawhierarchy(in:afterscreenupdates:)#Discussion)
- `DefaultViewRenderer` creates a graphics context, draws the view hierarchy, renders to image. `SentryViewPhotographer` then creates a graphics context, draws the rendered image into the context, draws masking overlays, renders to an image. Maybe the render passes can be reduced.
- Instead of `UIGraphicsImageRenderer` it might be faster to directly use `CGContext`

https://github.com/getsentry/sentry-cocoa/blob/07475f9c7b72f62c9c53e3a3e87c26915cee693e/Sources/Swift/Tools/SentryViewPhotographer.swift#L13-L21

[...]

PR #4940, opened 2025-03-04
feat(session-replay): Add experimental flags to use a more efficient view renderer for Session Replay

philprime:
## :scroll: Description

**TL;DR: Reduces the time required to render a session replay frame from ~160ms to ~30-36ms (benchmarked on an iPhone 8)**

- Implements a custom graphics image renderer.
- Adds a new experimental feature `enableExperimentalViewRenderer` to choose the view renderer (enables our graphics renderer)
- Adds a new experimental feature `enableFastViewRenderer` to choose the view draw method (switches from UIKit to CoreGraphics rendering, might be incomplete)
- Adds an experimental graphics image renderer
- Adds a view renderer using the experimental graphics image renderer
- Adds a mask renderer using the experimental graphics image renderer

> [!NOTE] 
> The mask renderer is using a points-to-pixel scale of 1, therefore any higher resolution view render will be downscaled.

See #4000 for more information for the main issue of frame performance.

To understand the changes, we need to look at the implementation of the `SentryDefaultViewRenderer` used by the `SentryViewPhotographer`:

```swift
@objcMembers
class SentryDefaultViewRenderer: NSObject, SentryViewRenderer {
    func render(view: UIView) -> UIImage {
        // START BLOCK - Setup
        let image = UIGraphicsImageRenderer(size: view.bounds.size).image { context in
            // START BLOCK - Draw
            view.drawHierarchy(in: view.bounds, afterScreenUpdates: false)
            // END BLOCK - Draw
        }
        // END BLOCK - Setup
        return image
    }
}
```

The rendering consists of two nested blocks: 
- `Setup`: to create a rendering context which is then used to convert rendered bitmap data into an in image
- `Draw`: which is drawing the view hierarchy into the bitmap context

Both need to be analysed for their performance and potential improvement candidates.

In addition we need to reconsider how the resolutions and coordinate system works:
- View has a `size` of a `width` and `height` in points (i.e. `375pt x 667pt` for iPhone 8)
- Windows have a `scale` at which the UI should be rendered (i.e. `2.0` for iPhone 8)
- Screens have a resolution of `size * scale` pixels (i.e. `750px x 1334px` for iPhone 8)

Therefore I also analyzed the impact of creating a scaled vs unscaled image, because the difference can be seen visually not only due to changes in resolution, but also due to blurriness:

*Example - iOS-Swift at scale `1.0`*:
scale-1

*Example - iOS-Swift at scale `2.0`*:
scale-2

The `drawHierarchy` performance has also been discussed in the [Apple Developer Forums](https://developer.apple.com/forums/thread/757394), leading to ReplayKit as an alternative (TBD)

I tried to create unit tests to test the view photographer using a complex UI built in a Storyboard file. I was not able to get it working, because unit tests have no render server, therefore trying to render a view to an image fails with the following error:

> Rendering a view (0x105223a80, UILayoutContainerView) that has not been committed to render server is not supported.

Instead I performed manual testing by adapting the `SentryViewPhotographer` to use this method, then looking at the values over time when running the sample `iOS-Swift` on an iPhone 8:

```swift
func image(view: UIView, onComplete: @escaping ScreenshotCallback) {
    let viewSize = view.bounds.size

    let startTime = DispatchTime.now().uptimeNanoseconds
    let startRedactTime = DispatchTime.now().uptimeNanoseconds
    let redact = redactBuilder.redactRegionsFor(view: view)
    let endRedactTime = DispatchTime.now().uptimeNanoseconds

    // The render method is synchronous and must be called on the main thread.
    // This is because the render method accesses the view hierarchy which is managed from the main thread.
    let startRenderTime = DispatchTime.now().uptimeNanoseconds
    let renderedScreenshot = renderer.render(view: view)
    let endRenderTime = DispatchTime.now().uptimeNanoseconds
    let endTime = DispatchTime.now().uptimeNanoseconds

    printSummary(
        redactDiff: endRedactTime - startRedactTime,
        renderDiff: endRenderTime - startRenderTime,
        totalDiff: endTime - startTime
    )

    dispatchQueue.dispatchAsync { [maskRenderer] in
        // The mask renderer does not need to be on the main thread.
        // Moving it to a background thread to avoid blocking the main thread, therefore reducing the performance
        // impact/lag of the user interface.
        let maskedScreenshot = maskRenderer.maskScreenshot(screenshot: renderedScreenshot, size: viewSize, masking: redact)
        onComplete(maskedScreenshot)
    }
}

private let sampleLimit = 120
private var redactDiffHistory = [UInt64]()
private var renderDiffHistory = [UInt64]()
private var totalDiffHistory = [UInt64]()

func printSummary(redactDiff: UInt64, renderDiff: UInt64, totalDiff: UInt64) {
    redactDiffHistory = (redactDiffHistory + [redactDiff]).suffix(sampleLimit)
    renderDiffHistory = (renderDiffHistory + [renderDiff]).suffix(sampleLimit)
    totalDiffHistory = (totalDiffHistory + [totalDiff]).suffix(sampleLimit)

    let redactDiffHistoryMin = redactDiffHistory.min() ?? 0
    let redactDiffHistoryMax = redactDiffHistory.max() ?? 0
    let redactDiffHistoryAverage = Double(redactDiffHistory.reduce(0, +)) / Double(max(redactDiffHistory.count, 1))
    let sortedRedactDiffHistory = redactDiffHistory.sorted()
    let redactDiffHistoryP50 = sortedRedactDiffHistory[redactDiffHistory.count / 2]
    let redactDiffHistoryP75 = sortedRedactDiffHistory[Int(Double(redactDiffHistory.count) * 0.75)]
    let redactDiffHistoryP95 = sortedRedactDiffHistory[Int(Double(redactDiffHistory.count) * 0.95)]
    let redactDiffHistoryLast = redactDiffHistory.last ?? 0

    let renderDiffHistoryMin = renderDiffHistory.min() ?? 0
    let renderDiffHistoryMax = renderDiffHistory.max() ?? 0
    let renderDiffHistoryAverage = Double(renderDiffHistory.reduce(0, +)) / Double(max(renderDiffHistory.count, 1))
    let sortedRenderDiffHistory = renderDiffHistory.sorted()
    let renderDiffHistoryP50 = sortedRenderDiffHistory[renderDiffHistory.count / 2]
    let renderDiffHistoryP75 = sortedRenderDiffHistory[Int(Double(renderDiffHistory.count) * 0.75)]
    let renderDiffHistoryP95 = sortedRenderDiffHistory[Int(Double(renderDiffHistory.count) * 0.95)]
    let renderDiffHistoryLast = renderDiffHistory.last ?? 0

    let totalDiffHistoryMin = totalDiffHistory.min() ?? 0
    let totalDiffHistoryMax = totalDiffHistory.max() ?? 0
    let totalDiffHistoryAverage = Double(totalDiffHistory.reduce(0, +)) / Double(max(totalDiffHistory.count, 1))
    let sortedTotalDiffHistory = totalDiffHistory.sorted()
    let totalDiffHistoryP50 = sortedTotalDiffHistory[totalDiffHistory.count / 2]
    let totalDiffHistoryP75 = sortedTotalDiffHistory[Int(Double(totalDiffHistory.count) * 0.75)]
    let totalDiffHistoryP95 = sortedTotalDiffHistory[Int(Double(totalDiffHistory.count) * 0.95)]
    let totalDiffHistoryLast = totalDiffHistory.last ?? 0
    
    func f(_ value: UInt64) -> String {
        String(format: "%8.4f ms", Double(value) / 1_000_000.0)
    }

    func f(_ value: Double) -> String {
        String(format: "%8.4f ms", Double(value) / 1_000_000.0)
    }

    let samples = String(format: "%4.i Samples", redactDiffHistory.count)
    print("| \(samples) | Redact      | Render      | Total       |")
    print("|--------------|-------------|-------------|-------------|")
    print("| Min          | \(f(redactDiffHistoryMin)) | \(f(renderDiffHistoryMin)) | \(f(totalDiffHistoryMin)) |")
    print("| Max          | \(f(redactDiffHistoryMax)) | \(f(renderDiffHistoryMax)) | \(f(totalDiffHistoryMax)) |")
    print("| Avg          | \(f(redactDiffHistoryAverage)) | \(f(renderDiffHistoryAverage)) | \(f(totalDiffHistoryAverage)) |")
    print("| p50          | \(f(redactDiffHistoryP50)) | \(f(renderDiffHistoryP50)) | \(f(totalDiffHistoryP50)) |")
    print("| p75          | \(f(redactDiffHistoryP75)) | \(f(renderDiffHistoryP75)) | \(f(totalDiffHistoryP75)) |")
    print("| p95          | \(f(redactDiffHistoryP95)) | \(f(renderDiffHistoryP95)) | \(f(totalDiffHistoryP95)) |")
    print("| Last         | \(f(redactDiffHistoryLast)) | \(f(renderDiffHistoryLast)) | \(f(totalDiffHistoryLast)) |")
}
```

### Baseline Performance

These are the results of using the currently released `SentryViewPhotographer.image(view:onComplete:)`. We are not measuring the `maskRenderer.maskScreenshot(...)` because it is run on the background thread and therefore not blocking the main thread.

If a screen uses 60 frames per seconds to render graphics, the maximum time used to render one frame should be less than `1 / 60 = 16.667 ms`. If a frame render takes longer, the next frame will be not be rendered (= a dropped frame) to catch up the lost time. This results in a stuttering user interface and visible to the user.

**Results at screen scale (i.e. `2.0` for iPhone 8):**

```swift
let image = UIGraphicsImageRenderer(size: view.bounds.size).image { _ in
    view.drawHierarchy(in: view.bounds, afterScreenUpdates: false)
}
```

The data shows that we have a significant frame delay with up to 161.5842ms = ~9 frames dropped every second.

|  120 Samples | Redact      | Render      | Total       |
|--------------|-------------|-------------|-------------|
| Min          |   3.0583 ms | 145.3815 ms | 151.4525 ms |
| Max          |   6.5138 ms | 155.8338 ms | 161.8351 ms |
| Avg          |   5.8453 ms | 149.8243 ms | 155.6732 ms |
| p50          |   6.0484 ms | 149.2103 ms | 154.1397 ms |
| p75          |   6.1136 ms | 151.9487 ms | 158.0255 ms |
| p95          |   6.2567 ms | 155.3496 ms | 161.3549 ms |
| Last         |   6.1190 ms | 150.4965 ms | 156.6198 ms |

**Results at scale `1.0`:**

```swift
let image = UIGraphicsImageRenderer(size: view.bounds.size, format: .init(for: .init(displayScale: 1))).image { _ in
    view.drawHierarchy(in: view.bounds, afterScreenUpdates: false)
}
```

No significant changes compared to the native screen scale.

|  120 Samples | Redact      | Render      | Total       |
|--------------|-------------|-------------|-------------|
| Min          |   5.9785 ms | 146.0440 ms | 152.2056 ms |
| Max          |   6.4990 ms | 156.3045 ms | 162.4112 ms |
| Avg          |   6.1184 ms | 149.6858 ms | 155.8079 ms |
| p50          |   6.1055 ms | 148.6769 ms | 155.1323 ms |
| p75          |   6.1640 ms | 151.9197 ms | 158.0497 ms |
| p95          |   6.3104 ms | 155.7726 ms | 161.8109 ms |
| Last         |   5.9875 ms | 147.1455 ms | 153.1360 ms |

#### Using `UIView.snapshotView(afterScreenUpdates:)`

The first attempt to optimize the performance was using the method [UIView.snapshotView(afterScreenUpdates:)](https://developer.apple.com/documentation/uikit/uiview/snapshotview(afterscreenupdates:)) to create a dummy view hierarchy as stated in the documentation, to see if the dummy view would render faster than the normal one:

 > This method very efficiently captures the current rendered appearance of a view and uses it to build a new snapshot view. You can use the returned view as a visual stand-in for the current view in your app. [..] Because **the content is captured from the already rendered content, this method reflects the current visual appearance of the view** and is not updated to reflect animations that are scheduled or in progress. However, calling this method is faster than trying to render the contents of the current view into a bitmap image yourself.

During testing the snapshot view never displayed any visual data when rendered into an `UIImage`.

An Apple Engineer mentioned in the [Apple Developer Forum](https://developer.apple.com/forums/thread/108278?answerId=613818022#613818022) that a snapshot **can not** be rendered to an image due to security reasons.

**Results:**

Discarded this approach because not possible.

#### Reuse the [UIGraphicsImageRenderer](https://developer.apple.com/documentation/uikit/uigraphicsimagerenderer)

The documentation mentions that this renderer uses a built-in cache, therefore recommends to reuse renderer instances:

> After initializing an image renderer, you can use it to draw multiple images with the same configuration. An image renderer keeps a cache of Core Graphics contexts, so reusing the same renderer can be more efficient than creating new renderers.

But the `UIGraphicsImageRenderer` is set to a fixed size and in our case the size of the view could eventually change. 

To reduce the memory footprint I decided to have a maximum of one cached renderer, discarding and re-creating it whenever the size changes. In the worst case it has to re-create the renderer every time, yielding the same performance as without the cache. In the best case the size never changes and the renderer is reused forever.

**Results:**

There is no significant change compared to the baseline.

|  120 Samples | Redact      | Render      | Total       |
|--------------|-------------|-------------|-------------|
| Min          |   5.9499 ms | 146.9310 ms | 153.0000 ms |
| Max          |   6.7538 ms | 156.0019 ms | 161.9819 ms |
| Avg          |   6.0744 ms | 149.5189 ms | 155.5966 ms |
| p50          |   6.0541 ms | 148.0545 ms | 154.0874 ms |
| p75          |   6.1128 ms | 151.6945 ms | 157.7179 ms |
| p95          |   6.2532 ms | 155.3220 ms | 161.3656 ms |
| Last         |   6.0766 ms | 147.0542 ms | 153.1340 ms |

#### Experimental View Renderer

UIKit is built on top of [CoreGraphics](https://developer.apple.com/documentation/coregraphics) and [CoreAnimation](https://developer.apple.com/documentation/QuartzCore) also known as the `QuartzCore`.

The `UIGraphicsImageRenderer` has been introduced with `iOS 10.0` to wrap around the `CGContext` provided by CoreGraphics, as the setup can be tedious and complicated.

The `SentryGraphicsImageRenderer` is a custom implementation creating a `CGContext` pixel buffer and also converting it into an `UIImage` afterwards.

**Results at native scale (i.e. `2.0`):**

```swift
let scale = (view as? UIWindow ?? view.window)?.screen.scale ?? 1
let image = SentryGraphicsImageRenderer(size: view.bounds.size, scale: scale).image { context in
    view.drawHierarchy(in: view.bounds, afterScreenUpdates: false)
}
```

- Significant decrease in render time compared to base line performance.

|  120 Samples | Redact      | Render      | Total       |
|--------------|-------------|-------------|-------------|
| Min          |   2.1284 ms |  14.7570 ms |  16.8885 ms |
| Max          |   6.7335 ms |  32.5775 ms |  38.6342 ms |
| Avg          |   6.0055 ms |  25.4156 ms |  31.4264 ms |
| p50          |   6.0432 ms |  24.5575 ms |  30.6436 ms |
| p75          |   6.0950 ms |  27.3424 ms |  33.3627 ms |
| p95          |   6.2285 ms |  30.3218 ms |  36.5009 ms |
| Last         |   6.0572 ms |  24.3487 ms |  30.4101 ms |

**Results at `1.0` scale:**

```swift
let scale = 1.0
let image = SentryGraphicsImageRenderer(size: view.bounds.size, scale: scale).image { context in
    view.drawHierarchy(in: view.bounds, afterScreenUpdates: false)
}
```

- Significant decrease in render time compared to base line performance.
- Slower than native scale, probably due to downscaling of internal graphics data

|  120 Samples | Redact      | Render      | Total       |
|--------------|-------------|-------------|-------------|
| Min          |   2.1135 ms |  27.0460 ms |  29.1636 ms |
| Max          |   7.2708 ms |  48.6616 ms |  54.7815 ms |
| Avg          |   6.0159 ms |  38.8028 ms |  44.8226 ms |
| p50          |   6.1011 ms |  38.4735 ms |  44.7876 ms |
| p75          |   6.2221 ms |  40.3663 ms |  46.3873 ms |
| p95          |   6.4524 ms |  44.4222 ms |  50.5087 ms |
| Last         |   6.1820 ms |  42.6977 ms |  48.8843 ms |

#### Replacing `view.drawHierarchy(in:afterScreenUpdates:)` with `view.layer.render(in:)`

Instead of drawing the view using the `drawHierarchy` provided on the UIKit-level, we can directly call the rendering of the `layer` used by the UIView provided by CoreGraphics.

> [!WARNING] 
> During testing we noticed that the render is incomplete, in particular the tab bar icons have not been rendered. The exact impact is not known and can vary.
> 
> image_5

**Results at native scale (i.e. `2.0`):**

```swift
let image = UIGraphicsImageRenderer(size: view.bounds.size).image { context in
    view.layer.render(in: context.cgContext)
}
```

- Significant decrease in render time compared to the base line performance. 
- Faster than the `SentryGraphicsImageRenderer` + `drawHierarchy`

|  120 Samples | Redact      | Render      | Total       |
|--------------|-------------|-------------|-------------|
| Min          |   2.0926 ms |  13.1221 ms |  15.2177 ms |
| Max          |   6.3896 ms |  24.6149 ms |  30.6909 ms |
| Avg          |   6.0791 ms |  20.5175 ms |  26.6001 ms |
| p50          |   6.0992 ms |  19.8446 ms |  25.9533 ms |
| p75          |   6.1418 ms |  21.5083 ms |  27.6816 ms |
| p95          |   6.3071 ms |  23.0865 ms |  29.1678 ms |
| Last         |   6.0721 ms |  24.6149 ms |  30.6909 ms |

**Results at `1.0` scale:**

```swift
let image = UIGraphicsImageRenderer(size: view.bounds.size, format: .init(for: .init(displayScale: 1))).image { context in
    view.layer.render(in: context.cgContext)
}
```

- Slightly slower than native scale, probably due to downscaling of internal graphics data

|  120 Samples | Redact      | Render      | Total       |
|--------------|-------------|-------------|-------------|
| Min          |   2.6087 ms |  12.5871 ms |  15.1984 ms |
| Max          |   6.7715 ms |  26.2617 ms |  32.3283 ms |
| Avg          |   6.0303 ms |  22.2413 ms |  28.2754 ms |
| p50          |   6.0662 ms |  21.3212 ms |  27.3122 ms |
| p75          |   6.1138 ms |  24.0657 ms |  30.1942 ms |
| p95          |   6.3048 ms |  26.2106 ms |  32.2714 ms |
| Last         |   6.7715 ms |  20.5220 ms |  27.2970 ms |

#### Experimental Renderer + `view.layer.render(in:)`

Combining the previous two approaches.

> [!WARNING] 
> During testing we noticed that the render is incomplete, in particular the tab bar icons have not been rendered. The exact impact is not known and can vary.

**Results at native scale (i.e. `2.0`):**

```swift
let scale = (view as? UIWindow ?? view.window)?.screen.scale ?? 1
let image = SentryGraphicsImageRenderer(size: view.bounds.size, scale: scale).image { context in
    view.layer.render(in: context.cgContext)
}
```

- Significant decrease in render time compared to the base line performance. 
- Large improvements compared to experimental renderer with `view.drawHierarchy(...)`
- No improvements compared to `UIGraphicsImageRenderer` + `view.layer.render(in:)`

|  120 Samples | Redact      | Render      | Total       |
|--------------|-------------|-------------|-------------|
| Min          |   5.9651 ms |  18.5304 ms |  24.5483 ms |
| Max          |   6.5470 ms |  24.9161 ms |  30.9267 ms |
| Avg          |   6.0764 ms |  20.7397 ms |  26.8196 ms |
| p50          |   6.0500 ms |  19.8425 ms |  26.0733 ms |
| p75          |   6.1109 ms |  22.4187 ms |  28.5533 ms |
| p95          |   6.3331 ms |  24.6614 ms |  30.6779 ms |
| Last         |   6.0334 ms |  18.7154 ms |  24.7522 ms |

**Results at `1.0` scale:**

```swift
let scale = 1.0
let image = SentryGraphicsImageRenderer(size: view.bounds.size, scale: scale).image { context in
    view.layer.render(in: context.cgContext)
}
```

- Slower than native scale, probably due to downscaling of internal graphics data

|  120 Samples | Redact      | Render      | Total       |
|--------------|-------------|-------------|-------------|
| Min          |   5.9296 ms |  19.2106 ms |  25.2905 ms |
| Max          |   6.2822 ms |  27.0288 ms |  33.0322 ms |
| Avg          |   6.0384 ms |  21.6527 ms |  27.6942 ms |
| p50          |   6.0178 ms |  20.7255 ms |  26.7658 ms |
| p75          |   6.0690 ms |  22.8727 ms |  28.9003 ms |
| p95          |   6.2106 ms |  26.0749 ms |  32.0846 ms |
| Last         |   6.1801 ms |  23.6582 ms |  29.8419 ms |

### Detailed Analysis of Implementation

Looking at one of the samples using `drawViewHierarchyInRect` in detail we can analyze the duration of the different calls in detail. In particular we notice that the most expensive calls are inside UIKit and CoreGraphics.

```
26.00 ms  100,0 % Sentry            |  SentryGraphicsImageRenderer.image(actions:)
25.00 ms   96,2 % Sentry            |   closure #1 in SentryExperimentalViewRenderer.render(view:)
25.00 ms   96,2 % UIKitCore         |    -[UIView drawViewHierarchyInRect:afterScreenUpdates:]
21.00 ms   80,8 % UIKitCore         |     -[UIImage drawInRect:blendMode:alpha:]
21.00 ms   80,8 % CoreGraphics      |      CGContextDrawImageWithOptions
20.00 ms   76,9 % CoreGraphics      |       ripc_DrawImage
15.00 ms   57,7 % CoreGraphics      |        ripc_AcquireRIPImageData
 5.00 ms   19,2 % CoreGraphics      |        RIPLayerBltImage
 1.00 ms    3,8 % libdispatch.dylib |       _dispatch_once_callout
 4.00 ms   15,4 % UIKitCore         |     _UIRenderViewImageAfterCommit
 1.00 ms    3,8 % Sentry            |   SentryGraphicsImageRenderer.Context.currentImage.getter
 1.00 ms    3,8 % CoreGraphics      |    CGBitmapContextCreateImage
```

It seems like `UIView` is backed by an `UIImage` from a previous render, then draws it again into the bitmap context. As this is is a private API, we can not access it directly. No further optimization possible.


Sources/Swift/Integrations/SessionReplay/Preview/SentryMaskingPreviewView.swift
```diff
@@ -27,7 +27,8 @@ class SentryMaskingPreviewView: UIView {
     init(redactOptions: SentryRedactOptions) {
         self.photographer = SentryViewPhotographer(
             renderer: PreviewRenderer(),
-            redactOptions: redactOptions
+            redactOptions: redactOptions,
+            enableExperimentalMasking: false
```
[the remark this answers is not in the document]
philipphofmann: We have access to the redactOptions which stem from the SentryOptions. So in the `SentrySessionReplayIntegration` we should have access to the SentryOptions and then this flag.

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -19,6 +19,30 @@ public class SentryExperimentalOptions: NSObject {
...
+    /**
+     * Enables the experimental view renderer used by the Session Replay integration.
+     *
+     * Rendering the view hierarchy is an expensive operation and can impact the performance of your app.
+     * The experimental view renderer is optimized for performance, but might not render all views correctly.
+     *
+     * - Experiment: This is an experimental feature and is therefore disabled by default. In case you are noticing issues with the experimental
+     *               view renderer, please report the issue on [GitHub](https://github.com/getsentry/sentry-cocoa).
+     */
+    public var enableExperimentalViewRenderer = false
```
philipphofmann, line 31: `m`: Shouldn't these rather be moved to `SentryReplayOptions` cause they are related to SR?

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -19,6 +19,30 @@ public class SentryExperimentalOptions: NSObject {
...
+     *
+     * This flag can be used to switch between drawing the view hierarchy using `CALayer.render(in:)` or `UIView.drawHierarchy(in:afterScreenUpdates:)`.
+     *
+     * - Warning: Rendering the view hiearchy using the `CALayer.render(in:)` method can lead to rendering issues, especially when using custom views.
+     *              For complete rendering, it is recommended to set this option to `false`. In case you prefer performance over completeness, you can
+     *              set this option to `true`.
+     * - Experiment: This is an experimental feature and is therefore disabled by default. In case you are noticing issues with the experimental
+     *               view renderer, please report the issue on [GitHub](https://github.com/getsentry/sentry-cocoa).
+     */
+    public var enableFastViewRenderer = false
```
philipphofmann, line 44: `m`: Why do we need two options? Why not enable the fastViewRenderer also with `enableExperimentalViewRenderer` or vice versa? What's the use case for only enabling one of the two introduced options?

Sources/Swift/Tools/ViewCapture/SentryExperimentalMaskRenderer.swift
```diff
@@ -0,0 +1,71 @@
+#if canImport(UIKit) && !SENTRY_NO_UIKIT
+#if os(iOS) || os(tvOS)
+
+import UIKit
+
+class SentryExperimentalMaskRenderer: NSObject, SentryMaskRenderer {
+    func maskScreenshot(screenshot image: UIImage, size: CGSize, masking: [RedactRegion]) -> UIImage {
+        let image = SentryGraphicsImageRenderer(size: size, scale: 1).image { context in
```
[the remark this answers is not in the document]
philipphofmann: Please add this info as a comment.

Sources/Swift/Tools/ViewCapture/SentryExperimentalMaskRenderer.swift
```diff
@@ -0,0 +1,71 @@
+#if canImport(UIKit) && !SENTRY_NO_UIKIT
+#if os(iOS) || os(tvOS)
+
+import UIKit
+
+class SentryExperimentalMaskRenderer: NSObject, SentryMaskRenderer {
```
philipphofmann, line 6: `m`: Isn't this class almost a duplicate of `SentryDefaultMaskRenderer`? The only difference at a first glance is that this class uses the `SentryGraphicsImageRenderer`. If yes, can't we simply pass in the `SentryGraphicsImageRenderer` instead of the `UIGraphicsImageRenderer` and ditch the duplication?

Sources/Swift/Tools/ViewCapture/SentryGraphicsImageRenderer.swift
```diff
@@ -0,0 +1,88 @@
+#if canImport(UIKit) && !SENTRY_NO_UIKIT
+#if os(iOS) || os(tvOS)
+
+import UIKit
+
+/**
+ * Class similar to the ``UIKit/UIGraphicsImageRenderer`` class, but optimized for Sentry.
+ *
+ * We introduced this class, because the ``UIGraphicsImageRenderer`` caused performance issues due to internal caching mechanisms.
+ * During testing we noticed a significant performance improvement by creating the bitmap context directly.
```
philipphofmann, line 10: `m`: Please elaborate on what exactly the internal caching mechanisms are. Are they iOS specific or do they live in our SDK?

Sources/Swift/Tools/ViewCapture/SentryViewPhotographer.swift
```diff
@@ -16,25 +16,31 @@ class SentryViewPhotographer: NSObject, SentryViewScreenshotProvider {
...
         let viewSize = view.bounds.size
         let redact = redactBuilder.redactRegionsFor(view: view)
+        // The render method is synchronous and must be called on the main thread.
+        // This is because the render method accesses the view hierarchy which is managed from the main thread.
         let renderedScreenshot = renderer.render(view: view)
 
         dispatchQueue.dispatchAsync { [maskRenderer] in
+            // The mask renderer does not need to be on the main thread.
+            // Moving it to a background thread to avoid blocking the main thread, therefore reducing the performance
+            // impact/lag of the user interface.
```
philipphofmann, line 38: Thanks for the comment.

Sources/Swift/Tools/ViewCapture/SentryViewPhotographer.swift
```diff
@@ -16,25 +16,31 @@ class SentryViewPhotographer: NSObject, SentryViewScreenshotProvider {
...
             let maskedScreenshot = maskRenderer.maskScreenshot(screenshot: renderedScreenshot, size: viewSize, masking: redact)
             onComplete(maskedScreenshot)
         }
     }
-    
+
     func image(view: UIView) -> UIImage {
```
philipphofmann, line 44: `m`: Can't we also move masking the screenshot here to a bg thread? Currenlty this is only called from capturing a screenshot, so maybe adding a comment explaining why we don't move it to a bg thread here would be helpful.

philipphofmann (commented):
This looks great. I think it's OK to skip automated test for now. Once we know the solution is working correctly, we can come up with proper tests to avoid regressions when changing the code.

Sources/Swift/SentryExperimentalOptions.swift
philprime: I considered introducing only a single feature flag to reduce the amount of configuration options a customer needs to consider. There is a major difference between the two flags:

- `enableExperimentalViewRenderer` uses our custom graphics renderer (I will rename the flag to `enableCustomViewRenderer`) instead of the one provided by UIKit, which should act the same and produce the same output, but be faster. The flag is intended so that customers can opt-in and provide real-world feedback on its stability. Eventually we will want to switch from the default view renderer to the custom one.
- `enableFastViewRenderer` has known issues as it does not fully render the UIKit view hierarchy, but instead renders the underlying CoreGraphics/CoreAnimation layers. This has performance advantages, but my research has shown that the rendered UI is incomplete. This feature will most likely stay off, even after experimental state, due to this issue. But we still want to provide customers an option who prefer faster performance rather than complete UI renders, as it can still provide value to see partial screens. I decided against changing this flag based on some heuristics to leave the decision when to use it to the customer.

I also updated the PR description.

Sources/Swift/Tools/ViewCapture/SentryExperimentalMaskRenderer.swift
philprime: At this point in time, they are mostly equal while using a different renderer. I did consider merging it and only switch the renderer, so we can do that.

But I do not want to change the default implementation which is GA now and make the mask renderer a replaceable/injectable component.

I will re-evaluate to DRY

Sources/Swift/SentryExperimentalOptions.swift
philprime: I renamed `enableFastViewRenderer` to `enableFastViewRendering` so it doesn't seem like an alternative to to the experimental view renderer but an additional benefit.

Sources/Swift/Tools/ViewCapture/SentryGraphicsImageRenderer.swift
philprime: I have added a declaration in the PR description at "Reuse the UIGraphicsImageRenderer", but to summarize, it's stated in the Apple Documentation in the last paragraph of the _Overview_ here:

https://developer.apple.com/documentation/uikit/uigraphicsimagerenderer

Sources/Swift/Tools/ViewCapture/SentryViewPhotographer.swift
philprime: Most likely yes, but this PR focuses on session replay and refactoring this method to be async too (= dropping it in favor of the other one), would require refactoring in the screenshot integration which is out of scope for this PR

Sources/Swift/Tools/ViewCapture/SentryGraphicsImageRenderer.swift
philprime: I'll add a reference to the code docs

Sources/Swift/Integrations/SessionReplay/Preview/SentryMaskingPreviewView.swift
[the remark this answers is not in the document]
philprime: I changed it to do that but I am not sure if this is the way to go. The SentryRedactOptions is actually a public protocol used to expose options to `PreviewRedactOptions`. Furthermore I noticed null exceptions due to the conditional initialization of the session replay options:

https://github.com/getsentry/sentry-cocoa/blob/c964075f766455bae09be6d409d8823eb75a62ed/Sources/Sentry/SentryOptions.m#L143-L145

philipphofmann (commented):
I don't have enough insights yet into SR to clearly answer the following question: The `SentryViewPhotographer.image` already dispatches masking to the background. Can't we dispatch more heavy work of capturing the screenshot to a BG thread?

Sources/Swift/Tools/ViewCapture/SentryExperimentalMaskRenderer.swift
```diff
@@ -0,0 +1,20 @@
...
+
+import UIKit
+
+class SentryExperimentalMaskRenderer: SentryDefaultMaskRenderer {
+    override func maskScreenshot(screenshot image: UIImage, size: CGSize, masking: [RedactRegion]) -> UIImage {
+        // The `SentryDefaultMaskRenderer` is also using an display scale of 1, therefore we also use 1 here.
+        // This could be evaluated in future iterations to view performance impact vs quality.
+        let image = SentryGraphicsImageRenderer(size: size, scale: 1).image { context in
+            applyMasking(to: context, image: image, size: size, masking: masking)
+        }
```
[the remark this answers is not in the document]
philprime: The `SentryViewPhotographer` already dispatches this to the background where possible

Sources/Swift/Tools/ViewCapture/SentryGraphicsImageRenderer.swift
```diff
@@ -0,0 +1,93 @@
...
+        let bitsPerComponent = 8 // 8 bits for each of RGB component
+
+        // Allocate memory for raw image data and initializes every byte in the allocated memory to 0.
+        guard let rawData = calloc(pixelsPerColumn * bytesPerRow, MemoryLayout<UInt8>.size) else {
+            SentryLog.error("Unable to allocate memory for image data")
+            return UIImage()
+        }
+        defer {
+            free(rawData) // Release the memory when done
+        }
```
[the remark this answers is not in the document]
philprime: That's a good idea, but could also cause fragments of previous renders to leak into new ones if not every pixel is overdrawn. I would have to add a clear/reset of all values to 0 too. Not sure if that is more heavy than a reallocation.

philprime:
@philipphofmann the capturing of the screenshot consists of three steps:

1. calculate the redaction frames from the view hierarchy
2. render the current view hierarchy to a bitmap context
3. draw the redaction frames into the bitmap context to hide sensible data.

Step 1 and 2 must be run on main thread because all access to the view hierarchy must be on main. Step 3 can be performed in the background (as it is for session replay, not for screenshots), because it only uses the output of the previous two steps.

Sources/Swift/Tools/ViewCapture/SentryGraphicsImageRenderer.swift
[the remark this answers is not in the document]
philprime: [This StackOverflow](https://stackoverflow.com/a/2688522) posts explains that we should use `calloc` multiple time, because the Kernel scrubs the data for us and is more efficient than using `memset` to reset it.

Therefore I see two options:
1. Reuse the buffer and not care about previous render passes (assuming they are always the same size anyways)
2. Keep it as is.

I am for 2.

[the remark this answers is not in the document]
philipphofmann: Option 1 doesn't work, as we could run into weird problems. We would need to create a new instance of the SentryGraphicsImageRenderer whenever the size changes. We could make this work, but we would need to refactor the code for a bit. Let's keep it as it, and come back to it when required.

philipphofmann (approved):
LGTM, as this is experimental. Once we know the approaches are working, we should think about how we can test this. Please also double check CI before merging.

Sources/Swift/Tools/ViewCapture/SentryExperimentalMaskRenderer.swift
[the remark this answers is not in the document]
philprime: Not sure where the misunderstanding lies. The SentryViewPhotographer has three steps
1. redact, 2. render, 3. mask, with step 1 and 2 being done on the main thread while 3 is dispatched to a background thread.

Step 3 then uses this method to create a masked image, so we are already on the background thread when this method is called, therefore I do not see a point in dispatching it here again.

Sources/Swift/Tools/ViewCapture/SentryGraphicsImageRenderer.swift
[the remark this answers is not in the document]
philprime: Re-creating the renderer did not show any significant improvement for the `UIImageGraphicsRenderer`, so maybe it also won't have any impact for our renderer (assumption that needs testing!).

Leaving it as is.

philprime:
Ready to merge after #4968

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add extension for `Data` to track file I/O operations with Sentry (#4862)
 - Send fatal app hang session updates (#4921) only when enabling the option `enableAppHangTrackingV2`.
+- Add experimental flag `options.sessionReplay.enableExperimentalViewRenderer` to enable up to 5x times more performance in Session Replay (#4940)
```
philipphofmann, line 9: 🥇

philipphofmann (approved):
Let's ship this

brustolin:
Hey, this is amazing!
I’m a little late to the party, but here’s a small contribution:

“layer.render” was the first strategy used in the POV, and it seems fine at first with the Sentry sample, which is a small project. However, after trying it in other projects like Pocketcast or my own app, its performance was worse than `drawViewHierarchy`.

So I suggest double-checking this—maybe it won’t be an issue with the new GraphicRenderer.

P.S.:  It’s missing some elements with “layer.render” because it doesn’t render CoreGraphics filters, so no blur, shadow, tinting (which I believe is related to the tabbar issue), and others.

brustolin:
So, I was super curious about this one and did the checking already with my Bible app:

Running with `drawViewHierarchy`:

0.0411989688873291 seconds
0.04057502746582031 seconds
0.039093971252441406 seconds
0.04010593891143799 seconds
0.03964495658874512 seconds
0.039234042167663574 seconds
0.04171907901763916 seconds
0.037222981452941895 seconds
0.03608500957489014 seconds
0.03732001781463623 seconds
0.040248990058898926 seconds
0.040567994117736816 seconds
0.038839101791381836 seconds
0.037380099296569824 seconds
0.03842198848724365 seconds
0.038452982902526855 seconds

Running with `layer.render(in:)`:

0.05045497417449951 seconds
0.048934102058410645 seconds
0.04951000213623047 seconds
0.05310797691345215 seconds
0.04890894889831543 seconds
0.04867994785308838 seconds
0.048406004905700684 seconds
0.050428032875061035 seconds
0.054461002349853516 seconds
0.04904603958129883 seconds
0.04909801483154297 seconds
0.05078601837158203 seconds
0.048236966133117676 seconds
0.050390005111694336 seconds
0.04842102527618408 seconds
0.04951298236846924 seconds
0.04840099811553955 seconds
0.047763943672180176 seconds
0.04880094528198242 seconds

Old Approach with `UIGraphicsImageRenderer` and `drawViewHierarchy`:
0.04103994369506836 seconds
0.04091799259185791 seconds
0.04353797435760498 seconds
0.040132999420166016 seconds
0.03941702842712402 seconds
0.04117608070373535 seconds
0.040914058685302734 seconds
0.04170703887939453 seconds
0.04350292682647705 seconds
0.04094207286834717 seconds
0.040766000747680664 seconds
0.04295706748962402 seconds
0.04352700710296631 seconds
0.04115605354309082 seconds
0.043032050132751465 seconds
0.04268503189086914 seconds
0.04489099979400635 seconds

The performance seems to be a little bit in favor of new SentryGraphicsRenderer and `drawViewHierarchy` (that also yields a better result), but the difference from the old `UIGraphicsImageRenderer` it's negligible.

Tested with iPhone 14 Pro Max
Each line is a call to "image(actions:)"

kahest:
@brustolin hey thanks for chiming in and running the tests :) the perf improvements are definitely more relevant on lower-end devices, in iPhone 14 Pro Max it's expected to have less of an impact 👍

brustolin:
@kahest Hey, its a pleasure to participate.

 > the perf improvements are definitely more relevant on lower-end devices
 
Yeah, I can imagine that. My only take here is to stick to `drawViewHierarchy`, better performance on complex View and better results on image fidelity.

PR #4941, opened 2025-03-05
fix: Correctly finish TTFD span when no new frame

philipphofmann:
## :scroll: Description

This PR fixes wrongly finishing the TTFD span as the deadline exceeded when the user calls finishSpansIfNotFinished, and the UIViewControllerPerformanceTracker calls finishSpansIfNotFinished before a new frame gets drawn. This happened for a customer using a SplashScreenUIViewController. Now, the TimeToDisplayTracker finishes the TTFD span in finishSpansIfNotFinished when the user called finishSpansIfNotFinished.

A customer reported this bug via Slack.

Unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryTimeToDisplayTracker.m
+++ b/Sources/Sentry/SentryTimeToDisplayTracker.m
@@ -140,6 +140,8 @@
 
 - (void)finishSpansIfNotFinished
 {
+    [SentryDependencyContainer.sharedInstance.framesTracker removeListener:self];
+
     if (self.initialDisplaySpan.isFinished == NO) {
         [self.initialDisplaySpan finish];
     }
@@ -160,8 +162,6 @@
 
         [self.fullDisplaySpan finishWithStatus:kSentrySpanStatusDeadlineExceeded];
     }
-
-    [SentryDependencyContainer.sharedInstance.framesTracker removeListener:self];
 }
 
 - (void)framesTrackerHasNewFrame:(NSDate *)newFrameDate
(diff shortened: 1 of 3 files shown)


Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -86,9 +86,9 @@ - (BOOL)startForTracer:(SentryTracer *)tracer
     [tracer setShouldIgnoreWaitForChildrenCallback:^(id<SentrySpan> span) {
         if ([span.origin isEqualToString:SentryTraceOriginAutoUITimeToDisplay]) {
             return YES;
-        } else {
-            return NO;
         }
+
+        return NO;
```
philprime, line 91: Thanks for this little change making the early-exit cleaner 💯

Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -143,12 +143,23 @@ - (void)finishSpansIfNotFinished
     }
 
     if (self.fullDisplaySpan.isFinished == NO) {
+        if (_fullyDisplayedReported) {
+            SENTRY_LOG_DEBUG(
+                @"SentrySDK.reportFullyDisplayed() was called but didn't receive a new frame to "
+                @"finish the TTFD span. Finishing the full display span so the SDK can start a new "
+                @"time to display tracker.");
+            [self.fullDisplaySpan finish];
```
philprime, line 151: `m`: Not sure if I have enough insights into TTID/TTFD and frame tracking, but shouldn't we also remove the listener here as we do in L162?

philprime (approved):
LGTM, left comments but probably fine to continue

Sources/Sentry/SentryTimeToDisplayTracker.m
philipphofmann: Good catch yes. I will add this at the top of the method.

PR #4945, opened 2025-03-05
chore: Add changelog not about IP address backfill cutoff

kahest:
This PR adds a note to the changelog that starting with version `8.47.0` of the SDK, storage of user's ip address requires `sendDefaultPii=true` because Relay removes backfilling empty ip address to `{{auto}}` with that version: https://github.com/getsentry/relay/pull/4519

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Behavioral changes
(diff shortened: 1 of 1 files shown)


kahest:
changing to draft since we need to postpone this on the Relay side

PR #4951, opened 2025-03-06
tests: Add test isolation and cleanup to Data tracing tests

philprime:
- Adds cleanup of stale files
- Adds pre-conditions for tests

#skip-changelog


Tests/SentryTests/Integrations/Performance/IO/DataSentryTracingIntegrationTests.swift
```diff
@@ -87,6 +102,25 @@ class DataSentryTracingIntegrationTests: XCTestCase {
...
+            }
+            if fileUrlToWrite != nil && manager.fileExists(atPath: fileUrlToWrite.path) {
+                try manager.removeItem(at: fileUrlToWrite)
+            }
+            if ignoredFileUrlToRead != nil && manager.fileExists(atPath: ignoredFileUrlToRead.path) {
+                try manager.removeItem(at: ignoredFileUrlToRead)
+            }
+            if ignoredFileUrlToWrite != nil && manager.fileExists(atPath: ignoredFileUrlToWrite.path) {
+                try manager.removeItem(at: ignoredFileUrlToWrite)
+            }
```
philipphofmann, line 122: `m`: Why do you check if the file exists before removing it? Can't we simply just remove it? That would be less code.

philipphofmann (approved):
LGTM

Tests/SentryTests/Integrations/Performance/IO/DataSentryTracingIntegrationTests.swift
philprime: Because it throws with "file not found"

philipphofmann: Oh my 🤦. It's late sorry. I thought it would simply return false. Ignore me.

philprime:
Looking at the failing tests it seems that some tests are now failing because the `basePath` and `sentryPath` are the same and therefore trigger the ignored file case.

Using the `TestConstants.dsnAsString(username: testName)` to create a unique folder has been a great attempt, but it's simply causing too many issues with ignored file paths.

PR #4955, opened 2025-03-07
fix: profiling sampling only noncontinuous

armcknight:
I noticed this working on another branch, we were computing trace profile sample rates even when unnecessary due to continuous profiling being used.

#skip-changelog


philprime (approved):
LGTM

philipphofmann (approved):
LGTM

PR #4961, opened 2025-03-07
test: disallow starting continuous profiler if SDK is closed

armcknight:
It already worked that way, but we didn't have a test explicitly for it.

#skip-changelog

PR #4963, opened 2025-03-08
format: remove all leading whitespace on empty lines

armcknight:
I got tripped up because somehow a ton of these lines got edited in my Xcode while I prepared #4961 and is still causing me pain in #4855. i decided it'd just be easiest to wipe them out across the codebase and add the lint rule to prevent them from creeping back in.

I recommend enabling the trim whitespace-only lines option in Xcode or equivalent IDE (it wasn't previously enabled for me so I don't know why they got trimmed in the first place):

image

#skip-changelog


philprime (approved):
LGTM, thanks for doing this! It also started to bother me.

As the amount of files is massive, I used the "Hide whitespace" feature of GitHub review to show only non-whitespace changes.

image

philipphofmann (approved):
LGTM, thanks 🙏

Issue #4937, opened 2025-03-04
App Hangs: Ensure scope isn't applied twice for fatal app hangs


philipphofmann:
### Description

Ensure the client doesn't apply the scope again when capturing fatal app hang events. The client applies the scope, but it also does this for crash events. The `applyToEvent` has some mechanisms to avoid overriding data, but maybe that's not sufficient.

https://github.com/getsentry/sentry-cocoa/blob/89acc847f14a330e7fe560a4ccc02f5f47a57742/Sources/Sentry/SentryClient.m#L760

PR #4969, opened 2025-03-11
fix: Wrongly applying scope to crash events

philipphofmann:
## :scroll: Description

The client wrongly applied the scope to crash events, which wasn't a problem in most cases because applying the scope only overwrites event data such as tags, user, environment, level, and context if these properties are nil. The client no longer applies the scope to crash events to avoid setting irrelevant values.

Fixes GH-4937

Unit tests, smoke test on a simulator

This feature branch
https://sentry-sdks.sentry.io/issues/5992437707/?project=5428557&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=0
CleanShot 2025-03-11 at 14 23 53@2x

Main Branch
https://sentry-sdks.sentry.io/issues/5992437707/events/543c0f451cc64233b984df68da2970b8/?project=5428557

CleanShot 2025-03-11 at 14 24 30@2x

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryScope.m
+++ b/Sources/Sentry/SentryScope.m
@@ -583,18 +583,6 @@ NS_ASSUME_NONNULL_BEGIN
         event.level = level;
     }
 
-    NSMutableDictionary *newContext = [self context].mutableCopy;
-    if (event.context != nil) {
-        [SentryDictionary mergeEntriesFromDictionary:event.context intoDictionary:newContext];
-    }
-
-    // Don't add the trace context of a current trace to a crash event because crash events are from
-    // a previous run.
-    if (event.isCrashEvent) {
-        event.context = newContext;
-        return event;
-    }
-
     id<SentrySpan> span;
 
     if (self.span != nil) {
@@ -609,6 +597,11 @@ NS_ASSUME_NONNULL_BEGIN
                 event.transaction = [[(SentryTracer *)span transactionContext] name];
             }
         }
+    }
+
+    NSMutableDictionary *newContext = [self context].mutableCopy;
+    if (event.context != nil) {
+        [SentryDictionary mergeEntriesFromDictionary:event.context intoDictionary:newContext];
     }
 
     newContext[@"trace"] = [self buildTraceContext:span];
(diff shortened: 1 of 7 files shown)


Sources/Sentry/SentryScope.m
```diff
@@ -583,18 +583,6 @@ - (SentryEvent *__nullable)applyToEvent:(SentryEvent *)event
...
-    if (event.context != nil) {
-        [SentryDictionary mergeEntriesFromDictionary:event.context intoDictionary:newContext];
-    }
-
-    // Don't add the trace context of a current trace to a crash event because crash events are from
-    // a previous run.
-    if (event.isCrashEvent) {
-        event.context = newContext;
-        return event;
-    }
```
philipphofmann, line 596: We don't need this anymore because we don't call this for crash events anymore in the client.

Tests/SentryTests/SentryScopeSwiftTests.swift
```diff
@@ -231,17 +231,6 @@ class SentryScopeSwiftTests: XCTestCase {
         XCTAssertEqual(trace?["span_id"] as? String, fixture.transaction.spanId.sentrySpanIdString)
     }
     
-    func testApplyToEvent_ScopeWithSpan_NotAppliedToCrashEvent() {
```
philipphofmann, line 234: I moved this test to the client tests `testCaptureCrashEventWithSession_ScopeWithSpan_NotAppliedToCrashEvent`

Sources/Sentry/SentryScope.m
philprime: How can we assure it won't be called for crashes again in the future?

Tests/SentryTests/SentryScopeSwiftTests.swift
philprime: Thanks for mentioning.

philprime (approved):
LGTM, left one feedback comment

Sources/Sentry/SentryScope.m
philipphofmann: I will add an early exit at the top of the method.

philprime (approved):

Issue #3846, opened 2024-04-11
Bump minimum OS versions


philipphofmann:
### Description

As you soon can only publish to the App Store with Xcode 15, we can bump the minimum OS of the SDK to the versions [supported by Xcode 15](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes), which are iOS 12 and later, tvOS 12 and later, and watchOS 4 and later. 

We don't need a major bump for this, but we should clearly point this out in the Changelog and the docs.

armcknight:
Continuing the discussion here on this comment: https://github.com/getsentry/sentry-cocoa/pull/3845#pullrequestreview-1993989747

> Xcode 15 allows you to target macOS 10.13, iOS 12, tvOS 12, and watchOS 4. We must support these versions. This is also in our [SDK Philosophy and Design Principles](https://develop.sentry.dev/sdk/philosophy/#compatibility-is-king).

I am on board with compatibility, be we have to be able to test the platforms we claim to support. We should either prioritize getting that working, or compromise on the stance. it can still be a guideline without being absolute dogma. I think if we must choose between compromising on validated releases via automated testing, or compromising on supporting the earliest possible platform version, then automated testing should win out over compatibility.

I think this philosophy is covered in the section right above it, under "Degrade Gracefully":
> If this cannot be achieved we need to make sure that we inform the user ahead of time by alerting about such situations at installation time. If for instance an SDK requires a minimum version to even build, we need to ensure that the user is informed and that we do not cause confusing errors later.

IMO This is why this should be a breaking change with a bump to 9.0.0, and then in the highly unlikely event we hear from a customer that they need either a new feature or a patch on e.g. iOS <13, we could backport a hotfix to 8.x.y. That, to me, is how you really hit the sweet spot between Compatibility and Graceful Degradation.

bruno-garcia:
I love these principles but Compatibility is King doesn't mean support stuff no one uses.

philipphofmann:
> I am on board with compatibility, be we have to be able to test the platforms we claim to support.

It's essential we test our code on the most used / most recent platforms, of course, but I think it's acceptable not to test on all platforms we support, especially the ones almost nobody uses anymore. The major reason why we don't test our code on iOS 12 and such is because GH actions doesn't have good support for that. With our test suite in place, it mayn't reveal bugs on older, less-used OS versions, but that chance also exists on the most adopted OS versions.

[...]

PR #4971, opened 2025-03-12
meta: add decision for major-4 version support

armcknight:
INRE recent slack conversations and https://github.com/getsentry/sentry-cocoa/issues/3846

#skip-changelog

### DIFF
--- a/develop-docs/DECISIONS.md
+++ b/develop-docs/DECISIONS.md
@@ -390,7 +390,7 @@ We can also start with this option to evaluate Swift Codable and switch to optio
 Date: March 11, 2025
 Contributors: @armcknight, @philipphofmann, @kahest
 
-We will support versions of each platform going back 4 major versions. There are 3 considerations:
+We will support versions of each platform going back 4 major versions, but we support no version which is not debuggable by the current Xcode required to submit apps to stores. There are 3 considerations:
 
 1. The distribution of events we receive from the various versions of iOS etc in the wild.
 1. Xcode support. As of the time of this writing, the oldest version of Xcode that can still submit apps to the app store is Xcode 15, which supports back to iOS 12, while the current is iOS 18.
--- a/develop-docs/DECISIONS.md
+++ b/develop-docs/DECISIONS.md
@@ -384,3 +384,20 @@ We can also start with this option to evaluate Swift Codable and switch to optio
 #### Cons
 
 1. Duplicate code.
+
+## Platform version support
+
+Date: March 11, 2025
+Contributors: @armcknight, @philipphofmann, @kahest
+
+We will support versions of each platform going back 4 major versions. There are 3 considerations:
+
+1. The distribution of events we receive from the various versions of iOS etc in the wild.
+1. Xcode support. As of the time of this writing, the oldest version of Xcode that can still submit apps to the app store is Xcode 15, which supports back to iOS 12, while the current is iOS 18.
+1. GitHub Actions support. This dictates which versions we can automatically test. Their oldest [macOS runner image](https://github.com/actions/runner-images/tree/main/images/macos) is `macos-13` with support going back to iOS 16.1
+
+Our major-4 standard would place us right in the middle of the earliest Xcode and GitHub Actions supported versions, which seems like a reasonable standard.
+
+Those versions that cannot be automatically tested with GitHub Actions shall be declared as "best effort" support.
+
+See previous discussion at https://github.com/getsentry/sentry-cocoa/issues/3846.


armcknight:
Should probably add something about cadence, like around WWDCs in June and the September GA launches, and how we've typically used discussions to gather other things to ship in a forthcoming major, see the latest one: https://github.com/getsentry/sentry-cocoa/discussions/2946

kahest (approved):
lgtm, thanks for adding!

the note about cadence makes sense as a side note but isn't part of the decision IMO, so I'll leave it to you

develop-docs/DECISIONS.md
```diff
@@ -384,3 +384,20 @@ We can also start with this option to evaluate Swift Codable and switch to optio
...
+## Platform version support
+
+Date: March 11, 2025
+Contributors: @armcknight, @philipphofmann, @kahest
+
+We will support versions of each platform going back 4 major versions. There are 3 considerations:
+
+1. The distribution of events we receive from the various versions of iOS etc in the wild.
+1. Xcode support. As of the time of this writing, the oldest version of Xcode that can still submit apps to the app store is Xcode 15, which supports back to iOS 12, while the current is iOS 18.
+1. GitHub Actions support. This dictates which versions we can automatically test. Their oldest [macOS runner image](https://github.com/actions/runner-images/tree/main/images/macos) is `macos-13` with support going back to iOS 16.1
```
philipphofmann, line 397: `l`:
```suggestion
1. The distribution of events we receive from the various versions of iOS etc in the wild.
2. Xcode support. As of the time of this writing, the oldest version of Xcode that can still submit apps to the app store is Xcode 15, which supports back to iOS 12, while the current is iOS 18.
3. GitHub Actions support. This dictates which versions we can automatically test. Their oldest [macOS runner image](https://github.com/actions/runner-images/tree/main/images/macos) is `macos-13` with support going back to iOS 16.1
```

develop-docs/DECISIONS.md
```diff
@@ -384,3 +384,20 @@ We can also start with this option to evaluate Swift Codable and switch to optio
...
+
+## Platform version support
+
+Date: March 11, 2025
+Contributors: @armcknight, @philipphofmann, @kahest
+
+We will support versions of each platform going back 4 major versions. There are 3 considerations:
+
+1. The distribution of events we receive from the various versions of iOS etc in the wild.
+1. Xcode support. As of the time of this writing, the oldest version of Xcode that can still submit apps to the app store is Xcode 15, which supports back to iOS 12, while the current is iOS 18.
```
philipphofmann, line 396: `m`: I that's key. You can still submit apps with Xcode 15, which still supports debugging apps on iOS 12. I think we can reasonably always bump it to the min debuggable Xcode version with which you can still submit apps to the app store. So by the end or mid of April, when you can only submit with Xcode 16, it's OK to raise this to iOS 15. For now it's still iOS 12. WDYT, @kahest and @armcknight?

philipphofmann: If we would ship a major now, I would bump to iOS 14 for sure, because you won't be able to submit apps to the app store with Xcode 15 soon.

develop-docs/DECISIONS.md
armcknight: i just hate having to fix up the ordering when things move around ¯\_(ツ)_/¯

develop-docs/DECISIONS.md
kahest: I would bump to iOS 14 even if it was possible to use Xcode 15 longer than a few months, because supporting 4 major versions back whenever we ship a major should be more than enough backwards compatibility in pretty much all cases, which is why I suggest the major-4 rule as default. do you agree?

develop-docs/DECISIONS.md
```diff
@@ -384,3 +384,20 @@ We can also start with this option to evaluate Swift Codable and switch to optio
 #### Cons
 
 1. Duplicate code.
+
+## Platform version support
+
+Date: March 11, 2025
+Contributors: @armcknight, @philipphofmann, @kahest
+
+We will support versions of each platform going back 4 major versions. There are 3 considerations:
```
kahest, line 393:
```suggestion
We will support versions of each platform going back 4 major versions, but we support no version which is not debuggable by the current Xcode required to submit apps to stores. There are 3 considerations:
```

develop-docs/DECISIONS.md
philipphofmann: OK, then let's keep it simple. Make it major four, and if somebody complains, we revisit. We don't do a major that often, and clients could stick to an older version.

PR #4976, opened 2025-03-12
tests: change test coverage targets to Sentry and SentrySwiftUI

philprime:
According to the Codecov reports we are collecting coverage data about the test files, rather than only the framework.

This PR should reduce the collected code coverage to the targets `Sentry` and `SentrySwiftUI`.

#skip-changelog


philprime:
Closing this because I do not have the time resources available anymore to look into it

PR #4978, opened 2025-03-14
test: fix continuous launch profile test

armcknight:
I noticed while working on #4856 that this test wasn't functioning correctly. It's a combination of things:
 - first, https://github.com/getsentry/sentry-cocoa/pull/4024 changed a launch argument in iOS-Swift that switches between trace and continuous profiling, but didn't change it in the UI test. what wound up happening is that the UI test would run a trace profile, and since trace and continuous profile chunk data were both written to the same debug file, the assertions didn't know it was looking at a trace profile when they should've been checking a continuous profile. to mitigate this in the future, i've changed `SentryProfilerTestHelperssentry_writeProfileFile` so that it either writes to `trace-profiles/` or `continuous-profiles/`.
 - then, we changed the behavior of calling stop on a continuous profiler in https://github.com/getsentry/sentry-cocoa/pull/4214 so that it doesn't actually stop the profile, it lets the chunk finish. due to the bug above, the UI tests didn't fail. it didn't occur to us that this new asynchronous behavior needed to be accounted for in the UI test.
 - then, we extended the duration of a profile chunk from 10s to 60s in https://github.com/getsentry/sentry-cocoa/pull/4826. this doesn't really have any bearing on the issue, but would've been the most recent opportunity to fix this test
 
Because this is a UI test, we can't simply inject `TestSentryNSTimerFactory` from the test code like we do with unit tests, and really can't even do it using launch args handled in the app/sdk code, since that mock utility only exists in the unit tests. For now, I've settled on conditionally compiling behavior, additionally protected with a launch argument set from the UI tests, that will immediately stop a continuous profile when stop is called.

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift-UITests/ProfilingUITests.swift
+++ b/Samples/iOS-Swift/iOS-Swift-UITests/ProfilingUITests.swift
@@ -8,8 +8,9 @@ class ProfilingUITests: BaseUITest {
     override func setUp() {
         super.setUp()
 
-        // make sure there are no previous configuration files or profile files written
-        app.launchArguments.append("--io.sentry.wipe-data")
+        app.launchArguments.append(contentsOf: [
+            "--io.sentry.wipe-data" // make sure there are no previous configuration files or profile files written
+        ])
     }
     
     func testAppLaunchesWithTraceProfiler() throws {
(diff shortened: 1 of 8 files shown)


philprime (approved):
LGTM

PR #4981, opened 2025-03-14
feat(profiling): new launch profiling option

armcknight:
Introduce the new `SentryProfileOptions.profileAppStarts` option.

I realized I would need to be able to distinguish between continuous profiling V1 with launch profiling enabled and this V2 launch profiling, so I had to rework the way `SentryProfileOptions` are configured in SentryOptions, so I used a configuration block pattern similar to user feedback and replay, so that I could look for the presence of the configuration by nilchecking it.

Note that while the diff size on this is large, almost all of it is test code.

#skip-changelog; for https://github.com/getsentry/sentry-cocoa/issues/4853


Sources/Sentry/SentryProfiler.mm
```diff
@@ -36,13 +36,39 @@
...
 
+} // namespace
+
+#    pragma mark - Public
+
 void
-_sentry_configureContinuousProfiling(SentryOptions *options)
+sentry_reevaluateSessionSampleRate(float sessionSampleRate)
+{
+    // TODO: carry over computed sample decision from last launch profile configuration, if there
```
philprime, line 46: `m`: should we resolve this TODO before merge?

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
```diff
@@ -106,13 +106,6 @@ final class SentryContinuousProfilerTests: XCTestCase {
         try assertContinuousProfileStoppage()
     }
 
-    func testBackgroundNotificationStopsProfile() {
```
philprime, line 109: `m`: why is this test removed?

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -31,20 +31,26 @@
...
-sentry_config(NSNumber *profilesRate, NSNumber *profilesRand)
+sentry_configForLaunchProfilerForTrace(
+    NSNumber *profilesRate, NSNumber *profilesRand, SentryProfileOptions *_Nullable profileOptions)
 {
     SentryTracerConfiguration *config = [SentryTracerConfiguration defaultConfiguration];
     config.profilesSamplerDecision =
         [[SentrySamplerDecision alloc] initWithDecision:kSentrySampleDecisionYes
                                           forSampleRate:profilesRate
                                          withSampleRand:profilesRand];
+    config.profileOptions = profileOptions;
```
philprime, line 53: `m`: does this create a new copy of the options? Is it a problem if it is the original object?

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.swift
```diff
@@ -14,25 +14,98 @@ final class SentryAppLaunchProfilingSwiftTests: XCTestCase {
...
+        XCTAssertFalse(appLaunchProfileConfigFileExists())
+        sentry_sdkInitProfilerTasks(options, TestHub(client: nil, andScope: nil))
+        XCTAssertTrue(appLaunchProfileConfigFileExists())
+        _sentry_nondeduplicated_startLaunchProfile()
+        XCTAssert(SentryTraceProfiler.isCurrentlyProfiling())
+    }
+}
+
+// MARK: transaction based profiling iOS-only
+#if !os(macOS)
```
philprime, line 97: `m`: instead of removing the entire tests it would be great to keep the tests and use `XCTSkip` in each test instead

Sources/Sentry/SentryProfiler.mm
armcknight: fixed in https://github.com/getsentry/sentry-cocoa/pull/5008

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
armcknight: i moved it and then a merge from upstream brought in the old copy as well. it also exists at https://github.com/getsentry/sentry-cocoa/blob/d21e64c07e78ddf99d12034d0533d8438f9b2f70/Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift#L83

Sources/Swift/Integrations/Performance/SentryProfileOptions.swift
```diff
@@ -60,4 +64,18 @@ public class SentryProfileOptions: NSObject {
...
+    /// during your app's launch.
+    /// - warning: Continuous profiling is an experimental feature and may still contain bugs.
+    /// - note: `sessionSampleRate` is evaluated on the previous launch and only takes effect when
+    /// app start profiling activates on the next launch.
+    /// - note: If `lifecycle` is `manual`, profiling is started automatically on startup, but you
+    /// must manually call `SentrySDK.stopProfiler()` whenever you app startup to be complete. If
+    /// `lifecycle` is `trace`, profiling is started automatically on startup, and will
+    /// automatically be stopped when the root span that is associated with app startup ends.
+    /// - note: Profiling is automatically disabled if a thread sanitizer is attached.
+    public var profileAppStarts: Bool = false
```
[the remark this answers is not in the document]
armcknight: we used launch before but are moving to start to align with the rest of the SDK (see eg https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentryAppStartTracker.m)

Tests/SentryProfilerTests/SentryAppLaunchProfilingTests.swift
armcknight: the API inside the test bodies doesn't compile on other platforms, we'd have to also surround the test body by `#if available`, i don't think there's value in doing this

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -58,53 +64,118 @@
...
 
+SentryLaunchProfileConfig
+sentry_launchShouldHaveContinuousProfilingV2(SentryOptions *options)
+{
+    if (!options.profiling.profileAppStarts) {
+        SENTRY_LOG_DEBUG(@"Continuous profiling v2 enabled but disabled app start profiling, "
+                         @"won't profile launch.");
+        return (SentryLaunchProfileConfig) { NO, nil, nil };
+    }
+    if (options.profiling.lifecycle == SentryProfileLifecycleTrace) {
```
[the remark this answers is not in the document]
armcknight: we nilcheck it before calling the enclosing function here, and there are unit tests exercising every config combination

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -58,53 +64,118 @@
...
 
-    SENTRY_LOG_DEBUG(@"Will start trace profile next launch.");
+    SENTRY_LOG_DEBUG(@"Will start transaction profile next launch; will profile launch.");
     return (SentryLaunchProfileConfig) { YES, tracesSamplerDecision, profilesSamplerDecision };
 }
 
+SentryLaunchProfileConfig
+sentry_launchShouldHaveContinuousProfilingV2(SentryOptions *options)
+{
+    if (!options.profiling.profileAppStarts) {
```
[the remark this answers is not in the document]
armcknight: https://github.com/getsentry/sentry-cocoa/pull/4981#discussion_r2023599486

Sources/Configuration/SDK.xcconfig
```diff
@@ -163,7 +161,8 @@ SWIFT_OPTIMIZATION_LEVEL_Test = -Onone
 SWIFT_OPTIMIZATION_LEVEL_TestCI = -Onone
 SWIFT_OPTIMIZATION_LEVEL = $(SWIFT_OPTIMIZATION_LEVEL_$(CONFIGURATION))
 
-SWIFT_TREAT_WARNINGS_AS_ERRORS = YES
+GCC_TREAT_WARNINGS_AS_ERRORS = NO
```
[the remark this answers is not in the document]
armcknight: thanks for reminding me about this, this will not be merged into main, and i thought i'd reverted it, but one of the many merges must've brought it back. i'll fix it here now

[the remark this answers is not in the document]
armcknight: done in https://github.com/getsentry/sentry-cocoa/pull/4981/commits/2588a27db59ce132f5f4fd8bf733cdb4df75bea6

armcknight:
thanks @philprime for the review! addressed all your comments. as for the CI issues, i just merged in #5008 which fixed the remaining issue for it.

philprime (approved):
LGTM

Issue #4582, opened 2024-11-27
Fix crash in sentry_toIso8601String


philipphofmann:
### Description

The SDK sometimes crashes with SIGABRT in `sentry_toIso8601String`. 

```
Exception Type: EXC_CRASH (SIGABRT)

Thread 0 Crashed:
0   libsystem_kernel.dylib          0x1e54a2974         __pthread_kill
1   libsystem_pthread.dylib         0x1f8ee10e8         pthread_kill
2   libsystem_c.dylib               0x1a4f5bb7c         abort
3   libsystem_malloc.dylib          0x1ad186c64         malloc_vreport
4   libsystem_malloc.dylib          0x1ad186f0c         malloc_zone_error
5   libsystem_malloc.dylib          0x1ad17ca40         nanov2_guard_corruption_detected
6   libsystem_malloc.dylib          0x1ad162f80         nanov2_allocate_outlined
7   CoreFoundation                  0x19cfa8ab4         _CFRuntimeCreateInstance
8   CoreFoundation                  0x19cfa8540         __CFStringCreateImmutableFunnel3
9   CoreFoundation                  0x19d01ef04         CFDateFormatterCreateStringWithAbsoluteTime
10  Foundation                      0x19bedaab4         -[NSDateFormatter stringForObjectValue:]
11  Sentry.framework                0x1078f7670         -[NSDate(SentryExtras) sentry_toIso8601String]
12  Sentry.framework                0x1078e6a58         -[SentryBreadcrumb serialize]
13  Sentry.framework                0x107908c48         -[SentryScope addBreadcrumb:]
14  Sentry.framework                0x1078d9a88         -[SentryHub addBreadcrumb:]
15  Sentry.framework                0x107907bf4         +[SentrySDK addBreadcrumb:]
```

We also see this in our [internal SDK crashes](https://sentry.sentry.io/issues/?groupStatsPeriod=auto&project=4505469596663808&query=is%3Aunresolved%20sentry_toIso8601String&referrer=issue-list&sort=user&statsPeriod=30d&viewId=35146).

philipphofmann:
I double-checked the code and didn't find anything that we can do to avoid this issue. Looking at more stacktraces, it looks like the `NSDateFormatter` crashes because it can't allocate the memory for the string. I added an extra test in https://github.com/getsentry/sentry-cocoa/pull/4983 to ensure the method doesn't crash when passing nil to it.

[...]

PR #4983, opened 2025-03-14
test: Add testToIso8601StringNil_ReturnsNil

philipphofmann:
Add a test to ensure we don't crash when passing nil to sentry_toIso8601String.

Came up while investigating https://github.com/getsentry/sentry-cocoa/issues/4582.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/SentryTests.m
+++ b/Tests/SentryTests/SentryTests.m
@@ -130,7 +130,7 @@
     XCTAssertEqual([date timeIntervalSince1970], 1582803326.0);
 }
 
-- (void)testToIs08691StringNil_ReturnsNil
+- (void)testToIso8601StringNil_ReturnsNil
 {
 #pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/SentryTests.m
```diff
@@ -130,4 +130,13 @@ - (void)testDateCategoryCompactibility
     XCTAssertEqual([date timeIntervalSince1970], 1582803326.0);
 }
 
+- (void)testToIs08691StringNil_ReturnsNil
```
philprime, line 133:
```suggestion
- (void)testToIso8691StringNil_ReturnsNil
```

philprime (approved):
Thanks for adding this test, but why is this not part of #4982?

philipphofmann:
> Thanks for adding this test, but why is this not part of #4982?

This is toISO the PR above is about fromISO. Two different methods.

PR #4985, opened 2025-03-14
chore: Remove no-commit-to-branch pre-commit

philipphofmann:
The  hook prevents committing directly to the main branch, but this can be annoying for developers who use a different Git workflow. If you accidentally try to commit to main, you have to create a new branch and rerun all pre-commit checks.

Since we already enforce branch protection rules for main via GitHub, this hook adds little value while reducing flexibility. Removing it allows for a more streamlined workflow without unnecessary friction.

#skip-changelog


philprime (approved):
LGTM

PR #4987, opened 2025-03-14
test: Use XCTUnwrap in NetworkTrackerTests

philipphofmann:
Use XCTUnwrap in
testGetRequest_SpanCreatedAndBaggageHeaderAdded
instead of ? to get better failing asserts.

Came up while investigating flaky tests.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Integrations/Performance/Network/SentryNetworkTrackerIntegrationTests.swift
+++ b/Tests/SentryTests/Integrations/Performance/Network/SentryNetworkTrackerIntegrationTests.swift
@@ -173,10 +173,10 @@ class SentryNetworkTrackerIntegrationTests: XCTestCase {
         dataTask.resume()
         wait(for: [expect], timeout: 5)
         
-        let children = Dynamic(transaction).children as [Span]?
-        
-        XCTAssertEqual(children?.count, 1) //Span was created in task resume swizzle.
-        let networkSpan = try XCTUnwrap(children?.first)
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/Integrations/Performance/Network/SentryNetworkTrackerIntegrationTests.swift
```diff
@@ -173,10 +173,10 @@ class SentryNetworkTrackerIntegrationTests: XCTestCase {
         dataTask.resume()
         wait(for: [expect], timeout: 5)
         
-        let children = Dynamic(transaction).children as [Span]?
-        
-        XCTAssertEqual(children?.count, 1) //Span was created in task resume swizzle.
-        let networkSpan = try XCTUnwrap(children?.first)
+        let children = try XCTUnwrap(Dynamic(transaction).children as [Span]?)
```
philprime, line 176: Could this also be an optional cast?

```suggestion
        let children = try XCTUnwrap(Dynamic(transaction).children as? [Span])
```

philprime (approved):

Tests/SentryTests/Integrations/Performance/Network/SentryNetworkTrackerIntegrationTests.swift
philipphofmann: Nope, that doesn't compile

```
Cast from 'Dynamic' to unrelated type '[any Span]' always fails
```

PR #4990, opened 2025-03-17
test: Improve touch tracker lock tests

philipphofmann:
The test flaked here https://github.com/getsentry/sentry-cocoa/actions/runs/13897038496/job/38880559094.

Instead of only simply increasing the timeout, I also improved the test a bit.

Instead of having three async operations iterating 1000 times over each method, add an async operation for each method all running concurrently to increase the likeliness of identifying synchronization issues potentially leading to crashes.
Furthermore, increase the timeout to 5 seconds so the test times out less frequently in CI.

#skip-changelog


philprime (approved):
LGTM

PR #4993, opened 2025-03-17
fix(profiling): start cancels prior stop request

armcknight:
After some discussion around how stop should work for the upcoming API changes, I realized this was incorrectly coded (or, correctly coded an incorrect assumption).

Stopping the continuous profiler doesn't immediately stop it, rather, it allows the current in-flight chunk to complete, then stops it. If a new call to start comes in, it should "cancel" the pending stop request, and simply allow the same continuous profiler to continue running.

### DIFF
--- a/Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
+++ b/Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
@@ -161,6 +161,34 @@ final class SentryContinuousProfilerTests: XCTestCase {
         SentryContinuousProfiler.stop()
         try assertContinuousProfileStoppage()
     }
+
+    func testStoppingAndStartingAgainBeforeFinalChunkCompletesResultsInOneProfile() throws {
+        // arrange
+        SentryContinuousProfiler.start()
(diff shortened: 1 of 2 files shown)


Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
```diff
@@ -161,6 +161,34 @@ final class SentryContinuousProfilerTests: XCTestCase {
...
+        fixture.currentDateProvider.advanceBy(interval: 1)
+        SentryContinuousProfiler.stop()
+
+        fixture.currentDateProvider.advanceBy(interval: 1)
+        XCTAssert(SentryContinuousProfiler.isCurrentlyProfiling())
+
+        fixture.currentDateProvider.advanceBy(interval: 1)
+        SentryContinuousProfiler.start()
+
+        fixture.currentDateProvider.advanceBy(interval: 60)
```
armcknight, line 180: This is the important assertion. Before making the change, this would fail, because the profiler never canceled the stop request on a subsequent call to start. Now, it passses.

indragiek (approved):

PR #4995, opened 2025-03-18
ref(profiling): public methods for new API

armcknight:
per some discussion about API terminology, we will not introduce new methods named `startProfileSession`/`stopProfileSession`, instead continuing with `startProfiler`/`stopProfiler`. also explain session a bit better in the SentryProfileOptions headerdocs

#skip-changelog; for #4851


Sources/Swift/Integrations/Performance/SentryProfileOptions.swift
```diff
@@ -42,8 +46,10 @@ public class SentryProfileOptions: NSObject {
     
     /// The % of user sessions in which to enable profiling.
     /// - warning: Continuous profiling is an experimental feature and may still contain bugs.
-    /// - note: Whether or not the session is sampled is determined once, when the SDK is initially
-    /// configured.
+    /// - note: The decision whether or not to sample profiles is computed using this sample rate
+    /// when the SDK is started, and applies until the app resigns its active status. It is then
+    /// reevaluated on subsequent launch events that start the SDK. The time between these two
```
indragiek, line 51: > It is then reevaluated on subsequent launch events that start the SDK
    
It's not clear to me from this that it would be re-evaluated upon foregrounding, since that doesn't "start the SDK" (which only happens on launch)

Sources/Sentry/Public/SentrySDK.h
```diff
@@ -372,51 +372,43 @@ SENTRY_NO_INIT
...
+ * launch profiles configured with @c SentryOptions.enableAppLaunchProfiling . If you need sampling,
+ * you must calculate your own sampling decisions to call this method or enable app launch
+ * profiling. @b Warning: this behavior relies on deprecated options and will change in a future
+ * version. In the future, there will be internal sampling and the default sample rate will be 0.
+ * @note If neither @c SentryOptions.profilesSampleRate nor @c SentryOptions.profilesSampler are
+ * set, and @c SentryOptions.configureProfiling is set, this method does nothing if the profiling
+ * session is not sampled with respect to @c SentryOptions.profileSessionSampleRate , (which
+ * defaults to a sample rate of 0), or if it is sampled but the profiler is already running.
+ * @note If neither @c SentryOptions.profilesSampleRate nor @c SentryOptions.profilesSampler are
+ * set, and @c SentryOptions.configureProfiling is set, this method does nothing if
```
indragiek, line 390: This documentation looks like parts of this are duplicated (e.g. the `@warning`) and there's other stuff that doesn't apply: `In the future, there will be internal sampling and the default sample rate will be 0.` -- we do have internal sampling now

armcknight: This part is tricky. I'm trying to warn people using v1 of continuous profiling, where we had no sampling facility, that in SDK version 9.0.0, the behavior of this method will switch out from under them, because it will go from having no sampling (continuous profiling v1) to having a default sample rate of 0 (continuous profiling v2) so if they don't configure the profile session sample rate when they upgrade, any calls to this method they already had will just go dark.

indragiek: IMO we should put that in the changelog/release notes - having this here for new customers who don't have the context will be more confusing. We're OK with breaking the prior behavior because it was a beta feature (and we weren't billing for usage).

armcknight: Good point 👍🏻 That'll simplify these a lot.

CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
...
+>
+> - `SentryOptions.enableProfiling`
+> - `SentryOptions.isProfilingEnabled`
+> - `SentryOptions.profilesSampleRate`
+> - `SentryOptions.profilesSampler`
+> - `SentryOptions.enableLaunchProfiling`
+>
+> Additionally, note that the behavior of `SentrySDK.startProfiler()` will change once the above API are removed, as follows: before adding the new configuration API (`SentryProfileOptions`), `SentrySDK.startProfiler()` would unconditionally start a continuous profile if both `SentryOptions.profilesSampleRate` and `SentryOptions.profilesSampler` were `nil`, or no-op if either was non-`nil` (meaning the SDK would operate under original, transaction-based, profiling model). In the next major version, `SentryOptions.profilesSampleRate` and `SentryOptions.profilesSampler` will be removed, and `SentrySDK.startProfile()` will become a no-op unless you configure `SentryProfileOptions.sessionSampleRate` to a value greater than zero (which is it's default). If you already have calls to `SentrySDK.startProfiler()` in your code, ensure you properly configure `SentryProfileOptions` via `SentryOptions.configureProfiling` to avoid losing profiling coverage.
```
philipphofmann, line 18: `l`: Two small typos
```suggestion
> Additionally, note that the behavior of `SentrySDK.startProfiler()` will change once the above APIs are removed, as follows: before adding the new configuration API (`SentryProfileOptions`), `SentrySDK.startProfiler()` would unconditionally start a continuous profile if both `SentryOptions.profilesSampleRate` and `SentryOptions.profilesSampler` were `nil`, or no-op if either was non-`nil` (meaning the SDK would operate under original, transaction-based, profiling model). In the next major version, `SentryOptions.profilesSampleRate` and `SentryOptions.profilesSampler` will be removed, and `SentrySDK.startProfile()` will become a no-op unless you configure `SentryProfileOptions.sessionSampleRate` to a value greater than zero (which is its default). If you already have calls to `SentrySDK.startProfiler()` in your code, ensure you properly configure `SentryProfileOptions` via `SentryOptions.configureProfiling` to avoid losing profiling coverage.
```

Tests/SentryProfilerTests/SentryProfilingPublicAPITests.swift
```diff
@@ -114,7 +115,10 @@ extension SentryProfilingPublicAPITests {
         SentrySDK.startProfiler()
         XCTAssertFalse(SentryContinuousProfiler.isCurrentlyProfiling())
     }
+}
 
+// MARK: continuous profiling v2
```
philipphofmann, line 120: Yes V1 and V2 not old and new 🙏 😃

philipphofmann (approved):
LGTM

PR #4998, opened 2025-03-19
test(profiling): fix broken tests for new API

armcknight:
In https://github.com/getsentry/sentry-cocoa/pull/4856#discussion_r1998977235, I changed the code incorrectly from using an assert to just a conditional check, causing some test failures. This fixes that, along with some cosmetic improvements to the tests.

#skip-changelog; for #4853


philprime (approved):
LGTM

Issue #4936, opened 2025-03-04
App Hangs: Rename client.captureCrashEvent


philipphofmann:
### Description

Capturing fatal app hang events use client.captureCrashEvent. Rename the method so it communicates it's not only about capturing crashes. Maybe captureFatalEvent does the trick.

PR #5001, opened 2025-03-20
ref: Improve ClientAttachmentProcessor

philipphofmann:
Specify types for NSArrays, rename attachmentsForEvent to processAttachmentsForEvent, only pass attachments to processAttachmentsForEvent, remove not required nullable for SentryClientAttachmentProcessor.

Came up while investigating GH-4936.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryClient.m
+++ b/Sources/Sentry/SentryClient.m
@@ -458,7 +458,8 @@ NSString *const DropSessionLogMessage = @"Session has no release name. Won't sen
 
     SentryTraceContext *traceContext = [self getTraceStateWithEvent:event withScope:scope];
 
-    NSArray *attachments = [self attachmentsForEvent:preparedEvent scope:scope];
+    NSArray<SentryAttachment *> *attachments = [self processAttachmentsForEvent:preparedEvent
+                                                                    attachments:scope.attachments];
 
     [self.transportAdapter sendEvent:preparedEvent
                         traceContext:traceContext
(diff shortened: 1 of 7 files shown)


Sources/Sentry/SentryClient.m
```diff
@@ -468,17 +469,6 @@ - (SentryId *)sendEvent:(SentryEvent *)event
     return preparedEvent.eventId;
 }
 
-- (NSArray *)attachmentsForEvent:(SentryEvent *)event scope:(SentryScope *)scope
```
philipphofmann, line 471: I pushed this method down close to `addAttachmentProcessor` and `removeAttachmentProcessor`.

Sources/Sentry/SentryScreenshotIntegration.m
```diff
@@ -59,8 +59,9 @@ - (void)uninstall
     [client removeAttachmentProcessor:self];
 }
 
-- (NSArray<SentryAttachment *> *)processAttachments:(NSArray<SentryAttachment *> *)attachments
-                                           forEvent:(nonnull SentryEvent *)event
+- (nonnull NSArray<SentryAttachment *> *)processAttachments:
```
philipphofmann, line 62: This class doesn't define `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`. Adding these would require even more changes, and I want to keep this PR small. The same applies for `SentryViewHierarchyIntegration.m`.

philprime (approved):
Thanks for adding these types!

Issue #5004, opened 2025-03-20
Remove truncation logic for SentryMessage


philipphofmann:
### Description

Let Relay truncate the max char length, as clarified in https://github.com/getsentry/sentry-docs/pull/13068. The benefit is that users will see that the message was truncated.

With SDK side truncation
![Image](https://github.com/user-attachments/assets/0b156e41-b448-4435-81d6-4f76e2ac84ce)

Without SDK side truncation Relay truncates

![Image](https://github.com/user-attachments/assets/912bf98f-b57e-40eb-8629-9ddb4222f067)

PR #5005, opened 2025-03-20
impr: Remove char limit of 8192 for SentryMessage

philipphofmann:
## :scroll: Description

Let Relay truncate the max char length. The benefit is that users see that the message was truncated.

Fixes GH-5004

Unit tests

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/Public/SentryMessage.h
+++ b/Sources/Sentry/Public/SentryMessage.h
@@ -20,7 +20,7 @@ SENTRY_NO_INIT
 /**
  * Returns a @c SentryMessage with setting formatted.
  * @param formatted The fully formatted message. If missing, Sentry will try to interpolate the
- * message. It must not exceed 8192 characters. Longer messages will be truncated.
+ * message. The backend will truncate messages longer than 8192 characters.
  */
 - (instancetype)initWithFormatted:(NSString *)formatted;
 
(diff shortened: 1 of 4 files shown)


Sources/Sentry/Public/SentryMessage.h
```diff
@@ -20,7 +20,7 @@ SENTRY_NO_INIT
 /**
  * Returns a @c SentryMessage with setting formatted.
  * @param formatted The fully formatted message. If missing, Sentry will try to interpolate the
- * message. It must not exceed 8192 characters. Longer messages will be truncated.
+ * message. The backend will truncate messages longer than 8192 characters.
```
philprime, line 23: `m`: I am aware of the reasoning for changing this, but what would happen if a message with a huge payload, i.e. 1MB of data, is uploaded? Is there an upper limit which the backend will drop? Is "no limit" better than a super-high limit?

philipphofmann: The goal is to move this power to the backend so we can easily change it. The backend will drop envelopes exceeding a certain limit yes. This problem exists for all properties the user can set. We currently only truncate SentryMessage. There is an edge case when a user sets a very very long message and the backend will drop your message now because the SDK doesn't truncate it anymore, but you need to use a message exceeding 1MB.

philprime: Thanks for clarifying.

philprime (approved):
LGTM

Sources/Sentry/Public/SentryMessage.h
philipphofmann: Thanks for double checking 🙏

PR #5007, opened 2025-03-21
ref(profiling): function renames

armcknight:
I was getting lost trying to debug launch profiling stuff so i renamed some things to try to help bring more clarity. Tried to keep this free of anything but renames, no logic should be changed.

#skip-changelog; for #4853


philipphofmann (approved):
Thanks for opening an extra PR 💯. LGTM

PR #5008, opened 2025-03-21
fix(profiling): add ui tests for new launch profiling modes

armcknight:
with some small fixes needed in logic

also converted more bare asserts to warnings in production with early exits where needed.

#skip-changelog; for #4853


Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -245,10 +279,24 @@
...
+        }
+
+        if (![SentryContinuousProfiler isCurrentlyProfiling]) {
+#    if SENTRY_TEST || SENTRY_TESTCI
+            SENTRY_CASSERT(
+                NO, @"Expected a continuous profiler to be running for this configuration.");
+#    else
+            SENTRY_LOG_WARN(
+                @"Expected a continuous profiler to be running for this configuration.");
+#    endif // SENTRY_TEST || SENTRY_TESTCI
```
philipphofmann, line 298: `l`: Maybe we could define a macro that calls `SENTRY_CASSERT` when `SENTRY_TEST || SENTRY_TESTCI` and calls `SENTRY_LOG_WARN` otherwise to reduce some duplication.

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
```diff
@@ -110,6 +110,8 @@
         return @"backgrounded";
     case SentryProfilerTruncationReasonTimeout:
         return @"timeout";
+    case SentryProfilerTruncationReasonPanic:
+        return @"panic";
```
philipphofmann, line 114: `m`: It would be great to have a test for this.

Sources/Sentry/SentryOptions.m
```diff
@@ -705,6 +705,12 @@ - (BOOL)isContinuousProfilingV2Enabled
     return [self isContinuousProfilingEnabled] && _profiling != nil;
 }
 
+- (BOOL)isProfilingCorrelatedToTraces
+{
+    return ![self isContinuousProfilingEnabled]
+        || _profiling.lifecycle == SentryProfileLifecycleTrace;
+}
+
```
philipphofmann, line 713: `m`: It would be great to have some unit tests for this logic.

philipphofmann (approved):
LGTM

Sources/Sentry/Profiling/SentryProfilerSerialization.mm
armcknight: It's actually not needed. Removed.

PR #5012, opened 2025-03-24
ci: Use xcodes in ci-select-xcode.sh

philipphofmann:
Replace manual selecting Xcode version with xcodes.

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,10 +29,10 @@ jobs:
   # With this we catch potential issues already in the PR.
   ios-swift-release:
     name: Release Build of iOS Swift
-    runs-on: macos-13
+    runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
-      - run: ./scripts/ci-select-xcode.sh
+      - run: ./scripts/ci-select-xcode.sh 16.2
       - uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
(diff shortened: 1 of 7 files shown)


philprime (approved):
LGTM, but maybe we should add a check that the Xcode switch actually happened, i.e. `xcrun --show-sdk-version` or any other of these:

```
$ xcrun
Usage: xcrun [options] <tool name> ... arguments ...

Find and execute the named command line tool from the active developer
directory.

The active developer directory can be set using `xcode-select`, or via the
DEVELOPER_DIR environment variable. See the xcrun and xcode-select manual
pages for more information.

Options:
  -h, --help                  show this help message and exit
  --version                   show the xcrun version
  -v, --verbose               show verbose logging output
  --sdk <sdk name>            find the tool for the given SDK name
  --toolchain <name>          find the tool for the given toolchain
  -l, --log                   show commands to be executed (with --run)
  -f, --find                  only find and print the tool path
  -r, --run                   find and execute the tool (the default behavior)
  -n, --no-cache              do not use the lookup cache
  -k, --kill-cache            invalidate all existing cache entries
  --show-sdk-path             show selected SDK install path
  --show-sdk-version          show selected SDK version
  --show-sdk-build-version    show selected SDK build version
  --show-sdk-platform-path    show selected SDK platform path
  --show-sdk-platform-version show selected SDK platform version
```

philprime:
Great, maybe add that as expected behaviour as a comment so we remember in the future in case xcodes default behavior gets changed

philipphofmann:
> Great, maybe add that as expected behaviour as a comment so we remember in the future in case xcodes default behavior gets changed

Still a draft PR 😃. Yes, I will.

PR #5015, opened 2025-03-25
ci: Use installed simulators for critical UI tests

philipphofmann:
Stop creating simulators, which can take around 5 minutes, and use the already installed simulators on the GH action.

#skip-changelog

### DIFF
--- a/.github/workflows/ui-tests-critical.yml
+++ b/.github/workflows/ui-tests-critical.yml
@@ -51,12 +51,17 @@ jobs:
       fail-fast: false
       matrix:
         include:
+
+          # As of 25th March 2025, the preinstalled iOS simulator version is 16.4 for macOS 13 and Xcode 14.3.1; see 
+          # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#installed-sdks
           - runs-on: macos-13
             xcode: "14.3.1"
             device: "iPhone 14"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ui-tests-critical.yml
```diff
@@ -54,20 +54,16 @@ jobs:
...
             device: "iPhone 14"
-            os-version: "16.4"
-            force-sim-runtime: true
 
           # macos-14 iOS 17 not included due to the XCUIServerNotFound errors causing flaky tests
 
           - runs-on: macos-15
             xcode: "16.2"
             device: "iPhone 16"
-            os-version: "18.2"
```
philprime, line 65: `h`: This removes testing for different OS versions!

philipphofmann: I added a comment to clarify that it doesn't.

philprime (approved):
Without checking the default simulator version, this might lead to issues down the road, because the runner image is outside of our control (e.g. they update it to support newer iOS versions without a new macOS version). The risk is probably not to impactful.

philipphofmann:
> Without checking the default simulator version, this might lead to issues down the road, because the runner image is outside of our control (e.g. they update it to support newer iOS versions without a new macOS version). The risk is probably not to impactful.

Yes, indeed. I also thought about that. My assumption is that creating simulators increases flakiness. Also it takes quite some time. I want to remove the create-simulator.sh script in an upcoming PR. I will add a note to the internal develop docs, explaining the approach.

PR #5016, opened 2025-03-25
ci: Use installed simulators for UI tests

philipphofmann:
Stop creating simulators which can take around 5 minutes and use the already installed simulators on the GH action.

#skip-changelog

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -113,7 +113,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # As of 25th March 2025, the preinstalled iOS simulator version is 16.4 for macOS 13 and Xcode 15.2; see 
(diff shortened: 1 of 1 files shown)


.github/workflows/ui-tests.yml
```diff
@@ -107,29 +107,24 @@ jobs:
...
       fail-fast: false
       matrix:
         include:
           - runs-on: macos-13
             xcode: "15.2"
-            device: "iPhone 14 (16.4)"
 
           - runs-on: macos-14
             xcode: "15.4"
-            device: "iPhone 15 (17.2)"
```
philprime, line 122: `h`: This removes testing for different iOS versions (16 and 17)

philipphofmann: No it doesn't. I added a comment to clarify this.

philipphofmann: I added a comment for clarification

philprime (approved):
See my comment here https://github.com/getsentry/sentry-cocoa/pull/5015#pullrequestreview-2714380243

PR #5018, opened 2025-03-25
fix(session-replay): improve multi-threading of session replay processing

### LINKED ISSUE #4730: Session Replay thread priority inversions
### Platform

iOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

8.43.0

### Xcode Version

16.2

_No response_

### Steps to Reproduce

Here is the config:

```
        SentrySDK.start { options in
            guard let self = self else { return }
            options.dsn = ""
            options.enableWatchdogTerminationTracking = false
#if os(iOS)
            options.sessionReplay.onErrorSampleRate = 1.0
            options.sessionReplay.sessionSampleRate = 0.1
            options.sessionReplay.maskAllText = false
            options.sessionReplay.maskAllText = false
#endif
        }
```

### Expected Result

Should upload session replay without any errors or warnings.

### Actual Result

Getting this warning printed in the xcode console:

```
Thread Performance Checker: Thread running at User-initiated quality-of-service class waiting on a lower QoS thread running at Default quality-of-service class. Investigate ways to avoid priority inversions
PID: 4838, TID: 1230510
Backtrace
=================================================================
0   Calorimeter.debug.dylib             0x000000010c7e95cc $s6Sentry0A14OnDemandReplayC11renderVideo33_70FE3B80E922CEF5576FF378226AFAE1LL4with4from2atAA0aF4InfoCSgSayAA0aD5FrameVG_Siz10Foundation3URLVtKF + 1756
1   Calorimeter.debug.dylib             0x000000010c7e8cb4 $s6Sentry0A14OnDemandReplayC15createVideoWith9beginning3endSayAA0aF4InfoCG10Foundation4DateV_ALtKF + 1180
2   Calorimeter.debug.dylib             0x000000010c7e8e44 $s6Sentry0A14OnDemandReplayC15createVideoWith9beginning3endSayAA0aF4InfoCG10Foundation4DateV_ALtKFTo + 156
3   Calorimeter.debug.dylib             0x000000010c7cfd14 $s6Sentry0A13SessionReplayC16createAndCapture33_0C34F63BF2CAC7EBBCDDA9F08A50850ALL9startedAt10replayTypey10Foundation4DateV_AA0acQ0OtFyycfU_ + 284
4   Calorimeter.debug.dylib             0x000000010a50c5a4 $sIeg_IeyB_TR + 48
5   Calorimeter.debug.dylib             0x000000010c7172f4 __53-[SentryDispatchQueueWrapper dispatchAsyncWithBlock:]_block_invoke + 36
6   libdispatch.dylib                   0x000000010303888c _dispatch_call_block_and_release + 32
7   libdispatch.dylib                   0x000000010303a578 _dispatch_client_callout + 20
8   libdispatch.dylib                   0x0000000103042454 _dispatch_lane_serial_drain + 840
9   libdispatch.dylib                   0x000000010304325c _dispatch_lane_invoke + 408
10  libdispatch.dylib                   0x00000001030506fc _dispatch_root_queue_drain_deferred_wlh + 328
11  libdispatch.dylib                   0x000000010304fd0c _dispatch_workloop_worker_thread + 580
12  libsystem_pthread.dylib             0x0000000219ccf680 _pthread_wqthread + 288
13  libsystem_pthread.dylib             0x0000000219ccd474 start_wqthread + 8

```

_No response_

philprime:
## :scroll: Description

- Changes the priorities of the background queues of session replay.
- Refactors the async dispatch handling of session replay.

Some of the methods of the session replay video processing are dispatched to a low priority background queue, which then dispatches additional AV work on another dispatch queue.

The current implementation does not respect queue priorities and can cause thread starvation, because the processing queue and the work queue are either on the same priority, or the work queue is on a lower priority.

In the worst case, the processing queue is blocked and waits for the work queue, but the work queue has a lower priority than the processing queue and is awaiting execution on the main thread.

This is what the thread inversion warning is showing.

Closes #4730

Manual testing.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentrySessionReplayIntegration.m
+++ b/Sources/Sentry/SentrySessionReplayIntegration.m
@@ -57,6 +57,9 @@ static SentryTouchTracker *_touchTracker;
     // This is the easiest way to ensure segment 0 will always reach the server, because session
     // replay absolutely needs segment 0 to make replay work.
     BOOL _rateLimited;
+    id<SentryCurrentDateProvider> _dateProvider;
+    SentryDispatchQueueWrapper *_replayProcessingQueue;
+    SentryDispatchQueueWrapper *_replayAssetWorkerQueue;
 }
 
 - (instancetype)init
@@ -120,6 +123,19 @@ static SentryTouchTracker *_touchTracker;
     }
 
     _notificationCenter = SentryDependencyContainer.sharedInstance.notificationCenterWrapper;
+    _dateProvider = SentryDependencyContainer.sharedInstance.dateProvider;
+
+    // The asset worker queue is used to work on video and frames data.
+    // Use a relative priority of -1 to make it lower than the default background priority.
+    _replayAssetWorkerQueue = [SentryDispatchQueueWrapper
+        createBackgroundDispatchQueueWithName:"io.sentry.session-replay.asset-worker"
+                             relativePriority:-1];
+    // The dispatch queue is used to asynchronously wait for the asset worker queue to finish its
+    // work. To avoid a deadlock, the priority of the processing queue must be lower than the asset
+    // worker queue. Use a relative priority of -2 to make it lower than the asset worker queue.
+    _replayProcessingQueue = [SentryDispatchQueueWrapper
+        createBackgroundDispatchQueueWithName:"io.sentry.session-replay.processing"
+                             relativePriority:-2];
 
     [self moveCurrentReplay];
     [self cleanUp];
@@ -189,30 +205,54 @@ static SentryTouchTracker *_touchTracker;
     }
 
     SentryOnDemandReplay *resumeReplayMaker =
-        [[SentryOnDemandReplay alloc] initWithContentFrom:lastReplayURL.path];
+        [[SentryOnDemandReplay alloc] initWithContentFrom:lastReplayURL.path
+                                          processingQueue:_replayProcessingQueue
+                                         assetWorkerQueue:_replayAssetWorkerQueue
+                                             dateProvider:_dateProvider];
     resumeReplayMaker.bitRate = _replayOptions.replayBitRate;
     resumeReplayMaker.videoScale = _replayOptions.sizeScale;
 
     NSDate *beginning = hasCrashInfo
         ? [NSDate dateWithTimeIntervalSinceReferenceDate:crashInfo.lastSegmentEnd]
         : [resumeReplayMaker oldestFrameDate];
-
     if (beginning == nil) {
         return; // no frames to send
     }
+    NSDate *end = [beginning dateByAddingTimeInterval:duration];
 
-    SentryReplayType _type = type;
-    int _segmentId = segmentId;
+    // This method is called from a background thread, so we can synchronize the creation of the
+    // video with a dispatch group.
+    __block NSArray<SentryVideoInfo *> *videos;
+    __block NSError *_Nullable error;
 
-    NSError *error;
-    NSArray<SentryVideoInfo *> *videos =
-        [resumeReplayMaker createVideoWithBeginning:beginning
-                                                end:[beginning dateByAddingTimeInterval:duration]
-                                              error:&error];
-    if (videos == nil) {
-        SENTRY_LOG_ERROR(@"Could not create replay video: %@", error);
+    dispatch_group_t group = dispatch_group_create();
+    dispatch_group_enter(group);
+    [resumeReplayMaker
+        createVideoAsyncWithBeginning:beginning
+                                  end:end
+                           completion:^(NSArray<SentryVideoInfo *> *_Nullable resultVideos,
+                               NSError *_Nullable resultError) {
+                               videos = resultVideos;
+                               error = resultError;
+                               dispatch_group_leave(group);
+                           }];
+    // Wait for the video creation to finish without a timeout, because the video creation is
+    // expected to finish in a reasonable time frame.
+    dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
+
+    // Either error or videos should be set.
+    if (error != nil) {
+        SENTRY_LOG_ERROR(@"Could not create replay video, reason: %@", error);
         return;
     }
+    if (videos == nil) {
+        SENTRY_LOG_ERROR(@"Could not create replay video, reason: no videos available");
+        return;
+    }
+
+    // For each segment we need to create a new event with the video.
+    int _segmentId = segmentId;
+    SentryReplayType _type = type;
     for (SentryVideoInfo *video in videos) {
         [self captureVideo:video replayId:replayId segmentId:_segmentId++ type:_type];
         // type buffer is only for the first segment
@@ -224,8 +264,10 @@ static SentryTouchTracker *_touchTracker;
         [NSDictionary dictionaryWithObjectsAndKeys:replayId.sentryIdString, @"replay_id", nil];
     event.context = eventContext;
 
-    if ([NSFileManager.defaultManager removeItemAtURL:lastReplayURL error:&error] == NO) {
-        SENTRY_LOG_ERROR(@"Can`t delete '%@': %@", SENTRY_LAST_REPLAY, error);
+    NSError *_Nullable removeError;
+    BOOL result = [NSFileManager.defaultManager removeItemAtURL:lastReplayURL error:&removeError];
+    if (result == NO) {
+        SENTRY_LOG_ERROR(@"Can`t delete '%@': %@", SENTRY_LAST_REPLAY, removeError);
     }
 }
 
@@ -316,30 +358,27 @@ static SentryTouchTracker *_touchTracker;
                                                      error:nil];
     }
 
-    SentryOnDemandReplay *replayMaker = [[SentryOnDemandReplay alloc] initWithOutputPath:docs.path];
+    SentryOnDemandReplay *replayMaker =
+        [[SentryOnDemandReplay alloc] initWithOutputPath:docs.path
+                                         processingQueue:_replayProcessingQueue
+                                        assetWorkerQueue:_replayAssetWorkerQueue
+                                            dateProvider:_dateProvider];
     replayMaker.bitRate = replayOptions.replayBitRate;
     replayMaker.videoScale = replayOptions.sizeScale;
     replayMaker.cacheMaxSize
         = (NSInteger)(shouldReplayFullSession ? replayOptions.sessionSegmentDuration + 1
                                               : replayOptions.errorReplayDuration + 1);
 
-    dispatch_queue_attr_t attributes = dispatch_queue_attr_make_with_qos_class(
-        DISPATCH_QUEUE_SERIAL, DISPATCH_QUEUE_PRIORITY_LOW, 0);
-    SentryDispatchQueueWrapper *dispatchQueue =
-        [[SentryDispatchQueueWrapper alloc] initWithName:"io.sentry.session-replay"
-                                              attributes:attributes];
-
-    self.sessionReplay = [[SentrySessionReplay alloc]
-        initWithReplayOptions:replayOptions
-             replayFolderPath:docs
-           screenshotProvider:screenshotProvider
-                  replayMaker:replayMaker
-          breadcrumbConverter:breadcrumbConverter
-                 touchTracker:_touchTracker
-                 dateProvider:SentryDependencyContainer.sharedInstance.dateProvider
-                     delegate:self
-                dispatchQueue:dispatchQueue
-           displayLinkWrapper:[[SentryDisplayLinkWrapper alloc] init]];
+    SentryDisplayLinkWrapper *displayLinkWrapper = [[SentryDisplayLinkWrapper alloc] init];
+    self.sessionReplay = [[SentrySessionReplay alloc] initWithReplayOptions:replayOptions
+                                                           replayFolderPath:docs
+                                                         screenshotProvider:screenshotProvider
+                                                                replayMaker:replayMaker
+                                                        breadcrumbConverter:breadcrumbConverter
+                                                               touchTracker:_touchTracker
+                                                               dateProvider:_dateProvider
+                                                                   delegate:self
+                                                         displayLinkWrapper:displayLinkWrapper];
 
     [self.sessionReplay
         startWithRootView:SentryDependencyContainer.sharedInstance.application.windows.firstObject
--- a/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
+++ b/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
@@ -7,27 +7,17 @@ import CoreGraphics
 import Foundation
 import UIKit
 
-struct SentryReplayFrame {
-    let imagePath: String
-    let time: Date
-    let screenName: String?
-}
-
-enum SentryOnDemandReplayError: Error {
-    case cantReadVideoSize
-    case cantCreatePixelBuffer
-    case errorRenderingVideo
-}
-
+// swiftlint:disable type_body_length
 @objcMembers
 class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
         
     private let _outputPath: String
     private var _totalFrames = 0
     private let dateProvider: SentryCurrentDateProvider
-    private let workingQueue: SentryDispatchQueueWrapper
+    private let processingQueue: SentryDispatchQueueWrapper
+    private let assetWorkerQueue: SentryDispatchQueueWrapper
     private var _frames = [SentryReplayFrame]()
-    
+
     #if SENTRY_TEST || SENTRY_TEST_CI || DEBUG
     //This is exposed only for tests, no need to make it thread safe.
     var frames: [SentryReplayFrame] {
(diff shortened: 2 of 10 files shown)


Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -189,30 +205,54 @@ - (void)resumePreviousSessionReplay:(SentryEvent *)event
...
-                                                end:[beginning dateByAddingTimeInterval:duration]
-                                              error:&error];
+    NSDate *end = [beginning dateByAddingTimeInterval:duration];
+
+    // This method is called from a background thread, so we can synchronize the creation of the
+    // video with a dispatch group.
+    __block NSArray<SentryVideoInfo *> *videos;
+    __block NSError *_Nullable error;
+
+    dispatch_group_t group = dispatch_group_create();
```
philprime, line 228: I would appreciate feedback if you asee any issues with this approach for synchronizing the async call

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -316,30 +358,27 @@ - (void)startWithOptions:(SentryReplayOptions *)replayOptions
...
+                                         processingQueue:_replayProcessingQueue
+                                        assetWorkerQueue:_replayAssetWorkerQueue
+                                            dateProvider:_dateProvider];
     replayMaker.bitRate = replayOptions.replayBitRate;
     replayMaker.videoScale = replayOptions.sizeScale;
     replayMaker.cacheMaxSize
         = (NSInteger)(shouldReplayFullSession ? replayOptions.sessionSegmentDuration + 1
                                               : replayOptions.errorReplayDuration + 1);
 
-    dispatch_queue_attr_t attributes = dispatch_queue_attr_make_with_qos_class(
```
philprime, line 326: Moved this setup to the SentryDispatchQueueWrapper to make it replaceable

philprime:
This PR grew into a a rather large refactor, I'll create additional PRs to split the changes into smaller digestible parts

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -120,128 +130,260 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
...
         let videoHeight = image.size.height * CGFloat(videoScale)
-        
-        let videoWriter = try AVAssetWriter(url: outputFileURL, fileType: .mp4)
+        let pixelSize = CGSize(width: videoWidth, height: videoHeight)
+
+        let videoWriter: AVAssetWriter
+        do {
+            videoWriter = try AVAssetWriter(url: outputFileURL, fileType: .mp4)
+        } catch {
+            SentryLog.debug("[Session Replay] Failed to create video writer, reason: \(error)")
```
philprime, line 221:
```suggestion
            SentryLog.error("[Session Replay] Failed to create video writer, reason: \(error)")
```

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -120,128 +130,260 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
...
     }
-    
-    private func renderVideo(with videoFrames: [SentryReplayFrame], from: inout Int, at outputFileURL: URL) throws -> SentryVideoInfo? {
-        guard from < videoFrames.count, let image = UIImage(contentsOfFile: videoFrames[from].imagePath) else { return nil }
+
+    // swiftlint:disable function_body_length cyclomatic_complexity
+    private func renderVideo(with videoFrames: [SentryReplayFrame], from: Int, at outputFileURL: URL, completion: @escaping (Result<SentryRenderVideoResult, Error>) -> Void) {
+        SentryLog.debug("[Session Replay] Rendering video with \(videoFrames.count) frames, from index: \(from), to output url: \(outputFileURL)")
+        guard from < videoFrames.count, let image = UIImage(contentsOfFile: videoFrames[from].imagePath) else {
+            SentryLog.debug("[Session Replay] Failed to render video, reason: index out of bounds or can't read image at path: \(videoFrames[from].imagePath)")
```
philprime, line 206:
```suggestion
            SentryLog.error("[Session Replay] Failed to render video, reason: index out of bounds or can't read image at path: \(videoFrames[from].imagePath)")
```

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -120,128 +130,260 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
...
-            let outputFileURL = URL(fileURLWithPath: _outputPath.appending("/\(videoFrames[frameCount].time.timeIntervalSinceReferenceDate).mp4"))
-            if let videoInfo = try renderVideo(with: videoFrames, from: &frameCount, at: outputFileURL) {
-                videos.append(videoInfo)
-            } else {
-                frameCount++
-            }  
+
+    func createVideoAsyncWith(beginning: Date, end: Date, completion: @escaping ([SentryVideoInfo]?, Error?) -> Void) {
+        // Note: In Swift it is best practice to use `Result<Value, Error>` instead of `(Value?, Error?)`
+        //       Due to interoperability with Objective-C and @objc, we can not use Result here.
```
philprime, line 142:
```suggestion
        //       Due to interoperability with Objective-C and @objc, we can not use Result for the completion callback.
```

philipphofmann:
I'm unsure if we're choosing the right approach here by adding two more DispatchQueues. If I'm not mistaken, the root cause is using the default DispatchQueue of the `SentryDispatchQueueWrapper` here: 
https://github.com/getsentry/sentry-cocoa/blob/38fc771279e2920cdb07f11c9cbe12ce4836a3e7/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift#L169-L170

`SentrySessionReplayIntegration.resumePreviousSessionReplay` already runs on a BG thread scheduled on the `SentryDispatchQueueWrapper.queue`. This `resumePreviousSessionReplay` then calls `SentryOnDemandReplay.createVideoWithBeginning`, which calls `SentryOnDemandReplay.renderVideo`. `renderVideo` then renders the video with `AVAssetWriterInput.requestMediaDataWhenReady`, which uses `SentryDispatchQueueWrapper.queue`. As this queue is serial, I'm not surprised that, especially when the SDK starts and there are a couple of envelopes that need to be sent, we receive a priority inversion warning. 

https://github.com/getsentry/sentry-cocoa/blob/38fc771279e2920cdb07f11c9cbe12ce4836a3e7/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift#L147-L206

I don't think we need to call `createVideoAsyncWithBeginning` in the `SentrySessionReplayIntegration`. Creating the video already happens on a BG thread. Making it async requires using a DispatchGroup below, so it's more complex again.

https://github.com/getsentry/sentry-cocoa/blob/44c4af72f68c588083c5c351b853ed8b76821275/Sources/Sentry/SentrySessionReplayIntegration.m#L231-L244

Instead, I think we need to use one extra dedicated queue for video processing in `SentryOnDemandReplay` and this should solve the issue. Maybe I'm missing something. I'm also happy to brainstorm this on a call.

philprime:
I'll try to go into your point, but yes I'll contact you regarding a call to discuss this in detail.


> If I'm not mistaken, the root cause is using the default DispatchQueue of the SentryDispatchQueueWrapper here:

Yes and no. The actual problem causing the warning is this line...

https://github.com/getsentry/sentry-cocoa/blob/38fc771279e2920cdb07f11c9cbe12ce4836a3e7/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift#L202

... as it was called from a queue which was created here:

https://github.com/getsentry/sentry-cocoa/blob/38fc771279e2920cdb07f11c9cbe12ce4836a3e7/Sources/Sentry/SentrySessionReplayIntegration.m#L333-L337

This serial queue with priority low was still somehow on a higher priority than the one used by the asset writer in `requestMediaDataWhenReady`, while being unclear on which priority. 

This `DISPATCH_QUEUE_PRIORITY_LOW` seems to not even exist in Swift [`DispatchQueue.QoSClass`](https://developer.apple.com/documentation/dispatch/dispatchqos/qosclass-swift.enum) which is why I changed it to `QOS_CLASS_BACKGROUND` here:

https://github.com/getsentry/sentry-cocoa/blob/44c4af72f68c588083c5c351b853ed8b76821275/Sources/Sentry/SentryDispatchQueueWrapper.m#L112-L113

This change alone might already fix the Xcode warning, but this PR is in general to improve multi-threading to ensure that thread inversion can not happen.

> I'm unsure if we're choosing the right approach here by adding two more DispatchQueues. 

> I don't think we need to call `createVideoAsyncWithBeginning` in the `SentrySessionReplayIntegration`. Creating the video already happens on a BG thread. Making it async requires using a DispatchGroup below, so it's more complex again.

We are not adding two more queues, we are only adding the `asset worker queue`.

The `processing queue` already existed here:

https://github.com/getsentry/sentry-cocoa/blob/38fc771279e2920cdb07f11c9cbe12ce4836a3e7/Sources/Sentry/SentrySessionReplayIntegration.m#L333-L337

But instead of using it in the `SentrySessionReplay.swift` here (only use case in the class):

https://github.com/getsentry/sentry-cocoa/blob/38fc771279e2920cdb07f11c9cbe12ce4836a3e7/Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift#L226-L242

I moved the background dispatching into the `createVideoWith(beginning:)` so that the class `SentrySessionReplay` **does not do any dispatching anymore**, but instead uses a callback approach to process the video segments ([PR diff](https://github.com/getsentry/sentry-cocoa/pull/5018/files#diff-7e90bc4012112a489d229dd525a4bb5c21ee3478152aae0f1c0b75b34aa85947R221-R239)).

The processing queue is therefore now moved to the `SentryOnDemandReplay`, but instead of just creating it in the initializer, I made it injectable from `SentrySessionReplayIntegration.m`:

https://github.com/getsentry/sentry-cocoa/blob/44c4af72f68c588083c5c351b853ed8b76821275/Sources/Sentry/SentrySessionReplayIntegration.m#L134-L139

> Instead, I think we need to use one extra dedicated queue for video processing in `SentryOnDemandReplay` and this should solve the issue. Maybe I'm missing something. I'm also happy to brainstorm this on a call.

That's what the newly added `asset worker queue` is for:

https://github.com/getsentry/sentry-cocoa/blob/44c4af72f68c588083c5c351b853ed8b76821275/Sources/Sentry/SentrySessionReplayIntegration.m#L129-L133

As mentioned above, the `createVideoAsyncWith(beginning:)` is dispatching into the background processing queue to iterate all the frames and convert them into segments...

https://github.com/getsentry/sentry-cocoa/blob/44c4af72f68c588083c5c351b853ed8b76821275/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift#L140C10-L147

... and we need to loop through multiple frames and segments, while waiting for the `renderVideo` callback to respond, therefore it uses a semaphore to synchronize it.

The reason why `renderVideo` is async, is caused by the AVAssetWriter using some self-configured background queue inside which is outside our control (as I did not see any way of configuring it), therefore instead we can only pass a callback queue to `requestMediaDataWhenReady`. Now as the `processing queue` is already waiting using the semaphore, the completion block of `requestMediaDataWhenReady` can not be dispatched on this blocked queue, and instead we need the `asset worker queue` to process media, then call the callback of `renderVideo` to then releasing the semaphore and the dispatch queue.

If we go back to the previous implementation of `renderVideo` which was not async with a callback but synchronized itself using DispatchGroup...

https://github.com/getsentry/sentry-cocoa/blob/38fc771279e2920cdb07f11c9cbe12ce4836a3e7/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift#L164-L171

... that would be the same logic as if we do the synchronization 
e could go back to the previous implementation of the `renderVideo` which did not offer a callback but synchronized itself using dispatch group, but that would still be the same as the logic in `createVideoAsyncWith`.

This is therefore also a shift of synchronization logic, but I don't understand why `renderVideo` should be synchronized in the first place, if the used `viewWriterInput` is also using callback-style async.

Finally the `asset worker queue` has an explicitly defined higher priority than the `processing queue`, so when the `processing queue` waits for the `asset worker queue` the dispatcher will prefer the `asset worker` to finish and release the other one.

philprime:
I found [this PR comment](https://github.com/getsentry/sentry-cocoa/pull/4194#discussion_r1700050798) and #4439 which I'll need to go through in detail again because the comments do include additional information on the choices made.

I'll try to summarize the reasonings and add more comments to the code itself, so it is documented for future maintenance.

philprime:
I also added more logging where relevant to debug issues

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@
 - Correctly rate limit envelopes from the new UI profiling system (#5131)
 - Race condition in ANRTrackerV1 (#5137)
 
+### Fixes
+
+- Fix thread inversion warning in session replay (#5018)
```
philipphofmann, line 23: `m`: There is already a fixes section for the unreleased section

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -121,6 +123,19 @@ - (void)setupWith:(SentryReplayOptions *)replayOptions
...
+    // The asset worker queue is used to work on video and frames data.
+    // Use a relative priority of -1 to make it lower than the default background priority.
+    _replayAssetWorkerQueue = [SentryDispatchQueueWrapper
+        createBackgroundDispatchQueueWithName:"io.sentry.session-replay.asset-worker"
+                             relativePriority:-1];
+    // The dispatch queue is used to asynchronously wait for the asset worker queue to finish its
+    // work. To avoid a deadlock, the priority of the processing queue must be lower than the asset
+    // worker queue. Use a relative priority of -2 to make it lower than the asset worker queue.
+    _replayProcessingQueue = [SentryDispatchQueueWrapper
+        createBackgroundDispatchQueueWithName:"io.sentry.session-replay.processing"
```
philipphofmann, line 137: `h`: The comment mentions that setting the wrong priorities here can lead to a deadlock. I think it's essential to have a test fail when using the wrong priorities to avoid regressions. We could assert that the `SentryDispatchQueueWrapper` is initialized with the correct attributes, which tests implementation details, and has the downside of having to change the test when the implementation details change. We could also have a test causing a deadlock when interacting with this class when the priorities are wrongly set, which could be a bit harder to achieve, but has the advantage of not testing implementation details.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -122,150 +141,257 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
...
+                    currentError = error
+                }
+                group.leave()
+            }
+
+            // Calling group.wait will block the `processingQueue` until the video rendering completes or a timeout occurs.
+            // It is imporant that the renderVideo completion block signals the group.
+            // The queue used by render video must have a higher priority than the processing queue to reduce thread inversion.
+            // Otherwise, it could lead to queue starvation and a deadlock/timeout.
+            guard group.wait(timeout: .now() + 120) == .success else {
```
philipphofmann, line 210: `m`: 120 seems a bit high to me. Shouldn't 30 be enough?

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -121,6 +123,19 @@ - (void)setupWith:(SentryReplayOptions *)replayOptions
     }
 
     _notificationCenter = SentryDependencyContainer.sharedInstance.notificationCenterWrapper;
+    _dateProvider = SentryDependencyContainer.sharedInstance.dateProvider;
+
+    // The asset worker queue is used to work on video and frames data.
+    // Use a relative priority of -1 to make it lower than the default background priority.
+    _replayAssetWorkerQueue = [SentryDispatchQueueWrapper
+        createBackgroundDispatchQueueWithName:"io.sentry.session-replay.asset-worker"
+                             relativePriority:-1];
```
philipphofmann, line 132: `h`: That's so easy to break. If we ever change the priority of the default background priority here, we could end up with the same problem again. And when I change the priority in the `SentryDispatchQueueWrapper` how should I know that I I need to change it here too?

https://github.com/getsentry/sentry-cocoa/blob/8856e5a7cf210e0417952d909d5924e76a7d1620/Sources/Sentry/SentryDispatchQueueWrapper.m#L10-L13

A test should fail if these priorities don't match our requirements here. If you need ideas for testing this properly, please let me know.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -122,150 +141,257 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
...
+                SentryLog.error("[Session Replay] Timeout while waiting for video rendering to finish.")
+                throw SentryOnDemandReplayError.errorRenderingVideo
```
philipphofmann, line 212: `m`: If we timeout, we could still return the videos we have and send what we have instead of discarding everything.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -122,150 +141,257 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
...
+            }
+
+            // If there was an error, throw it to exit the loop.
+            if let error = currentError {
+                SentryLog.error("[Session Replay] Error while rendering video: \(error), cancelling video segment creation")
+                throw error
+            }
```
philipphofmann, line 219: `m`: Same here, we could still return the videos we have and try to send them instead of throwing away all of them.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
philipphofmann, line 210: `m`: What drove you to add the DispatchGroup inside the `while frameCount < videoFrames.count {` and not outside? Now we have a timeout of 120 for every loop iteration, which can be a very long time. Shouldn't we have one timeout for the whole process?

philipphofmann (commented):
This is going in the right direction. I have a few questions on error handling and testing. Apart from that it already looks way better than before.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
philprime: Thanks for pointing this out! 🙏 
I had to set this higher because it was set to 2 seconds, and when debugging the timeout would always hit, but forgot to revert it afterwards.

Sources/Sentry/SentrySessionReplayIntegration.m
philprime: I would highly appreciate an example how you would test the priorities of queues.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
philprime: The timeout is wrong as mentioned in the other comment. It was at 2 seconds before, which is also a very tight window to do video conversion. I will reconsider a longer timeout for the entire process instead.

Sources/Sentry/SentrySessionReplayIntegration.m
philipphofmann: Using a factory method that I can swap out in the tests or something like that. Then I can assert that I call the factory with the correct parameters.

philprime: Thinking about this again, I am starting to question if another factory makes sense, because the `createBackgroundDispatchQueueWithName` already is a factory.
But I do see your point of this breaking easily and that we do not have tests to validate the behavior.

Therefore I am going to implement this:
- I'll add a new protocol `SentryDispatchQueueProviderProtocol` and an instance of `SentryDispatchQueueProvider`
- The provider will then use the `SentryDispatchQueueWrapper` under the hood to create dispatch queues
- This will make it replaceable in tests so we can assert that the queues are created with the correct arguments.
- The provider will live in the container, as there is no possibility to inject a dependency here, as we are already in the top-level integration

Sources/Sentry/SentrySessionReplayIntegration.m
philprime: See my answer to the other comments on how I'll make sure that SentryDispatchQueueWrapper is called correctly.

I am not sure yet if it's viable to create a failing test if the priorities are wrong, as we are doing this because the inner workings of AVFoundation are not transparent, but Xcode shows a warning.

Sources/Sentry/SentrySessionReplayIntegration.m
philprime: Nevermind, I just noticed that the `SentryDispatchQueueWrapper` already is used as an instance to create queues, so I'll just change the new method to be non-static

philprime: I just found `SentryDispatchFactory`, will extend this

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
philprime: I changed it to 10 seconds, reasonable amount for a video conversion

Sources/Sentry/SentrySessionReplayIntegration.m
philprime: alright, added the protocol and tests for the queues.
I actually did not mock the factory in the unit tests, but instead just checked the priorities of the queues underneath. Still believe there is value in adopting the factory pattern.

Sources/Sentry/SentrySessionReplayIntegration.m
philprime: As mentioned in the other PR comment, I added tests

philprime:
After looking at the internals, I noticed that we are propagating the error from the video creation to the caller, then just log and discard it because there's nothing else we can do.

I removed the propagation of the error and logged it directly `SentryOnDemandReplay.createVideoWith`, but instead return any previously generated videos.

This allowed me to get rid of multiple error handlers and try-throws

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Fix thread inversion warning in session replay (#5018)
```
philipphofmann, line 7: `m`: Yes this fixes the warning, but what's the customer benefit? Faster video conversion and no deadlocks or such? If yes, can we rephrase this?

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -638,6 +638,29 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
         XCTAssertEqual(writtenLastData, currentData)
     }
 
+    func testQueuePriorities_processingQueueShouldHaveLowerPriorityThanWorkerQueue() throws {
```
philipphofmann, line 641: Thanks for the test 🥇

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -135,164 +146,264 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
...
+            // The queue used by render video must have a higher priority than the processing queue to reduce thread inversion.
+            // Otherwise, it could lead to queue starvation and a deadlock/timeout.
+            guard group.wait(timeout: .now() + 10) == .success else {
+                SentryLog.error("[Session Replay] Timeout while waiting for video rendering to finish, returning \(videos.count) videos")
+                return videos
             }
+
+            // If there was an error, log it and exit the loop.
+            if let error = currentError {
+                // In previous implementations the error was propagated to the completion block, discarding any generated video.
```
philipphofmann, line 217: `l`: Mentioning a specific SDK version makes this a bit more specific, because if I want to find out which one, I need to do some git blaming.

philipphofmann (approved):
Thanks, LGTM

CHANGELOG.md
philprime: Updated, should be good now

PR #5021, opened 2025-03-25
test: Remove force_unwrap for SessionTests

philipphofmann:
#skip-changelog

### DIFF
--- a/Tests/SentryTests/SentrySessionTests.swift
+++ b/Tests/SentryTests/SentrySessionTests.swift
@@ -60,7 +60,7 @@ class SentrySessionTestsSwift: XCTestCase {
         XCTAssertNotEqual(session, copiedSession)
     }
     
-    func testInitWithJson_Status_MapsToCorrectStatus() throws {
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/SentrySessionTests.swift
```diff
@@ -60,7 +60,7 @@ class SentrySessionTestsSwift: XCTestCase {
         XCTAssertNotEqual(session, copiedSession)
     }
     
-    func testInitWithJson_Status_MapsToCorrectStatus() {
+    func testInitWithJson_Status_MapsToCorrectStatus() throws {
```
philprime, line 63: `m`: does this method throw?

philprime (approved):
Thank you for removing the force unwraps!

Tests/SentryTests/SentrySessionTests.swift
philipphofmann: Good catch, no it doesn't.

PR #5019, opened 2025-03-25
test: Remove force_unwrap for Serialization tests

philipphofmann:
#skip-changelog

PR #5028, opened 2025-03-26
test: Use non optional data for string to data

philipphofmann:
Prefer using Data(foo.utf8) over using foo.data(using: .utf8)

As discussed in https://github.com/getsentry/sentry-cocoa/pull/5019#discussion_r2012468356. 

We can enable the SwiftLint rule [non_optional_string_data_conversion](https://realm.github.io/SwiftLint/non_optional_string_data_conversion.html)  after merging https://github.com/getsentry/sentry-cocoa/pull/5019.  

#skip-changelog


philprime (approved):
LGTM, thank you for this improvement!

PR #5031, opened 2025-03-27
ci: Boot simulators early

philipphofmann:
Boot the simulator before compiling the UI tests to know when booting fails and to exit earlier.

#skip-changelog


.github/workflows/ui-tests.yml
```diff
@@ -94,14 +100,14 @@ jobs:
         uses: actions/upload-artifact@v4
         if: ${{ failure() }}
         with:
-          name: ui-tests-ios-swiftui-${{matrix.device}}.xcresult
```
philprime, line 97: `m`: We should keep some value of the matrix in this name, otherwise when adding another matrix entry the upload will warn/fail with a "duplicate artifact name" error

.github/workflows/ui-tests.yml
```diff
@@ -94,14 +100,14 @@ jobs:
...
+          name: ui-tests-ios-swiftui.xcresult
           path: fastlane/test_results/ui-tests-ios-swiftui.xcresult
 
       - name: Archiving Raw Test Logs
         uses: actions/upload-artifact@v4
         if: ${{ failure() || cancelled() }}
         with:
-          name: ui-tests-ios-swiftui-raw-output-${{matrix.device}}
+          name: ui-tests-ios-swiftui-raw-output
```
philprime, line 110: `m`: We should keep some value of the matrix in this name, otherwise when adding another matrix entry the upload will warn/fail with a "duplicate artifact name" error

.github/workflows/ui-tests.yml
```diff
@@ -63,22 +66,25 @@ jobs:
 
   # SwiftUI only supports iOS 14+ so we run it in a separate matrix here
   ui-tests-swift-ui:
-    name: UI Tests for SwiftUI on ${{matrix.device}} Simulator
```
philprime, line 66: `m`: Either leave any value of the matrix here, so they are not named identically when adding new matrix values, or remove the matrix entirely

.github/workflows/ui-tests.yml
```diff
@@ -22,18 +22,21 @@ concurrency:
 
 jobs:
   ui-tests:
-    name: UI Tests for ${{matrix.target}} on Simulators
+    name: UI Tests for ${{matrix.target}}
```
philprime, line 25: `m`: Either leave any value of the matrix here, so they are not named identically when adding new matrix values, or remove the matrix entirely

philipphofmann: I'm confused. I have the `${{matrix.target}}` in the name. What should I change?

.github/workflows/ui-tests.yml
philipphofmann: I removed the matrix.

.github/workflows/ui-tests.yml
philipphofmann: Good point. I removed the matrix.

.github/workflows/ui-tests.yml
philprime: My bad, it's fine like it is.

philprime (approved):
LGTM

PR #5033, opened 2025-03-27
chore: Remove destination for analyzer

philipphofmann:
We don't need to specify the destination and analyze with Xcode 16.2.

#skip-changelog.


philprime (approved):
LGTM, thanks

Issue #3905, opened 2024-04-26
Missing part of crash message


sindresorhus:
### Platform

macOS

### Environment

Production

### Installed

Swift Package Manager

### Version

8.24.0

### Steps to Reproduce

Some code crashes in production for violating the Swift Hashable requirement.

Crash report: https://sindresorhus.sentry.io/share/issue/ab8eb435cc71469082a07cdbece72d70/

### Expected Result

I expected to see the full error message.

### Actual Result

Sentry only shows part of it, making it impossible to track down the issue.

Sentry shows:

> ' were found in a Dictionary.

While it should have shown:

> Duplicate keys of type 'SomeKeyType' were found in a Dictionary.

Kinda looks like some sanitizing going wrong.

sindresorhus:
Another example (also Hashable violation):

> EKCale > Fatal error >
Stack overflow in (null)

Interestingly, this one doesn't show the `were found in a Dictionary.` part at all. And `EKCale` is a cut off `EKCalendar`.

https://sindresorhus.sentry.io/share/issue/0e4faa9089734b2bbe498a1e876fb30e/

philipphofmann:
The error message stems from the register x26 and ends up in the crash report under the JSON property `notable_addresses`.

When writing the crash report, we write notable registers here:
https://github.com/getsentry/sentry-cocoa/blob/fd1c9946004b5d02e51e802930ded740913ce5b8/Sources/SentryCrash/Recording/SentryCrashReport.c#L1012-L1027

I printed the raw value of the x26 register, and I always get.
```
' were found in a Dictionary.
This usually means either that the type violates Hashable's requirements, or
that members of such a dictionary were mutated after insertion.
```
instead of
```
Fatal error: Duplicate keys of type 'Something' were found in a Dictionary.
This usually means either that the type violates Hashable's requirements, or
that members of such a dictionary were mutated after insertion.
```

I also investigated if the methods isNotableAddress, or `isValidString` might do something wrong, but again I couldn't find anything, and it seems the the error message is truncated at the x26 register. So I tried checking the memory around the x26 register with changing `writeNotableRegisters` to the following:

```
static void
writeNotableRegisters(const SentryCrashReportWriter *const writer,
    const struct SentryCrashMachineContext *const machineContext)
{
    char registerNameBuff[30];
    const char *registerName;
    const int numRegisters = sentrycrashcpu_numRegisters();
    for (int reg = 0; reg < numRegisters; reg++) {
[...]

PR #1596, opened 2021-12-29
feat: Support for fatalError, assert, precondition

philipphofmann:
## :scroll: Description

The SDK was not able to retrieve the messages from Swifts fatalError, assert, or
precondition. This is fixed now by getting the message from the crashInfoMessage
of the binary image of libswiftCore.dylib.

The changes to SentryCrash for accessing the crashInfoMessage are merged from https://github.com/kstenerud/KSCrash/pull/375. Special thanks to @GLinnik21, who did the PR, and @kstenerud, who reviewed the PR.

Previously
Screen Shot 2021-12-29 at 11 15 55

https://sentry.io/organizations/sentry-sdks/issues/2862440420/events/a61f6937edd24abc9f4c5284ca50868d/?project=5428557

Now
Screen Shot 2021-12-29 at 11 14 56

https://sentry.io/organizations/sentry-sdks/issues/2893407482/events/72b601c512ff44f9a8d6ccbe3a403e22/?project=5428557

## :bulb: Motivation and Context

Fixes GH-662

## :green_heart: How did you test it?
Unit tests, simulator, and real devices.

## :pencil: Checklist

- [x] I reviewed the submitted code
- [x] I added tests to verify the changes
- [ ] I updated the docs if needed
- [x] Review from the native team if needed
- [x] No breaking changes

## :crystal_ball: Next steps
Update feature list of Apple in docs https://docs.sentry.io/platforms/apple/guides/ios/

PR #5036, opened 2025-03-28
fix: Truncation of Swift crash messages

philipphofmann:
## :scroll: Description

With GH-1596, we introduced caching binary images to avoid calling async unsafe methods when writing crash reports. This [PR](https://github.com/getsentry/sentry-cocoa/pull/2939) removed getting the crash info from the binary images when crashing. This is fixed now by calling getCrashInfo while writing the binary images for a crash report.

Fixes GH-3905

Manually by triggering a crash in the SwiftiOS sample app. I also tested other types of crashes.

https://sentry-sdks.sentry.io/issues/6540727731/events/65a844dee9a74581babccce1e29edce9/?project=5428557
Screenshot 2025-04-15 at 11 33 07

Normal fatalError message still works: https://sentry-sdks.sentry.io/issues/6536640165/events/a4d56ab9b7184b658e882ff4b4f1d7aa/?project=5428557

Force unwrap optional still works: https://sentry-sdks.sentry.io/issues/6202264696/events/2756e9db5b484b97be4a311a7c1ce188/?project=5428557

Normal crash still works: https://sentry-sdks.sentry.io/issues/6044162827/events/1de102c17620494fb2e7f6c8ee9ce856/?project=5428557

## :pencil: Checklist


philipphofmann:
For a the Testfllight build I sadly don't get the assertion message. We simply get `throwFatalDuplicateKeyError`.

Screenshot 2025-04-11 at 15 25 06

But we get the fatal error message of directly calling `fatalError`. I need to further investigate why that is

https://sentry-sdks.sentry.io/issues/6536640165/?project=5428557&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=0

Screenshot 2025-04-11 at 15 27 08

philipphofmann:
I still experience weird behaviour. Sometimes the Swift error message is in the `__crash_info` and sometimes not. At least the problem is reproducable on both simulator and a real device for release builds. Either Swift isn't writing the error message to `__crash_info` consistently or we fail to read it. I need to further investigate what causes this problem.

philipphofmann:
Luckily, seeing Swift assert messages, sometimes there and sometimes not, was a false positive. The sample code for creating the assertion failure sometimes failed when initializing the dict, and then Swift doesn't write an assertion message, but instead we get the following error:

Screenshot 2025-04-15 at 11 29 30

Anyways, I really double checked our implementation of getting the crash info and it is legit.

Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -477,32 +477,6 @@ - (void)enhanceValueFromNotableAddresses:(SentryException *)exception
...
- * https://github.com/apple/swift/blob/d1bb98b11ede375a1cee739f964b7d23b6657aaf/stdlib/public/runtime/Errors.cpp#L365-L377
- * @seealso
- * https://github.com/apple/swift/blob/d1bb98b11ede375a1cee739f964b7d23b6657aaf/stdlib/public/runtime/Errors.cpp#L361
- * @seealso
- * https://github.com/apple/swift/blob/d1bb98b11ede375a1cee739f964b7d23b6657aaf/stdlib/public/runtime/Errors.cpp#L269-L293
- * @seealso
- * https://github.com/apple/swift/blob/d1bb98b11ede375a1cee739f964b7d23b6657aaf/stdlib/public/runtime/Errors.cpp#L264-L293
- * @seealso
- * https://github.com/apple/swift/blob/d1bb98b11ede375a1cee739f964b7d23b6657aaf/include/swift/Runtime/Debug.h#L29-L58
- */
```
philipphofmann, line 505: Moved to `Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c`.

Samples/iOS-Swift/iOS-Swift/ErrorsViewController.swift
```diff
@@ -84,6 +84,15 @@ class ErrorsViewController: UIViewController {
         fatalError("This is a fatal error. Oh no 😬.")
     }
 
+    @IBAction func throwFatalDuplicateKeyError(_ sender: Any) {
```
philprime, line 87: `l`: It would be helpful to have more explanation how this test case works, e.g. why does it cause a Duplicate Key error instead of just updating the existing value?
I believe explaining the sample cases could help us in the future to find similar cases for reported issues

Samples/macOS-SwiftUI/macOS-SwiftUI/ContentView.swift
```diff
@@ -15,7 +15,18 @@ struct ContentView: View {
             Button(action: reportNSException) {
                 Text("Report NSException")
             }
-            
+
+            Button(action: {
+                // Triggers: Fatal error: Duplicate keys of type 'Something' were found in a Dictionary.
```
philprime, line 20: `l`: See my comment in `ErrorsViewController.swift`

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
```diff
@@ -293,6 +293,50 @@ isValidCrashInfoMessage(const char *str)
...
+ * https://github.com/apple/swift/blob/d1bb98b11ede375a1cee739f964b7d23b6657aaf/stdlib/public/runtime/Errors.cpp#L269-L293
+ * @seealso
+ * https://github.com/apple/swift/blob/d1bb98b11ede375a1cee739f964b7d23b6657aaf/stdlib/public/runtime/Errors.cpp#L264-L293
+ * @seealso
+ * https://github.com/apple/swift/blob/d1bb98b11ede375a1cee739f964b7d23b6657aaf/include/swift/Runtime/Debug.h#L29-L58
+ *
+ * We also investigated using getsectbynamefromheader to get the crash info as Crashlytics
+ * (https://github.com/firebase/firebase-ios-sdk/blob/main/Crashlytics/Crashlytics/Components/FIRCLSBinaryImage.m#L245-L283)
+ * does it, but we get the same results. If the crash info is missing, we can't find it via
+ * getsectiondata or getsectbynamefromheader. We also saw Swift storing the error message into the
```
philprime, line 325: Could it be that the full crash message is actually somewhere else?

Tests/SentryTests/Integrations/SentryCrash/SentryCrashReportTests.swift
```diff
@@ -169,7 +171,26 @@ class SentryCrashReportTests: XCTestCase {
...
     }
-    
+
+#if os(iOS) && !targetEnvironment(macCatalyst)
+    // We can't really test writing the crash_info_message unless there is an actual crash.
+    // We noticed that the libsystem_sim_platform.dylib is a simulator library that has a crash info message even when there is no crash.
+    // This is simply a smoke test to ensure we can write the crash_info_message key to the report.
+    // This only works on iOS and also not macCatalyst.
+    // Actual testing must be done manually with a fatalError in Swift.
+    func testWriteCrashReport_ContainsCrashInfoMessage() throws {
```
philprime, line 181: `l`: Do you think there could be value in creating a workflow/test-runner-system which actually starts the sample app in the simulator in CI and let it crash, to then look at the data written to disk outside of Xcode? Maybe using something like NPM jest as the test runner, executing shell scripts like we do in the sentry-wizard E2E tests

philprime (approved):
LGTM, thank you for looking into this!

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
philipphofmann: After my investigation and research, the probability is low. I also tested this with a sample app on Crashlytics, and they also didn't report the full crash message. If you have any idea, please let me know.

Tests/SentryTests/Integrations/SentryCrash/SentryCrashReportTests.swift
philipphofmann: That should be `h` 😄. Yes indeed, but that's not straightforward. I opened an issue: https://github.com/getsentry/sentry-cocoa/issues/5096.

PR #5043, opened 2025-04-02
fix: add condtion to reference UIScreen

soo941226:
## :scroll: Description

- add a condition to reference UIScreen in `SentryCrashWrapper`

At first, I need your help. I'm using this SDK at face a runtime error like below.
스크린샷 2025-04-02 오후 2 07 09

I'm not familiar with Objective-C. and I can't understand your full codes, context, and philosophy. furthermore I'm not sure it's the problem only for session tracking.
If you agree this is the problem, could you help me to solve this problem?
thank you.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryCrashWrapper.m
+++ b/Sources/Sentry/SentryCrashWrapper.m
@@ -1,4 +1,5 @@
 #import "SentryCrashWrapper.h"
+#import "Sentry/SentryOptions.h"
 #import "SentryCrash.h"
 #import "SentryCrashBinaryImageCache.h"
 #import "SentryCrashIntegration.h"
@@ -106,7 +107,7 @@ NS_ASSUME_NONNULL_BEGIN
     sentrycrashbic_stopCache();
 }
 
-- (void)enrichScope:(SentryScope *)scope
+- (void)enrichScope:(SentryScope *)scope withOption:(SentryOptions *)options;
 {
     // OS
     NSMutableDictionary *osData = [NSMutableDictionary new];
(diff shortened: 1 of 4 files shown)


philprime:
Hi @soo941226, thank your opening up a pull request with improvements!

I'll take a look at it soon and get back to you with feedback

philprime:
@soo941226 looking at your changes and the screenshot of the crash report, I have a few questions:

- Can you please share your `SentrySDK.configure`? It seems to me that `SentrySDK.capture(error:)` is called before the `scope` is configured during bootstrapping
- Can you confirm that you call `SentrySDK.capture(error:)` from a background thread?
- Do you call `SentrySDK.capture(error:)` without bootstrapping the SDK first?

Your changes only silence the "errors", the actually problem is that `enrichScope` is called from a background thread, but that can only happen if the SentrySDK client is not initialized.

Sources/Sentry/SentryCrashWrapper.m
```diff
@@ -183,16 +184,18 @@ - (void)enrichScope:(SentryScope *)scope
...
+    if (options.enableAutoSessionTracking) {
+        NSArray<UIWindow *> *appWindows
+            = SentryDependencyContainer.sharedInstance.application.windows;
+        if ([appWindows count] > 0) {
+            UIScreen *appScreen = appWindows.firstObject.screen;
+            if (appScreen != nil) {
+                [deviceData setValue:@(appScreen.bounds.size.height)
+                              forKey:@"screen_height_pixels"];
+                [deviceData setValue:@(appScreen.bounds.size.width) forKey:@"screen_width_pixels"];
+            }
```
philprime, line 196: `h`: this option is not relevant here, we should ensure the `enrichScope` is called from the main thread instead.

philprime (changes requested):

Sources/Sentry/SentryCrashWrapper.m
philipphofmann: Yes, indeed. The `enrichScope` is called on the main thread when starting the SDK. We could also cache `screen_height_pixels` and `screen_width_pixels`.

philprime:
Happy to hear that and thank you that you considered improving our SDK!

PR #5047, opened 2025-04-03
feat(ci): add scheduled workflow to update clang-format and swiftlint

philprime:
## :scroll: Description

- Adds a scheduled GitHub Actions workflow to automatically update clang-format and swifltint via a pull request.
- Adds relative path resolving to reference version files relative to the script files for `check-tooling-versions.sh
- Adds utility script `update-tooling-versions.sh`
- Adds `make init-ci-format` with a `Brewfile` **including swiftlint, because it is not pre-installed on macos-15 runners**

Running `make init` writes the version of `clang-format` and `swiftlint` to files in `./scripts/.clang-format-version` and `./scripts/.swiftlint-version`.

AFAIK we do not have any automation, so I added this one.

Manually dispatched the workflow via UI.
See #5049 for an example (leaving it open until this PR is merged)

#skip-changelog


.github/workflows/format-code.yml
```diff
@@ -2,30 +2,31 @@ name: format code
...
+      - "Samples/**"
+      - ".github/workflows/format-code.yml"
 
 jobs:
   # Formats Swift, Objective-C, C++, and C code and commits the formatted code
   # if necessary
   format-code:
     name: Format Code
-    runs-on: macos-13
+    runs-on: macos-15
```
philprime, line 16: Using the opportunity to bump this to macos-15, will require installation of `clang-format`, `swiftlint` and `rbenv` as these are not preinstalled anymore.

Makefile
```diff
@@ -24,6 +23,16 @@ init-ci-test:
 init-ci-deploy:
 	brew bundle --file Brewfile-ci-deploy
 
+# installs the tools needed to run CI format tasks locally (note that swiftlint is preinstalled in github actions)
```
philprime, line 26:
```suggestion
# installs the tools needed to run CI format tasks locally
```

philipphofmann (commented):
We use the [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated) to update other dependencies, which is configured in the `dependabot.yml`. What drove you to update the dependencies manually instead of using the dependabot?

philprime:
Maybe I do not know enough about Dependabot. Can it be configured to execute custom scripts and update custom version files like we do for `.clang-format-version` and `.swiftlint-version`?

For my personal projects I've been using [Renovate](https://docs.renovatebot.com/) with custom regex managers for years due to limited capabilities of dependabot, but maybe Dependabot caught up in functionality.

philipphofmann:
> Can it be configured to execute custom scripts and update custom version files like we do for .clang-format-version and .swiftlint-version?

No idea. We use the dependabot in most sentry repos. We try to minimize our dependencies on non-GitHub maintained GH actions as much as possible due to security risks. Also sticking to one way to update our dependencies reduces complexity. It would be great to evaluate if we can use the dependabot for this. If not, we should quickly document in our Cocos SDK develop docs decision log, why we chose a custom script to update these dependencies.

.github/workflows/auto-update-tools.yml
```diff
@@ -0,0 +1,55 @@
+name: "Automation: Update tooling versions"
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
```
philipphofmann, line 6: `m`: I think it makes sense to explain directly here why we can't use the dependabot.

philipphofmann (approved):
LGTM, when explaining why we can't use the dependabot.

PR #5048, opened 2025-04-03
fix: replace associated object with current TTD tracker

philprime:
## :scroll: Description

- Uses the `currentTTDTracker` instead of an associated object

We had a an edge case reported where a view controller was dismissed, and in the completion handler the methods `viewDidLoad()` and `viewDidLayoutSubviews()` were called. This methods were swizzled and triggered a TTID report.

As the view controller was already being deallocated, a race-condition occurred and the `objc_getAssociatedObject` accessed a memory with "random" objects instead, leading to an SDK crash.

TODO

## :pencil: Checklist


philprime:
Progress update:

- Added UI tests to reproduce the crash
- Added a fix to stop the crashing
- `SentryUIViewControllerPerformanceTrackerTests.test_waitForFullDisplay_NewViewControllerLoaded_BeforeReportTTFD` is failing --> probably due to refactoring
- Otherests in `SentryUIViewControllerPerformanceTrackerTests` are not failing if executed individually, but fail if executed all at once --> indicates that the test is not cleaning up correctly.

philprime:
Blocked by #5162 and #5163

philprime:
I reverted the changes of moving the performance tracker which are done in #5163 to see if I can get this fix on track without refactoring the shared dependency.

Tests/SentryTests/Integrations/Performance/UIViewController/SentryUIViewControllerPerformanceTrackerTests.swift
```diff
@@ -399,28 +399,6 @@ class SentryUIViewControllerPerformanceTrackerTests: XCTestCase {
         XCTAssertEqual(tracer?.children.filter { $0.operation.contains("initial_display") }.count, 0, "Tracer must not contain a TTID span")
         XCTAssertEqual(tracer?.children.filter { $0.operation.contains("full_display") }.count, 0, "Tracer must not contain a TTFD span")
     }
-
-    func testSecondViewController() {
```
philprime, line 403: IIUC this test only checks if the view controller has a tracker associated. This way of storing a reference to the TTD tracker is outdated, so I removed the test

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -40,17 +54,15 @@ - (instancetype)initWithTracker:(SentryPerformanceTracker *)tracker
...
+
+        _ttdTrackers = [NSMapTable weakToStrongObjectsMapTable];
+        _spanIds = [NSMapTable weakToStrongObjectsMapTable];
+        _spansInExecution = [NSMapTable weakToStrongObjectsMapTable];
+        _layoutSubviewSpanIds = [NSMapTable weakToStrongObjectsMapTable];
     }
     return self;
 }
 
-- (SentrySpan *)viewControllerPerformanceSpan:(UIViewController *)controller
```
philprime, line 47: Used for only one case --> inlined the code

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -18,12 +18,26 @@
 #    import <UIKit/UIKit.h>
 #    import <objc/runtime.h>
 
+// Instead of using associated objects, we use a global map to store the time to display tracker,
+// spanId, spans in execution, and layout subview spanId to avoid memory issues with associated
+// objects accessed from different threads.
+//
+// Using a NSMapTable allows weak references to the keys, which means we don't need to remove the
+// entries when the UIViewController is deallocated.
```
philipphofmann, line 26: `l`: I would put that comment directly above the NSMapTable properties, so they are colocated.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -18,12 +18,26 @@
 #    import <UIKit/UIKit.h>
 #    import <objc/runtime.h>
 
+// Instead of using associated objects, we use a global map to store the time to display tracker,
+// spanId, spans in execution, and layout subview spanId to avoid memory issues with associated
```
philipphofmann, line 22: `l`: I think we could change the wording here to state that in an earlier version we used associated objects and explain the problems we had with it.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -469,11 +466,104 @@ - (void)measurePerformance:(NSString *)description
...
+// - MARK: - Getter and Setter Helpers
+
+- (SentryTimeToDisplayTracker *_Nullable)getTimeToDisplayTrackerForController:
+    (UIViewController *)controller
+{
+    SENTRY_LOG_DEBUG(
+        @"[UIViewController Performance] Getting time to display tracker for controller: %@",
+        controller);
+    // Use a global map to store the time to display tracker to avoid memory issues with associated
+    // objects.
```
philipphofmann, line 492: `l`: I think mentioning once that we had memory issues with associated object at the top of the class is enough. IMO, we can remove these comments here. It's mostly noise.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
```diff
@@ -40,17 +54,15 @@ - (instancetype)initWithTracker:(SentryPerformanceTracker *)tracker
 
         _alwaysWaitForFullDisplay = NO;
         _dispatchQueueWrapper = dispatchQueueWrapper;
+
+        _ttdTrackers = [NSMapTable weakToStrongObjectsMapTable];
+        _spanIds = [NSMapTable weakToStrongObjectsMapTable];
+        _spansInExecution = [NSMapTable weakToStrongObjectsMapTable];
+        _layoutSubviewSpanIds = [NSMapTable weakToStrongObjectsMapTable];
```
philipphofmann, line 61: `h`: The docs of `weakToStrongObjectsMapTable` state:

> Use of weak-to-strong map tables is not recommended. The strong values for weak keys which get zeroed out continue to be maintained until the map table resizes itself. - https://developer.apple.com/documentation/foundation/nsmaptable/weaktostrongobjects()?language=objc#Discussion

If I'm not mistaken, we never call `removeObjectForKey` on any of these `NSMapTable`. With this approach ,we depend on the NSMapTable to resize itself, which isn't guaranteed to happen at some point. This usually happens when adding entries or the `NSMapTable` has so many entries that it needs to resize. This means that we store plenty of tddTrackers in memory until they get deallocated. Add the following log statement to `getTimeToDisplayTrackerForController`:
```
SENTRY_LOG_DEBUG(@"[UIViewController Performance] size: %lu", self.ttdTrackers.count);
```

Then launch the iOSSwift sample app, open a bunch of UIViewControllers and the log output proves my theory
```
[Sentry] [debug] [timeIntervalSince1970:1747142571.015678] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 0
[Sentry] [debug] [timeIntervalSince1970:1747142571.0505772] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 1
[Sentry] [debug] [timeIntervalSince1970:1747142576.089128] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 1
[Sentry] [debug] [timeIntervalSince1970:1747142576.358041] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 2
[Sentry] [debug] [timeIntervalSince1970:1747142577.219908] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 2
[Sentry] [debug] [timeIntervalSince1970:1747142577.268588] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 3
[Sentry] [debug] [timeIntervalSince1970:1747142578.243167] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 3
[Sentry] [debug] [timeIntervalSince1970:1747142578.3024678] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 4
[Sentry] [debug] [timeIntervalSince1970:1747142584.534718] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 4
[Sentry] [debug] [timeIntervalSince1970:1747142584.59974] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 5
[Sentry] [debug] [timeIntervalSince1970:1747142589.257824] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 5
[Sentry] [debug] [timeIntervalSince1970:1747142589.298486] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 6
[Sentry] [debug] [timeIntervalSince1970:1747142594.11501] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 6
[Sentry] [debug] [timeIntervalSince1970:1747142596.76362] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 7
[Sentry] [debug] [timeIntervalSince1970:1747142596.794146] [SentryUIViewControllerPerformanceTracker:491] [UIViewController Performance] size: 7
```

philipphofmann (commented):
I added a couple of minor comments, but we have one major issue to address. The current approach eats plenty of memory. We have to clean up the objects in `NSMapTable` at some point.

Samples/iOS-Swift/iOS-Swift/ViewControllers/ViewLifecycleTestViewController.swift
```diff
@@ -0,0 +1,30 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        print("This statement is printed in viewDidLoad")
+
+        // We need to ensure that view controller swizzling is active.
+        // Otherwise this test is rather useless.
+        let stacktraceContainsUIViewControllerSwizzling = Thread.callStackSymbols.contains { frame in
+            frame.contains("SentryUIViewControllerSwizzling swizzleViewDidLoad")
+        }
+        assert(stacktraceContainsUIViewControllerSwizzling, "viewDidLoad should be swizzled")
```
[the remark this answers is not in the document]
philprime: This is a pre-condition. iOS-Swift can be started without swizzling, in case the `SentrySDKWrapper` is changed (by accident) or on purpose.
As this view controller is used in tests using launch arguments, we need to assert that they are actually passed all the way to this method.
If swizzling is disabled, this test can not crash as it should without a fixed implementation.

Samples/iOS-Swift/iOS-Swift-UITests/ViewLifecycleUITests.swift
```diff
@@ -0,0 +1,36 @@
+import XCTest
+
+class ViewLifecycleUITests: BaseUITest {
+    override var automaticallyLaunchAndTerminateApp: Bool { false }
```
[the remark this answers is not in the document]
philprime: We need this to be able to define the launch arguments in `setUp`. If we do not set it, it will launch the apps with the arguments in the `BaseUITest`

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
philprime: It's amazing how it looses the entire purpose of having weak references.
I was expecting it to clean up by itself. Thanks for pointing this out.

I'll investigate how hight the memory impact actually is. Having "dead" refs in the table might have a neglectable impact.

philprime: Okay it seems I can stop further research, NSMapTable is not suitable for this use-case.
http://cocoamine.net/blog/2013/12/13/nsmaptable-and-zeroing-weak-references/

philipphofmann: > I was expecting it to clean up by itself. Thanks for pointing this out.

That's what code reviews are for 😄

philipphofmann:
> every time we access the map it will trigger pruning.

That's the most reliable way to do that, IMO. I don't think debouncing is required, as these operations should be fast.

philprime:
Further inspection of the NSTableMap showed that it uses some internal heuristic to clean up the storage, therefore causing the `count` to be larger than the actual content of the map.

I created a wrapper `SentryWeakMap` with automatic pruning on access or mutation, and replaced the internal storage with a dictionary.

Sources/Sentry/SentryUIViewControllerPerformanceTracker.m
philprime: Using the new weak map the storage is now cleaned up properly on every access or mutation.

antonis (approved):
I went through the implementation and the previous review feedback and everything LGTM 🚢 

> I created a wrapper `SentryWeakMap` with automatic pruning on access or mutation, and replaced the internal storage with a dictionary.

The `SentryWeakMap` looks solid too me 👍 

Thank you for your work on this @philprime 🙇

philprime:
Thank you @antonis for the review!

PR #5050, opened 2025-04-03
chore: add flag to disable app hang tracking v2 in sample

philprime:
I just noticed we are not enabling ANR v2 in the sample (maybe on purpose?) and also don't offer a launch argument to disable it.

#skip-changelog


Samples/iOS-Swift/iOS-Swift/SentrySDKWrapper.swift
```diff
@@ -60,6 +60,7 @@ struct SentrySDKWrapper {
         options.attachViewHierarchy = enableAttachViewHierarchy
         options.enableTimeToFullDisplayTracing = enableTimeToFullDisplayTracing
         options.enablePerformanceV2 = enablePerformanceV2
+        options.enableAppHangTrackingV2 = enableAppHangTrackingV2
```
armcknight, line 63: Needs the launch arg added to the scheme. I added it in b41924378

philprime: Thank you!

philprime:
@philipphofmann just to double-check, is there a reason v2 is not enabled in the sample?

philipphofmann:
> @philipphofmann just to double-check, is there a reason v2 is not enabled in the sample?

No specific reason, no. Could you enable it per default in this PR please?

PR #5052, opened 2025-04-03
test(iOS-Swift): in app profile options

armcknight:
Adds in-app configurability of the same things that launch args and environment variables can be used for when testing the various configuration options for profiling. This allows us to test different configurations with builds of the sample app from TestFlight, as generated from Release workflow runs, as a final acceptance test check before sending out an SDK release. It also serves as a helpful visual indicator of how profiling was configured for the given launch.

Uses UserDefaults as an override point that takes precedence over launch args and env variable values.

Also, the Profiling view controller was getting crowded, so I split out the stuff needed for the profiling benchmark tests into a new Benchmarking view controller with its own tab bar tab.

|before|after (profiling tab)|after (benchmarking tab)|
|---|---|---|
|Simulator Screenshot - iPhone SE (3rd generation) - 2025-04-04 at 14 40 24|Simulator Screenshot - iPhone SE (3rd generation) - 2025-04-04 at 14 38 48|Simulator Screenshot - iPhone SE (3rd generation) - 2025-04-04 at 14 38 50|

#skip-changelog


Samples/iOS-Swift/iOS-Swift/Profiling/BenchmarkingViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+                cpuWorkthreads.insert(thread)
+                thread.start()
+            }
+        } else {
+            let absDiff = abs(diff)
+            for _ in 0 ..< absDiff {
+                let thread = cpuWorkthreads.removeFirst()
+                thread.cancel()
+            }
+        }
```
philprime, line 116: `l`: nit pick to make early exit more visible

```suggestion
        if diff == 0 {
            return
        }
        
        if diff > 0 {
            for _ in 0 ..< diff {
                let thread = WorkThread()
                thread.qualityOfService = .userInteractive
                cpuWorkthreads.insert(thread)
                thread.start()
            }
        } else {
            let absDiff = abs(diff)
            for _ in 0 ..< absDiff {
                let thread = cpuWorkthreads.removeFirst()
                thread.cancel()
            }
        }
```

Samples/iOS-Swift/iOS-Swift/SentrySDKWrapper.swift
```diff
@@ -420,30 +420,30 @@ extension SentrySDKWrapper {
     }
     
     var profilesSampleRate: NSNumber? {
-        if args.contains("--io.sentry.enableContinuousProfiling") {
```
philprime, line 423: `l`: nit pick, might be easier/debug to read with early-exit pattern:

```swift
        if SentrySDKOverrides.Profiling.useContinuousProfilingV1 {
            return nil
        }
        if let profilesSampleRateOverride = SentrySDKOverrides.Profiling.sampleRate {
            return NSNumber(value: profilesSampleRateOverride)
        }
        return 1
```

philprime (approved):
The code changes look good to me, but can you maybe create a writeup how the benchmarking view works, what the sliders do/purpose of the inputs and what the outputs mean? A simple "user manual"/readme could be helpful.

PR #5054, opened 2025-04-04
chore(session-replay): GA of better session replay view renderer V2

philprime:
- After our evaluation phase we should enable the experimental view renderer by default to roll out to a larger audience.

cc @kahest @krystofwoldrich


CHANGELOG.md
```diff
@@ -8,7 +8,14 @@
 
 ### Improvements
 
+> ![Important]
+> This version enables the experimental view renderer used by Session Replay by default.
```
philipphofmann, line 12: `h`: Rolling out a renderer called experimental sounds a bit weird to me. Maybe we rename it to V2 or something.

philprime: It's still an experimental view renderer just enabled by default. Why should this be renamed if it is still in experimental state?

philipphofmann: It feels weird to enable something experimental per default for a GA product. If we're confident enough to enable it per default for everyone it shouldn't be experimental anymore IMO.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -107,17 +107,16 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+     * **up to 4-5x faster rendering** (reducing `~160ms` to `~36ms` per frame) on older devices.
      *
-     * - Experiment: This is an experimental feature and is therefore disabled by default. In case you are noticing issues with the experimental
-     *               view renderer, please report the issue on [GitHub](https://github.com/getsentry/sentry-cocoa). Eventually, we will
-     *               remove this feature flag and use the experimental view renderer by default.
+     * - Experiment: In case you are noticing issues with the new view renderer, please report the issue on [GitHub](https://github.com/getsentry/sentry-cocoa). 
+     *               Eventually, we will remove this feature flag and use the new view renderer by default.
      */
-    public var enableExperimentalViewRenderer = false
+    public var enableViewRendererV2 = true
```
krystofwoldrich, line 119: Might be worth considering only deprecating `enableExperimentalViewRenderer`, making it an alias for `enableViewRendererV2`, so we don't break apps which used it.

philprime: Added the `enableExperimentalViewRenderer` as an deprecated alias back for `enableViewRendererV2`

philipphofmann (approved):
LGTM

philprime:
Blocked by #5099

philprime:
Not blocked anymore, will need to update this soon.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+> ![Important]
+> This version makes the new view renderer used by Session Replay generally available.
```
philipphofmann, line 6: `m`: I think it is essential to briefly mention that the V2 renderer is better than V1. Something like this could work.
```suggestion
> This version enables the better view renderer V2 used by Session Replay generally available and enables it by default.
```

CHANGELOG.md
```diff
@@ -11,6 +17,7 @@
 - More logging for Session Replay video info (#5132)
 - Improve session replay frame presentation timing calculations (#5133)
 - Use wider compatible video encoding options for Session Replay (#5134)
+- GA of new session replay view renderer (#5054)
```
philipphofmann, line 20: `m`: Please change the working to V2 and also the PR title.

philipphofmann (approved):
Yeah GA :shipit: LGTM

CHANGELOG.md
philprime: Not sure if I understand correctly, is the updated version ok?

CHANGELOG.md
philprime: Happy to iterate, is the updated version okay?

PR #5055, opened 2025-04-04
chore(session-replay): add sample app with AVFoundation preview layer

### LINKED ISSUE #4883: Assess and improve session replay performance using camera view finder
### Description

- We need to assess how session replay recording affects video capturing previews, especially using AVFoundation and [AVCaptureVideoPreviewLayer](https://developer.apple.com/documentation/avfoundation/avcapturevideopreviewlayer)
- Masking preview or disabling recording of preview could be complicated, as it is implemented as a [CALayer](https://developer.apple.com/documentation/QuartzCore/CALayer), rather than UIView

philprime:
Adds a sample project with a simple preview layer to view camera input.
This was used to test session replay with camera inputs.

Closes #4883 

#skip-changelog


philipphofmann (approved):
LGTM, maybe it makes sense to compile the sample in CI to ensure we don't break it by mistake.

philipphofmann (approved):
Again LGTM

.github/workflows/benchmarking.yml
```diff
@@ -7,16 +7,16 @@ on:
   pull_request:
     paths:
       # test changes to Sentry SDK sources
-      - 'Sources/**'
```
armcknight, line 10: please split these kinds of lint diffs into separate PRs in the future

armcknight:
@philprime why create a separate sample app project for this instead of adding it to iOS-Swift?

philprime:
@armcknight simple lightweight examples, iOS-Swift is massive with tons of debug output and all the features enabled. This example was only there to focus one one thing: testing performance of an app with a AVFoundation preview layer.

Also there is no point in running this app in UI tests, because there will not be any camera available, therefore adding it to the iOS-Swift example would just bloat it up

.github/workflows/benchmarking.yml
philprime: I am actually going to do it for this PR too.

philprime: https://github.com/getsentry/sentry-cocoa/pull/5072

armcknight:
The bloat factor is worse when we have more project settings to maintain and more CI config and compute time with each separate project.

philprime:
@armcknight I disagree with the general sentiment, as this sample app has nothing to do with the others, therefore it should also be isolated with its code and settings. This sample app is also not going to be used as a daily driver for UI tests or similar, and will most likely also not be adapted over time. If it passes CI now, but eventually doesn't, that would actually mean we messed up somewhere.

But I do agree that additional CI config is not beneficial, therefore I merged it with the existing building of sample apps.

PR #5062, opened 2025-04-08
test(iOS-Swift): default to ui profiling, trace lifecycle and app sta…

armcknight:
Iterating on the in-app override work done in https://github.com/getsentry/sentry-cocoa/pull/5052

I realized after doing some testing that the in-app profiling overrides were trying to do too much state management to avoid "invalid" configurations, where really we should be able to test every combination of settings.

Also, this defaults iOS-Swift and iOS-ObjectiveC to use the new UI profiling stuff, with overrides for all the prior, deprecated profiling methods.

Also also correctly distinguishes between a falsy value written to user defaults vs the key not existing, which also returns false by default, but we need to be able to tell the two use cases apart. And, introduces a way to make the schema args/vars take precedence over the user defaults settings. There are valid reasons to want to do it both ways, so I'd rather do this than simply change the precedence order.

#skip-changelog


Samples/iOS-ObjectiveC/iOS-ObjectiveC/AppDelegate.m
```diff
@@ -33,39 +33,32 @@ - (BOOL)application:(UIApplication *)application
...
+        if (env[@"--io.sentry.profilesSamplerValue"] != nil) {
+            options.profilesSampler
+                = ^NSNumber *_Nullable(SentrySamplingContext *_Nonnull samplingContext)
+            {
+                return @([env[@"--io.sentry.profilesSamplerValue"] floatValue]);
+            };
+        }
+
+        if ([args containsObject:@"--io.sentry.disable-app-start-profiling"]) {
+            options.enableAppLaunchProfiling = YES;
```
indragiek, line 61: This looks flipped, the option says disable but you're enabling it

philprime: +1

Seems that you want `if (![args containsObject...`

Samples/iOS-ObjectiveC/iOS-ObjectiveC/AppDelegate.m
```diff
@@ -33,39 +33,32 @@ - (BOOL)application:(UIApplication *)application
...
+
+        if (env[@"--io.sentry.profilesSamplerValue"] != nil) {
+            options.profilesSampler
+                = ^NSNumber *_Nullable(SentrySamplingContext *_Nonnull samplingContext)
+            {
+                return @([env[@"--io.sentry.profilesSamplerValue"] floatValue]);
+            };
+        }
+
+        if ([args containsObject:@"--io.sentry.disable-app-start-profiling"]) {
```
philprime, line 60:
```suggestion
        if (![args containsObject:@"--io.sentry.disable-app-start-profiling"]) {
```

Samples/iOS-Swift/iOS-Swift/SentrySDKOverrides.swift
```diff
@@ -76,53 +72,82 @@ enum SentrySDKOverrides {
...
             get {
-                getOverride(for: Profiling.Key.launchProfiling)
+                !getOverride(for: Profiling.Key.disableAppStartProfiling)
             }
             set(newValue) {
-                UserDefaults.standard.set(newValue, forKey: Profiling.Key.launchProfiling.rawValue)
+                setOverride(for: Profiling.Key.disableAppStartProfiling, value: !newValue)
             }
         }
+    }
```
philprime, line 94: `l`: Please stay consistent in code style, either use explicit returns or not, both are fine (depending on our Swift lint setup)

philprime (approved):
LGTM, left some small comments you might want to consider to resolve first.

Samples/iOS-Swift/iOS-Swift/Profiling/ProfilingViewController.swift
```diff
@@ -182,4 +145,45 @@ class ProfilingViewController: UIViewController, UITextFieldDelegate {
...
+            tracesSampleRateField.text = String(format: "%.2f", sampleRate.floatValue)
+            tracesSampleRateField.isEnabled = true
+            tracesSampleRateSwitch.isOn = true
+        } else {
+            tracesSampleRateField.text = "nil"
+            tracesSampleRateField.isEnabled = false
+            tracesSampleRateSwitch.isOn = false
+        }
+
+        if let v2Options = options.profiling {
```
philipphofmann, line 172: `l`: This reads a bit better, IMO.
```suggestion
        if let optionsV2 = options.profiling {
```

philipphofmann (commented):
Not much to add.

Samples/iOS-ObjectiveC/iOS-ObjectiveC/AppDelegate.m
armcknight: yep, forgot to flip it after changing the sample app's default setting. thanks!

PR #5063, opened 2025-04-08
fix(profiling): ui profiling trace lifecycle root span tracking

armcknight:
Fixes a couple edge cases where bookkeeping of root spans became imbalanced for trace lifecycle in UI profiling.

#skip-changelog

### DIFF
--- a/Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
+++ b/Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
@@ -249,11 +249,8 @@ sentry_stopProfilerDueToFinishedTransaction(
 #    endif // SENTRY_HAS_UIKIT
 )
 {
-    if (isProfiling && [hub.getClient.options isProfilingCorrelatedToTraces]) {
-        if (![SentryContinuousProfiler isCurrentlyProfiling]) {
-            SENTRY_TEST_FATAL(
-                @"Expected a continuous profiler to be running for this configuration.");
-        }
+    if (isProfiling && [hub.getClient.options isContinuousProfilingV2Enabled] &&
+        [hub.getClient.options isProfilingCorrelatedToTraces]) {
         SENTRY_LOG_DEBUG(@"Stopping tracking root span tracer with profilerReferenceId %@",
             transaction.trace.profilerReferenceID.sentryIdString);
         sentry_stopTrackingRootSpanForContinuousProfilerV2();
--- a/Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
+++ b/Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
@@ -330,19 +330,14 @@ SentryId *_Nullable sentry_startProfilerForTrace(SentryTracerConfiguration *conf
         // codepath
         return _sentry_startContinuousProfilerV2ForTrace(
             configuration.profileOptions, transactionContext);
-    } else if ([hub.getClient.options isContinuousProfilingEnabled]) {
+    } else if ([hub.getClient.options isContinuousProfilingV2Enabled]) {
         // non launch profile
-        SentryProfileOptions *profileOptions = hub.getClient.options.profiling;
-        if (profileOptions == nil) {
-            SENTRY_LOG_DEBUG(
-                @"Continuous profiling v1 configured; will not start automatically for trace.");
-            return nil;
-        }
         if (transactionContext.parentSpanId != nil) {
             SENTRY_LOG_DEBUG(@"Not a root span, will not start automatically for trace lifecycle.");
             return nil;
         }
-        return _sentry_startContinuousProfilerV2ForTrace(profileOptions, transactionContext);
+        return _sentry_startContinuousProfilerV2ForTrace(
+            hub.getClient.options.profiling, transactionContext);
     } else {
         BOOL profileShouldBeSampled
             = configuration.profilesSamplerDecision.decision == kSentrySampleDecisionYes;
--- a/Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
+++ b/Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
@@ -80,17 +80,13 @@ _unsafe_cleanUpTraceProfiler(SentryProfiler *profiler, NSString *tracerKey)
 void
 _unsafe_cleanUpContinuousProfilerV2()
 {
-    _gInFlightRootSpans -= 1;
-
-    if (_gInFlightRootSpans < 0) {
-#    if SENTRY_TEST || SENTRY_TEST_CI
-        SENTRY_CASSERT(NO, @"Attempted to decrement count of root spans to less than zero.");
-#    else
-        SENTRY_LOG_DEBUG(@"Attempted to decrement count of root spans to less than zero.");
-#    endif // SENTRY_TEST || SENTRY_TEST_CI
+    if (_gInFlightRootSpans == 0) {
+        SENTRY_TEST_FATAL(@"Attempted to decrement count of root spans to less than zero.");
+    } else {
+        _gInFlightRootSpans -= 1;
     }
 
-    if (_gInFlightRootSpans <= 0) {
+    if (_gInFlightRootSpans == 0) {
         SENTRY_LOG_DEBUG(@"Last root span ended, stopping profiler.");
         [SentryContinuousProfiler stop];
     } else {
@@ -133,10 +129,12 @@ SentryId *_Nullable _sentry_startContinuousProfilerV2ForTrace(
         return nil;
     }
 
+    SentryId *profilerReferenceId = [[SentryId alloc] init];
     SENTRY_LOG_DEBUG(
-        @"Starting continuous profiler for tracer %@", transactionContext.traceId.sentryIdString);
+        @"Starting continuous profiler for root span tracer with profilerReferenceId %@",
+        profilerReferenceId.sentryIdString);
     sentry_trackRootSpanForContinuousProfilerV2();
-    return [[SentryId alloc] init];
+    return profilerReferenceId;
 }
 
 } // namespace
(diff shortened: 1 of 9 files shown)


Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -80,17 +80,13 @@
...
-    if (_gInFlightRootSpans < 0) {
-#    if SENTRY_TEST || SENTRY_TEST_CI
-        SENTRY_CASSERT(NO, @"Attempted to decrement count of root spans to less than zero.");
-#    else
-        SENTRY_LOG_DEBUG(@"Attempted to decrement count of root spans to less than zero.");
-#    endif // SENTRY_TEST || SENTRY_TEST_CI
+    if (_gInFlightRootSpans == 0) {
+        SENTRY_TEST_FATAL(@"Attempted to decrement count of root spans to less than zero.");
+    } else {
+        _gInFlightRootSpans -= 1;
```
armcknight, line 86: This was a bad part, since _gInFlightRootSpans is an unsigned int, decrementing it from 0 would underflow it to UINT_MAX

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -170,21 +168,15 @@
...
-        if (!notLaunchContinuousProfileV2TraceLifecycle && !isProfiling) {
-            SENTRY_LOG_DEBUG(@"Continuous profiler v1 won't be stopped with a tracer.");
-            return;
-        }
-
+        SENTRY_LOG_DEBUG(@"Stopping tracking discarded tracer with profileReferenceId %@",
+            internalTraceId.sentryIdString);
         _unsafe_cleanUpContinuousProfilerV2();
-    } else if (internalTraceId != nil && isProfiling) {
+    } else if (internalTraceId != nil) {
```
armcknight, line 179: Two things were wrong here, one is that hub is nil for launch profiles, so the condition checked on line 181 was wrong. I also removed the `isProfiling` parameter and only call this function from SentryTracer if it is `true`.

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -257,16 +249,10 @@
...
-        }
-
-        if (![SentryContinuousProfiler isCurrentlyProfiling]) {
-            SENTRY_TEST_FATAL(
-                @"Expected a continuous profiler to be running for this configuration.");
-        }
+    if (isProfiling && [hub.getClient.options isContinuousProfilingV2Enabled] &&
+        [hub.getClient.options isProfilingCorrelatedToTraces]) {
+        SENTRY_LOG_DEBUG(@"Stopping tracking root span tracer with profilerReferenceId %@",
+            transaction.trace.profilerReferenceID.sentryIdString);
```
armcknight, line 255: There was no good reason to even do anything for invalid combinations, it would come down to user error I think, which we already warn on in logs. Simply make the check for the right configuration and don't mess with anything otherwise here.

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -351,6 +338,10 @@
                 @"Continuous profiling v1 configured; will not start automatically for trace.");
             return nil;
         }
+        if (transactionContext.parentSpanId != nil) {
+            SENTRY_LOG_DEBUG(@"Not a root span, will not start automatically for trace lifecycle.");
+            return nil;
+        }
```
armcknight, line 344: In some cases we were tracking child spans, checking for parents fixed issues related to that.

armcknight (commented):
A lot of the diff is just logging I used to find the problems, which I think are valuable to leave in. I've pointed out the actual fixes in comments.

Sources/Sentry/SentrySpan.m
```diff
@@ -110,6 +110,13 @@ - (instancetype)initWithContext:(SentrySpanContext *)context
         }
 #endif // SENTRY_TARGET_PROFILING_SUPPORTED
     }
+
+    if (_parentSpanId == context.parentSpanId) {
```
philprime, line 114: `m`: Xcode Analyzer seems to leave a warning here:

> Access to instance variable '_parentSpanId' results in a dereference of a null pointer (loaded from variable 'self') [core.NullDereference]

philprime (approved):
LGTM, left some notes.

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -80,17 +80,13 @@
...
-        SENTRY_LOG_DEBUG(@"Attempted to decrement count of root spans to less than zero.");
-#    endif // SENTRY_TEST || SENTRY_TEST_CI
+    if (_gInFlightRootSpans == 0) {
+        SENTRY_TEST_FATAL(@"Attempted to decrement count of root spans to less than zero.");
+    } else {
+        _gInFlightRootSpans -= 1;
     }
 
-    if (_gInFlightRootSpans <= 0) {
+    if (_gInFlightRootSpans == 0) {
```
philipphofmann, line 89: `m`: This looks weird now. We have the same if condition twice. Instead, we can merge the two into one.
```
if (_gInFlightRootSpans == 0) {
    SENTRY_TEST_FATAL(@"Attempted to decrement count of root spans to less than zero.");
} else {
    _gInFlightRootSpans -= 1;
}

if (_gInFlightRootSpans == 0) {
    SENTRY_LOG_DEBUG(@"Last root span ended, stopping profiler.");
    [SentryContinuousProfiler stop];
} else {
    SENTRY_LOG_DEBUG(@"Waiting for remaining root spans to finish before stopping profiler.");
}
```

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -80,17 +80,13 @@
...
-    _gInFlightRootSpans -= 1;
-
-    if (_gInFlightRootSpans < 0) {
-#    if SENTRY_TEST || SENTRY_TEST_CI
-        SENTRY_CASSERT(NO, @"Attempted to decrement count of root spans to less than zero.");
-#    else
-        SENTRY_LOG_DEBUG(@"Attempted to decrement count of root spans to less than zero.");
-#    endif // SENTRY_TEST || SENTRY_TEST_CI
+    if (_gInFlightRootSpans == 0) {
+        SENTRY_TEST_FATAL(@"Attempted to decrement count of root spans to less than zero.");
```
philipphofmann, line 84: `m`: I know it was already like that before, but now that I look at this again, I think it would be better to not call assert here. Incrementing and decrementing `_gInFlightRootSpans` gets called from many different places in the SentryTracer. One race condition could easily lead to `_gInFlightRootSpans` being 0 when calling `_unsafe_cleanUpContinuousProfilerV2`. Instead, I think we should simply log a warning and continue. I know the assert only fails when running tests, but even then I think exiting the program with an assert is an overkill. WDTY @philprime?

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
```diff
@@ -133,10 +129,12 @@
...
     }
 
+    SentryId *profilerReferenceId = [[SentryId alloc] init];
     SENTRY_LOG_DEBUG(
-        @"Starting continuous profiler for tracer %@", transactionContext.traceId.sentryIdString);
+        @"Starting continuous profiler for root span tracer with profilerReferenceId %@",
+        profilerReferenceId.sentryIdString);
     sentry_trackRootSpanForContinuousProfilerV2();
-    return [[SentryId alloc] init];
+    return profilerReferenceId;
```
philipphofmann, line 137: Thanks for the log message. This can be super helpful for debugging.

philipphofmann (commented):
This PR doesn't contain any changes in tests, which looks suspicious. The PR fixes some edge cases, so tests should have to change accordingly. Furthermore, it would be great to have some tests for these edge cases.

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
philprime: @philipphofmann How is this calling assert in production? DO you mean during tests?

```
#if SENTRY_TEST || SENTRY_TEST_CI
#    define SENTRY_TEST_FATAL(...) SENTRY_CASSERT(NO, __VA_ARGS__)
#else
#    define SENTRY_TEST_FATAL(...) SENTRY_LOG_WARN(__VA_ARGS__)
#endif // SENTRY_TEST || SENTRY_TEST_CI
```

Sources/Sentry/SentrySpan.m
armcknight: Oof, i went to fix the analyzer warning and changed the wrong thing. this should read `if (context.parentSpanId == nil)`, fixed

Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm
armcknight: Yes, but the condition means things at each spot. The first one checks before decrementing, if it's about to go below 0, which should never happen. The second one checks if it's 0 after decrementing. If we combine them it will have to have the second check in the else branch of the first, plus additional checks in the first to make sure the profiler is actually stopped if something has gone awry:

```
if (_gInFlightRootSpans == 0) {
    SENTRY_TEST_FATAL(@"Attempted to decrement count of root spans to less than zero, ensuring profiler is stopped.");
    if ([SentryContinuousProfiler isCurrentlyProfiling]) {
        [SentryContinuousProfiler stop];
    }
} else {
    _gInFlightRootSpans -= 1;
    if (_gInFlightRootSpans == 0) {
        SENTRY_LOG_DEBUG(@"Last root span ended, stopping profiler.");
        [SentryContinuousProfiler stop];
    } else {
        SENTRY_LOG_DEBUG(@"Waiting for remaining root spans to finish before stopping profiler.");
    }
}
```

PR #4760, opened 2025-01-28
fix: add NSNull handling to sentry_sanitize

philprime:
## :scroll: Description

- Adds handling of `[NSNull null]` as parameter for `sentry_sanitize`
- Adds unit tests for `sentry_sanitize`

## :bulb: Motivation and Context

Closes #4748 

## :green_heart: How did you test it?

- Additional unit tests

## :pencil: Checklist

You have to check all boxes before merging:

- [X] I added tests to verify the changes.
- [X] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [X] I updated the docs if needed.
- [X] I updated the wizard if needed.
- [X] Review from the native team if needed.
- [X] No breaking change or entry added to the changelog.
- [X] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

_none_

PR #5064, opened 2025-04-08
fix: Crash in setMeasurement when name is nil

philipphofmann:
## :scroll: Description

Although the name for setMeasurement is nonnull, we saw occurrences when it was nil, which led to crashes. Now, when the name is nil, the SDK discards the measurement and logs an error.

A customer reported these crashes.

Unit tests

## :pencil: Checklist


Sources/Sentry/SentryTracer.m
```diff
@@ -423,13 +423,27 @@ - (nullable SentryTraceContext *)traceContext
...
     SentryMeasurementValue *measurement = [[SentryMeasurementValue alloc] initWithValue:value
                                                                                    unit:unit];
+    [self setMeasurement:name measurement:measurement];
+}
+
+- (void)setMeasurement:(NSString *)name measurement:(SentryMeasurementValue *)measurement
+{
+    // Although name is nonnull we saw edge cases where it was nil and then leading to crashes. If
+    // the name is nil we can discard the measurement
+    if (name == nil) {
```
philprime, line 442: `l`: Does this also include `NULL`, `[NSNull null]`? Does it make sense to change the `setMeasurement` signature to use `_Nonnull` for the parameters?

Tests/SentryTests/Transaction/SentryTracer+Test.h
```diff
@@ -10,4 +10,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+void testing_setMeasurementWithNilName(SentryTracer *tracer, NSNumber *value);
```
philprime, line 13: `l`: Should we mark this explicitly as `_Nullable`?

philprime (approved):
LGTM, left some comments

Sources/Sentry/SentryTracer.m
philipphofmann: We already use _Nonnull, but somehow we still sometimes get `nil` here. It could be related to Kotlin Multiplatform.

Tests/SentryTests/Transaction/SentryTracer+Test.h
philipphofmann: The tracer and value must not be nil. The implementation passes `nil` to the setMeasurement. So not it doesn't make sense.

Sources/Sentry/SentryTracer.m
philipphofmann: Why do you think we should also include `NULL` and `NSNull null` here? We're not dealing with dictionaries here.

philprime: No direct reason, just double-checking after implementing #4760. If you say it is fine, I agree with you.

philipphofmann: I think it's fine.

philipphofmann:
Sorry about the commit messup, I accidentially merged https://github.com/getsentry/sentry-cocoa/pull/5067 into this one before merging this PR.

PR #5071, opened 2025-04-08
test(profiling): reproduce root span tracking issue

armcknight:
#skip-changelog

trying to repro a failure mode that was fixed in https://github.com/getsentry/sentry-cocoa/pull/5063 so we can have a unit test for it. it needs to integrate SentryTimeToDisplayTracker and SentryUIViewControllerSwizzling and related subsystems together with launch profiling.


Tests/SentryProfilerTests/SentryProfiledTracerConcurrencyTests.swift
```diff
@@ -0,0 +1,245 @@
...
+        _sentry_nondeduplicated_startLaunchProfile()
+
+        // Assert
+        XCTAssertEqual(_gInFlightRootSpans, 1)
+        XCTAssertTrue(SentryContinuousProfiler.isCurrentlyProfiling())
+
+        let uivc = ProfiledViewController(nibName: nil, bundle: nil)
+        uivc.loadView()
+        uivc.viewWillAppear(false)
+        fixture.displayLinkWrapper.call()
```
philipphofmann, line 133: `m`: As you pointed out on Slack, you want to have input on that test. I think we could make our lives easier here by limiting the setup and dependencies we require for testing. Ideally, yes we test with the started SDK, an UIViewController to start and finish a trace, but if this causes trouble, we can also emulate what that code is doing by directly interacting with the code we want to test. 
If the code we want to test requires all the setup, we could try to refactor the code so it is easier to test. I'm sorry it's a bit hard to explain in GH comment what exactly would be required here.

armcknight: Yes, I tried initially starting with just the components I had in mind, things kept failing in weird ways due to the tightly coupled nature of many parts of the SDK.

philipphofmann: Maybe we can refactor parts of the code we want to test so it depends on less code.

PR #5067, opened 2025-04-08
fix: Make setMeasurement thread safe

philipphofmann:
## :scroll: Description

Calling setMeasurement on transactions ans spans could lead to crashes. This is fixed now by making it thread safe.

## :bulb: Motivation and Context

A customer reported this crash and we see a few occurrences in our internal [SDK crash detection](https://sentry.sentry.io/issues/4684233446/?project=4505469596663808&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D%20setmeasurement&referrer=issue-stream&sort=user&stream_index=0).

## :green_heart: How did you test it?
Unit tests.

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #5078, opened 2025-04-10
fix: Make setMeasurement thread safe

philipphofmann:
Addition to #5067 to also make the getter of measurements thread safe.


philprime (approved):
LGTM

PR #5081, opened 2025-04-10
feat: Allow Hybrid SDKs to `setTrace`

bitsandfoxes:
## :scroll: Description

This PR allows Hybrid SDKs to explicitly set the `PropagationContext` on the scope via the `PrivateSentrySDKOnly`. 

The basic idea is that the Cocoa SDK, when brought in via a Hybrid SDK, should accept a trace from the Head SDK to connect errors on all layers of the application. This PR makes use of the fact that the Hybrid SDKs rely on their own performance instrumentation right now, allowing us to leverage the propagation context to achieve this.

Originates from the Unity SDK: https://github.com/getsentry/sentry-unity/issues/1998 
Unblocks the React Native SDK: https://github.com/getsentry/sentry-react-native/issues/3918
Mirror-PR on the Java SDK: https://github.com/getsentry/sentry-java/pull/4137

## :pencil: Checklist

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryScope.m
+++ b/Sources/Sentry/SentryScope.m
@@ -154,6 +154,17 @@ NS_ASSUME_NONNULL_BEGIN
     }
 }
 
+- (void)setPropagationContext:(SentryPropagationContext *)propagationContext
+{
+    @synchronized(_propagationContext) {
+        _propagationContext = propagationContext;
+
+        for (id<SentryScopeObserver> observer in self.observers) {
+            [observer setTraceContext:[self.propagationContext traceContextForEvent]];
+        }
+    }
+}
+
 - (nullable id<SentrySpan>)span
 {
     @synchronized(_spanLock) {
--- a/Sources/Sentry/SentryScope.m
+++ b/Sources/Sentry/SentryScope.m
@@ -159,8 +159,11 @@ NS_ASSUME_NONNULL_BEGIN
     @synchronized(_propagationContext) {
         _propagationContext = propagationContext;
 
-        for (id<SentryScopeObserver> observer in self.observers) {
-            [observer setTraceContext:[self.propagationContext traceContextForEvent]];
+        if (self.observers.count > 0) {
+            NSDictionary *traceContext = [self.propagationContext traceContextForEvent];
+            for (id<SentryScopeObserver> observer in self.observers) {
+                [observer setTraceContext:traceContext];
+            }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


Sources/Sentry/include/SentryScope+Private.h
```diff
@@ -20,7 +20,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (atomic, strong) SentryUser *_Nullable userObject;
 
-@property (atomic, strong) SentryPropagationContext *propagationContext;
+@property (nonatomic, strong) SentryPropagationContext *propagationContext;
```
philprime, line 23: `m`: Why did you change this? If there is a specific reason, it would be great if you could add comment so we know in the future

Sources/Sentry/include/HybridPublic/PrivateSentrySDKOnly.h
```diff
@@ -103,6 +104,11 @@ typedef void (^SentryOnAppStartMeasurementAvailable)(
  */
 + (NSDictionary *)getExtraContext;
 
+/**
+ * Allow Hybrids SDKs to set the current Trace.
+ */
++ (void)setTrace:(SentryId *)traceId spanId:(SentrySpanId *)spanId;
```
philprime, line 110: `l`: We might want to add a comment to the comments if this method is thread-safe

Sources/Sentry/SentryScope.m
```diff
@@ -154,6 +154,17 @@ - (void)setSpan:(nullable id<SentrySpan>)span
     }
 }
 
+- (void)setPropagationContext:(SentryPropagationContext *)propagationContext
+{
+    @synchronized(_propagationContext) {
+        _propagationContext = propagationContext;
+
+        for (id<SentryScopeObserver> observer in self.observers) {
+            [observer setTraceContext:[self.propagationContext traceContextForEvent]];
```
philprime, line 163: `m`: `traceContextForEvent` is creating a new dictionary on every call. Can we reuse the same dictionary when sending it to observers to reduce allocations?

philprime (commented):
Please add tests for the `SentryPropagationContext.initWithTraceId` to ensure that the `traceId` and `spanId` is actually set and not falling back to randomly generated ones (could be annoying to debug).

Sources/Sentry/SentryScope.m
bitsandfoxes: Nice catch, thanks!

PR #5076, opened 2025-04-09
test: more feature flags

armcknight:
Following on the work started in https://github.com/getsentry/sentry-cocoa/pull/5052, move more things governed by launch args and environment variables to SentrySDKOverrides so they can also be set in-app using UserDefaults.

Slightly reworks the API of the nested enums and accessor variables to be more streamlined and scalable, as well as the rules around how to combine schema settings and user default settings. See the develop-docs/README.md updates.

Also adds a screen that allows viewing the currently configured overrides, as well as setting them (which sets the overrides in user defaults): 
image

#skip-changelog

PR #5086, opened 2025-04-12
test: duplicate the schema config from iOS-Swift to the other swift apps

armcknight:
Following #5076/#5077/#5085

Duplicate the launch args and environment variables from the iOS-Swift Xcode schema to each other sample app's. Putting these in their own PR because it's a large diff. (This would be way less onerous if we used XcodeGen to generate sample app projects on the fly; they could be defined in exactly one place).

#skip-changelog


armcknight:
This will significantly change after moving to xcodegen, so will redo in a new PR when the time comes.

PR #5088, opened 2025-04-14
chore(file-io): add data and file manager swizzling to enabled features

### LINKED ISSUE #4906: Track adoption of enableFileManagerSwizzling
### Description

Track the adoption of the feature via the [SentryEnabledFeaturesBuilder](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Swift/Helper/SentryEnabledFeaturesBuilder.swift). When the option enableFileManagerSwizzling is enabled, please append it to the list, which gets added to the SDK event payload and gets indexed in Looker.

philprime:
## :scroll: Description

Adds two flags for data and file manager swizzling adoption tracking

Closes #4906

Added tests

## :pencil: Checklist

#skip-changelog

### DIFF
--- a/Sources/Swift/Helper/SentryEnabledFeaturesBuilder.swift
+++ b/Sources/Swift/Helper/SentryEnabledFeaturesBuilder.swift
@@ -57,7 +57,18 @@ import Foundation
             features.append("fastViewRendering")
         }
 #endif // #if os(iOS) && !SENTRY_NO_UIKIT
+
+        addFileIOFeatures(options: options, to: &features)
         return features
     }
     // swiftlint:enable cyclomatic_complexity
+
+    private static func addFileIOFeatures(options: Options, to features: inout [String]) {
+        if options.experimental.enableDataSwizzling {
+            features.append("dataSwizzling")
+        }
+        if options.experimental.enableFileManagerSwizzling {
+            features.append("fileManagerSwizzling")
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


Sources/Swift/Helper/SentryEnabledFeaturesBuilder.swift
```diff
@@ -57,7 +57,18 @@ import Foundation
             features.append("fastViewRendering")
         }
 #endif // #if os(iOS) && !SENTRY_NO_UIKIT
+
+        addFileIOFeatures(options: options, to: &features)
```
philipphofmann, line 61: `l`: IMO, the extra method harms readability a bit. Now I have to jump around to see which features are added to the list. For me, it's OK to have all features in one method. No need to add an extra one.

philipphofmann (approved):
LGTM, thanks.

Sources/Swift/Helper/SentryEnabledFeaturesBuilder.swift
philprime: I also don't like this but swiftlint complained about function body length and cyclomatic complexity is already disabled too, so maybe it's worth starting to extract into multiple methods.

philprime: As we both don't like this I'll change it and force-ignore the linter

PR #5092, opened 2025-04-15
test: Return create-simulator to compare flakiness

krystofwoldrich:
## :scroll: Description

Returns sim creation removed in https://github.com/getsentry/sentry-cocoa/pull/5015 and https://github.com/getsentry/sentry-cocoa/pull/5030

Based on CI stats, the Xcode 14.3.1 test became very flaky; with this change, we can compare if creating the sim improves it.

Previous:
Screenshot 2025-04-15 at 13 39 59

Current:
Screenshot 2025-04-15 at 13 39 37

#skip-changelog 

Superseded by: https://github.com/getsentry/sentry-cocoa/pull/5093

### DIFF
--- a/.github/workflows/ui-tests-critical.yml
+++ b/.github/workflows/ui-tests-critical.yml
@@ -86,8 +86,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+
+      - name: Boot Default Simulator
+        if: ${{ matrix.device == 'default' && matrix.os-version == 'default' }}
+        run: ./scripts/ci-boot-simulator.sh --xcode ${{matrix.xcode}}
+
       - name: Create ${{matrix.device}} (${{matrix.os-version}}) Simulator using Xcode ${{matrix.xcode}}
-        if: ${{ matrix.os-version != 'default' }} # Only create simulator if os-version is set
+        if: ${{ matrix.device != 'default' && matrix.os-version != 'default' }} # Only create simulator if os-version is set
         run: ./scripts/create-simulator.sh "${{matrix.xcode}}" "${{matrix.os-version}}" "${{matrix.device}}" "${{matrix.force-sim-runtime}}"
 
       - name: Install tooling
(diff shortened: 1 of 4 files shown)


.github/workflows/ui-tests-critical.yml
```diff
@@ -83,17 +113,21 @@ jobs:
         run: |
           maestro test ${{env.MAESTRO_FLOWS_PATH}} --format junit --debug-output ${{env.MAESTRO_LOGS_PATH}}
 
+      - name: Delete Simulator
+        if: always()
+        run: ./scripts/delete-simulator.sh
```
philprime, line 115: `l`: why do we have to delete the simulator? Wouldn't it get discarded anyways by the CI runner?

scripts/ci-boot-simulator.sh
```diff
@@ -14,37 +14,51 @@ set -euo pipefail
 # Parse named arguments
 XCODE_VERSION="16.2" # Default value
 
+# Only allow one argument type
+if [[ $# -gt 2 ]]; then
+    echo "Error: Cannot specify both --xcode and --device"
+    echo "Usage: $0 [-x|--xcode <version>] OR [-d|--device <device>]"
+    exit 1
+fi
+
```
philprime: `m`: This needs to be remove the next time another param is added. Please instead do a check after the parsing of the the options to check it's XOR, and throw the error if both variables are set.

philprime (approved):
I added comments.

We might want to consider the setup used by fastlane to create simulators for automated screenshots:

- https://github.com/fastlane/fastlane/blob/master/snapshot/lib/snapshot/simulator_launchers/simulator_launcher_base.rb#L51
- https://github.com/fastlane/fastlane/blob/95cab31c184e3b0fd0ccebce9dd8ade5f224db1d/fastlane_core/lib/fastlane_core/device_manager.rb#L224

scripts/ci-boot-simulator.sh
```diff
@@ -48,3 +48,9 @@ esac
 
 echo "Booting simulator $SIMULATOR"
 xcrun simctl boot "$SIMULATOR"
+
+# Wait for the simulator to boot
+xcrun simctl bootstatus "$SIMULATOR"
```
krystofwoldrich, line 53: Maybe the issue was in waiting for the simulator boot, which was removed when removing the sim creation; only time will tell.

.github/workflows/ui-tests-critical.yml
krystofwoldrich: Yes, in theory GH should remove it.

krystofwoldrich:
One more observation,

for the latest macos-15 the failure rate basically hasn't changed, but it also boots the sim much faster, based on my observation.

Let's first see if waiting for the boot improves the flakiness.
- https://github.com/getsentry/sentry-cocoa/pull/5093

PR #5094, opened 2025-04-15
test:  Async safe log write to console

philipphofmann:
Ensure setting SENTRY_ASYNC_SAFE_LOG_ALSO_WRITE_TO_CONSOLE = 1 doesn't get committed by accident.

#skip-changelog


philprime:
`m`: While I do like this approach in using tests to cause the failure of CI, where else is this `SENTRY_ASYNC_SAFE_LOG_ALSO_WRITE_TO_CONSOLE` used and where else can it become a problem?
What if it is not set on `SentryTests` but set on `Sentry`?

I believe I need more context here.

philipphofmann:
Thanks for the feedback, @philprime . Is it clear now https://github.com/getsentry/sentry-cocoa/pull/5094/commits/b6e222bd10f81c032c58c0eb5e9bb70395b8d1f8?

PR #5105, opened 2025-04-17
ci: build sample apps on latest macOS image and Xcode version

armcknight:
- use `macos-15` and Xcode 16.3 for sample app build jobs (they were a mixture of older ones, but this is just to make sure they compile at all, so we should be able to use the same thing for them all, and that should be what we're all presumably using on our local machines as well)
- remove the default for the xcode version parameter in ci-select-xcode.sh: it's better to see exactly what version is being used in the workflow config files. if we were to change the default in the bash script, that change is invisible in the workflow configs. also, we didn't have any naked invocations anyways

`make session-replay-camera-test-xcode` will generate the project, although in the future we might want a small script to do this that can take some parameters like platform/version/language/UI-lib

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,10 +29,10 @@ jobs:
   # With this we catch potential issues already in the PR.
   ios-swift-release:
     name: Release Build of iOS Swift
-    runs-on: macos-14
+    runs-on: macos15
     steps:
       - uses: actions/checkout@v4
-      - run: ./scripts/ci-select-xcode.sh 16.2
+      - run: ./scripts/ci-select-xcode.sh 16.3
       - uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -29,10 +29,10 @@ jobs:
   # With this we catch potential issues already in the PR.
   ios-swift-release:
     name: Release Build of iOS Swift
-    runs-on: macos-14
+    runs-on: macos15
```
philipphofmann, line 32: `m`: I think this should be
```suggestion
    runs-on: macos-15
```

philipphofmann (approved):
Thanks for the cleanup

.github/workflows/build.yml
armcknight: Thanks!

PR #5106, opened 2025-04-17
ci: camera sample with xcodegen

armcknight:
Demonstrate using XcodeGen to generate one of our sample app projects.

As the first in a series of such changes, it also introduces a collection of xcconfig files for us to use as a shared set for all the sample apps.


.gitignore
```diff
@@ -224,3 +224,4 @@ dist
 .yarn/build-state.yml
 .yarn/install-state.gz
 .pnp.*
+Samples/SessionReplay-CameraTest/SessionReplay-CameraTest.xcodeproj
```
philprime, line 227: `h`: In previous years I've also been ignoring xcodegen-generated Xcode projects therefore I understand why you added this. Over the time it got harder to checkout the project and get it up and running, because it always caused friction with generating the project on every clone and branch checkout.
Therefore I would prefer if we could check in the generated project as well.

.github/workflows/build.yml
```diff
@@ -66,17 +66,31 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: ./scripts/ci-select-xcode.sh 16.3
+      - run: brew install xcodegen
+        if: ${{ contains(matrix.scheme, 'SessionReplay-CameraTest') }}
+      - run: make session-replay-camera-test-xcode
+        if: ${{ contains(matrix.scheme, 'SessionReplay-CameraTest') }}
```
philprime, line 72: `m`: It might make sense to add a flag `generate` to the matrix and set it to `true` only for the `SessionReplay-CameraTest`, instead of doing a name-based check to keep the complexity lower

.github/workflows/build.yml
```diff
@@ -66,17 +66,31 @@ jobs:
...
           build
+      - if: ${{ contains(matrix.scheme, 'SessionReplay-CameraTest') }}
+        run: >-
+          env NSUnbufferedIO=YES
+          xcodebuild
+          -project Samples/${{matrix.scheme}}/${{matrix.scheme}}.xcodeproj
+          -scheme '${{matrix.scheme}}'
+          -configuration Debug
+          CODE_SIGNING_ALLOWED="NO"
+          build
```
philprime, line 93: `l`: Does it make sense to migrate this to `sentry-xcodebuild.sh` in this PR?

Samples/SessionReplay-CameraTest/Config/SessionReplay-CameraTest.xcconfig
```diff
@@ -0,0 +1,27 @@
+#include "Architectures.xcconfig"
```
philprime, line 1: `m`: Why do we need to create so many different config files instead of having it in one?

Samples/SessionReplay-CameraTest/project.yml
```diff
@@ -0,0 +1,28 @@
+name: SessionReplay-CameraTest
```
philprime, line 1: `m`: Couple of options we might want to add:

- [`options.bundleIdPrefix`](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#options) - locking down the prefix use by all the targets so it's not autogenerated
- [`options.generateEmptyDirectories`](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#options) - Can be helpful when creating empty directories in Finder, but want to create files afterwards in Xcode. By Default empty directories are not showing up in Xcode.
- [`options.createIntermediateGroups`](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#options) - We should test this option to see the difference in group strucutre.
- [`options.groupOrdering`](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#groupOrdering) - This one is rather subjective but could be useful to improve our structure.
- [`schemes`](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#scheme) - To explicitly define the generated schemes so we are all using the same.

Sentry.xcworkspace/contents.xcworkspacedata
```diff
@@ -34,9 +34,6 @@
       <FileRef
          location = "group:Samples/iOS-Swift/iOS-Swift.xcodeproj">
       </FileRef>
-      <FileRef
```
philprime, line 37: `h`: Why did you remove this project?

armcknight: hmm, i removed them from the workspace, but i suppose we could leave the references in place.

Samples/SessionReplay-CameraTest/project.yml
armcknight: i actually did use `schemes` in a subsequent PR to coordinate test targets for iOS-Swift, since things are a little more complicated there. i think schemes are autogenerated otherwise. after making the change mentioned in https://github.com/getsentry/sentry-cocoa/pull/5106#discussion_r2055063454 i'll see if CI works, I'd say let's only explicitly define the schemes if it doesn't autocreate or work otherwise.

i wouldn't care about groupOrdering since it introduces the need to manually maintain two sets of strings, and I personally don't use the file navigator, i just use quick open (⌘ ⇧ O) and then if I need to e.g. create a new sibling file, jump to the right spot in the file navigator with ⌘ ⇧ J, but i could see the value in createIntermediateGroups/generateEmptyDirectories for the latter kind of purpose.

Samples/SessionReplay-CameraTest/Config/SessionReplay-CameraTest.xcconfig
armcknight: one thing this helps with is scoping the git history of changes for a logical grouping of settings (as they are grouped in Xcode). and i suppose, just keeping the requirements on one's working memory manageable.

everything else can be done either way. for instance, one danger is adding the same setting multiple times with different values and not realizing it. this is easy to do in a single long list, and in different files. but a lint script could be written to check either kind of structure. so i don't see this as a motivating factor one way or the other.

.github/workflows/build.yml
armcknight: no, i'd prefer not to touch that script here, and based on https://github.com/getsentry/sentry-cocoa/pull/5106#discussion_r2055063454, might even change this back.

.github/workflows/build.yml
armcknight: i thought about doing this, but this is just a temporary measure while we have a mix of xcodegen-managed projects and otherwise. once they're all across the finish line, this will disappear anyways.

.gitignore
armcknight: I prefer not to, because then you can easily get mismatched state if someone checks in one thing but not the other, or if someone directly modifies the xcodeproj files. 

In either case, you have to know whether or not you need to regenerate a given project locally. `xcodegen` is basically instantaneous, and if something in a spec gets broken, then _hopefully_ that wouldn't get checked into `main` as CI should fail. 

Are there other kinds of friction you've experienced in the past?

philprime: > I prefer not to, because then you can easily get mismatched state if someone checks in one thing but not the other, or if someone directly modifies the xcodeproj files.

That is in my opinion exactly the reason why we should check them in, so that we are all using the same project and have less "works on my machine" situations. We should just not use any machine-specific paths, i.e. absolute paths.

> In either case, you have to know whether or not you need to regenerate a given project locally.

How would you do this?

> Are there other kinds of friction you've experienced in the past?

The largest pain-point was that I never know if the currently generated Xcode project is the latest one, so I constantly had to re-generate the project, especially on every branch change → my main issue with not checking-in the Xcode project.

I also have the case that re-generating the project while Xcode is open is constantly crashing Xcode in my personal projects. I have to restart Xcode probably at least 5 times per day.

I also did participate in [this issue](https://github.com/yonaskolb/XcodeGen/issues/1184).

Samples/SessionReplay-CameraTest/Config/SessionReplay-CameraTest.xcconfig
philprime: Okay while it is still a lot of files to look into I see your reasoning.

.gitignore
armcknight: > less "works on my machine" situations

This is what CI is for.

> The largest pain-point was that I never know if the currently generated Xcode project is the latest one, so I constantly had to re-generate the project

Regenerate it if you pull the main branch or merge in from upstream. In practice you wouldn't even need to do it every time though. Regarding this:

> I also have the case that re-generating the project while Xcode is open is constantly crashing Xcode in my personal projects

I have been iterating on the yml specs and regenerating projects while they're open without crashing. Just have to hit spacebar to allow Xcode to reload from disk with that dialog it pops up.

> I also did participate in https://github.com/yonaskolb/XcodeGen/issues/1184.

I haven't observed this issue, but probably because I only use the xcconfig files to set build settings, with nothing coming from the yml spec.

Sentry.xcworkspace/contents.xcworkspacedata
philprime: When I started to maintain this SDK I assumed it's best to use the workspace to have access to all relevant files. That's my main reason, but I assume the CLI usage would also have to be changed.

> i'll make this change 👍🏻

Thank you

.gitignore
philprime: > Regenerate it if you pull the main branch or merge in from upstream. In practice you wouldn't even need to do it every time though.

I'm not sure how to express this better, in the last 3 years of working with xcodegen, I had to re-generate my projects multiple times per day. Especially when changing branches, because there are most certainly new files in a different branch which would then not show up after checkout.

> I have been iterating on the yml specs and regenerating projects while they're open without crashing.

That is also what I am expecting, I am just telling you from my personal experience. If I could reliably reproduce, I would have even tried to fix it. Sometimes it crashed every time I regenerate, sometimes it does not.

**To summarize my point:**
We had the Xcode project checked in before and we could be sure that if we check out a branch, the exact project would be there.
Yes, Xcode projects cause merge conflicts, but that can easily be fixed by simply re-generate in case of a conflict, then merge the fixed project.
For reference, if you have a `yarn.lock` file with merge conflicts, running `yarn install` will automatically fix the lockfile, exact same process.

Sentry.xcworkspace/contents.xcworkspacedata
armcknight: The CLI flow could even be made simpler. Instead of a separate make task per project as I am currently proceeding, I can just make it regenerate all the projects in one go.

Worth noting though, that if we ditch the workspace, we can still generate projects that have access to all the relevant files.

Samples/SessionReplay-CameraTest/project.yml
armcknight: > this could lead to use using different schemes working on this SDK

Could you explain what you mean here? If we don't define a scheme explicitly, and XcodeGen automatically makes its own, where does the difference come from?

.gitignore
armcknight: > I'm not sure how to express this better

You've expressed your position adequately. I simply think your tradeoffs are inferior to mine, based on my own experience. You are not the only one with years of experience with XcodeGen, I've been using it for 6-7 years now, for everything from personal projects to commercial projects based in Bazel that would generate XcodeGen specs from Jinja templates producing Xcode projects scoped down to the minimal dependency tree needed to build a module. And sorry, I've just not had the bad experience you describe. What we have right now in this repository is painfully antiquated to me, after having done things in better ways for so long.

Xcode crashing because the underlying files changed has nothing to do with XcodeGen. When you change branches, the files also change underneath Xcode. This has historically also caused crashes when changing branches with a project open–I know, I've experienced that plenty as well. You are simply advocating to have git be the one changing the files instead of XcodeGen. This doesn't solve the problem in any way whatsoever.

Tracking the project along with the spec is simply adding a second source of truth that must be maintained in parallel. We'd have to vigilantly enforce conventions or write more CI checks. It may produce huge diffs every time someone decides to regenerate the projects using a different Xcode version than what was used to generate the last one checked in.

You're in terminal, you change branches: run make xcode again. We can add it to .git/post-checkout if you don't use terminal. This sounds like a non-issue to me.

Sentry.xcworkspace/contents.xcworkspacedata
armcknight: This is now done, for this and all subsequent PRs, there is now just the one `make xcode` task, and the projects stay linked in the xcworkspace so they can be used there.

Samples/SessionReplay-CameraTest/project.yml
armcknight: FWIW, I do explicitly define schemes for apps that have associated test targets, to tie the test targets to the apps' schemes' test actions, in [iOS-Swift](#5115), [iOS-Swift6](#5116) and (iOS13-Swift)[#5117]. This project for the camera test I don't believe needs it as it currently doesn't have any UI tests.

armcknight: Reading the docs for `bundleIdPrefix` again, I'm not sure how it accomplishes what you mention:

> locking down the prefix use by all the targets so it's *not* autogenerated

emphasis mine.

from the docs:

> If this is specified then any target that doesn't have an PRODUCT_BUNDLE_IDENTIFIER (via all levels of build settings) will get an autogenerated one by combining bundleIdPrefix and the target name: bundleIdPrefix.name. The target name will be stripped of all characters that aren't alphanumerics, hyphens, or periods. Underscores will be replaced with hyphens.

So it sounds like it _does_ autogenerate? I don't think we want this behavior so I'm leaving it out for now.

test.txt
```diff
@@ -0,0 +1 @@
+test
```
philprime, line 1: `m`: clean this up please

.gitignore
philprime: Ok, while we still not agree with you or your arguments, if this becomes a continuous problem again like I experienced before, I'll raise an internal discussion and we might have to revisit this. Until then I'll accept this.

philprime (approved):

philipphofmann:
@armcknight, what's stopping us from merging this PR?

Issue #4767, opened 2025-01-29
feedback: implement bring-your-own-button


armcknight:
### Description

we currently provide a customizable button widget, but there is also a spec for customers to provide a button they create and insert in the view hierarchy, that we essentially attach a handler to

PR #5107, opened 2025-04-21
feat(feedback): bring your own button

armcknight:
For #4767, implements the configuration option to assign the hook to open the feedback form from a button a customer creates and places in their own app how/where they like best.

Required some refactoring of our own classes for the overlay window and root view controller to abstract out the behavior from the display, so we can reroute the behavioral codepaths to this new config point.


Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
```diff
@@ -131,7 +131,7 @@ extension UserFeedbackUITests {
 
         fillInFields(testMessage, testName, testEmail)
         
-        submit()
+        sendButton.tap()
```
philprime, line 134: `h`: In the implementation of `submit()` you are asserting for `            XCTAssert(widgetButton.waitForExistence(timeout: 1))` if the parameter `expectingError` is set to `false` (it is by default). That means, replacing this call removes the assertion. Please double-check that you actually intended to remove the assertion for the widget button when not using a custom button.

Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
```diff
@@ -169,7 +169,7 @@ extension UserFeedbackUITests {
         let testMessage = "UITest user feedback"
         fillInFields(testMessage)
         
-        submit()
+        sendButton.tap()
```
philprime, line 172: `h`: See comment above

Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
```diff
@@ -229,7 +259,7 @@ extension UserFeedbackUITests {
         messageTextView.tap()
         messageTextView.typeText("UITest user feedback")
         
-        submit()
+        sendButton.tap()
```
philprime, line 262: `h`: See comment above

Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
```diff
@@ -318,7 +348,7 @@ extension UserFeedbackUITests {
         let testMessage = "UITest user feedback"
         fillInFields(testMessage)
         
-        submit()
+        sendButton.tap()
```
philprime, line 351: `h`: See comment above

Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
```diff
@@ -332,7 +362,7 @@ extension UserFeedbackUITests {
...
```
philprime, line 365: `h`: See comment above

Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
```diff
@@ -454,11 +484,33 @@ extension UserFeedbackUITests {
         messageTextView.tap()
         messageTextView.typeText("UITest user feedback")
         
-        submit()
+        sendButton.tap()
```
philprime, line 487: `h`: See comment above

Samples/iOS-Swift/iOS-Swift/SentrySDKWrapper.swift
```diff
@@ -239,6 +248,10 @@ extension SentrySDKWrapper {
         config.configureForm = configureFeedbackForm(config:)
         config.configureTheme = configureFeedbackTheme(config:)
         configureHooks(config: config)
+
+        if ProcessInfo.processInfo.arguments.contains("--io.sentry.ui-test.use-custom-feedback-button") {
+            config.customButton = feedbackButton
+        }
```
philprime, line 254: `m`: Would it make sense to move this to the `SentrySDKOverrides.swift`?

Samples/iOS-Swift/iOS-Swift/ErrorsViewController.swift
```diff
@@ -13,6 +13,13 @@ class ErrorsViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         addDSNDisplay(self, vcview: dsnView)
+
+        if ProcessInfo.processInfo.arguments.contains("--io.sentry.ui-test.use-custom-feedback-button") {
```
philprime, line 17: `m`: Would it make sense moving this live to `SentrySDKOverrides.swift`?

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
```diff
@@ -37,7 +37,14 @@ public class SentryUserFeedbackConfiguration: NSObject {
...
     public var showFormForScreenshots: Bool = false
-    
+
+    /**
+     * Install a hook for the specified button to show the form when it is pressed.
+     * - note: If this is set, `configureWidget` is ignored.
+     * - note: Default: `nil`
+     */
+    public var customButton: UIButton?
+
```
philprime, line 47: `l`: Using a button seems to conver most use-cases. To widen the possibilities without impacting the SDK we could also go for [`UIControl`](https://developer.apple.com/documentation/uikit/uicontrol) instead, allowing [other components](https://developer.apple.com/documentation/uikit/uicontrol#inherited-by) such as [UISwitch](https://developer.apple.com/documentation/uikit/uiswitch) or custom controls to be used as well.

This might be an edge case for most users, but as we are not using any `UIButton`-specific features, I don't see a reason to be restrictive either.

```suggestion
    public var customButton: UIControl?
```

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
```diff
@@ -151,6 +58,57 @@ struct SentryUserFeedbackWidget {
...
+                constraints.append(button.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor, constant: -config.widgetConfig.layoutUIOffset.horizontal))
+            }
+            if config.widgetConfig.location.contains(.leading) {
+                constraints.append(button.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor, constant: config.widgetConfig.layoutUIOffset.horizontal))
+            }
+            NSLayoutConstraint.activate(constraints)
+        }
+
+        required init?(coder: NSCoder) {
+            fatalError("init(coder:) has not been implemented")
```
philprime, line 92: `l`: We should add a comment or change the error message to explain why it wasn't implemented so it does not look like an undocumented TODO.

For example:
```suggestion
            fatalError("init(coder:) is not supported for the Sentry user feedback widget window")
```

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
```diff
@@ -151,6 +58,57 @@ struct SentryUserFeedbackWidget {
             return result
```
philprime, line 58: `m`: I have implemented a `PassThroughWindow` in the past (I'll share it with you internally in Slack) and encountered an edge case where events did not correctly pass through between iOS 17 and iOS 18. 

Here's a Apple Developer Forum thread I would like you to read through and double-check it does/does not apply to this hit test logic.

https://developer.apple.com/forums/thread/762292

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
```diff
@@ -151,6 +58,57 @@ struct SentryUserFeedbackWidget {
...
         }
     }
+
+    func setWidget(visible: Bool, animated: Bool) {
+        rootVC.setWidget(visible: visible, animated: animated)
+    }
+
+    class RootViewController: UIViewController, UIAdaptivePresentationControllerDelegate {
+        let defaultWidgetSpacing: CGFloat = 8
+        unowned let button: SentryUserFeedbackWidgetButtonView
```
philprime, line 68: `m`: Why `unowned` and not weak? While we are in control in the reference, it is easy to break unowned references, which can result in SDK crashes. Using `weak` with nil-unwrapping would be safer to use.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
```diff
@@ -39,59 +35,131 @@ class SentryUserFeedbackIntegrationDriver: NSObject, SentryUserFeedbackWidgetDel
         if let darkThemeOverrideBuilder = configuration.configureDarkTheme {
             darkThemeOverrideBuilder(configuration.darkTheme)
         }
-        
-        if configuration.widgetConfig.autoInject {
-            createWidget()
+
+        if let customButton = configuration.customButton {
+            customButton.addTarget(self, action: #selector(showForm(sender:)), for: .touchUpInside)
```
philprime, line 40: `h`: We should wrap `self` in a weak box, because of this comment in the documentation of `addTarget`:

> The control does not retain the object in the target parameter. It is your responsibility to maintain a strong reference to the target object while it is attached to a control.

If the SentrySDK is closed and uninstalled, I would expect it to dealloc the user feedback integration driver, therefore this can cause `self` to be deallocated, leaving a dangling target action.

Alternative could be unregistering from the `customButton` on `dealloc`.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
```diff
@@ -39,59 +35,131 @@ class SentryUserFeedbackIntegrationDriver: NSObject, SentryUserFeedbackWidgetDel
...
-            || config.location.contains(.top)
-            || config.location.contains(.bottom)
+        || config.location.contains(.leading)
+        || config.location.contains(.top)
+        || config.location.contains(.bottom)
         let notAll = !config.location.contains(.all)
         let valid = noOpposingVerticals && noOpposingHorizontals && atLeastOneLocation && notAll
-        #if DEBUG
+#if DEBUG
         assert(valid, "Invalid widget location specified: \(config.location). Must specify either one edge or one corner of the screen rect to place the widget.")
```
philprime, line 114: `m`: Why is this wrapped in a `DEBUG` check?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
```diff
@@ -39,59 +35,131 @@ class SentryUserFeedbackIntegrationDriver: NSObject, SentryUserFeedbackWidgetDel
...
+            guard let resolvedResponder = responder else { break }
+            let klass = type(of: resolvedResponder)
+            guard klass.isSubclass(of: UIViewController.self) else {
+                responder = resolvedResponder.next
+                continue
+            }
+            return resolvedResponder as? UIViewController
+        }
+        return nil
     }
```
philprime, line 163: `m`: Please add somme comments why the next responder lookup is implemented like this. Is there no internal implementation to find the next responder? Why is the method called `controller` which I would expect to be the controller managing this view? Some additional context could be helpful down the line to maintain this SDK.

philprime (commented):
I added a couple of comments I need you to look at.

In general I would highly appreciate you adding more documentation/comments in the code, not to explain what the code does, but instead the context of it. This would help us and future maintainers tremendously when coming back to the code in a couple of months.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
armcknight: to try to avoid crashing in production. we have various macros like this in objc code, but it's clumsier in swift

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
armcknight: I'll use the weak ref

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
armcknight: > Please add somme comments why the next responder lookup is implemented like this. 

I'll add it as a comment, but to explain here, it's because we need the UIViewController instance that owns the view in question, which will either be a custom button or our own widget button, so that we can call `presentViewController` to display the form. 

> Is there no internal implementation to find the next responder? 

Not that I know of, or I would've used it, but someone can feel free to point one out if there is one.

> Why is the method called controller which I would expect to be the controller managing this view? 

It is returning the deepest nested UIViewController, see my comment above. I will rename it to `viewController`.

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
armcknight: This is out of scope for this PR, please file an issue and/or discuss this idea in the project's slack room.

Samples/iOS-Swift/iOS-Swift/ErrorsViewController.swift
armcknight: Yep!

Samples/iOS-Swift/iOS-Swift-UITests/UserFeedbackUITests.swift
armcknight: Ah, you're right, this is probably why i saw the timing issue come back that i fixed in one of the new test cases.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackWidget.swift
armcknight: I did take a look through that thread, I hadn't seen it before, but this is out of scope for this PR. The code here is not new, it's already shipped, it only showed up in the diff because due to the refactor it was outdented one scope level.

Worringly, we only run the ui tests on iOS 16.4 and 17.5 currently, but I actually just opened a separate PR to update the 17.5 run to 18.3:

armcknight: See https://github.com/getsentry/sentry-cocoa/pull/5125

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
philprime: Would not have been any changes other than changing the type `UIButton` to `UIControl`. I'll raise it in Slack to spread awareness, rather than requesting it.

philprime (approved):
Thank you for integrating all the feedback, LGTM

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
armcknight: We only attach a touch-up-inside action trigger to the UI element. if we instead use a UIControl, then we have to consider other actions like value changed for UISwitch, so a straight swap is not possible without surprising side effects for end users. This would expand the design scope of the feature as we have to consider what it really means for people to use a switch versus a button. They are different entities in the UI framework because they are different UX. So I disagree that they can be treated interchangeably. If you feel strongly that we should support something like a switch (which really makes zero sense to me here) then please raise the issue with our design team and PMs, not on a PR like this.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Features
+
+- Added ability to bring your own button for user feedback form display (#5107)
```
```suggestion
- Added ability to bring your own button for user feedback form display (#5107)

```

PR #5116, opened 2025-04-22
ci: ios swift6 xcodegen

armcknight:
Following #5115, extracting more projects out of the old iOS-Swift xcode project (which really housed multiple separate apps–this separates each one's project spec) to be generated by xcodegen


Plans/iOS-Swift6_Base.xctestplan
```diff
@@ -14,9 +14,9 @@
   "testTargets" : [
     {
       "target" : {
-        "containerPath" : "container:iOS-Swift.xcodeproj",
-        "identifier" : "7B64386726A6C544000D0F65",
-        "name" : "iOS-Swift-UITests"
```
armcknight, line 19: I'm not sure how this even worked before. It seems to have been targeting the wrong test target 🤔

Samples/iOS-Swift6.xcconfig
```diff
@@ -0,0 +1,27 @@
...
+#include "Shared/Config/ClangWarningsObjC.xcconfig"
+#include "Shared/Config/AssetCatalog.xcconfig"
+#include "Shared/Config/ClangAnalyzer.xcconfig"
+#include "Shared/Config/Swift.xcconfig"
+#include "Shared/Config/Metal.xcconfig"
+
+PRODUCT_BUNDLE_IDENTIFIER = io.sentry.sample.iOS-Swift6
+INFOPLIST_FILE = iOS-Swift6/Info.plist
+SUPPORTED_PLATFORMS = iphoneos iphonesimulator
+SWIFT_VERSION = 6
```
philprime, line 27: `m`: Can we somehow make `iOS-Swift6` use the exact same configuration as `iOS-Swift` and only set the `SWIFT_VERSION` to `6`? I think this sample is used not as a separate platform like macOS/tvOS etc.

Samples/iOS-Swift6/Info.plist
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
philprime, line 1: `m`: Just so we are aligned, if we are not including the Xcode project in the repository anymore, because it can be generated, shouldn't we also remove Info.plist because it is also generated?

Samples/iOS-Swift6.xcconfig
armcknight: we can, i considered rolling up the xcconfigs to a common config that's included in each target config and then overriding only what's necessary, but wanted to do it in a later refactor so that change is isolated and to keep these PRs as compartmentalized as possible

Samples/iOS-Swift6/Info.plist
armcknight: for most of these, i turned off plist generation. i don't feel strongly one way or the other, as we can have a shared xcconfig with plist settings, or a shared plist file, in the future. let's make the decision in a later refactor PR.

Samples/iOS-Swift6.xcconfig
armcknight: noted this in https://github.com/getsentry/sentry-cocoa/issues/5165#issuecomment-2852482982

Sentry.xcodeproj/project.pbxproj
```diff
@@ -2010,7 +2010,6 @@
 		D48E8B9C2D3E82AC0032E35E /* SentrySpanOperation.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SentrySpanOperation.swift; sourceTree = "<group>"; };
 		D4A236012D5F838200D55C58 /* macOS-Swift_Base.xctestplan */ = {isa = PBXFileReference; lastKnownFileType = text; path = "macOS-Swift_Base.xctestplan"; sourceTree = "<group>"; };
 		D4A236062D5F846200D55C58 /* iOS-ObjectiveC_Base.xctestplan */ = {isa = PBXFileReference; lastKnownFileType = text; path = "iOS-ObjectiveC_Base.xctestplan"; sourceTree = "<group>"; };
-		D4A236072D5F846F00D55C58 /* iOS-Swift6_Base.xctestplan */ = {isa = PBXFileReference; lastKnownFileType = text; path = "iOS-Swift6_Base.xctestplan"; sourceTree = "<group>"; };
```
philprime, line 2013: `m`: This might be an unexpected removal, because the test plan still exists. Can be wrong here, leaving this at your discretion.

philprime (approved):

PR #5117, opened 2025-04-22
ci: move iOS13-Swift under xcodegen mgmt

armcknight:
following #5116, moving app projects to xcodegen. this is the last app that was under iOS-Swift.xcodeproj, so that xcodeproj file is now able to be removed here.

Also, renamed the directory containing the ui test sources from iOS13-SwiftTests to iOS13-Swift-UITests for consistency and clarity.

#skip-changelog; for https://github.com/getsentry/sentry-cocoa/issues/5165

PR #5118, opened 2025-04-22
ci: ios swift xcodegen cleanup

armcknight:
now that the original tracked iOS-Swift.xcodeproj is gone, we don't need to disambiguate the other projects with a _xcodeproj suffix (really was only necessary for the new xcodegen-managed iOS-Swift xcode project but i used it for all of them anyways) and move the iOS-Swift sources one level up from //Samples/iOS-Swift/iOS-Swift to //Samples/iOS-Swift.

Follows the xcodegen work after #5117


.github/workflows/benchmarking.yml
```diff
@@ -11,7 +11,7 @@ on:
 
       # test changes to benchmarking implementation
       - "Samples/iOS-Swift/**"
-      - "Samples/iOS-Swift/iOS-Benchmarking/**"
+      - "Samples/iOS-Benchmarking/**"
```
philprime, line 14: `h`: This looks wrong to me, if I checkout the branch and run `make xcode`, the folder is at `iOS-Swift/iOS-Benchmarking`

armcknight: hmm, you're right, but anyways this be redundant with line 13, so i'm actually just going to remove line 14 altogether.

philprime (approved):

PR #5121, opened 2025-04-22
fix(session-tracker): add session start for SDK start after didBecomeActive

### LINKED ISSUE #5069: Session Replay is not cleaned up correctly after calling Sentry.close
### Description

I created a sample in #5055 which provides a function to reload the Sentry SDK:

```swift
static func reloadSentrySDK() {
    if SentrySDK.isEnabled {
        print("SentrySDK already started, closing it")
        SentrySDK.close()
    }

    SentrySDK.start { options in
        options.debug = true

        options.sessionReplay.sessionSampleRate = Self.isSessionReplayEnabled ? 1.0 : 0.0
        options.sessionReplay.enableExperimentalViewRenderer = Self.isExperimentalViewRendererEnabled
    }
}
```

When closing the SDK the on-demand replay capturing is not restarted.
An initial investigation showed that uninstalling the `SentrySessionReplayIntegration` calls the method `SentrySessionReplay.pause()` which invalidates the `CADisplayLink` used as the trigger for creating frames:

https://github.com/getsentry/sentry-cocoa/blob/0782001b8903e178a3bb8db34f2f366ce1cc04ab/Sources/Sentry/SentrySessionReplayIntegration.m#L523C9-L527

https://github.com/getsentry/sentry-cocoa/blob/0782001b8903e178a3bb8db34f2f366ce1cc04ab/Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift#L108-L117

When installing the SDK again, there seems to be left-over state and the session replay recorder is not bootstrapped correctly.

When calling `stop()` instead of `pause()` in the `SentrySessionReplayIntegration`, the object gets deallocated and discarded, but it is not started again correctly.

philprime:
## :scroll: Description

On SDK startup we check if the app is already in foreground and manually trigger the logic of `didBecomeActive`.

Currently the session tracker expects the `didBecomeActive` notification from the application life cycle after the SDK start. If we start the SDK after the app is opened, e.g. after fetching a remote config and initializing the SDK again, or if the user is asked for consent first, session tracking will not start until the app is in background for a while again.

Please see [this comment](https://github.com/getsentry/sentry-cocoa/issues/5069#issuecomment-2821420256) for further details.

Closes #5069

- Added unit tests
- Used the `start` and `close` buttons in the sample apps

## :pencil: Checklist


philprime:
@philipphofmann I agree, it seems like somehow other changes leaked into the PR during rebasing

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
```diff
@@ -361,7 +361,21 @@ class SentrySessionTrackerTests: XCTestCase {
         
         assertNotificationNames(notificationNames)
     }
-    
+
+    func testForegroundBeforeStart() {
+        // -- Arrange --
+        goToForeground()
```
philprime, line 367: @philipphofmann should we just mock the `SentryDependencyContainer.sharedInstance.application` here and create an UI test with a real app running?

philipphofmann: I don't think we need a UI test with a real app for this. What would be the benefit? I would simply pass in the `SentryUIApplication` as a dependency and then use a mock implementation as in SentryUIViewControllerSwizzlingTests for example. We could also extract a common `SentryUIApplication` for testing from `SentryUIViewControllerSwizzlingTests` or use a local one in here. Up to you.

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -30,7 +31,9 @@ @interface SentrySessionTracker ()
 
 @end
 
-@implementation SentrySessionTracker
+@implementation SentrySessionTracker {
+    BOOL isStarted;
```
philipphofmann, line 35: `l`: Any specific reason not to define this as a regular property, like the ones above?

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -215,4 +249,16 @@ - (void)willTerminate
...
 
+- (BOOL)isAppActive
+{
+    BOOL isAppActive = NO;
+#if SENTRY_HAS_UIKIT
+    SentryUIApplication *application = SentryDependencyContainer.sharedInstance.application;
+    if (application.applicationState == UIApplicationStateActive) {
+        isAppActive = YES;
+    }
+#endif
```
philipphofmann, line 260: `m`: We could make this work on macOS by using the NSApplication with something like
```
NSApplication *application = [NSApplication sharedApplication];
if (application.isActive) {
    isAppActive = YES;
}
```
But maybe it makes sense to also create a wrapper class as `SentryUIApplication` for testability. It's also acceptable to leave out this edge case for now and follow up.

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -173,8 +200,15 @@ - (void)didBecomeActive
                 timeIntervalSinceDate:self.lastInForeground];
 
         if (secondsInBackground * 1000 >= (double)(self.options.sessionTrackingIntervalMillis)) {
+            SENTRY_LOG_DEBUG(@"[Session Tracker] App was in the background for %f seconds. "
```
philipphofmann, line 203: Thanks for the log messages 💯

Sources/Sentry/SentrySessionTracker.m
philipphofmann, line 35: `l`: wasStarted seems a bit cleaner to me. No strong opinion.
```suggestion
    BOOL wasStarted;
```

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -215,4 +249,16 @@ - (void)willTerminate
...
+- (BOOL)isAppActive
+{
+    BOOL isAppActive = NO;
+#if SENTRY_HAS_UIKIT
+    SentryUIApplication *application = SentryDependencyContainer.sharedInstance.application;
+    if (application.applicationState == UIApplicationStateActive) {
+        isAppActive = YES;
+    }
+#endif
+    return isAppActive;
```
philipphofmann, line 261: `l`: A bit shorter and cleaner, IMO.
```suggestion
- (BOOL)isAppActive
{
#if SENTRY_HAS_UIKIT
    SentryUIApplication *application = SentryDependencyContainer.sharedInstance.application;
    if (application.applicationState == UIApplicationStateActive) {
        return YES;
    }
#endif
    return NO;
```

philipphofmann (commented):
There are a few tests failing. Either something is wrong with the test setup or the session logic is broken.

Sources/Sentry/SentrySessionTracker.m
philprime: No particular reason, refactoring it.

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -151,7 +165,18 @@ - (void)didBecomeActive
...
     // synchronize to be safe.
     @synchronized(self) {
+        // If the SDK became active before started, we ignore the notification.
+        // This can happen if the SDK is cloesd and started, or if the start of the SDK was delayed,
+        // e.g. asking a user for consent first.
+        if (!self->isStarted) {
+            SENTRY_LOG_DEBUG(@"[Session Tracker] Ignoring didBecomeActive notification because the "
+                             @"tracker is not started.");
+            return;
+        }
```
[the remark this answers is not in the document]
philprime: We do need it for the already existing unit tests, as they manually call didBecomeActive. Either I have to leave this in, or change the tests.

Sources/Sentry/SentrySessionTracker.m
philprime: I agree

Sources/Sentry/SentrySessionTracker.m
philprime: Instead of a wrapper I created a protocol `SentryApplication.h` which is implemented by `SentryUIApplication.h` and `SentryNSApplication.h`, so that the dependency container does not have to care about multiple types. What do you think about this approach?

philprime:
After adding application state simulation to the unit tests, they fail with and without the changes of this PR. I'll will need to look into it with @philipphofmann to see if the tests are actually correct.

philprime:
After a discussion the existing tests are valid, therefore indicating that my changes in this PR are not valid yet.

philprime:
I manually confirmed that this actually fixes the issue with session replay by using the `Close SDK` and `Start SDK` buttons in the Sample App > Extras tab.

SentryTestUtils/ArrayAccesses.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 
 public extension Array {
     func element(at index: Int) -> Self.Element? {
-        guard count >= index else {
+        guard count > index else {
```
noahsmartin: Just flagging this to make sure it was intentional, seems like `>=` would be expected check?

Sources/Sentry/SentryNSApplication.m
```diff
@@ -0,0 +1,17 @@
+#import "SentryNSApplication.h"
+
+#if TARGET_OS_OSX
+
+#    import <AppKit/AppKit.h>
+
+@implementation SentryNSApplication
+
+- (BOOL)isActive
```
noahsmartin, line 9: Why not make these new types in Swift?

noahsmartin (approved):

SentryTestUtils/ArrayAccesses.swift
philprime: Thanks for double-checking. It's intended, because we want to make sure the index is smaller than the count, otherwise we would get an out-of-bounds exception.

Maybe this way is clearer:

```suggestion
        guard index < count else {
```

Sources/Sentry/SentryNSApplication.m
philprime: That's fair feedback. I did start from the `SentryUIApplication.m` so it was out of reflex to create the macOS variant also this way. I'll consider changing it to Swift so it's aligned with our long-term goal of migrating to Swift 👍

philprime: I am running into issues with the following error:

```
.../Build/Products/Debug/Sentry.framework/Headers/Sentry-Swift.h:288:9 Cyclic dependency in module '_SentryPrivate': _SentryPrivate -> Sentry -> _SentryPrivate
```

```
#if os(macOS)

/**
 * A helper tool to retrieve informations from the application instance.
 */
@objcMembers
class SentryNSApplication: NSObject, SentryApplication {
    func isActive() -> Bool {
        return NSApplication.shared.isActive
    }
}

#endif
```

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -83,6 +84,16 @@ - (void)start
         addObserver:self
            selector:@selector(willTerminate)
                name:SentryNSNotificationCenterWrapper.willTerminateNotificationName];
+
+    // Edge case: When starting the SDK after the app did become active, we need to call
+    //            didBecomeActive manually to start the session. This is the case when
+    //            closing the SDK and starting it again.
+    //
+    // We need to use the global app state tracker to check if the app is active, because
+    // we need to know if the app has become active before the SDK was started.
```
armcknight: I think this would be more appropriate as a headerdoc/comment at the implementation for `isAppActive`.

Also, what is an example of a scenario where `isAppActive` returns `NO`? I'd have guessed the SDK wouldn't be able to have any code executing 🤔

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -152,17 +163,26 @@ - (void)didBecomeActive
     // synchronize to be safe.
     @synchronized(self) {
         if (self.wasDidBecomeActiveCalled) {
+            SENTRY_LOG_DEBUG(@"[Session Tracker] Ignoring didBecomeActive notification because it "
```
armcknight: You shouldn't need to specify `[SessionTracker]`, the log should already have `[SentrySessionTracker.m:166]` in this message.

```suggestion
            SENTRY_LOG_DEBUG(@"Ignoring didBecomeActive notification because it "
```

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -152,17 +163,26 @@ - (void)didBecomeActive
...
+    [self startSession];
+}
+
+- (void)startSession
+{
     SentryHub *hub = [SentrySDK currentHub];
     self.lastInForeground = [[[hub getClient] fileManager] readTimestampLastInForeground];
 
-    if (nil == self.lastInForeground) {
+    if (!self.lastInForeground) {
```
armcknight: I prefer `nil == lastInForeground` as it's more obvious that it's an object being compared and not a boolean. Similar to how you compare `someBoolVar == false` above. `nil` coming before the comparison is a more controversial choice but the reasoning I've heard is that it's harder to accidentally edit it to assignment to `nil` when it comes before the object reference variable.

```suggestion
    if (nil != self.lastInForeground) {
```

Sources/Sentry/include/SentryApplication.h
```diff
@@ -0,0 +1,62 @@
...
+
+// MARK: - Shared methods
+
+- (BOOL)isActive;
+
+// MARK: - UIKit-specific methods
+
+#if SENTRY_HAS_UIKIT
+/**
+ * Returns the application state available at @c NSApplication.sharedApplication.applicationState
```
armcknight: Is it important that this implementation detail be documented here?
```suggestion
 * Returns the application state available at @c UIApplication.sharedApplication.applicationState
```

Sources/Sentry/include/SentryApplication.h
```diff
@@ -0,0 +1,62 @@
...
+ */
+@property (nonatomic, readonly) UIApplicationState applicationState;
+
+/**
+ * All application open windows.
```
armcknight: There isn't such thing as a closed window. Should this read "visible windows"?

Sources/Sentry/include/SentryApplication.h
```diff
@@ -0,0 +1,62 @@
...
+ * Retrieves connected scenes for given UIApplication
+ */
+- (NSArray<UIScene *> *)getApplicationConnectedScenes:(UIApplication *)application
+    API_AVAILABLE(ios(13.0), tvos(13.0));
+
+/**
+ * Use @c [SentryUIApplication relevantViewControllers] and convert the
+ * result to a string array with the class name of each view controller.
+ */
+- (nullable NSArray<NSString *> *)relevantViewControllersNames;
```
armcknight, line 57:
```suggestion
relevantViewControllerNames
```

Sources/Sentry/include/SentryApplication.h
```diff
@@ -0,0 +1,62 @@
...
+@property (nonatomic, readonly, nullable) NSArray<UIWindow *> *windows;
+
+/**
+ * Retrieves the application delegate for given UIApplication
+ */
+- (nullable id<UIApplicationDelegate>)getApplicationDelegate:(UIApplication *)application;
+
+/**
+ * Retrieves connected scenes for given UIApplication
+ */
```
armcknight, line 49: These are pretty redundant with the method names. If they don't add anything they should just be omitted. The one below makes more sense about uivc names

```suggestion
- (nullable id<UIApplicationDelegate>)getApplicationDelegate:(UIApplication *)application;

```

Sources/Sentry/include/SentryApplication.h
armcknight: I see you merely moved these instead of authoring them. Still, I offer you the suggestions if you want to click the buttons.

Tests/SentryTests/Integrations/Breadcrumbs/SentrySystemEventBreadcrumbsTest.swift
```diff
@@ -115,8 +120,8 @@ class SentrySystemEventBreadcrumbsTest: XCTestCase {
         
         sut = fixture.getSut(currentDevice: currentDevice)
         
-        NotificationCenter.default.post(Notification(name: UIDevice.batteryStateDidChangeNotification, object: currentDevice))
-        
+        fixture.notificationCenterWrapper.post(Notification(name: UIDevice.batteryStateDidChangeNotification, object: currentDevice))
```
armcknight, line 123: Nice find here!

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
```diff
@@ -438,26 +465,93 @@ class SentrySessionTrackerTests: XCTestCase {
...
+        sut.stop()
         #if os(iOS) || targetEnvironment(macCatalyst) || os(tvOS)
         // When the app stops, the app state is `inactive`.
         // This can be observed by viewing the application state in `UIAppDelegate.applicationDidEnterBackground`.
         fixture.application.applicationState = .inactive
+        #else
+        // The Sentry SDK should be initialized in the `NSApplicationDelegate.applicationDidFinishLaunching`
+        // At this point the app is not active yet.
+        //
+        // This can be observed by checking `NSApplication.shared.isActive` in `NSApplicationDelegate.didFinishLaunchingWithOptions`.
```
armcknight: Is this what's happening here and there are just different code flows for UIApplication and NSApplication, or is this a copypasta?

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
```diff
@@ -467,6 +561,12 @@ class SentrySessionTrackerTests: XCTestCase {
...
+        fixture.application.setIsActive(false)
```
armcknight: Same question here. If there is a difference between UIKit and AppKit, please also explain that in these comments

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## Fixes
+
+- Fix auto-start for session tracker when SDK is started after app did become active (#5121)
```
```suggestion
- Fix auto-start for session tracker when SDK is started after `appDidBecomeActive` notification was already posted (#5121)
```

SentryTestUtils/ArrayAccesses.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 
 public extension Array {
     func element(at index: Int) -> Self.Element? {
-        guard count >= index else {
+        guard index < count else {
```
armcknight: Yikes that's embarrassing! Good catch.

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -152,17 +163,26 @@ - (void)didBecomeActive
...
         // Cause we don't want to track sessions if the app is in the background we need to wait
         // until the app is in the foreground to start a session.
+        SENTRY_LOG_DEBUG(@"[Session Tracker] App was in the foreground for the first time. "
```
armcknight:
```suggestion
        SENTRY_LOG_DEBUG(@"App was in the foreground for the first time. "
```

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -173,8 +193,15 @@ - (void)didBecomeActive
                 timeIntervalSinceDate:self.lastInForeground];
 
         if (secondsInBackground * 1000 >= (double)(self.options.sessionTrackingIntervalMillis)) {
+            SENTRY_LOG_DEBUG(@"[Session Tracker] App was in the background for %f seconds. "
```
armcknight:
```suggestion
            SENTRY_LOG_DEBUG(@"App was in the background for %f seconds. "
```

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -173,8 +193,15 @@ - (void)didBecomeActive
...
+                             @"Starting a new session.",
+                secondsInBackground);
             [hub endSessionWithTimestamp:self.lastInForeground];
             [hub startSession];
+        } else {
+            SENTRY_LOG_DEBUG(@"[Session Tracker] App was in the background for %f seconds. Not "
```
armcknight:
```suggestion
            SENTRY_LOG_DEBUG(@"App was in the background for %f seconds. Not "
```

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -215,4 +242,10 @@ - (void)willTerminate
     self.wasDidBecomeActiveCalled = NO;
 }
 
+- (BOOL)isAppActive
+{
+    id<SentryApplication> application = SentryDependencyContainer.sharedInstance.application;
+    return application.isActive;
```
armcknight: Any reason not to golf it? 🏌🏻 

```suggestion
     return SentryDependencyContainer.sharedInstance.application.isActive;
```

Sources/Sentry/include/SentryApplication.h
armcknight: Reading the implementation it seems it's windows connected to scenes only:
```suggestion
 * All windows connected to scenes.
```

armcknight: (We could consider also just renaming the property to `sceneWindows`)

armcknight (commented):
Thanks for doing this work, some nice improvements in here. There are a couple test cases I wanted to understand better but these are the rest of the comments I have.

Samples/macOS-Swift/Shared/Info.plist
```diff
@@ -25,7 +25,7 @@
 	<key>GIT_COMMIT_HASH</key>
 	<string>&lt;sha&gt;</string>
 	<key>GIT_STATUS_CLEAN</key>
-	<string>&lt;status&gt;</string>
+	<string>0</string>
```
armcknight: Please revert this change. Maybe we should add a check for this.

```suggestion
	<string>&lt;status&gt;</string>
```

Sources/Sentry/SentrySessionTracker.m
philprime: When bootstrapping the SDK from `AppDelegate.didFinishLaunchingWithOptions` the `isAppActive` is `NO`, because the app becomes active afterwards.

If the SDK is bootstrapped later, e.g. after the user has been asked for privacy consent, the `isAppActive` is `YES`. This also applies to SDK restarts in our sample apps.

Sources/Sentry/SentrySessionTracker.m
philprime: Fine to me, it's just not the code style I am used to so it was out of reflex. (I always go for the `IF <VARIABLE_IN_FOCUS> <OPERATOR> <EXPECTED_VALUE> THEN`, but it's not worth enforcing a strict ruleset here).

I can't use the suggestion because `!self.lastInForeground` is `nil == self.lastInForeground`, so your suggestion would flip the conditional.

Sources/Sentry/SentrySessionTracker.m
philprime: I also just noticed the comment for the global app state tracker was outdated from another approach, thanks for the comment.

CHANGELOG.md
philprime: Is it important for SDK users to know that we use the UIApplication life cycle notifications for this? If not, I do not see a reason to be so specific.

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philprime: Not sure what happened here and why it's the wrong comment.
I'll update it accordingly.

Sources/Sentry/include/SentryApplication.h
philprime: I appreciate the feedback, and yes this file definitely needs some cleanup.
Changing this will increase the PR size by touching unrelated files, so let's do this in a followup PR.

Sources/Sentry/include/SentryApplication.h
philprime: As mentioned in the other comment, I'll change the docs but I won't change the method signatures in this PR.
Thanks again for mentioning.

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -215,4 +237,9 @@ - (void)willTerminate
     self.wasDidBecomeActiveCalled = NO;
 }
 
+- (BOOL)isAppActive
+{
+    return SentryDependencyContainer.sharedInstance.application.isActive;
```
philippsuess: `m`: I would prefer passing the application in as a dependency via the constructor for testability, please.

Sources/Sentry/SentrySessionTracker.m
philippsuess: I think that's legit. Thanks.

SentryTestUtils/ArrayAccesses.swift
philippsuess: `m`: I think it would make sense to add a unit test for the test code to avoid a regression. We've been bitten by bugs in our test code quite often.

SentryTestUtilsTests/TestNSNotificationCenterWrapperTests.swift
```diff
@@ -0,0 +1,152 @@
+@testable import SentryTestUtils
+import XCTest
+
+class TestNSNotificationCenterWrapperTests: XCTestCase {
```
philippsuess, line 6: Thanks for adding tests for the test code 🥇

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -1,15 +1,18 @@
...
 #import "SentryFileManager.h"
 #import "SentryHub+Private.h"
 #import "SentryInternalNotificationNames.h"
 #import "SentryLog.h"
+#import "SentryNSApplication.h"
 #import "SentryNSNotificationCenterWrapper.h"
 #import "SentryOptions+Private.h"
 #import "SentrySDK+Private.h"
 #import "SentrySwift.h"
+#import "SentryUIApplication.h"
```
philippsuess: `m`: We only need the protocol. We don't need to import the headers for the implementations.
```suggestion
#import "SentryNSNotificationCenterWrapper.h"
#import "SentryOptions+Private.h"
#import "SentrySDK+Private.h"
#import "SentrySwift.h"
```

Sources/Sentry/include/SentryNSNotificationCenterWrapper.h
```diff
@@ -18,6 +18,12 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly, copy, class) NSNotificationName willTerminateNotificationName;
 #endif
 
+#if SENTRY_HAS_UIKIT
+// macOS apps can not enter the background, therefore this notification
+// is not available there.
```
philippsuess, line 23: Thanks for the comment.

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
```diff
@@ -196,8 +202,18 @@ class SentrySessionTrackerTests: XCTestCase {
...
+        // This causes multiple session tracker to be registered as observers, until they are fully released and
+        // weak references are nil.
+        // Using an autoreleasepool to ensure that the deallocation happens before the test continues.
+        autoreleasepool {
+            crashInBackground()
+            assertNoSessionSent()
+
+            // Manually deallocate the previous sut to avoid race-conditions of duplicate observers
+            sut = nil
+        }
```
philippsuess: `m`: Shouldn't we then fix this for all test cases? There are a few tests calling `sut = fixture.getSut()`. I think we don't have to rely on dealloc to call `removeObserver`. We could add a method called setNewSut, that calls stop before creating and assigning a new sut. Or we could simply call stop every time before we get a new sut. Up to you.
```
private func setNewSut() {
    sut.stop()
    sut = fixture.getSut()
}
```

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
```diff
@@ -438,35 +465,109 @@ class SentrySessionTrackerTests: XCTestCase {
...
+        XCTAssertEqual(event.status, SentrySessionStatus.ok)
+
+        stopSut()
+        // Ideally we would expect the SDK to send the `exited` session status when calling `stop`.
+        //
+        // As the test case `testKillAppWithoutNotificationsAndNoCrash_EndsWithAbnormalSession` is calling `stop` and
+        // expects that to result in an `abnormal` session next time the tracker is started, we have a contradiction.
+        //
+        // To keep existing tests working as intended, we decided to keep the `abnormal` status being send after the
+        // restart has happened.
```
philippsuess: `h`: I think that's actually a bug you found here. When the user closes the SDK, which then calls stop, the session tracker stops working and doesn't listen to notifications anymore. Instead, I think we should end the session as healthy when closing the SDK, but we should double-check this with Android, for example. Previously, we simply didn't create a session. Now when users start the SDK, we would create plenty of abnormal sessions, which will confuse users.

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -152,17 +160,25 @@ - (void)didBecomeActive
...
+                @"Ignoring didBecomeActive notification because it was already called.");
             return;
         }
         self.wasDidBecomeActiveCalled = YES;
     }
 
+    [self startSession];
+}
+
+- (void)startSession
```
philippsuess: `h`: Previously, we only started a session if `self.wasDidBecomeActiveCalled == NO` and we always set `wasDidBecomeActiveCalled = YES` before starting a session. If the hybrid SDKs post `SentryHybridSdkDidBecomeActiveNotification` after we call `startSession` when the app is active, we then end the current session and start a new one, which isn't correct. Also see https://github.com/getsentry/sentry-cocoa/blob/ba8a46b7ac0863cf2c8e45b5386ffb4ad6c141d4/Sources/Sentry/include/SentryInternalNotificationNames.h#L1-L6

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
```diff
@@ -438,35 +465,109 @@ class SentrySessionTrackerTests: XCTestCase {
...
+        XCTAssertEqual(event.status, SentrySessionStatus.abnormal)
+
+        event = try XCTUnwrap(fixture.client.captureSessionInvocations.invocations.element(at: 2))
+        XCTAssertEqual(event.status, SentrySessionStatus.ok)
+
+        event = try XCTUnwrap(fixture.client.captureSessionInvocations.invocations.element(at: 3))
+        XCTAssertEqual(event.status, SentrySessionStatus.exited)
+
+        // Assert that there are no more invocations
+        XCTAssertEqual(fixture.client.captureSessionInvocations.invocations.count, 4)
```
philippsuess: `h`: That doesn't seem right. When starting the SessionTracker we first end the session as abnormal. Then we start a new one. There shouldn't be an exited session. The root cause is that the hub keeps a reference to the current session internally via `_session`. It then ends the `_session` as exited when calling `hub.startSession`. When the user calls `stopSut`, he does that via `SentrySDK.close`, which removes the hub from the SDK. So your test code needs to set a new hub before calling `sut.start()`. We can achieve this by adding `fixture.setNewHubToSDK()` to the `stopSut` method. Then we can adapt the test case to the following


```suggestion
        event = try XCTUnwrap(fixture.client.captureSessionInvocations.invocations.element(at: 1))
        XCTAssertEqual(event.status, SentrySessionStatus.abnormal)

        event = try XCTUnwrap(fixture.client.captureSessionInvocations.invocations.element(at: 2))
        XCTAssertEqual(event.status, SentrySessionStatus.ok)

        // Assert that there are no more invocations
        XCTAssertEqual(fixture.client.captureSessionInvocations.invocations.count, 3)
```

It took me a while to figure this out. I actually thought the code is broken, which would be bad 😅. I'm happy it's the tests.

philippsuess: This is how the test could look like:
```
func testRestartInForeground_shouldStartNewSession() throws {
    // -- Arrange --
    startSutInAppDelegate()
    goToForeground()

    XCTAssertEqual(fixture.client.captureSessionInvocations.invocations.count, 1)
    var event = try XCTUnwrap(fixture.client.captureSessionInvocations.invocations.element(at: 0))
    XCTAssertEqual(event.status, SentrySessionStatus.ok)

    stopSut()
    fixture.setNewHubToSDK()

    // Ideally we would expect the SDK to send the `exited` session status when calling `stop`.
    //
    // As the test case `testKillAppWithoutNotificationsAndNoCrash_EndsWithAbnormalSession` is calling `stop` and
    // expects that to result in an `abnormal` session next time the tracker is started, we have a contradiction.
    //
    // To keep existing tests working as intended, we decided to keep the `abnormal` status being send after the
    // restart has happened.

    // -- Act --
    sut.start()

    // -- Assert --
    event = try XCTUnwrap(fixture.client.captureSessionInvocations.invocations.element(at: 1))
    XCTAssertEqual(event.status, SentrySessionStatus.abnormal)

    event = try XCTUnwrap(fixture.client.captureSessionInvocations.invocations.element(at: 2))
    XCTAssertEqual(event.status, SentrySessionStatus.ok)

    // Assert that there are no more invocations
    XCTAssertEqual(fixture.client.captureSessionInvocations.invocations.count, 3)
}
```

Sources/Sentry/SentrySessionTracker.m
philprime: Good idea, I'll apply that.

SentryTestUtils/ArrayAccesses.swift
philprime: I added tests in https://github.com/getsentry/sentry-cocoa/pull/5431

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philprime: I tried different combinations of using `autoreleasepool` and they didn't work.
Even just changing this case, will fail the test:

```swift
        crashInBackground()
        assertNoSessionSent()

        autoreleasepool {
            // Manually deallocate the previous sut to avoid race-conditions of duplicate observers
            sut = nil
        }
```

The problem isn't the `sut.stop()`, but instead that the `TestNSNotificationCenterWrapper.addObserver*Invocations` hold weak references to the `sut`, so when the new `sut` is created and observes for notifications, we have two `sut` in the observers.


While writing this comment I realized that we might actually miss the correct `removeObserver` calls, causing the `WeakReferences` to stay in the list observers, while their wrapped value turns `nil`.

I'll check if extending the `removeObserver` mocking will allow the removal of `autoreleasepool`

image

philprime: The problem here is, that we are abusing the invocations as the list of current observers.

image

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philprime: Discussing this internally, will follow up here for documentation.

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philprime: I was able to remove the usage of `autoreleasepool` by not using the invocations as means to track the observers, and by correctly overriding the remaining removeObserver calls.

Sources/Sentry/SentrySessionTracker.m
philprime: I agree this is not solved correctly now and I would propose to move the `self.wasDidBecomeActiveCalled` to the `startSession` method to debounce that as well.

But it also raises another edge case now, when the SDK is (re-)started after the app has become active (which should start sessions as of this PR), but we actually don't want to start it yet, because we want to wait for the Hybrid SDK to post the notification instead.

This sort of delayed init from the Hybrid SDKs seems patched together.

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philprime: Internal discussion is inconclusive so far, but I would agree that recording a ton of abnormal session is definitely not the way to go as it can indicate false-positive issues.

I see two options here:
1) Clean exit the session on SDK shutdown (either flushing the session on close or on next restart)
2) Mark as "exited-on-close" and on SDK restart it continues the previous session.

I would go for option 1

Sources/Sentry/SentrySessionTracker.m
philprime: I moved the check for `wasDidBecomeActiveCalled` to the `startSession` and added a `self. wasDidBecomeActiveCalled = NO` to the `stop` method.

All tests did pass locally.

Sources/Sentry/SentrySessionTracker.m
```diff
@@ -105,6 +118,9 @@ - (void)stop
         removeObserver:self
                   name:SentryNSNotificationCenterWrapper.willTerminateNotificationName];
 #endif
+    // Reset the `wasDidBecomeActiveCalled` flag to ensure that the next time
+    // `didBecomeActive` is called, it will start a new session.
+    self.wasDidBecomeActiveCalled = NO;
```
philippsuess: `m`: I think we should find a better name for `wasDidBecomeActiveCalled` cause it's now set to true in `startSession` even when not calling `didBecomeActive`. `wasStartSessionCalled` is my first shot.

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philippsuess: Yes, I would also go for option 1. When you close the SDK, we simply end the session as it is and send it.

Sources/Sentry/SentrySessionTracker.m
philippsuess: This should work now. Ideally there would be a test that fails with the previous approach, but I leave this up to you.

philipphofmann (commented):
Almost LGTM. There are still two unsesolved comments. Once we fix these, we should be goog do merge.

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philprime: I had to add a flag `gracefully` to the stop method so we can differentiate between simulated crashes in the tests `crashSut()` to normal stop `stopSut()`.

Sources/Sentry/SentrySessionTracker.m
philprime: I added tests.

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philippsuess: That looks a bit weird. Only the tests use `gracefully:false`. I think the tests previously called stop to unsubscribe for all notifications. Instead, I would add another method named removeObservers or something like this, which the tests can call. I don't think that dealloc should call the stop, ending the session. That's changing the logic. I think only stop called from the SessionTrackingIntegration should end the session.

philippsuess: So basically this: https://github.com/getsentry/sentry-cocoa/pull/5461

philprime: > That looks a bit weird. Only the tests use gracefully:false. I think the tests previously called stop to unsubscribe for all notifications.

I'll change it to a `reset` method

>  I don't think that dealloc should call the stop, ending the session. That's changing the logic

You added this in 2022, how would calling `stop()` from `dealloc` be changing the logic?

https://github.com/getsentry/sentry-cocoa/blame/0c0834c833b3f7b2fbc12ab4427a535e809fc11a/Sources/Sentry/SentrySessionTracker.m#L112

philipphofmann (commented):
Almost LGTM. Just one open comment https://github.com/getsentry/sentry-cocoa/pull/5121#discussion_r2161035116.

Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
philippsuess: @philprime, stop now also ends the session, which it didn't do previously.

CHANGELOG.md
philippsuess, line 7: `m`: Maybe we should also add that this now correctly ends sessions when you close the SDK.

philprime: I agree, I'll add a second changelog line

PR #5125, opened 2025-04-22
ci: add ui test run on macos 15 / xcode 16.3

armcknight:
Noticed we aren't running UI tests on the latest toolchains.

#skip-changelog

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -111,20 +111,17 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # As of 25th March 2025, the preinstalled iOS simulator version is 16.4 for macOS 13 and Xcode 14.3.1; see
-          # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#installed-sdks
           - runs-on: macos-13
             xcode: "14.3.1"
+            device: "iPhone 14 (16.4)"
 
-          # As of 25th March 2025, the default iOS simulator version is 17.5 for macOS 14 and Xcode 15.4; see
-          # https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md#installed-sdks
           - runs-on: macos-14
             xcode: "15.4"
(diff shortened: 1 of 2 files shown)


armcknight:
We need to explicitly define the destination on these xcodebuild invocations, leaving it to chance isn't cutting it anymore:
```
[00:01:16]: ▸ --- xcodebuild: WARNING: Using the first of multiple matching destinations:
[00:01:16]: ▸ { platform:macOS, arch:arm64, variant:Mac Catalyst, id:0000FE00-B211F5635431B634, name:My Mac }
[00:01:16]: ▸ { platform:macOS, arch:x86_64, variant:Mac Catalyst, id:0000FE00-B211F5635431B634, name:My Mac }
[00:01:16]: ▸ { platform:macOS, arch:arm64, variant:Designed for [iPad,iPhone], id:0000FE00-B211F5635431B634, name:My Mac }
```
and then it tries to run iOS-Swift against macOS, which fails the build

philipphofmann (approved):
Thanks

armcknight:
I wound up discovering the device specs per matrix entry were removed in https://github.com/getsentry/sentry-cocoa/pull/5031, so I put it back to fix the issue.

.github/workflows/ui-tests.yml
```diff
@@ -105,21 +105,23 @@ jobs:
...
-          # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#installed-sdks
           - runs-on: macos-13
             xcode: "14.3.1"
+            device: "iPhone 14 (16.4)"
 
-          # As of 25th March 2025, the default iOS simulator version is 17.5 for macOS 14 and Xcode 15.4; see
-          # https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md#installed-sdks
           - runs-on: macos-14
             xcode: "15.4"
+            device: "iPhone 15 (17.2)"
```
armcknight, line 120:
```suggestion
            device: "iPhone 15 (17.5)"
```

PR #5126, opened 2025-04-23
fix: change identifier iOS-SwiftUITests to iOS-Swift-UITests

philprime:
While refreshing code signing certificates and provisioning profiles I noticed an inconsistency in the identifiers of `iOS-Swift-UITests` and `iOS-SwiftUITests`.

As the one with two dashes is more prominent in the repository, I removed `iOS-SwiftUITests`.

#skip-changelog


philipphofmann (approved):
Thanks

PR #5130, opened 2025-04-23
Revert "fix: change identifier iOS-SwiftUITests to iOS-Swift-UITests …

armcknight:
This change was similarly made in https://github.com/getsentry/sentry-cocoa/pull/5115 which I already had open, and is now making it difficult to merge that branch with `main`. I would appreciate if we could back this out and let it happen as part of that PR.

#skip-changelog

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift.xcodeproj/project.pbxproj
+++ b/Samples/iOS-Swift/iOS-Swift.xcodeproj/project.pbxproj
@@ -742,7 +742,7 @@
 			name = "iOS-Swift-UITests";
 			packageProductDependencies = (
 			);
-			productName = "iOS-Swift-UITests";
+			productName = "iOS-SwiftUITests";
 			productReference = 7B64386826A6C544000D0F65 /* iOS-Swift-UITests.xctest */;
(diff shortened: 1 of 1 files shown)


philprime:
I do not understand why we should revert a small fix in favor of a large refactoring of the Xcode projects. The PRs to use xcodegen will take more time, due to the complexity of making sure that all the relevant settings are still correct. 
Ideally the xcodegen PRs should output almost the same projects (with different ordering of files) as before, therefore if the xcodegen PRs *does not* do this correctly, we'll have to revisit this again for the third time.

armcknight:
The xcodegen PRs currently only cover sample projects, they are low stakes. I don't have any current plans to migrate the SDK project to being generated via XcodeGen. I would sooner simply remove the project altogether and have it integrated into the sample projects using SPM via a local path dependency. In the past, I have written scripts to validate that all the settings remain unchanged when making such refactors using `xcodebuild -showBuildSettings`. I briefly touched on that in https://github.com/getsentry/sentry-cocoa/discussions/2024. But this is all a discussion for another day.

PR #5131, opened 2025-04-23
fix(profiling): use correct rate limit key for UI profiling

armcknight:
We discovered that the mobile clients need to use a new key for rate limiting envelopes sending the new UI profiling chunks.

TBD if we need to also change the envelope type: https://github.com/getsentry/sentry-cocoa/blob/0a19aee3686feb5d51083f8a75d7b4f35e2ac19d/Sources/Sentry/Profiling/SentryProfilerSerialization.mm#L345

Answer: no


phacops (approved):

phacops:
> Do you also need to update `kSentryDataCategoryNameProfileChunk`?

I wouldn't update the variable name. We'll only have this one, for the `profile_chunk` item type.

> I'm a bit surprised that no tests need an update. `SentryContinuousProfilerTests` and `SentryDataCategoryMapperTests` use the constant `profile_chunk`.

The hard-coded `profile_chunk` is used in the tests to check the value of the item type and this doesn't change. The data category mapping test uses the variable `.profileChunk` which has the new value `profile_chunk_ui` now since we just changed the value of the constant without renaming, so no need to adjust anything.

philipphofmann:
> I wouldn't update the variable name. We'll only have this one, for the `profile_chunk` item type. 

I'm sorry. Copy paste mistake. I wanted to write `SentryEnvelopeItemTypeProfileChunk` https://github.com/getsentry/sentry-cocoa/blob/0a19aee3686feb5d51083f8a75d7b4f35e2ac19d/Sources/Sentry/include/HybridPublic/SentryEnvelopeItemType.h#L16

> > I'm a bit surprised that no tests need an update. `SentryContinuousProfilerTests` and `SentryDataCategoryMapperTests` use the constant `profile_chunk`.
> 
> The hard-coded `profile_chunk` is used in the tests to check the value of the item type and this doesn't change. The data category mapping test uses the variable `.profileChunk` which has the new value `profile_chunk_ui` now since we just changed the value of the constant without renaming, so no need to adjust anything.

Ah this test uses the envelope item type. Therefore, no change required: https://github.com/getsentry/sentry-cocoa/blob/0a19aee3686feb5d51083f8a75d7b4f35e2ac19d/Tests/SentryTests/Networking/SentryDataCategoryMapperTests.swift#L12

phacops:
> I wanted to write `SentryEnvelopeItemTypeProfileChunk` 

So indeed, no need to change since this the item type stays as `profile_chunk`.

Issue #2508, opened 2022-12-09
detectANRs thread safe issue


iDevPro:
### Platform

iOS

### Installed

CocoaPods

### Version

7.25.1

### Steps to Reproduce

1. Configure sentry in code like this:
```
SentrySDK.start { options in 
     options.enableAppHangTracking = true
}
```
2. Toggle checkbox ThreadSanitizer in Scheme => Edit scheme => Run (debug) => Diagnostics => Thread Sanitizer

3. Run application and wait detectANR function called :)
4. Result:
![Снимок экрана 2022-12-09 в 14 39 37](https://user-images.githubusercontent.com/1888230/206694519-f4a7011e-6d44-41d0-9d7e-c22b9fca69aa.png)
![Снимок экрана 2022-12-09 в 14 40 23](https://user-images.githubusercontent.com/1888230/206694863-45a4f45b-e626-446d-8555-55921af810c2.png)
![Снимок экрана 2022-12-09 в 14 40 40](https://user-images.githubusercontent.com/1888230/206694876-e34c3eb7-aa30-41ee-9c4e-145f9be7ee2a.png)
![Снимок экрана 2022-12-09 в 14 40 58](https://user-images.githubusercontent.com/1888230/206694903-d7bda57b-daac-4669-b97d-1cab0ee2833b.png)

### Expected Result

No thread issues

### Actual Result

Thread issue

philipphofmann:
Thanks for reporting this, @iDevPro. I just tried that with version 7.31.3 and couldn't reproduce the issue. Please try the latest version and let us know if the problem persists.

iDevPro:
[philipphofmann](https://github.com/philipphofmann), Hello.
I tried check version that you said me, but problem persists :(

<img width="146" alt="Снимок экрана 2023-01-12 в 14 40 58" src="https://user-images.githubusercontent.com/1888230/

<img width="1121" alt="Снимок экрана 2023-01-12 в 14 40 49" src="https://user-images.githubusercontent.com/1888230/212057812-74d8902e-e05e-4bde-9b74-7e8724a2768e.png">
212057802-dd718f10-ba2b-45d0-abb1-f732c1426ce9.png">

philipphofmann:
Thanks for the update, @iDevPro. Although I can't reproduce this, I think I know what the problem is
We read and write to reported in different threads

Main Thread
[...]

PR #5137, opened 2025-04-25
fix: Race condition in ANRTrackerV1

philipphofmann:
## :scroll: Description

Use an atomic_bool to fix race conditions for accessing reported in the ANRTrackerV1.

Fixes GH-2508

With testWithActualDispatchQueueAndThreadWrapper locally having the thread sanitizer enabled.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixes
 
-- Race conditions in ANRTrackerV1 (#5137)
+- Race condition in ANRTrackerV1 (#5137)
 
 ### Improvements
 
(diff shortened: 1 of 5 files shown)


philprime:
@philipphofmann IIRC this should actually have been experimental in the first place.
This kept getting pushed back, @kahest should we even add the experimental note?

philprime (approved):
LGTM, but do we know why Codecov is adding annotations that the code is not covered by tests?

philipphofmann:
> LGTM, but do we know why Codecov is adding annotations that the code is not covered by tests?

No idea, that part of the code is covered 100% with tests.

PR #5139, opened 2025-04-25
meta: correct version setting

armcknight:
I noticed something about how we version some things while going through the XcodeGen refactor: we were setting `CURRENT_PROJECT_VERSION = 8.49.1` for the SentrySDK and SentrySwiftUI framework products. However, that setting is supposed to only be an integer value. The semantic version is supposed to go into MARKETING_VERSION.

We also had a couple disparate places where we would write the semver. So, this unifies them all into a single source of truth, and uses the correct build settings.


Utils/VersionBump/main.swift
```diff
@@ -16,14 +16,14 @@ let files = [
...
-    "./Sources/Configuration/SDK.xcconfig",
+    "./Samples/Config/Versioning.xcconfig",
+    "./Sources/Configuration/SDK.xcconfig", // T O D O: use a single file here that is included into either other one. or even one single file for everything
     "./Sources/Configuration/SentrySwiftUI.xcconfig"
 ]
 
 let args = CommandLine.arguments
 
-let semver: StaticString = "([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*))?(?:\\+[0-9A-Za-z-]+)?"
+let semver: StaticString = "MARKETING_VERSION = ([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*))?(?:\\+[0-9A-Za-z-]+)?"
```
philprime, line 26: `h`: This breaks the release flow!!

This utility is used in the release flow [like you can see here](https://github.com/getsentry/sentry-cocoa/actions/runs/14735024914/job/41358848051#step:3:3).

The `semver` is used to parse the argument, not the file! Please test these kind of changes locally like this:

```
$ make bump-version TO=8.49.2
--> Clean VersionBump
cd Utils/VersionBump && rm -rf .build && swift build
Fetching https://github.com/sharplet/Regex.git from cache
Fetching https://github.com/kareman/SwiftShell.git from cache
Fetched https://github.com/kareman/SwiftShell.git from cache (0.45s)
Fetched https://github.com/sharplet/Regex.git from cache (0.45s)
Computing version for https://github.com/sharplet/Regex.git
Computed https://github.com/sharplet/Regex.git at 2.1.1 (1.06s)
Computing version for https://github.com/kareman/SwiftShell.git
Computed https://github.com/kareman/SwiftShell.git at 5.1.0 (0.62s)
Creating working copy for https://github.com/kareman/SwiftShell.git
Working copy of https://github.com/kareman/SwiftShell.git resolved at 5.1.0
Creating working copy for https://github.com/sharplet/Regex.git
Working copy of https://github.com/sharplet/Regex.git resolved at 2.1.1
Building for debugging...
[27/27] Applying VersionBump
Build complete! (9.05s)
--> Bumping version from 8.49.2
./Utils/VersionBump/.build/debug/VersionBump 8.49.2
/Volumes/Developer/getsentry/sentry-cocoa/Utils/VersionBump/main.swift:29: version number must fit x.x.x format
make: *** [bump-version] Error 1
```

armcknight: Well that's embarrassing. I marked this PR for review on accident. In any case I addressed the TODO left for myself and fixed this issue. 

Please be kind in code reviews. We all make mistakes, we're all one team.

Ideally we have unit tests for this so we can also validate it in CI: https://github.com/getsentry/sentry-cocoa/issues/5177

philprime: > Well that's embarrassing. I marked this PR for review on accident. In any case I addressed the TODO left for myself and fixed this issue.

Thank you for fixing the issue. Yes it can happen to accidentally mark something as ready while being TODO.

> Please be kind in code reviews. We all make mistakes, we're all one team.

It was not my intention to seem rude. This is just critical code so I wanted to make sure we are aware that this could easily break our release flow and so we should all be extra careful when touching it.

philprime (approved):
LGTM, I just want to point out that with this PR we are not only changing the sample projects but also configurations of the SDK (`SDK.xcconfig`).
I don't see this as an issue, it just doesn't match with internal communication of this refactor not touching the SDK projects yet.

I would appreciate a second review by @philipphofmann on this PR please.

Sources/Configuration/Versioning.xcconfig
```diff
@@ -1,2 +1,2 @@
 CURRENT_PROJECT_VERSION = 1
-MARKETING_VERSION = 8.49.1
+MARKETING_VERSION = 8.49.2
```
philipphofmann, line 2: `h`: Something is off here. Here we have `8.49.2` and sometimes we have `8.49.3`, but we already released 8.50.0. All versions should already be `8.50.0`.

Package.swift
```diff
@@ -12,12 +12,12 @@ let package = Package(
     targets: [
         .binaryTarget(
                     name: "Sentry",
-                    url: "https://github.com/getsentry/sentry-cocoa/releases/download/8.49.2/Sentry.xcframework.zip",
+                    url: "https://github.com/getsentry/sentry-cocoa/releases/download/8.49.3/Sentry.xcframework.zip",
                     checksum: "85cf16b4fa2dd0ae58b1e5c62fd8c829c94e5dc92251c3a1de8e71b3fed195ef" //Sentry-Static
```
philipphofmann, line 16: `h`: If the URL to the zip changes, the checksum must also change. Otherwise, we could break SPM, when compiling from the main branch. Anyways this should be 8.50.0, and we shouldn't update this manually. Only the release workflow should do that.

Sentry.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "Sentry"
-  s.version      = "8.49.2"
+  s.version      = "8.49.3"
```
philipphofmann, line 3: `h`: We shouldn't have to update this manually. Only the GitHub release workflow should do that.

philipphofmann (commented):
The approach looks good, but I don't think we can touch the version numbers manually. Only the release workflow should do this.

I'm not too much worried about the change of `CURRENT_PROJECT_VERSION` in `Sources/Configuration/SDK.xcconfig` because as already pointed out by @armcknight it should be an integer of a float, which it wasn't. So it was already broken.

Utils/VersionBump/main.swift
armcknight: That's what `h` is for, chill with all the exclamation marks next time, that does nothing to raise extra awareness, it only looks like you're screaming. You can also reach out via slack if you need to escalate.

Sources/Configuration/Versioning.xcconfig
armcknight: I haven't merged from `main` since 8.50.0 was released, and I did a find/replace on some of these I think, so I'll get the versions straightened out here and for the other ones you mentioned.

armcknight:
I'm converting this back to a draft to move it out of the way of the xcodegen PRs.

philprime:
@armcknight is this PR still relevant or should we close it?

armcknight:
I think it is still valid but this branch is so hopelessly behind i'll just try it again later.

armcknight:
OK, made most of the changes in #5700 and #5701

Issue #5142, opened 2025-04-26
macOS launch profiling can trigger inadvertently


armcknight:
### Description

In order to know whether a launch profile should start before the call to start the Sentry SDK, we write a config file to check in a `+ [load]` call. We construct a path for this file using NSCachesDirectory. On iOS, this is automatically scoped to the app, since all apps operate in a sandbox. However, with macOS apps distributed outside the Mac App Store, there is no automatic sandboxing. In such apps, we wind up writing this file to the top level of the caches directory. This means that any macOS app that ships the Sentry SDK can read/write this file, and one app can leave a configuration that results in any other app running the launch profiler.

The fix will entail constructing the path for the config file using the app's bundle ID.

kahest:
this has been fixed in [8.49.2](https://github.com/getsentry/sentry-cocoa/releases/tag/8.49.2)

PR #5143, opened 2025-04-28
fix(profiling): Change path of non-sandboxed macOS app launch profile configuration

armcknight:
reproduces the issue reported in https://github.com/getsentry/sentry-cocoa/issues/5142 and fixed in https://github.com/getsentry/sentry-cocoa/pull/5144

This extends the macOS-Swift project to have four app targets: two that are sandboxed, and two that are not.

This allows us to simulate the scenario where one mac app configures launch profiling thus writing the launch profile config file, and then a different app with a different bundle id tries to configure the SDK to never profile launches, but then still reads the config file from the common cache location and has the profiler run when it shouldn't. 

The simulation runs the same way for each pair of apps (sandboxed and nonsandboxed). Only the nonsandboxed pair test should fail, since the sandboxed pair gets an automatically namespaced caches path.

#skip-changelog


Samples/macOS-Swift/Shared/AppDelegate.swift
```diff
@@ -5,6 +5,7 @@ import Sentry
 class AppDelegate: NSObject, NSApplicationDelegate {
 
     func applicationDidFinishLaunching(_ aNotification: Notification) {
+        print("ProcessInfo.processInfo.arguments: \(ProcessInfo.processInfo.arguments)")
```
philprime, line 8: `m`: Looks like a left-over to me, if not necessary let's remove it to keep the code minimal

philprime:
I went ahead and fixed the paths to the Entitlements, as they were incorrect on my case-sensitive file-system.

Plans/iOS13-Swift_Base.xctestplan
```diff
@@ -11,10 +11,7 @@
   "defaultOptions" : {
     "codeCoverage" : false,
     "commandLineArgumentEntries" : [
-      {
```
philprime, line 14: This argument is not used anymore

Samples/macOS-Swift/macOS-Swift.xcodeproj/project.pbxproj
```diff
@@ -211,27 +427,127 @@
...
 			dependencies = (
+				84B5A9ED2DBD79E7003327C9 /* PBXTargetDependency */,
 				6308534D2440C5F500DDE4CE /* PBXTargetDependency */,
 			);
 			name = "macOS-Swift";
 			productName = "macOS-Swift";
 			productReference = 637AFDEB243B04320034958B /* macOS-Swift.app */;
 			productType = "com.apple.product-type.application";
 		};
+		84B5A8F32DBD1A86003327C9 /* macOS-Swift-Other */ = {
```
philprime, line 438: This new target is necessary for having two different macOS apps in ui tests to verify isolation of stored data

philprime (approved):
LGTM, still need wait for green CI

Sources/Sentry/include/SentryFileManager.h
```diff
@@ -161,6 +162,12 @@ SENTRY_EXTERN void writeAppLaunchProfilingConfigFile(
  */
 SENTRY_EXTERN void removeAppLaunchProfilingConfigFile(void);
 
+SENTRY_EXTERN NSString *_Nullable sentryStaticBasePath(void);
+
+#    if defined(SENTRY_TEST) || defined(SENTRY_TEST_CI)
+SENTRY_EXTERN void removeBasePath(void);
```
philipphofmann, line 168: `m`: This name is a bit confusing. Maybe something like this is cleaner.
```suggestion
SENTRY_EXTERN void removeSentryStaticBasePath(void);
```

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+> ![Important]
+> Version 8.21.0 introduced an issue for macOS apps that run without a sandbox (i.e. distributed outside the Mac App Store). We recommend upgrading to at least this version.
```
philipphofmann, line 6: `m`: I think it's essential to point out what the issue is. Otherwise, people will wonder what issue?? Same applies for 8.21.0 please.
```suggestion
> Version 8.21.0 introduced an issue for app launch profiling **only for macOS apps that run without a sandbox** (i.e. distributed outside the Mac App Store). This issue could lead to starting the app launch profiler even when it's not configured via the options. We recommend upgrading to at least this version.
```

philipphofmann (approved):
I only looked at the changelog and the actual fix but not at the sample apps. I trust @philprime s review.

Samples/macOS-Swift/Shared/AppDelegate.swift
```diff
@@ -5,23 +5,32 @@ import Sentry
...
-                options.profilesSampleRate = 1
+
+            // Set the profilesSampleRate to `nil` to enable the configureProfiling block
+            options.profilesSampleRate = nil
+            options.configureProfiling = {
+                $0.profileAppStarts = ProcessInfo.processInfo.arguments.contains("--io.sentry.enable-profile-app-starts")
+                $0.sessionSampleRate = 1
             }
+
+            options.enableUncaughtNSExceptionReporting = false
```
philipphofmann, line 23: `l`: The default is false. We can remove this.
```suggestion
```

PR #5144, opened 2025-04-28
fix(profiling): fix macOS launch profiling issue

armcknight:
fixes issue reported in https://github.com/getsentry/sentry-cocoa/issues/5142 and reproduced in the tests in https://github.com/getsentry/sentry-cocoa/pull/5143

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 - Correctly rate limit envelopes from the new UI profiling system (#5131)
 - Race condition in ANRTrackerV1 (#5137)
--  Issue in non-sandboxed macOS apps that could inadvertently run the launch profiler (#5144)
+- Issue in non-sandboxed macOS apps that could inadvertently run the launch profiler (#5144)
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 - Correctly rate limit envelopes from the new UI profiling system (#5131)
 - Race condition in ANRTrackerV1 (#5137)
-- Cross-configuration issue in nonsandboxed macOS apps that would inadvertently run the launch profiler (#5144)
+-  Issue in non-sandboxed macOS apps that could inadvertently run the launch profiler (#5144)
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,30 @@
 # Changelog
 
+## Unreleased
+
+> ![Important]
+> This version enables the better view renderer V2 used by Session Replay by default.
+> You can disable it by setting the option `options.sessionReplay.enableViewRendererV2` to `false`.
+>
+> In case you are noticing issues with view rendering, please report them on [GitHub](https://github.com/getsentry/sentry-cocoa).
+
+### Features
+
+- Added ability to bring your own button for user feedback form display (#5107)
+
+### Fixes
+
+- Correctly rate limit envelopes from the new UI profiling system (#5131)
+- Race condition in ANRTrackerV1 (#5137)
+- Cross-configuration issue in nonsandboxed macOS apps that would inadvertently run the launch profiler (#5144)
+
+### Improvements
+
+- More logging for Session Replay video info (#5132)
+- Improve session replay frame presentation timing calculations (#5133)
+- Use wider compatible video encoding options for Session Replay (#5134)
+- GA of better session replay view renderer V2 (#5054)
+
 ## 8.49.1
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+> ![Important]
+> Version 8.21.0 introduced an issue for macOS apps that run without a sandbox (i.e. distributed outside the Mac App Store). We recommend upgrading to at least this version.
+
 ### Fixes
 
 - Non-sandboxed macOS app launch profile configuration are now respected (#5144)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,28 +2,9 @@
 
 ## Unreleased
 
-> ![Important]
-> This version enables the better view renderer V2 used by Session Replay by default.
-> You can disable it by setting the option `options.sessionReplay.enableViewRendererV2` to `false`.
->
-> In case you are noticing issues with view rendering, please report them on [GitHub](https://github.com/getsentry/sentry-cocoa).
-
-### Features
-
-- Added ability to bring your own button for user feedback form display (#5107)
-
 ### Fixes
 
-- Correctly rate limit envelopes from the new UI profiling system (#5131)
-- Race condition in ANRTrackerV1 (#5137)
-- Issue in non-sandboxed macOS apps that could inadvertently run the launch profiler (#5144)
-
-### Improvements
-
-- More logging for Session Replay video info (#5132)
-- Improve session replay frame presentation timing calculations (#5133)
-- Use wider compatible video encoding options for Session Replay (#5134)
-- GA of better session replay view renderer V2 (#5054)
+- Non-sandboxed macOS app launch profile configuration are now respected (#5144)
 
 ## 8.49.1
 
@@ -42,7 +23,7 @@
 - New continuous profiling configuration API (#4952 and #5063)
 
 > [!Important]
-> With the addition of the new profiling configuation API, the previous profiling API are deprecated and will be removed in the next major version of the SDK:
+> With the addition of the new profiling configuration API, the previous profiling API are deprecated and will be removed in the next major version of the SDK:
 >
 > - `SentryOptions.enableProfiling`
 > - `SentryOptions.isProfilingEnabled`
@@ -921,6 +902,9 @@ The following two features, disabled by default, were mistakenly added to the re
 - Add context to event with CrashIntegration disabled (#3699)
 
 ## 8.21.0
+
+> ![Important]
+> This version introduced an issue for macOS apps that run without a sandbox (i.e. distributed outside the Mac App Store). We recommend upgrading to at least version 8.49.2.
 
 ### Features
 
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -634,6 +634,11 @@ _non_thread_safe_removeFileAtPath(NSString *path)
     return [NSFileManager.defaultManager fileExistsAtPath:path isDirectory:&isDir] && isDir;
 }
 
+/**
+ * @note This method must be statically accessible because it will be called during app launch,
+ * before any instance of  ``SentryFileManager`` exists, and so wouldn't be able to access this path
+ * from an objc property on it like the other paths.
+ */
 NSString *_Nullable sentryStaticCachesPath(void)
 {
     static NSString *_Nullable sentryStaticCachesPath = nil;
@@ -650,29 +655,45 @@ NSString *_Nullable sentryStaticCachesPath(void)
         NSString *sandboxedCachesDirectory = cachesDirectory;
 
 #if TARGET_OS_OSX
-        // for macOS apps, we need to ensure our own sandbox so that this path is not shared between
-        // all apps that ship the SDK
-        NSString *bundleIdentifier = NSBundle.mainBundle.bundleIdentifier;
-        if (bundleIdentifier == nil) {
-            SENTRY_LOG_WARN(@"No bundle identifier, cannot read/write launch profile config file.");
-            sentryStaticCachesPath = nil;
-            return;
-        }
-        if (bundleIdentifier.length == 0) {
-            SENTRY_LOG_WARN(@"Bundle identifier exists but is zero length, cannot construct path "
-                            @"for caches directory.");
-            sentryStaticCachesPath = nil;
-            return;
-        }
+        // For macOS apps, we need to ensure our own sandbox so that this path is not shared between
+        // all apps that ship the SDK.
 
-        if (![cachesDirectory containsString:bundleIdentifier]) {
-            // the mac app is not sandboxed, we need to create a caches path scoped by the bundle ID
-            sandboxedCachesDirectory =
-                [cachesDirectory stringByAppendingPathComponent:bundleIdentifier];
+        // Only apps running in a sandboxed environment have the `APP_SANDBOX_CONTAINER_ID` set as a
+        // process environment variable. Reference implementation:
+        // https://github.com/realm/realm-js/blob/a03127726939f08f608edbdb2341605938f25708/packages/realm/binding/apple/platform.mm#L58-L74
+        NSProcessInfo *processInfo = [NSProcessInfo processInfo];
+        BOOL isSandboxed = processInfo.environment[@"APP_SANDBOX_CONTAINER_ID"] != nil;
+        if (!isSandboxed) {
+            // If the macOS app is not sandboxed, we need to manually create a scoped cache
+            // directory. The cache path must be unique an stable over app launches, therefore we
+            // can not use any changing identifier.
+            SENTRY_LOG_DEBUG(
+                @"App is not sandboxed, extending default cache directory with bundle identifier.");
+
+            // The bundle identifier is used to create a unique cache directory for the app.
+            // If the bundle identifier is not available, we use the name of the executable.
+            NSString *identifier = [NSBundle.mainBundle bundleIdentifier];
+            if (identifier == nil) {
+                SENTRY_LOG_WARN(@"No bundle identifier found, using main bundle executable name.");
+                identifier = [[[NSBundle mainBundle] executablePath] lastPathComponent];
+            } else if (identifier.length == 0) {
+                SENTRY_LOG_WARN(@"Bundle identifier exists but is zero length, using main bundle "
+                                @"executable name.");
+                identifier = [[[NSBundle mainBundle] executablePath] lastPathComponent];
+            }
+            // If neither the bundle identifier nor the executable name are available, we can't
+            // create a unique and stable cache directory.
+            if (identifier == nil) {
+                SENTRY_LOG_ERROR(@"No bundle identifier found, cannot create cache directory.");
+                return;
+            }
+
+            sandboxedCachesDirectory = [cachesDirectory stringByAppendingPathComponent:identifier];
         }
 #endif // TARGET_OS_OSX
 
         sentryStaticCachesPath = sandboxedCachesDirectory;
+        SENTRY_LOG_DEBUG(@"Using static cache directory: %@", sentryStaticCachesPath);
     });
     return sentryStaticCachesPath;
 }
@@ -706,49 +727,10 @@ removeBasePath(void)
 
 NSURL *_Nullable sentryLaunchConfigFileURL = nil;
 
-#    if TARGET_OS_OSX
-/**
- * Per https://github.com/getsentry/sentry-cocoa/issues/5142, we previously were writing
- * config files to the top-level cache directory for non-sandboxed apps. we need to delete
- * anything that was previously there. this isn't needed for any app shipping this version
- * of the SDK or later, but will help other apps that haven't yet gotten an update.
- */
-void
-fixBadCachesDirectory(void)
-{
-    NSString *cachesDirectory
-        = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES).firstObject;
-    if (cachesDirectory == nil) {
-        SENTRY_LOG_WARN(@"No caches directory location reported.");
-        return;
-    }
-    NSString *bundleIdentifier = NSBundle.mainBundle.bundleIdentifier;
-    if (bundleIdentifier == nil) {
-        SENTRY_LOG_ERROR(@"App does not contain an app bundle, cannot fix bad cache path.");
-        return;
-    }
-    if ([cachesDirectory containsString:bundleIdentifier]) {
-        SENTRY_LOG_DEBUG(
-            @"NSSearchPathForDirectoriesInDomains returns a sandboxed path, will not remove.");
-        return;
-    }
-    NSString *oldTopLevelSentryCache =
-        [cachesDirectory stringByAppendingPathComponent:@"io.sentry"];
-    if ([NSFileManager.defaultManager fileExistsAtPath:oldTopLevelSentryCache]) {
-        _non_thread_safe_removeFileAtPath(oldTopLevelSentryCache);
-    }
-}
-#    endif // TARGET_OS_OSX
-
 NSURL *_Nullable launchProfileConfigFileURL(void)
 {
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-
-#    if TARGET_OS_OSX
-        fixBadCachesDirectory();
-#    endif // TARGET_OS_OSX
-
         NSString *basePath = sentryStaticBasePath();
         if (basePath == nil) {
             SENTRY_LOG_WARN(@"No location available to write a launch profiling config.");
@@ -762,6 +744,7 @@ NSURL *_Nullable launchProfileConfigFileURL(void)
         }
         sentryLaunchConfigFileURL =
             [NSURL fileURLWithPath:[basePath stringByAppendingPathComponent:@"profileLaunch"]];
+        SENTRY_LOG_DEBUG(@"Launch profile config file URL: %@", sentryLaunchConfigFileURL);
     });
     return sentryLaunchConfigFileURL;
 }
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -656,15 +656,14 @@ NSString *_Nullable sentryStaticCachesPath(void)
             return;
         }
 
-        // We need to ensure our own sandbox so that this path is not shared between other apps on
-        // the same system.
-        NSString *_Nullable sandboxedCachesDirectory
-            = sentryGetScopedCachesDirectory(cachesDirectory);
-        if (!sandboxedCachesDirectory) {
+        // We need to ensure our own scoped directory so that this path is not shared between other
+        // apps on the same system.
+        NSString *_Nullable scopedCachesDirectory = sentryGetScopedCachesDirectory(cachesDirectory);
+        if (!scopedCachesDirectory) {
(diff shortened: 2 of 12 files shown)


Sources/Sentry/SentryFileManager.m
```diff
@@ -680,17 +706,62 @@ - (BOOL)isDirectory:(NSString *)path
...
+    if ([cachesDirectory containsString:bundleIdentifier]) {
+        SENTRY_LOG_DEBUG(
+            @"NSSearchPathForDirectoriesInDomains returns a sandboxed path, will not remove.");
+        return;
+    }
+    NSString *oldTopLevelSentryCache =
+        [cachesDirectory stringByAppendingPathComponent:@"io.sentry"];
+    if ([NSFileManager.defaultManager fileExistsAtPath:oldTopLevelSentryCache]) {
+        _non_thread_safe_removeFileAtPath(oldTopLevelSentryCache);
+    }
```
philprime, line 739: `h`: This will remove any data for any macOS app using Sentry. This could result in app `foo` with an older SDK version creating the directory while another app `bar` deleting it while it is in use. I propose we do not delete the path as it is already located in the caches directory, and leave the cleanup to macOS.
In addition I would propose we use a different top-level path like `io.sentry.data/<bundle-id>` or similar so we are not interfering with shipped but broken paths.

Sources/Sentry/include/SentryFileManager.h
```diff
@@ -120,7 +120,8 @@ BOOL createDirectoryIfNotExists(NSString *path, NSError **error);
 /**
  * Path for a default directory Sentry can use in the app sandbox' caches directory.
  * @note This method must be statically accessible because it will be called during app launch,
- * before any instance of @c SentryFileManager exists, and so wouldn't be able to access this path
+ * before any instance of @c SentryFileManager exists, and so wouldn't be able to access this path.
+ * @note For unsandboxed macOS apps, the path has the form @c ~/Library/Caches/<app-bundle-id> .
```
philprime, line 124: `m`: Isn't it prefixed with `io.sentry` after the fix?

```suggestion
 * @note For unsandboxed macOS apps, the path has the form @c ~/Library/Caches/io.sentry/<app-bundle-id> .
```

Tests/SentryTests/SentryOptionsTest.m
```diff
@@ -273,8 +273,13 @@ - (void)testCacheDirectoryPath
 
 - (NSString *)getDefaultCacheDirectoryPath
 {
+    // #if TARGET_OS_OSX
```
philprime, line 276: `h`: Not sure why this commented out. Please remove it if not used anymore, or add a comment why it is commented out.

philprime (commented):
We're getting there. Thanks for looking into this.

Sources/Sentry/SentryFileManager.m
```diff
@@ -680,17 +706,62 @@ - (BOOL)isDirectory:(NSString *)path
...
+/**
+ * Per https://github.com/getsentry/sentry-cocoa/issues/5142, we previously were writing
+ * config files to the top-level cache directory for non-sandboxed apps. we need to delete
+ * anything that was previously there. this isn't needed for any app shipping this version
+ * of the SDK or later, but will help other apps that haven't yet gotten an update.
+ */
+void
+fixBadCachesDirectory(void)
+{
+    NSString *cachesDirectory
```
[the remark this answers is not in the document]
philprime: Implementation candidate:

```objc
#if TARGET_OS_OSX
        // For macOS apps, we need to ensure our own sandbox so that this path is not shared between
        // all apps that ship the SDK
        NSProcessInfo *processInfo = [NSProcessInfo processInfo];
        BOOL isSandboxed = processInfo.environment[@"SENTRY_SANDBOXED"] != nil;
        if (!isSandboxed) {
            // The mac app is not sandboxed, we need to create a caches path scoped by the bundle ID
            SENTRY_LOG_DEBUG(@"App is not sandboxed, extending default cache directory with bundle identifier.");

            NSString *bundleIdentifier = NSBundle.mainBundle.bundleIdentifier;
            if (bundleIdentifier == nil) {
                SENTRY_LOG_WARN(@"No bundle identifier, cannot read/write launch profile config file.");
                sentryStaticCachesPath = nil;
                return;
            }
            if (bundleIdentifier.length == 0) {
                SENTRY_LOG_WARN(@"Bundle identifier exists but is zero length, cannot construct path "
                                @"for caches directory.");
                sentryStaticCachesPath = nil;
                return;
            }

            sandboxedCachesDirectory = [cachesDirectory stringByAppendingPathComponent:bundleIdentifier];
        }
#endif // TARGET_OS_OSX
```

Sources/Sentry/SentryFileManager.m
philipphofmann: > and leave the cleanup to macOS.

macOS might never clean this up. To my knowledge, the user must manually delete the caches.

Sources/Sentry/SentryOptions.m
```diff
@@ -209,9 +210,7 @@ - (instancetype)init
         SentryHttpStatusCodeRange *defaultHttpStatusCodeRange =
             [[SentryHttpStatusCodeRange alloc] initWithMin:500 max:599];
         self.failedRequestStatusCodes = @[ defaultHttpStatusCodeRange ];
-        self.cacheDirectoryPath
-            = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES)
-                  .firstObject;
+        self.cacheDirectoryPath = sentryStaticCachesPath();
```
philipphofmann, line 213: `h`: This can lead to lost data from SentryCrash or any envelope from previous SDK versions, as the SentryFileManager and SentryCrash both use the cacheDirectoryPath. I think @philprime already pointed that out in https://github.com/getsentry/sentry-cocoa/pull/5144/files#r2063231377. We can't simply ignore or delete the data. We must migrate that data.

Sources/Sentry/include/SentryFileManager.h
philprime: This was wrong from my end.

Sources/Sentry/SentryFileManager.m
philprime: I agree, AFAIK it only deletes caches if manually done by user or full disk.

Sources/Sentry/SentryOptions.m
armcknight: We're actually not going to mess with this path, since it actually does get some kind of namespacing using our DSN directly below this.

Sources/Sentry/SentryFileManager.m
```diff
@@ -634,6 +634,11 @@ - (BOOL)isDirectory:(NSString *)path
     return [NSFileManager.defaultManager fileExistsAtPath:path isDirectory:&isDir] && isDir;
 }
 
+/**
+ * @note This method must be statically accessible because it will be called during app launch,
+ * before any instance of  ``SentryFileManager`` exists, and so wouldn't be able to access this path
+ * from an objc property on it like the other paths.
+ */
```
philipphofmann, line 641: Thanks for the comment.

Sources/Sentry/SentryFileManager.m
```diff
@@ -646,7 +651,49 @@ - (BOOL)isDirectory:(NSString *)path
...
+
+#if TARGET_OS_OSX
+        // For macOS apps, we need to ensure our own sandbox so that this path is not shared between
+        // all apps that ship the SDK.
+
+        // Only apps running in a sandboxed environment have the `APP_SANDBOX_CONTAINER_ID` set as a
+        // process environment variable. Reference implementation:
+        // https://github.com/realm/realm-js/blob/a03127726939f08f608edbdb2341605938f25708/packages/realm/binding/apple/platform.mm#L58-L74
+        NSProcessInfo *processInfo = [NSProcessInfo processInfo];
+        BOOL isSandboxed = processInfo.environment[@"APP_SANDBOX_CONTAINER_ID"] != nil;
```
philipphofmann, line 665: `m`: We could add this to `SentryNSProcessInfoWrapper` for better testability and for consistency.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 
 - Correctly rate limit envelopes from the new UI profiling system (#5131)
 - Race condition in ANRTrackerV1 (#5137)
+- Cross-configuration issue in nonsandboxed macOS apps that would inadvertently run the launch profiler (#5144)
```
philipphofmann, line 19: `l`: A bit cleaner, IMO
```suggestion
-  Issue in non-sandboxed macOS apps that could inadvertently run the launch profiler (#5144)
```

Sources/Sentry/SentryOptions.m
philipphofmann: I don't understand why we're not going to mess with this path? The cacheDirectoryPath is used by SentryCrash and the SentryFileManager. This would change where we store that data for non-sandboxed macOS apps, and it is a problem. Therefore, I would like to remove that change from this PR.

Sources/Sentry/SentryFileManager.m
philprime: You are welcome. I actually got this from [this PR](https://github.com/getsentry/sentry-cocoa/commit/f8833c443d047f0c8dac5710e52b4c0624fc52ce#diff-98c5d4f2fdf1f254efbb66e82cbc8109f224038b61270c4886cce083c12664e3R732-R753), but not sure when it got lost

Sources/Sentry/SentryFileManager.m
philipphofmann: Ah sorry, it's in `load`. Just skip it.

philipphofmann: Maybe add a comment why we don't do that and all good.

Sources/Sentry/SentryFileManager.m
```diff
@@ -646,7 +651,49 @@ - (BOOL)isDirectory:(NSString *)path
...
+            }
+            // If neither the bundle identifier nor the executable name are available, we can't
+            // create a unique and stable cache directory.
+            if (identifier == nil) {
+                SENTRY_LOG_ERROR(@"No bundle identifier found, cannot create cache directory.");
+                return;
+            }
+
+            sandboxedCachesDirectory = [cachesDirectory stringByAppendingPathComponent:identifier];
+        }
```
philipphofmann, line 692: `m`: We could write unit tests for the logic in here by extracting a method that takes `isSandboxed`, `identifier`, and `lastPathComponent` and returns the `sandboxedCachesDirectory`. WDTY @philprime

Sources/Sentry/SentryFileManager.m
philprime: added comment ✅

Sources/Sentry/SentryOptions.m
philprime: This is resolved now that I removed the changes in `SentryOptions.m`, right? We have #5145 now to follow up

Sources/Sentry/SentryFileManager.m
armcknight: i think extracting the logic that's inside the dispatch_once makes sense to unit test it, and then the UI tests should validate correct integrated behavior

CHANGELOG.md
armcknight: Sounds good to me 👍🏻

Sources/Sentry/SentryFileManager.m
philipphofmann: The advantage is that you can easily test all the edge cases of the logic here. The method must not use dispatch_once. Basically something like this

```
NSString *_Nullable sentryGetSandboxedCachesDirectory(
    NSString *cachesDirectory,
    NSProcessInfo *processInfo,
    BOOL isSandboxed,
    NSString *_Nullable identifier,
    NSString *_Nullable lastPathComponent)
{
    NSString *sandboxedCachesDirectory = cachesDirectory;
    
    if (!isSandboxed) {
        // If the macOS app is not sandboxed, we need to manually create a scoped cache
        // directory. The cache path must be unique an stable over app launches, therefore we
        // can not use any changing identifier.
        SENTRY_LOG_DEBUG(
            @"App is not sandboxed, extending default cache directory with bundle identifier.");

        if (identifier == nil) {
            SENTRY_LOG_WARN(@"No bundle identifier found, using main bundle executable name.");
            identifier = lastPathComponent;
        } else if (identifier.length == 0) {
            SENTRY_LOG_WARN(@"Bundle identifier exists but is zero length, using main bundle "
                            @"executable name.");
            identifier = lastPathComponent;
        }
        // If neither the bundle identifier nor the executable name are available, we can't
        // create a unique and stable cache directory.
        if (identifier == nil) {
            SENTRY_LOG_ERROR(@"No bundle identifier found, cannot create cache directory.");
            return nil;
        }

        sandboxedCachesDirectory = [cachesDirectory stringByAppendingPathComponent:identifier];
    }
    
    return sandboxedCachesDirectory;
}
```

Call it like this
```
        // We can not use the SentryNSProcessInfoWrapper here because this method is called before
        // the SentryDependencyContainer is initialized.
        NSProcessInfo *processInfo = [NSProcessInfo processInfo];
        BOOL isSandboxed = processInfo.environment[@"APP_SANDBOX_CONTAINER_ID"] != nil;
        
        // The bundle identifier is used to create a unique cache directory for the app.
        // If the bundle identifier is not available, we use the name of the executable.
        NSString *identifier = [NSBundle.mainBundle bundleIdentifier];
        NSString *lastPathComponent = [[[NSBundle mainBundle] executablePath] lastPathComponent];
        
        sandboxedCachesDirectory = sentryGetSandboxedCachesDirectory(
            cachesDirectory, processInfo, isSandboxed, identifier, lastPathComponent);
```

philipphofmann: Then you use `sentryGetSandboxedCachesDirectory` for your tests.

armcknight (commented):
@philprime I think based on my question, we will need an additional test case, where we call `sentryBuildScopedCachesDirectoryPath` twice, the first time with a nil bundle id and a nonnil lastPathComponent, then the second time with a nonnil bundle path and the same lastPathComponent, and assert that we still use the lastPathComponent version of the static base path.

That also leads to other combinations but unsure what is actually possible and how defensive we want to get with this logic.

Sources/Sentry/SentryFileManager.m
```diff
@@ -634,23 +634,111 @@ - (BOOL)isDirectory:(NSString *)path
...
+    }
+
+    // If neither the bundle identifier nor the executable name are available, we can't
+    // create a unique and stable cache directory.
+    // We do not fall back to any default path, because it could be shared with other apps
+    // and cause leaks impacting other apps.
+    if (identifier == nil) {
+        SENTRY_LOG_ERROR(@"No bundle identifier found, cannot create cache directory.");
+        return nil;
+    }
```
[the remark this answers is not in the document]
armcknight: added this in https://github.com/getsentry/sentry-cocoa/pull/5144/commits/0398fdaba6613425f5ba8b5e77e80f6e3d8e1594

CHANGELOG.md
```diff
@@ -897,6 +903,9 @@ The following two features, disabled by default, were mistakenly added to the re
 
 ## 8.21.0
 
+> ![Important]
+> This version introduced an issue for macOS apps that run without a sandbox (i.e. distributed outside the Mac App Store). We recommend upgrading to at least version 8.49.2.
```
philipphofmann, line 907: We should also update the GitHub release after merging and releasing this https://github.com/getsentry/sentry-cocoa/releases/tag/8.21.0.

Sources/Sentry/SentryFileManager.m
```diff
@@ -634,23 +634,132 @@ - (BOOL)isDirectory:(NSString *)path
...
         if (cachesDirectory == nil) {
             SENTRY_LOG_WARN(@"No caches directory location reported.");
             return;
         }
-        sentryStaticCachesPath = cachesDirectory;
+
+        // We need to ensure our own sandbox so that this path is not shared between other apps on
+        // the same system.
+        NSString *_Nullable sandboxedCachesDirectory
+            = sentryGetScopedCachesDirectory(cachesDirectory);
```
philipphofmann, line 662: `m`: This can be a bit confusing. On a non-sandboxed macOS app this directory isn't sandboxed. If you're not familiar with the code, you could assume we create our own sandbox, which isn't true. Instead, we use the public caches folder and scope the directory per app, so it isn't shared across apps. We explain this in detail below, but if you only read this section of the code, you could assume we create our own sandbox.
```suggestion
        // We need to ensure our own scoped directory so that this path is not shared between other apps on
        // the same system.
        NSString *_Nullable scopedCachesDirectory
            = sentryGetScopedCachesDirectory(cachesDirectory);
```

Sources/Sentry/SentryFileManager.m
```diff
@@ -634,23 +634,132 @@ - (BOOL)isDirectory:(NSString *)path
...
+    }
+
+    // It's unlikely that the executable name will be zero length, but we'll cover this case anyways
+    if (identifier.length == 0) {
+        SENTRY_LOG_ERROR(@"Executable name was zero length.");
+        return nil;
+    }
+
+    return [cachesDirectory stringByAppendingPathComponent:identifier];
+}
```
philipphofmann, line 761: `l`: This code has little to do with the rest of the FileManager. We could put it into an extra file to not bloat the FileManager even more.

Tests/SentryTests/Helper/SentryFileManagerTests.swift
```diff
@@ -1209,6 +1209,121 @@ extension SentryFileManagerTests {
...
+        let result = sentryGetScopedCachesDirectory(cachesDirectoryPath)
+
+        // -- Assert
+        XCTAssertEqual(result, cachesDirectoryPath)
+#endif // os(macOS)
+    }
+
+    func testSentryGetScopedCachesDirectory_targetIsMacOS_shouldReturnPath() throws {
+#if !os(macOS)
+        throw XCTSkip("Test is disabled for iOS")
```
philipphofmann, line 1230: `l`: It's disabled for everything but macOS.
```suggestion
        throw XCTSkip("Test is disabled for non macOS")
```

Tests/SentryTests/Helper/SentryFileManagerTests.swift
```diff
@@ -1209,6 +1209,121 @@ extension SentryFileManagerTests {
...
+            // for sandboxed scenarios, always return the original path
+            (isSandboxed: true, bundleIdentifier: "com.example.app", lastPathComponent: "AppBinaryName", expected: "some/path/to/caches"),
+            (isSandboxed: true, bundleIdentifier: "", lastPathComponent: "AppBinaryName", expected: "some/path/to/caches"),
+            (isSandboxed: true, bundleIdentifier: nil, lastPathComponent: "AppBinaryName", expected: "some/path/to/caches"),
+            (isSandboxed: true, bundleIdentifier: "com.example.app", lastPathComponent: "", expected: "some/path/to/caches"),
+            (isSandboxed: true, bundleIdentifier: "", lastPathComponent: "", expected: "some/path/to/caches"),
+            (isSandboxed: true, bundleIdentifier: nil, lastPathComponent: "", expected: "some/path/to/caches"),
+            (isSandboxed: true, bundleIdentifier: "com.example.app", lastPathComponent: nil, expected: "some/path/to/caches"),
+            (isSandboxed: true, bundleIdentifier: "", lastPathComponent: nil, expected: "some/path/to/caches"),
+            (isSandboxed: true, bundleIdentifier: nil, lastPathComponent: nil, expected: "some/path/to/caches")
```
philipphofmann, line 1292: `l`: I'm confused. What's the benefit of testing all combinations? That test is tough to read and maintain. Testing all combinations instead of the most important ensures all combinations work, but it comes with the cost of lots of code and a couple of redundant tests.

Tests/SentryTests/Helper/SentryFileManagerTests.swift
```diff
@@ -1209,6 +1209,121 @@ extension SentryFileManagerTests {
...
+            false,
+            bundleId,
+            executableName
+        )
+
+        // -- Assert --
+        XCTAssertEqual(result, "~/some/path/to/caches/MyAppBinary")
+
+        // -- Cleanup --
+        try FileManager.default.removeItem(atPath: "~/some/path/to/caches/MyAppBinary")
```
philipphofmann, line 1325: `m`: Better to move this directly below `ry FileManager.default.createDirectory(atPath: "~/some/path/to/caches/MyAppBinary", withIntermediateDirectories: true)` and wrap it in a `defer` to ensure it always runs.

Sources/Sentry/SentryFileManager.m
```diff
@@ -634,23 +634,132 @@ - (BOOL)isDirectory:(NSString *)path
...
+        // is in the app identifier now.
+        NSString *executableNameBasedPath =
+            [cachesDirectory stringByAppendingPathComponent:lastPathComponent];
+        BOOL isDirectory = NO;
+        if ([NSFileManager.defaultManager fileExistsAtPath:executableNameBasedPath
+                                               isDirectory:&isDirectory]
+            && isDirectory) {
+            return executableNameBasedPath;
+        }
+    }
```
philipphofmann, line 733: `h`: What drove us to choose this approach? This will cause one extra `fileExistsAtPath` during app launch for every non-sandbox macOS app, no matter if they use profiling or not. It would be great if we could avoid this.
What about always using the bundle identifier and only using the lastPathComponent as a fallback? If the bundle identifier changes, you would potentially lose one app launch profile and the config file would stay at the previous path.

philipphofmann (commented):
I maybe found one issue to address, but close to LGTM.

Sources/Sentry/SentryFileManager.m
philprime: Yes, but in another PR please.

CHANGELOG.md
philprime: We should also consider adding a note to the hotfix release changes.
[Internal discussion](https://sentry.slack.com/archives/C08QHFV4W9E/p1745915959876679)

Sources/Sentry/SentryFileManager.m
philprime: I agree, left-over from refactoring

Sources/Sentry/SentryFileManager.m
philprime: > What drove us to choose this approach? 

IUUC it's consistency between app starts, so that if an app is fixed and suddenly has a bundle identifier available (unrealistic unless it's a CLI without an Info.plist suddenly getting one) it won't loose the previous approach. 

> This will cause one extra fileExistsAtPath during app launch for every non-sandbox macOS app, no matter if they use profiling or not. It would be great if we could avoid this.

If performance is your issue here, we could also use `access(path. F_OK) == 0` here, as we already do in `appLaunchProfileConfigFileExists`.

> What about always using the bundle identifier and only using the lastPathComponent as a fallback?

That's what we already do in the next couple of lines.

> If the bundle identifier changes, you would potentially lose one app launch profile and the config file would stay at the previous path.

I would consider the bundle identifier as the identity of the app, therefore a change of a bundle identifier is not the same app anyways. So yes I agree we just skip the app launch.

kahest: IMO both approaches are fine.
- an additional `fileExistsAtPath` is bad on iOS, but for non-sandboxed macOS apps it seems fine - or am I missing something?
- OTOH, losing one app launch profile in the edge-case that bundle id changes (because suddenly there's a info.plist) is also fine. again, am I missing something?

CHANGELOG.md
philprime: Added a note to the unreleased section.

Tests/SentryTests/Helper/SentryFileManagerTests.swift
philprime: I also don't see the value in testing every combination, as there are no code paths leading to these combinations.

Tests/SentryTests/Helper/SentryFileManagerTests.swift
```diff
@@ -1209,6 +1209,121 @@ extension SentryFileManagerTests {
...
+            // -- Assert --
+            XCTAssertEqual(result, testCase.expected, "Inputs: (isSandboxed: \(testCase.isSandboxed), bundleIdentifier: \(String(describing: testCase.bundleIdentifier)), lastPathComponent: \(String(describing: testCase.lastPathComponent)), expected: \(String(describing: testCase.expected))); Output: \(String(describing: result))")
+        }
+    }
+
+    func test_sentryBuildScopedCachesDirectoryPath_afterAlreadyUsingExecutableName() throws {
+        // -- Arrange --
+        let executableName = "MyAppBinary"
+        let bundleId = "com.my.app"
+        try FileManager.default.createDirectory(atPath: "~/some/path/to/caches/MyAppBinary", withIntermediateDirectories: true)
```
philprime, line 1311: `h`: Let's not pollute the user home directory with test files. If we actually create files, let's go for `/tmp`

philprime:
Summary from internal Slack discussion on keep using the executable path name even if the bundle identifier becomes available:

We think maintenance cost outweighs the benefit.

- AFAIK it's a rare edge case because every macOS app needs to have a bundle identifier, therefore it's not possible for a macOS app to not have a bundle identifier defined and then suddenly have one.
- CLI tools do not use bundle identifiers, but still the edge case would only happen if a developer suddenly adds a bundle to the CLI and then marks it to be the main bundle.
- The file-exists will be run at every app start of every macOS app without a sandbox, unnecessarily slow down app launch (even if only minimal impact)
- In the worst-case when an app changes from having no bundle identifier to having one, it would change the path and skip one app launch. Then it writes the `profileLaunch` again, so it is fixed with the next app start
- Having additional logic to maintain for a rare edge case just increases complexity of maintenance for little benefit. As we want to revisit the file hierarchy in #5145 this could also further increase complexity for any future changes.

To conclude, it's not worth the effort and after careful reconsidering I am now going to revert this handling mechanism.

Sources/Sentry/SentryFileManager.m
philprime: Discussed in calls and internal Slack channels. Posted an update directly to the PR.

Tests/SentryTests/Helper/SentryFileManagerTests.swift
philprime: While I would also prefer to use equivalent classes instead of testing every permutation of the inputs, let's keep it as is for now and get this PR done.

philprime (approved):

PR #5147, opened 2025-04-29
ci: when commenting about dangerous files, remove previous comments

armcknight:
Remove previous commnents to avoid this noisy iteration:
image

#skip-changelog


philipphofmann (approved):
Thanks 🥇

PR #5150, opened 2025-04-29
ci: Use non xl macos-15 for benchmarks

philipphofmann:
Switch to a non xl runner for benchmarking cause it's not required and xl runners are expensive.

#skip-changelog


philprime (approved):
LGTM

PR #5151, opened 2025-04-29
ci: Build XCFramework with Xcode 15.2 on macos-14

philipphofmann:
Replace the xlarge runner with macos-15 and compile the XCFramework on Xcode 16.3.

If this works fine, we can also consider replacing https://github.com/getsentry/sentry-cocoa/blob/500abd8a216eb7e05a4a3383039c79a84998d71b/.github/workflows/release.yml#L19 but I think we should do that in a minor release and not a bugfix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,6 +94,11 @@ jobs:
     runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


philipphofmann:
That's the plan: https://github.com/getsentry/sentry-cocoa/pull/5152

philprime (approved):
Thanks for the comments, LGTM

Issue #5124, opened 2025-04-22
feedback: ensure functionality is disabled on call to SentrySDK.close


armcknight:
### Description

at a minimum, remove any UI elements from the view hierarchy

PR #5158, opened 2025-04-30
chore: add toggle to enable/disable Sentry in sample on-demand

### LINKED ISSUE #5069: Session Replay is not cleaned up correctly after calling Sentry.close

philprime:
Adds a toggle switch to the camera sample app to disable Sentry SDK after it started.

I use this to see how well the SDK is cleaned up to fix #5069
Can also be used for #5124 

This is a side-product so it's a different PR for easy review, extracted from draft PRs

#skip-changelog

### DIFF
--- a/Samples/SessionReplay-CameraTest/Sources/AppDelegate.swift
+++ b/Samples/SessionReplay-CameraTest/Sources/AppDelegate.swift
@@ -4,6 +4,7 @@ import UIKit
 @main
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
+    static var isSentryEnabled = true
     static var isSessionReplayEnabled = true
     static var isViewRendererV2Enabled = true
 
@@ -17,6 +18,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if SentrySDK.isEnabled {
             print("SentrySDK already started, closing it")
             SentrySDK.close()
+        }
+
+        if !isSentryEnabled {
+            print("SentrySDK disabled")
+            return
         }
 
         SentrySDK.start { options in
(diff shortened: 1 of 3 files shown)


philipphofmann (approved):
LGTM

PR #5173, opened 2025-05-05
chore(session-replay): Add detailed debug logging for session replay

philprime:
## :scroll: Description

Adds debug logging to session replay integration.

It must become easier to debug issues with the integration.
I tried to add only relevant information for SDK users trying not to produce spam-like amount of logs.

Ran sample app in iOS simulator, looked at logs.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentrySessionReplayIntegration.m
+++ b/Sources/Sentry/SentrySessionReplayIntegration.m
@@ -198,7 +198,7 @@ static SentryTouchTracker *_touchTracker;
             @"[Session Replay] Previous session replay is a buffer, using error sample rate");
         float errorSampleRate = [jsonObject[@"errorSampleRate"] floatValue];
         if ([SentryDependencyContainer.sharedInstance.random nextNumber] >= errorSampleRate) {
-            SENTRY_LOG_DEBUG(
+            SENTRY_LOG_INFO(
                 @"[Session Replay] Buffer session replay event not sampled, dropping replay");
             return;
         }
--- a/Sources/Sentry/SentrySessionReplayIntegration.m
+++ b/Sources/Sentry/SentrySessionReplayIntegration.m
@@ -102,9 +102,12 @@ static SentryTouchTracker *_touchTracker;
 
     id<SentryViewRenderer> viewRenderer;
     if (enableViewRendererV2) {
+        SENTRY_LOG_DEBUG(@"[Session Replay] Setting up view renderer v2, fast view rendering: %@",
+            enableFastViewRendering ? @"YES" : @"NO");
         viewRenderer =
             [[SentryViewRendererV2 alloc] initWithEnableFastViewRendering:enableFastViewRendering];
     } else {
+        SENTRY_LOG_DEBUG(@"[Session Replay] Setting up default view renderer");
         viewRenderer = [[SentryDefaultViewRenderer alloc] init];
     }
     // We are using the flag for the experimental view renderer also for the experimental mask
@@ -115,6 +118,8 @@ static SentryTouchTracker *_touchTracker;
                                                     enableMaskRendererV2:enableViewRendererV2];
 
     if (touchTracker) {
+        SENTRY_LOG_DEBUG(
+            @"[Session Replay] Setting up touch tracker, scale: %f", replayOptions.sizeScale);
         _touchTracker = [[SentryTouchTracker alloc] initWithDateProvider:_dateProvider
                                                                    scale:replayOptions.sizeScale];
         [self swizzleApplicationTouch];
@@ -148,6 +153,7 @@ static SentryTouchTracker *_touchTracker;
     NSData *lastReplay = [NSData dataWithContentsOfURL:lastReplayUrl];
 
     if (lastReplay == nil) {
+        SENTRY_LOG_DEBUG(@"[Session Replay] No last replay info found");
         return nil;
     }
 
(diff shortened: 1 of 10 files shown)


Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -185,8 +194,12 @@ - (void)resumePreviousSessionReplay:(SentryEvent *)event
     int segmentId = hasCrashInfo ? crashInfo.segmentId + 1 : 0;
 
     if (type == SentryReplayTypeBuffer) {
+        SENTRY_LOG_DEBUG(
+            @"[Session Replay] Previous session replay is a buffer, using error sample rate");
         float errorSampleRate = [jsonObject[@"errorSampleRate"] floatValue];
         if ([SentryDependencyContainer.sharedInstance.random nextNumber] >= errorSampleRate) {
+            SENTRY_LOG_DEBUG(
```
philipphofmann, line 201: `l`: Could be level info

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -102,9 +102,12 @@ - (void)setupWith:(SentryReplayOptions *)replayOptions
 
     id<SentryViewRenderer> viewRenderer;
     if (enableViewRendererV2) {
+        SENTRY_LOG_DEBUG(@"[Session Replay] Setting up view renderer v2, fast view rendering: %@",
```
philipphofmann, line 105: `m`: Maybe we should define a constant for `[Session Replay]` or maybe even extra macros to ensure we're using the `[Session Replay] ` for filtering. It doesn't have to be in this PR. My main concern is that it's very easy to not follow the patten of using the `[Session Replay]` prefix, and I would like to find a way to ensure this pattern. We could also have a linter checking that log messages start with this pattern if the class is a SR class.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -212,6 +238,7 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
     }
         
     private func finishVideo(outputFileURL: URL, usedFrames: [SentryReplayFrame], videoHeight: Int, videoWidth: Int, videoWriter: AVAssetWriter) -> Result<SentryVideoInfo?, Error> {
+        SentryLog.debug("[Session Replay] Finishing video with output file URL: \(outputFileURL), used frames count: \(usedFrames.count), video height: \(videoHeight), video width: \(videoWidth)")
```
philipphofmann, line 241: `l`: That could be level info.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -125,22 +134,34 @@ class SentrySessionReplay: NSObject {
             return
         }
         
-        guard !reachedMaximumDuration else { return }
-        guard !isRunning else { return }
+        guard !reachedMaximumDuration else { 
+            SentryLog.debug("[Session Replay] Reached maximum duration, not resuming")
```
philipphofmann, line 138: `l`: Maybe level `info` or `warn`.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -125,22 +134,34 @@ class SentrySessionReplay: NSObject {
...
     func captureReplayFor(event: Event) {
-        guard isRunning else { return }
+        SentryLog.debug("[Session Replay] Capturing replay for event: \(event)")
+        guard isRunning else { 
+            SentryLog.debug("[Session Replay] Session replay is not running, not capturing replay")
+            return 
+        }
 
         if isFullSession {
+            SentryLog.debug("[Session Replay] Session replay is full, setting event context")
```
philipphofmann, line 158: `l`: Again maybe level `info`.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -305,29 +354,37 @@ class SentrySessionReplay: NSObject {
     }
     
     private func takeScreenshot() {
-        guard let rootView = rootView, !processingScreenshot else { return }
+        SentryLog.debug("[Session Replay] Taking screenshot")
```
philipphofmann, line 357: `m`: This could be below the guard statements, otherwise you get two log message when not taking a screenshot. Basically 
```
[Session Replay] Taking screenshot
[Session Replay] Not taking screenshot, reason: root view is nil or processing screenshot
```
Which is a bit confusing and verbose.

Or you change the log message to entering taking screenshot or something like that.

philipphofmann (approved):
LGTM, with a couple of suggestions.

Sources/Sentry/SentrySessionReplayIntegration.m
philprime: I would propose we go one step further and introduce "categories" for logs, similar to OSLog [1]. Maybe we can do this easily be introducing a parameter for the macro instead of using a  string-based approach. I know this will introduce other issues down the road, so yes different PR.

My intention with the pattern was not introducing a standardized approach, I just needed a way to filter logs for session replay.

[1] https://www.avanderlee.com/debugging/oslog-unified-logging/

philipphofmann: Let's solve this in another PR.

philipphofmann (approved):
Again, LGTM. Thanks for including the feedback.

PR #5157, opened 2025-04-30
fix(ci): fix changes-in-high-risk-code.yml with no previous comments

philprime:
Failed here: https://github.com/getsentry/sentry-cocoa/actions/runs/14752051972/job/41411624802?pr=5154

#skip-changelog

PR #5183, opened 2025-05-06
fix(ci): Use response data in changes-in-high-risk-code.yml

philprime:
- Adds the `changes-in-high-risk-code.yml` to the list of risky files to make sure the action is executed when changing the workflow too
- Uses the `comments.data` instead of `comments`. See [this workflow run](https://github.com/getsentry/sentry-cocoa/actions/runs/14859106773/job/41719320930) for details on the actual value of `comments` and the error.

The bug was introduced in #5147 and we already tried to fix it in #5157.

#skip-changelog

### DIFF
--- a/.github/file-filters.yml
+++ b/.github/file-filters.yml
@@ -1,6 +1,7 @@
 # This is used by the action https://github.com/dorny/paths-filter
 
 high_risk_code: &high_risk_code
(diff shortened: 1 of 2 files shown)


philipphofmann (approved):
LGTM

PR #5197, opened 2025-05-07
test: clean up xcconfigs

armcknight:
Two clean up items:
- somehow the configs crept back into SessionReplay-CameraTest after they were moved from there to Samples/Shared/Config
- after i moved the version bump pr to the end of the chain, some of the changes crept back in. the `#include` of the versioning configs are fixed now

Nothing was broken but these were just duplicates and premature changes I'd like to keep straight for subsequent work.

#skip-changelog


philprime (approved):
Thanks for the cleanup

philipphofmann (approved):
Oh yes, thanks for the cleanup. 💯

PR #5181, opened 2025-05-06
test: shared sdk lib

armcknight:
Take 2 of sharing SentrySDKWrapper. Finally getting back to it after moving the sample app projects to xcodegen to solve the issues hit in https://github.com/getsentry/sentry-cocoa/pull/5077 and https://github.com/getsentry/sentry-cocoa/pull/5085

#skip-changelog

PR #5198, opened 2025-05-07
test: share sdk wrapper xcodegen

armcknight:
After linking the shared library to the sample apps in #5181, use the shared SentrySDKWrapper.shared.startSentry() call from all swift apps. leaves iOS-ObjectiveC as-is for now as it's the only one using ObjC to initialize the SDK and we need to ensure all those APIs work in ObjC.

Note: there were some differences between the way the SDK wrapper initializes the SDK by default without overrides, and the way the Camera sample would do it. That is addressed in #5201 and #5202

#skip-changelog


philipphofmann (approved):
LGTM

PR #5199, opened 2025-05-07
test: share sdk overrides

armcknight:
Separately from using the shared SentrySDKWrapper to initialize the SDK in the sample apps in #5198, supply the launch args and environment vars to each sample app's scheme so SentrySDKOverrides can be used during debug. These were previously only available to iOS-Swift. Uses XcodeGen spec `include` and `schemeTemplates` directives to avoid duplicating the list for each one.

Also moves the FeaturesViewController and cell classes to the shared lib so it can be presented in any sample app running on iOS, although the storyboard/code changes haven't yet been done for them yet (mac/tv/watch will have to be recoded separately for various reasons).

#skip-changelog


philipphofmann (approved):
LGTM

PR #5201, opened 2025-05-07
test: more override config

armcknight:
Adding more config option overrides and reorganizing in the yml spec. This originally started when I was moving the Camera sample app to use the sdk wrapper and shared overrides.

# TODO
- [x] Ensure that any option set in SentrySDKWrapper.swift has a corresponding override in SentrySDKOverrides
- Gonna leave this for a subsequent PR to keep this one from getting too big: ~- [ ]  Do it again, but for iOS-ObjectiveC, since we need to make sure usage of options from objc compiles, but we don't yet have other objc sample apps, so don't need an objc wrapper just yet~

#skip-changelog


Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -66,7 +66,7 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
         /**
          * Used by Hybrid SDKs.
          */
-        static func fromName(_ name: String) -> SentryReplayOptions.SentryReplayQuality {
+        public static func fromName(_ name: String) -> SentryReplayOptions.SentryReplayQuality {
```
armcknight, line 69: How bad would this SDK change be @philipphofmann

armcknight:
@philprime It's getting long enough that I can't quite remember why these calls are already in this branch, but I'm pretty sure the reason it's not showing up for SwiftUI is due to this change not having been merged yet: https://github.com/getsentry/sentry-cocoa/pull/5223

philprime:
@armcknight so what would be the best approach to continue? Merging the PR with the known issue and fixing it then?

armcknight:
@philprime i'm just going to remove user feedback from the SwiftUI test apps until we find a solution that works.

Samples/SentrySampleShared/SentrySampleShared/SentrySDKWrapper.swift
```diff
@@ -41,30 +41,40 @@ public struct SentrySDKWrapper {
...
 #if !os(macOS) && !os(watchOS) && !os(visionOS)
-        if #available(iOS 16.0, *), !SentrySDKOverrides.Other.disableSessionReplay.boolValue {
+        if #available(iOS 16.0, *), !SentrySDKOverrides.SessionReplay.disableSessionReplay.boolValue {
             options.sessionReplay = SentryReplayOptions(
-                sessionSampleRate: 0,
-                onErrorSampleRate: 1,
-                maskAllText: true,
-                maskAllImages: true
+                sessionSampleRate: SentrySDKOverrides.SessionReplay.sessionReplaySampleRate.floatValue ?? 0,
+                onErrorSampleRate: SentrySDKOverrides.SessionReplay.sessionReplayOnErrorSampleRate.floatValue ?? 1,
```
philprime: `l`: Duplicate session replay scoping

```suggestion
                sessionSampleRate: SentrySDKOverrides.SessionReplay.sessionSampleRate.floatValue ?? 0,
                onErrorSampleRate: SentrySDKOverrides.SessionReplay.onErrorSampleRate.floatValue ?? 1,
```

armcknight:
> I expect to review it in https://github.com/getsentry/sentry-cocoa/pull/5202 again.

This is not meant to merge into #5202... anyone reviewing this should review it as its own changeset, don't leave anything for later.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
armcknight: `@_spi(Private)` can't be used here:
```
Internal static method cannot be declared '@_spi' because only public and open declarations can be '@_spi'
```

I also tried making it `open` but:
```
Only classes and overridable class members can be declared 'open'; use 'public'
```

armcknight: Oh. I guess this PR doesn't even use this function any more ¯\\_(ツ)_/¯ Just removing the change.

PR #5195, opened 2025-05-07
ci: Remove CameraTest xcconfig from version bump

philipphofmann:
## :scroll: Description

The CameraTest doesn't contain a proper marketing version. We can ignore it. 

I also added a GH action to run the version util via the make file.

## :green_heart: How did you test it?

By intentionally breaking it here https://github.com/getsentry/sentry-cocoa/actions/runs/14886048089/job/41805819992?pr=5195.

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #5204, opened 2025-05-08
ci: Add separate prettier linting workflow

philipphofmann:
The linter didn't run for this PR https://github.com/getsentry/sentry-cocoa/pull/5195 which contained lint errors for the yml file. That caused the precommit hook to fail locally because it runs prettier, and I had to fix this with https://github.com/getsentry/sentry-cocoa/pull/5203. Therefore, I opened this PR to avoid future issues like these.

This commit introduces a new GitHub Actions workflow for running Prettier on specific file types (YAML, JSON, Markdown) during pushes and pull requests to the main branch. The Prettier job has been removed from the existing lint workflow to optimize performance by isolating it from other linters.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/5203 which contains prettier lint fixes. We see that the CI job fails when not containing these fixes in https://github.com/getsentry/sentry-cocoa/actions/runs/14900491723/job/41851170858. Now it's green.


philprime (approved):
Thank you for the improvement 👍

PR #5205, opened 2025-05-08
ci: Move shellcheck into its on workflow

philipphofmann:
Move the shellcheck into its on workflow so it always runs when sh files change. We can also run it on ubuntu and its preinstalled.

#skip-changelog


philprime:
Can you please give more context why it should be run in its own workflow rather than changing the paths of the existing workflow?

philipphofmann:
Sorry my bad. Is it clearer now with adding the comment?

philprime (approved):
Now it's clearer, thank you, LGTM!

PR #5170, opened 2025-05-03
ci: macos swift xcodegen

armcknight:
for https://github.com/getsentry/sentry-cocoa/issues/5165

#skip-changelog

PR #5208, opened 2025-05-08
fix: Add missing App Sandbox entitlements for macOS

philprime:
## :scroll: Description

- Changes the XcodeGen settings to generate an Entitlements file with App Sandbox enabled for `macOS-Swift-Sandboxed` and `macOS-Swift-Sandboxed-Other`
- Adds a second tab `Metadata` to the macOS UI which can be used to display and verify generic checks 
- Add a verification of the sandbox status to the UI tests

We are testing the behaviour of macOS apps with and without App Sandbox. During the refactoring of the sample apps to macOS, the entitlements where lost. This PR recovers this and ensure it won't happen again.

Added and removed the App Sandbox from the targets and watched the newly added tests fail.

## :pencil: Checklist

#skip-changelog


philipphofmann (approved):
LGTM

armcknight (commented):
A couple questions
- was something broken that this fixes? In the original [PR](https://github.com/getsentry/sentry-cocoa/pull/5143) that introduced the mac app variants, i actually removed the entitlements file altogether and only set ENABLE_APP_SANDBOX = YES in the project's build settings and confirmed that was able to reproduce the issue, which I moved to the xcconfig files in the xcodegen refactor ([PR](https://github.com/getsentry/sentry-cocoa/pull/5170)) so nothing was accidentally lost, you can see no entitlements file appears in that diff at all.
- why does the entitlements file for macOS-Swift also have `com.apple.security.app-sandbox` set to `<true />`? https://github.com/getsentry/sentry-cocoa/blob/eb43e8b9da5f52d12b5ffa25dae2625b2bcd547b/Samples/macOS-Swift/macOS-Swift.entitlements#L5-L6
    - actually, looking again, i think macOS-Swift.entitlements is simply unused, removing in [this PR](https://github.com/getsentry/sentry-cocoa/pull/5212)
- why do we provide the entitlement files with the settings, but then also specify the setting again in the xcodegen spec? https://github.com/getsentry/sentry-cocoa/blob/eb43e8b9da5f52d12b5ffa25dae2625b2bcd547b/Samples/macOS-Swift/macOS-Swift.yml#L100-L103
    - ETA: Ah I see what happened here, you checked in the generated files. Please don't commit multiple sources of truth to the repository as it can cause confusion and drift, as I mentioned in https://github.com/getsentry/sentry-cocoa/pull/5106#discussion_r2055073160

philprime:
Thanks for the feedback, I'll answer your questions in #5213

PR #5210, opened 2025-05-08
fix: Fix default values and dictionary values for session replay options

philprime:
# :scroll Description

This commit introduces several enhancements to the SentryReplayOptions class, including new properties for better configuration and handling of session replay options. Key changes include:

- Added internal methods for converting raw values to SentryReplayQuality.
- Introduced new properties for bitrate and size scale calculations.
- Updated initializers to support dictionary-based configuration.
- Added comprehensive unit tests for SentryReplayOptions and related classes, ensuring default values and custom configurations are correctly handled.
- Introduced new test files for Objective-C compatibility and SwiftUI integration.

Covered cases:
- Init without args in Swift
- Init with all args in Swift
- Init with some args ommitted in Swift (fallback to default value)
- No nullable args
- Init without args in Objective-C
- Init with all args in Objective-C
- Init with missing value in dictionary, fallback to default value

- The default values are inconsistent (i.e. `enableViewRendererV2` was not enabled by default everywhere)
- The default values set in the Swift initializer made an initializer without arguments unavailable to Objective-C, and the replacement fix (empty init) was setting default values inconsistently.

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
+++ b/Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
@@ -9,7 +9,7 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
      */
     @objc
     public enum SentryReplayQuality: Int, CustomStringConvertible {
-        fileprivate static let defaultQuality: SentryReplayQuality = .medium
+        internal static let defaultQuality: SentryReplayQuality = .medium
 
         /**
          * Video Scale: 80%
@@ -48,6 +48,28 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
             default: return defaultQuality
             }
         }
+
+        /// Converts a nullable Int to a SentryReplayQuality.
+        ///
+        /// This method extends the ``SentryReplayQuality.init(rawValue:)`` by supporting nil values.
+        ///
+        /// - Parameter rawValue: The raw value to convert.
+        /// - Returns: Corresponding ``SentryReplayQuality`` or `nil` if not a valid raw value or no value is provided.
+        fileprivate static func from(rawValue: Int?) -> SentryReplayOptions.SentryReplayQuality? {
+            guard let rawValue = rawValue else {
+                return nil
+            }
+            return SentryReplayOptions.SentryReplayQuality(rawValue: rawValue)
+        }
+
+        fileprivate var bitrate: Int {
+            self.rawValue * 20_000 + 20_000
+        }
+
+        fileprivate var sizeScale: Float {
+            self == .low ? 0.8 : 1.0
+        }
+
     }
 
     /**
@@ -74,7 +96,7 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
      *
      * - note: The default is true
      */
-    public var maskAllText = true
+    public var maskAllText: Bool
 
     /**
      * Indicates whether session replay should redact all non-bundled image
@@ -82,13 +104,13 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
      *
      * - note: The default is true
      */
-    public var maskAllImages = true
+    public var maskAllImages: Bool
 
     /**
      * Indicates the quality of the replay.
      * The higher the quality, the higher the CPU and bandwidth usage.
      */
-    public var quality = SentryReplayQuality.defaultQuality
+    public var quality: SentryReplayQuality
 
     /**
      * A list of custom UIView subclasses that need
@@ -96,7 +118,7 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
      * By default Sentry already mask text and image elements from UIKit
      * Every child of a view that is redacted will also be redacted.
      */
-    public var maskedViewClasses = [AnyClass]()
+    public var maskedViewClasses: [AnyClass]
 
     /**
      * A list of custom UIView subclasses to be ignored
@@ -104,7 +126,7 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
      * The views of given classes will not be redacted but their children may be.
      * This property has precedence over `redactViewTypes`.
      */
-    public var unmaskedViewClasses = [AnyClass]()
+    public var unmaskedViewClasses: [AnyClass]
 
     /**
      * Alias for ``enableViewRendererV2``.
@@ -125,19 +147,19 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
     /**
      * Enables the up to 5x faster new view renderer used by the Session Replay integration.
      *
-     * Enabling this flag will reduce the amount of time it takes to render each frame of the session replay on the main thread, therefore reducing 
+     * Enabling this flag will reduce the amount of time it takes to render each frame of the session replay on the main thread, therefore reducing
      * interruptions and visual lag. [Our benchmarks](https://github.com/getsentry/sentry-cocoa/pull/4940) have shown a significant improvement of
      * **up to 4-5x faster rendering** (reducing `~160ms` to `~36ms` per frame) on older devices.
      *
-     * - Experiment: In case you are noticing issues with the new view renderer, please report the issue on [GitHub](https://github.com/getsentry/sentry-cocoa). 
+     * - Experiment: In case you are noticing issues with the new view renderer, please report the issue on [GitHub](https://github.com/getsentry/sentry-cocoa).
      *               Eventually, we will remove this feature flag and use the new view renderer by default.
      */
-    public var enableViewRendererV2 = true
+    public var enableViewRendererV2: Bool
 
     /**
      * Enables up to 5x faster but incommpelte view rendering used by the Session Replay integration.
-     * 
-     * Enabling this flag will reduce the amount of time it takes to render each frame of the session replay on the main thread, therefore reducing 
+     *
+     * Enabling this flag will reduce the amount of time it takes to render each frame of the session replay on the main thread, therefore reducing
      * interruptions and visual lag. [Our benchmarks](https://github.com/getsentry/sentry-cocoa/pull/4940) have shown a significant improvement of
      * up to **5x faster render times** (reducing `~160ms` to `~30ms` per frame).
      *
@@ -153,21 +175,21 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
      *               view renderer, please report the issue on [GitHub](https://github.com/getsentry/sentry-cocoa). Eventually, we will
      *               mark this feature as stable and remove the experimental flag, but will keep it disabled by default.
      */
-    public var enableFastViewRendering = false
+    public var enableFastViewRendering: Bool
 
     /**
      * Defines the quality of the session replay.
      * Higher bit rates better quality, but also bigger files to transfer.
      */
-    var replayBitRate: Int {
-        quality.rawValue * 20_000 + 20_000
+    internal var replayBitRate: Int {
+        quality.bitrate
     }
 
     /**
      * The scale related to the window size at which the replay will be created
      */
-    var sizeScale: Float {
-        quality == .low ? 0.8 : 1.0
+    internal var sizeScale: Float {
+        quality.sizeScale
     }
 
     /**
@@ -175,38 +197,78 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
      * The more the havier the process is.
      * The minimum is 1, if set to zero this will change to 1.
      */
-    var frameRate: UInt = 1 {
+    internal var frameRate: UInt {
         didSet {
-            if frameRate < 1 { frameRate = 1 }
+            if frameRate < 1 {
+                frameRate = 1
+            }
         }
     }
 
     /**
      * The maximum duration of replays for error events.
      */
-    let errorReplayDuration = TimeInterval(30)
+    internal var errorReplayDuration: TimeInterval
 
     /**
      * The maximum duration of the segment of a session replay.
      */
-    let sessionSegmentDuration = TimeInterval(5)
+    internal var sessionSegmentDuration: TimeInterval
 
     /**
      * The maximum duration of a replay session.
      */
-    let maximumDuration = TimeInterval(3_600)
+    internal var maximumDuration: TimeInterval
 
     /**
      * Used by hybrid SDKs to be able to configure SDK info for Session Replay
      */
-    var sdkInfo: [String: Any]?
-    
+    internal var sdkInfo: [String: Any]?
+
     /**
-     * Inittialize session replay options disabled
+     * Initialize session replay options disabled
+     *
+     * - Note: This initializer is added for Objective-C compatibility, as constructors with default values
+     *         are not supported in Objective-C.
      */
-    public override init() {
-        self.sessionSampleRate = 0
-        self.onErrorSampleRate = 0
+    public convenience override init() {
+        self.init(
+            // We need to set at least one parameter to avoid the initializer to call itself.
+            //
+            // MAKE SURE THIS DEFAULT VALUE IS ALIGNED WITH THE DEFAULT VALUE IN THE INITIALIZER BELOW.
+            sessionSampleRate: 0,
+        )
+    }
+
+    /// Initializes a new instance of ``SentryReplayOptions`` using a dictionary.
+    ///
+    /// - Parameter dictionary: A dictionary containing the configuration options for the session replay.
+    ///
+    /// - Warning: This initializer is primarily used by Hybrid SDKs and is not intended for public use.
+    convenience init(dictionary: [String: Any]) {
+        // This initalizer is calling the one with optional parameters, so that defaults can be applied
+        // for absent values.
+        self.init(
+            sessionSampleRate: (dictionary["sessionSampleRate"] as? NSNumber)?.floatValue,
+            onErrorSampleRate: (dictionary["errorSampleRate"] as? NSNumber)?.floatValue,
+            maskAllText: (dictionary["maskAllText"] as? NSNumber)?.boolValue,
+            maskAllImages: (dictionary["maskAllImages"] as? NSNumber)?.boolValue,
+            enableViewRendererV2: (dictionary["enableViewRendererV2"] as? NSNumber)?.boolValue
+            ?? (dictionary["enableExperimentalViewRenderer"] as? NSNumber)?.boolValue,
+            enableFastViewRendering: (dictionary["enableFastViewRendering"] as? NSNumber)?.boolValue,
+            maskedViewClasses: (dictionary["maskedViewClasses"] as? NSArray)?.compactMap({ element in
+                NSClassFromString((element as? String) ?? "")
+            }),
+            unmaskedViewClasses: (dictionary["unmaskedViewClasses"] as? NSArray)?.compactMap({ element in
+                NSClassFromString((element as? String) ?? "")
+            }),
+            quality: SentryReplayQuality.from(rawValue: dictionary["quality"] as? Int),
+            sdkInfo: dictionary["sdkInfo"] as? [String: Any],
+            frameRate: (dictionary["frameRate"] as? NSNumber)?.uintValue,
+            errorReplayDuration: (dictionary["errorReplayDuration"] as? NSNumber)?.doubleValue,
+            sessionSegmentDuration: (dictionary["sessionSegmentDuration"] as? NSNumber)?.doubleValue,
+            maximumDuration: (dictionary["maximumDuration"] as? NSNumber)?.doubleValue
+        )
     }
 
     /**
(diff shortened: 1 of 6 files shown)


philprime:
cc @kahest so you know this PR exists.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -216,39 +278,108 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+        onErrorSampleRate: Float = 0,
+        maskAllText: Bool = true,
+        maskAllImages: Bool = true,
+        enableViewRendererV2: Bool = true,
+        enableFastViewRendering: Bool = false
+    ) {
+        // Note for future maintainers:
+        // - This initializer is publicly available for Swift, but not for Objective-C.
+        // - Each parameter has a default value, so the parameter can be omitted.
+        // - The values are not optional, because SDK users should not be able to set them to nil.
```
armcknight, line 292: The fact that the declarations are not optional should be sufficient, no need to restate it here. Additionally, if they ever become optional for some reason, there is no automatic way to know this comment needs to be changed, and so could become a lie in the future. I would just remove it.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -216,39 +278,108 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+    public init(
+        sessionSampleRate: Float = 0,
+        onErrorSampleRate: Float = 0,
+        maskAllText: Bool = true,
+        maskAllImages: Bool = true,
+        enableViewRendererV2: Bool = true,
+        enableFastViewRendering: Bool = false
+    ) {
+        // Note for future maintainers:
+        // - This initializer is publicly available for Swift, but not for Objective-C.
```
armcknight, line 290: I think it is common knowledge that Swift initializers with default values are not available to ObjC, we don't need to write this in every spot.

Additionally, if it is indeed intended that we allow consumers to instantiate an instance of SentryReplayOptions from objc, we should code that in the iOS-ObjectiveC sample app to avoid breaking things in the future.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -153,60 +175,100 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
-    public var enableFastViewRendering = false
+    public var enableFastViewRendering: Bool
 
     /**
      * Defines the quality of the session replay.
      * Higher bit rates better quality, but also bigger files to transfer.
      */
-    var replayBitRate: Int {
-        quality.rawValue * 20_000 + 20_000
+    internal var replayBitRate: Int {
```
armcknight, line 184: Why add all these? Isn't `internal` the default?

armcknight (commented):
Is this change entirely motivated by the inconsistency in the default value being set for `enableViewRendererV2`? Personally, I find these changes confusing and potentially dangerous. We cannot rely on comments to enforce consistency in the business logic, those kinds of constraints need to be coded. Yes, you can catch them in unit tests, but this introduces almost 1,000 lines of test; that is also code that is prone to errors, requires its own maintenance, and slows down future changes. A lot of that checking can be done by the compiler if we keep the target code simpler. And we wouldn't need the hacks to avoid recursive convenience init calls.

My approach here would probably involve deduplicating magic values that are used as defaults in variable declarations and default parameter values into static variables or even into a structure called Defaults. One source of truth. Here's what that would look like: https://github.com/getsentry/sentry-cocoa/pull/5214. A lot of your tests might still provide value so I don't necessarily think you need to throw all that effort away.

It's unfortunate that this class was ever written with publicly available initializers and mutable properties, as this is the kind of mess it leads to. Better would have been to use a builder pattern instead of offering multiple types of initializers. We use that elsewhere so I'm not sure why we didn't here. Sorry you have to deal with that now 😭

philprime:
Thanks for the feedback @armcknight, I already had a discussion with @philipphofmann yesterday, that this situation is quite complex and due the some of these initializers being public and already shipped to production made it even harder and frustrating to find a solution that covers all. 

> Is this change entirely motivated by the inconsistency in the default value being set for enableViewRendererV2? Personally, I find these changes confusing and potentially dangerous. We cannot rely on comments to enforce consistency in the business logic, those kinds of constraints need to be coded. Yes, you can catch them in unit tests, but this introduces almost 1,000 lines of test; that is also code that is prone to errors, requires its own maintenance, and slows down future changes. A lot of that checking can be done by the compiler if we keep the target code simpler. And we wouldn't need the hacks to avoid recursive convenience init calls.

This is the second time that a default value was only partially updated and therefore not consistent, so it must be covered somehow. I agree with most of your statements, these options are just so badly designed that any change can cause public facing issues. IMHO some of the initializers should not have been public in the first place.

> My approach here would probably involve deduplicating magic values that are used as defaults in variable declarations and default parameter values into static variables or even into a structure called Defaults. One source of truth. Here's what that would look like: https://github.com/getsentry/sentry-cocoa/pull/5214. A lot of your tests might still provide value so I don't necessarily think you need to throw all that effort away.

Yes, I had this one approach in my drafts as well but it had two additional downsides:

1. We would introduce another public type for the constants, something I wanted to avoid because SDK users should not care about the default values like that (we can not use an internal type, the type must be public).
2. A person looking at the Xcode documentation of the initializer will not see anymore what the default value is:

image

> It's unfortunate that this class was ever written with publicly available initializers and mutable properties, as this is the kind of mess it leads to. Better would have been to use a builder pattern instead of offering multiple types of initializers. We use that elsewhere so I'm not sure why we didn't here. Sorry you have to deal with that now 😭

Yes, this is terrible to maintain, but it is what is now. I added [a note](https://github.com/getsentry/sentry-cocoa/discussions/2946#discussioncomment-13085656) for v9

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: My intention was to be explicit what's public facing and what not, so that any future maintainer visiting this property will reconsider if making it public is necessary. If no visibility is defined, some might be careless and think something like: "Oh we missed adding a `public` here, that's why it's not available".

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: Yes, common knowledge if you have background in using Objective-C. But in my case I didn't know why the method was there in the first place, because Sesssion Replay is mostly written and used in Swift (in the SDK), but then I realized the constructor is there because of `SentryOptions.m`.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: Code is changed easily, this is context to make maintainers aware we decided on purpose not to make it optional, and have a default fallback in the initializer body.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -153,60 +175,100 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
      */
-    public override init() {
-        self.sessionSampleRate = 0
-        self.onErrorSampleRate = 0
+    public convenience override init() {
+        self.init(
+            // We need to set at least one parameter to avoid the initializer to call itself.
+            //
+            // MAKE SURE THIS DEFAULT VALUE IS ALIGNED WITH THE DEFAULT VALUE IN THE INITIALIZER BELOW.
+            sessionSampleRate: 0
```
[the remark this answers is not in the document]
philprime: I agree with comments not being good quality control. If we want a static variable in the constructor, it must be public, see [my PR comment](https://github.com/getsentry/sentry-cocoa/pull/5210#issuecomment-2865523968).

philipphofmann:
I don't believe this code will be terrible to maintain, because we're not going to change that code often, as it's a public API. The tests can only change when the requirements change, which should only be the case in the next major release. As these APIs are a bit of a mess and we can't fix that now, let's agree on a somewhat OK solution and fix it later.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philipphofmann: As internal is the default, it's a bit redundant, but I don't care too much about that.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philipphofmann: I think it would be good to comment on why it's not available in ObjC.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -216,39 +278,108 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+        sessionSampleRate: Float = 0,
+        onErrorSampleRate: Float = 0,
+        maskAllText: Bool = true,
+        maskAllImages: Bool = true,
+        enableViewRendererV2: Bool = true,
+        enableFastViewRendering: Bool = false
+    ) {
+        // Note for future maintainers:
+        // - This initializer is publicly available for Swift, but not for Objective-C.
+        // - Each parameter has a default value, so the parameter can be omitted.
```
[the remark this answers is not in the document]
philipphofmann: I agree with both of you. If it helps you @philprime for future context, we can keep it. It doesn't hurt much.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -216,39 +278,108 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+        // - The values are not optional, because SDK users should not be able to set them to nil.
+
+        // WHEN CHANGING ANY DEFAULT VALUE IN THE INITIALIZER, UPDATE THE PROPERTY DOCUMENTATION
```
philipphofmann, line 294: `l`: Maybe a bit more specific.
```suggestion
        // WHEN CHANGING ANY DEFAULT VALUE IN THE INITIALIZER, UPDATE THE PROPERTY CODE DOCUMENTATION ABOVE
```

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -216,39 +278,108 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+
         self.sessionSampleRate = sessionSampleRate
         self.onErrorSampleRate = onErrorSampleRate
         self.maskAllText = maskAllText
         self.maskAllImages = maskAllImages
         self.enableViewRendererV2 = enableViewRendererV2
         self.enableFastViewRendering = enableFastViewRendering
+
+        // These are the default values for the properties.
+        // Do not add any default values at the property declaration, because they will be overridden by the initializer.
```
philipphofmann, line 304: `m`: Now looking at this again, I'm reconsidering my statement yesterday. I think having the default values in one place as suggested in https://github.com/getsentry/sentry-cocoa/pull/5214, which has the downside of not showing the default values in the code docs, is easier to maintain. The current approach can break easily, but we're also not going to change this often. We can work around the code doc default value problem, by adding code docs for the initializers and the properties. The SentryOptions are written in ObjC and you can also only get the default value by looking at the code docs. I have no strong opinion on this one. I let you decide @philprime.

Tests/SentrySwiftUITests/PreviewRedactOptionsTests.swift
```diff
@@ -0,0 +1,140 @@
...
+            unmaskedViewClasses: [UIButton.self],
+            enableViewRendererV2: true
+        )
+        // -- Assert --
+        XCTAssertFalse(options.maskAllText)
+        XCTAssertFalse(options.maskAllImages)
+        XCTAssertEqual(options.maskedViewClasses.count, 1)
+        for maskedViewClass in options.maskedViewClasses {
+            XCTAssertEqual(String(describing: maskedViewClass), String(describing: UIView.self))
+        }
```
philipphofmann, line 32: `l`: Maybe we can use `assertArrayEquals` or some other helper method to avoid having to write this assertion code multiple times.

Tests/SentryTests/Integrations/SessionReplay/SentryReplayOptionsObjcTests.m
```diff
@@ -0,0 +1,41 @@
...
+
+@interface SentryReplayOptionsObjcTests : XCTestCase
+
+@end
+
+@implementation SentryReplayOptionsObjcTests
+
+- (void)testInit_withoutArguments_shouldUseDefaults
+{
+    // Note for future maintainers:
```
philipphofmann, line 13: `l`: Every note is for future maintainers 😅 . No need to point that out.
```suggestion
```

Tests/SentryTests/Integrations/SessionReplay/SentryReplayOptionsTests.swift
```diff
@@ -3,34 +3,217 @@ import Foundation
...
+        XCTAssertEqual(options.maskedViewClasses.count, 0)
+        XCTAssertEqual(options.unmaskedViewClasses.count, 0)
+        XCTAssertEqual(options.quality, SentryReplayOptions.SentryReplayQuality.defaultQuality)
+        XCTAssertEqual(options.frameRate, 1)
+        XCTAssertEqual(options.errorReplayDuration, 30)
+        XCTAssertEqual(options.sessionSegmentDuration, 5)
+        XCTAssertEqual(options.maximumDuration, 60 * 60)
+    }
+
+    func testInit_sessionSampleRateOmitted_shouldUseDefaultValues() {
```
philipphofmann, line 58: `l`: Can't we reduce the test cases by omitting all except one parameter in one test, and then omitting just this parameter in another test. We basically just need two test cases. So basically something like this

Test omitting all other than `sessionSampleRate`
```
let options = SentryReplayOptions(
            sessionSampleRate: 0.5,
        )
```

Tests omitting `sessionSampleRate`
```
let options = SentryReplayOptions(
           
            onErrorSampleRate: 0.8,
            maskAllText: false,
            maskAllImages: false,
            enableViewRendererV2: false,
            enableFastViewRendering: true
        )
```

philipphofmann (approved):
LGTM, we need so many tests because the logic is so complex. Without all these tests,  we can't be sure everything works. We can reduce the number of tests to omit some of the parameters, but the rest is required, IMO. FYI @armcknight.

Thanks for adding all the tests, @philprime

philipphofmann:
> It's unfortunate that this class was ever written with publicly available initializers and mutable properties, as this is the kind of mess it leads to. Better would have been to use a builder pattern instead of offering multiple types of initializers. We use that elsewhere so I'm not sure why we didn't here. Sorry you have to deal with that now 😭

And yes, that's the root cause. We should fix this in V9.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -216,39 +278,108 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
      *  - errorSampleRate Indicates the percentage in which a 30 seconds replay will be send with
      * error events.
      */
-    public init(sessionSampleRate: Float = 0, onErrorSampleRate: Float = 0, maskAllText: Bool = true, maskAllImages: Bool = true, enableViewRendererV2: Bool = false, enableFastViewRendering: Bool = false) {
+    public init(
+        sessionSampleRate: Float = 0,
+        onErrorSampleRate: Float = 0,
+        maskAllText: Bool = true,
+        maskAllImages: Bool = true,
+        enableViewRendererV2: Bool = true,
```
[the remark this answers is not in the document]
philprime: Thank you for testing!

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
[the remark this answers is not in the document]
philprime: I want to keep this, added a note related to unavailability in ObjC so the comment is a comparision between Swift and ObjC and not just an explanaiton of the Swift compiler.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: Removed the internal

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: We are trading maintainability for usability. I'll create a well-documented defaults type so that users can also easily understand the values. I will NOT add the default values to the code documentation for this initializer, because then we are back at the original issue of having multiple fields to maintain.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Detect AppHangsV2 when tracing not enabled (#5184)
+- Fix default and dictionary values for session replay options (#5210)
```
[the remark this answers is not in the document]
philprime: I added two entries, one for the fix of unaligned default values and one for the improvements regarding the dictionary initializer for Hybrid SDKs

Tests/SentryTests/Integrations/SessionReplay/SentryReplayOptionsTests.swift
philprime: I believe we should test them omitted individually because removing most of the defaults could easily cause a different initializer being tested instead, e.g. removing all would cause the default initializer to be triggered instead of using all default values --> the original problem.

Relying on at least one value being set could work, which was the previous attempt aswell.

At least I can remove the assertions for unrelated fields, so we only assert the default value of each omitted field

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: I decided to add a comment to each initializer and property pointing to the default values type.

philprime:
Based on your review feedback I refactored the initializers to use the default values in the new public type `SessionReplayOptions.DefaultValues`. I added a reference to all of the initializers and properties so SDK users can find more details

antonis (approved):
Thank you for all the changes @philprime 🙇 
The changes on this PR LGTM and worked as expected [with RN](https://github.com/getsentry/sentry-react-native/pull/4815) in my tests 🚀

Tests/SentryTests/Integrations/SessionReplay/SentryReplayOptionsTests.swift
philipphofmann: Let's keep them then.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 ### Fixes
 
 - Detect AppHangsV2 when tracing not enabled (#5184)
+- Add shared default values for session replay options initializers (#5210)
```
philipphofmann, line 8: `m`: Can we point out which default values we fixed, please?
```suggestion
- Fix default values for session replay options (#5210) TODO add the fixed ones.
```

Sources/SentrySwiftUI/Preview/PreviewRedactOptions.swift
```diff
@@ -1,14 +1,66 @@
...
 public class PreviewRedactOptions: SentryRedactOptions {
+    /**
+     * Flag to redact all text in the app by drawing a black rectangle over it.
+     *
+     * - Note: See ``SentryReplayOptions.DefaultValues.maskAllText`` for the default value.
+     */
     public let maskAllText: Bool
+
+    /**
+     * Flag to redact all non-bundled image in the app by drawing a black rectangle over it.
```
philipphofmann, line 18: `l`: A typo, I believel
```suggestion
     * Flag to redact all non-bundled images in the app by drawing a black rectangle over them.
```

Sources/SentrySwiftUI/Preview/PreviewRedactOptions.swift
```diff
@@ -1,14 +1,66 @@
...
+    
+    /**
+     * The classes of views to exclude from masking.
+     *
+     * - Note: See ``SentryReplayOptions.DefaultValues.unmaskedViewClasses`` for the default value.
+     */
     public let unmaskedViewClasses: [AnyClass]
+    
+    /**
+     * Enables the up to 5x faster new view renderer used by the Session Replay integration.
```
philipphofmann, line 39: `l`: It soon won't be new, so we can ditch it.
```suggestion
     * Enables the up to 5x faster view renderer used by the Session Replay integration.
```

Sources/SentrySwiftUI/Preview/PreviewRedactOptions.swift
```diff
@@ -1,14 +1,66 @@
...
+     */
     public let enableViewRendererV2: Bool
 
-    public init(maskAllText: Bool = true, maskAllImages: Bool = true, maskedViewClasses: [AnyClass] = [], unmaskedViewClasses: [AnyClass] = [], enableViewRendererV2: Bool = false) {
+    /**
+     * Initializes a new instance of ``PreviewRedactOptions`` with the specified parameters.
+     *
+     * - Parameters:
+     *   - maskAllText: Flag to redact all text in the app by drawing a black rectangle over it.
+     *   - maskAllImages: Flag to redact all non-bundled image in the app by drawing a black rectangle over it.
```
philipphofmann, line 50: `l`: Same typo here.
```suggestion
     *   - maskAllImages: Flag to redact all non-bundled images in the app by drawing a black rectangle over them.
```

Sources/SentrySwiftUI/Preview/PreviewRedactOptions.swift
```diff
@@ -1,14 +1,66 @@
...
+     *   - maskedViewClasses: The classes of views to mask.
+     *   - unmaskedViewClasses: The classes of views to exclude from masking.
+     *   - enableViewRendererV2: Enables the up to 5x faster new view renderer used by the Session Replay integration.
```
philipphofmann, line 53: `l`: Also ditch new.
```suggestion
     *   - enableViewRendererV2: Enables the up to 5x faster view renderer used by the Session Replay integration.
```

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -152,103 +207,206 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+            maximumDuration: (dictionary["maximumDuration"] as? NSNumber)?.doubleValue
+        )
+    }
+
+    /**
+     * Initializes a new instance of ``SentryReplayOptions`` with the specified parameters.
+     *
+     * - Parameters:
+     *   - sessionSampleRate: Indicates the percentages in which replay for the session will be created.
+     *   - onErrorSampleRate: Indicates the percentages in which a buffer of 30 seconds replay will be send with error events.
```
philipphofmann, line 337:
```suggestion
     *   - sessionSampleRate: Indicates the percentage in which replay for the session will be created.
     *   - onErrorSampleRate: Indicates the percentage in which a buffer of 30 seconds replay will be sent with error events.
```

philipphofmann, line 337: `m`: I'm a bit confused by the wording here. I don't really understand this. Maybe try to rephrase it, please.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -152,103 +207,206 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+     *   - maskAllText: Flag to redact all text in the app by drawing a black rectangle over it.
+     *   - maskAllImages: Flag to redact all non-bundled image in the app by drawing a black rectangle over it.
```
philipphofmann, line 339: `l`:
```suggestion
     *   - maskAllImages: Flag to redact all non-bundled images in the app by drawing a black rectangle over them.
```

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -152,103 +207,206 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+     *   - enableViewRendererV2: Enables the up to 5x faster new view renderer used by the Session Replay integration.
```
philipphofmann, line 340:
```suggestion
     *   - enableViewRendererV2: Enables the up to 5x faster view renderer used by the Session Replay integration.
```

philipphofmann (approved):
LGTM, with a couple of suggestions. Thanks

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: I agree, these were already in the class before I started refactoring. Will replace them.

philipphofmann: > these were already in the class before I started refactoring

That explains it, then 😅 .  I'm used to higher standard comments from you 😄

CHANGELOG.md
philprime: I adapted it, it's only the `enableViewRendererV2` which changed in the initializer.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: I rewrote them, should be more concise now.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -333,12 +339,12 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
      *
      * - Parameters:
-     *   - sessionSampleRate: Indicates the percentages in which replay for the session will be created.
-     *   - onErrorSampleRate: Indicates the percentages in which a buffer of 30 seconds replay will be send with error events.
-     *   - maskAllText: Flag to redact all text in the app by drawing a black rectangle over it.
-     *   - maskAllImages: Flag to redact all non-bundled image in the app by drawing a black rectangle over it.
-     *   - enableViewRendererV2: Enables the up to 5x faster new view renderer used by the Session Replay integration.
-     *   - enableFastViewRendering: Enables faster but incomplete view rendering used by the Session Replay integration.
+     *   - sessionSampleRate: Sample rate used to determine the percentage of replays of sessions that will be uploaded.
+     *   - onErrorSampleRate: Sample rate used to determine the percentage of replays of error events that will be uploaded.
```
philipphofmann, line 343: Better thanks

philipphofmann (approved):
Thanks

armcknight:
Nice work getting this across the finish line @philprime! 🥇

PR #5213, opened 2025-05-08
chore: Add manually configured entitlement files and configurations

armcknight:
These were added in https://github.com/getsentry/sentry-cocoa/pull/5208 but the entitlement files aren't actually needed for the test to pass.

#skip-changelog


philprime:
To answer your questions in this [PR comment](https://github.com/getsentry/sentry-cocoa/pull/5208#pullrequestreview-2826494007).

> - was something broken that this fixes? In the original https://github.com/getsentry/sentry-cocoa/pull/5143 that introduced the mac app variants, i actually removed the entitlements file altogether and only set ENABLE_APP_SANDBOX = YES in the project's build settings and confirmed that was able to reproduce the issue, which I moved to the xcconfig files in the xcodegen refactor (https://github.com/getsentry/sentry-cocoa/pull/5170) so nothing was accidentally lost, you can see no entitlements file appears in that diff at all.

I was looking at sandbox settings due to the discovery of [sentry-wizard#988](https://github.com/getsentry/sentry-wizard/issues/988), where we realized that the SentrySDK can not be used when sandbox is enabled without checking "Outgoing Connections (Client)".

That's when I realized that the Entitlements file where absent and that the App Sandbox settings where also not there where I expected to find them in Xcode:

Screenshot 2025-05-09 at 09 36 34

So basically when checking out this PR, I am not able to use `macOS-Swift-Sandboxed` to perform tests with data being send to Sentry.

I was not aware that `ENABLE_APP_SANDBOX` exists as I am used to manage Info.plist and Entitlements as actual files instead of letting Xcode generate them from build settings. Thanks for that, will keep it in mind when fixing the wizard 👍 

But to be able to use the Sentry SDK and have it actually send envelopes in a sandboxed macOS app, we will have to check the "Outgoing Connections (Client)" in the App Sandbox settings. Can we do that from the build settings as well, when using `ENABLE_APP_SANDBOX`?

> - why does the entitlements file for macOS-Swift also have com.apple.security.app-sandbox set to <true />? actually, looking again, i think macOS-Swift.entitlements is simply unused, removing in

Ah yes, that file was a left-over from local testing and accidentally committed by me and missed in PR review. Thanks for pointing this out, that one should be removed as it isn't used.

> why do we provide the entitlement files with the settings, but then also specify the setting again in the xcodegen spec?

see answers above

> ETA: Ah I see what happened here, you checked in the generated files. Please don't commit multiple sources of truth to the repository as it can cause confusion and drift, as I mentioned in https://github.com/getsentry/sentry-cocoa/pull/5106#discussion_r2055073160

That's why I reached out on Slack, as I wanted to circle back to you that it doesn't lead to a misunderstanding or perceives as undermining of the previous discussion. It wasn't entirely clear which files are checked in and which ones not:

| File | Can be generated? | Is being generated? | Is ignored? |
|---|---|---|---| 
| .xcodeproj | YES | YES | YES |
| Info.plist | YES | NO | NO |
| .entitlements | YES | NO | YES |

armcknight:
@philprime Thanks for the detailed response. I'm totally fine using entitlements for sandboxing and the outgoing client connection thing as I didn't know about the work being done around that.

My preference would be to check in an entitlements file and reference it with CODE_SIGN_ENTITLEMENTS in an xcconfig, instead of declaring the settings in the XcodeGen spec to have it generate an entitlements file, but I don't feel too strongly about this.

The one thing I will continue to request is, if we go the XcodeGen-managed route, that we don't check in the generated entitlements files.

Would you be ok with me modifying this PR as such, with checked in entitlements files with all the needed settings, and referencing them from the projects?

philprime:
In that case let's treat it the same way as we do with `Info.plist` and will not generate the file using xcodegen and directly configure it using `CODE_SIGN_ENTITLEMENTS`.

> Would you be ok with me modifying this PR as such, with checked in entitlements files with all the needed settings, and referencing them from the projects?

Yes, I can do that.

Samples/iOS15-SwiftUI/iOS15-SwiftUI.xcconfig
```diff
@@ -22,7 +22,8 @@
 #include "../Shared/Config/Metal.xcconfig"
 
 PRODUCT_BUNDLE_IDENTIFIER = io.sentry.iOS15-SwiftUI
-INFOPLIST_FILE = iOS-SwiftUI/Info.plist
+INFOPLIST_FILE = iOS15-SwiftUI/Info.plist
```
philprime, line 25: It seems to me like the Info.plist was misconfigured too

Samples/watchOS-Swift/watchOS-Swift-App.xcconfig
```diff
@@ -22,7 +22,8 @@
 #include "../Shared/Config/Metal.xcconfig"
 
 PRODUCT_BUNDLE_IDENTIFIER = io.sentry.watchOS-Swift.watchkitapp
-INFOPLIST_FILE = "watchOS=Swift WatchKit App/Info.plist"
+INFOPLIST_FILE = watchOS-Swift WatchKit App/Info.plist
```
philprime, line 25: Found a typo in the INFOPLIST_FILE

philprime:
@armcknight I went through all the configs and manually tested if I can configure capabilities from the Xcode UI

Samples/watchOS-Swift/watchOS-Swift-App.xcconfig
armcknight: Nice finds on these 👍🏻

Samples/iOS15-SwiftUI/iOS15-SwiftUI.xcconfig
armcknight: Nice finds on these 👍🏻

philprime (approved):
As we @armcknight switched roles I take your enabling of the auto-merge as an "approval" for the changes, and actually approve this PR to get it merged.

PR #5214, opened 2025-05-09
ref: propose using a defaults struct for SentryReplayOptions

armcknight:
Proposal for how to centralize default values.

#skip-changelog


philprime:
I applied this feedback in #5210. Thank you for the sample.

Issue #4925, opened 2025-03-01
Parallelize components of xcframework build


armcknight:
### Description

The CI jobs for building an xcframework for both the build.yml and release.yml workflows takes about 20 minutes. It builds for multiple platforms, and multiple architectures in many of them. I don't yet know why each linker command appears 3 times (using the raw logs downloaded from https://github.com/getsentry/sentry-cocoa/actions/runs/13595258247/job/38010561537):
```
❯ grep 'clang++ -Xlinker -reproducible -target' ~/Desktop/job-logs.txt | awk -F' ' '{print $6};' | sort | uniq -c
   3 arm64-apple-ios11.0
   3 arm64-apple-ios11.0-simulator
   3 arm64-apple-ios13.1-macabi
   3 arm64-apple-macos10.13
   3 arm64-apple-tvos11.0
   3 arm64-apple-tvos11.0-simulator
   3 arm64-apple-watchos4.0
   3 arm64-apple-watchos4.0-simulator
   3 arm64-apple-xros1.0
   3 arm64-apple-xros1.0-simulator
   3 arm64_32-apple-watchos4.0
   3 arm64e-apple-ios11.0
   3 arm64e-apple-ios13.1-macabi
   3 arm64e-apple-macos10.13
   3 arm64e-apple-tvos11.0
   3 arm64e-apple-watchos4.0
   3 arm64e-apple-xros1.0
   3 armv7k-apple-watchos4.0
   3 i386-apple-watchos4.0-simulator
   3 x86_64-apple-ios11.0-simulator
   3 x86_64-apple-ios13.1-macabi
   3 x86_64-apple-macos10.13
   3 x86_64-apple-tvos11.0-simulator
   3 x86_64-apple-watchos4.0-simulator
   3 x86_64-apple-xros1.0-simulator
```

armcknight:
I guess the multiple runs for each are for these variants (each call to generate_xcframework runs through all the platform architectures): https://github.com/getsentry/sentry-cocoa/blob/8e55ea25fc048dfd83625cd52022a736f048f0b1/scripts/build-xcframework.sh#L125-L134

armcknight:
Future possible tasks:

[...]

PR #5218, opened 2025-05-09
ci: parallelize xcframework slice builds

armcknight:
The job to build the xcframework builds all the slices in serial, which takes upwards of 30 minutes. There are at least 6 slices, so if we parallelize these, we should be able to cut that time down to a small fraction.

Before: [30 minutes 52 seconds](https://github.com/getsentry/sentry-cocoa/actions/runs/14936883321/job/41966429544?pr=5218)
After: [6 minutes 17 seconds](https://github.com/getsentry/sentry-cocoa/actions/runs/15150880219)

One thing I noticed but haven't yet addressed is that the SentrySwiftUI build actually produced Sentry.xcframework as well, because to archive the SentrySwiftUI target, Sentry target must be built as well as it's a dependency. So, we could avoid the duplicated tasks with some tweaks to the workflow logic. I just haven't figured out how to do that yet.

Another future task could be caching individual slices and assembled variants, as not all will always need to be rebuilt (think, if uikit code is modified, then the UIKitless variant won't need to be rebuilt).

#skip-changelog

Fixes GH-4925


philipphofmann:
That would be super nice to get done. Thanks for starting this @armcknight.

armcknight:
Got the slices all building in parallel, working on downloading the artifacts for assembly together. It's looking like the total job time will be around 5-6 minutes now based on https://github.com/getsentry/sentry-cocoa/actions/runs/15080146739

armcknight:
Fiiiiinally got all the things in the right path locations to get it all passing again. Final result: 6m17s.

armcknight:
The failing tests are due to having removed the old build-xcframework.sh script but not yet updating benchmarking.yml or ui-tests.yml to use the new reusable workflows that parallelize the steps from it; #5277 fixes that and should be merged into this branch before merging this one to `main`.

.github/workflows/release.yml
```diff
@@ -1,6 +1,8 @@
 name: Release
 run-name: Release ${{ github.event.inputs.version }}
 
+# TODO: change this to dispatch the build-xcframework.yml workflow, and then depend on it here with the workflow dependency setup described at https://stackoverflow.com/a/64733705
```
armcknight, line 4: done and removed in #5277

itaybre (approved):

scripts/create-carthage-json.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -euo pipefail
 
-echo "{ \"1.0\": \"file:///$(pwd)/Carthage/Sentry.framework.zip?alt=file:///$(pwd)/Carthage/Sentry.xcframework.zip\" }" > ./Samples/Carthage-Validation/Sentry.Carthage.json
-echo "{ \"1.0\": \"file:///$(pwd)/Carthage/SentrySwiftUI.framework.zip?alt=file:///$(pwd)/Carthage/SentrySwiftUI.xcframework.zip\" }" > ./Samples/Carthage-Validation/SentrySwiftUI.Carthage.json
+echo "{ \"1.0\": \"file://$(pwd)/Carthage/Sentry.framework.zip?alt=file://$(pwd)/Carthage/Sentry.xcframework.zip\" }" > ./Samples/Carthage-Validation/Sentry.Carthage.json
```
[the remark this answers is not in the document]
armcknight: There is a validation step that builds a project using Carthage and the local assets.

Makefile
```diff
@@ -98,18 +98,6 @@ analyze:
 	rm -rf analyzer
 	set -o pipefail && NSUnbufferedIO=YES xcodebuild analyze -workspace Sentry.xcworkspace -scheme Sentry -configuration Release CLANG_ANALYZER_OUTPUT=html CLANG_ANALYZER_OUTPUT_DIR=analyzer CODE_SIGNING_ALLOWED="NO" 2>&1 | xcbeautify && [[ -z `find analyzer -name "*.html"` ]]
 
-# Since Carthage 0.38.0 we need to create separate .framework.zip and .xcframework.zip archives.
-# After creating the zips we create a JSON to be able to test Carthage locally.
-# For more info check out: https://github.com/Carthage/Carthage/releases/tag/0.38.0
-build-xcframework:
```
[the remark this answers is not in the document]
armcknight: I agree. I kept a version of it around about halfway through this PR and then gave up after some point due to the flux of changes. I'll work on adding it back in.

.github/workflows/test.yml
```diff
@@ -240,7 +240,6 @@ jobs:
           path: |
             raw-build-output.log
             raw-build-for-testing-output.log
-            raw-test-output.log
```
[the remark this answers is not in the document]
armcknight: Good catch. I don't remember what happened here, fixed.

.github/workflows/assemble-xcframework-variant.yml
```diff
@@ -0,0 +1,49 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          pattern: xcframework-${{inputs.variant-id}}-slice-*
+          path: xcframework-slices
+
+      - name: Assemble XCFramework
+        run: ./scripts/assemble-xcframework.sh "${{github.workspace}}/xcframework-slices" "${{inputs.name}}${{inputs.suffix}}"
+        shell: bash
+
+      - name: Archive XCFramework
```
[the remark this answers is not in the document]
armcknight: I think we've simply cargo-culted "archive" all over the place for any steps that use `actions/upload-artifact` 😅. I can fix here.

.github/workflows/build-xcframework-variant-slices.yml
```diff
@@ -0,0 +1,100 @@
...
+      # We also can't use Xcode 16.x because validating the XCFramework then fails with Xcode 15.x.
+      - run: ./scripts/ci-select-xcode.sh 15.2
+        shell: bash
+
+      - name: Build ${{inputs.name}}${{inputs.suffix}} XCFramework slice for ${{matrix.sdk}}
+        if: startsWith(github.ref, 'refs/heads/release/') == false
+        run: ./scripts/build-xcframework-slice.sh ${{matrix.sdk}} ${{inputs.name}} "${{inputs.suffix}}" ${{inputs.macho-type}} ${{inputs.configuration-suffix}}
+        shell: bash
+
+      - name: Remove Sentry.framework from SentrySwiftUI build
```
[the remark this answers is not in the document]
armcknight: I added a comment. Keep in mind this is how this used to work, I left a note that we could improve the process later.

Makefile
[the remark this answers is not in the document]
armcknight: Done, currently called build-xcframework-local.sh, even though it's also used in a couple new CI workflows, but those could be improved in the future to use a cached build product from a `main` run, or these new parallel tasks if there's no cache available.

armcknight:
@philipphofmann agreed on the long names. GHA just truncates with no option to resize the pane 😭 I whittled it down to barebones: 
image
|||
|---|---|
|image|image|

scripts/assemble-xcframework.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -x
+
+search_path="$1"
+xcframework_name="$2"
+
+xcodebuild_cmd="xcodebuild -create-xcframework"
+
+for slice_dir in "$search_path"/xcframework-*; do
+    framework_path=$(find "$slice_dir" -name '*.framework' -print -quit)
```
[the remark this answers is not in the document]
armcknight: OK, I did wind up specifying the list of SDKs to search for frameworks, rather than using `find`. This also validates all the slices are present that we expect.

armcknight:
OK, I fixed the missing maccatalyst problem (refactoring how that's handled–making it a first class sdk slice job), and rebuilt the local version of the script which is called under the old `make build-xcframework` (and `make build-signed-xcframework`).

.github/workflows/release.yml
```diff
@@ -15,46 +15,64 @@ on:
...
-        run: bundle exec fastlane prepare_signed_xcframework
-        env:
-          APP_STORE_CONNECT_KEY_ID: ${{ secrets.APP_STORE_CONNECT_KEY_ID }}
-          APP_STORE_CONNECT_ISSUER_ID: ${{ secrets.APP_STORE_CONNECT_ISSUER_ID }}
-          APP_STORE_CONNECT_KEY: ${{ secrets.APP_STORE_CONNECT_KEY }}
-          FASTLANE_KEYCHAIN_PASSWORD: ${{ secrets.FASTLANE_KEYCHAIN_PASSWORD }}
-          MATCH_GIT_PRIVATE_KEY: ${{ secrets.MATCH_GIT_PRIVATE_KEY }}
-          MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
-          MATCH_USERNAME: ${{ secrets.MATCH_USERNAME }}
-        shell: sh
```
philippsuess, line 38: `h`: I'm a bit confused. This PR renames the fastlane workflow to `prepare_xcframework_signing` for signing, but I don't see any invocations of `prepare_xcframework_signing` nor do I see anything signing the XCFramework during the release.yml. Am I missing something? If I'm not mistaken, this PR removes signing of the XCFramework or am I missing something?

scripts/assemble-xcframework.sh
```diff
@@ -0,0 +1,70 @@
...
+IFS=',' read -r -a sdks <<< "$4"
+
+# on ci, the xcarchives live in paths like the following:
+#   /path/to/.../xcframework-slices/xcframework-sentry-swiftui-slice-maccatalyst/Library/Frameworks/SentrySwiftUI.framework
+#   /path/to/.../xcframework-slices/xcframework-sentry-swiftui-slice-macosx/Library/Frameworks/SentrySwiftUI.framework
+#   /path/to/.../xcframework-slices/xcframework-sentry-swiftui-slice-iphoneos/Library/Frameworks/SentrySwiftUI.framework
+# in the local build script they're in something like:
+#   /path/to/.../Carthage/archive/Sentry-WithoutUIKitOrAppKit/iphoneos.xcarchive
+#   /path/to/.../Carthage/archive/Sentry-WithoutUIKitOrAppKit/macos.xcarchive
+# the issue is that we need to inject the sdk name once into the local version, and twice into the ci version. a template string satisfies this requirement.
```
philippsuess, line 17: Thanks for the comment.

philipphofmann (commented):
When I compare the contents of the [Sentry-Dynamic.xcframework.zip](https://github.com/getsentry/sentry-cocoa/releases/download/8.52.1/Sentry-Dynamic.xcframework.zip) from the latest release and this PR, I see that the PRs xcframework has some parts missing. It would be great to double check this with all XCFrameworks.

Screenshot 2025-06-06 at 14 14 35


Before merging this PR, we should trigger an alpha release directly from this branch and manually test the signed XCFrameworks with an SPM sample application. We should do this for all different XCFrameworks to ensure we're not breaking something.

Screenshot 2025-06-06 at 14 18 05

.github/workflows/release.yml
armcknight: You're not missing anything, I think I removed the job that prepares for signing somehow. But, the product is now signed by the assembly workflow with this input: https://github.com/getsentry/sentry-cocoa/pull/5218/files?diff=split&w=0#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R57

scripts/assemble-xcframework.sh
armcknight: it's actually a bit outdated, it wound up being the case that the SDK name only appears once in each path. i might look at moving the files to a common location to avoid having to do this altogether, although at least it does work

armcknight:
I modified `release.yml` so it can run everything for PRs except the release action and diffed the build result against the last release, everything is there:
image

armcknight:
We looked at the difference in the macosx slice's size and discovered that the upload artifact action zips up the files and in the process resolves symlinks, so was duplicating things. We are instead ditto-ing the files before uploading them.

armcknight:
Fixed the remaining symlink issues.

Manually tested Sentry-Dynamic, Sentry (static) and SentrySwiftUI in a test project for each platform, including mac catalyst.

armcknight:
Also I realized that now that `release.yml` would run most of its jobs in PRs, it would duplicate the build in `build-xcframework.yml`. Also, in releases, `build-xcframework.yml` would still run. So, I moved everything into `release.yml` in https://github.com/getsentry/sentry-cocoa/pull/5218/commits/45e2a9eaf5a9bcc4cf656c4e506d47b09dfa11ea

armcknight:
i've locally validated with the integration tests in https://github.com/getsentry/sentry-cocoa/pull/5384 that each kind of app and package manager can correctly link each type of SDK variant.

.github/workflows/assemble-xcframework-variant.yml
```diff
@@ -0,0 +1,98 @@
...
+      variant-id:
+        description: |-
+          The ID of the variant to build an XCFramework slice for. Used to collect appropriate slices for final deliverable assembly.
+          E.g. "sentry-static", "sentry-dynamic" or "sentry-withoutuikit-dynamic"
+        required: true
+        type: string
+
+      signed:
+        description: |-
+          Whether or not the assembled XCFramework should be signed.
```
philprime, line 36: `l`: Not sure what the valid inputs here are, but IIUC it is `--sign`, so I would add this as an example aswell

```suggestion
          Whether or not the assembled XCFramework should be signed.
          E.g. `--sign`
```

.github/workflows/assemble-xcframework-variant.yml
```diff
@@ -0,0 +1,98 @@
...
+        run: bundle exec fastlane prepare_xcframework_signing
+        env:
+          APP_STORE_CONNECT_KEY_ID: ${{ secrets.APP_STORE_CONNECT_KEY_ID }}
+          APP_STORE_CONNECT_ISSUER_ID: ${{ secrets.APP_STORE_CONNECT_ISSUER_ID }}
+          APP_STORE_CONNECT_KEY: ${{ secrets.APP_STORE_CONNECT_KEY }}
+          FASTLANE_KEYCHAIN_PASSWORD: ${{ secrets.FASTLANE_KEYCHAIN_PASSWORD }}
+          MATCH_GIT_PRIVATE_KEY: ${{ secrets.MATCH_GIT_PRIVATE_KEY }}
+          MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
+          MATCH_USERNAME: ${{ secrets.MATCH_USERNAME }}
+        shell: sh
```
philprime: `l`: Reason why this is explicitly `sh` and not the default shell?

.gitignore
```diff
@@ -46,8 +46,7 @@ Carthage/Checkouts
 Carthage
 Sentry.Carthage.json
 SentrySwiftUI.Carthage.json
-Sentry.framework.zip
-Sentry.xcframework.zip
+*xcframework*
```
philprime: `m`: This can cause issues as we do have scripts using the keyword `xcframework`, like `assemble-xcframework-variant.yaml` or `build-xcframework-local.sh`

These files are already checked in so they are tracked and not ignored, but any new file, or maybe even for a file move/rename this can lead to files being ignored and the next maintainer will have to search this line here again.

Maybe we can narrow this glob down.

scripts/zip_built_sdk.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
philprime: `l`: Nit pick, as we use `ditto` instead of `zip` we could call this script `archive_built_sdk.sh` instead.

.github/workflows/assemble-xcframework-variant.yml
armcknight: I'd actually documented this in #5384, but it's always bothered me that i wound up using a string for a boolean. I just made it a boolean here instead.

.github/workflows/assemble-xcframework-variant.yml
armcknight: GitHub actions complained that this didn't explicitly declare a shell. I'll try just removing this declaration again, maybe it was unrelated.

.gitignore
armcknight: I can put it back to the explicit list of all of them 👍🏻

scripts/zip_built_sdk.sh
armcknight: I didn't name this file, but I renamed it once and I'll do it again 😏

.github/workflows/assemble-xcframework-variant.yml
armcknight: Yep, seems to work now

PR #5220, opened 2025-05-12
test: Add separate test plan for HTTPTransport unit tests

philprime:
# Description

- Changes the `Sentry_Base.xctestplan` to skip tests defined for `HTTPTransport`. This test plan is still configured to include all newly added tests to this test plan.
- Adds a separate test plan `Sentry_HTTPTransport.xctestplan` with explicitly defined tests to run separately.

We need to discuss the impact on the developer experience, because running tests in Xcode will only execute the default scheme (`Sentry_Base`).

# Context

- Test plans do not support name globs
- Executing tests in separate test plans will isolate execution context, therefore reduce tests interfering with each other.
- This is not intended to improve test isolation, just to reduce failing CI

#skip-changelog

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -317,6 +317,9 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: false
           verbose: true
+          # Ignore test coverage of tests
+          files: slather/coverage/**/*.xml
+          exclude: "*.xctest.coverage.txt"
 
       # Sometimes codecov uploads etc can fail. Retry one time to rule out e.g. intermittent network failures.
       - name: Push code coverage to codecov
@@ -327,3 +330,5 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
           verbose: true
+          files: slather/coverage/**/*.xml
+          exclude: "*.xctest.coverage.txt"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -253,13 +253,16 @@ jobs:
           done
 
       - name: List generated test reports
-        run: ls -laR build/reports
+        run: ls -lR build/reports
+
+      - name: List generated test coverage
+        run: ls -lR slather/coverage
 
       - name: Publish Test Report
         uses: mikepenz/action-junit-report@cf701569b05ccdd861a76b8607a66d76f6fd4857 # v5.5.1
         if: always()
         with:
-          report_paths: "**/build/reports*/junit.xml"
+          report_paths: "**/build/reports/**/*.xml"
           fail_on_failure: true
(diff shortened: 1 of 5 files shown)


philprime:
@philipphofmann this is a proposal based on one of our 1:1 chats we had a while ago. We could also go for a matrix so that test plans are individual jobs that can be restarted.

Plans/Sentry_Base.xctestplan
```diff
@@ -33,7 +33,64 @@
         "SentryCrashReportStore_Tests\/testCrashReportCount1",
         "SentryCrashReportStore_Tests\/testDeleteAllReports",
         "SentryCrashReportStore_Tests\/testStoresLoadsMultipleReports",
+        "SentryHttpTransportTests",
```
armcknight, line 36: I think instead of doing this we should also move them into a new unit test target. This will not age well.

.github/workflows/test.yml
```diff
@@ -146,6 +160,8 @@ jobs:
             xcode: "15.4"
             test-destination-os: "17.5"
             scheme: "Sentry"
+            additional-test-plans:
+              - "Sentry_HTTPTransport"
 
           # iOS 17
           - runs-on: macos-14
```
armcknight, line 167: This PR highlights to me that we have a job matrix of N runs that run the Sentry unit tests, and one that runs SentrySwiftUI. I would prefer we extract that SentrySwiftUI run to its own job, then the remaining will all also run the HTTP transport tests, so we don't have to use a matrix variable for it, or the extra logic in the run step. If we combine that with using a new test target to run the HTTP tests, like I've previously done with the profiler tests, this could be greatly simplified. It also moves us in the right direction for modularizing the SDK: I think our unit tests need to be broken down into targets per component anyways.

armcknight (commented):
Some ideas for your consideration.

Plans/Sentry_Base.xctestplan
philprime: I am not entirely happy with this aswell. My idea was not having to manage multiple targets and making sure the settings match (even though xcodegen makes this easier already) and instead just run specific test suites separately.

.github/workflows/test.yml
philprime: > I would prefer we extract that SentrySwiftUI run to its own job

I agree with this 👍 

> If we combine that with using a new test target to run the HTTP tests...

This would work for me too, I do not see much of a difference in two test plans vs two targets, except for less target configuration management as mentioned in the other comment.

philprime:
@philipphofmann any inputs on this PR? We can always postpone it to another time.

philipphofmann:
Do we really need this currently? Are the HTTPTransport tests flaky at the moment?

philprime:
I don't think so, this was our idea a while ago, we can restart if becomes a problem again.

PR #5222, opened 2025-05-13
test: iOS sample app debug menu

armcknight:
Makes the feature flag table view shared in #5199 available to all iOS sample apps. Moved it from its spot in iOS-Swift where a button in the Extras tab would push it to a nav controller, to showing a floating button in an overlay window that presents the list modally (dismissable with a swipe down).

Future work should move the DSN view on each screen of iOS-Swift into this modal as well. (see https://github.com/getsentry/sentry-cocoa/pull/5228)

It is shown in iOS-ObjectiveC but be aware that that app doesn't yet respect all the possible flags. This should be taken care of in #5201 .

#skip-changelog


armcknight:
I believe the remaining CI failures are flakes

PR #5223, opened 2025-05-13
fix(feedback): user feedback widget in SwiftUI

### LINKED ISSUE #5113: feedback: SwiftUI support for widget
### Description

Our current auto injection of the widget doesn't work with SwiftUI, as reported in https://github.com/getsentry/sentry-cocoa/issues/5053

Shipping bring-your-own-button as a workaround, but we can also fix auto injection to inject into the root UIHostingController in a SwiftUI app, while keeping the overlay UIWindow strategy for UIKit apps.

armcknight:
In an attempt to fix #5113, use the correct API to display a floating UIWindow in the SwiftUI app.

~Requires consumers to start the sentry api from a configured scene delegate, as demonstrated in the sample swiftui apps in #5222. This would need to be added to the docs for auto injection, and may warrant a separate API to display the feedback widget (this was originally planned, but i removed it as i'd only implemented auto injection. seems we just identified a need for it) if we want to give consumers the option of starting the SDK as early as possible (in a swiftui app's init method) and be able to display the feedback widget as well.~

Update: The API to manually display a widget was done in https://github.com/getsentry/sentry-cocoa/pull/5236. This PR will fix that functionality in SwiftUI apps. A subsequent PR may attempt to fix the auto-injection situation for SwiftUI apps.

**Current status**: ~a build issue detailed in the linked issue~ this is resolved as of #5524


armcknight:
New UI test passing: https://github.com/getsentry/sentry-cocoa/actions/runs/16009048791/job/45162392977?pr=5223#step:12:799

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
```diff
@@ -42,6 +42,16 @@ import UIKit
...
             widgetConfigBuilder(configuration.widgetConfig)
             validate(configuration.widgetConfig)
+
+            /*
+             * We cannot currently automatically inject a widget into a SwiftUI application, because at the recommended time to start the Sentry SDK (SwiftUIApp.init) there is nowhere to put a UIWindow overlay. SwiftUI apps must currently declare a UIApplicationDelegateAdaptor that returns a UISceneConfiguration, which we can then extract a connected UIScene from into which we can inject a UIWindow.
+             *
+             * At the time this integration is being installed, if there is no UIApplicationDelegate and no connected UIScene, it is very likely we are in a SwiftUI app, but it's possible we could instead be in a UIKit app that has some nonstandard launch procedure or doesn't call SentrySDK.start in a place we expect/recommend, in which case they will need to manually display the widget when they're ready by calling SentrySDK.feedback.showWidget.
+             */
+            if UIApplication.shared.connectedScenes.isEmpty && UIApplication.shared.delegate == nil {
+                return
```
noahsmartin, line 52: I'm surprised codecov is saying this line isn't covered. Shouldn't it be covered by the UI test you added for the SwiftUI app? My understanding is this case is reached for apps written entirely in SwiftUI, so just wanted to double check the test is running/working

armcknight: yeah that should be taken upon calling SentrySDK.start, let me have a quick look, thanks for pointing it out

armcknight: @noahsmartin I think it's because code coverage is explicitly disabled for this sample app's test plan; fixed in https://github.com/getsentry/sentry-cocoa/pull/5544 and then we can rerun this afterwards

armcknight:
rebased on top of https://github.com/getsentry/sentry-cocoa/pull/5544 to get code coverage working

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
armcknight: Well it seems to still be reporting no coverage here, but I verified it locally with a breakpoint and Xcode's own code coverage view ¯\\_(ツ)_/¯:
image
image

PR #5224, opened 2025-05-13
chore: Remove unused Xcode test schemes

philprime:
- The scheme `SentryTests` and `SentryProfilerTests` are unused because the tests are added as test targets to scheme `Sentry`.
- The test plan `Sentry_Base.xctestplan` is configured to run SentryProfilerTests

#skip-changelog


philipphofmann (commented):
CI is green and this
Screenshot 2025-05-13 at 15 28 31

What else do I need 😄. LGTM

PR #5225, opened 2025-05-13
fix: Threading issues in SentryDependencyContainer

philipphofmann:
## :pencil: Checklist


Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -101,191 +139,103 @@ - (instancetype)init
...
-        @synchronized(sentryDependencyContainerLock) {
-            if (_threadInspector == nil) {
-                _threadInspector =
-                    [[SentryThreadInspector alloc] initWithOptions:SentrySDK.options];
-            }
-        }
-    }
-    return _threadInspector;
+    SENTRY_LAZY_INIT(
+        _fileManager, [[SentryFileManager alloc] initWithOptions:SentrySDK.options error:nil]);
```
armcknight, line 175: This `nil` in the constructor is crashing the tests.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -54,6 +53,24 @@
...
+ */
+#define SENTRY_LAZY_INIT(instance, initBlock)                                                      \
+    if (instance == nil) {                                                                         \
+        @synchronized(sentryDependencyContainerDependenciesLock) {                                 \
+            if (instance == nil) {                                                                 \
+                instance = initBlock;                                                              \
+            }                                                                                      \
+        }                                                                                          \
+    }                                                                                              \
+    return instance;
```
armcknight, line 69: Returning inside the macro will make things trickier, for instance, dealing with the inout error parameter in the crashing file manager property init.

armcknight: Proposed the change in https://github.com/getsentry/sentry-cocoa/pull/5246

armcknight (approved):
Generally in favor of the problem being solved here. i left some thoughts for where we might go in the future.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -63,34 +80,55 @@ @interface SentryDependencyContainer ()
...
 
 + (instancetype)sharedInstance
 {
-    return instance;
+    // This synchronization adds around 5% slowdown compared to no synchronization.
+    // As we don't call this method in a tight loop, it's acceptable. Measured with self.measure in
+    // unit tests.
+    @synchronized(sentryDependencyContainerInstanceLock) {
+        return instance;
+    }
```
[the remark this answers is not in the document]
armcknight: Note, answered in https://github.com/getsentry/sentry-cocoa/pull/5225#discussion_r2093588992

Sources/Sentry/SentryDependencyContainer.m
armcknight: Proposed the change in https://github.com/getsentry/sentry-cocoa/pull/5246

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -63,34 +80,55 @@ @interface SentryDependencyContainer ()
...
+        // Why don't we reset all dependencies manually so we can avoid using a lock in
+        // sharedInstance? Good question. This approach comes with the following problems:
+        //
+        // 1. We need a lock for all properties, including the init dependencies.
+        //
+        // 2. When adding a new dependency it is very easy to forget adding it to the reset list
+        //
+        // 3. The lock in sharedInstance only caused around a 5% overhead compared to not using a
+        // lock. Measured with self.measure in unit tests. So not having locks for all the init
+        // dependencies is as efficient as having a lock for the instance in sharedInstance.
```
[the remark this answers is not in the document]
philipphofmann: Yes, I also thought about something like that. W would need to move the static functions from the SentryDependencyContainer to the SentrySDK class, and it would have the responsibility of keeping the DependencyContainer. I didn't want to do it in this PR, because I first wanted to fix the threading issues, and doing that could easily be a bigger refactoring even though it looks small.

Sources/Sentry/SentryDependencyContainer.m
philipphofmann: Thanks for the proposal. I found a different approach that doesn't require moving returning the instance out of the macro https://github.com/getsentry/sentry-cocoa/blob/157c2d3301c5510a1e66e90d42d4a236347e87cf/Sources/Sentry/SentryDependencyContainer.m#L172-L183

I think this approach is also legit. Thanks for opening a PR for the suggestions.

PR #5227, opened 2025-05-13
cleanup integration protocol

noahsmartin:
## :scroll: Description

It's best practice not to import the Swift generated header into an objc header to avoid circular dependencies, as described here: https://developer.apple.com/documentation/swift/importing-swift-into-objective-c#Include-Swift-Classes-in-Objective-C-Headers-Using-Forward-Declarations
(More info here: https://cjwirth.com/tech/circular-references-swift-objc)

This protocol was only ever conformed to by types that also inherited from SentryBaseIntegration. To clean it up a bit I moved the conformance into the base class (which already implemented the required methods), and converted it into an objc header. Since a few header files no longer needed to import the swift generated header, a few other missing imports showed up which I fixed.

Cleans up the code, helps with building using SPM

Built locally

## :pencil: Checklist

#skip-changelog


armcknight (approved):
LGTM. CI failures look like flakes.

philprime (approved):
LGTM

PR #5228, opened 2025-05-13
test: DSN view in debug menu

armcknight:
Previously, I'd added this DSN customization view/flow to a couple screens in iOS-Swift in #4320. However, it is only fixed to those screens, which is a subset of all areas of the app.

This moves the view into the shared lib's new in app debug menu so it's accessible from anywhere in an app that uses the debug menu, which is getting automatically shared to more of the sample apps.

image

#skip-changelog


philprime:
I am not sure if the changes in this PR aren't outdated already. Please update the PR, resolve conflicts, then ping me again to review it.

armcknight:
@philprime this branch is updated now and also includes the fix requested in https://github.com/getsentry/sentry-cocoa/pull/5202#discussion_r2088474799

philprime (approved):
LGTM, thank you for the improvements.

Please double-check the failing UI test, it might be related to the view controller changes

armcknight:
Thanks. Looks like more flakes (although, we really shouldn't show the debug menu for UI tests unless they actually need it, same for feedback widget)

armcknight:
@philipphofmann, FWIW, I'm routinely asked by my profiling backend teammates to send a test profile to an alternate project to validate something. In any case the original functionality predated me, i've just refactored it to improve the UX over time.

philipphofmann:
Ah OK, it's OK to keep it then. I hardly ever use it.

PR #5233, opened 2025-05-14
Remove log swift header

noahsmartin:
## :scroll: Description

Remove swift import from last header file (SentryLog.h). Removing this triggered many missing imports in .m/.mm files which are now added

Same as https://github.com/getsentry/sentry-cocoa/pull/5227 and https://github.com/getsentry/sentry-cocoa/pull/5232 this is to address circular dependencies in swift/objc references and follow best practices from https://developer.apple.com/documentation/swift/importing-swift-into-objective-c#Include-Swift-Classes-in-Objective-C-Headers-Using-Forward-Declarations

Locally building

## :pencil: Checklist

#skip-changelog

### DIFF
--- a/Sources/Sentry/include/SentryLog.h
+++ b/Sources/Sentry/include/SentryLog.h
@@ -1,19 +1,27 @@
 #import "SentryDefines.h"
-#import "SentrySwift.h"
 
-#define SENTRY_LOG(_SENTRY_LOG_LEVEL, ...)                                                         \
-    if ([SentryLog willLogAtLevel:_SENTRY_LOG_LEVEL]) {                                            \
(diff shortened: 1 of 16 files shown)


Sources/Sentry/include/SentryLog.h
```diff
@@ -1,19 +1,22 @@
 #import "SentryDefines.h"
-#import "SentrySwift.h"
```
noahsmartin, line 2: The main change is the removal of this import, because swift isn't being imported in a header file anymore a few other implementation files that were using swift needed to have the import explicitly added

armcknight (approved):
Looks good and I see the logs working locally, :shipit:

Issue #5234, opened 2025-05-14
CI: test changes in hybrid SDK samples


armcknight:
### Description

- [ ] react native
- [ ] flutter
- [ ] .NET

buenaflor:
For Flutter we have some native ios tests in `flutter/example/ios`: 

the steps to build the sample would be
- install flutter:
```yml
- name: Set up Flutter
    uses: subosito/flutter-action@v2
    with:
      channel: stable
```
- clone the repo
- run `flutter pub get` in `flutter/example`

PR #5235, opened 2025-05-14
ci: react native tests

armcknight:
Something I've thought about for a long time. See https://github.com/getsentry/sentry-cocoa/issues/5234

#skip-changelog

### DIFF
--- a/.github/workflows/test-cross-platform.yml
+++ b/.github/workflows/test-cross-platform.yml
@@ -5,54 +5,57 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
+    paths:
+      - '.github/workflows/test-cross-platform.yml'
+      - 'Sources/**'
+      - 'Sentry.podspec'
 
 jobs:
-  test:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


philprime:
@krystofwoldrich could you please review this PR? I believe you might have more insights on how to do this kind of testing with react native

armcknight:
@philprime Sorry, I removed the comments I had stating this, but this is essentially a copy of the [react native ios test workflow](https://github.com/getsentry/sentry-react-native/blob/8dffe6b21d5f2e2c24b65910060ebf149e9fafbf/.github/workflows/native-tests.yml#L19-L53), with a couple sed scripts carrying out the steps outlined in the [CONTRIBUTING.md doc](https://github.com/getsentry/sentry-react-native/blob/main/CONTRIBUTING.md#develop-with-sentry-cocoa) to use the local sentry sdk, so it's essentially exactly what is prescribed by their instructions.

Also note that Kryštof is out for the rest of the week. I was hoping to use this and similar ones for dart/.net to validate Noah's PRs #5232 and #5233.

PR #5238, opened 2025-05-15
Remove unnecessary version check

noahsmartin:
We only support > iOS 10.0 now (and same for the other platforms) so we can delete this conditional and use `clock_gettime_nsec_np` directly now to prevent the extra swift/objc interop

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentrySysctl.m
+++ b/Sources/Sentry/SentrySysctl.m
@@ -1,9 +1,9 @@
 #import "SentrySysctl.h"
 #import "SentryCrashSysCtl.h"
+#import "SentrySwift.h"
 #import "SentryTime.h"
 #include <stdio.h>
 #include <time.h>
-#import "SentrySwift.h"
 
 static NSDate *moduleInitializationTimestamp;
 static uint64_t runtimeInitSystemTimestamp;
--- a/Sources/Sentry/SentrySysctl.m
+++ b/Sources/Sentry/SentrySysctl.m
@@ -3,6 +3,7 @@
 #import "SentryTime.h"
 #include <stdio.h>
 #include <time.h>
+#import "SentrySwift.h"
 
 static NSDate *moduleInitializationTimestamp;
 static uint64_t runtimeInitSystemTimestamp;
@@ -35,7 +36,7 @@ sentryModuleInitializationHook(void)
     // there's no guarantee on whether that or this load method will be called first, the difference
     // in time has been observed to only be on the order of single milliseconds, not significant
     // enough to make a difference in outcomes
-    runtimeInitSystemTimestamp = getAbsoluteTime();
+    runtimeInitSystemTimestamp = [[[SentryDefaultCurrentDateProvider alloc] init] systemTime];
 }
 
 - (NSDate *)runtimeInitTimestamp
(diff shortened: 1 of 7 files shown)


armcknight (commented):
I'm down with removing the availability check, that's great, but not with fragmenting the system time access–we use `getAbsoluteTime` in many places.

noahsmartin:
Updated to move the other call sites to the swift implementation

Sources/Sentry/SentrySysctl.m
```diff
@@ -35,7 +36,7 @@ + (void)load
     // there's no guarantee on whether that or this load method will be called first, the difference
     // in time has been observed to only be on the order of single milliseconds, not significant
     // enough to make a difference in outcomes
-    runtimeInitSystemTimestamp = getAbsoluteTime();
+    runtimeInitSystemTimestamp = [[[SentryDefaultCurrentDateProvider alloc] init] systemTime];
```
armcknight, line 39: We'll want a static access to `clock_gettime_nsec_np` since we want to do as little work as possible in the launch profile codepath. I see this as being a static method in SentryDefaultCurrentDateProvider that since there's really no underlying state anyways. This way, TestCurrentDateProvider could still override that for testing purposes (i haven't fully through this through, but it would need to provide access to the underlying state stored in the test mock subclass instance stored in the instance held by SentryDependencyContainer).

Tests/SentryTests/Networking/SentryHttpTransportTests.swift
```diff
@@ -833,9 +833,9 @@ class SentryHttpTransportTests: XCTestCase {
         fixture.requestManager.responseDelay = fixture.flushTimeout + 0.2
         
         SentryLog.withoutLogs {
-            let beforeFlush = getAbsoluteTime()
+            let beforeFlush = SentryDefaultCurrentDateProvider().systemTime()
```
armcknight, line 836: All the test usages should access SentryDependencyContainer.sharedInstance().dateProvider, as it often holds a mocked subclass instance of SentryCurrentDateProvider, TestCurrentDateProvider. Some tests don't currently use a mock, like SentrySamplingProfilerTests above, but this class (SentryHttpTransportTests) does, see https://github.com/getsentry/sentry-cocoa/blob/ba1994d08ceb0638ff37a55e1c5ff194b07ea410/Tests/SentryTests/Networking/SentryHttpTransportTests.swift#L20 and https://github.com/getsentry/sentry-cocoa/blob/ba1994d08ceb0638ff37a55e1c5ff194b07ea410/Tests/SentryTests/Networking/SentryHttpTransportTests.swift#L58

I'd day for consistency let's use that dep container instance in all tests.

Still, thanks for replacing the bare invocations of `getAbsoluteTime()`, that was an oversight to ever use that in the tests and probably are leftover from before we delivered the system time via SentryCurrentDateProvider 🙏🏻

armcknight (commented):
We're getting closer, just need to use the right pathways to the now better encapsulated function!

Tests/SentryTests/Networking/SentryHttpTransportTests.swift
noahsmartin: Wasn't totally sure about this, these tests appear to be ensuring certain functions run fast enough (the thing that is test is the time different) so you wouldn't want that to be mocked right? I converted everything to the static Swift function, but left these non-mocked, lmk what you think!

armcknight: You're absolutely right! Now I just wonder why we didn't just use XCTestExpectation with a `wait` call. Honestly even that I think is inappropriate for a unit test. These should be performance tests. Oh well, I agree they're fine as-is with your changes!

armcknight (approved):
LGTM, the tests look like flakes

PR #5242, opened 2025-05-16
fix: Add persistence for context to fix watchdog termination events

philprime:
## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -113,6 +113,11 @@ _non_thread_safe_removeFileAtPath(NSString *path)
 @property (nonatomic, assign) NSUInteger maxEnvelopes;
 @property (nonatomic, weak) id<SentryFileManagerDelegate> delegate;
 
+@property (nonatomic, copy) NSString *contextFilePathOne;
+@property (nonatomic, copy) NSString *contextFilePathTwo;
+@property (nonatomic, copy) NSString *previousContextFilePathOne;
+@property (nonatomic, copy) NSString *previousContextFilePathTwo;
+
 @end
 
 @implementation SentryFileManager
@@ -184,6 +189,13 @@ _non_thread_safe_removeFileAtPath(NSString *path)
     self.appHangEventFilePath =
         [self.sentryPath stringByAppendingPathComponent:@"app.hang.event.json"];
     self.envelopesPath = [self.sentryPath stringByAppendingPathComponent:EnvelopesPathComponent];
+
+    self.contextFilePathOne = [self.sentryPath stringByAppendingPathComponent:@"contexts.1.state"];
+    self.contextFilePathTwo = [self.sentryPath stringByAppendingPathComponent:@"contexts.2.state"];
+    self.previousContextFilePathOne =
+        [self.sentryPath stringByAppendingPathComponent:@"previous.contexts.1.state"];
+    self.previousContextFilePathTwo =
+        [self.sentryPath stringByAppendingPathComponent:@"previous.contexts.2.state"];
 }
 
 - (void)setDelegate:(id<SentryFileManagerDelegate>)delegate
@@ -499,6 +511,51 @@ _non_thread_safe_removeFileAtPath(NSString *path)
     }
 
     return breadcrumbs;
+}
+
+#pragma mark - Contexts
+
+- (void)moveContextFileToPreviousContextFile
+{
+    @synchronized(self.contextFilePathOne) {
+        [self moveState:self.contextFilePathOne toPreviousState:self.previousContextFilePathOne];
+        [self moveState:self.contextFilePathTwo toPreviousState:self.previousContextFilePathTwo];
+    }
+}
+
+- (NSDictionary<NSString *, NSDictionary<NSString *, id> *> *_Nullable)readPreviousContext;
+{
+    SENTRY_LOG_DEBUG(@"Reading previous context");
+
+    if ([[NSFileManager defaultManager] fileExistsAtPath:self.previousContextFilePathOne]) {
+        SENTRY_LOG_DEBUG(@"Reading data of previous context file one");
+        NSData *data = [NSData dataWithContentsOfFile:self.previousContextFilePathOne];
+        NSError *error;
+        NSDictionary<NSString *, NSDictionary<NSString *, id> *> *_Nullable context =
+            [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+        if (error) {
+            SENTRY_LOG_ERROR(@"Error deserializing context: %@", error);
+            return nil;
+        }
+        SENTRY_LOG_DEBUG(@"Previous context: %@", context);
+        return context;
+    }
+
+    if ([[NSFileManager defaultManager] fileExistsAtPath:self.previousContextFilePathTwo]) {
+        SENTRY_LOG_DEBUG(@"Reading data of previous context file two");
+        NSData *data = [NSData dataWithContentsOfFile:self.previousContextFilePathTwo];
+        NSError *error;
+        NSDictionary<NSString *, NSDictionary<NSString *, id> *> *_Nullable context =
+            [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+        if (error) {
+            SENTRY_LOG_ERROR(@"Error deserializing context: %@", error);
+            return nil;
+        }
+        SENTRY_LOG_DEBUG(@"Previous context: %@", context);
+        return context;
+    }
+
+    return nil;
 }
 
 #pragma mark - TimezoneOffset
--- a/Sources/Sentry/SentryWatchdogTerminationScopeObserver.m
+++ b/Sources/Sentry/SentryWatchdogTerminationScopeObserver.m
@@ -9,10 +9,14 @@
 @interface SentryWatchdogTerminationScopeObserver ()
 
 @property (strong, nonatomic) SentryFileManager *fileManager;
-@property (strong, nonatomic) NSFileHandle *fileHandle;
+
+@property (strong, nonatomic) NSFileHandle *fileHandleBreadcrumbs;
+@property (strong, nonatomic) NSString *activeFilePathBreadcrumbs;
 @property (nonatomic) NSInteger maxBreadcrumbs;
 @property (nonatomic) NSInteger breadcrumbCounter;
-@property (strong, nonatomic) NSString *activeFilePath;
+
+@property (nonatomic, strong) NSFileHandle *fileHandleContext;
+@property (strong, nonatomic) NSString *activeFilePathContext;
 
 @end
 
@@ -22,9 +26,10 @@
                            fileManager:(SentryFileManager *)fileManager
 {
     if (self = [super init]) {
-        self.maxBreadcrumbs = maxBreadcrumbs;
         self.fileManager = fileManager;
+
         self.breadcrumbCounter = 0;
+        self.maxBreadcrumbs = maxBreadcrumbs;
 
         [self switchFileHandle];
     }
@@ -34,55 +39,106 @@
 
 - (void)dealloc
 {
-    [self.fileHandle closeFile];
+    [self.fileHandleBreadcrumbs closeFile];
+    [self.fileHandleContext closeFile];
 }
 
 // PRAGMA MARK: - Helper methods
 
 - (void)deleteFiles
 {
-    [self.fileHandle closeFile];
-    self.fileHandle = nil;
-    self.activeFilePath = nil;
+    [self deleteFilesBreadcrumbs];
+    [self deleteFilesContexts];
+}
+
+- (void)deleteFilesBreadcrumbs
+{
+    [self.fileHandleBreadcrumbs closeFile];
+    self.fileHandleBreadcrumbs = nil;
+    self.activeFilePathBreadcrumbs = nil;
     self.breadcrumbCounter = 0;
 
     [self.fileManager removeFileAtPath:self.fileManager.breadcrumbsFilePathOne];
     [self.fileManager removeFileAtPath:self.fileManager.breadcrumbsFilePathTwo];
 }
 
+- (void)deleteFilesContexts
+{
+    [self.fileHandleContext closeFile];
+    self.fileHandleContext = nil;
+    self.activeFilePathContext = nil;
+
+    [self.fileManager removeFileAtPath:self.fileManager.contextFilePathOne];
+    [self.fileManager removeFileAtPath:self.fileManager.contextFilePathTwo];
+}
+
 - (void)switchFileHandle
 {
-    if ([self.activeFilePath isEqualToString:self.fileManager.breadcrumbsFilePathOne]) {
-        self.activeFilePath = self.fileManager.breadcrumbsFilePathTwo;
+    [self switchFileHandleBreadcrumbs];
+    [self switchFileHandleContexts];
+}
+
+- (void)switchFileHandleBreadcrumbs
+{
+    if ([self.activeFilePathBreadcrumbs isEqualToString:self.fileManager.breadcrumbsFilePathOne]) {
+        self.activeFilePathBreadcrumbs = self.fileManager.breadcrumbsFilePathTwo;
     } else {
-        self.activeFilePath = self.fileManager.breadcrumbsFilePathOne;
+        self.activeFilePathBreadcrumbs = self.fileManager.breadcrumbsFilePathOne;
     }
 
     // Close the current filehandle (if any)
-    [self.fileHandle closeFile];
+    [self.fileHandleBreadcrumbs closeFile];
 
     // Create a fresh file for the new active path
-    [self.fileManager removeFileAtPath:self.activeFilePath];
-    [[NSFileManager defaultManager] createFileAtPath:self.activeFilePath
+    [self.fileManager removeFileAtPath:self.activeFilePathBreadcrumbs];
+    [[NSFileManager defaultManager] createFileAtPath:self.activeFilePathBreadcrumbs
                                             contents:nil
                                           attributes:nil];
 
     // Open the file for writing
-    self.fileHandle = [NSFileHandle fileHandleForWritingAtPath:self.activeFilePath];
+    self.fileHandleBreadcrumbs =
+        [NSFileHandle fileHandleForWritingAtPath:self.activeFilePathBreadcrumbs];
 
-    if (!self.fileHandle) {
-        SENTRY_LOG_ERROR(@"Couldn't open file handle for %@", self.activeFilePath);
+    if (!self.fileHandleBreadcrumbs) {
+        SENTRY_LOG_ERROR(@"Couldn't open file handle for %@", self.activeFilePathBreadcrumbs);
     }
 }
 
-- (void)store:(NSData *)data
+- (void)switchFileHandleContexts
+{
+    SENTRY_LOG_DEBUG(@"Switching file handle for context");
+    if ([self.activeFilePathContext isEqualToString:self.fileManager.contextFilePathOne]) {
+        self.activeFilePathContext = self.fileManager.contextFilePathTwo;
+    } else {
+        self.activeFilePathContext = self.fileManager.contextFilePathOne;
+    }
+    SENTRY_LOG_DEBUG(@"New active file path for context: %@", self.activeFilePathContext);
+
+    // Close the current filehandle (if any)
+    [self.fileHandleContext closeFile];
+
+    // Create a fresh file for the new active path
+    [self.fileManager removeFileAtPath:self.activeFilePathContext];
+    [[NSFileManager defaultManager] createFileAtPath:self.activeFilePathContext
+                                            contents:nil
+                                          attributes:nil];
+
+    // Open the file for writing
+    self.fileHandleContext = [NSFileHandle fileHandleForWritingAtPath:self.activeFilePathContext];
+
+    if (!self.fileHandleContext) {
+        SENTRY_LOG_ERROR(@"Couldn't open file handle for %@", self.activeFilePathContext);
+    }
+}
+
+- (void)storeBreadcrumb:(NSData *)data
 {
     unsigned long long fileSize;
     @try {
-        fileSize = [self.fileHandle seekToEndOfFile];
+        fileSize = [self.fileHandleBreadcrumbs seekToEndOfFile];
 
-        [self.fileHandle writeData:data];
-        [self.fileHandle writeData:[@"\n" dataUsingEncoding:NSASCIIStringEncoding]];
+        [self.fileHandleBreadcrumbs writeData:data];
+        [self.fileHandleBreadcrumbs writeData:[@"\n" dataUsingEncoding:NSASCIIStringEncoding]];
 
         self.breadcrumbCounter += 1;
     } @catch (NSException *exception) {
@@ -93,6 +149,22 @@
             [self switchFileHandle];
             self.breadcrumbCounter = 0;
         }
+    }
+}
+
+- (void)storeContext:(NSData *)data
+{
+    @try {
+        SENTRY_LOG_DEBUG(@"Storing context data: %@", data);
+        // Override the entire file content by starting from the beginning
+        [self.fileHandleContext seekToFileOffset:0];
+        [self.fileHandleContext writeData:data];
+
+        // Truncate the file to the current size
+        [self.fileHandleContext truncateFileAtOffset:data.length];
+        SENTRY_LOG_DEBUG(@"Written context data to file: %@", self.activeFilePathContext);
+    } @catch (NSException *exception) {
+        SENTRY_LOG_ERROR(@"Error while writing data to context file: %@", exception.description);
     }
 }
 
@@ -110,9 +182,9 @@
 
     if (error) {
         SENTRY_LOG_ERROR(@"Error serializing breadcrumb: %@", error);
-    } else {
-        [self store:jsonData];
+        return;
     }
+    [self storeBreadcrumb:jsonData];
 }
 
 - (void)clear
@@ -128,47 +200,61 @@
 
 - (void)setContext:(nullable NSDictionary<NSString *, id> *)context
 {
-    // Left blank on purpose
+    SENTRY_LOG_DEBUG(@"Setting context: %@", context);
+    if (![NSJSONSerialization isValidJSONObject:context]) {
+        SENTRY_LOG_ERROR(@"Context is not a valid JSON object: %@", context);
+        return;
+    }
+
+    NSError *error;
+    NSData *jsonData = [NSJSONSerialization dataWithJSONObject:context options:0 error:&error];
+
+    if (error) {
+        SENTRY_LOG_ERROR(@"Error serializing context: %@", error);
+        return;
+    }
+
+    [self storeContext:jsonData];
 }
 
 - (void)setDist:(nullable NSString *)dist
 {
-    // Left blank on purpose
+    SENTRY_LOG_DEBUG(@"Setting dist: %@", dist);
 }
 
 - (void)setEnvironment:(nullable NSString *)environment
 {
-    // Left blank on purpose
+    SENTRY_LOG_DEBUG(@"Setting environment: %@", environment);
 }
 
 - (void)setExtras:(nullable NSDictionary<NSString *, id> *)extras
 {
-    // Left blank on purpose
+    SENTRY_LOG_DEBUG(@"Setting extras: %@", extras);
 }
 
 - (void)setFingerprint:(nullable NSArray<NSString *> *)fingerprint
 {
-    // Left blank on purpose
+    SENTRY_LOG_DEBUG(@"Setting fingerprint: %@", fingerprint);
 }
 
 - (void)setLevel:(enum SentryLevel)level
 {
-    // Left blank on purpose
+    SENTRY_LOG_DEBUG(@"Setting level: %@", @(level));
 }
 
 - (void)setTags:(nullable NSDictionary<NSString *, NSString *> *)tags
 {
-    // Left blank on purpose
+    SENTRY_LOG_DEBUG(@"Setting tags: %@", tags);
 }
 
 - (void)setUser:(nullable SentryUser *)user
 {
-    // Left blank on purpose
+    SENTRY_LOG_DEBUG(@"Setting user: %@", user);
 }
 
 - (void)setTraceContext:(nullable NSDictionary<NSString *, id> *)traceContext
 {
-    // Left blank on purpose
+    SENTRY_LOG_DEBUG(@"Setting trace context: %@", traceContext);
 }
 
 @end
--- a/Sources/Sentry/SentryWatchdogTerminationTracker.m
+++ b/Sources/Sentry/SentryWatchdogTerminationTracker.m
@@ -69,6 +69,13 @@
                 event.timestamp = sentry_fromIso8601String(timestampIso8601String);
             }
 
+            // Load the previous context from disk
+            NSDictionary<NSString *, NSDictionary<NSString *, id> *> *_Nullable context =
+                [self.fileManager readPreviousContext];
+            if (context != nil) {
+                event.context = context;
+            }
+
             SentryException *exception =
                 [[SentryException alloc] initWithValue:SentryWatchdogTerminationExceptionValue
                                                   type:SentryWatchdogTerminationExceptionType];
--- a/Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
+++ b/Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
@@ -85,8 +85,14 @@ NS_ASSUME_NONNULL_BEGIN
             initWithMaxBreadcrumbs:options.maxBreadcrumbs
                        fileManager:[[[SentrySDK currentHub] getClient] fileManager]];
 
-    [SentrySDK.currentHub configureScope:^(
-        SentryScope *_Nonnull outerScope) { [outerScope addObserver:scopeObserver]; }];
+    [SentrySDK.currentHub configureScope:^(SentryScope *_Nonnull outerScope) {
+        // Add the observer to the scope so that it can be notified when the scope changes.
+        [outerScope addObserver:scopeObserver];
+
+        // Sync the current context to the observer to capture context modifications that happened
+        // before installation.
+        [scopeObserver setContext:outerScope.contextDictionary];
+    }];
 
     return YES;
 }
--- a/Sources/Sentry/SentryWatchdogTerminationLogic.m
+++ b/Sources/Sentry/SentryWatchdogTerminationLogic.m
@@ -72,9 +72,9 @@
 
     // Restarting the app in development is a termination we can't catch and would falsely
     // report watchdog termiations.
-    if (previousAppState.isDebugging) {
-        return NO;
-    }
+    //    if (previousAppState.isDebugging) {
+    //        return NO;
+    //    }
 
     // The app was terminated normally
     if (previousAppState.wasTerminated) {
@@ -98,9 +98,9 @@
         return NO;
     }
 
-    if (previousAppState.isANROngoing) {
-        return NO;
-    }
+    //    if (previousAppState.isANROngoing) {
+    //        return NO;
+    //    }
 
     // When calling SentrySDK.start twice we would wrongly report a Watchdog Termination. We can
     // only report a Watchdog Termination when the SDK is started the first time.
--- a/Sources/Sentry/include/SentryFileManager.h
+++ b/Sources/Sentry/include/SentryFileManager.h
@@ -28,10 +28,16 @@ SENTRY_NO_INIT
 
 @property (nonatomic, readonly) NSString *basePath;
 @property (nonatomic, readonly) NSString *sentryPath;
+
 @property (nonatomic, readonly) NSString *breadcrumbsFilePathOne;
 @property (nonatomic, readonly) NSString *breadcrumbsFilePathTwo;
 @property (nonatomic, readonly) NSString *previousBreadcrumbsFilePathOne;
 @property (nonatomic, readonly) NSString *previousBreadcrumbsFilePathTwo;
+
+@property (nonatomic, readonly) NSString *contextFilePathOne;
+@property (nonatomic, readonly) NSString *contextFilePathTwo;
+@property (nonatomic, readonly) NSString *previousContextFilePathOne;
+@property (nonatomic, readonly) NSString *previousContextFilePathTwo;
 
 - (nullable instancetype)initWithOptions:(SentryOptions *)options error:(NSError **)error;
 
@@ -96,6 +102,10 @@ SENTRY_NO_INIT
 #pragma mark - Breadcrumbs
 - (void)moveBreadcrumbsToPreviousBreadcrumbs;
 - (NSArray *)readPreviousBreadcrumbs;
+
+#pragma mark - Contexts
+- (void)moveContextFileToPreviousContextFile;
+- (NSDictionary<NSString *, NSDictionary<NSString *, id> *> *_Nullable)readPreviousContext;
 
 #pragma mark - TimezoneOffset
 - (NSNumber *_Nullable)readTimezoneOffset;


Sources/Sentry/SentryWatchdogTerminationScopeObserver.m
```diff
@@ -96,6 +152,22 @@ - (void)store:(NSData *)data
     }
 }
 
+- (void)storeContext:(NSData *)data
+{
+    @try {
+        SENTRY_LOG_DEBUG(@"Storing context data: %@", data);
+        // Override the entire file content by starting from the beginning
+        [self.fileHandleContext seekToFileOffset:0];
+        [self.fileHandleContext writeData:data];
```
philipphofmann, line 161: As the context doesn't change that frequently, I think we could consider storing that data on a BG thread, so we might not need using a file handle. Simple file IO operations could be enough.

philprime: Works for me, was considering this too, thought it might be safer to have a file handle open, so we do not have to do file IO and just write data over time.

Do we need the swap files and write it twice then?

Sources/Sentry/include/HybridPublic/SentryDependencyContainer.h
```diff
@@ -101,6 +104,7 @@ SENTRY_NO_INIT
 @property (nonatomic, strong) SentryThreadInspector *threadInspector;
 @property (nonatomic, strong) SentryFileIOTracker *fileIOTracker;
 @property (nonatomic, strong) SentryCrash *crashReporter;
+@property (nonatomic, strong) SentryScopeContextPersistentStore *scopeContextStore;
```
philipphofmann, line 107: `m`: Please add the new properties to `testThreadSafety_ResetAndAccessProperties` to ensure testing the thread safety.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -347,4 +348,42 @@ - (SentryMXManager *)metricKitManager SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOC
 
 #endif // SENTRY_HAS_METRIC_KIT
 
+#if SENTRY_HAS_REACHABILITY
+- (SentryReachability *)reachability SENTRY_DISABLE_THREAD_SANITIZER(
+    "double-checked lock produce false alarms")
```
philipphofmann, line 353: `l`: Please use `SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOCK` instead.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -347,4 +348,42 @@ - (SentryMXManager *)metricKitManager SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOC
...
+        [[SentryScopeContextPersistentStore alloc] initWithFileManager:self.fileManager]);
+}
+
+#if SENTRY_HAS_UIKIT
+- (SentryWatchdogTerminationBreadcrumbProcessor *)
+    getWatchdogTerminationBreadcrumbProcessorWithMaxBreadcrumbs:(NSInteger)maxBreadcrumbs
+{
+    // This method is only a factory, therefore do not keep a reference.
+    // The processor will be created each time it is needed.
+    @synchronized(sentryDependencyContainerInstanceLock) {
```
philipphofmann, line 371: `m`: I think you could ditch the `@synchronized` then or consider not adding the method at all.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    init(fileManager: SentryFileManager) {
+        self.fileManager = fileManager
+    }
+
+    // MARK: - Context
+
+    func readPreviousContextFromDisk() -> [String: [String: Any]]? {
+        let fm = FileManager.default
+        guard fm.fileExists(atPath: contextFileURL.path) else {
```
philipphofmann, line 15: `h`: That should be the `previousContextFilePath` if I'm not mistaken. We should add a test for this, because a test should fail cause we're reading the current `contextFileURL` instead of the previous one.

Sources/Sentry/SentryFileManager.m
```diff
@@ -501,6 +508,16 @@ - (NSArray *)readPreviousBreadcrumbs
...
 
+#pragma mark - Contexts
+
+- (void)moveContextFileToPreviousContextFile
+{
+    SENTRY_LOG_DEBUG(@"Moving context file to previous context file");
+    @synchronized(self.contextFilePath) {
+        [self moveState:self.contextFilePath toPreviousState:self.previousContextFilePath];
+    }
+}
```
philipphofmann, line 519: `m`: Can we add a unit test to the `SentryFileManagerTests` for this method, please?

Sources/Sentry/SentrySDK.m
```diff
@@ -240,6 +240,7 @@ + (void)startWithOptions:(SentryOptions *)options
     SentryClient *newClient = [[SentryClient alloc] initWithOptions:options];
     [newClient.fileManager moveAppStateToPreviousAppState];
     [newClient.fileManager moveBreadcrumbsToPreviousBreadcrumbs];
+    [newClient.fileManager moveContextFileToPreviousContextFile];
```
philipphofmann, line 243: `m`: It would be valuable to add a test in SentrySDKTests to ensure we're actually calling this here.

Sources/Sentry/SentryWatchdogTerminationTracker.m
```diff
@@ -97,6 +81,46 @@ - (void)start
...
+    }
+}
+
+- (void)addContextToEvent:(SentryEvent *)event
+{
+    // Load the previous context from disk, or create an empty one if it doesn't exist
+    NSDictionary<NSString *, NSDictionary<NSString *, id> *> *previousContext =
+        [self.scopeContextStore readPreviousContextFromDisk];
+    NSMutableDictionary *context =
+        [[NSMutableDictionary alloc] initWithDictionary:previousContext ?: @{}];
```
philipphofmann, line 111: `l`: `readPreviousContextFromDisk` could return an empty dict instead of nil. Then we could simplify the logic here.

Sources/Sentry/SentrySDK.m
philipphofmann, line 243: `m`: This adds another fileIO operation on the calling thread when starting the SDK. I know we simply follow the same approach, but I think we should consider moving all these operations do a BG thread, but not in this PR. WDTY?

Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
```diff
@@ -80,13 +84,26 @@ - (BOOL)installWithOptions:(SentryOptions *)options
...
+                       contextProcessor:contextProcessor];
+
+    [SentrySDK.currentHub configureScope:^(SentryScope *_Nonnull outerScope) {
+        // Add the observer to the scope so that it can be notified when the scope changes.
+        [outerScope addObserver:scopeObserver];
 
-    [SentrySDK.currentHub configureScope:^(
-        SentryScope *_Nonnull outerScope) { [outerScope addObserver:scopeObserver]; }];
+        // Sync the current context to the observer to capture context modifications that happened
+        // before installation.
```
philipphofmann, line 104: Thanks for the comment.

Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
```diff
@@ -80,13 +84,26 @@ - (BOOL)installWithOptions:(SentryOptions *)options
...
+        [scopeObserver setContext:outerScope.contextDictionary];
```
philipphofmann, line 105: `m`: If I'm not mistaken, no tests validate this functionality. It's a bit suspicious that we have no changes in `SentryWatchdogTerminationIntegrationTests`.

Sources/Swift/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationContextProcessor.swift
```diff
@@ -0,0 +1,42 @@
...
+    init(
+        withDispatchQueueWrapper dispatchQueueWrapper: SentryDispatchQueueWrapper,
+        scopeContextStore: SentryScopeContextPersistentStore
+    ) {
+        self.dispatchQueueWrapper = dispatchQueueWrapper
+        self.scopeContextStore = scopeContextStore
+
+        super.init()
+
+        clear()
```
philipphofmann, line 19: `m`: I don't think we need to call `clear` in init actually, because we anyways overwrite the context here https://github.com/getsentry/sentry-cocoa/blob/76fc6d930d0da34f652ff03eeae6e1d14ca3f072/Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m#L103-L105

This an extra file IO operation on the calling thread, and I'd like to avoid it.

If we want to call clear I would rather call it in `SentryWatchdogTerminationTrackingIntegration`, cause doing that in init seems a bit strange to me.

philipphofmann: Ah I think I get why we need it, `setContext` happens async, so we could end up in a dirty state.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
```diff
@@ -0,0 +1,83 @@
+@_implementationOnly import _SentryPrivate
+
+@objcMembers
+class SentryScopeContextPersistentStore: NSObject {
```
philipphofmann, line 4: `m`: Codecov complains rightfully about the test coverage of this class. Can we maybe add some unit tests, covering the edge cases?

philipphofmann (commented):
I added a couple of comments. Thanks for doing this 💪

philipphofmann:
> I believe the additional file is fine for now and can be refactored again at a future point.

Once we also add dist, user, fingerprint, tags, etc. I think we can refactor the solution a bit to reuse the logic. I don't think we need one extra file for each of them. We could combine a couple of properties into one file.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Add persistence for scope context missing in watchdog termination events (#5242)
```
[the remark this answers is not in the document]
armcknight: Sell the experience, not the implementation

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
```diff
@@ -0,0 +1,83 @@
...
+            return nil
+        }
+    }
+
+    func writeContextToDisk(context: [String: [String: Any]]) {
+        guard let data = encode(context: context) else {
+            return
+        }
+        do {
+            try data.write(to: contextFileURL, options: .atomic)
```
armcknight, line 32: Where does the path to this file get created? If the sdk is started with a custom cache path that's a deeply nested path where none of the directories are on disk yet, does this still work? Because `-[SentryFileManager writeData:toPath:]` creates such directory paths on the fly, but since we're not calling that here (which I'd recommend anyways) I don't think this will work in the general case.

Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
```diff
@@ -64,12 +65,15 @@ - (BOOL)installWithOptions:(SentryOptions *)options
         [[SentryWatchdogTerminationLogic alloc] initWithOptions:options
                                                    crashAdapter:crashWrapper
                                                 appStateManager:appStateManager];
+    SentryScopeContextPersistentStore *scopeContextStore =
+        [[SentryScopeContextPersistentStore alloc] initWithFileManager:fileManager];
```
armcknight, line 69: Why aren't we using the instance managed by `SentryDependencyContainer` here, that is injected into init the SDC-managed `SentryWatchdogTerminationContextProcessor`? Doesn't that mean we'll have two instances competing for access to the same `contextFileURL`?

Sources/Sentry/include/SentryWatchdogTerminationTracker.h
```diff
@@ -26,7 +27,8 @@ SENTRY_NO_INIT
        watchdogTerminationLogic:(SentryWatchdogTerminationLogic *)watchdogTerminationLogic
                 appStateManager:(SentryAppStateManager *)appStateManager
            dispatchQueueWrapper:(SentryDispatchQueueWrapper *)dispatchQueueWrapper
-                    fileManager:(SentryFileManager *)fileManager;
+                    fileManager:(SentryFileManager *)fileManager
+              scopeContextStore:(SentryScopeContextPersistentStore *)serialization;
```
armcknight, line 31:
```suggestion
              scopeContextStore:(SentryScopeContextPersistentStore *)scopeContextStore;
```

Sources/Sentry/SentryWatchdogTerminationTracker.m
```diff
@@ -33,13 +34,15 @@ - (instancetype)initWithOptions:(SentryOptions *)options
...
                     fileManager:(SentryFileManager *)fileManager
+              scopeContextStore:(SentryScopeContextPersistentStore *)serialization
 {
     if (self = [super init]) {
         self.options = options;
         self.watchdogTerminationLogic = watchdogTerminationLogic;
         self.appStateManager = appStateManager;
         self.dispatchQueue = dispatchQueueWrapper;
         self.fileManager = fileManager;
+        self.scopeContextStore = serialization;
```
armcknight, line 45:
```suggestion
              scopeContextStore:(SentryScopeContextPersistentStore *)scopeContextStore
{
    if (self = [super init]) {
        self.options = options;
        self.watchdogTerminationLogic = watchdogTerminationLogic;
        self.appStateManager = appStateManager;
        self.dispatchQueue = dispatchQueueWrapper;
        self.fileManager = fileManager;
        self.scopeContextStore = scopeContextStore;
```

Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
armcknight: Since `SentryWatchdogTerminationTracker` has a different dispatch queue wrapper instance (`io.sentry.watchdog-termination-tracker`) than  the one in `SentryWatchdogTerminationContextProcessor` (`io.sentry.watchdog-termination-tracking.context-processor`), and none of the operations in `SentryScopeContextPersistentStore` are thread safe, there could be issues here.

armcknight (commented):
I haven't been able to go through this with a fine-toothed comb but these are some things that jump out at me.

Sources/Sentry/SentryDependencyContainer.m
philprime: Why remove it? You (the previous maintainers) introduced the `SentryDispatchFactory` also as a factory to create instances instead of `SentryDispatchQueueWrapper`

Sources/Sentry/SentryWatchdogTerminationTracker.m
philprime: Separate of concern. For the `readPreviousContextFromDisk` an empty context and a `nil` context is not the same.

Sources/Sentry/SentrySDK.m
philprime: I don't think we can do this, because it should be a synchronous call during start so that the `contextFilePath` is ready to be used by the SDK. If we do this async, it could happen that the file is overwritten with new data, before it is moved to the `previousContextFilePath`.

Sources/Swift/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationContextProcessor.swift
philprime: The `SentryWatchdogTerminationBreadcrumbProcessor` was my reference here. It is calling `switchFileHandle` on `init` which also removes and creates a new file for the active path. This is the same logic but written differently.

Sources/Swift/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationContextProcessor.swift
```diff
@@ -0,0 +1,42 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+@objcMembers
+class SentryWatchdogTerminationContextProcessor: NSObject {
```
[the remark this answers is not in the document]
philprime: This class is new and we should prefer Swift over ObjC for our long-term strategy, but I did not want to rewrite the `SentryWatchdogTerminationBreadcrumbProcessor` which is just a refactoring of the already existing ObjC code

Sources/Sentry/include/HybridPublic/SentryDependencyContainer.h
```diff
@@ -38,6 +39,8 @@
 @class SentryUIApplication;
 @class SentryViewHierarchy;
 @class SentryUIViewControllerPerformanceTracker;
+@class SentryWatchdogTerminationContextProcessor;
+@class SentryWatchdogTerminationBreadcrumbProcessor;
```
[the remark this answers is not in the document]
philprime: Right now it is only used for the watchdog termination so I also don't see much reason to _not_ couple it with that. IMHO renaming classes like that looses context on their purpose.

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
```diff
@@ -0,0 +1,125 @@
...
+        return;
+    }
+
+    NSError *error;
+    NSData *jsonData = [NSJSONSerialization dataWithJSONObject:crumb options:0 error:&error];
+
+    if (error) {
+        SENTRY_LOG_ERROR(@"Error serializing breadcrumb: %@", error);
+        return;
+    }
```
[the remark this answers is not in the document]
philprime: Thanks for pointing this out, I didn't see that class. I'll see if I can use it for this purpose.

Tests/SentryTests/Integrations/WatchdogTerminations/SentryWatchdogTerminationsScopeObserverTests.swift
```diff
@@ -1,174 +1,140 @@
...
-        }
-
-        func testWritingToFullFileSystem() throws {
-            let breadcrumb = try XCTUnwrap(fixture.breadcrumb.serialize() as? [String: String])
-
-            sut.addSerializedBreadcrumb(breadcrumb)
+        // -- Assert --
+        XCTAssertEqual(fixture.contextProcessor.setContextInvocations.count, 1)
+        let invocation = try XCTUnwrap(fixture.contextProcessor.setContextInvocations.first)
+        // Use NSDictionary to erase the type information and compare the dictionaries
```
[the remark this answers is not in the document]
philprime: Not sure if I understand this PR comment. I see unit tests as isolated units, so when one fails I want jump there and see what's going on. That's why I add comments to each test even if it applies to all of them. If a comment is somewhere else in the file, high chance to miss it.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
philprime: I agree, thanks for pointing this out.

Yes I'll add more tests

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
philprime: `contextFileURL` is a an alias/wrapper for `fileManager.contextFilePath` (see end of file). The `contextFilePath` is initialized in `SentryFileManager` like every other path based on the `sentryPath` which is backed by the `options.cacheDirectoryPath` and dsn hash.

I'll see if I can reuse logic of SentryFileManager (if it's not too specific for other use cases, I do not want to pollute new logic with old one).

Tests/SentryTests/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationContextProcessorTests.swift
```diff
@@ -0,0 +1,202 @@
...
+        // -- Act --
+        sut.setContext(fixture.context)
+        sut = nil
+
+        // Execute the block after the processor is deallocated
+        fixture.dispatchQueueWrapper.invokeLastDispatchAsync()
+
+        // -- Assert --
+        // This assertion is a best-effort check to see if the block was executed, as there is not other
+        // mechanism to assert this case
```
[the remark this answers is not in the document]
philprime: I renamed the test and added a comment for more details 👍

SentryTestUtilsTests/TestContstantTests.swift
```diff
@@ -0,0 +1,38 @@
...
+        // -- Act --
+        let dsn = TestConstants.dsnForTestCase(type: FooTests.self)
+
+        // -- Assert --
+        XCTAssertEqual(dsn, "https://FooTests:password@app.getsentry.com/12345")
+    }
+
+    func testDsnForTestCase_externalFrameworkObjectiveCClass_shouldUseTypeNameInDSN() {
+        // -- Arrange --
+        class FooTests: XCTestCase {}
```
philipphofmann, line 19: `l`: FooTests isn't used here. Same applies for testDsnForTestCase_externalFrameworkSwiftClass_shouldUseTypeNameInDSN
```suggestion
```

Sources/Sentry/include/HybridPublic/SentryDependencyContainer.h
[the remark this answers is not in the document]
philipphofmann: Those classes are internal, we can always rename them later. As they only for the watchdog termination logic, I think it's good to be specific with the naming and keep it as is.

Sources/Sentry/SentryDependencyContainer.m
philipphofmann: Because we don't need the `@synchronized`; it's obsolete. You can safely access this method from multiple threads. Each thread gets a different instance of `SentryWatchdogTerminationBreadcrumbProcessor` and the `[self fileManager]` is thread safe. Unless I'm missing something.

philipphofmann: > introduced the SentryDispatchFactory also as a factory to create instances instead of SentryDispatchQueueWrapper

Mainly for testing purposes. I don't mind keeping it, no strong opinion.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
```diff
@@ -0,0 +1,83 @@
...
+        // The error handler is required due but seems not to be executed.
+        guard JSONSerialization.isValidJSONObject(context) else {
+            SentryLog.error("Failed to serialize context, reason: context is not valid json: \(context)")
+            return nil
+        }
+        do {
+            return try JSONSerialization.data(withJSONObject: context, options: [])
+        } catch {
+            SentryLog.error("Failed to serialize context, reason: \(error)")
+            return nil
```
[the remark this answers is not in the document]
philipphofmann: Good catch, we can reuse this https://github.com/getsentry/sentry-cocoa/blob/ee00d2bf45900a3f751bf67ca5e035f572e792d3/Sources/Sentry/SentrySerialization.m#L19-L33

I totally missed that thanks, @armcknight.

Sources/Sentry/SentryFileManager.m
```diff
@@ -501,6 +508,16 @@ - (NSArray *)readPreviousBreadcrumbs
     return breadcrumbs;
 }
 
+#pragma mark - Contexts
+
+- (void)moveContextFileToPreviousContextFile
+{
+    SENTRY_LOG_DEBUG(@"Moving context file to previous context file");
+    @synchronized(self.contextFilePath) {
```
philipphofmann, line 516: `m`: We now have the logic for accessing the `contextFilePath` split in two classes. This one here uses `@synchronized`, but the `SentryScopeContextPersistentStore` doesn't use any locks. It's very easy to mess this up in the future. Anybody could write more code accessing the `contextFilePath` and `previousContextFilePath` and it's hard to realize that the `SentryScopeContextPersistentStore` also accesses this. Maybe it's better to keep the file paths only in the `SentryScopeContextPersistentStore` and only ask the `fileManager` for the base path there.

Sources/Sentry/SentrySDK.m
philipphofmann: We don't have to fix this now. DispatchQueues are serial by default. We can move parts of these operations to a BG thread, when all of these run on the same serial DispatchQueue it should be doable.

philipphofmann (commented):
Another round. Holy moly, this is way more complex than I thought it would be 🤦

philprime:
Switching this back to draft so you don't feel like reviewing while I am still working on it.

Tests/SentryTests/Integrations/WatchdogTerminations/SentryWatchdogTerminationsScopeObserverTests.swift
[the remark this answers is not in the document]
philprime: Ah fair point, I see it now. Thank you.

philprime:
This PR grew quite in size. I am now going to extract smaller PRs for necessary changes than can be seen as isolated changes (i.e. fixing compiler flags or serialization).

metcalfealex:
👋 Any idea when this fix might get merged? We are very interested in it 😄 as its blocking us from upgrading Sentry from `8.43.0` to the latest version (we are using Sentry RN and would like to upgrade to pick up the Xcode 16.3 fixes)

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -347,4 +349,53 @@ - (SentryMXManager *)metricKitManager SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOC
...
+                       scopeContextStore:self.scopeContextPersistentStore])
+}
+#endif
+
+#if SENTRY_TEST || SENTRY_TEST_CI
++ (instancetype)createForTesting;
+{
+    return [[SentryDependencyContainer alloc] init];
+}
+#endif
```
philippsuess: `m`: What drove you to add this? You should be able to use `reset` and `sharedInstance` for testing.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -347,4 +349,53 @@ - (SentryMXManager *)metricKitManager SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOC
...
+
+- (SentryWatchdogTerminationBreadcrumbProcessor *)
+    getWatchdogTerminationBreadcrumbProcessorWithMaxBreadcrumbs:(NSInteger)maxBreadcrumbs
+{
+    // This method is only a factory, therefore do not keep a reference.
+    // The processor will be created each time it is needed.
+    return [[SentryWatchdogTerminationBreadcrumbProcessor alloc]
+        initWithMaxBreadcrumbs:maxBreadcrumbs
+                   fileManager:self.fileManager];
+}
```
philippsuess, line 380: `m`: This opens the door for creating multiple instances of the `SentryWatchdogTerminationScopeObserver`, which would cause problems as far as I get it, because they both would open a file handle to the same file. I think keeping the initialization of these two classes in the `SentryWatchdogTerminationTrackingIntegration` would improve this a bit, because you can't simply call the dependency container asking for a new instance. You have to do it manually.

Sources/Sentry/SentrySDK.m
```diff
@@ -240,6 +240,8 @@ + (void)startWithOptions:(SentryOptions *)options
     SentryClient *newClient = [[SentryClient alloc] initWithOptions:options];
     [newClient.fileManager moveAppStateToPreviousAppState];
     [newClient.fileManager moveBreadcrumbsToPreviousBreadcrumbs];
+    [SentryDependencyContainer.sharedInstance
+            .scopeContextPersistentStore moveCurrentFileToPreviousFile];
```
philippsuess, line 244: `l`: That's so awkward now. First, we access the file manager from the client and then internally from the Dependency Container here. We should really unify this with https://github.com/getsentry/sentry-cocoa/pull/5315 after merging this PR.

Sources/Sentry/SentryFileManager.m
```diff
@@ -191,6 +191,13 @@ - (void)setDelegate:(id<SentryFileManagerDelegate>)delegate
     _delegate = delegate;
 }
 
+#pragma mark - Convenience Accessors
+
+- (NSURL *)getSentryPathAsURL
+{
+    return [NSURL fileURLWithPath:self.sentryPath];
+}
```
philippsuess, line 199: `m`: It would be great to have a test(s) for this in the `SentryFileManagerTests`, cause if this returns garbage the `SentryScopeContextPersistentStore` or any other future implementation relying on this will break.

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
```diff
@@ -0,0 +1,120 @@
...
+}
+
+- (void)storeBreadcrumb:(NSData *_Nonnull)data
+{
+    unsigned long long fileSize;
+    @try {
+        fileSize = [self.fileHandle seekToEndOfFile];
+
+        [self.fileHandle writeData:data];
+        [self.fileHandle writeData:[@"\n" dataUsingEncoding:NSASCIIStringEncoding]];
```
philippsuess, line 106: `l`: As a follow up, I think we could already append the `\n' to the NSData and avoid calling `writeData`. I think it already was like that, but that's something that caught my eye. We don't have to fix this in this PR.

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
```diff
@@ -0,0 +1,120 @@
...
+
+    [self.fileManager removeFileAtPath:self.fileManager.breadcrumbsFilePathOne];
+    [self.fileManager removeFileAtPath:self.fileManager.breadcrumbsFilePathTwo];
+}
+
+- (void)storeBreadcrumb:(NSData *_Nonnull)data
+{
+    unsigned long long fileSize;
+    @try {
+        fileSize = [self.fileHandle seekToEndOfFile];
```
philippsuess, line 103: `l`: Also as a follow up: we should evaluate if the really need `seekToEndOfFile`, because it adds some overhead. While it is cheap, it would be gread if we could avoid it.

Tests/SentryTests/Integrations/WatchdogTerminations/TestSentryWatchdogTerminationBreadcrumbProcessor.swift
```diff
@@ -0,0 +1,23 @@
+@testable import Sentry
+import SentryTestUtils
+
+// Note: This file should ideally live in SentryTestUtils, but this would lead to circular imports.
+// When refactoring the project structure, consider moving this to SentryTestUtils.
```
philippsuess, line 5: Thanks for the comment

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
```diff
@@ -0,0 +1,96 @@
...
+    private let fileManager: SentryFileManager
+
+    init(fileManager: SentryFileManager) {
+        self.fileManager = fileManager
+    }
+
+    // MARK: - Context
+
+    func moveCurrentFileToPreviousFile() {
+        SentryLog.debug("Moving context file to previous context file")
```
philippsuess, line 14: `l`: The fileManager.moveState has a very similar log message already. We could ditch it
```suggestion
```

https://github.com/getsentry/sentry-cocoa/blob/f96ab191e62125a645c846318e73eb96dc0a5794/Sources/Sentry/SentryFileManager.m#L922-L924

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
```diff
@@ -0,0 +1,96 @@
...
+            SentryLog.error("Failed to deserialize context, reason: data is not valid json")
+            return nil
+        }
+        // Validate that the deserialized values are of the expected type.
+        for (key, value) in deserialized {
+            guard value is [String: Any] else {
+                SentryLog.error("Failed to deserialize context, reason: value for key \(key) is not a valid dictionary")
+                return nil
+            }
+        }
```
philippsuess: `h`: I'm curious what drove you to add this check? Wouldn't it be better to only throw avay the specific key? And if we want to keep this, keep in mind that the context can be nested, so we would need to also check nested dictionaries. We don't do this for the SentryCrashScopeObserver for example.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
```diff
@@ -0,0 +1,96 @@
...
+
+    // MARK: - Helpers
+
+    /**
+     * Path to a state file holding the latest context observed from the scope.
+     *
+     * This path is used to keep a persistent copy of the scope context on disk, to be available after
+     * restart of the app.
+     */
+    var contextFileURL: URL {
```
philippsuess, line 102: `l`: Shouldn't this and the `previousContextFileURL` be private? We only use it in this class and in the tests. Ideally, the tests don't have to access this, IMO. If they do, they're checking implementation details. If we really want to check the filename, we can also assert that certain files are written when using `writeContextToDisk` and the other methods.

Sources/Sentry/SentryDependencyContainer.m
philprime: Tests don't need shared public static and should be isolated. I added tests for the added logic in the container, e.g. the `testScopeContextStore`. Setting up an entire static container and resetting adds IMO overhead that's unnecessary, as the container should be a reusable component anyways.

Sources/Sentry/SentryDependencyContainer.m
philprime: I moved it here so that dependency container works as a factory using dependencies required by components. 
Simply moving the initialization to the `SentryWatchdogTerminationTrackingIntegration` does not fix the potential issue you mentioned, because I could still create additional instances in other parts of the SDK. This way the initialization is at least consolidated to one class (the container).

Sources/Sentry/SentrySDK.m
philprime: I added this based on this previous comment by you:

https://github.com/getsentry/sentry-cocoa/pull/5242#discussion_r2104413738

I'll keep it this way, as we are going in circles now.

Sources/Sentry/SentryFileManager.m
philprime: Good catch, I thought I've added the tests already. Adding them now.

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
philprime: Yes, it was already like that:

https://github.com/getsentry/sentry-cocoa/pull/5242/files/ee02a8692038fb6a132876834884b013b324eb97#diff-5a49ba79ff9cffa820eb8f92c3644d3c4a3175025632149daf46c51511c15c46L84-L85

I would propose we refactor the breadcrumb processor to a later point in time and revisit the topic then. Do you want me to create an issue or just add a comment?

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
philprime: I agree, we might not need to seek to the end because when creating the file handle the file is deleted first, so the file handle is never created with existing data.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
philprime: This is here on purpose so that the execution path is traceable from the logs, i.e. the `moveState` is used by multiple methods, but together with the log line in `moveCurrentFileToPreviousFile()` it's clear what the caller is.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
philprime: The return type of the `decodeContext` method is `[String: [String: Any]]?`.
The return type of `SentrySerialization.deserializeDictionary` is `[AnyHashable : Any]?` because it wraps `NSJSONSerialization` which isn't typed and just returns `id`.

This means that the chain of casting is this:

```
+ (id) NSJSONSerialization.JSONObjectWithData
⬇️
+ (NSDictionary *_Nullable)SentrySerialization.deserializeDictionaryFromJsonData
⬇️
SentryScopeContextPersistentStore.decodeContext(...) -> [String: [String: Any]]?
```

My intention was making sure that at least the generic dictionary returned from `SentrySerialization.deserializeDictionaryFromJsonData` is properly validated to respect the return type of `SentryScopeContextPersistentStore.decodeContext`.

> ...keep in mind that the context can be nested, so we would need to also check nested dictionaries

No we do not, because the validation is only making sure that the values of the dictionary are `[String: Any]` and any nested dictionary would be `Any`.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
philprime: This feedback isn't clear to me. If we want to make sure that the file is written to the correct location, we need to get the URL where we expect it to be written.

If I have to rebuild the URL in the tests, the feedback could be "don't repeat yourself". Now I can wrap this getter in a compiler flag to only make it exist for tests, but what's the point there if it's an internal class anyways?

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
philprime: See https://github.com/getsentry/sentry-cocoa/pull/5242#discussion_r2121047950 regarding followup

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
philippsuess: An issue would be great, please.

Sources/Sentry/SentryFileManager.m
philippsuess: Thanks

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
philippsuess: Yes, let's put that into an issue please.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
philippsuess: It's OK to keep it. No strong opinion.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
philippsuess: Ah, now I get it, thanks. It's OK to keep it then, IMO, but I would add this info as a comment, cause it wasn't obvious. I think it's OK to throw away the whole dict in that case, because it means something is clearly wrong.

Sources/Sentry/SentryDependencyContainer.m
philippsuess: > Tests don't need shared public static and should be isolated.

I generally agree with this sentiment, but tests should also test how the code is used. In that case, yes, it's a public global state, but you only the public global state once to retrieve the container in the tests. Then you keep a reference to the container. I think when you change your tests to the following it would be less code and closer to how the code gets used
```
SentryDependencyContainer.reset()
let container = SentryDependencyContainer.sharedInstance()
let processor1 = container.watchdogTerminationContextProcessor
let processor2 = container.watchdogTerminationContextProcessor
```

Sources/Sentry/SentryDependencyContainer.m
philippsuess: Yes, it would only improve the situation slightly and doesn't solve the underlying problem. Up to you.

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
philprime: https://github.com/getsentry/sentry-cocoa/issues/5333

Sources/Sentry/Processors/SentryWatchdogTerminationBreadcrumbProcessor.m
philprime: https://github.com/getsentry/sentry-cocoa/issues/5333

Sources/Sentry/SentrySDK.m
philippsuess: As already pointed out, we can keep this. We don't have to fix it now. It's something that could cause problems in the future. I only wanted to point this out, so we can maybe fix it in the future.

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
philprime: I'll add a comment 👍

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
philippsuess: The tests in `SentryScopeContextPersistentStoreTests.testPreviousContextFileURL_returnsURLWithCorrectPath` already know that the URL is `URL(fileURLWithPath: fixture.fileManager.sentryPath).appendingPathComponent("previous.context.state")`. The other tests use `SentryScopeContextPersistentStore.previousContextFileURL`, but instead, they could also use `URL(fileURLWithPath: fixture.fileManager.sentryPath).appendingPathComponent("previous.context.state")`, which could be a variable in `SentryScopeContextPersistentStoreTests`. Then the advantage would be that no other source code / not tests can access the private variables `contextFileURL` and `previousContextFileURL`. 

That's basically the principle of least knowledge, but maybe we're getting too puristic here. You can also simply ignore me. Not worth the effort of going back and forth.

Sources/Sentry/SentryDependencyContainer.m
philprime: Okay, I'll change it and we can reconsider this another time. Not the focus of this PR

metcalfealex:
Thanks for getting this in @philprime !

Any idea when a release will be out with this fix? To complicate things we would also need the sentry RN version to pick it up 🙏

philprime:
@metcalfealex I'll check internally regarding any additional work that should be done before the next release and will get back to you when I have an ETA.

philprime:
@metcalfealex we released this change in the latest version https://github.com/getsentry/sentry-cocoa/releases/tag/8.52.0

Please notice the warning on the in the release notes, as we are currently working on a hotfix 8.52.1, you might want to wait for.

metcalfealex:
Thanks @philipphofmann! we really appreciate the quick turnaound!
btw any chance we can get a new patch release for sentry RN to pick up the patched 8.52.1 version?

krystofwoldrich:
@metcalfealex The RN SDK version was just released.

https://github.com/getsentry/sentry-react-native/releases/tag/6.15.1

metcalfealex:
Thanks @krystofwoldrich @philprime 

we tested this update today and notice that not all tags are being persisted / set on WatchDog fatal events (we are seeing the user info though).

We confirmed that we are calling Sentry.setTag well before we replicate a WatchDog error. Is it expected that these custom tags will be persisted and visible in this WatchDog events? 

Note on sentry-cocoa 8.43.0 we can see these custom tags on the WatchDog errors

philprime:
Thanks @metcalfealex.

High chance that this is missing due to the limited implementation here:

https://github.com/getsentry/sentry-cocoa/blob/e91f42ab86995f99783df80e8a69c15d82453c5c/Sources/Sentry/SentryWatchdogTerminationScopeObserver.m#L80-L83

We had to do this in individual PRs because it became clear that this will take more time, and we are short on resources.

I'll create another issue where we can track this. Thank you for the patience and support 🙏

PR #5243, opened 2025-05-16
Split swift files into folders

noahsmartin:
This moves the swift code that has no objc dependencies into a folder called "Core". It also adds an option to compile those files with swift package manager. This is a small but independent step for getting everything to use SPM. It did require making some changes, I pulled out a bit of code that depended on objc to be separate from files that otherwise had no objc deps.

#skip-changelog


philipphofmann:
@noahsmartin this PR has huge potential for merge conflicts. To avoid this, we could either merge this quickly or split it up into multiple smaller PRs.

noahsmartin:
Thanks @philipphofmann I updated it to be about half the size, and removed the draft tag, it's ready for review now so hopefully we can merge it before too many conflicts show up

itaybre (approved):

PR #5244, opened 2025-05-16
ref: remove replay import in uikitless umbrella header

armcknight:
Since Session Replay implementation requires UIKit, it doesn't make sense to include this file in the UIKitless SDK variant's  umbrella header.

#skip-changelog


armcknight:
At this point I get errors like:
```
Sources/Swift/Integrations/UserFeedback/SentryFeedback.swift Could not build Objective-C module 'SentryWithoutUIKit'

```
and i'm not sure what else is referencing the header. Maybe just because it's in the path and gets pulled into the build. It's not worth the effort i think to push this further, it'll probably just shake out if we split the sdk up into modules, one of which being a UIKit [umbrella] module.

PR #5246, opened 2025-05-16
fix: move return out of macro

armcknight:
to fix the test crashes in #5225 

#skip-changelog


philipphofmann:
Thanks @armcknight for the suggestion. I could solve it without the extra returns. We can close this.

PR #5251, opened 2025-05-19
test: Refactor session tracker tests with application states

philprime:
While working on #5121 I refactored the session tracker tests for readability and adding application state to mimic the states when using actual app to trigger `didBecomeActive` etc.

It does not adapt the tests themselves.

#skip-changelog


Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
```diff
@@ -482,29 +615,28 @@ class SentrySessionTrackerTests: XCTestCase {
...
-            XCTAssertEqual(SentrySessionStatus.ok.description, session.status.description)
-            XCTAssertEqual(0, session.errors)
-            XCTAssertNil(session.timestamp)
-            XCTAssertNil(session.duration)
-            assertSessionFields(session: session)
-        } else {
-            XCTFail("No session init sent.")
+        guard let session = fixture.client.captureSessionInvocations.last else {
+            return XCTFail("No session init sent.")
         }
```
philipphofmann, line 622: `l`: We could also use XCTUnwrap, but then of course we need throws here and all the tests need throws. Up to you.

philipphofmann (approved):
Thanks for doing this in an extra PR. LGTM.

philipphofmann (commented):
Again, LGTM.

philipphofmann (approved):

Issue #4517, opened 2024-11-07
Improve C++ Exception Handling with Fishhook


philipphofmann:
We could use [fishhook](https://github.com/facebook/fishhook) to enhance C++ exception. Currently, the Sentry SDK uses `dlsym` to look up `__cxa_throw`. https://github.com/getsentry/sentry-cocoa/blob/f31a3eff586235b76a9d9dfb0ec82d7bc326afee/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_CPPException.cpp#L92

This approach falls short in applications with multiple dynamic libraries, as exceptions thrown in those separate modules aren’t caught by the main exception handler, but it works fine when Sentry is linked statically into the same module.

By using [fishhook](https://github.com/facebook/fishhook), which allows functions to be “hooked” by name as they are loaded across various modules, we could dynamically find and intercept all instances of `__cxa_throw`, even in different dynamic libraries. This would provide more comprehensive and consistent exception handling across an app’s runtime, regardless of the number of C++ modules or how they’re linked. 

We used fishhook in the past for trying to get async stacktraces (https://github.com/getsentry/sentry-cocoa/pull/998), but we removed it again because it was not working reliably. The problem wasn't fishhook, but how we used it. We now use `backtrace_async` for async stacktrace in Swift.

This could maybe help with:
* https://github.com/getsentry/sentry-cocoa/issues/935
* https://github.com/getsentry/sentry-cocoa/issues/1267
* https://github.com/getsentry/sentry-cocoa/issues/1562
* https://github.com/getsentry/sentry-cocoa/issues/4226

philippsuess:
We just merged the PR to fix this. In the next release, you can enable `options.experimental.enableUnhandledCPPExceptionsV2 = true` to get proper stacktraces for fatal C++ exceptions. Please keep in mind that this feature is still experimental. We would welcome feedback so we can make the feature stable with https://github.com/getsentry/sentry-cocoa/issues/5309.

PR #5256, opened 2025-05-20
feat: Capturing fatal CPPExceptions via cxa_throw

philipphofmann:
## :scroll: Description

Use fishhook to swap the __cxa_throw C++ exception handler for getting
stacktraces for unhandled C++ exceptions. This implementation is based
on the code in https://github.com/kstenerud/KSCrash.

The API stability check fails because I added a new experimental option `enableUnhandledCPPExceptionsV2`. 

There are still a few open issues to address before making this GA; see https://github.com/getsentry/sentry-cocoa/issues/5309

Fixes GH-4517

<details><summary>Testing Details macOS sample</summary>
<p>

### Without the feature 
 Screenshot 2025-05-20 at 09 00 00

https://sentry-sdks.sentry.io/issues/6590753526/events/d1bb9784083e40a8abfdb28ac2b9c34e/?project=5428557 

#### Invalid Argument

https://github.com/getsentry/sentry-cocoa/blob/e70c3aaf9742fd87cc1ff1c81df986154a565eab/Samples/macOS-Swift/Shared/CppSample.cpp#L4-L8

Screenshot 2025-05-20 at 09 02 18 https://sentry-sdks.sentry.io/issues/6545237791/events/a393feb8d8b2401c8d6cedca5ce181a4/?project=5428557

#### rethrowNoActiveCPPException

https://github.com/getsentry/sentry-cocoa/blob/e70c3aaf9742fd87cc1ff1c81df986154a565eab/Samples/macOS-Swift/Shared/CppSample.cpp#L22-L28

Screenshot 2025-05-20 at 09 06 59

https://sentry-sdks.sentry.io/issues/6617441025/events/30bc9db8d17b4dcd80d6473c57a3c36f/?project=5428557

#### C++ Exception from BG Thread

Screenshot 2025-05-20 at 09 08 15

https://sentry-sdks.sentry.io/issues/6565152294/events/691f845d8dc04305b0cf376e2d21732b/?project=5428557

#### NoExcept C++ Exception

https://github.com/getsentry/sentry-cocoa/blob/e70c3aaf9742fd87cc1ff1c81df986154a565eab/Samples/macOS-Swift/Shared/CppSample.cpp#L16-L20

Screenshot 2025-05-20 at 09 10 17

https://sentry-sdks.sentry.io/issues/6565163235/events/d9dc157068994ed2879842dbaf5f477b/?project=5428557

</p>
</details> 

Unhandled C++Exception from testflight iOS-Swift
https://sentry-sdks.sentry.io/issues/6623122757/events/85451edf3f324ad98a6293f1e2cc4b55/?project=5428557

Screenshot 2025-05-22 at 10 12 16

## :pencil: Checklist

### DIFF
--- a/.github/workflows/ui-tests-critical.yml
+++ b/.github/workflows/ui-tests-critical.yml
@@ -38,6 +38,8 @@ jobs:
         with:
           name: SwiftUITestSample
           path: TestSamples/SwiftUITestSample/DerivedData/Build/Products/Debug-iphonesimulator/SwiftUITestSample.app
+          # If no files are found, fail the job with an error, because they are required for the next step
+          if-no-files-found: error
 
   run-tests:
     name: Test iOS on Xcode ${{matrix.xcode}} - V2 # Up the version with every change to keep track of flaky tests
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -192,6 +192,8 @@ platform :ios do
       skip_archive: true,
       skip_package_dependencies_resolution: true
     )
+
+    delete_keychain(name: "fastlane_tmp_keychain") unless is_ci
   end
 
   desc "Upload iOS-Swift to TestFlight and symbols to Sentry"


Sources/SentryCrash/Recording/Tools/SentryCrashCxaThrowSwapper.c
```diff
@@ -0,0 +1,383 @@
...
+    if (g_cxa_throw_handler != NULL) {
+        g_cxa_throw_handler(thrown_exception, tinfo, dest);
+    }
+
+    void *backtraceArr[k_requiredFrames];
+    int count = backtrace(backtraceArr, k_requiredFrames);
+
+    Dl_info info;
+    if (count >= k_requiredFrames) {
+        if (dladdr(backtraceArr[k_requiredFrames - 1], &info) != 0) {
```
philipphofmann, line 150: @supervacuus I'm a bit unsure about this approach here. If we don't call the original `cxa_throw_type` apps crash when throwing C++ exceptions. It's very easy to reproduce. Basically swap cxa_throw, comment out `original(thrown_exception, tinfo, dest);` and then throw any C++ exception in a try catch. Then the app cashes. Do you have any idea what could be the motivation for only calling this if the `count >= k_requiredFrames`?

philprime:
I'll wait for @supervacuus feedback on your PR comment before doing the review in case it needs another revision first.

philipphofmann:
> I'll wait for @supervacuus feedback on your PR comment before doing the review in case it needs another revision first.

This is mostly native stuff. I would wait for the review of @supervacuus before adding your review.

supervacuus (commented):
Quick update: I have already started "investigating" the changes. I hope that my response to your comment makes sense. This conversation should ideally lead to inline comments. 

I am still in the process, but most parts look relatively sane. A first questionable impression is the lack of thread synchronization, considering that we access quite a few globals. `cxa_throw` itself can and will be called from multiple threads simultaneously, which might be less of a problem if we only read. However, there is still a chance that setup and teardown happen concurrently. Did you consider these possible interactions already? It could be fine if you were sure to delineate reads and writes with certainty, but that is rarely the case.

supervacuus (changes requested):
This looks great—only minor things. The only unanswered aspect I currently see is that during sentry initialization, when you rebind `__cxa_throw` with your decorator, another thread could already be throwing and entering your decorator. At the same time, the `g_cxa_originals` structure is still not finished. 

The problem with that is that we _must_ call an "original"; otherwise, we crash. We can either guarantee that the SDK initialization is completed before any other thread can throw, or we create an explicit `abort()` path that logs the situation, so users aren't confronted with an indecipherable crash in this scenario. And to be clear, this is an edge case, but still a race.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_CPPException.cpp
```diff
@@ -71,6 +73,20 @@ static SentryCrashStackCursor g_stackCursor;
...
 // ============================================================================
 
+static NEVER_INLINE void
+captureStackTrace(void *, std::type_info *tinfo, void (*)(void *)) KEEP_FUNCTION_IN_STACKTRACE
+{
+    SENTRY_ASYNC_SAFE_LOG_TRACE("Entering captureStackTrace");
+
+    if (tinfo != nullptr && strcmp(tinfo->name(), "NSException") == 0) {
+        return;
+    }
```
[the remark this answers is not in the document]
philipphofmann: Yes exactly. I added a comment.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_CPPException.cpp
```diff
@@ -79,18 +95,21 @@ void __cxa_throw(void *thrown_exception, std::type_info *tinfo, void (*dest)(voi
...
+    void *thrown_exception, std::type_info *tinfo, void (*dest)(void *)) KEEP_FUNCTION_IN_STACKTRACE
 {
-    if (g_captureNextStackTrace) {
-        sentrycrashsc_initSelfThread(&g_stackCursor, 1);
-    }
+    SENTRY_ASYNC_SAFE_LOG_TRACE("Entering __cxa_throw");
 
     static cxa_throw_type orig_cxa_throw = NULL;
+    if (sentrycrashct_is_cxa_throw_swapped() == false) {
+        captureStackTrace(thrown_exception, tinfo, dest);
```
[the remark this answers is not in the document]
philipphofmann: Yes, exactly. Again I added a comment.

supervacuus (approved):

philipphofmann:
I have no clue why, but the unit tests for Mac Catalyst keep failing. I need to figure out why that is before merging this.

Sources/SentryCrash/Recording/Tools/SentryCrashCxaThrowSwapper.c
```diff
@@ -68,77 +68,69 @@
...
-    SENTRY_ASYNC_SAFE_LOG_TRACE("Freed memory for stored pairs");
-}
-
-static uintptr_t
-findAddress(void *address)
-{
-    SENTRY_ASYNC_SAFE_LOG_TRACE("Finding address for %p", address);
+    SENTRY_ASYNC_SAFE_LOG_DEBUG(
+        "Adding pair for image with base address: %p, and originalCxaThrowFunction: %p",
+        (void *)pair.cxa_throw_original_function, (void *)pair.cxa_throw_original_function);
```
supervacuus:
```suggestion
    SENTRY_ASYNC_SAFE_LOG_DEBUG(
        "Adding pair for image with base address: %p, and originalCxaThrowFunction: %p",
        (void *)pair.image_dli_fbase_address, (void *)pair.cxa_throw_original_function);
```

Sources/Sentry/SentryCrashIntegration.m
```diff
@@ -165,6 +169,10 @@ - (void)startCrashHandler:(NSString *)cacheDirectory
         }
 #endif // TARGET_OS_OSX
 
+        if (enableCppExceptionsV2) {
+            sentrycrashcm_cppexception_enable_swap_cxa_throw();
```
philprime, line 176: `l`: We should add a debug log statement here to indicate that it was enabled.

Sources/Sentry/include/SentryCompiler.h
```diff
@@ -237,6 +237,11 @@
 #define NEVER_INLINE
 #endif
 
+/* KEEP_FUNCTION_IN_STACKTRACE */
+#define KEEP_FUNCTION_IN_STACKTRACE __attribute__((disable_tail_calls))
+
+#define THWART_TAIL_CALL_OPTIMISATION __asm__ __volatile__("");
+
```
philprime: `m`: We should add documentation to macros so future maintainers understand the purpose. 
I'll created a draft based on how I understand it, please consider applying and improving if not correct.

```suggestion
/**
 * KEEP_FUNCTION_IN_STACKTRACE
 *
 * This macro disables tail call optimization (TCO) for the function it decorates.
 * It ensures that the function remains present in the stack trace, even if it
 * could otherwise be optimized away by the compiler.
 *
 * This attribute is supported by Clang and GCC.
 *
 * Example:
 *   KEEP_FUNCTION_IN_STACKTRACE
 *   void capture_error(Error *err) {
 *       // Stack frame will be preserved in the captured stack trace.
 *       sentry_send_event(err);
 *   }
 */
#define KEEP_FUNCTION_IN_STACKTRACE __attribute__((disable_tail_calls))

/**
 * THWART_TAIL_CALL_OPTIMISATION
 *
 * This macro inserts a volatile no-op inline assembly instruction, acting as a
 * barrier to tail call optimization at a specific call site.
 *
 * This can be used to force intermediate utility or wrapper functions to remain
 * visible in stack traces, which helps developers trace the full logical flow of
 * their application when viewing captured exceptions or performance data in Sentry.
 *
 * This is especially useful in hand-written C code where macros, error wrappers,
 * or utility functions would otherwise be optimized into tail calls and removed
 * from the stack trace.
 *
 * Example:
 *   void send_wrapped_event(Event *evt) {
 *       sentry_send_event(evt);
 *       THWART_TAIL_CALL_OPTIMISATION;
 *   }
 */
#define THWART_TAIL_CALL_OPTIMISATION __asm__ __volatile__("");

```

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_CPPException.cpp
```diff
@@ -237,11 +282,29 @@ setEnabled(bool isEnabled)
         } else {
             std::set_terminate(g_originalTerminateHandler);
             g_originalTerminateHandler = NULL;
+
+            // This method is a no-op if cxa_throw is not swapped.
```
philprime, line 286: Thanks for this comment

Sources/SentryCrash/Recording/Tools/SentryCrashCxaThrowSwapper.c
```diff
@@ -411,7 +409,8 @@ sentrycrashct_unswap_cxa_throw(void)
         rebind_symbols_for_image(header, slide, false);
     }
 
-    sentrycrashct_clear_pairs();
+    // We MUST NOT clear the pairs because a if we can't unswap one of the cxa_throw handlers, we
+    // still MUST call the original cxa_throw handler.
```
supervacuus: Not sure if you want to remark here that this isn't supposed to be that way, but investigated as part of #5309.

supervacuus (approved):
Inspected the changes since the last review. Except for one copypasta everything looks sensible.

Sources/SentryCrash/Recording/Tools/SentryCrashPlatformSpecificDefines.h
```diff
@@ -42,4 +42,8 @@ typedef struct nlist nlist_t;
 #    define LC_SEGMENT_ARCH_DEPENDENT LC_SEGMENT
 #endif /* __LP64__ */
 
+#ifndef SEG_DATA_CONST
```
philprime, line 51: `l`: It would be helpful to have a comment here explaining the purpose of this constant, to give additional context without having to look up `__DATA`

Sources/Swift/Core/SentryExperimentalOptions.swift
```diff
@@ -21,6 +21,12 @@ public class SentryExperimentalOptions: NSObject {
      */
     public var enableFileManagerSwizzling = false
 
+    /// A more reliable way to report unhandled C++ exceptions.
+    ///
+    /// This approach hooks into all intances of the `__cxa_throw` function, which provides a more comprehensive and consistent exception handling across an app’s runtime, regardless of the number of C++ modules or how they’re linked. It helps in obtaining accurate stack traces.
+    ///
```
philprime: `m`: Please use Swift documentation features to mark this as an release. For consistency in this file I would use `/** */` instead of `///` while both is valid. Also add the notes about the topics we discussed earlier regarding caching of debug symbols.

```suggestion
    /**
     * A more reliable way to report unhandled C++ exceptions.
     *
     * This approach hooks into all instances of the `__cxa_throw` function, which provides a more comprehensive and consistent exception handling across an app’s runtime, regardless of the number of C++ modules or how they’re linked. It helps in obtaining accurate stack traces.
     *
     * - Note: The mechanism of hooking into `__cxa_throw` could cause issues with symbolication on iOS due to caching of symbol references.
     * - Experiment: This is an experimental feature and is therefore disabled by default. We'll enable it by default in a future major release.
     */
```

Tests/SentryTests/SentryCrash/SentryCrashCxaThrowSwapper_Tests.mm
```diff
@@ -0,0 +1,311 @@
...
+        NSString *typeName = [NSString stringWithCString:tinfo->name()
+                                                encoding:NSUTF8StringEncoding];
+
+        g_exceptionHandlerBlock(errorMessage, typeName);
+    }
+
+    THWART_TAIL_CALL_OPTIMISATION
+}
+
+static NEVER_INLINE void
```
philprime, line 39: `l`: Why do we need to add this inline annotation here, but not in the SDK itself? Could we use `KEEP_FUNCTION_IN_STACKTRACE` here?

Sources/SentryCrash/Recording/Tools/SentryCrashCxaThrowSwapper.c
```diff
@@ -0,0 +1,423 @@
...
+
+static void
+__cxa_throw_decorator(void *thrown_exception, void *tinfo, void (*dest)(void *))
+{
+    const int k_requiredFrames = 2;
+
+    if (g_cxa_throw_handler != NULL) {
+        SENTRY_ASYNC_SAFE_LOG_DEBUG(
+            "Not __cxa_throw decorator, because no g_cxa_throw_handler set.");
+        g_cxa_throw_handler(thrown_exception, tinfo, dest);
```
philprime, line 142: `m`: Is there a change that the `g_cxa_throw_handler` could call the original implementation inside, therefore calling it a second time at the end of this decorator method?

Sources/SentryCrash/Recording/Tools/SentryCrashCxaThrowSwapper.c
```diff
@@ -0,0 +1,423 @@
...
+        return;
+    }
+
+    SENTRY_ASYNC_SAFE_LOG_DEBUG(
+        "Adding pair for image with base address: %p, and originalCxaThrowFunction: %p",
+        (void *)pair.cxa_throw_original_function, (void *)pair.cxa_throw_original_function);
+
+    if (g_cxa_originals_count == g_cxa_originals_capacity) {
+        g_cxa_originals_capacity *= 2;
+        g_cxa_originals = (SentryCrashImageToOriginalCxaThrowPair *)realloc(g_cxa_originals,
```
philprime, line 115: `m`: IIUC if this realloc returns NULL it has already overwritten the `g_cxa_originals`. Not sure if this has an impact.

Sources/SentryCrash/Recording/Tools/SentryCrashCxaThrowSwapper.c
philippsuess: I think it's OK to keep it that way for now, because it's a safety net. I would rather use a bit more memory than crash.

Sources/Sentry/include/SentryCompiler.h
philippsuess: Thanks, I copied this from KSCrash and didn't think about adding comments.

Sources/Swift/Core/SentryExperimentalOptions.swift
philippsuess: Whoops. Thanks for pointing that out. Since when do we use `/** */` for Swift??? I'm used to `///`. Will fix this for consitency.

Tests/SentryTests/SentryCrash/SentryCrashCxaThrowSwapper_Tests.mm
philippsuess: We also use `KEEP_FUNCTION_IN_STACKTRACE` here, and we also use `static NEVER_INLINE` in the SDK.

https://github.com/getsentry/sentry-cocoa/blob/8a01402c3129788bc8ad2791ed5763b2cda0910b/Tests/SentryTests/SentryCrash/SentryCrashCxaThrowSwapper_Tests.mm#L39-L41

https://github.com/getsentry/sentry-cocoa/blob/8a01402c3129788bc8ad2791ed5763b2cda0910b/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_CPPException.cpp#L76-L77

Sources/SentryCrash/Recording/Tools/SentryCrashCxaThrowSwapper.c
philippsuess: Good catch. The [Cpp reference on realloc](https://en.cppreference.com/w/c/memory/realloc) states: 

> On failure, returns a null pointer. The original pointer ptr remains valid and may need to be deallocated with [free](https://en.cppreference.com/w/c/memory/free.html) or realloc.

According to my understanding, as we pass `g_cxa_originals` as the pointer and then assign the result to it, we would leak memory if `realloc` fails, because it returns NULL, overwrites `g_cxa_originals` to `NULL` and we loose the pointer to it's previous memory. @supervacuus can you help us out here please?

Sources/SentryCrash/Recording/Tools/SentryCrashCxaThrowSwapper.c
philippsuess: The `g_cxa_throw_handler` is what we set it to. In our case, it's `captureStackTrace` inside `SentryCrashMonitor_CPPException`. We don't have access to the original implementation inside `captureStackTrace`, but only the exception. If you would go crazy and somehow get access to it, then yes. An actual problem I see is what happens when we throw an unintentional CppException in `captureStackTrace`. I added this to https://github.com/getsentry/sentry-cocoa/issues/5309 as point 7 to validte before GA.

philippsuess:
The API Stability Check is supposed to fail due to adding a new experimental options `enableUnhandledCPPExceptionsV2`.

Sources/SentryCrash/Recording/Tools/SentryCrashCxaThrowSwapper.c
supervacuus: Yes, the correct pattern for `realloc` would be

```C
SentryCrashImageToOriginalCxaThrowPair *tmp_cxa_originals = (SentryCrashImageToOriginalCxaThrowPair *)realloc(g_cxa_originals, sizeof(SentryCrashImageToOriginalCxaThrowPair) * g_cxa_originals_capacity);
if (!tmp_cxa_originals) {
    SENTRY_ASYNC_SAFE_LOG_ERROR(
        "Failed to realloc memory for g_cxa_originals: %s", strerror(errno));
    free(g_cxa_originals);
    return;
}
g_cxa_originals = tmp_cxa_originals;
```

However, in practice, if you fail to allocate memory for such a small structure, the leak caused by an inappropriate `realloc` use will be the least of your problems. 

In particular, it might be prudent to escalate (i.e., not only return, but return an error and handle it appropriately in rebind), since you have now freed the global structure, you must also stop the rebinding process (which doesn't make sense anyway if allocation returns `NULL`, since exceptions must also be allocated).

philippsuess: Thanks @supervacuus. I added a comment that we intentionally ignore this for now and added this as a todo in https://github.com/getsentry/sentry-cocoa/issues/5309.

zkacso-dev:
I don't see this flag mentioned in the documentation https://docs.sentry.io/platforms/apple/features/experimental-features/ .
Is it ready to use in a production app?

kahest:
@zkacso-dev thanks for the ping - we'll add the option to the Experimental Features page. We recommend testing this feature in a pre-production environment and appreciate feedback you might have. We have a small [list of items](https://github.com/getsentry/sentry-cocoa/issues/5309) we want to follow up with before we make the feature non-experimental.

philippsuess:
Related docs PR to add this to the list of experimental features https://github.com/getsentry/sentry-docs/pull/14731

PR #5257, opened 2025-05-20
fix: change http transport flush timeout log to warning

philprime:
It should be a warning

#skip-changelog


philipphofmann (approved):
LGTM, thanks.

PR #5258, opened 2025-05-20
feat: add preferred frame rate to display link wrapper

philprime:
## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryFramesTracker.m
+++ b/Sources/Sentry/SentryFramesTracker.m
@@ -168,7 +168,10 @@ slowFrameThreshold(uint64_t actualFramesPerSecond)
 
     _isRunning = YES;
 
-    [_displayLinkWrapper linkWithTarget:self selector:@selector(displayLinkCallback)];
+    // Set the display link to use the default frame rate of 60 fps.
+    [_displayLinkWrapper linkWithTarget:self
+                               selector:@selector(displayLinkCallback)
+               preferredFramesPerSecond:-1];
 }
 
 - (void)displayLinkCallback
--- a/Sources/Sentry/include/SentryDisplayLinkWrapper.h
+++ b/Sources/Sentry/include/SentryDisplayLinkWrapper.h
@@ -2,6 +2,8 @@
 
 #if SENTRY_HAS_UIKIT
 
+#    import <QuartzCore/QuartzCore.h>
+
 NS_ASSUME_NONNULL_BEGIN
 
 /**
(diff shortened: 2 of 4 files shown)


philipphofmann (commented):
> If the display link is not configured with a target frame rate, it can cause apps to run at 60 FPS even if not necessary.

Can you prove this? How can we be sure that's the case? According to a recent investigation of @noahsmartin we should actually swap out the CADisplayLink and use a [RunLoopObserver](https://developer.apple.com/documentation/corefoundation/cfrunloopobserver?language=objc) instead; see internal [Notion doc](https://www.notion.so/sentry/iOS-Perf-Performance-1ee8b10e4b5d805f8917c1440fee7479#1ee8b10e4b5d8044871fd7e2f4f20036). I think it makes sense to decide on our future approach before touching this right now.

philprime:
> Can you prove this?
I observed this behavior during testing of #4940 that the sample app starts off with less than 60 FPS frame rate, then when SR starts it spins up to 60 FPS. But it could also be caused from unoptimized 

I would have to create a reproducible project to proof this with numbers, but I believe the documentation [1] states enough reason for this change:

> Use lower refresh rates whenever possible to save power, because higher refresh rates can result in significant power consumption.

- [1] https://developer.apple.com/documentation/quartzcore/optimizing-promotion-refresh-rates-for-iphone-13-pro-and-ipad-pro#Enable-faster-ProMotion-refresh-rates

> According to a recent investigation of @noahsmartin we should actually swap out the CADisplayLink and use a [RunLoopObserver](https://developer.apple.com/documentation/corefoundation/cfrunloopobserver?language=objc) instead; see internal [Notion doc](https://www.notion.so/sentry/iOS-Perf-Performance-1ee8b10e4b5d805f8917c1440fee7479#1ee8b10e4b5d8044871fd7e2f4f20036). I think it makes sense to decide on our future approach before touching this right now.

I agree with swapping it out for performance tracing, I am not convinced yet that it is the right approach for visual timers and per-frame updates. We need to ensure that we do get the target frame rate, as we are trying to achieve constant FPS. This is out-of-scope for this PR.

philipphofmann:
As discussed in an internal meeting, I think I wouldn't change the preferred frame rate for SR right now. I think we first should investigate https://github.com/getsentry/sentry-cocoa/issues/5263 before tackling this PR.

philprime:
Closing this PR as stale, we can reconsider using this when #5263 is done

PR #5261, opened 2025-05-20
Redo split files

noahsmartin:
Redo of https://github.com/getsentry/sentry-cocoa/pull/5243 this time with a fix for the not recognized selector issue.

The problem was I had a source file that only contained an extension and was only referenced through an objc selector. This caused the linker to consider it dead code and remove it, but only when using static linking. The issue is described here: https://github.com/swiftlang/swift/issues/48561

In that swift issue they recommend using the linker flag -ObjC to work around it, and I confirmed that worked in our test app. However, we don't want to recommend all users of the static SDK use that flag. I first moved the extension code into a different swift file that contained a class which was referenced and that fixed the issue, however I didn't want to put this logging related code in some random other Swift file. This fix changes it so the file doesn't contain only an extension, and instead has a helper class. I added a comment to explain why this helper class is there.

#skip-changelog


philipphofmann (approved):
LGTM, thanks for the comment explaining why we can't use an extension.

PR #5264, opened 2025-05-20
ci: give fastlane more time before timing out

armcknight:
We get a lot of UI test check run failures due to this timeout elapsing, so let's see if giving them more time helps. E.g. https://github.com/getsentry/sentry-cocoa/actions/runs/15141987547/job/42568403480#step:7:51

#skip-changelog

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1,6 +1,10 @@
 default_platform(:ios)
 
 platform :ios do
+  before_all do
+    ENV["FASTLANE_XCODE_LIST_TIMEOUT"] = "120" # sometimes running xcodebuild commands is slow on CI runners. give them more time than the default of 10 seconds. see https://github.com/fastlane-old/gym/issues/188#issuecomment-190320592 and https://docs.fastlane.tools/advanced/lanes/
+  end
+
   ios_swift_infoplist_path = "./Samples/iOS-Swift/iOS-Swift/Info.plist"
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -1,6 +1,10 @@
 default_platform(:ios)
 
 platform :ios do
+  before_all do
+    ENV["FASTLANE_XCODE_LIST_TIMEOUT"] = "120" # sometimes running xcodebuild commands is slow on CI runners. give them more time than the default of 10 seconds. see https://github.com/fastlane-old/gym/issues/188#issuecomment-190320592 and https://docs.fastlane.tools/advanced/lanes/
```
noahsmartin, line 5: This is in seconds? We might want something lower like 30 seconds? Just to avoid really long waits if it is actually going to fail

armcknight: yes, it's seconds, but also doesn't even seem to be respected the way i've set it here. see https://github.com/getsentry/sentry-cocoa/actions/runs/15145687017/job/42580715842?pr=5264#step:7:44 
image

armcknight: It should've been 120 instead of 3

armcknight: ah, wrong (or outdated) env var has proliferated online. the right one should be FASTLANE_XCODEBUILD_SETTINGS_TIMEOUT

armcknight: works https://github.com/getsentry/sentry-cocoa/actions/runs/15146553023/job/42583600066#step:7:44
image

PR #5267, opened 2025-05-20
ci: fastlane match fixes

armcknight:
This is an intermittent failure that's plagued us for a long time. The latest fix was attempted in https://github.com/getsentry/sentry-cocoa/pull/4597 but that still didn't get it, I see a failure today in https://github.com/getsentry/sentry-cocoa/actions/runs/15146553011/job/42583551906?pr=5264#step:10:96

`clone_branch_directly` attempts to create a new branch, which seems like it should _always_ fail since master is the default branch for the repo.

specifying `git_branch` as `master` is just reiterating the default value for that parameter.

Also, centralizes these settings in Matchfile since they're copied at every invocations of `sync_code_signing`

#skip-changelog

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -82,13 +82,7 @@ platform :ios do
 
     sync_code_signing(
       type: "appstore",
-      readonly: true,
       app_identifier: ["io.sentry.sample.iOS-Swift",  "io.sentry.sample.iOS-Swift.Clip"],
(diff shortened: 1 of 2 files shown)


philprime (approved):
LGTM, thanks for looking into this again.

philipphofmann (approved):
Thank you @armcknight

PR #5268, opened 2025-05-20
ci: Verify bumped version on CI

itaybre:
## :scroll: Description

Adds a verify parameter to check if a the version was properly bumped.

Make sure our bump utility is working properly

Ran the bump util locally and then ran the verify step.

## :pencil: Checklist

### DIFF
--- a/Utils/VersionBump/main.swift
+++ b/Utils/VersionBump/main.swift
@@ -2,6 +2,38 @@ import Foundation
 import Regex
 import SwiftShell
 
+protocol ErrorHandling: Error {
+    var message: String { get }
+}
+
+enum FileError: Error, ErrorHandling {
+    case fileNotFound(String)
+    case unknownFile(String)
+    
+    var message: String {
+        switch self {
+        case .fileNotFound(let file):
+            return "File not found: \(file)"
+        case .unknownFile(let file):
+            return "Unknown file: \(file)"
+        }
+    }
+}
+
+enum VersionError: Error, ErrorHandling {
+    case versionNotFound(String)
+    case versionMismatch(String, String)
+    
+    var message: String {
+        switch self {
+        case .versionNotFound(let file):
+            return "No version found for \(file)"
+        case .versionMismatch(let file, let versionFound):
+            return "Unexpected version \(versionFound) found for file \(file)"
+        }
+    }
+}
+
 let fromVersionFile = "./Sentry.podspec"
 
 let files = [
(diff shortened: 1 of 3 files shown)


Utils/VersionBump/main.swift
```diff
@@ -62,3 +71,50 @@ func extractVersionOnly(_ version: String) -> String {
...
+            }
+        } else if file == "./Package.swift" {
+            regexString = "https:\\/\\/github\\.com\\/getsentry\\/sentry-cocoa\\/releases\\/download\\/(?<version>[a-zA-z0-9\\.\\-]+)\\/Sentry"
+        } else if file == "./Sources/Sentry/SentryMeta.m" {
+            regexString = "static NSString \\*versionString = @\"(?<version>[a-zA-z0-9\\.\\-]+)\""
+        }
+        let match = try? Regex(string: regexString, options: [.dotMatchesLineSeparators]).firstMatch(in: fileContent)
+        let version = match?.captures[0] ?? "Version Not Found"
+        if version != expectedVersion {
+            exit(errormessage: "Unexpected version \(version) found for file '\(file)'")
```
armcknight, line 97: I would either fail faster on the `try?` by converting it to `try` or, my preference, gather all the failures in the for loop and then print a summary at the end. It could be helpful to know if only a subset of the versions weren't updated correctly, vs failing fast and then having to push multiple fixes until all problems are solved.

That would indicate we should change the `try` on line 79 to a `try?` as well, or catch the error and store for later while trying the rest of the files.

Utils/VersionBump/main.swift
```diff
@@ -62,3 +71,50 @@ func extractVersionOnly(_ version: String) -> String {
...
+    for file in restrictFiles {
+        let fileHandler = try open(file)
+        let fileContent: String = fileHandler.read()
+        
+        let marketingRegex = try? Regex(string: "MARKETING_VERSION\\s=\\s(?<version>[a-zA-z0-9\\.\\-]+)", options: [.dotMatchesLineSeparators])
+        let currentProjectRegex = try? Regex(string: "CURRENT_PROJECT_VERSION\\s=\\s(?<version>[a-zA-z0-9\\.\\-]+)", options: [.dotMatchesLineSeparators])
+        let match = marketingRegex?.firstMatch(in: fileContent) ?? currentProjectRegex?.firstMatch(in: fileContent)
+        let version = match?.captures[0] ?? "Version Not Found"
+        if version != exactVersion {
+            exit(errormessage: "Unexpected version \(version) found for file '\(file)'")
```
armcknight, line 113: Same comment here, consider iterating through all files so we get a more holistic view of everything that failed.

armcknight (approved):
Looks fine, I have some suggestions. I also noticed, through no fault of yours, that version-bump-util.yml just runs on every single PR? We should add trigger paths for just the files involved. We can always do that in a separate PR.

PR #5269, opened 2025-05-20
Add sentry private SPM

noahsmartin:
## :pencil: Checklist


philipphofmann (commented):
I'm a bit unsure about the naming of `_SentryPrivate`, because it contains public ObjC header files. That confuses me and feels wrong. 

Also a bit confusing to me is that the Xcode structure doesn't match the folder structure. It can get confusing where to add a new public header file or a new header file in general. Do you think we should consider reflecting the actual folder structure with the Xcode groups so it's less confusing?

Apart from that, I think this PR goes in the right direction.

philprime:
> Also a bit confusing to me is that the Xcode structure doesn't match the folder structure. It can get confusing where to add a new public header file or a new header file in general. 

I would highly appreciate the Xcode project structure reflecting the file hierarchy. We can not use synchronized folders yet, due to using older Xcode versions, but at some point we could migrate.

PR #5271, opened 2025-05-21
ci: build and sign xcframework

itaybre:
## :scroll: Description

Sign xcframework when building it

## :pencil: Checklist


philprime (commented):
Thanks for looking into this.

1. How long are frameworks signed?
2. What happens when frameworks are not signed anymore?
3. What happens if a certificate is revoked?

We should make sure that older versions of the framework can still be used.

Makefile
```diff
@@ -111,6 +111,14 @@ build-xcframework:
 	ditto -c -k -X --rsrc --keepParent Carthage/SentrySwiftUI.xcframework Carthage/SentrySwiftUI.xcframework.zip
 	ditto -c -k -X --rsrc --keepParent Carthage/Sentry-WithoutUIKitOrAppKit.xcframework Carthage/Sentry-WithoutUIKitOrAppKit.zip
 
+build-xcframework-sign:
+	@echo "--> Carthage: creating Signed Sentry xcframework"
+	./scripts/build-xcframework.sh | tee build-xcframework.log
+	./scripts/sign_and_zip.sh Carthage/Sentry.xcframework
+	./scripts/sign_and_zip.sh Carthage/Sentry-Dynamic.xcframework
+	./scripts/sign_and_zip.sh Carthage/SentrySwiftUI.xcframework
+	./scripts/sign_and_zip.sh Carthage/Sentry-WithoutUIKitOrAppKit.xcframework
```
philipphofmann, line 120: `m`: We could modify the bash script so it takes a parameter that specifies if you want to sign it or not. Then both `build-xcframework` and `build-xcframework-sign` would use the same script and we have less duplication of the code.

.github/workflows/test_sign_xcframework.yml
```diff
@@ -0,0 +1,46 @@
+name: Test Build Signed XCFramework
+run-name: Test Build Signed XCFramework
+
+on:
+  pull_request:
+
+jobs:
+  prepare_framework:
+    runs-on: macos-13-xlarge
```
philipphofmann, line 9: `h`: Please don't use an xlarge runner. Running this on every PR costs easily around 2k $ a month or most likely more. We obviously can spend 2k a month, but only if it makes sense. Let's use a `macos-14` please. We can't use macos-13 as mentioned here
https://github.com/getsentry/sentry-cocoa/blob/aa48954db865ed82e38feb313885dc6cb372cc8b/.github/workflows/build.yml#L101-L105

philipphofmann (commented):

.github/workflows/test_sign_xcframework.yml
```diff
@@ -0,0 +1,46 @@
+name: Test Build Signed XCFramework
+run-name: Test Build Signed XCFramework
+
+on:
+  pull_request:
```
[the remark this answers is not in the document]
itaybre: As mentioned on the other comment, this workflow was only to make sure this worked.
I will delete it and update the release one to generate signed releases before merging

.github/workflows/test_sign_xcframework.yml
philprime: @itaybre you can convert it back to draft

fastlane/Fastfile
```diff
@@ -9,6 +9,16 @@ platform :ios do
   ios_swift_clip_infoplist_path = "./Samples/iOS-Swift/iOS-SwiftClip/Info.plist"
   configuration = if is_ci then 'TestCI' else 'Test' end
 
+  lane :prepare_sign_xcframework do
```
philipphofmann, line 12: ``l`: o align the naming with the make file
```suggestion
  lane :prepare_signed_xcframework do
```

philipphofmann (approved):
Thanks for doing this. After merging this, I think a beta release and manually including the signed XCFramwork via SPM in a sample project is mandatory. We also must communicate this change to the Hybrid SDKs and let them try it with this beta release. Only when all this succeeds, I'm confident enough to ship a normal release.

philippsuess:
Smoke test. I manually included the XCFrameworks from this release https://github.com/getsentry/sentry-cocoa/releases/tag/8.52.0-beta into a sample project and Xcode says the signature is valid


image

image

For reference

PR #5272, opened 2025-05-21
ci: Fix paths for prettier

philipphofmann:
The path filters only applied to the root dir. Now they work for any files of .yml, .yaml, .json or .md in any directory.

#skip-changelog


philprime (approved):
LGTM, thanks for the fix

PR #5275, opened 2025-05-21
Remove extra imports

noahsmartin:
I noticed quite a few swift files imported _SentryPrivate despite not using any objc code, this removes those unnecessary imports 

#skip-changelog


philprime (approved):
LGTM, thanks for the cleanup.

Let's make sure that the CI goes ✅ before merge, because there's a high chance this was added at some point for a reason.

PR #5277, opened 2025-05-21
ci: refactor xcframework build workflows

armcknight:
Separately from https://github.com/getsentry/sentry-cocoa/pull/5218, and also for #4925, try to consolidate the new workflow implementation so it's not duplicated in the 3 places it's needed (build, release, benchmark)

One extra benefit here is that the UI test workflow that build the xcframework seems to have built two variants of it: Sentry-Dynamic and Sentry (the static lib). But, only the static lib was actually being used. So in addition to getting a faster build, we halved the amount of compute being used.

Another is that it's more obvious what each workflow is actually using, instead of having to dig through makefile, then build script logic, to see which xcframework variant is being build for "iOSOnly" or "gameOnly" parameters and which SDKs it's building for, including maccatalyst.

A third is that we don't need to build the mac catalyst version of the xcframework for the app size increase metric. The first comparison after merging to main will be smaller than the last one build with mac catalyst, but then going forward when like is compared to like, this check will be fine again, just run faster.

Improvements in CI times for the jobs using the reusable workflows:
- benchmarking.yml (xcframework build wall time): [before](https://github.com/getsentry/sentry-cocoa/actions/runs/15190771099) 2m14s; [after](https://github.com/getsentry/sentry-cocoa/actions/runs/15203928659) 2m0s
- ui-tests.yml (xcframework build wall time for duplicated sdk test): [before](https://github.com/getsentry/sentry-cocoa/actions/runs/15190771065/job/42722716406) 10m11s , [after](https://github.com/getsentry/sentry-cocoa/actions/runs/15203928669) 1m25s
- release.yml: similar to the speedup mentioned in #4295 for build.yml

Also, an improvement on #5218 is that the reusable workflow invocations with two stages of matrices are grouped by outer jobs instead of all being in one flat list, unordered (new vs old): 

|new|old|
|---|---|
_image|image|

#skip-changelog


itaybre (approved):
LGTM
I would try to remove `Carthage` from the paths at some point, but that can be done later

Issue #4424, opened 2024-10-10
Fix SDK crash in encodeObject


philipphofmann:
### Description

The SDK sometimes crashes with. See also [internal SDK crashes](https://sentry.sentry.io/issues/5058264690/?project=4505469596663808&query=is%3Aunresolved%20sdk.version%3A8.36.0&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=0) for reference. This is our most frequent SDK crash in our latest stable release [8.36.0](https://sentry.sentry.io/issues/?project=4505469596663808&query=is%3Aunresolved%20sdk.version%3A8.36.0&referrer=issue-stream&statsPeriod=14d&stream_index=0).
```
Exception Type: EXC_CRASH (SIGABRT)

Thread 0 Crashed:
0   libsystem_kernel.dylib          0x1e5e2c558         __pthread_kill
1   libsystem_pthread.dylib         0x206ab2114         pthread_kill
2   libsystem_c.dylib               0x1ae636174         abort
3   libsystem_malloc.dylib          0x1b55740e0         malloc_vreport
4   libsystem_malloc.dylib          0x1b5574388         malloc_zone_error
5   libsystem_malloc.dylib          0x1b5573cec         _tiny_check_and_zero_inline_meta_from_freelist
6   libsystem_malloc.dylib          0x1b55592a8         tiny_malloc_from_free_list
7   libsystem_malloc.dylib          0x1b5559894         tiny_malloc_should_clear
8   libsystem_malloc.dylib          0x1b555ca58         szone_malloc_should_clear
9   libsystem_malloc.dylib          0x1b556c6f8         nanov2_allocate_outlined
10  Foundation                      0x1a141367c         -[NSConcreteMutableData initWithLength:]
11  Foundation                      0x1a14134e4         -[NSString(NSStringOtherEncodings) dataUsingEncoding:allowLossyConversion:]
12  Sentry.framework                0x101d37738         encodeObject (Sentry.framework:326)
13  Sentry.framework                0x101d37c54         encodeObject (Sentry.framework:384)
14  Sentry.framework                0x101d37608         +[SentryCrashJSONCodec encode:options:error:] (Sentry.framework:427)
15  Sentry.framework                0x101d497dc         -[SentryCrashScopeObserver toJSONEncodedCString:] (Sentry.framework:149)
[...]

PR #5283, opened 2025-05-23
test: Add tests for encodeObject with NSString

philipphofmann:
GH-4424 reports a crash when calling dataUsingEncoding, which is related to invalid memory. This PR adds two more tests to ensure we can encode Japanese characters and even an invalid JSONString, which makes dataUsingEncoding return nil.

#skip-changelog


philipphofmann:
> It looks like we'd want to transition to UTF16 to support possible foreign language inputs to multiple SDK subsystems

Maybe yes, but not for now 😅. I think UTF8 is good enough for now.

PR #5284, opened 2025-05-23
fix(ci): Add missing update parameter to bump.sh

philprime:
The release workflow uses `make bump` while the craft script uses `./scripts/bump.sh`.
The version bump utility was updated and tested for the Makefil, but not for the script.

This PR adds a fix and extends the test workflows to cover both.
It also adds tests for the `update-package-sha.sh` which is used by `bump.sh`.

#skip-changelog

### DIFF
--- a/.github/workflows/version-bump-util.yml
+++ b/.github/workflows/version-bump-util.yml
@@ -64,8 +64,8 @@ jobs:
       - name: Create fake xcframework for update-package-sha.sh
         run: |
           mkdir -p Carthage
-          echo "<FAKE ZIP>" > Carthage/Sentry.xcframework.zip
-          echo "<FAKE ZIP>" > Carthage/Sentry-Dynamic.xcframework.zip
+          echo "<FAKE STATIC ZIP>" > Carthage/Sentry.xcframework.zip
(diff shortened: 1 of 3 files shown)


.github/workflows/version-bump-util.yml
```diff
@@ -19,18 +19,63 @@ on:
...
+          echo "OLD_VERSION=${OLD_VERSION}" >> $GITHUB_OUTPUT
+
+          NEW_VERSION="100.0.$(date +%Y%m%d%H%M%S)"
+          echo "Generated new version: ${NEW_VERSION}"
+          echo "NEW_VERSION=${NEW_VERSION}" >> $GITHUB_OUTPUT
+
+      - name: Verify OLD_VERSION is defined
+        if: ${{ steps.generate-version-number.outputs.OLD_VERSION == '' }}
+        run: |
+          echo "OLD_VERSION is not defined"
```
philprime, line 61:
```suggestion
          echo "OLD_VERSION is not defined. Make sure this script is reading the version from the correct file."
```

philipphofmann (approved):
LGTM, cause I don't want to block the release. We can follow up on https://github.com/getsentry/sentry-cocoa/pull/5284#pullrequestreview-2864043126.

PR #5287, opened 2025-05-23
ci: remove format code workflow

armcknight:
We recently decided to stop pushing reformat commits to branches in PR as it doesn't cause all required CI checks to rerun, thus presenting a false sense of security in the merge area check display.

#skip-changelog


itaybre (commented):
In this case `Brewfile-ci-format` can also be deleted, right?

PR #5297, opened 2025-05-26
impr: Remove lock in SentryLog.configure

philipphofmann:
## :scroll: Description

Remove unnecessary lock in SentryLog.configure as it's not required because we don't call configure from multiple threads anyways. Furthermore, SentryLog isn't fully thread safe anyways when calling configure and log from multiple threads. We accept these race conditions not leading to crashes because the SentrySDK only calls configure during start and making these methods thread-safe by using locks would add a significant overhead to all log messages. This PR ignores SentryLog for the ThreadSanitizer.

The thread sanitizer is disabled in CI. This brings us one step closer for enabling it again.

Unit tests and running the thread sanitizer locally.

## :pencil: Checklist

-  [x] I added tests to verify the changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Improvements
 
-- Remove unnecessary lock in SentryLog.configure (#5297)
+- Slightly reduce performance impact by removing unnecessary lock in SentryLog.configure (#5297)
 
 ## 8.51.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Improvements
+
+- Remove unnecessary lock in SentryLog.configure (#5297)
+
 ## 8.51.1
 
 ### Fixes
--- a/Sources/Swift/Core/Tools/SentryLog.swift
+++ b/Sources/Swift/Core/Tools/SentryLog.swift
@@ -1,5 +1,15 @@
 import Foundation
 
+/// A note on the thread safety:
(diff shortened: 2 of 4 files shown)


Tests/SentryTests/Helper/SentryLogTests.swift
```diff
@@ -111,5 +111,30 @@ class SentryLogTests: XCTestCase {
...
+                    SentryLog.log(message: "This is another test message", andLevel: SentryLevel.info)
+                }
+
+                expectation.fulfill()
+            }
+        }
+
+        dispatchQueue.activate()
+
+        wait(for: [expectation], timeout: 5.0)
```
itaybre, line 137: nit: I don't think 5 seconds are needed here.
Runs in less than 0.1s on my mac

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - XCFrameworks are now signed (#5271)
 
+### Improvements
+
+- Remove unnecessary lock in SentryLog.configure (#5297)
```
philprime, line 11: `m`: The changelog should mention the SDK user benefit, so maybe add something like:

```suggestion
- Reduce performance impact by removing unnecessary lock in SentryLog.configure (#5297)
```

Sources/Swift/Core/Tools/SentryLog.swift
```diff
@@ -1,5 +1,15 @@
 import Foundation
 
+/// A note on the thread safety:
```
philprime, line 3: Thank you for this inline explanation comment.

philprime (approved):
LGTM, small feedback.

PR #5298, opened 2025-05-26
feat: Add class loaded twice validator

### LINKED ISSUE #4566: Assertion message when Sentry is loaded twice in app
### Description

We fixed loading Sentry integration from the same binary with https://github.com/getsentry/sentry-cocoa/pull/4541, but when Sentry is loaded twice it still could happen that Sentry isn't working correctly. It would be great to detect this in the SDK, and consider logging an error or doing an assert to crash the app in debug. We had a customer who didn't realize they loaded Sentry twice, and it they didn't receive any crash reports cause the client was nil in the `SentryCrashReportSink`

https://github.com/getsentry/sentry-cocoa/blob/166d67abd7e722717fe4cf460c20fcc28412ac1b/Sources/Sentry/SentryCrashReportSink.m#L78-L82

itaybre:
## :scroll: Description

Add a log message when Sentry SDK is found to be present twice + Added a breakpoint

Fixes #4566 

Running duplicated SDK app + Unit Tests cases

## :pencil: Checklist

### DIFF
--- a/Sentry.xcodeproj/project.pbxproj
+++ b/Sentry.xcodeproj/project.pbxproj
@@ -991,6 +991,7 @@
 		D8F8F5572B835BC600AC5465 /* SentryMsgPackSerializerTests.m in Sources */ = {isa = PBXBuildFile; fileRef = D8F8F5562B835BC600AC5465 /* SentryMsgPackSerializerTests.m */; };
 		D8FC98AB2CD0DAB30009824C /* BreadcrumbExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = D8FC98AA2CD0DAAC0009824C /* BreadcrumbExtension.swift */; };
 		D8FFE50C2703DBB400607131 /* SwizzlingCallTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D8FFE50B2703DAAE00607131 /* SwizzlingCallTests.swift */; };
+		F4046F9B2DE4F6C2000B855B /* LoadValidator.swift in Sources */ = {isa = PBXBuildFile; fileRef = F4046F9A2DE4F6C2000B855B /* LoadValidator.swift */; };
 		FA034AC82DD3DB4900FE3107 /* SentryIntegrationProtocol.h in Headers */ = {isa = PBXBuildFile; fileRef = FA034AC72DD3DB4900FE3107 /* SentryIntegrationProtocol.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		FA67DCC12DDBD4C800896B02 /* SentryLog+Configure.swift in Sources */ = {isa = PBXBuildFile; fileRef = FA67DCC02DDBD4C800896B02 /* SentryLog+Configure.swift */; };
 		FA67DCF42DDBD4EA00896B02 /* SentryExperimentalOptions.swift in Sources */ = {isa = PBXBuildFile; fileRef = FA67DCF12DDBD4EA00896B02 /* SentryExperimentalOptions.swift */; };
@@ -2192,6 +2193,7 @@
 		D8F8F5562B835BC600AC5465 /* SentryMsgPackSerializerTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = SentryMsgPackSerializerTests.m; sourceTree = "<group>"; };
 		D8FC98AA2CD0DAAC0009824C /* BreadcrumbExtension.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BreadcrumbExtension.swift; sourceTree = "<group>"; };
 		D8FFE50B2703DAAE00607131 /* SwizzlingCallTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SwizzlingCallTests.swift; sourceTree = "<group>"; };
+		F4046F9A2DE4F6C2000B855B /* LoadValidator.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LoadValidator.swift; sourceTree = "<group>"; };
 		FA034AC72DD3DB4900FE3107 /* SentryIntegrationProtocol.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; name = SentryIntegrationProtocol.h; path = Public/SentryIntegrationProtocol.h; sourceTree = "<group>"; };
 		FA67DCC02DDBD4C800896B02 /* SentryLog+Configure.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "SentryLog+Configure.swift"; sourceTree = "<group>"; };
 		FA67DCC22DDBD4EA00896B02 /* NSLock.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NSLock.swift; sourceTree = "<group>"; };
@@ -4391,6 +4393,7 @@
 				FA67DCED2DDBD4EA00896B02 /* UIImageHelper.swift */,
 				FA67DCEE2DDBD4EA00896B02 /* UrlSanitized.swift */,
 				FA67DCEF2DDBD4EA00896B02 /* URLSessionTaskHelper.swift */,
+				F4046F9A2DE4F6C2000B855B /* LoadValidator.swift */,
 			);
 			path = Tools;
 			sourceTree = "<group>";
@@ -5297,6 +5300,7 @@
 				03F84D3527DD4191008FE43F /* SentryThreadHandle.cpp in Sources */,
 				0A2D8DA9289BC905008720F6 /* SentryViewHierarchy.m in Sources */,
 				D84D2CDD2C2BF7370011AF8A /* SentryReplayEvent.swift in Sources */,
+				F4046F9B2DE4F6C2000B855B /* LoadValidator.swift in Sources */,
 				D8BC28CC2BFF78220054DA4D /* SentryRRWebTouchEvent.swift in Sources */,
 				8EA1ED0B2668F8C400E62B98 /* SentryUIViewControllerSwizzling.m in Sources */,
 				7B98D7CF25FB650F00C5A389 /* SentryWatchdogTerminationTrackingIntegration.m in Sources */,
--- a/Sources/Swift/Core/Tools/LoadValidator.swift
+++ b/Sources/Swift/Core/Tools/LoadValidator.swift
@@ -1,0 +1,69 @@
+//
+//  LoadValidator.swift
+//  Sentry
+//
+//  Created by Itay Brenner on 26/5/25.
+//  Copyright © 2025 Sentry. All rights reserved.
+//
+
+import Foundation
+import MachO
+
+@objc
+class LoadValidator: NSObject {
+    // Any class should be fine
+    static let targetClassName = "PrivateSentrySDKOnly"
+    
+    @objc
+    class func validateLoadedOnce() {
+        let imageCount = _dyld_image_count()
+        
+        var imagesWhereSentryIsFound: [String] = [String]()
+        
+        // Iterate over all images to make sure none of them contains a Sentry implementation twice
+        for i in 0..<imageCount {
+            guard let cImageName = _dyld_get_image_name(i) else { continue }
+            
+            // Skip Apple Frameworks
+            let imagePath = String(cString: cImageName)
+            
+            let systemLibraryPath = "/usr/lib/"
+#if targetEnvironment(simulator)
+            let ignoredPath = "/Library/Developer/CoreSimulator/Profiles/Runtimes/"
+#else
+            let ignoredPath = "/System/Library/"
+#endif
+            if imagePath.contains(ignoredPath) || imagePath.hasPrefix(systemLibraryPath) {
+                continue
+            }
+            
+            var classCount: UInt32 = 0
+            if let classNames = objc_copyClassNamesForImage(cImageName, &classCount) {
+                for j in 0..<Int(classCount) {
+                    let name = String(cString: classNames[j])
+                    if name.contains(self.targetClassName) {
+                        imagesWhereSentryIsFound.append(imagePath)
+                        break
+                    }
+                }
+                free(classNames)
+            }
+        }
+        
+        if imagesWhereSentryIsFound.count > 1 {
+#if DEBUG
+            var message = ["❌ Sentry SDK was loaded multiple times in the binary ❌"]
+            message.append("⚠️ This can cause undefined behavior, crashes, or duplicate reporting.")
+            message.append("Ensure the SDK is linked only once")
+            imagesWhereSentryIsFound.forEach { path in
+                message.append("  - \(path)")
+            }
+            SentryLog.log(message: message.joined(separator: "\n"), andLevel: SentryLevel.debug)
+            // Raise a debugger breakpoint
+            raise(SIGTRAP)
+#else
+            SentryLog.log(message: "❌ Sentry SDK was loaded multiple times in the binary ❌", andLevel: SentryLevel.warning)
+#endif
+        }
+    }
+}
--- a/Sources/Sentry/SentrySDK.m
+++ b/Sources/Sentry/SentrySDK.m
@@ -225,6 +225,8 @@ static NSDate *_Nullable startTimestamp = nil;
     // thread could lead to deadlocks.
     SENTRY_LOG_DEBUG(@"Starting SDK...");
 
+    [LoadValidator validateLoadedOnce];
+
 #if defined(DEBUG) || defined(SENTRY_TEST) || defined(SENTRY_TEST_CI)
     SENTRY_LOG_DEBUG(@"Configured options: %@", options.debugDescription);
 #endif // defined(DEBUG) || defined(SENTRY_TEST) || defined(SENTRY_TEST_CI)


Sources/Swift/Core/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,67 @@
+//
+//  LoadValidator.swift
+//  Sentry
+//
+//  Created by Itay Brenner on 26/5/25.
+//  Copyright © 2025 Sentry. All rights reserved.
+//
+
```
philprime, line 8:
```suggestion
```

Sources/Swift/Core/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,67 @@
...
+        }
+        
+        if imagesWhereSentryIsFound.count > 1 {
+            var message = ["❌ Sentry SDK was loaded multiple times in the binary ❌"]
+            message.append("⚠️ This can cause undefined behavior, crashes, or duplicate reporting.")
+            message.append("Ensure the SDK is linked only once")
+            imagesWhereSentryIsFound.forEach { path in
+                message.append("  - \(path)")
+            }
+            print(message.joined(separator: "\n"))
```
philprime, line 60: `m`: Couldn't we use `SentryLog.fatal(...)` here? It would also allow us to assert log messages using `TestLogOutput`.

philprime (commented):
Thanks @itaybre for working on this! Left a couple of comments.

Tests/DuplicatedSDKTest/Unit Tests/Unit_Tests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+final class TestLog: XCTestCase {
+    func testDuplicatedLoadMessageOnSDKInit() throws {
+      
+      let output = captureStandardOutput {
+        SentrySDK.start(options: Options())
+      }
+      
+      XCTAssertTrue(output.contains("❌ Sentry SDK was loaded multiple times in the binary ❌"))
+    }
```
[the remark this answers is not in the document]
itaybre: Because the test for this uses [the xcframework](https://github.com/getsentry/sentry-cocoa/blob/main/.github/workflows/ui-tests.yml#L226) (because we need to link it twice) I didn't have access to the internal classes.
I can try setting it up without the xcframework and see if it links twice somehow

Sources/Swift/Core/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,59 @@
...
+#if DEBUG
+            // Raise a debugger breakpoint
+            raise(SIGTRAP)
```
noahsmartin, line 55: This might end up being very noisy since some apps have no choice but to include the SDK twice when they use third party frameworks that themselves pull in the SDK

Sources/Swift/Core/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,59 @@
...
+            let ignoredPath = "/Library/Developer/CoreSimulator/Profiles/Runtimes/"
+#else
+            let ignoredPath = "/System/Library/"
+#endif
+            if imagePath.contains(ignoredPath) || imagePath.hasPrefix(systemLibraryPath) {
+                continue
+            }
+            
+            var classCount: UInt32 = 0
+            if let classNames = objc_copyClassNamesForImage(cImageName, &classCount) {
```
noahsmartin, line 33: Should this whole thing be done only in debug builds? Just to prevent overhead for release builds

Sources/Swift/Core/Tools/LoadValidator.swift
armcknight: Check out e.g. `testSetUserBeforeStartingSDK_LogsFatalMessage` in case you hadn't seen that. In any case, we probably shouldn't use `print`, but also what happens with the logger if there are multiple copes of the sentry framework loaded 🤔

Sources/Swift/Core/Tools/LoadValidator.swift
armcknight: The other thing is that `DEBUG` as a compiler directive may not correlate to `SentryOptions.debug` or to a debugger even being attached (we have facilities to check for that btw). I assumed the "debug" being referred to here was _Sentry_ debug mode:
image

Sources/Swift/Core/Tools/LoadValidator.swift
philippsuess: Yes exactly. You can use the TestLogOutput as we do in testSetUserBeforeStartingSDK_LogsFatalMessage

Sources/Swift/Core/Tools/LoadValidator.swift
philippsuess: We must run this on a BG thread. The overhead of this code can be significant. Iterating over all classes of only one image in a large project can take a couple of ms. We also look for subclasses in SentrySubClassFinder, which is the heavy part, but my point is that this code can add some overhead; see https://github.com/getsentry/sentry-cocoa/blob/ba8a46b7ac0863cf2c8e45b5386ffb4ad6c141d4/Sources/Sentry/SentrySubClassFinder.m#L62-L64

I also think we should only do this for debug builds and when the `option.debug = true`.

Sources/Swift/Core/Tools/LoadValidator.swift
philippsuess: Yes, I think a log message should be a good start for now. We only have one known occurrence of this problem.

Sources/Swift/Core/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,59 @@
...
+import MachO
+
+@objc
+class LoadValidator: NSObject {
+    // Any class should be fine
+    static let targetClassName = "PrivateSentrySDKOnly"
+    
+    @objc
+    class func validateLoadedOnce() {
+        let imageCount = _dyld_image_count()
```
philippsuess: `h`: The `SentryCrashBinaryImageCache` uses `sentry_dyld_register_func_for_add_image`, which also the `SentryBinaryImageCache` uses. I think it would be better to do this check inside `SentryBinaryImageCache.binaryImageAdded` so we can also check binary images loaded later on.

philprime:
I believe we should update the PR branch before continuing review

Sentry.xcodeproj/project.pbxproj
```diff
@@ -2680,7 +2689,6 @@
 		6327C5C91EB8A783004E799B = {
 			isa = PBXGroup;
 			children = (
-				84A899032E2090E9009A551E /* CLAUDE.md */,
```
armcknight: I assume this was a conflict resolution, but please don't remove this from the project

Tests/DuplicatedSDKTest/Unit Tests/Unit_Tests.swift
```diff
@@ -0,0 +1,93 @@
...
+        options.debug = true
+        options.diagnosticLevel = .debug
+        SentrySDK.start(options: options)
+        
+        // Force loading another library to trigger validation
+        let path = Bundle(for: type(of: self)).bundlePath
+        let fullpath = "\(path)/../../Frameworks/ModuleA.framework/ModuleA"
+        guard dlopen(fullpath, RTLD_NOW) != nil else {
+            let error = String(cString: dlerror())
+            fatalError("Could not open framework: \(error)")
```
armcknight: Please package this into a Swift error and throw it from the test function so it doesn't crash the test run

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
+@_implementationOnly import _SentryPrivate
+import Darwin
+import Foundation
+import MachO
+
+@objc
+@_spi(Private) public final class LoadValidator: NSObject {
+    // Any class should be fine
+    static let targetClassName = "PrivateSentrySDKOnly"
```
armcknight: I hate string literals lol
```suggestion
    static let targetClassName = NSStringFromClass(LoadValidator.self)
```

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+                        }
+                    }
+                    free(classNames)
+                }
+            }
+        }
+    }
+    
+    /**
+     * Returns a pointer to a function inside the __TEXT segment of the binary containing this class
```
armcknight: The underscores make the comment bolded, which it shouldn't
```suggestion
     * Returns a pointer to a function inside the `__TEXT` segment of the binary containing this class
```

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+            imageName.withCString { cImageName in
+                if let classNames = objcRuntimeWrapper.copyClassNames(forImage: cImageName, amount: &classCount) {
+                    for j in 0..<Int(classCount) {
+                        guard let className = classNames[j] else {
+                            continue
+                        }
+                        let name = String(cString: UnsafeRawPointer(className).assumingMemoryBound(to: UInt8.self))
+                        if name.contains(self.targetClassName) {
+                            if name == self.targetClassName && isCurrentImageContainingLoadValidator {
+                                // Skip our implementation of `PrivateSentrySDKOnly`
```
armcknight:
```suggestion
                                // Skip the implementation of the class we are using as a proxy for being loaded that exists in the same binary that this instance of LoadValidator was loaded in
```

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+        internalValidateSDKPresenceIn(image, objcRuntimeWrapper: objcRuntimeWrapper, qos: .userInitiated) { duplicateFound in
+            result = duplicateFound
+            semaphore.signal()
+        }
+        semaphore.wait()
+        return result
+    }
+    
+    class private func internalValidateSDKPresenceIn(_ image: SentryBinaryImageInfo, objcRuntimeWrapper: SentryObjCRuntimeWrapper, qos: DispatchQoS.QoSClass = .background, resultHandler: ((Bool) -> Void)? = nil) {
+        DispatchQueue.global(qos: qos).async {
```
armcknight: Can/should we use SentryDispatchQueueWrapper? We've run into a bunch of problems in the past, specifically with SentryFileManager, where work was dispatched to some queue and then executed while later tests ran, and it wrecked the test suite for a long time 😕 A lot of what you see now in TestSentryDispatchQueueWrapper came out of that.

Sources/Swift/Tools/LoadValidator.swift
itaybre: Will probably use a different class, but good idea

Tests/DuplicatedSDKTest/Unit Tests/Unit_Tests.swift
philippsuess: You could also use XCTFail

Sources/Sentry/Public/SentryBinaryImageInfo.h
```diff
@@ -0,0 +1,14 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentryBinaryImageInfo : NSObject
+@property (nonatomic, strong) NSString *name;
+@property (nonatomic, copy) NSString *UUID;
+@property (nonatomic) uint64_t vmAddress;
+@property (nonatomic) uint64_t address;
+@property (nonatomic) uint64_t size;
+
+@end
```
philippsuess: `h`: Please don't make that public for everyone. This should be in `include/HybridPublic`.

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -83,6 +84,11 @@ - (void)binaryImageAdded:(const SentryCrashBinaryImage *)image
 
         [_cache insertObject:newImage atIndex:left];
     }
+
+    if ([SentrySDK.currentHub getClient].options.debug == YES) {
```
philippsuess: `m`: I would passing if we're running in debug or not via the `start` method, instead of checking this in here. SentrySDK.startWithOptions calls `SentryBinaryImageCache.start`. We have access to the options in there. So we can then add a parameter isDebug to `SentryBinaryImageCache.start` easily.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+import MachO
+
+@objc
+@_spi(Private) public final class LoadValidator: NSObject {
+    // Any class should be fine
+    static let targetClassName = "PrivateSentrySDKOnly"
+    
+    @objc
+    @_spi(Private) public class func validateSDKPresenceIn(_ image: SentryBinaryImageInfo) {
+        internalValidateSDKPresenceIn(image, objcRuntimeWrapper: SentryDefaultObjCRuntimeWrapper.sharedInstance())
```
philippsuess: `m`: I would prefer passing the objcRuntimeWrapper as a dependency via the constructor for `LoadValidator` so we clearly communicate its dependencies. We would like to get rid of sharedInstance / singletons as much as possible, cause we've been bitten by that quite often in our unit tests, because it's a risk for adding global state that's hard to undo.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+    @discardableResult class func validateSDKPresenceInSync(_ image: SentryBinaryImageInfo, objcRuntimeWrapper: SentryObjCRuntimeWrapper) -> Bool {
+        var result = false
+        let semaphore = DispatchSemaphore(value: 0)
+        internalValidateSDKPresenceIn(image, objcRuntimeWrapper: objcRuntimeWrapper, qos: .userInitiated) { duplicateFound in
+            result = duplicateFound
+            semaphore.signal()
+        }
+        semaphore.wait()
+        return result
+    }
```
philippsuess: `m`: When you inject the `DispatchQueue` as a dependency via the constructor you can easily swap it out in the tests and you can move this method to the tests. It's basically what @armcknight suggested in https://github.com/getsentry/sentry-cocoa/pull/5298/files#r2211847357.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+            let systemLibraryPath = "/usr/lib/"
+#if targetEnvironment(simulator)
+            let ignoredPath = "/Library/Developer/CoreSimulator/Profiles/Runtimes/"
+#else
+            let ignoredPath = "/System/Library/"
+#endif
+            let imageName = image.name
+            guard !imageName.contains(ignoredPath) && !imageName.hasPrefix(systemLibraryPath) else {
+                return
+            }
```
philippsuess: `m`: What about doing that check before moving to async code via the dispatch queue. These checks are almost free, and then we would avoid enqueuing loads of work items which is a bit of an overhead.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+        }
+        semaphore.wait()
+        return result
+    }
+    
+    class private func internalValidateSDKPresenceIn(_ image: SentryBinaryImageInfo, objcRuntimeWrapper: SentryObjCRuntimeWrapper, qos: DispatchQoS.QoSClass = .background, resultHandler: ((Bool) -> Void)? = nil) {
+        DispatchQueue.global(qos: qos).async {
+            var duplicateFound = false
+            defer {
+                resultHandler?(duplicateFound)
```
philippsuess, line 31: `l`: It seems like you're using the resultHandler so you can test this code. If we pass in the `DispatchQueue` via the constructor, I think we can remove this code.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+                            var message = ["❌ Sentry SDK was loaded multiple times in the binary ❌"]
+                            message.append("⚠️ This can cause undefined behavior, crashes, or duplicate reporting.")
+                            message.append("Ensure the SDK is linked only once, found classes in image paths: \(imageName)")
+                            SentrySDKLog.warning(message.joined(separator: "\n"))
+                            duplicateFound = true
+                            
+                            break
+                        }
+                    }
+                    free(classNames)
```
philippsuess: `l`: We could move this up to a defer block directly below when initializing the class names to ensure we don't exit without calling free in the future.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+            let isCurrentImageContainingLoadValidator = (loadValidatorAddressValue >= imageAddress) && (loadValidatorAddressValue < (imageAddress + imageSize))
+
+            var classCount: UInt32 = 0
+            imageName.withCString { cImageName in
+                if let classNames = objcRuntimeWrapper.copyClassNames(forImage: cImageName, amount: &classCount) {
+                    for j in 0..<Int(classCount) {
+                        guard let className = classNames[j] else {
+                            continue
+                        }
+                        let name = String(cString: UnsafeRawPointer(className).assumingMemoryBound(to: UInt8.self))
```
philippsuess: `m`: We have to be very careful when dealing with the classes here. We had weird crashes and bugs when calling `NSClassFromString` or accessing the actual classes in SentrySubClassFinder:
https://github.com/getsentry/sentry-cocoa/blob/bb78c914bce81dad34caaee83b94a4e68ead4717/Sources/Sentry/SentrySubClassFinder.m#L53-L75

What we do now is OK, but maybe it's worth adding a comment to reference the `SentrySubClassFinder` and explaining the risks.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+                        if name.contains(self.targetClassName) {
+                            if name == self.targetClassName && isCurrentImageContainingLoadValidator {
+                                // Skip our implementation of `PrivateSentrySDKOnly`
+                                continue
+                            }
```
philippsuess: `l`: I think we could reduce the nesting a bit:
```suggestion
                        if name == self.targetClassName && isCurrentImageContainingLoadValidator {
                            // Skip our implementation of `PrivateSentrySDKOnly`
                            continue
                        }

                        if name.contains(self.targetClassName) {
```

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+                            var message = ["❌ Sentry SDK was loaded multiple times in the binary ❌"]
```
philippsuess: `l`: Isn't it in the **same** binary?
```suggestion
                            var message = ["❌ Sentry SDK was loaded multiple times in the same binary ❌"]
```

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+                            message.append("⚠️ This can cause undefined behavior, crashes, or duplicate reporting.")
+                            message.append("Ensure the SDK is linked only once, found classes in image paths: \(imageName)")
```
philippsuess: `l`: It's only one path:
```suggestion
                            message.append("Ensure the SDK is linked only once, found classes in image path: \(imageName)")
```

philippsuess: `l`: I think we should add the class name that we found to the log message.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+import Foundation
+import MachO
+
+@objc
+@_spi(Private) public final class LoadValidator: NSObject {
+    // Any class should be fine
+    static let targetClassName = "PrivateSentrySDKOnly"
+    
+    @objc
+    @_spi(Private) public class func validateSDKPresenceIn(_ image: SentryBinaryImageInfo) {
```
philippsuess: `m`: I'm a bit confused by the method name. This method and this class validate that the SDK isn't loaded twice. Does it really validate the SDK presence? It prints a warning when it finds the SentrySDK loaded in two binaries.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+            let imageName = image.name
+            guard !imageName.contains(ignoredPath) && !imageName.hasPrefix(systemLibraryPath) else {
+                return
+            }
+            
+            let loadValidatorAddress = self.getCurrentFrameworkTextPointer()
+            let imageAddress = image.address
+            let imageSize = image.size
+            let loadValidatorAddressValue = UInt(bitPattern: loadValidatorAddress)
+            let isCurrentImageContainingLoadValidator = (loadValidatorAddressValue >= imageAddress) && (loadValidatorAddressValue < (imageAddress + imageSize))
```
philippsuess, line 40: `m`: Couldn't we add an early return if this is true? Below we continue when this is true and `name == self.targetClassName`.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+    
+    /**
+     * This synchronous version is intended to be used in tests.
+     * It uses a higher QoS and wait for the result
+     */
+    @discardableResult class func validateSDKPresenceInSync(_ image: SentryBinaryImageInfo, objcRuntimeWrapper: SentryObjCRuntimeWrapper) -> Bool {
+        var result = false
+        let semaphore = DispatchSemaphore(value: 0)
+        internalValidateSDKPresenceIn(image, objcRuntimeWrapper: objcRuntimeWrapper, qos: .userInitiated) { duplicateFound in
```
philippsuess: `m`: I think this can be `qos: .background`. I think it's OK that the log message appears a bit later. This could be a bit expensive and it's better to let it run with low priority in the background.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,90 @@
...
+                        let name = String(cString: UnsafeRawPointer(className).assumingMemoryBound(to: UInt8.self))
+                        if name.contains(self.targetClassName) {
+                            if name == self.targetClassName && isCurrentImageContainingLoadValidator {
+                                // Skip our implementation of `PrivateSentrySDKOnly`
+                                continue
+                            }
+                            var message = ["❌ Sentry SDK was loaded multiple times in the binary ❌"]
+                            message.append("⚠️ This can cause undefined behavior, crashes, or duplicate reporting.")
+                            message.append("Ensure the SDK is linked only once, found classes in image paths: \(imageName)")
+                            SentrySDKLog.warning(message.joined(separator: "\n"))
```
philippsuess: `h`: I think this should be an error, because when this is really the case, users must see it, because they might not receive any crashes. If it's a warning, it could be overlooked.
```suggestion
                            SentrySDKLog.error(message.joined(separator: "\n"))
```

Sources/Sentry/Public/SentryBinaryImageInfo.h
itaybre: I can try again, but I couldn't, make it visible for Swift when I placed it in `HybridPublic`.
I might have done something wrong

Sources/Swift/Tools/LoadValidator.swift
itaybre: This is a test only class, so it doesn't really matter the qos since it should not be used in prod

Sources/Swift/Tools/LoadValidator.swift
philippsuess: Follow up: Maybe my comment is a case of premature optimization. But I think we could use hasPrefix also for the `ignoredPath` because checking a prefix is cheaper than doing a `contains`.

Tests/DuplicatedSDKTest/Unit Tests/Unit_Tests.swift
itaybre: Just used XCTFail

Sources/Swift/Tools/LoadValidator.swift
itaybre: The result handler is so I can get the result, it is in fact only for testing but unrelated to the dispatchqueue

Sources/Swift/Tools/LoadValidator.swift
itaybre: Sounds good

Sources/Swift/Tools/LoadValidator.swift
itaybre: We could find a different name in the same image.
For example when static linking and someone added a copy of sentry with a slightly different name (like a prefix)

Sources/Swift/Tools/LoadValidator.swift
itaybre: Renamed to `checkForDuplicatedSDK`

Sources/Sentry/include/HybridPublic/SentryBinaryImageInfo.h
```diff
@@ -0,0 +1,14 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentryBinaryImageInfo : NSObject
```
philippsuess: `m`: Now that we simply pass the required properties as parameters to the LoadValidator, do we still need this extra class? If not, we could undo these changes and keep it as is.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,81 @@
...
+import Foundation
+import MachO
+
+@objc
+@_spi(Private) public final class LoadValidator: NSObject {
+    // Any class should be fine, ObjC classes are better
+    static let targetClassName = NSStringFromClass(SentryDependencyContainerSwiftHelper.self)
+    
+    // This function is used to check for duplicated SDKs in the binary.
+    // Since `SentryBinaryImageInfo` is not public and only available through the Hybrid SDK, we use the expanded parameters.
```
philippsuess, line 12: Thanks for the comment 🥇

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,81 @@
...
+    @objc
+    @_spi(Private) public class func checkForDuplicatedSDK(imageName: String, imageAddress: NSNumber, imageSize: NSNumber) {
+        internalCheckForDuplicatedSDK(imageName, imageAddress.uint64Value, imageSize.uint64Value,
+                                      objcRuntimeWrapper: SentryDefaultObjCRuntimeWrapper.sharedInstance(),
+                                      dispatchQueueWrapper: SentryDispatchQueueWrapper())
```
philippsuess: `m`: I would prefer not having a default parameter for the `SentryDispatchQueueWrapper`, because we now don't pass it in from the `SentryBinaryImageCache`, and we create an extra instance. Instead, we should reuse the one from the `SentryDependencyContainer`. We could pass the `SentryDispatchQueueWrapper` in via a init method and do the same for the objcRuntimeWrapper. Then the `LoadValidator` doesn't have to know how to create these dependencies and the code clearly communicates it's dependencies.

Sources/Swift/Tools/LoadValidator.swift
philippsuess: Yes, but many users set `options.debug = true` when setting up Sentry the first time and there are also multiple users keeping this flag on for their debug builds. If this slows them down, they will notice it. I quickly checked for how many images we enqueue an async work item and it's only a few. I think it should be OK to not use a lower priority and simply schedule this to the dispatch queue of the dispatch queue wrapper. It would be good to add this info as a code comment.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+    
+    class func internalCheckForDuplicatedSDK(_ imageName: String, _ imageAddress: UInt64, _ imageSize: UInt64, objcRuntimeWrapper: SentryObjCRuntimeWrapper, dispatchQueueWrapper: SentryDispatchQueueWrapper, resultHandler: ((Bool) -> Void)? = nil) {
+        let systemLibraryPath = "/usr/lib/"
+#if targetEnvironment(simulator)
+        let ignoredPath = "/Library/Developer/CoreSimulator/Profiles/Runtimes/"
+#else
+        let ignoredPath = "/System/Library/"
+#endif
+        guard !imageName.contains(ignoredPath) && !imageName.hasPrefix(systemLibraryPath) else {
```
philippsuess: `m`: `hasPrefix` should be close to O(1) in most cases while `contains` can be O(n). I think we should be able to use `hasPrefix` here.
```suggestion
        guard !imageName.hasPrefix(ignoredPath) && !imageName.hasPrefix(systemLibraryPath) else {
```

Tests/SentryTests/Swift/Core/Tools/LoadValidatorTests.swift
```diff
@@ -0,0 +1,197 @@
...
+        // Act
+        let validationResult = LoadValidator.checkForDuplicatedSDKInSync(imageName,
+                                                                         defaultImageAddress,
+                                                                         defaultImageSize,
+                                                                         objcRuntimeWrapper: testObjCRuntimeWrapper,
+                                                                         dispatchQueueWrapper: dispatchQueueWrapper)
+        
+        // Assert
+        XCTAssertFalse(validationResult, "Validation should skip for system libraries")
+        XCTAssertFalse(getClassListCalled, "ObjectiveC Wrapper shouldd not be called for a system library")
```
philippsuess: `l`: The same typo exists multiple times.
```suggestion
        XCTAssertFalse(getClassListCalled, "ObjectiveC Wrapper should not be called for a system library")
```

Tests/SentryTests/Swift/Core/Tools/LoadValidatorTests.swift
```diff
@@ -0,0 +1,197 @@
...
+        XCTAssertFalse(testOutput.loggedMessages.contains { $0.contains("❌ Sentry SDK was loaded multiple times") })
+        XCTAssertEqual(dispatchQueueWrapper.dispatchAsyncInvocations.count, 0)
+    }
+ 
+#if targetEnvironment(simulator)
+    func testValidateSDKPresenceIn_SimulatorPath_DoesNotValidate() {
```
philippsuess: `m`: We don't need to limit the test to the simulator environment. This test shouldn't validate no matter the environment.
```suggestion
    func testValidateSDKPresenceIn_SimulatorPath_DoesNotValidate() {
```

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,81 @@
...
+                                      dispatchQueueWrapper: SentryDispatchQueueWrapper())
+    }
+    
+    class func internalCheckForDuplicatedSDK(_ imageName: String, _ imageAddress: UInt64, _ imageSize: UInt64, objcRuntimeWrapper: SentryObjCRuntimeWrapper, dispatchQueueWrapper: SentryDispatchQueueWrapper, resultHandler: ((Bool) -> Void)? = nil) {
+        let systemLibraryPath = "/usr/lib/"
+#if targetEnvironment(simulator)
+        let ignoredPath = "/Library/Developer/CoreSimulator/Profiles/Runtimes/"
+#else
+        let ignoredPath = "/System/Library/"
+#endif
```
philippsuess: `m`: What about always checking both paths, no matter the environment? The checks are cheap and do no harm.

Tests/SentryTests/Swift/Core/Tools/LoadValidatorTests.swift
```diff
@@ -0,0 +1,197 @@
...
+                                                              objcRuntimeWrapper: SentryObjCRuntimeWrapper,
+                                                              dispatchQueueWrapper: SentryDispatchQueueWrapper) -> Bool {
+        var result = false
+        let semaphore = DispatchSemaphore(value: 0)
+        internalCheckForDuplicatedSDK(imageName, imageAddress, imageSize, objcRuntimeWrapper: objcRuntimeWrapper,
+                                      dispatchQueueWrapper: dispatchQueueWrapper) { duplicateFound in
+            result = duplicateFound
+            semaphore.signal()
+        }
+        semaphore.wait()
```
philippsuess: `m`: We could replace this now with test expectations and use a timeout. Otherwise, we could block the tests indefinitely.

sdk_api.json
```diff
@@ -43468,13 +43468,27 @@
         "declKind": "Import",
         "moduleName": "Sentry"
       },
+      {
+        "kind": "Import",
+        "name": "Darwin",
+        "printedName": "Darwin",
+        "declKind": "Import",
+        "moduleName": "Sentry"
+      },
```
philippsuess: `m`: I think we should be able to undo all the changes in the api.json files, because the current solution doesn't need any API changes.

philipphofmann (commented):
Thanks for including all the feedback. We're getting close to LGTM.

Sources/Swift/Tools/LoadValidator.swift
itaybre: Technically yes, but it doesn't hurt to check the real path only

sdk_api.json
itaybre: This api change seems needed, probably due to LoadValidator imports

philippsuess: Ah that makes sense thanks.

Sources/Swift/Tools/LoadValidator.swift
itaybre: Ready!

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -83,6 +88,16 @@ - (void)binaryImageAdded:(const SentryCrashBinaryImage *)image
 
         [_cache insertObject:newImage atIndex:left];
     }
+
+    if (self.isDebug) {
+        // Only validate if debug is enabled
```
philippsuess: `l`: Maybe we should mention this code adds a bit of of overhead to explain why we only do it in debug

Sources/Swift/Tools/LoadValidator.swift
philippsuess: It would be great to add this info as a comment.

Sources/Swift/Tools/LoadValidator.swift
```diff
@@ -0,0 +1,78 @@
...
+@_spi(Private) public final class LoadValidator: NSObject {
+    // Any class should be fine, ObjC classes are better
+    static let targetClassName = NSStringFromClass(SentryDependencyContainerSwiftHelper.self)
+    
+    // This function is used to check for duplicated SDKs in the binary.
+    // Since `SentryBinaryImageInfo` is not public and only available through the Hybrid SDK, we use the expanded parameters.
+    @objc
+    @_spi(Private) public class func checkForDuplicatedSDK(imageName: String, imageAddress: NSNumber, imageSize: NSNumber, dispatchQueueWrapper: SentryDispatchQueueWrapper) {
+        internalCheckForDuplicatedSDK(imageName, imageAddress.uint64Value, imageSize.uint64Value,
+                                      objcRuntimeWrapper: SentryDependencyContainerSwiftHelper.objcRuntimeWrapper(),
```
philippsuess, line 16: `m`: For consistency I think we should also remove the default parameter for the `objcRuntimeWrapper` and pass it in via the method.

philipphofmann (approved):
Thanks @itaybre, LGTM

Sources/Swift/Tools/LoadValidator.swift
itaybre: Unfortunately I can't, due to ObjC and Swift limitations, I can't use a public argument with SentryObjCRuntimeWrapper in Swift.

philippsuess: 🤦 I think we already had that discussion

Issue #5301, opened 2025-05-27
Include React Native text and images classes in the Replay options out of the box


krystofwoldrich:
### Description

`sentry-cocoa` SDK can be initialized manually in RN applications, since we are unable to detect RN text and images automatically, we should include the classes out of the box so RN SDK init is not required and well as users don't have to specify the RN classes manually.

```
"RNSVGSvgView",
"RCTImageView",
"RCTTextView",
"RCTParagraphComponentView"
```

Related to https://github.com/getsentry/sentry-react-native/issues/3608

PR #5302, opened 2025-05-27
feat(replay): Redact React Native text and images by default

krystofwoldrich:
## :scroll: Description

The cocoa SDK can be manually initialized in RN applications with or without the RN SDK. This can lead to unexpected behavior where RN text and images are not masked correctly.

This PR adds the RN text and images classes to be redacted when `maskAllText` and `maskAllImages` are used.

I have not included the RN SVG class as cocoa doesn't have `maskAllVectors`. 

fixes: https://github.com/getsentry/sentry-cocoa/issues/5301

Added unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Core/Tools/ViewCapture/UIRedactBuilder.swift
+++ b/Sources/Swift/Core/Tools/ViewCapture/UIRedactBuilder.swift
@@ -75,6 +75,9 @@ class UIRedactBuilder {
         
         if options.maskAllText {
             redactClasses += [ UILabel.self, UITextView.self, UITextField.self ]
+            // These classes are used by React Native to display text.
+            // We are including them here to avoid leaking text from RN apps with manually initialized sentry-cocoa.
+            redactClasses += ["RCTTextView", "RCTParagraphComponentView"].compactMap(NSClassFromString(_:))
         }
         
         if options.maskAllImages {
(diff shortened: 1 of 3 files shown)


Sources/Swift/Core/Tools/ViewCapture/UIRedactBuilder.swift
```diff
@@ -75,6 +75,9 @@ class UIRedactBuilder {
         
         if options.maskAllText {
             redactClasses += [ UILabel.self, UITextView.self, UITextField.self ]
+            // These classes are used by React Native to display text.
+            // We are including them here to avoid leaking text from RN apps with manually initialized sentry-cocoa.
+            redactClasses += ["RCTTextView", "RCTParagraphComponentView"].compactMap(NSClassFromString(_:))
```
romtsn, line 80: can we make `redactClasses` a Set here actually (I think it's an array now) to avoid duplicates? Also seems like redactClasses could have strings instead of Classes to make it even lighter, but Classes is also fine I guess

philprime: See line 108, the classes are mapped to a Set of ObjectIdentifier there.

This entire building of identifiers seems to be quite unclean, but I would propose cleaning it up in another PR.

PR #5304, opened 2025-05-27
ci: Remove unused script boot simulator

philipphofmann:
Nobody uses the script. We have git history if we need it again.


philprime (approved):
LGTM, thanks for cleaning up our repository!

philprime:
Experience shows, it was in fact used. We'll have to roll this back.

https://github.com/getsentry/sentry-cocoa/actions/runs/15298894151/job/43034485831

Issue #5058, opened 2025-04-07
Sentry catches wrong stacktrace of NSException


ShiBody:
### Platform

macOS

### Environment

Production

### Installed

Manually

### Version

8.40.0

### Xcode Version

16.1

### Steps to Reproduce

1. Raise a NSGenericException with customized fail reason
2. Debug in Xcode could see stacktrace in Exception Backtrace clearly. We have 58 elements in NSExcaption.

![Image](https://github.com/user-attachments/assets/8f46bd2a-2874-4123-b005-4080b5844868)

3. While after sending to Sentry, we could only see an invalid stacktrace
https://webex.sentry.io/issues/6504297906/events/6a3940bbdb38430b8a84f9177eb2d854/?project=4507095483809792

<img width="1659" alt="Image" src="https://github.com/user-attachments/assets/bb820931-c1b0-4f60-8892-f0eb8b949de4" />

### Expected Result

Same stacktrace as Exception Backtrace shows

### Actual Result

An invalid runloop stacktrace captured.

philprime:
Hi @ShiBody, thanks for opening up this issue.

Looking at the crash report in Sentry, have you tried to upload debug symbols to see if it helps with symbolicating the full stack trace? While I agree that the amount of frames seems off, testing the crash with debug symbols might help with pinpointing the issue.

philprime:
Thank you for getting back to us, we'll have to investigate this.

ShiBody:
Hi team, do you have any update for this bug?

philprime:
Hi @ShiBody, sorry no update yet.

philipphofmann:
@ShiBody, as this is on macOS, have you tried the [Uncaught NSException Reporting Option](https://docs.sentry.io/platforms/apple/guides/macos/usage/#uncaught-nsexception-reporting-option)?

```
options.enableUncaughtNSExceptionReporting = true
```

philipphofmann:
I just found out that we don't mention this in the getting started page on the docs for macOS, so I opened a PR https://github.com/getsentry/sentry-docs/pull/13546.

ShiBody:
> [@ShiBody](https://github.com/ShiBody), as this is on macOS, have you tried the [Uncaught NSException Reporting Option](https://docs.sentry.io/platforms/apple/guides/macos/usage/#uncaught-nsexception-reporting-option)?
> 
> ```
> options.enableUncaughtNSExceptionReporting = true
> ```

[...]

PR #5306, opened 2025-05-27
feat: Use callstack from NSException

itaybre:
## :scroll: Description

Add a new method to collect crashes on `_crashOnException` and report the stacktrace from the exception instead of the current thread

Exception captured on `_crashOnException` do no collect the proper stacktraces/
Fixes: #5058

On macOS App

## :pencil: Checklist

## TODO:
- [ ] Update changelog
- [ ] Add test case


Sources/Sentry/Public/SentrySDK.h
```diff
@@ -222,6 +222,18 @@ SENTRY_NO_INIT
                 withScopeBlock:(void (^)(SentryScope *scope))block
     NS_SWIFT_NAME(capture(exception:block:));
 
+#if TARGET_OS_OSX
```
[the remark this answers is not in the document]
itaybre: Could be used anywhere, but the issue doesn't happens outside of macOS since there is no _onCrashException for catching exceptions

Sources/Sentry/SentryExceptionWrapper.m
```diff
@@ -0,0 +1,80 @@
...
+                                          inAppExcludes:options.inAppExcludes];
+    SentryCrashStackEntryMapper *crashStackEntryMapper =
+        [[SentryCrashStackEntryMapper alloc] initWithInAppLogic:inAppLogic];
+
+    return crashStackEntryMapper;
+}
+
+@end
+
+#endif
```
armcknight, line 80:
```suggestion
#endif // TARGET_OS_OSX
```

Sources/Sentry/include/SentryExceptionWrapper.h
```diff
@@ -0,0 +1,21 @@
...
+ * This is a helper class to identify exceptions that should use the stacktrace within
+ */
+@interface SentryExceptionWrapper : NSException
+
+- (instancetype)initWithException:(NSException *)exception;
+- (NSArray<SentryThread *> *)buildThreads;
+
+@end
+
+#endif
```
armcknight, line 19:
```suggestion
#endif // TARGET_OS_OSX
```

armcknight (approved):
Works and LGTM, just some small suggestions from me.

philipphofmann (commented):

Sources/Sentry/include/SentryUseNSExceptionCallstackWrapper.h
```diff
@@ -0,0 +1,31 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+#if TARGET_OS_OSX
+
+@class SentryThread;
+
+@protocol ExceptionProtocol <NSObject>
```
philippsuess: `h`: Please prefix this with Sentry otherwise we could run into naming collisions, but this comment could become obsolete by addressing https://github.com/getsentry/sentry-cocoa/pull/5306#discussion_r2121371769 first.

Tests/SentryTests/SentryClientTests.swift
```diff
@@ -2094,6 +2094,38 @@ class SentryClientTest: XCTestCase {
...
+        XCTAssertEqual(actual.threads?[0].current, true)
+        XCTAssertEqual(actual.threads?[0].isMain, true)
+        // Make sure the stacktrace is not empty
+        XCTAssertGreaterThan(actual.threads?[0].stacktrace?.frames.count ?? 0, 1)
+        // We will need to update it if the test class / module changes
+        let testMangledName = "$s11SentryTests0A10ClientTestC011testCaptureA16WrappedExceptionyyKF"
+        let frameWithTestFunction = actual.threads?[0].stacktrace?.frames.first { frame in
+            frame.function == testMangledName
+        }
+        XCTAssertNotNil(frameWithTestFunction)
```
philippsuess: `m`: If this assertion fails, it could be a bit tricky to find out what's going on. The assert message could point out that it didn't find any frame with the mangled name `testMangledName`.

Tests/SentryTests/SentryCrashExceptionApplicationTests.swift
```diff
@@ -0,0 +1,58 @@
+@testable import Sentry
+import SentryTestUtils
+import XCTest
+
+#if os(macOS)
+
+class SentryCrashExceptionApplicationHelperTests: XCTestCase {
```
philippsuess, line 7: thanks for the tests

Tests/SentryTests/SentryCrashExceptionApplicationTests.swift
```diff
@@ -0,0 +1,58 @@
...
+    
+    override func tearDown() {
+        super.tearDown()
+        
+        clearTestState()
+        resetUserDefaults()
+        UserDefaults.standard.removeObject(forKey: "NSApplicationCrashOnExceptions")
+    }
+    
+    func testCrashOnException() throws {
```
philippsuess, line 17: `l`: FYI, started using // Arrange // Act // Assert in tests. It's not a hard requirement and we don't have it written down somewhere, but if you like the pattern, it would be great to use it for your tests.

Sources/Sentry/SentryUseNSExceptionCallstackWrapper.m
```diff
@@ -0,0 +1,80 @@
...
+@end
+
+@implementation SentryUseNSExceptionCallstackWrapper
+
+- (instancetype)initWithException:(NSObject<ExceptionProtocol> *)exception
+{
+    if (self = [super initWithName:exception.name
+                            reason:exception.reason
+                          userInfo:exception.userInfo]) {
+        _originalException = exception;
```
philippsuess: `m`: It would be great if we don't keep a reference to the original exception so it can be garbage collected and we remove a risk for potential future memory leaks. We only need the `callStackReturnAddresses` below, so we could copy these here into our own array or already convert the callStackReturnAddresses directly to SentryFrames. Another advantage is that we don't need the `ExceptionProtocol`.

Tests/SentryTests/SentryCrashExceptionApplicationTests.swift
itaybre: I usually use that too, maybe my spacing was odd.

- Setup SUT
- Run
- Assert

philippsuess: No it's not about the spacing. We actually put the comments into the tests: https://github.com/getsentry/sentry-cocoa/blob/654ec5fa902bb5b6e435154c0e25a0e9521ad71d/Tests/SentryTests/SentrySessionTests.swift#L123-L132

philipphofmann (approved):
Thank you @itaybre. LGTM

Tests/SentryTests/SentryCrashExceptionApplicationTests.swift
itaybre: Ohh, I can do that

PR #5311, opened 2025-05-28
Commit API interface

noahsmartin:
#skip-changelog


.github/workflows/api-stability.yml
```diff
@@ -21,49 +21,24 @@ jobs:
...
         run: |
-          xcrun --sdk iphoneos swift-api-digester \
-            -diagnose-sdk \
-            -o result.json \
-            -input-paths out_base.json \
-            -input-paths out_head.json \
-            -json \
-            -v
-          cat result.json
+          if cmp -s "sdk_api_base.json" "sdk_api.json"; then
```
philipphofmann, line 31: `m`: Maybe it makes sense to use diff instead of cmp to get a hint on what changed, but I'm unsure how useful the output would be in CI.

.github/workflows/api-stability.yml
```diff
@@ -21,49 +21,24 @@ jobs:
...
+            echo "No API changes detected."
+          else
+            echo "❌ API changes are detected, run ./scripts/update-api.sh and commit the changes."
```
philipphofmann, line 34: `m`: Maybe it makes sense to add a command to the makefile for this, which would be easier to type and find.

philipphofmann (commented):
I'm happy trying this approach and see how it goes. If it causes problems we can always rollback.

PR #5314, opened 2025-05-28
style: Run linter only on modified swift files on pre hook

itaybre:
## :scroll: Description

Add a new Makefile script to run linter only on modified files

Running `time git commit -m "Any message"`

Previous time: 46s
Actual time: 26s

## :pencil: Checklist

#skip-changelog

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,9 +55,8 @@ format-swift:
 	swiftlint --fix
 
 format-swift-staged:
-	git diff --cached --diff-filter=d --name-only | grep '\.swift$$' | while IFS= read -r filename; do \
-		swiftlint --fix "$$filename"; \
-	done
+	echo "Running swiftlint --fix on staged files"
+	swiftlint --fix $(shell git diff --cached --diff-filter=d --name-only | grep '\.swift$$')
 
 # Format Markdown
 format-markdown:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -54,6 +54,10 @@ format-clang:
 format-swift:
 	swiftlint --fix
 
+format-swift-staged:
+	echo "Running swiftlint --fix on staged files"
+	swiftlint --fix $(shell git diff --cached --diff-filter=d --name-only | grep '\.swift$$')
```
philipphofmann, line 59: `h`: This doesn't seem to make a difference on my machine. I get almost the same log output as when using `make format-swift` and it takes the same amount of time.

philipphofmann (commented):
I think that is not improving anything sadly on my machine.

Makefile
itaybre: Did you had any swift file staged?
If no files are staged, it will run on all files

philipphofmann: Ah yes, then it works. Can we maybe skip this when there are no staged swift files? My PRs often don't touch swift files.

itaybre: Let me check that, I think the pre commit hook already does this

itaybre: yeah, the command shouldn't run in that case
image

PR #5315, opened 2025-05-28
refactor: Add factory for SentryClient to dependency container

### LINKED ISSUE #5296: Use file manager of dependency container for SentryClient
### Description

I noticed today that we have a `SentryFileManager` in the `SentryDependencyContainer` which is used for most cases, but not for the `SentryClient` because it initializes it's own file manager in `SentrySDK`:

https://github.com/getsentry/sentry-cocoa/blob/1148435da4f6a10e06e56f669ea588813e6ad1ed/Sources/Sentry/SentrySDK.m#L240-L242

https://github.com/getsentry/sentry-cocoa/blob/1148435da4f6a10e06e56f669ea588813e6ad1ed/Sources/Sentry/SentryClient.m#L80-L96

It got introduced [here](https://github.com/getsentry/sentry-cocoa/pull/2424/files#diff-47eb3eb3ea905acecc12679b7d5eac4688cb64f396c5e02f3213aefc5d0a639aR84-R88) before being last changed [here](https://github.com/getsentry/sentry-cocoa/pull/3163/files#diff-47eb3eb3ea905acecc12679b7d5eac4688cb64f396c5e02f3213aefc5d0a639aR86-R88).

Found this while working on [this PR feedback comment](https://github.com/getsentry/sentry-cocoa/pull/5242/files#r2104413738) as `moveContextFileToPreviousContextFile` is invoked from `SentrySDK`. 

We should reuse the file manager of the dependency container so that the client uses the same as all other components. Right now the file manager in the dependency container is the one created for `SentrySDK.options`, and even if the options change, the file manager does not, leading to internal inconsistencies.

The SDK resets the dependency container on close. This is only a problem if you call start with a different DSN multiple times in a row. And if you don't use the static API. This could also be a reason with flaky tests not cleaning up correctly.

Image
Image

philprime:
- This PR is a draft of changes while working on #5242.
- The draft is moving the initialization of the SentryClient to the dependency container.
- I added a factory to create file manager using options.
- This clearly shows that the file manager is a global object that relies on `SentrySDK.options`. While this is set during bootstrapping of the SDK, it can lead to inconsistencies if other parts rely on different options (especially relevant in isolated tests).
- If we want to use factory pattern for the file manager, we should consider keeping weak references to the file manager per DSN (encapsulated in an hash) so we can reuse them.
- The nullability of the client is also inconsistent.

Closes #5296

@philipphofmann @armcknight I'm looking for feedback here on what you think of the approach of using a file manager factory where possible (e.g. in the ANR integration in the diff).

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryDependencyContainer.m
+++ b/Sources/Sentry/SentryDependencyContainer.m
@@ -1,6 +1,7 @@
 #import "SentryANRTrackerV1.h"
 
 #import "SentryBinaryImageCache.h"
+#import "SentryClient.h"
 #import "SentryDispatchFactory.h"
 #import "SentryDispatchQueueWrapper.h"
 #import "SentryDisplayLinkWrapper.h"
@@ -12,6 +13,7 @@
 #import "SentryNSProcessInfoWrapper.h"
 #import "SentryNSTimerFactory.h"
 #import "SentryOptions+Private.h"
+#import "SentryOptions.h"
 #import "SentryRandom.h"
 #import "SentrySDK+Private.h"
 #import "SentrySwift.h"
@@ -172,14 +174,27 @@ static NSObject *sentryDependencyContainerInstanceLock;
 - (SentryFileManager *)fileManager SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOCK
 {
     SENTRY_LAZY_INIT(_fileManager, ({
-        NSError *error;
-        SentryFileManager *manager = [[SentryFileManager alloc] initWithOptions:SentrySDK.options
-                                                                          error:&error];
-        if (manager == nil) {
-            SENTRY_LOG_DEBUG(@"Could not create file manager - %@", error);
+        SentryOptions *options = SentrySDK.options;
+        if (options == nil) {
+            SENTRY_LOG_FATAL(
+                @"SentryDependencyContainer.fileManager called before SentrySDK.options was set.");
         }
-        manager;
+        [self getFileManagerForOptions:options];
     }));
+}
+
+- (SentryFileManager *_Nullable)getFileManagerForOptions:(SentryOptions *)options
+    SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOCK
+{
+    NSError *error;
+    SentryFileManager *manager =
+        [[SentryFileManager alloc] initWithOptions:options
+                              dispatchQueueWrapper:self.dispatchQueueWrapper
+                                             error:&error];
+    if (manager == nil) {
+        SENTRY_LOG_FATAL(@"Could not create file manager - %@", error);
+    }
+    return manager;
 }
 
 - (SentryAppStateManager *)appStateManager SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOCK
@@ -346,5 +361,10 @@ static NSObject *sentryDependencyContainerInstanceLock;
 }
 
 #endif // SENTRY_HAS_METRIC_KIT
+
+- (SentryClient *_Nullable)getClientWithOptions:(SentryOptions *)options
+{
+    return [[SentryClient alloc] initWithOptions:options];
+}
 
 @end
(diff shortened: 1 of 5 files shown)


Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -172,16 +174,29 @@ - (instancetype)init
...
+{
+    NSError *error;
+    SentryFileManager *manager =
+        [[SentryFileManager alloc] initWithOptions:options
+                              dispatchQueueWrapper:self.dispatchQueueWrapper
+                                             error:&error];
+    if (manager == nil) {
+        SENTRY_LOG_FATAL(@"Could not create file manager - %@", error);
+    }
+    return manager;
```
philipphofmann, line 197: `h`: The way the synchronization currently works in the SentryFileManager, we must ensure only having one SentryFileManager instance per DSN, as you already pointed out in the PR description. Otherwise, we end up with race conditions when accessing the files.
I think your idea of keeping weak references using the DSN with the key and the file manager as a value could work.

Sources/Sentry/SentrySDK.m
```diff
@@ -237,7 +237,8 @@ + (void)startWithOptions:(SentryOptions *)options
     startInvocations++;
     startTimestamp = [SentryDependencyContainer.sharedInstance.dateProvider date];
 
-    SentryClient *newClient = [[SentryClient alloc] initWithOptions:options];
+    SentryClient *newClient =
+        [SentryDependencyContainer.sharedInstance getClientWithOptions:options];
```
philipphofmann, line 241: `m`: I'm unsure if that solves anything, because any user can also manually init a client and we must support that use case. Using the SentryDependencyContainer internally increases complexity, and I don't see the benefit. What drove you adding this?

philipphofmann (commented):
> I'm looking for feedback here on what you think of the approach of using a file manager factory where possible (e.g. in the ANR integration in the diff).

I think that's OK. We have a couple of locations in the code simply just pulling the file manager from the dependency container. Now you need to pass in the Options, which you can easily get via SentrySDK.options but at least then you know where they come from.

Sources/Sentry/SentrySDK.m
philipphofmann: > is this an actual SDK requirement, or just due to the fact that we don't want to change the current public API in our implementation to avoid a major rev?

It's part of the [unified API](https://develop.sentry.dev/sdk/miscellaneous/unified-api/), which we deprecated some time ago, because we're merging the hub and the scope. Once we do the next major we can revisit if we want to keep the client public. There are some use cases for initializing your own client, which I can't recall from the top of my head. Most users on Cocoa interact with the static API.

philprime:
Linking https://github.com/getsentry/sentry-cocoa/pull/5242#discussion_r2120400966 as it is related.

Sources/Sentry/SentrySDK.m
philprime: To summarize what we are trying to fix:

- Our current Dependency Injection + Factory Pattern implementation is quite messy because of opportunistic refactoring.
- Some instances are managed by integration instances
- Some are part of the dependency injection container
- Some are created using factories
- Some instances in the static container are bounds to global options, but are used with local options (which can be different).
- For local instances we fetch the dependencies from the container, then initialize it locally. This is duplicated in the code base, instead of delegating it to a factory with access to the dependency container (e.g. [here](https://github.com/getsentry/sentry-cocoa/blob/076c6ae915ca2a56b3fc0bf3cb82e6f5e1ba29ea/Sources/Sentry/SentrySessionReplayIntegration.m#L232-L236) and [here](https://github.com/getsentry/sentry-cocoa/blob/076c6ae915ca2a56b3fc0bf3cb82e6f5e1ba29ea/Sources/Sentry/SentrySessionReplayIntegration.m#L380-L384).

For reference, if instances are properly managed in a dependency container the entire integration can be delegate to testable and mockable factories.

This technical documentation of Swinject might be helpful in seeing the advantages:

https://github.com/Swinject/Swinject/tree/master/Documentation

philprime:
Closing this PR because I am planning to do a full overhaul of the dependency container which will also cover these topics.

PR #5316, opened 2025-05-28
test: Remove clearing state in serialization tests

philipphofmann:
The SentrySerializationTests have no global state and don't need to clear the test state.


philipphofmann:
Ups, CI is telling me we need this.

Issue #3637, opened 2024-02-14
Stop profiles when SentrySDK.startWithOptions is not called


armcknight:
`h`: What if a user never calls `startWithOptions`? Then the tracer and launch profile would run forever if I'm not mistaken. The auto-generated transactions have a timeout of 30 seconds

https://github.com/getsentry/sentry-cocoa/blob/ac8fd5353ed60b08a13a4ad891b7ca374e258f01/Sources/Sentry/SentryTracer.m#L160-L162

but only if they have an idle timeout or `waitForChildren`

https://github.com/getsentry/sentry-cocoa/blob/ac8fd5353ed60b08a13a4ad891b7ca374e258f01/Sources/Sentry/SentryTracer.m#L231-L234

The `SentryTracerConfiguration.defaultConfiguration` doesn't have an idle timeout or `waitForChildren`. One way to fix this would be to enable `waitForChildren`, or to add an extra flag that `isAutoGeneratedTransaction` check.

This is an edge case; I would be OK with fixing it with an upcoming PR.

_Originally posted by @philipphofmann in https://github.com/getsentry/sentry-cocoa/pull/3621#discussion_r1486022257_

kahest:
@armcknight is this still relevant?

armcknight:
@kahest honestly i don't think it's relevant. this is a misconfiguration of the SDK as far as I'm concerned, to call start once and configure launch profiling, and then on the subsequent launch to not call start. Maybe we can message better about this in docs.

kahest:
@armcknight makes sense, feel free to close this or keep it as a reminder to add docs

PR #5318, opened 2025-05-29
impr(profiling): always remove launch profile config after starting

armcknight:
I realized after rereading https://github.com/getsentry/sentry-cocoa/issues/3637#issuecomment-2787837554 that even if we consider a launch profile unable to be stopped as a misconfiguration we can't currently avoid at runtime, this leads to every launch afterwards also starting the launch profiling. We should instead consider not starting the SDK as disabling launch profiling, the same as if the SDK had _never_ been started in an app.

It also increases safety such that if the app crashes between the time a launch profile starts (especially if it's a crash _due_ to the profiler), and a call to SentrySDK.startWithOptions would've disabled launch profiling, launch profiling won't keep starting again on every subsequent launch.

To achieve this we should always remove the config file once we're done with it _for that launch_. Configuring to disable launch profiling already removed any file present; now, starting the launch profiler also removes it.

This also helps with issues like https://github.com/getsentry/sentry-cocoa/issues/5366

#skip-changelog


Samples/iOS-Swift/iOS-Swift-UITests/ProfilingUITests.swift
```diff
@@ -152,72 +137,104 @@ extension ProfilingUITests {
...
+        case ui // aka "continuous v2"
+    }
+
+    func performTest(profileType: ProfilingType, lifecycle: SentryProfileOptions.SentryProfileLifecycle? = nil) throws {
+        try launchAndConfigureSubsequentLaunches(shouldProfileThisLaunch: false, shouldProfileNextLaunch: true, profileType: profileType, lifecycle: lifecycle)
+        try launchAndConfigureSubsequentLaunches(terminatePriorSession: true, shouldProfileThisLaunch: true, shouldProfileNextLaunch: false, profileType: profileType, lifecycle: lifecycle)
+    }
+
+    fileprivate func setAppLaunchParameters(_ profileType: ProfilingUITests.ProfilingType, _ lifecycle: SentryProfileOptions.SentryProfileLifecycle?, _ shouldProfileNextLaunch: Bool) {
+        app.launchArguments.append(contentsOf: [
```
philippsuess, line 145: `m`: I'm not a real fan of using methods with multiple parameters for tests. It harms the readability of the tests. When I look at one single test I need to jump around a lot in the file to find out how the test works. I think some code duplication is acceptable for tests, because when you look at one test case you know what's going on. Tests only need to change when the requirements change. They usually don't change often. Also, with this approach, you end up having logic in tests that can contain bugs, and then the tests don't test what you want. Tests can be repetitive but they should be stupid.

That's a personal preference and maybe a bigger discussion. This is up to you and shouldn't block the PR.

Sources/Sentry/Profiling/SentryLaunchProfiling.m
```diff
@@ -307,6 +307,8 @@
     sentry_launchTracer = [[SentryTracer alloc] initWithTransactionContext:context
                                                                        hub:nil
                                                              configuration:config];
+
+    removeAppLaunchProfilingConfigFile();
```
philippsuess: `m`: I think it would be great to add the major points of the PR description and explain here as a code comment why we call `removeAppLaunchProfilingConfigFile`.

philipphofmann (approved):
Doesn't this deserve a changelog entry? Some comments, but nothing blocking. LGTM

Samples/iOS-Swift/iOS-Swift-UITests/ProfilingUITests.swift
armcknight: I agree with the readability aspect. Usually I'm doing this because I want to make it easy to see that the entirety of the search space is covered, which is harder to see in individual functions. But I will look into unrolling them in separate work.

Tests/SentryProfilerTests/SentryApplaunchProfilingMalformedConfigFileTests.swift
```diff
@@ -0,0 +1,289 @@
...
+#if os(iOS) || os(macOS) || targetEnvironment(macCatalyst)
+class SentryAppLaunchProfilingMalformedConfigFileTests: XCTestCase {
+    override func setUp() {
+        super.setUp()
+        removeAppLaunchProfilingConfigFile()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        clearTestState()
```
philippsuess, line 13: `m`: Do we really need to call `clearTestState` here? I would prefer calling the specific clearing methods, cause clearTestState does so much and could have potential side effects.

Tests/SentryProfilerTests/SentryApplaunchProfilingMalformedConfigFileTests.swift
```diff
@@ -0,0 +1,289 @@
...
+        try corruptedData.write(to: configURL)
+
+        // Verify file exists before the call
+        XCTAssertTrue(appLaunchProfileConfigFileExists())
+
+        // Verify that sentry_appLaunchProfileConfiguration returns nil for corrupted file
+        XCTAssertNil(sentry_appLaunchProfileConfiguration())
+
+        // Call _sentry_nondeduplicated_startLaunchProfile
+        _sentry_nondeduplicated_startLaunchProfile()
```
philippsuess: `l`: I think some of these code comments are redundant.
```suggestion
        _sentry_nondeduplicated_startLaunchProfile()
```

Tests/SentryProfilerTests/SentryApplaunchProfilingMalformedConfigFileTests.swift
```diff
@@ -0,0 +1,289 @@
...
+        XCTAssertFalse(appLaunchProfileConfigFileExists())
+    }
+
+    func testMalformedConfigFile_TraceProfilingMissingProfilesRand_DoesNotStartProfilingAndRemovesFile() throws {
+        // Create a config file for trace profiling but missing profiles sample rand
+        let configDict: [String: Any] = [
+            kSentryLaunchProfileConfigKeyProfilesSampleRate: 1.0,
+            kSentryLaunchProfileConfigKeyTracesSampleRate: 1.0,
+            kSentryLaunchProfileConfigKeyTracesSampleRand: 0.5
+            // Missing: kSentryLaunchProfileConfigKeyProfilesSampleRand
```
philippsuess, line 137: Super useful comment. Thanks 👍

Tests/SentryProfilerTests/SentryApplaunchProfilingMalformedConfigFileTests.swift
```diff
@@ -0,0 +1,289 @@
+import SentryTestUtils
+import XCTest
+
+#if os(iOS) || os(macOS) || targetEnvironment(macCatalyst)
+class SentryAppLaunchProfilingMalformedConfigFileTests: XCTestCase {
```
philippsuess, line 5: I prefer these test a lot. Although they have duplicated code, each test case has everything that it requires. I can read one test by itself without having to jump around in the code and I know what it's testing. Thanks 👍

armcknight:
For some reason tests are failing In CI but not locally, i'm debugging.

Samples/iOS-Swift/iOS-Swift-UITests/BaseUITest.swift
```diff
@@ -46,10 +46,17 @@ extension BaseUITest {
...
         // proxy for whether this is running in CI.
-        if !isDebugging() {
+        if !isDebugging() && activateBeforeLaunch {
+            // activate() appears to drop launch args and environment variables, so save them beforehand and reset them before subsequent calls to launch()
+            let launchArguments = app.launchArguments
+            let launchEnvironment = app.launchEnvironment
+
             // App prewarming can sometimes cause simulators to get stuck in UI tests, activating them
             // before launching clears any prewarming state.
             app.activate()
```
armcknight, line 56: @noahsmartin I was unable to resolve the UI test failures until adding this facility to avoid the double-launch. I think what was happening is that the files written were wiped on each launch. I didn't see another way to deal with all the implications of an individual launch for these tests, since they're sensitive to every launch of the app, and try to control that exactly to launch each app twice per test method.

PR #5319, opened 2025-05-29
meta: ignore code-formatting commits for git blame

armcknight:
Start collecting commit hashes for commits that only formatted code. This often interferes with my search for the origins of a piece of code. For instance, I was looking at `SentryCrashExceptionApplication._crashOnException`:
image

after making this change: 
image

Locally, you need to run `git config blame.ignoreRevsFile .git-blame-ignore-revs` but i was hesitant to add that to our `make init`. But this also gets taken up by GitHub: 
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

There are a few other commits I've seen like the one I already added to the file, they can just be added as they're encountered.

#skip-changelog


armcknight:
hah, talk about arcane tricks. i didnt notice that part on the page, but i also didn't foresee needing it so didn't go looking. what's a situation where you'd want to see the change from a format-only commit in the blame?

armcknight:
I think that as adding to this file would require a review, we can avoid accidentally adding a commit hash to this that has format and functional changes intermixed, which should always be discouraged anyways.

Evolution due solely to code format is uninteresting at best but more often is a noisy roadblock, as I described above.

PR #5320, opened 2025-05-29
Experimental options in Swift

noahsmartin:
As described in the SPM migration doc, this moves Swift types from the objc interface into extensions, making it visible from Swift in SPM. It unfortunately requires the force cast, but I think is very safe and the public API does not change, which is why none of the call-sites of these properties had to change.

#skip-changelog

### DIFF
--- a/Sources/Sentry/Public/SentryEvent.h
+++ b/Sources/Sentry/Public/SentryEvent.h
@@ -27,7 +27,7 @@ NS_SWIFT_NAME(Event)
 /**
  * This will be set by the initializer.
  */
-@property (nonatomic, strong) SentryId *eventId;
+@property (nonatomic, strong) NSObject *_swiftEventId;
 
 /**
  * Message of the event.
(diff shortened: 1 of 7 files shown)


philipphofmann (commented):
I am still a bit skeptical that this won't cause any breaking changes, also for ObjC.

What about only trying this with the experimental options first and doing a release. If you're using an experimental feature, it's more OK to be broken. If we don't hear any negative feedback, we're more confident and can proceed. Also when we know its working I think we should explain this briefly in the internal develop docs, cause it's not obvious.

Also I think that somebody could access the `_` properties by mistake and the we crash when force casting. I think we should make it a bit harder to shoot yourself into the foot.

noahsmartin:
Thanks for taking a look @philipphofmann I agree that moving just one property is safer, so I split that into this other PR: https://github.com/getsentry/sentry-cocoa/pull/5329 I'm still not too worried about this changing the API, explained why in the other PR description. Also made the rest of the changes in the other PR. I'll close this one for now in favor of the other one and can re-open this one when we are ready to update the other property

PR #5329, opened 2025-05-31
Experimental options in Swift

noahsmartin:
As described in the SPM migration doc, this moves a Swift type from the objc interface into an extension, making it visible from Swift in SPM. It unfortunately requires the force cast, but I think is very safe and the public API does not change, which is why none of the call-sites of these properties had to change.

The objc property is `readonly` so we don't have to worry about it being set anywhere to an incorrect object, other than in `SentryOptions.m` itself which only sets it one time.

I verified the objc interface does not change using the sample app, which access this property when configuring options. It's part of the iOS-ObjectiveC sample already

#skip-changelog


Sources/Sentry/Public/SentryOptions.h
```diff
@@ -782,11 +781,9 @@ typedef void (^SentryProfilingConfigurationBlock)(SentryProfileOptions *_Nonnull
...
 @property (nonatomic, copy) NSString *spotlightUrl;
 
-/**
- * This aggregates options for experimental features.
- * Be aware that the options available for experimental can change at any time.
- */
-@property (nonatomic, readonly) SentryExperimentalOptions *experimental;
+// Do not use this directly, instead use the non-underscored `experimental` property that is
+// defined through a Swift extension.
+@property (nonatomic, readonly) NSObject *_swiftExperimentalOptions;
```
philippsuess, line 786: Once we know this approach works, we could add a typealias for the NSObject, or define a macro to clearly wrap these NSObject wrapper properties. We don't need to do this in this PR.

philipphofmann (approved):
Let's give this approach a shot and see how it goes.

I think we still should add a changelog entry mentioning this as an improvement, so we know for reference in the changelog when we changed this.

PR #5330, opened 2025-05-31
Add more SPI

noahsmartin:
A follow up from https://github.com/getsentry/sentry-cocoa/pull/5307 making a few swift types that I missed `public` and adding spi where necessary (only when the enclosing class wasn't already spi)

#skip-changelog


philipphofmann (approved):
LGTM

philippsuess:
If these JSON diffs continue to be so huge, I think we should revisit the decision to commit that JSON to the repo and simply compare the diffs of the different commits instead. 
Screenshot 2025-06-03 at 15 47 45

PR #5331, opened 2025-05-31
Add thread info helper

noahsmartin:
Moves some swift imports out of ObjC++ by defining a helper that does not import Swift to perform the C++ code

#skip-changelog


Sources/Sentry/ThreadInfoHelper.mm
```diff
@@ -0,0 +1,17 @@
+#import "ThreadInfoHelper.h"
+#import "SentryThread.h"
+#include "SentryThreadHandle.hpp"
+
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+
+@implementation ThreadInfoHelper
```
philprime: `m`: We should prefix SDK classes with `Sentry`, because the prefix is used in our SDK Crash Detection to identify frames of the SDK.

```suggestion
@implementation SentryThreadInfoHelper
```

Sources/Sentry/ThreadInfoHelper.mm
```diff
@@ -0,0 +1,17 @@
...
+
++ (SentryThread *)threadInfo
+{
+    const auto threadID = sentry::profiling::ThreadHandle::current()->tid();
+    return [[SentryThread alloc] initWithThreadId:@(threadID)];
+}
```
philprime: `m`: Couldn't this live in `SentryThread` instead of being a separate class?

philprime (approved):
Left some minor comments, feel free to proceed.

Sources/Sentry/ThreadInfoHelper.mm
noahsmartin: Yes good call! I moved it to a private category on `SentryThread`

PR #5339, opened 2025-06-03
ci: Check Swift and Clang formatted code

philipphofmann:
Let CI fail if there is wrongly formatted Swift or code formatted with clang-format.

#skip-changelog

### DIFF
--- a/.github/workflows/lint-clang-formatting.yml
+++ b/.github/workflows/lint-clang-formatting.yml
@@ -28,13 +28,12 @@ jobs:
     runs-on: macos-15
     steps:
       - uses: actions/checkout@v4
-      
       - name: Install tooling
         run: make init-ci-format
+
       - name: Format with Clang
         run: make format-clang
     
-
       - name: Check for Formatting Changes
         run: |
           if [[ -n "$(git status --porcelain)" ]]; then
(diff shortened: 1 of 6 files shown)


.github/workflows/lint-clang-formatting.yml
```diff
@@ -0,0 +1,48 @@
+name: Check Code Formatting
```
philprime, line 1: `m`: Are you going to add workflows for `format-markdown`, `format-json`, and `format-yaml` aswell?

philprime (approved):
LGTM, thank you.

.github/workflows/lint-clang-formatting.yml
philippsuess: No we already have linters failing for these: 
* https://github.com/getsentry/sentry-cocoa/blob/main/.github/workflows/lint-dprint.yml
* https://github.com/getsentry/sentry-cocoa/blob/main/.github/workflows/lint-shellcheck.yml

philippsuess: Do you think the naming could be better? If yes, how would you name the workflows?

philprime: I think the naming is alright, I just missed that we already have the other two checks.

PR #5073, opened 2025-04-09
fix(session-replay): add masking improvements

philprime:
## :scroll: Description

This is a WIP PR to review analysis attempts and improvements for session replay masking

## :bulb: Motivation and Context

See #4864 

#skip-changelog

PR #5341, opened 2025-06-03
fix(session-replay): Use frame rate for cache max size

philprime:
- Adds missing configuration of the replay maker frame rate
- Uses the configured frame rate for calculating the maximum cache size of the replay maker frame rate.

The frame rate is not publicly exposed, but it can cause issues when changing the frame rate internally, e.g. during testing.

I did not add additional tests in this PR, because the session replay integration dependency architecture was poorly designed and it would require a larger refactoring. I confirmed the changes by manually debugging set values.

These changes are derived from #5073

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -10,7 +10,8 @@ coverage:
   round: down
 
 ignore:
-  - "Tests/*"
+  - "SentryTestUtilsTests/**"
+  - "Tests/**"
 
 comment:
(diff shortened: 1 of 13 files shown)


philipphofmann (approved):
LGTM, thanks for the comments and explaining why you didn't add any tests, cause that's the first thing I would have asked 😅

PR #5342, opened 2025-06-03
fix(session-replay): Use timestamp of screenshot for frames

philprime:
- When configuring a higher frame rate for session replay, the async dispatch of frame rendering caused delays in processing.
- The delay was then affecting the timestamps of frames as the date provider was used at a later stage than the actual screenshot
- This PR changes the timestamp of a frame to the timestamp when the screenshot was taken, which should also better align with other data points such as breadcrumbs.

### DIFF
--- a/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
+++ b/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
@@ -15,7 +15,6 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
         
     private let _outputPath: String
     private var _totalFrames = 0
-    private let dateProvider: SentryCurrentDateProvider
     private let processingQueue: SentryDispatchQueueWrapper
     private let assetWorkerQueue: SentryDispatchQueueWrapper
     private var _frames = [SentryReplayFrame]()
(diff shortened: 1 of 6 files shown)


Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -35,27 +34,23 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
...
         withContentFrom outputPath: String,
         processingQueue: SentryDispatchQueueWrapper,
-        assetWorkerQueue: SentryDispatchQueueWrapper,
-        dateProvider: SentryCurrentDateProvider
+        assetWorkerQueue: SentryDispatchQueueWrapper
     ) {
         self.init(
             outputPath: outputPath,
             processingQueue: processingQueue,
             assetWorkerQueue: assetWorkerQueue,
```
philprime:
```suggestion
            assetWorkerQueue: assetWorkerQueue
```

Tests/SentryTests/Integrations/SessionReplay/SentryOnDemandReplayTests.swift
```diff
@@ -31,15 +30,14 @@ class SentryOnDemandReplayTests: XCTestCase {
...
-            dateProvider: dateProvider
+            assetWorkerQueue: trueDispatchQueueWrapper ? SentryDispatchQueueWrapper() : TestSentryDispatchQueueWrapper()
         )
     }
     
     func testAddFrame() {
         let sut = getSut()
-        sut.addFrameAsync(image: UIImage.add)
-       
+        sut.addFrameAsync(timestamp: Date(timeIntervalSinceReferenceDate: 0xBAAD_F00D), image: UIImage.add)
```
philippsuess, line 39: `m`: Why 0xBAAD_F00D?

philipphofmann (approved):
LGTM

Tests/SentryTests/Integrations/SessionReplay/SentryOnDemandReplayTests.swift
philprime: No apparent reason, just any number other than 0

philprime: https://en.wikipedia.org/wiki/Hexspeak

philippsuess: OK thanks for the info.

PR #5343, opened 2025-06-03
refactor(session-replay): Add separate files and missing type prefix

philprime:
Blocked by:

- [ ] #5341
- [ ] #5342

#skip-changelog

### DIFF
--- a/Sources/Swift/Core/Tools/ViewCapture/SentryRedactRegion.swift
+++ b/Sources/Swift/Core/Tools/ViewCapture/SentryRedactRegion.swift
@@ -1,0 +1,35 @@
+#if canImport(UIKit) && !SENTRY_NO_UIKIT
+#if os(iOS) || os(tvOS)
+import Foundation
+import ObjectiveC.NSObjCRuntime
+import UIKit
+
+class SentryRedactRegion {
+    enum CodingKeys: CodingKey {
+        case size
+        case transform
+        case type
+        case color
+        case name
+    }
+
+    let size: CGSize
+    let transform: CGAffineTransform
+    let type: SentryRedactRegionType
+    let color: UIColor?
+    let name: String
+
+    init(size: CGSize, transform: CGAffineTransform, type: SentryRedactRegionType, color: UIColor? = nil, name: String) {
+        self.size = size
+        self.transform = transform
+        self.type = type
+        self.color = color
+        self.name = name
+    }
+
+    func canReplace(as other: SentryRedactRegion) -> Bool {
+        size == other.size && transform == other.transform && type == other.type
+    }
+}
+#endif
+#endif
--- a/Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
+++ b/Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
@@ -7,45 +7,7 @@ import UIKit
 import WebKit
 #endif
 
-enum RedactRegionType {
-    /// Redacts the region.
-    case redact
-    
-    /// Marks a region to not draw anything.
-    /// This is used for opaque views.
-    case clipOut
-    
-    /// Push a clip region to the drawing context.
-    /// This is used for views that clip to its bounds.
-    case clipBegin
-    
-    /// Pop the last Pushed region from the drawing context.
-    /// Used after prossing every child of a view that clip to its bounds.
-    case clipEnd
-    
-    /// These regions are redacted first, there is no way to avoid it.
-    case redactSwiftUI
-}
-
-struct RedactRegion {
-    let size: CGSize
-    let transform: CGAffineTransform
-    let type: RedactRegionType
-    let color: UIColor?
-    
-    init(size: CGSize, transform: CGAffineTransform, type: RedactRegionType, color: UIColor? = nil) {
-        self.size = size
-        self.transform = transform
-        self.type = type
-        self.color = color
-    }
-    
-    func canReplace(as other: RedactRegion) -> Bool {
-        size == other.size && transform == other.transform && type == other.type
-    }
-}
-
-class UIRedactBuilder {
+class SentryUIRedactBuilder {
     ///This is a wrapper which marks it's direct children to be ignored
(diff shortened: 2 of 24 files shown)


Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -256,32 +220,57 @@ class UIRedactBuilder {
         var enforceRedact = forceRedact
         
         if !ignore && redact {
-            redacting.append(RedactRegion(size: layer.bounds.size, transform: newTransform, type: swiftUI ? .redactSwiftUI : .redact, color: self.color(for: view)))
+            redacting.append(SentryRedactRegion(
+                size: layer.bounds.size,
+                transform: newTransform,
+                type: swiftUI ? .redactSwiftUI : .redact,
+                color: self.color(for: view),
+                name: layer.debugDescription
```
philprime, line 228: This `name` is an addition, it helps to debug redaction issues.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -256,32 +220,57 @@ class UIRedactBuilder {
...
             if isAxisAligned(newTransform) && finalViewFrame == rootFrame {
                 //Because the current view is covering everything we found so far we can clear `redacting` list
                 redacting.removeAll()
             } else {
-                redacting.append(RedactRegion(size: layer.bounds.size, transform: newTransform, type: .clipOut))
+                redacting.append(SentryRedactRegion(
+                    size: layer.bounds.size,
+                    transform: newTransform,
+                    type: .clipOut,
+                    name: layer.debugDescription
```
philprime, line 245: See other comment

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -256,32 +220,57 @@ class UIRedactBuilder {
...
         let clipToBounds = view.clipsToBounds
         if clipToBounds {
             /// Because the order in which we process the redacted regions is reversed, we add the end of the clip region first.
             /// The beginning will be added after all the subviews have been mapped.
-            redacting.append(RedactRegion(size: layer.bounds.size, transform: newTransform, type: .clipEnd))
+            redacting.append(SentryRedactRegion(
+                size: layer.bounds.size,
+                transform: newTransform,
+                type: .clipEnd,
+                name: layer.debugDescription
```
philprime, line 261: See other comment

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -256,32 +220,57 @@ class UIRedactBuilder {
...
         for subLayer in subLayers.sorted(by: { $0.zPosition < $1.zPosition }) {
             mapRedactRegion(fromLayer: subLayer, relativeTo: layer, redacting: &redacting, rootFrame: rootFrame, transform: newTransform, forceRedact: enforceRedact)
         }
         if clipToBounds {
-            redacting.append(RedactRegion(size: layer.bounds.size, transform: newTransform, type: .clipBegin))
+            redacting.append(SentryRedactRegion(
+                size: layer.bounds.size,
+                transform: newTransform,
+                type: .clipBegin,
+                name: layer.debugDescription
```
philprime, line 272: See other comment

Sources/Swift/Core/Tools/ViewCapture/SentryRedactRegion.swift
```diff
@@ -0,0 +1,35 @@
...
+
+class SentryRedactRegion {
+    enum CodingKeys: CodingKey {
+        case size
+        case transform
+        case type
+        case color
+        case name
+    }
+
```
armcknight: I was able to compile without this, is it needed? `SentryRedactRegion` isn't `Codable`, and the names are the same as the props.

```suggestion
```

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
armcknight: It looks like we're now passing two different properties of the same layer in the two params (size, name). we could just pass `layer` as long as name is always going to be the layer's debugDescription

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -74,22 +74,22 @@ class SentryOnDemandReplay: NSObject, SentryReplayVideoMaker {
...
-        SentryLog.debug("[Session Replay] Adding frame to replay, screen: \(screen ?? "nil") at timestamp: \(timestamp)")
-        guard let data = rescaleImage(image)?.pngData() else {
+    private func addFrame(timestamp: Date, maskedViewImage: UIImage, forScreen screen: String?) {
+        SentryLog.debug("[Session Replay] Adding frame to replay, screen: \(screen ?? "nil")")
+        guard let data = rescaleImage(maskedViewImage)?.pngData() else {
             SentryLog.error("[Session Replay] Could not rescale image, dropping frame")
             return
         }
-
+
```
armcknight:
```suggestion

```

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: Unfortunately we can not. Passing the `layer` will keep a strong reference to the `CALayer` and when accessing it from a background thread, it will cause main thread warnings. I also observed crashes from doing that.

Sources/Swift/Core/Tools/ViewCapture/SentryRedactRegion.swift
philprime: Thanks for pointing this out. Seems to be a left-over I missed to remove 👍

PR #5346, opened 2025-06-03
chore: Change codecov.yml to recursively ignore test files

philprime:
- Codecov is picking up coverage of test files, which is not helpful in PR comments when trying to hit patch coverage goals.
- According to internal support we need to match the file structure recursively with double asteriks.

#skip-changelog

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -10,6 +10,7 @@ coverage:
   round: down
 
 ignore:
+  - "SentryTestUtilsTests/**"
   - "Tests/**"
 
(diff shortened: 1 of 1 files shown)


philprime:
Looks like the codecov file tree is now correct 👍 

https://app.codecov.io/gh/getsentry/sentry-cocoa/commit/b23b5866fad0b7ff32e231893ad46727b3b81538/tree

nsdeschenes (approved):
🥳

PR #5359, opened 2025-06-04
Full spm

noahsmartin:
This enables SPM support for building the entire SDK, and gets it building in CI. The approach follows from our slack discussion about introducing a new flag for any changes that are breaking the public API.

The PR is separated into 3 commits for easier reviewing. The first commit is the changes to `Decodable` conformances which is the part that changes the public API. The second commit is enabling accessing sentryId from Swift. Previously I had done this with a force cast, but this is a simpler way to get it compiling in CI without needing to introduce the force cast. The last commit just sets up CI/Package.swift

If we didn't want to modify the public API we could convert all these classes to Swift, as I started doing here: https://github.com/getsentry/sentry-cocoa/pull/5390 but the focus of this PR is just to get CI building with SPM without making as extensive changes.

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,6 +92,7 @@ jobs:
     runs-on: macos-15
     steps:
       - uses: actions/checkout@v4
+      # Only watchOS is compiling with SPM so far, the rest of the targets need profiling support which is still a work in progress due to the ObjC++ usage
       - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace && EXPERIMENTAL_SPM_BUILDS=1 xcodebuild build -scheme SentrySPM -sdk watchos -destination 'generic/platform=watchOS'
         shell: sh
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,6 +87,13 @@ jobs:
           name: raw-build-output-scheme-${{matrix.scheme}}
           path: |
             raw-build-output.log
+  build-spm:
+    name: Build with SPM
+    runs-on: macos-15
+    steps:
+      - uses: actions/checkout@v4
+      - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace && EXPERIMENTAL_SPM_BUILDS=1 xcodebuild build -scheme SentrySPM -sdk watchos -destination 'generic/platform=watchOS'
+        shell: sh
 
   check-debug-without-UIKit:
     name: Check no UIKit linkage (DebugWithoutUIKit)
--- a/Package.swift
+++ b/Package.swift
@@ -1,41 +1,90 @@
 // swift-tools-version:5.3
+import Darwin.C
 import PackageDescription
+
+var products: [Product] = [
+    .library(name: "Sentry", targets: ["Sentry"]),
+    .library(name: "Sentry-Dynamic", targets: ["Sentry-Dynamic"]),
+    .library(name: "SentrySwiftUI", targets: ["Sentry", "SentrySwiftUI"])
+]
+
+var targets: [Target] = [
+    .binaryTarget(
+        name: "Sentry",
+        url: "https://github.com/getsentry/sentry-cocoa/releases/download/8.52.1/Sentry.xcframework.zip",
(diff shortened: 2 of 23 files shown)


Package.swift
```diff
@@ -1,41 +1,84 @@
...
+              name: "Sentry",
+              url: "https://github.com/getsentry/sentry-cocoa/releases/download/8.52.1/Sentry.xcframework.zip",
+              checksum: "48a6c6693148a3f9096108164eb938a931b964395fcf38e169383b9e4cffcfc5" //Sentry-Static
+          ),
+  .binaryTarget(
+              name: "Sentry-Dynamic",
+              url: "https://github.com/getsentry/sentry-cocoa/releases/download/8.52.1/Sentry-Dynamic.xcframework.zip",
+              checksum: "b9d9054c65ee5ac0591c27826edddc54490a96c2a83dee25519cae0c1a593231" //Sentry-Dynamic
+          ),
+  .target ( name: "SentrySwiftUI",
```
philprime: `m`: This file might need some formatting

```suggestion
.target(name: "SentrySwiftUI",
```

Sources/Sentry/SentryEventSwiftHelper.m
```diff
@@ -0,0 +1,17 @@
+#import "SentryEventSwiftHelper.h"
+#import "SentryEvent.h"
+#import "SentrySwift.h"
+
+@implementation SentryEventSwiftHelper
```
philprime, line 12: `m`: Why do we need this helper class? It seems so little functionality. Please add some internal documentation

Sources/Swift/Exports.swift
```diff
@@ -0,0 +1,3 @@
+#if SENTRY_SWIFT_PACKAGE
```
philprime, line 6: `m`: A header comment with documentation of the file purpose would be convenient.

Sources/Swift/Protocol/Codable/SentryBreadcrumbCodable.swift
```diff
@@ -1,7 +1,16 @@
 @_implementationOnly import _SentryPrivate
 import Foundation
 
-extension Breadcrumb: Decodable {
+#if SENTRY_SWIFT_PACKAGE
```
philprime, line 5: `m`: We should document somewhere in this repository why this approach with `#if-else`  is used, so we can easily understand it again when revisiting it in a couple of years.

philprime: Maybe a page in the `develop-docs` directory would be the best place to fully explain it, but then we add a comment to each of the decodable helper classes with a reference.

.github/workflows/build.yml
```diff
@@ -87,6 +87,13 @@ jobs:
           name: raw-build-output-scheme-${{matrix.scheme}}
           path: |
             raw-build-output.log
+  build-spm:
+    name: Build with SPM
+    runs-on: macos-15
+    steps:
+      - uses: actions/checkout@v4
+      - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace && EXPERIMENTAL_SPM_BUILDS=1 xcodebuild build -scheme SentrySPM -sdk watchos -destination 'generic/platform=watchOS'
```
philippsuess, line 96: `m`: Why do we specify watchOS? If there's a good reason, please add the reply as a comment.

philippsuess, line 96: `m`: Maybe it makes sense to add this to `.scripts/sentry-xcodebuild.sh`?

noahsmartin: Because the non-watchOS builds are not working yet, they require removing Swift imports from ObjC++ used by profiling code. I plan to do that next and then update this to build for iOS. Will add a comment

noahsmartin: Since it's still pretty experimental I'd rather it be isolated for now, and then we can move it once there is better support (like for non watchOS)

noahsmartin:
I think we should hold off a bit longer on letting users try it, maybe once we have our own sample app using it that would be a good time to put it in the changelog/let people try it

PR #5360, opened 2025-06-04
refactor: Serialize envelope straight to a file

itaybre:
## :scroll: Description

Write files straight to a file on `storeEnvelope` instead of storing the data which may cause increased memory usage

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -207,7 +207,7 @@ _non_thread_safe_removeFileAtPath(NSString *path)
             [self.envelopesPath stringByAppendingPathComponent:[self uniqueAscendingJsonName]];
         SENTRY_LOG_DEBUG(@"Writing envelope to path: %@", path);
 
-        if ([SentrySerialization writeEnvelope:envelope toPath:path]) {
+        if (![SentrySerialization writeEnvelope:envelope toPath:path]) {
             SENTRY_LOG_ERROR(@"Failed to serialize and store envelope.");
             return nil;
         }


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Missing debug meta for non fatal events (#5352)
 
+### Improvements
+
+- Memory footprint decrese when serializing envelopes (#5360)
```
```suggestion
- Memory footprint decreased when serializing envelopes (#5360)
```

philippsuess:
I don't have the bandwidth to review this PR today, and I'm on PTO the whole next week. FYI, I tried something similar with a NSFileHandle last year and we had a very nasty severe incident https://github.com/getsentry/sentry-cocoa/pull/4219 with corrupted envelopes. Check the changelog of [8.35.0](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8350) and the 8.35.1. Serializing envelopes is a core part of the SDK. When we do this we have to be 100% confident that everything is working correctly. With CI being so flaky right now, I highly doubt that we can be so confident.

While it's highly admirable that you're trying to improve this, I would rather focus on making the CI stable and postpone such refactorings for now, please.

itaybre:
> I don't have the bandwidth to review this PR today, and I'm on PTO the whole next week. FYI, I tried something similar with a NSFileHandle last year and we had a very nasty severe incident #4219 with corrupted envelopes. Check the changelog of [8.35.0](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8350) and the 8.35.1. Serializing envelopes is a core part of the SDK. When we do this we have to be 100% confident that everything is working correctly. With CI being so flaky right now, I highly doubt that we can be so confident.
> 
> While it's highly admirable that you're trying to improve this, I would rather focus on making the CI stable and postpone such refactorings for now, please.

I agree it is a risky change, and to be honest, I would like to refactor how we read envelopes too.
It think the code can be improved and add some failsafes there.

Regarding the CI flakiness, we are improving there 💪

Sources/Sentry/SentryFileManager.m
```diff
@@ -202,20 +202,15 @@ - (NSURL *)getSentryPathAsURL
...
     @synchronized(self) {
         NSString *path =
             [self.envelopesPath stringByAppendingPathComponent:[self uniqueAscendingJsonName]];
         SENTRY_LOG_DEBUG(@"Writing envelope to path: %@", path);
 
-        if (![self writeData:envelopeData toPath:path]) {
-            SENTRY_LOG_WARN(@"Failed to store envelope.");
+        BOOL success = [SentrySerialization writeEnvelope:envelope toPath:path];
+
+        if (!success) {
```
armcknight: If we're not using the variable for anything else:

```suggestion
        if (![SentrySerialization writeEnvelope:envelope toPath:path]) {
```

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,81 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
     return sentryLevelForString(eventDictionary[@"level"]);
 }
 
++ (BOOL)writeObject:(id)jsonObject toStream:(NSOutputStream *)stream
+{
+    NSError *error = nil;
+    // Use an autoreleasepool to release memory as soon as possible
+    @autoreleasepool {
+        [NSJSONSerialization writeJSONObject:jsonObject toStream:stream options:0 error:&error];
```
armcknight: We should wrap this similar to `SentrySerialization.dataWithJSONObject` where we first check if the json object is valid. (I used to have code in there that would also catch NSExceptions in DEBUG, because there are cases where `isValidJSONObject` will return `YES` but the method _still_ throws exceptions. but that code was removed. just to provide more context on the additional safety checks available/possible there)

armcknight: I think also that wrapping it allows us to separate some of the intent of unit tests between serialization of data per se, vs envelopes as their own thing. Up to you how you want to do that for this PR. Just for consideration.

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,81 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+        SENTRY_LOG_ERROR(@"Internal error while serializing JSON: %@", error);
+        return NO;
+    }
+
+    return YES;
+}
+
++ (NSMutableDictionary *)buildEnvelopeSerialized:(SentryEnvelope *_Nonnull)envelope
+{
+    NSMutableDictionary *serializedData = [NSMutableDictionary new];
```
armcknight: `new` considered dangerous

```suggestion
    NSMutableDictionary *serializedData = [NSMutableDictionary dictionary];
```

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,81 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+    NSDate *sentAt = envelope.header.sentAt;
+    if (sentAt != nil) {
+        [serializedData setValue:sentry_toIso8601String(sentAt) forKey:@"sent_at"];
+    }
+    return serializedData;
+}
+
++ (BOOL)writeEnvelope:(SentryEnvelope *)envelope toPath:(NSString *)path
+{
+    NSOutputStream *outputStream = [NSOutputStream outputStreamToFileAtPath:path append:NO];
```
armcknight: `nil` check and early exit?

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,81 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+    BOOL headerSuccess = [self writeObject:serializedData toStream:outputStream];
+    if (!headerSuccess) {
+        [outputStream close];
+        [[NSFileManager defaultManager] removeItemAtPath:path error:nil];
+        SENTRY_LOG_ERROR(@"Envelope header cannot be converted to JSON.");
+        return NO;
+    }
+
+    for (int i = 0; i < envelope.items.count; ++i) {
+        [outputStream write:[@"\n" dataUsingEncoding:NSUTF8StringEncoding].bytes maxLength:1];
```
armcknight: I wonder if it's worth extracting this to a method that `dispatch_once`s this and statically stores the result? 
```
[@"\n" dataUsingEncoding:NSUTF8StringEncoding].bytes
```

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,81 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+
+        NSDictionary *serializedItemHeaderData = [envelope.items[i].header serialize];
+
+        BOOL itemSuccess = [self writeObject:serializedItemHeaderData toStream:outputStream];
+        if (!itemSuccess) {
+            [outputStream close];
+            [[NSFileManager defaultManager] removeItemAtPath:path error:nil];
```
armcknight: Annoying rabbit hole, but we should capture and handle the error here as well.

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,81 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+            SENTRY_LOG_ERROR(@"Envelope item header cannot be converted to JSON.");
+            return NO;
+        }
+
+        [outputStream write:[@"\n" dataUsingEncoding:NSUTF8StringEncoding].bytes maxLength:1];
+        [outputStream write:envelope.items[i].data.bytes maxLength:envelope.items[i].data.length];
+    }
```
armcknight: Just to simplify the repeated array indices a little bit, we could use fast enumeration here:

```suggestion
    for (SentryEnvelopeItem *envelopeItem in envelope.items) {
        [outputStream write:[@"\n" dataUsingEncoding:NSUTF8StringEncoding].bytes maxLength:1];

        NSDictionary *serializedItemHeaderData = [envelopeItem.header serialize];

        BOOL itemSuccess = [self writeObject:serializedItemHeaderData toStream:outputStream];
        if (!itemSuccess) {
            [outputStream close];
            [[NSFileManager defaultManager] removeItemAtPath:path error:nil];
            SENTRY_LOG_ERROR(@"Envelope item header cannot be converted to JSON.");
            return NO;
        }

        [outputStream write:[@"\n" dataUsingEncoding:NSUTF8StringEncoding].bytes maxLength:1];
        [outputStream write:envelopeItem.data.bytes maxLength:envelopeItem.data.length];
    }
```

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,81 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
++ (BOOL)writeEnvelope:(SentryEnvelope *)envelope toPath:(NSString *)path
+{
+    NSOutputStream *outputStream = [NSOutputStream outputStreamToFileAtPath:path append:NO];
+    [outputStream open];
+
+    NSMutableDictionary *serializedData = [self buildEnvelopeSerialized:envelope];
+    BOOL headerSuccess = [self writeObject:serializedData toStream:outputStream];
+    if (!headerSuccess) {
+        [outputStream close];
+        [[NSFileManager defaultManager] removeItemAtPath:path error:nil];
```
armcknight: Same comment here with the `error` param.

Sources/Sentry/SentrySerialization.m
armcknight: Every call to `-[NSStream write]` returns an integer status, we need to handle the return values.

armcknight (commented):
I will defer to @philipphofmann as the final reviewer here since he's got the battle scars already here 😄 but I think this is the right thing to do, we should figure it out and get it right and ship it: every time we move from keeping something in memory to writing it to disk, we increase the probability of retaining data in the face of a crash, versus losing that data at worst, or having to recompute it at best.

We'll likely need to mock NSOutputStream so we can unit test all the non-happy-paths.

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,153 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+            [self cleanupFailedWrite:outputStream tempPath:tempPath];
+            SENTRY_LOG_ERROR(@"Failed to write item data to stream.");
+            return NO;
+        }
+    }
+
+    [outputStream close];
+
+    NSError *moveError;
+    [[NSFileManager defaultManager] moveItemAtPath:tempPath toPath:path error:&moveError];
```
philprime, line 409: `m`: Please check if the `SentryFileManager` offers functionality that can be used here too.

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,153 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+
++ (NSData *)newlineData
+{
+    static NSData *newlineData = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{ newlineData = [@"\n" dataUsingEncoding:NSUTF8StringEncoding]; });
+    return newlineData;
+}
+
++ (BOOL)writeEnvelope:(SentryEnvelope *)envelope toPath:(NSString *)path
```
philprime, line 351: `m`: To be honest, I do not like that we are mixing a serialization utility class with actual file operations. IMO to have separation of concern, a serialize should ideally have two methods per type (serialize and deserialize X), mapping a type into data.
But as this is my opinion, I will not enforce this, but I would like to propose to move move most of the file management logic to e.g. `SentryFileManager`

Sources/Sentry/SentryFileManager.m
```diff
@@ -202,20 +202,13 @@ - (NSURL *)getSentryPathAsURL
...
-    }
-
     @synchronized(self) {
         NSString *path =
             [self.envelopesPath stringByAppendingPathComponent:[self uniqueAscendingJsonName]];
         SENTRY_LOG_DEBUG(@"Writing envelope to path: %@", path);
 
-        if (![self writeData:envelopeData toPath:path]) {
-            SENTRY_LOG_WARN(@"Failed to store envelope.");
+        if (![SentrySerialization writeEnvelope:envelope toPath:path]) {
```
philprime, line 210: `m`: In addition to the other comment, I would propose the following (if reasonable):

`SentryEnvelopePersistentStore` offers the `storeEnvelope` method which uses the `SentrySerialization` to convert the data and passes it to `SentryFileManager` to write it (see `SentryScopeContextPersistentStore` for reference).

That way we have a lower-level class to access the disk, a helper class to convert data structures to binary data and a wrapper-class with a simpler interface.

philprime (commented):
I left a couple of architecture/concept comments to consider. In general I believe to see the benefits this brings, so almost LGTM.

Sources/Sentry/SentryFileManager.m
itaybre: I think it makes more sense to do this on a follow up, not in this PR.
Mainly because this will need changes in many places and might make it harder to review

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,153 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+    }
+    if (writeResult < 0) {
+        SENTRY_LOG_ERROR(@"Internal error while writing object to stream");
+        return NO;
+    }
+
+    return YES;
+}
+
++ (NSMutableDictionary *)buildEnvelopeSerialized:(SentryEnvelope *_Nonnull)envelope
```
philippsuess, line 319: `m`: This method only uses the `SentryEnvelopeHeader`. We could rename it to `serializeEnvelopeHeader` and only pass in the `SentryEnvelopeHeader`, so it's easier to understand. And you can ditch the `_Nonnull` because we use `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` .
```suggestion
+ (NSMutableDictionary *)buildEnvelopeSerialized:(SentryEnvelopeHeader *)envelope
```

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,153 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+    return serializedData;
+}
+
++ (NSData *)newlineData
+{
+    static NSData *newlineData = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{ newlineData = [@"\n" dataUsingEncoding:NSUTF8StringEncoding]; });
+    return newlineData;
+}
```
philippsuess, line 349: `l`: Having this somewhat constant in the middle of the file is a bit confusing to me. I think we could move this to the top of the class.

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,153 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
...
+{
+    NSInteger writeResult;
+    NSError *error = nil;
+    // Use an autoreleasepool to release memory as soon as possible
+    @autoreleasepool {
+        writeResult = [NSJSONSerialization writeJSONObject:jsonObject
+                                                  toStream:stream
+                                                   options:0
+                                                     error:&error];
+    }
```
philippsuess, line 305: `m`: I'm curious on why the `@autoreleasepool` actually helps here. Do you have proof that this reduces the memory footprint?

Sources/Sentry/SentrySerialization.m
philippsuess: `h`: Yes, indeed. We must always call `isValidJSONObject` when serializing. Users sometimes put data into the scope or wherever that otherwise leads to crashes when trying to serialize the data to JSON.

Sources/Sentry/SentrySerialization.m
```diff
@@ -310,6 +292,153 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
     return sentryLevelForString(eventDictionary[@"level"]);
 }
 
++ (BOOL)writeObject:(id)jsonObject toStream:(NSOutputStream *)stream
+{
+    NSInteger writeResult;
+    NSError *error = nil;
```
philippsuess, line 298: `m`: What about changing this so also `dataWithJSONObject` uses it. So, basically `writeObject` just takes the output stream and writes to it. It doesn't matter if it's a file or memory. Then we have one central place for calling `writeJSONObject` and also checking `isValidJSONObject` before writing it.

Sources/Sentry/SentryFileManager.m
philippsuess: I agree with @philprime here. My proposal is to keep the temp file logic in the SentryFileManager and the `SentrySerialization.writeEnvelope` doesn't write to a file but instead to an `NSOutputStream`. This has more benefits:
1.  `SentrySerialization.dataWithEnvelope` can create a `NSOutputStream.outputStreamToMemory` reusing the same logic.
2. SentryMigrateSessionInit can easily switch to using `NSOutputStream` https://github.com/getsentry/sentry-cocoa/blob/701b301b0777a6cbe3bd97f45be62cb72edd1eb7/Sources/Sentry/SentryMigrateSessionInit.m#L102-L106
3. It allows us to also use a stream when sending the HTTP request to Sentry instead of again loading the entire envelope into memory https://github.com/getsentry/sentry-cocoa/blob/701b301b0777a6cbe3bd97f45be62cb72edd1eb7/Sources/Sentry/SentryNSURLRequestBuilder.m#L14-L21

We must not do these optimizations in this PR, but doing what I recommended, will make it easier.

philprime:
Should this PR be marked as draft again?

philprime:
@itaybre should we keep this around or clean it up?

philprime:
Closing this one as outdated

PR #5370, opened 2025-06-05
revert: Run checks for release branches

philprime:
This reverts commit 49e2bd0ee860b000d1ed31b9c628c636dfec8b9c because it causes the release workflow to fail:

https://github.com/getsentry/sentry-cocoa/actions/runs/15468574493/job/43548210375


markushi (approved):
🚀

PR #5373, opened 2025-06-05
ci: Update ios versions and handle prewarmed ui test launches

noahsmartin:
I looked at a few of the tests that have been failing on my recent PRs and found a couple opportunities for improvements.

1 - The duplicated SDK test was running on iOS 18.0 and on that failed test the entire simulator had crashed (see screenshot) I'm trying updating it to 18.5 which has some simulator fixes that might make it not crash as much
Screenshot 2025-06-05 at 10 48 17 AM

2 - Another test was stuck with "app event loop idle notification not received". We saw things like this at Emerge sometimes that was from the app launch being prewarmed. UI tests don't know how to handle this and get stuck, we can fix it by first activating the app, and then launching it. There might be other reasons this happens, I read one is from permissions dialogs popping up on the host macOS machine, but at least this will fix one possible cause.
Screenshot 2025-06-05 at 10 52 00 AM

#skip-changelog

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift-UITests/BaseUITest.swift
+++ b/Samples/iOS-Swift/iOS-Swift-UITests/BaseUITest.swift
@@ -37,6 +37,9 @@ extension BaseUITest {
         for (k, v) in env {
             app.launchEnvironment[k] = v
         }
+        // App prewarming can sometimes cause simulators to get stuck in UI tests, activating them
+        // before launching clears any prewarming state.
+        app.activate()
         app.launch()
         waitForExistenceOfMainScreen()
     }
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -118,7 +118,7 @@ jobs:
             ./fastlane/test_output/**
 
   ui-tests-swift:
-    name: UI Tests for iOS-Swift Xcode ${{matrix.xcode}} - V4 # Up the version with every change to keep track of flaky tests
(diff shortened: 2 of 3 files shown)


philippsuess:
> LGTM, but we should ask @philipphofmann regarding the test name to verify we should update the name or not

What do you mean by the test name? Increasing to V5? If yes, we do that so we can compare the stability in our CI data in Sentry https://sentry.sentry.io/dashboard/46009/?project=5899451&statsPeriod=7d.

Samples/iOS-Swift/iOS-Swift-UITests/BaseUITest.swift
```diff
@@ -37,6 +37,9 @@ extension BaseUITest {
         for (k, v) in env {
             app.launchEnvironment[k] = v
         }
+        // App prewarming can sometimes cause simulators to get stuck in UI tests, activating them
+        // before launching clears any prewarming state.
+        app.activate()
```
philippsuess, line 42: @noahsmartin, thanks a lot for this improvement. If this actually improves the stability, we should add this for all UI tests. We use `app.launch()` in multiple UI tests without calling `activate` before.

armcknight: I've just been trying to debug something locally, and found out that the call to activate immediately followed by launch causes the UI test to launch the test host app twice, with a SIGKILL in between. Investigating.

PR #5375, opened 2025-06-05
ci: Use common UI test action

itaybre:
## :scroll: Description

Creates a unified workflow to use from all our UI tests

All our UI tests are running the same steps (or very similar), meaning that any change there need to be replicated multiple times, making it harder to maintain. Using a single workflow will help maintaining this workflow.

Run CI workflows

## :pencil: Checklist


.github/workflows/ui-tests-common.yml
```diff
@@ -0,0 +1,103 @@
...
+      xcode_version:
+        description: "Xcode version"
+        required: true
+        type: string
+      build_with_make:
+        description: "Build with make"
+        required: false
+        default: false
+        type: boolean
+      build_with_script:
```
armcknight: could we make this name more obvious that it triggers an xcframework build?
```suggestion
      build_xcframework:
```

.github/workflows/ui-tests.yml
```diff
@@ -143,158 +63,30 @@ jobs:
...
-
-      - name: Store screenshot
-        uses: ./.github/actions/capture-screenshot
-        if: failure()
+    uses: ./.github/workflows/ui-tests-common.yml
+    with:
+      fastlane_command: duplication_test
+      xcresult_name: duplication-test
+      xcode_version: 16.4
+      build_with_script: true
```
armcknight: here's the other one
```suggestion
      build_xcframework: true
```

.github/workflows/ui-tests-common.yml
```diff
@@ -0,0 +1,103 @@
+name: UI Tests Common
+
+on:
+  workflow_call:
+    inputs:
+      fastlane_command:
+        description: "The fastlane command to run"
+        required: true
+        type: string
+      xcresult_name:
```
armcknight: We could probably get away without using this, and just form a name based on the other inputs to this workflow: `uitests-xcresult-$fastlane_command-$xcode_version-$macos_version`, wdyt?

armcknight: Looking again I see that one of the values has spaces, so we'd have to write a slugifier :/

armcknight: Can also just use some internal formulation, since the xcresult is only really needed for the `Publish Test Report` step in this reusable workflow.

armcknight (approved):
Great improvement! Please also add `.github/workflows/ui-tests-common.yml` to the list of trigger files in `.github/workflows/ui-tests.yml`

.github/workflows/ui-tests-common.yml
itaybre: I feel this can be reused, but there seem to be some different names, sometimes dashes are replaced by underscore.

itaybre: I have to find the proper code

itaybre: I think I got it in [01739d0](https://github.com/getsentry/sentry-cocoa/pull/5375/commits/01739d094809ee042556730c93991b7e8fb6c5da), I thought the name was automatic

.github/workflows/ui-tests.yml
```diff
@@ -25,111 +26,29 @@ concurrency:
...
-            ./fastlane/test_output/**
-
-      - name: Store screenshot
-        uses: ./.github/actions/capture-screenshot
-        if: failure()
+    with:
+      fastlane_command: ui_tests_${{matrix.target}}
+      xcode_version: 16.2
+      build_with_make: true
+      macos_version: macos-14
```
noahsmartin, line 37: Not relevant for this change, but perhaps we should bump this up

noahsmartin (approved):

PR #5384, opened 2025-06-08
test: package integration

armcknight:
Smoke test our binary XCFramework build for every combination of platform (ios, tvos etc) package manager (spm, cocoapods, carthage) and certain sdk variant combinations (static, dynamic, swiftui)

Add them to `release.yml`, so that this validation can happen after building the XCFramework for PRs, but before running the craft plugin for releases.

#skip-changelog


armcknight:
Current status here is that almost all integration projects are able to compile and link, except for the carthage projects targeting a platform other than iOS. I'm debugging the `carthage` executable to see what's up with it: it unzips the xcframeworks zip and then only retains the iOS slices, which are not the correct ones.

They work locally, just iterating on getting them working on CI.

armcknight:
Current status: all the tests pass except the Carthage non-iOS jobs, because somehow the xcframework that winds up in the Carthage/Build directory always only contains iOS images. Looking into how we build the xcframework, what gets uploaded/downloaded by the actions jobs for those, [un]zipping and the Carthage binary itself.

PR #5388, opened 2025-06-10
WIP ref: Fix Decodable conformances for SPM

noahsmartin:
**This is still a WIP, no need to review yet**

This is the next step of the SPM migration doc, moving protocol conformances to not be across module boundaries. With SPM objc and swift are in separate modules, so this conformance cannot be added to a type defined in objc. Instead we can create a new class inheriting from the objc class, and add the conformance to this new class.

#skip-changelog


philippsuess:
@noahsmartin, It would be great to write a short comment on why you closed this PR for future reference.

PR #5391, opened 2025-06-10
Convert SentryGeo to Swift

noahsmartin:
#skip-changelog


noahsmartin:
Closing because this is a public API so it's a breaking change

PR #5286, opened 2025-05-23
ci: use dprint

itaybre:
## :scroll: Description

Replace `prettier` with `dprint`

## :bulb: Motivation and Context

Using `prettier` requires installing node (and nvm for our tooling), while also being slower if we intend to run it on many files later on.
dprint can be installed with just brew.

## :pencil: Checklist

You have to check all boxes before merging:

- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #5394, opened 2025-06-10
chore: remove node.js gitignore contents

armcknight:
We previously removed `prettier` in https://github.com/getsentry/sentry-cocoa/pull/5286 but I just noticed we still had these gitignore entries.

#skip-changelog


philprime (approved):
LGTM, thanks

PR #5395, opened 2025-06-10
ci: revert duplication test to its own logic not using ui-tests-commo…

armcknight:
Need to revert the change from #5375 for the one test that requires an xcframework to be built, because there's no way to set conditional dependencies between jobs, which would be needed to merge it together with the changes from #5218.

#skip-changelog


armcknight:
I can actually just remove the condition in ui-tests-common.yml to unblock the other PR. this can be worked out later.

PR #5398, opened 2025-06-12
Replace Maestro tests with regular UI Tests

itaybre:
## :scroll: Description

Maestro is producing flaky tests due to our crash test.
Using UI Tests should prevent this.

## :pencil: Checklist


TestSamples/SwiftUITestSample/SwiftUITestSample.xcodeproj/project.pbxproj
```diff
@@ -3,15 +3,27 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 70;
+	objectVersion = 63;
 	objects = {
 
 /* Begin PBXBuildFile section */
```
itaybre: TODO: use xcodegen

philprime: Are you doing this in this PR?

TestSamples/SwiftUITestSample/SwiftUITestSampleUITests/XCUIApplication+uninstall.swift
```diff
@@ -0,0 +1,48 @@
+import XCTest
+
+// This snipped was obtained from https://www.jessesquires.com/blog/2021/10/25/delete-app-during-ui-tests/
+extension XCUIApplication {
+    func uninstall(name: String? = nil) {
```
philprime: Nice! We could use this for all flaky tests where we need a clean state instead of using the `--io.sentry.wipe-data` to simulate a cleanup

TestSamples/SwiftUITestSample/SwiftUITestSampleUITests/SwiftUITestSampleUITests.swift
```diff
@@ -0,0 +1,50 @@
...
+        continueAfterFailure = false
+    }
+
+    override func tearDownWithError() throws {
+        if app.state != .notRunning {
+            app.terminate()
+        }
+        
+        // Uninstall the app after every test to make sure we run on a clean state
+        app.uninstall()
```
philprime: `m`: We might have to find a way to do this before, because it could happen that the app is still installed from the previous run (or from normal testing) and then the first test runs will not have a clean state.

philprime (approved):
LGTM, small feedback comments.

TestSamples/SwiftUITestSample/SwiftUITestSampleUITests/XCUIApplication+uninstall.swift
armcknight: Whoa, this is an awesome find 👍🏻 👍🏻

TestSamples/SwiftUITestSample/SwiftUITestSampleUITests/SwiftUITestSampleUITests.swift
armcknight: Agreed, it seems like a better approach to do it in either the class or instance set up method on XCUIApplication, depending on if it needs to be done once before a set of test cases, or before each test case in a class.

TestSamples/SwiftUITestSample/SwiftUITestSample.xcodeproj/project.pbxproj
itaybre: Yes

PR #5401, opened 2025-06-12
feat(screenshot): Add masking options for screenshot provider

### LINKED ISSUE #4831: Allow configuration of screenshot masking for hybrid SDKs
### Description

[Currently](https://github.com/getsentry/sentry-cocoa/blob/950adcc4ad4c913056e819a6542212c6458c6b48/Sources/Sentry/SentryScreenshot.m#L20) the screenshots are always taken with masking enabled and hybrids SDKs options are ignored. Thus the screenshots are masked only partially.

https://github.com/getsentry/sentry-cocoa/blob/950adcc4ad4c913056e819a6542212c6458c6b48/Sources/Sentry/SentryScreenshot.m#L16-L23

### LINKED ISSUE #4832: Option to disable screenshot masking
### Description

[Currently](https://github.com/getsentry/sentry-cocoa/blob/950adcc4ad4c913056e819a6542212c6458c6b48/Sources/Sentry/SentryScreenshot.m#L20) the screenshots are always taken with masking enabled. And there is no option to it.

We also need to add documentation about masking and how to opt out.

philprime:
## :scroll: Description

- Adds `SentryScreenshotOptions` to configure the masking rules of screenshots taken for captured when `attachScreenshots` is true, and also for user feedback. The reference was sentry-flutter which uses a `options.privacy` approach. As I do not want to touch the options of session replay to consolidate into one `privacy` block, I added it as separate options. My thoughts behind this was also that a user might not want their PII show up in replays, but is fine to consent to PII submission when using the user feedback form. This is hereby delegate to the SDK user.
- We can consider merging the options for screenshots and session replay in a future (major) release.
- ~I added a `getScreenshotProviderForOptions` to the dependency container. Originally I wanted to use the same logic as `getANRTracker` with a single lazy initialized instance, but but that can be wrong if options change later on. As we need to fix this eventually I instead introduced a map with weak references to the screenshot provider using the options as the key.~
- Adds screenshot options to `SentrySDKWrapper` and `SentrySDKOverrides`
- Adds session replay options to `SentrySDKWrapper` and `SentrySDKOverrides` because I had to turn it off during development
- Uses explicity `nonnull` and `_Nullable` annotations so we can have more null-safety.
- This PR also enables the mask renderer V2 by default, as it is already enabled for session replay by default already
- Refactored session replay flags for consistency: replaced `disableSessionReplay` with `disable`, renamed and split sample rate flags, and adjusted related code to use the new flag names. [[1]](diffhunk://#diff-6a98d5ce6bc4fdba41da9f89ff68296fb4fcead2ccbdc3046158ad33126fa4beL101-R123) [[2]](diffhunk://#diff-3e2347a4fe0f477de6a9eb0fc6a2e9be48e9060580c48a999d1cc15adf77655bL52-L66) [[3]](diffhunk://#diff-dc0ec55fb85d0443e834d5b3a41c6e06305d0ce3897c7fd40ab43d659c643cc8R12-R28) [[4]](diffhunk://#diff-dc0ec55fb85d0443e834d5b3a41c6e06305d0ce3897c7fd40ab43d659c643cc8L84-R101)
- Expanded `SentrySDKOverrides` enums to include screenshot options, and updated override type and ignore logic for both session replay and screenshot features to match the new flags. [[1]](diffhunk://#diff-6a98d5ce6bc4fdba41da9f89ff68296fb4fcead2ccbdc3046158ad33126fa4beR44) [[2]](diffhunk://#diff-6a98d5ce6bc4fdba41da9f89ff68296fb4fcead2ccbdc3046158ad33126fa4beL101-R123) [[3]](diffhunk://#diff-6a98d5ce6bc4fdba41da9f89ff68296fb4fcead2ccbdc3046158ad33126fa4beL315-R341) [[4]](diffhunk://#diff-6a98d5ce6bc4fdba41da9f89ff68296fb4fcead2ccbdc3046158ad33126fa4beL394-R423)

Closes #4831
Closes #4832

- I used to xcscheme arguments to enable and disable masking for screenshots. Then I tapped on `capture exception` and verified the screenshot in Sentry is masked or unmasked.
- Unit tests

## :pencil: Checklist


philprime:
@cursor review

philprime:
@cursor review

SentryTestUtils/TestSentryViewRenderer.swift
```diff
@@ -0,0 +1,30 @@
...
+#if os(iOS) || os(tvOS)
+
+@_spi(Private) @testable import Sentry
+import UIKit
+
+@_spi(Private) public class TestSentryViewRenderer: NSObject, SentryViewRenderer {
+    /// Records invocations of `render(view:)`.
+    public let renderInvocations: Invocations<WeakReference<UIView>> = Invocations()
+
+    /// The value to be returned by `render(view:)`. If `nil`, `render(view:)` will
```
philippsuess: `l`: The comment seems a bit broken. The last sentence is not a sentence. I think we could ditch all the code comments in here, cause there not adding much value.

Sources/Swift/Core/Tools/ViewCapture/SentryScreenshotSource.swift
```diff
@@ -5,42 +5,51 @@
...
     private let photographer: SentryViewPhotographer
-    
-    public override init() {
-        photographer = SentryViewPhotographer(
+
+    public override convenience init() {
+        // We need to provide a init method without parameters for Obj-C compatibility.
+        // However, we want to force users to provide a photographer.
+        // `assertionFailure` will not crash the app in release builds, but at least
+        // it notifies the developer during testing and development.
```
philippsuess, line 15: `m`: I'm confused why we need this. I can compile the code when removing this init method.

Sources/Swift/Integrations/Screenshot/SentryScreenshotOptions.swift
```diff
@@ -0,0 +1,215 @@
...
+     * Enabling this flag will reduce the amount of time it takes to render the screenshot on the main thread, therefore reducing
+     * interruptions and visual lag. [Our benchmarks](https://github.com/getsentry/sentry-cocoa/pull/4940) have shown a significant improvement of
+     * **up to 4-5x faster rendering** (reducing `~160ms` to `~36ms`) on older devices.
+     *
+     * - Experiment: In case you are noticing issues with the new view renderer, please report the issue on [GitHub](https://github.com/getsentry/sentry-cocoa).
+     *               Eventually, we will remove this feature flag and use the new view renderer by default.
+     *
+     * - Note: See ``SentryViewScreenshotOptions.init`` for the default value.
+     */
+    public var enableViewRendererV2: Bool
```
philippsuess, line 34: `m`: I don't like that we have loads of duplicated code for the `SentryRedactOptions` now. We have it here and in the `SentryReplayOptions`. Also, the code comments can easily get out of sync. Can we solve this with inheritance for now, so that both SentryReplayOptions and SentryViewScreenshotOptions inherit from a new class? Maybe we can also rething this whole approach with a better solution in a next major.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -275,18 +286,53 @@ - (SentryCrash *)crashReporter SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOCK
...
-#    else
-    SENTRY_LOG_DEBUG(
-        @"SentryDependencyContainer.screenshot only works with UIKit enabled. Ensure you're "
-        @"using the right configuration of Sentry that links UIKit.");
-    return nil;
-#    endif // SENTRY_HAS_UIKIT
+    @synchronized(sentryDependencyContainerDependenciesLock) {
+        SCREENSHOTS_PROVIDER_KEY key = [self getScreenshotProviderKey:options];
+        SentryScreenshotSource *_Nullable provider = [_screenshotSourceStorage objectForKey:key];
+
```
philippsuess: `h`: I'm unsure about the `_screenshotSourceStorage`. The options usually don't change, and the SDK doesn't react properly to changes to the options after calling start. So most of the time, we will have precisely one `SentryScreenshotSource` in this storage. It also requires the implementation of equals and hashable for the `SentryViewScreenshotOptions`. Also if somebody changes the options multiple times, this storage wouldn't clear its unused dependencies. So we would need to come up with a solution to clear the cache, which can be tricky.

I'm unsure if we need this. What about simply always returning a new instance of `SentryViewScreenshotOptions`? Making the SDK reacting correctly to changes of the options on the fly is a bigger initiative. Please let me know if I'm missing something.

philipphofmann (commented):
Thanks for adding this. I'm a bit unsure about the `_screenshotSourceStorage` in the dependency container. Apart from that, it looks legit. We can also discuss this on a call if it speeds things up.

Sources/Swift/Integrations/Screenshot/SentryScreenshotOptions.swift
philprime: See the first paragraph in the PR description.

Session replay and screenshots are two different features, ideally they should be in two different modules. If we share configuration values, they should both inherit from the same type (e.g. privacy settings like in Flutter). 

We can do this in another major → https://github.com/getsentry/sentry-cocoa/discussions/2946#discussioncomment-14285130

Sources/Swift/Core/Tools/ViewCapture/SentryScreenshotSource.swift
philprime: Because of this thread:

https://github.com/getsentry/sentry-cocoa/pull/5943/files#r2276145470

Maybe @noahsmartin can explain it more here.

Sources/Sentry/SentryDependencyContainer.m
philprime: We discussed this in multiple occasions that our current dependency injection assumes to be configured for global singleton options. While you are correct that the options usually don't change, we should transition away from this. It's especially noticable for the SentryFileManager.

This storage is just a map to have instances associated with DSNs, but using a generic pattern. It's the same as if we store the DSN of the current instance in another property, but the main difference is **allowing** multiple instances, which can be useful for parallelizing unit tests.

> Also if somebody changes the options multiple times, this storage wouldn't clear its unused dependencies.

This is wrong, at this point we always require an SDK restart to change the options, therefore the dependency container is fully discarded.

If you prefer to keep the dependency container hard-linked to the `SentrySDK.options`, I am open to adapt the PR, but I can already tell you that this pattern will come up again in the next couple of months.

philippsuess: > > Also if somebody changes the options multiple times, this storage wouldn't clear its unused dependencies.
>
> This is wrong, at this point we always require an SDK restart to change the options, therefore the dependency container is fully discarded.

Yes, but only when you restart the SDK, so what's the point of keeping a hashtable which has only one reference that gets cleared when resetting the dependency container? I'm a bit confused.

> We discussed this in multiple occasions that our current dependency injection assumes to be configured for global singleton options. While you are correct that the options usually don't change, we should transition away from this. It's especially noticable for the SentryFileManager.

I agree. Fixing it in one place doesn't solve the underlying problem. I'm unsure if the dependency container is the right place for this. I would rather solve the underlying problem with a solution that works for most of the global state and the singletons. That seems like a bigger discussion I would like to avoid having in this PR.

philprime:
After an internal discussion between @philipphofmann and myself, it was decided that even though the `SentryScreenshotSource` is bound to options, I will not introduce a per-options solution for the dependency container in this PR.

SentryTestUtils/TestSentryViewRenderer.swift
philprime: This is not just a comment, it's documentation. The `TestSentryViewRenderer` is a type provided by the `SentryTestUtils` and used in other parts of the SDK, so it should have some documentation.

I'll fix it.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -271,18 +271,34 @@ - (SentryCrash *)crashReporter SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOCK
...
+                    initWithEnableFastViewRendering:options.enableFastViewRendering];
+            } else {
+                viewRenderer = [[SentryDefaultViewRenderer alloc] init];
+            }
+
+            SentryViewPhotographer *photographer =
+                [[SentryViewPhotographer alloc] initWithRenderer:viewRenderer
+                                                   redactOptions:options
+                                            enableMaskRendererV2:options.enableViewRendererV2];
+            _screenshotSource = [[SentryScreenshotSource alloc] initWithPhotographer:photographer];
```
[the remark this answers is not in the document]
philprime: This is expected.

Sources/Sentry/SentryScreenshotIntegration.m
```diff
@@ -42,6 +44,7 @@ - (BOOL)installWithOptions:(nonnull SentryOptions *)options
     SentryClient *client = [SentrySDKInternal.currentHub getClient];
     [client addAttachmentProcessor:self];
 
+    // TODO: Use the options to currify the saveScreenshot function.
```
philippsuess: `m`: There's still a todo in here. Did you leave this on purpose?

Sources/Sentry/SentryUserFeedbackIntegration.m
```diff
@@ -20,11 +20,15 @@ - (BOOL)installWithOptions:(SentryOptions *)options
         return NO;
     }
 
+    // The screenshot source is coupled to the options, but due to the dependency container being
+    // tightly to the options anwyas, it was decided to not pass it to the container.
```
philippsuess: Thanks for the comment.

Sources/Swift/Integrations/Screenshot/SentryScreenshotOptions.swift
```diff
@@ -0,0 +1,215 @@
...
+        }
+        return self.enableViewRendererV2 == other.enableViewRendererV2 &&
+        self.enableFastViewRendering == other.enableFastViewRendering &&
+        self.maskAllText == other.maskAllText &&
+        self.maskAllImages == other.maskAllImages &&
+        self.maskedViewClasses.count == other.maskedViewClasses.count &&
+        self.unmaskedViewClasses.count == other.unmaskedViewClasses.count &&
+        self.maskedViewClasses.map { NSStringFromClass($0) } == other.maskedViewClasses.map { NSStringFromClass($0) } &&
+        self.unmaskedViewClasses.map { NSStringFromClass($0) } == other.unmaskedViewClasses.map { NSStringFromClass($0) }
+    }
```
philippsuess: `m`: Do we still need this? If not, it would be great to delete it.

philipphofmann (approved):
Thanks @philprime. LGTM

Samples/Shared/feature-flags.yml
```diff
@@ -9,12 +9,21 @@ schemeTemplates:
...
+        "--io.sentry.screenshot.disable-mask-all-text": false
+
         # session replay
-        "--io.sentry.session-replay.disable-session-replay": false
-        "--io.sentry.session-replay.disableViewRendereV2": false
-        "--io.sentry.session-replay.enableFastViewRendering": false
-        "--io.sentry.session-replay.disable-mask-all-text": false
+        "--io.sentry.session-replay.disable": false
+        "--io.sentry.session-replay.on-error-sample-rate": true
+        "--io.sentry.session-replay.session-sample-rate": false
```
philprime:
```suggestion
```

PR #5402, opened 2025-06-12
Remove unnecessary simulators

itaybre:
## :scroll: Description

Github Runners seem to have little space available which might cause degraded performance, remove non necessary simulators.

## :pencil: Checklist

#skip-changelog


itaybre:
> I’m skeptical because if these simulators haven’t been used they shouldn’t take much space right? Worth a try though

There are about 30 simulators being deleted, I saw about 30GB freed by this. But after running a couple times, seems like this doesn't always free space, might depend which runner it works on.
Will try deleting unused Xcodes versions

armcknight:
If I'm reading this right from the before and after runs of `df`, we have before:
```
Filesystem        Size    Used   Avail Capacity iused ifree %iused  Mounted on
/dev/disk3s1s1   320Gi   9.6Gi    45Gi    18%    404k  467M    0%   /
```
and after
```
/dev/disk3s1s1   320Gi   9.6Gi   124Gi     8%    404k  1.3G    0%   /
```
So before, there were already 45 gibibytes available. Do we think that is causing issues? I too would want to see some experiments around the amount of free space and prevalence of flakes before running 3-5 minutes of deletion on every job (multiplied by X parallel jobs)

philippsuess:
Thanks a lot @itaybre for investigating flaky UI tests.

I'm also unsure about this approach. If this fixes some flakiness, we should ideally open an issue for the GH action runner https://github.com/actions/runner-images and tell them to fix the underlying issue. I don't think we should mess around too much with the GH action runners environment. If they then change something, we can easily get unexpected behaviour.

We can also simply try it and see how it goes. It's very easy to revert this change.

itaybre:
Doesn't seem to improve stability, closing this PR

philippsuess:
@itaybre, thanks for exploring this avenue, though.

PR #5408, opened 2025-06-13
fix(session-replay): Add redaction for SFSafariView

krystofwoldrich:
## :scroll: Description

SFSafariView doesn't use WebView component and thus is not redacted with the current set of classes to redact.

fixes: https://github.com/getsentry/sentry-react-native/issues/4920

tested with adjusted reproducible example: https://github.com/krystofwoldrich/expo-auth-session-sentry

example of replay before this pr: https://krystof-woldrich.sentry.io/explore/replays/9d5c14e2372c4d4ca23e6238bdf63aa9

example of replay after this pr: https://krystof-woldrich.sentry.io/explore/replays/7b883d4d68594c06a4869884788b4e9c

## :pencil: Checklist


philprime:
Thank you for the PR. I'll take over here and make sure we have an example in the sample apps and tests/verification

philprime:
- Added a `SFSafariViewController` to iOS sample
- Added a `ASWebAuthenticationSession` for web-based authentication to iOS sample
- Cleanup and tests

philprime (approved):

philprime:
I manually tested the redaction for iOS 16, 17 and 18.
Both SFSafariViewController and the ASWebAuthenticationSession are redacted correctly.

PR #5409, opened 2025-06-13
ci: Split tests into individual jobs

itaybre:
## :pencil: Checklist

### DIFF
--- a/TestSamples/SwiftUITestSample/SwiftUITestSampleUITestsCrashOnly/CrashTests.swift
+++ b/TestSamples/SwiftUITestSample/SwiftUITestSampleUITestsCrashOnly/CrashTests.swift
@@ -15,13 +15,14 @@ final class SwiftUITestSampleUITests: XCTestCase {
         app.buttons["Crash"].tap()
         
         launchApp()
-        XCTAssertTrue(app.staticTexts["Welcome!"].exists)
+        XCTAssertTrue(app.staticTexts["Welcome!"].waitForExistence(timeout: 10), "App did not load properly - Welcome text not found")
     }
     
     private func launchApp() {
         app.activate()
-        // Wait for the app to be ready
-        sleep(2)
+        // Make sure the app is ready before launching it
+        XCTAssertTrue(app.staticTexts["Welcome!"].waitForExistence(timeout: 10), "App did not activate properly")
+
         app.launch()
     }
 }


Plans/SwiftUITestSample_Crash.xctestplan
```diff
@@ -0,0 +1,28 @@
...
+    }
+  ],
+  "defaultOptions" : {
+    "uiTestingScreenshotsLifetime" : "keepAlways"
+  },
+  "testTargets" : [
+    {
+      "skippedTests" : [
+        "EnvelopeTest",
+        "EnvelopeTest\/testCorruptedEnvelope()"
```
armcknight: I don't love this, it's not going to scale well. We'll have to add every test case to every test plan to skip, which IMO should only be used for tests that are not functioning correctly for some reason. Could we split them into separate Xcode UI test targets instead? Then they just get automatically included without having to mess with this extra layer of indirection.

TestSamples/SwiftUITestSample/SwiftUITestSample.yml
```diff
@@ -30,14 +30,61 @@ targets:
...
+        SwiftUITestSampleUITestsEnvelopeOnly: test
+    test:
+      config: Test
+      testPlans:
+        - path: ../../Plans/SwiftUITestSample_Envelope.xctestplan
+      preActions:
+        - script: |
+            echo "Uninstalling app"
+            xcrun simctl uninstall booted io.sentry.tests.SwiftUITestSample
+  SwiftUITestSampleAll:
```
armcknight, line 87: I don't think anything uses this anymore, at least in CI. We still would have the make task `test-ui-critical` that runs `./scripts/test-ui-critical.sh`, which uses it, but we could also change it to run the two new targets you've created instead. Up to you.

armcknight (approved):
Good change, in general I like keeping test targets small and scoped. Better parallelizability, smaller blast radius for weird bugs in tests. 👍🏻

TestSamples/SwiftUITestSample/SwiftUITestSampleUITestsCrashOnly/CrashTests.swift
```diff
@@ -0,0 +1,27 @@
...
+        app.buttons["Crash"].tap()
+        
+        launchApp()
+        XCTAssertTrue(app.staticTexts["Welcome!"].exists)
+    }
+    
+    private func launchApp() {
+        app.activate()
+        // Wait for the app to be ready
+        sleep(2)
```
philippsuess: `m`: How can we be sure that sleep(2) is long enough if the CI is very busy? Using sleep can lead to flakiness from my experience. I would prefer not using it, if possible.

philipphofmann (commented):
Just gave this a quick glance and found one minor issue.

TestSamples/SwiftUITestSample/SwiftUITestSampleUITestsCrashOnly/CrashTests.swift
itaybre: Fair enough, I can use `waitForExistence`

TestSamples/SwiftUITestSample/SwiftUITestSample.yml
itaybre: There is already a script for `make test-ui-critical`

PR #5413, opened 2025-06-16
ci: Make hang triggering more deterministic

noahsmartin:
I've noticed tests related to app hangs frequently fail. Most recently there was this run that failed: https://github.com/getsentry/sentry-cocoa/actions/runs/15674734611/job/44152349015

The test failed because it tried to trigger an app hang, but then the app was hanging for too long and the UI test timed out trying to run the next step. The way we trigger an app hang does not create a very deterministic length hang, it will vary based on the performance of the machine. This PR just changes the app hang to be a `sleep` with a constant time, which I hope will fix these flaky tests by preventing the hang from taking too long.

#skip-changelog

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift/Tools/TriggerAppHang.swift
+++ b/Samples/iOS-Swift/iOS-Swift/Tools/TriggerAppHang.swift
@@ -20,18 +20,7 @@ func triggerNonFullyBlockingAppHang() {
     }
 }
 
-/// Schedules heavy UI rendering work on the main thread in a tight loop, which causes
-/// a fully blocking app hang without frames being rendered.
+/// Sleep on the main thread for long enough to trigger a hang
 func triggerFullyBlockingAppHang(button: UIButton) {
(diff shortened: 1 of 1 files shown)


philprime:
I am not sure if this change is valid.

Before the change we had a busy loop, which keeps the CPU busy at 100% usage, which reflects if a SDK user's app is doing a lot of busy work.

After the change the `sleep(5)` is suspending the current thread for the given duration, while having the CPU usage be at essentially zero. The thread is blocked but doesn't consume CPU cycles.

Therefore the test case is now something different! Are we certain this is what we want? If not, we should consider reverting.

cc @philipphofmann

philippsuess:
Yes, indeed, @philprime. Detecting app hangs with thread sleep is very easy. When the run loop is extremely busy, it's a bit more difficult, but our app hang detection detects this. We could meet in the middle and have one button using thread sleep, which the UI test uses, and keep the other for manual testing. We should add comments explaining why we have both, please.

Issue #5327, opened 2025-05-30
Check persisted launch profile config in SentryTimeToDisplayTracker


armcknight:
### Description

in SentryTimeToDisplayTracker, it looks at the current configured options to decide if the launch profile is correlated to a trace, in order to stop it. however it should be looking at the options from the previous launch that configured the current launch profile, since options settings can change between launches.

we need to persist a few more options, and then read out and keep in memory _all_ the persisted launch profile config options.

philippsuess:
Fixed with https://github.com/getsentry/sentry-cocoa/pull/5417. Please reopen @armcknight if I'm mistaken.

PR #5300, opened 2025-05-27
fix: Race condition for app launch profiling

philipphofmann:
## :scroll: Description

When having app launch profiling and TTFD enabled, a race condition when reading alwaysWaitForFullDisplay could occur. This is fixed now by reading if TTFD is enabled from the options instead of the UIViewControllerPerformanceTracker.

## :bulb: Motivation and Context

Running the tests with the thread sanitizer enabled surfaced this data race.

```
==================
WARNING: ThreadSanitizer: data race (pid=47792)
  Read of size 1 at 0x00030aa799f8 by thread T350:
    #0 -[SentryUIViewControllerPerformanceTracker alwaysWaitForFullDisplay] <null> (Sentry:arm64+0x18c298)
    #1 __sentry_sdkInitProfilerTasks_block_invoke <null> (Sentry:arm64+0x12fca8)
    #2 __53-[SentryDispatchQueueWrapper dispatchAsyncWithBlock:]_block_invoke <null> (Sentry:arm64+0x144bc)
    #3 __tsan::invoke_and_release_block(void*) <null> (libclang_rt.tsan_iossim_dynamic.dylib:arm64+0x7f97c)
    #4 _dispatch_client_callout <null> (libdispatch.dylib:arm64+0x1c274)

  Previous write of size 1 at 0x00030aa799f8 by main thread (mutexes: write M0):
    #0 -[SentryUIViewControllerPerformanceTracker initWithTracker:dispatchQueueWrapper:] <null> (Sentry:arm64+0x1860bc)
    #1 -[SentryDependencyContainer uiViewControllerPerformanceTracker] <null> (Sentry:arm64+0x178bc8)
    #2 SentryTests.SentrySDKTests.testReportFullyDisplayed() -> () <null> (SentryTests:arm64+0x783388)
    #3 @objc SentryTests.SentrySDKTests.testReportFullyDisplayed() -> () <null> (SentryTests:arm64+0x783728)
    #4 __invoking___ <null> (CoreFoundation:arm64+0x13a20c)

  Location is heap block of size 80 at 0x00030aa799f0 allocated by main thread:
    #0 calloc <null> (libclang_rt.tsan_iossim_dynamic.dylib:arm64+0x5ed24)
    #1 _malloc_type_calloc_outlined <null> (libsystem_malloc.dylib:arm64+0xf308)
    #2 SentryTests.SentrySDKTests.testReportFullyDisplayed() -> () <null> (SentryTests:arm64+0x783388)
[...]

PR #5417, opened 2025-06-16
fix(profiling): check previous options for launch profiles

armcknight:
For #5327 

I noticed while reviewing #5300 that we mix persisted and current options when looking at conditions to stop launch profiles, specifically w.r.t. TTID/TTFD.


Sources/Sentry/SentryProfiler.mm
```diff
@@ -206,13 +230,14 @@ - (void)stopForReason:(SentryProfilerTruncationReason)reason
...
     sentry_isTracingAppLaunch = NO;
     [self.metricProfiler stop];
-    self.truncationReason = reason;
 
     if (![self isRunning]) {
         SENTRY_LOG_WARN(@"Profiler is not currently running.");
         return;
     }
 
+    self.truncationReason = reason;
```
noahsmartin, line 241: Not clear to me why this is moved below the warning check. Is it expected that check is sometimes hit? If it is maybe we should change it form WARN to DEBUG

Sources/Sentry/SentryPerformanceTracker.m
```diff
@@ -47,8 +52,14 @@ - (SentrySpanId *)startSpanWithName:(NSString *)name
                              origin:(NSString *)origin
 {
     id<SentrySpan> activeSpan;
-    @synchronized(self.activeSpanStack) {
-        activeSpan = [self.activeSpanStack lastObject];
+#if SENTRY_TARGET_PROFILING_SUPPORTED
+    activeSpan = sentry_launchTracer;
```
armcknight, line 56: This was a crucial bug fix; SentryTimeToDisplayTracker trying to stop a trace-based profile wouldn't be able to access the trace to stop it.

Sources/Sentry/SentryProfiler.mm
armcknight: Yeah probably fine to move it to DEBUG. This was overwriting a truncation reason if the profiler had timed out, and a subsequent call to stop came in. Related to how trace-based profiles piggyback off trace start/stop.

SentryTestUtils/SentryLaunchProfiling+Tests.h
```diff
@@ -3,23 +3,13 @@
...
 
-typedef struct {
-    BOOL shouldProfile;
-    SentrySamplerDecision *_Nullable tracesDecision;
-    SentrySamplerDecision *_Nullable profilesDecision;
-} SentryLaunchProfileConfig;
-
-SENTRY_EXTERN SentryTracer *_Nullable sentry_launchTracer;
-
-SentryLaunchProfileConfig sentry_shouldProfileNextLaunch(SentryOptions *options);
```
armcknight, line 21: Moved to SentryProfiler+Private.h

Sources/Sentry/include/SentryTracerConfiguration.h
```diff
@@ -30,23 +30,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * via @c SentryProfilingOptions.sessionSampleRate) or non-launch trace-based profiles.
  */
 @property (nonatomic, strong, nullable) SentrySamplerDecision *profilesSamplerDecision;
-
-/**
- * For launch continuous v2 profiles that must start for trace lifecycle, we must explicitly be able
- * to indicate that to the tracer here, since there's no hub or options attached to it for the
- * profiler system to know whether it's a old-style trace profile or a trace continuous v2 profile.
- */
-@property (nonatomic, strong, nullable) SentryProfileOptions *profileOptions;
```
armcknight, line 49: we'll now retrieve these from the global `sentry_profileConfiguration`

Sources/Sentry/SentryPerformanceTracker.m
```diff
@@ -47,8 +52,14 @@ - (SentrySpanId *)startSpanWithName:(NSString *)name
...
+#endif // SENTRY_TARGET_PROFILING_SUPPORTED
+
+    if (activeSpan == nil) {
+        @synchronized(self.activeSpanStack) {
```
noahsmartin, line 60: Do we need to only access activeSpan in this synchronized block? I see we are synchronizing on "activeSpanStack" not "activeSpan", just pointing out it is a change from the code before and if this `startSpanWithName` method is called from multiple threads it could be a problem because you'd now be potentially setting/reading activeSpan concurrently

noahsmartin (approved):

Issue #1620, opened 2021-12-06
Shows unknown "Called from" stack frame in crash report


sindresorhus:
### Environment

SaaS (https://sentry.io/)

### Version

Sentry Cocoa 7.5.2

### Steps to Reproduce

- Go to https://sentry.io/organizations/sindresorhus/issues/2833675135/events/8f9b4afd74da49fbb3e3fef911c4433e/

### Expected Result

I expected to see the name of each stack frame.

### Actual Result

The first stack frame is "unknown" and "redacted".
<img width="944" alt="Screen Shot 2021-12-06 at 18 10 10" src="https://user-images.githubusercontent.com/170270/144890540-f58e836a-93af-460b-b728-ae6b9bf1ccd1.png">

This is a macOS app using the SwiftUI `App` protocol.

philipphofmann:
@jan-auer the logic for loading binary images is in [SentryCrashDynamicLinker](https://github.com/getsentry/sentry-cocoa/blob/7cf0f7441aed40e4c344404d10172547fb3edd4a/Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c#L336). The original file [KSDynamicLinker](https://github.com/kstenerud/KSCrash/commits/master/Source/KSCrash/Recording/Tools/KSDynamicLinker.c) doesn't contain any new patches that we don't have yet. I already applied the latest patch of `accumulative patch` with https://github.com/getsentry/sentry-cocoa/pull/1596.

jan-auer:
Sorry, I made a mistake when looking at this the first time. `libdyld` is there in all cases, it's just unreferenced and the UI hides it. If you reset filters you'll find it. 

This means, the bottom frame returned by the stack walker is either wrong, or it points at an image that's unloaded at the time of the crash. If you look at the address, it lies shortly after the `Sentry` framework, which suggests that this is a user-space library.

sindresorhus:
Please keep this open.

amadeu01:
@philipphofmann do you have a plan to fix this issue ?

philipphofmann:
No, not yet. We are going to discuss this again internally, as I also find it annoying, and we should fix it.

amadeu01:
@philipphofmann have you had any update on that one?

philipphofmann:
@amadeu01, sadly not yet, no.

philipphofmann:
[...]

PR #5418, opened 2025-06-16
Fix: Manually add Dyld image to images cache + locate addresses in dyld

itaybre:
## :scroll: Description

Manually add Dyld image to images cache + locate addresses in dyld

With these changes they now show up, instead of appearing as `unknown`
*Now:*
image
Note: there is still an unkown address sometimes, which not even Apple's crash reports find:
```
20  UIKitCore                     	       0x185c597d4 -[UIApplication _run] + 796
21  UIKitCore                     	       0x185c5dba0 UIApplicationMain + 124
22  iOS-Swift                     	       0x10025b9c4 main + 64 (AppDelegate.swift:5)
23  ???                           	       0x1004993d4 ???
24  dyld                          	       0x10060eb98 start + 6076
```

*Before:*
image

Since iOS 15 and macOS 12, there are prelinked libraries when running apps, which do not show up when calling `_dyld_image_count`, thus we do not show there in our reports.

Fixes: #1620 

Unit testing + Our test sample app

## :pencil: Checklist


Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
```diff
@@ -118,21 +145,46 @@ imageIndexContainingAddress(const uintptr_t address)
...
+            if (loadCmd->cmd == LC_SEGMENT) {
+                const struct segment_command *segCmd = (struct segment_command *)cmdPtr;
+                if (strcmp(segCmd->segname, "__TEXT") == 0) {
+                    uintptr_t segStart = (uintptr_t)dyldHeader + segCmd->vmaddr;
+                    if (address >= segStart && address < segStart + segCmd->vmsize) {
+                        return DYLD_INDEX;
+                    }
+                }
+            } else if (loadCmd->cmd == LC_SEGMENT_64) {
+                const struct segment_command_64 *segCmd = (struct segment_command_64 *)cmdPtr;
```
noahsmartin: I wonder if we could dedup all this arch specific code with some preprocessor checks, like:
```
#if __LP64__
#define SEGMENT_TYPE LC_SEGMENT_64
#else
#define SEGMENT_TYPE LC_SEGMENT
#end
```

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
```diff
@@ -118,21 +145,46 @@ imageIndexContainingAddress(const uintptr_t address)
...
- * This is required for any symtab command offsets.
+/** Get the segment base address of the specified image header.
  *
- * @param idx The image index.
+ * @param header The header to get the segment base address for.
  * @return The image's base address, or 0 if none was found.
  */
 static uintptr_t
-segmentBaseOfImageIndex(const uint32_t idx)
+segmentBaseOfHeader(const struct mach_header *header)
```
noahsmartin: Nothing calls this new function right? Why not keep it with just having the function `segmentBaseOfImageIndex` and not making this a separate function?

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
```diff
@@ -213,18 +279,51 @@ sentrycrashdl_dladdr(const uintptr_t address, Dl_info *const info)
...
-    if (idx == UINT_MAX) {
-        return false;
+
+    const struct mach_header *header = NULL;
+    uintptr_t imageVMAddrSlide = 0;
+    uintptr_t segmentBase = 0;
+
+    if (idx == DYLD_INDEX) {
+        // Handle dyld manually
+        struct dyld_all_image_infos *infos = getAllImageInfo();
```
noahsmartin: Is this signal safe? This sounds like a function that needs to be signal safe

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
itaybre: Yeah, I can revert this change.
I was using it initially in `sentrycrashdl_dladdr`, we can go back to the old implementation

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
itaybre: Updated to be signal safe

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
```diff
@@ -213,18 +271,50 @@ sentrycrashdl_dladdr(const uintptr_t address, Dl_info *const info)
...
+                    imageVMAddrSlide = (uintptr_t)header - vmaddr;
+                    segmentBase = (uintptr_t)header;
+                    break;
+                }
+            }
+            cmdPtr += loadCmd->cmdsize;
+        }
+
+        info->dli_fname = "dyld";
+    } else if (idx != UINT_MAX) {
```
noahsmartin, line 316: Do we need one more condition to match the behavior from before?
```
else {
    return false
}
```

noahsmartin (approved):

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
itaybre: Not technically, because `segmentBase` will be equal to 0 and it will return false in the next check, but it might be good to make it explicit with a special case in case we make changes here in the future.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
 - Move SentryExperimentalOptions to a property defined in Swift (#5329)
 - Fix building with Xcode 26 (#5386)
 
+### Fixes
+
+- Manually add Dyld image to images cache + locate addresses in dyld (#5418)
```
armcknight: should mention here that this solves some of the "unknown" images in stacktraces

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -150,6 +151,42 @@ sentrycrashbic_iterateOverImages(sentrycrashbic_imageIteratorCallback callback,
...
+ * Since Apple no longer includes dyld in the images listed by _dyld_image_count and related
+ * functions, we need to check if dyld is already present in our cache before adding it.
+ *
+ * @return true if dyld is not found in the loaded images and should be added to the cache,
+ *         false if dyld is already present in the loaded images.
+ */
+bool
+sentrycrashbic_shouldAddDyld(void)
+{
+    // dyld is different from libdyld.dylib, the later contains the public API
```
armcknight:
```suggestion
    // dyld is different from libdyld.dylib, the latter contains the public API
```

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -150,6 +151,42 @@ sentrycrashbic_iterateOverImages(sentrycrashbic_imageIteratorCallback callback,
...
+    // while the former contains the runtime information
```
armcknight: I don't quite understand the distinction between public API and runtime information, could you elaborate?

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.c
```diff
@@ -37,6 +38,14 @@
 #include "SentryCrashMemory.h"
 #include "SentryCrashPlatformSpecificDefines.h"
 
+#if __LP64__
+#    define SEGMENT_TYPE LC_SEGMENT_64
+#    define segment_command_t struct segment_command_64
```
armcknight: We should prefix these
```suggestion
#    define SENTRY_SEGMENT_TYPE LC_SEGMENT_64
#    define sentry_segment_command_t struct segment_command_64
```

Sources/SentryCrash/Recording/Tools/SentryCrashDynamicLinker.h
```diff
@@ -36,6 +36,8 @@ extern "C" {
 #include <stdbool.h>
 #include <stdint.h>
 
+#define DYLD_INDEX UINT_MAX - 1
```
armcknight: also prefix this
```suggestion
#define SENTRY_DYLD_INDEX UINT_MAX - 1
```

Tests/SentryTests/SentryCrash/SentryCrashDynamicLinkerTests.m
```diff
@@ -0,0 +1,154 @@
...
+extern uint32_t imageIndexContainingAddress(const uintptr_t address);
+extern bool sentrycrashbic_shouldAddDyld(void);
+extern uintptr_t firstCmdAfterHeader(const struct mach_header *const header);
+#if __LP64__
+#    define SEGMENT_TYPE LC_SEGMENT_64
+#    define segment_command_t struct segment_command_64
+#else
+#    define SEGMENT_TYPE LC_SEGMENT
+#    define segment_command_t struct segment_command
+#endif
```
armcknight: I think we could at least expose some of this in the implementation's header, or a +Test.h header so we don't have to worry about the definitions drifting

Tests/SentryTests/SentryCrash/SentryCrashDynamicLinkerTests.m
```diff
@@ -0,0 +1,154 @@
...
+        if (loadCmd->cmd == SEGMENT_TYPE) {
+            const segment_command_t *segCmd = (segment_command_t *)cmdPtr;
+            if (strcmp(segCmd->segname, "__TEXT") == 0) {
+                return (uintptr_t)header + segCmd->vmaddr;
+            }
+        }
+        cmdPtr += loadCmd->cmdsize;
+    }
+    return (uintptr_t)header;
+}
```
armcknight: This looks like a combination of two separate pieces of implementation from the SDK, any chance we can just expose the function through a +Test.h header?

armcknight (commented):
Nice work. We should probably address the codecov annotations here if at all possible, to address some of the non-happy-path possibilities, to make sure things don't break down if we return a NULL or 0 somewhere that doesn't handle it correctly.

PR #5427, opened 2025-06-17
test: Improve trigger fully blocking app hang

philipphofmann:
Add a second button to decode an image in a loop until 5 seconds have passed. This is a follow-up on GH-5413. 

#skip-changelog


philippsuess:
> Is a fixed time a better app hang use case than a fixed number of tasks?

The problem is that in CI the for loop with a fixed number of tasks can take way longer than 5 seconds. If CI is overloaded it can take x seconds. We only want the app to hang for 5 seconds.

I also updated the code comment to explain this.

Samples/iOS-Swift/iOS-Swift/Tools/TriggerAppHang.swift
```diff
@@ -20,7 +20,26 @@ func triggerNonFullyBlockingAppHang() {
...
-  sleep(5)
+    let buttonTitle = button.currentTitle
+    var i = 0
+
+    let currentTime = Date()
+    let timeToFinishUpdatingUI = currentTime.addingTimeInterval(5)
+
+    while timeToFinishUpdatingUI > Date() {
+        i += Int.random(in: 0...10)
+        i -= 1
```
noahsmartin: why do this minus 1 here? Might be nice to have a comment

noahsmartin (approved):
I don't think this would be flaky (although I worry a bit about all the calls to `setTitle` which we don't control it. It's possible that it causes some flakiness to call this in a tight loop, since that's definitely not how the API was intended to be used.

But more broadly, why is this more realistic than calling a sleep? I don't think it's possible for any frames to be rendered while this loop is executing. It runs on the main run loop, so it's no different than sleeping the main run loop, right?

Samples/iOS-Swift/iOS-Swift/Tools/TriggerAppHang.swift
```diff
@@ -20,7 +20,26 @@ func triggerNonFullyBlockingAppHang() {
     }
 }
 
-/// Sleep on the main thread for long enough to trigger a hang
+/// Do heavy work on the main thread, blocking it for 5 seconds.
+/// This is a different scenraio than calling sleep on the main thread.
+/// With sleep the main thread is blocked, but  not doing anything. With the
+/// following code, the main thread is too busy rendering frames.
```
noahsmartin: I don't think the main thread renders frames. Frames are rendered when the main thread is idle, by the render server. With both this code and the sleep the runloop is blocked, so no updates make it to the render server until this function returns

philprime:
> But more broadly, why is this more realistic than calling a sleep? I don't think it's possible for any frames to be rendered while this loop is executing. It runs on the main run loop, so it's no different than sleeping the main run loop, right?

Please see [my comment in the other PR](https://github.com/getsentry/sentry-cocoa/pull/5413#issuecomment-2979248616) why I do not think it's the same.

Also App Hang Detection is also used to detect if too much work is done on the main thread, e.g. a large image is decoded. This is a lot of busy work and not the same as sleep.

trevor-e:
The test case Noah added seems like the simplest base test case approach IMO that satisfies what ours docs say for the fully-blocking case:

> A fully-blocking app hang is when the main thread is stuck completely, and the app can't render a single frame.

The busy loop seems like another interesting but _separate_ case to include, maybe we can keep both as @philipphofmann mentioned in the other discussion. If we had to only pick one to use for CI I'd vote for the sleep implementation though.

armcknight:
Agree with @trevor-e and @noahsmartin here. It doesn't matter what is holding up the main thread from shuffling off a new CA commit to the render server. If a customer had a sleep on the main thread, we should also detect that as an app hang. If anyone can come up with a test case where a sleep _wouldn't_ be classified as a hang, then we have something to go on here. Otherwise it's a distinction without a difference.

Imagine if we simply mocked threads and run loops. In that case, we'd probably use a timer to fire updates from the mock main thread to the anr detector, or manipulate the date/time provider to fast-forward the system clock, which would all basically be the same thing as using sleep on the main thread itself.

philippsuess:
## Summary of internal Slack Discussion

When we built the app hang detection V1, we also investigated using CFRunLoopObserver with dispatch semaphores to detect app hangs. We identified that this solution doesn't work well when the run loop executes all events in time, but the app still seems stuck to the user. This approach didn't detect the for loop in this PR as an app hang. That's why we check if work on the main thread gets executed in time. That is app hang v1. The code docs explain this a bit: https://github.com/getsentry/sentry-cocoa/blob/ba8a46b7ac0863cf2c8e45b5386ffb4ad6c141d4/Sources/Sentry/include/SentryANRTrackerV1.h#L19-L22
Now we have app hang v2, which solely looks at delayed frames, which allows us to also detect app hangs that don't fully block the main thread: https://github.com/getsentry/sentry-cocoa/blob/ba8a46b7ac0863cf2c8e45b5386ffb4ad6c141d4/Sources/Sentry/include/SentryANRTrackerV2.h#L13-L22

@noahsmartin and I agreed that instead of setting the title, we should decode a large image to keep the main thread and the  CFRunLoopObserver busy, which is a common mistake in iOS development.

Issue #5393, opened 2025-06-10
Investigate strong references after SDK close


philprime:
### Description

During development I noticed that when closing the SDK using the `Close SDK` button in the iOS sample app, there are still strong references internally which are not cleaned up properly.

This was noticed by the following warning, which is also user-facing:

![Image](https://github.com/user-attachments/assets/97b0d860-4468-41fe-a69a-b7335bf9d2ed)

This is actually caused by session replay not cleaning up the background processing, with the scope being already discarded, causing it to create **and configure** a new scope, which then accesses the `UIWindow` from the background thread.

https://github.com/getsentry/sentry-cocoa/blob/603fec8f5dcd6e209a79e2dfa4f7089f01cc5d20/Sources/Sentry/SentryCrashWrapper.m#L187-L194

Furthermore the memory graph after the close still shows a lot of non-deallocated instances of the SDK:

![Image](https://github.com/user-attachments/assets/2a8ed8f8-6136-4b6a-a4bb-dced19b98211)

philprime:
An initial investigation has show that some of these references might actually be false positives as the dependency container gets recreated in `reset` rather than deallocated.

![Image](https://github.com/user-attachments/assets/4d246367-4e31-452a-b9c3-4b7def0cc377)

In addition I did find classes which have their own singleton, are not deallocated when calling `SentrySDK.close()`, e.g. the `SentryGlobalEventProcessor`. This will require refactoring to move them to the dependency container.

PR #5428, opened 2025-06-17
refactor: Move singleton of SentryGlobalEventProcessor to dependency container

philprime:
Moves the singleton managed using `dispatch_once` to the dependency container, so that it gets deallocated on `SentrySDK.close()` correctly and to allow mocking in unit tests.

Related to #5393

#skip-changelog

### DIFF
--- a/SentryTestUtils/ClearTestState.swift
+++ b/SentryTestUtils/ClearTestState.swift
@@ -44,7 +44,6 @@ class TestCleanup: NSObject {
         #endif // os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
         
         SentryDependencyContainer.reset()
-        Dynamic(SentryGlobalEventProcessor.shared()).removeAllProcessors()
         SentryPerformanceTracker.shared.clear()
 
         SentryTracer.resetAppStartMeasurementRead()
--- a/Sources/Sentry/SentryDependencyContainer.m
+++ b/Sources/Sentry/SentryDependencyContainer.m
@@ -25,6 +25,7 @@
 #import <SentryDebugImageProvider.h>
 #import <SentryDefaultRateLimits.h>
 #import <SentryDependencyContainer.h>
+#import <SentryGlobalEventProcessor.h>
 #import <SentryHttpDateParser.h>
 #import <SentryInternalDefines.h>
 #import <SentryNSNotificationCenterWrapper.h>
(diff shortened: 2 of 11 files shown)


itaybre (commented):
Should this be added as an improvement?

SentryTestUtils/ClearTestState.swift
```diff
@@ -44,7 +44,6 @@ class TestCleanup: NSObject {
         #endif // os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
         
         SentryDependencyContainer.reset()
-        Dynamic(SentryGlobalEventProcessor.shared()).removeAllProcessors()
```
itaybre, line 47: I don't see us calling removeAllProcessors() anymore.
Can this function be removed now?

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -429,4 +431,8 @@ - (SentryWatchdogTerminationContextProcessor *)watchdogTerminationContextProcess
 }
 #endif
 
+- (SentryGlobalEventProcessor *)globalEventProcessor SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOCK
+{
+    SENTRY_LAZY_INIT(_globalEventProcessor, [[SentryGlobalEventProcessor alloc] init])
```
philippsuess, line 436: `m`: Please add this property to `testThreadSafety_ResetAndAccessProperties`.

philipphofmann (approved):
Thanks for the cleanup.

SentryTestUtils/ClearTestState.swift
philprime: `removeAllProcessors()` calls `removeAllObjects` on the array of processors in the event processor. `SentryDependencyContainer.reset()` recreates the shared container. This deallocates the global event processor as of this PR, which will then also deallocate the array, therefore yielding the same result.

philprime:
@itaybre in case you mean adding this to the changelog as an improvement, I don't think we need to because there are no changes for the SDK user as the global event processor is an internal type.

SentryTestUtils/ClearTestState.swift
itaybre: @philprime I don't see why `removeAllProcessors()` can't be deleted now since no ones is calling it anymore.
It was only called by `TestCleanup` from what I saw in the codebase

PR #5431, opened 2025-06-18
fix: Add out of bounds handling for array accessor

philprime:
The current implementation of `Array.element(at index: Int)` does not handle out of bounds indicies for negative and positive numbers.

This PR adds correct handling and tests to verify the behaviour.

#skip-changelog


philipphofmann (approved):
Thanks for doing this in an extra PR.

PR #5432, opened 2025-06-18
Use (intel) large runners for UI tests

itaybre:
Test using macOS large runners to check stability

`macos-NUM-large` use intel machines
`macos-NUM-xlarge` use M1 machines

#skip-changelog


philprime:
FYI @itaybre this was tried and rollbacked with https://github.com/getsentry/sentry-cocoa/pull/4881. If you want try a different approach, go for it, just know we already looked into this.

itaybre:
> FYI @itaybre this was tried and rollbacked with #4881. If you want try a different approach, go for it, just know we already looked into this.

There is a slight difference, large runners use Intel machines while xlarge use M1 machines.
This is slightly different, but I still see failed tests either way :(

itaybre:
Closing this pr since we can't confirm this solves any issue right now

PR #5434, opened 2025-06-18
fix: Restore disable everything override

itaybre:
Disable everything flag was broken in the latest refactor.
Adding them back to the same flags as before.

#skip-changelog


Samples/SentrySampleShared/SentrySampleShared/SentrySDKOverrides.swift
```diff
@@ -285,3 +291,68 @@ extension SentrySDKOverrides.Special {
         }
     }
 }
+
+// MARK: Disable Everything Helpoer
```
armcknight:
```suggestion
// MARK: Disable Everything Helper
```

Issue #5366, opened 2025-06-05
Investigate and fix multiple calls to cleanup continuous profiler v2


philprime:
### Description

This issue is a follow-up for the workaround implemented in #5363.

Our UI Tests were failing because the `SentryFramesTracker` and `SentryTimeToDisplayTracker` both try to finish the same span, therefore causing the `SENTRY_TEST_FATAL` to fail the test.

https://github.com/getsentry/sentry-cocoa/blob/49e2bd0ee860b000d1ed31b9c628c636dfec8b9c/Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm#L80-L87

We have replaced the assertion with #5363 as it was blocking our release process, while not having any effect for production usage of the SDK, as the assertion was test-only.

@armcknight due to limited knowledge of @philipphofmann and myself, please investigate why the assertion is hit. I was able to reproduce the issue reliably by following these steps:

1. Revert the changes of #5363 
2. Run the UI test ProfilingUITests.testAppLaunchesWithTraceProfiler()
3. During execution, stop the tests
5. Set a breakpoint in `sentry_trackRootSpanForContinuousProfilerV2()` where the `_gInFlightRootSpans` is incremented
4. Set a breakpoint in `_unsafe_cleanUpContinuousProfilerV2()`  where the `_gInFlightRootSpans` is decremented
6. Run the UI test `LaunchUITests.testBreadcrumbData()`
7. The increment will be hit once
8. The decrement will be hit twice, which would reduce `_gInFlightRootSpans` to `-1` and causes the test run to crash due to `SENTRY_TEST_FATAL`

I am trying to reproduce right now and will add stack traces to this issue when available.

PR #5436, opened 2025-06-18
test(UI): always wipe all data on disk before starting a new UI test

armcknight:
The profile UI tests all wind up removing launch config files before they end. But if they abnormally exited before being able to do so, then another UI test would wind up running a launch profile.

This change makes sure there is no on-disk data leakage between UI tests in any way, whether that be launch profiles, envelopes, etc.

Noticed this while investigating https://github.com/getsentry/sentry-cocoa/issues/5366, but this doesn't fix the root problem of imbalanced bookkeeping mentioned there.

#skip-changelog

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift-UITests/BaseUITest.swift
+++ b/Samples/iOS-Swift/iOS-Swift-UITests/BaseUITest.swift
@@ -29,6 +29,9 @@ extension BaseUITest {
     func newAppSession() -> XCUIApplication {
         let app = XCUIApplication()
         app.launchEnvironment["--io.sentry.ui-test.test-name"] = name
+        app.launchArguments.append(contentsOf: [
+            "--io.sentry.wipe-data"
+        ])
         return app
     }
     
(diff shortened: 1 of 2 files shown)


philipphofmann (commented):
I don't think we can do this for all UI tests, cause some of the app launch profiling test need to run without wiping the data. That's maybe also why the UI tests are failing.

armcknight:
I think I originally had it in `BaseUITest.setUp` so that it would only set this for the _first_ launch per test case. For some reason I moved it to `newAppSession` but you're right, the profiling tests call that method multiple times per case. Moving it to `setUp` is the right call after all, and I think we should move forward with this.

Samples/iOS-Swift/iOS-Swift-UITests/BaseUITest.swift
```diff
@@ -12,6 +12,9 @@ class BaseUITest: XCTestCase {
         continueAfterFailure = false
         XCUIDevice.shared.orientation = .portrait
         app.launchEnvironment["--io.sentry.sdk-environment"] = "ui-tests"
+        app.launchArguments.append(contentsOf: [
+            "--io.sentry.wipe-data"
```
noahsmartin, line 16: I think @itaybre had tried uninstalling and then re-installing the app, which seems even more future proof, but maybe had other tradeoffs

noahsmartin (approved):

Samples/iOS-Swift/iOS-Swift-UITests/BaseUITest.swift
itaybre: Uninstalling ended up being flaky, maybe the cause was the CI.
We can try adding that later

PR #5442, opened 2025-06-20
ref: Use final when possible

noahsmartin:
## :scroll: Description

Switched non-public and non objc swift classes to final, to match general swift best practices and let the compiler avoid dynamic dispatch (makes method calls faster)

Locally built SDK

## :pencil: Checklist

#skip-changelog


itaybre (approved):
lgtm

philprime:
While I do agree with this change, we might have to change our mocking strategy at some point, because we are still using subclasses to track invocations in tests.

Ideally that will be solved with protocols and dependency injection instead.

PR #5443, opened 2025-06-20
chore: Bump versions

noahsmartin:
Update the minimum SDK versions to the earliest versions still supported by Xcode. This resolves a warning in Xcode that we are targeting a version that is not supported by Xcode.

Screenshot 2025-06-23 at 8 42 07 AM


philippsuess:
I think it's acceptable to do this even in a non major because you [have to use Xcode 16 to submit apps](https://developer.apple.com/news/upcoming-requirements/?id=02212025a&utm_source=chatgpt.com) and you now can't use iOS 11 anymore as the min OS version. We only must clearly point this out in the changelog.

noahsmartin:
We're going to wait until the next release (8.53.0) is out and then merge this

armcknight:
This should be a breaking change that shouldn't ship until V9.

ETA because I just saw Philipp's comment: two considerations:
- there may be people shipping the SDK to places other than app stores, where they can still use older versions of Xcode/SDKs
- if we really want to call this a fix vs a breaking change, we should wait until the version of Xcode that is forcing this is actually supported to ship to the app store (maybe it already is? sounds like it from some other convos but I don't know for sure)

philippsuess:
> if we really want to call this a fix vs a breaking change, we should wait until the version of Xcode that is forcing this is actually supported to ship to the app store (maybe it already is? sounds like it from some other convos but I don't know for sure)

Yes, you must use Xcode 16 now to submit to the app store https://github.com/getsentry/sentry-cocoa/pull/5443#issuecomment-2995632044

armcknight:
Ah, sorry, I thought this was due to Xcode 26. 

If this is a warning in Xcode 16, is this already breaking customer workflows in any way? Like, does it show up as an error in any way by some combination of a package manager and settings like treating warnings as errors? Or does it actually prevent submissions to the app store? Has this been reported by anyone? Essentially: what is the urgency here?

The thing is that this has the potential to _break_ some peoples' workflows, which is why I advocate to make this a major revision. You can always rationalize something as just a bugfix, but if it breaks contracts, it breaks contracts, there is no rationalizing away CI failures. We should just be up front about that by reporting it in the standardized way with semver. 

This is why I proposed a while ago to bump these minimum versions every fall with the new GAs of Xcode in our own major revisions in lockstep with Xcode. That can has already been kicked down the road for years, despite my prodding, in the name of backwards compatibility, and now we want to break that same compatibility in a poorly communicated fashion. I think we can wait a bit longer and do it the right way.

philippsuess:
Those are valid points, @armcknight. So let's do it properly and do this with a major release. It's worth noting that we don't have to tackle all major items in https://github.com/getsentry/sentry-cocoa/discussions/2946 for v9. Doing the most important ones is also acceptable and save some for v10.

kahest:
I agree with @armcknight and @philipphofmann here - we should prioritise shipping v9 and bump the versions as part of that

philprime:
Can we mark this PR as draft or close it until v9?

philippsuess:
I'm for closing PRs that we don't plan on merging soon. Too many open draft PRs are a bit of noise and I would like to keep the number low.

philprime:
I'll change it to draft because we are not ready for it yet.

PR #5446, opened 2025-06-20
Remove unused store endpoint

noahsmartin:
This is dead code, the store endpoint was deprecated and it's not longer used, also not part of the public API of the sdk

#skip-changelog


noahsmartin:
Initially the PR also removed the function on the DSN to get the store endpoint, but I removed that one so it's not touching public API

armcknight (approved):
Nice

PR #5448, opened 2025-06-20
ci: deduplicate xcframework builds

armcknight:
We build some form of xcframework in multiple workflows, which is redundant work and wasted compute. Because GitHub Actions doesn't support dependencies between workflows, this places any step that needs an XCFramework into the release.yml workflow where they're already built.

#skip-changelog


philprime (approved):
The `release.yml` starts to feel kind of polluted with tests now also being in there, but I agree with the value of deduplication in doing it this way.

armcknight:
Yeah I'm not really sure what to do with the release.yml growth :/ Maybe there's some further way we can reorganize the jobs/steps into reusable actions like "release-build", "release-validate" and "release-deploy".

armcknight:
Oh apparently there is some facility for workflow dependency: https://stackoverflow.com/a/64733705 and https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run

ETA: ofc there is a big caveat that kills it for us: 
> This event will only trigger a workflow run if the workflow file exists on the default branch

PR #5449, opened 2025-06-20
ci: cache xcframework builds

armcknight:
After parallelizing the build for https://github.com/getsentry/sentry-cocoa/issues/4925, one of the proposed further improvements was to cache build products. This tries to do that.

#skip-changelog


.github/workflows/assemble-xcframework-variant.yml
```diff
@@ -69,27 +69,44 @@ jobs:
           MATCH_GIT_PRIVATE_KEY: ${{ secrets.MATCH_GIT_PRIVATE_KEY }}
           MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
           MATCH_USERNAME: ${{ secrets.MATCH_USERNAME }}
+      
+      - name: Restore XCFramework cache
+        uses: actions/cache@v4
+        with:
+          key: ${{ runner.os }}-xcframework-${{inputs.variant-id}}-${{hashFiles('Sources/**')}}
```
itaybre: Shouldn't we consider using the sentry project config for the cache too?

armcknight: yes, that would be necessary 👍🏻

itaybre (approved):

PR #5450, opened 2025-06-20
ci: remove condition that would prevent xcframework slice builds for …

armcknight:
Noticed while working on https://github.com/getsentry/sentry-cocoa/pull/5449 that this condition would prevent actually building the xcframework for release. 

#skip-changelog


philprime (approved):
Nice catch, thank you, LGTM

Issue #5289, opened 2025-05-25
Objective-C interfaces for APIs used by Hybrid SDKs


jamescrosswell:
## Summary of the Problem

The current .NET bindings for the Cocoa SDK rely on [Objective Sharpie](https://github.com/xamarin/ObjectiveSharpie) to generate interop code from Objective-C headers. This process is imperfect and requires post-processing via PowerShell scripts to clean up incorrect or missing bindings (see [`generate-cocoa-bindings.ps1`](https://github.com/getsentry/sentry-dotnet/blob/e3ad69a89d42a879c199eb87a4e50a21c95e91d9/scripts/generate-cocoa-bindings.ps1)), but does automate much of the work.

Issues have arisen due to the Cocoa SDK’s migration of public types from Objective-C to Swift. This breaks the binding process for .NET, as Objective Sharpie does not support Swift. The bindings for any Swift APIs need to be coded by hand.

Two maintenance strategies are currently used to bump to new versions of the Cocoa SDK then:
1. **Auto-generate + patch** (Objective Sharpie + cleanup script).
    - For ObjectiveC APIs
3. **Manually hand-code** bindings where automation falls short.
    - For Private APIs and Swift APIs

This may also impacts other hybrid SDKs (e.g., Unity, Flutter, React Native), suggesting a broader need for stable interop boundaries.

---

## 💡 Potential Solutions

Swift is the long-term direction for public APIs in Cocoa. However, maintaining a *public Objective-C interface for hybrid SDKs only* could be a reasonable compromise.

1. **Introduce a Stable Objective-C Interface Layer**  
   Create and maintain an Objective-C header that exposes only the APIs needed by hybrid SDKs (including .NET). This would insulate them from changes in internal Swift implementations.

2. **CI Coverage for Bindings Compatibility**  
   Investigate ways to automate checks that ensure .NET (and other hybrid SDKs) bindings continue to work after SDK changes. For example:
   - Add CI tasks that build .NET bindings against the latest SDK
   - Use test projects or sample apps for validation ([sentry-cocoa#5234](https://github.com/getsentry/sentry-cocoa/issues/5234))

---

## Additional Context

- .NET adoption has grown from 0 to ~1600 orgs in 3 years, now outpacing Unity in org count.
- Growth is driven by MAUI adoption and App Center shutdown.
- Manual effort for .NET binding maintenance is high, especially for a small team.
- An Objective-C interface could allow ~95% of the binding process to be automated.

---

[...]

PR #5453, opened 2025-06-21
ref: Convert SentryExtraPackages to Swift

noahsmartin:
## :scroll: Description

Converts SentryExtraPackages to Swift, found this as a small internal class that is easy to migrate. I kept the API of the class the same so hybrid SDKs don't have to make any changes, since it is still exposed to objc with `@objc`

Tests already exist for this class

## :pencil: Checklist

#skip-changelog


Sources/Resources/Sentry.modulemap
```diff
@@ -27,7 +27,6 @@ framework module Sentry {
         header "SentrySessionReplayIntegration-Hybrid.h"
         
         header "SentrySdkInfo.h"
-        header "SentryExtraPackages.h"
```
philprime, line 30: `m`: Could this have implications for the Hybrid SDKs? I know Swift types are exposed, but are there any follow-up steps?

cc @antonis @buenaflor

Sources/Swift/Core/Helper/SentryExtraPackages.swift
```diff
@@ -0,0 +1,36 @@
...
+        let newPackage = ["name": name, "version": version]
+
+        syncQueue.async(flags: .barrier) {
+            extraPackages.insert(newPackage)
+        }
+    }
+
+    @objc
+    public static func getPackages() -> NSMutableSet {
+        var copy: Set<[String: String]> = []
```
philprime: `l`: Add a comment explaining why `sync` is used here

```suggestion
        var copy: Set<[String: String]> = []
        // Thread-safe synchronous read on a concurrent queue.
        // Multiple reads can happen concurrently unless a write barrier is active.
        // Returns a copy to prevent external mutation of internal state.
```

Sources/Swift/Core/Helper/SentryExtraPackages.swift
```diff
@@ -0,0 +1,36 @@
...
+    private static let syncQueue = DispatchQueue(label: "io.sentry.SentryExtraPackages.sync", attributes: .concurrent)
+
+    @objc
+    public static func addPackageName(_ name: String?, version: String?) {
+        guard let name, let version else {
+            return
+        }
+
+        let newPackage = ["name": name, "version": version]
+
```
philprime, line 14: `l`: Add a comment explaining why `async(flags: .barrier)` is used here.

```suggestion

        // Thread-safe write using a barrier block on a concurrent queue.
        // Ensures no other reads or writes happen during this insertion.
```

philprime (approved):
LGTM, added potential improvement comments

Sources/Resources/Sentry.modulemap
noahsmartin: They are all still @objc so it doesn't change the public interface of the hybrid SDK framework. Now that I'm writing this, I realize we probably could have an API stability check for the hybrid SDK same as the one we have for the main SDK

philprime: Could also help with https://github.com/getsentry/sentry-cocoa/issues/5289

noahsmartin: Yeah I think that issue is blocked by switching to SPM and the idea is once we are on SPM we'd have a better place to maintain objc compatible APIs. A new SPM target could then generate a header file from the Swift code that could be used by any codegen tools that only work with ObjC

armcknight: FWIW we have a ReactNative integration test since https://github.com/getsentry/sentry-cocoa/pull/5235, as long as that passes it makes me much more confident in the hybrid sdk API stability generally

Sources/Swift/Core/Helper/SentryExtraPackages.swift
```diff
@@ -0,0 +1,41 @@
...
+        // Thread-safe write using a barrier block on a concurrent queue.
+        // Ensures no other reads or writes happen during this insertion.
+        syncQueue.async(flags: .barrier) {
```
philippsuess: `m`: We have a NSRecursiveLock extension https://github.com/getsentry/sentry-cocoa/blob/fc0757dafef9fca8b8101395728bae8800c6a65a/Sources/Swift/Core/Extensions/Locks.swift#L22-L32

An NSRecursiveLock should have a smaller memory footprint than a DispatchQueue. WDYT?

Sources/Swift/Core/Helper/SentryExtraPackages.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    @objc
+    public static func getPackages() -> NSMutableSet {
+        var copy: Set<[String: String]> = []
+        // Thread-safe synchronous read on a concurrent queue.
+        // Multiple reads can happen concurrently unless a write barrier is active.
+        // Returns a copy to prevent external mutation of internal state.
+        syncQueue.sync {
+            copy = extraPackages
+        }
```
philippsuess: `m`: When you use an NSRecursiveLock you could simplify this to 
```suggestion
        return lock.synchronized {
            return NSMutableSet(set: extraPackages as NSSet)
        }
```

Sources/Swift/Core/Helper/SentryExtraPackages.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+
+@objc @_spi(Private) public final class SentryExtraPackages: NSObject {
+    private static var extraPackages = Set<[String: String]>()
+    private static let syncQueue = DispatchQueue(label: "io.sentry.SentryExtraPackages.sync", attributes: .concurrent)
```
philippsuess: `h`: I'm unsure about using the DispatchQueue for synchronization.

When I run this test, it times out. When I use an NSRecursive lock for which we already have an extension (https://github.com/getsentry/sentry-cocoa/pull/5453#discussion_r2163713165), the test succeeds.
```swift
func testAddPackageThreadSafe() {

    let expectation = self.expectation(description: "Package added")
    expectation.expectedFulfillmentCount = 20000

    let dispatchQueue = DispatchQueue(label: "com.sentry.test", attributes: [.concurrent, .initiallyInactive])

    for _ in 0..<10000 {

        dispatchQueue.async {
            SentryExtraPackages.addPackageName("test-package-1", version: "1.0.0")
            expectation.fulfill()
        }

        dispatchQueue.async {
            XCTAssertNotNil(SentryExtraPackages.getPackages())
            expectation.fulfill()
        }
    }

    dispatchQueue.activate()

    wait(for: [expectation], timeout: 1.0)
}
```

Sources/Swift/Core/Helper/SentryExtraPackages.swift
noahsmartin: I always prefer to use dispatch queues because in my experience they have been the fastest solution for something like this, and they are a bit more "pure" swift - not needing to use something from objc land with the NS prefix. However, the purpose of this PR is just to convert to Swift, so happy to change to any prior art we have like this in the codebase. One day I will try to get us to use swift concurrency instead 😂 

(that's a long ways out, I wouldn't risk using it right now...)

philippsuess: > One day I will try to get us to use swift concurrency instead 😂

Well, that could take years, but I hope we will get there 😄

Sources/Swift/Core/Helper/SentryExtraPackages.swift
noahsmartin: I would guess part of it is due to the "initiallyInactive", the locking version would be able to start running right away. As written this code would create 10k queued blocks and consume a lot of memory before any start running. Also I see in the docs for "activate" that it seems to force the queue to map to only one thread? Usually a queue could run on multiple threads. Haven't tried it myself, but that stands out as a difference

philippsuess: When you use `.concurrent` this test case spins up multiple threads for the same queue. When running it locally I even get 65 threads.

Screenshot 2025-06-24 at 16 12 37

I like to use `.initiallyInactive` and `.concurrent` for quick operations, as the one above, because then you truly get multiple threads executing yoru work at the same time. If you skip the `.initiallyInactive` and `activate` the dispatch queue already starts executing your work while you still adding new work items and then you don't have that many threads at the same time. When I remove the initiallyInactive, I get 42 threads for this test case. The number of threads varies and depends on the test code and your CPU, etc. When I write tests like these, I want to have as many threads as possible hammering the synchronization logic, so that's why I usually stick to .initiallyInactive, .concurrent` and `activate`.
Screenshot 2025-06-24 at 16 16 30

noahsmartin: With dispatch queues you do have to watch out for thread explosion and it seems like that's happening here. Because the max number of threads are being created before any of them start running the system is never able to create a thread for the queue internal to SentryExtraPackages, and it deadlocks. Swift concurrency promises to solve this by managing the threads for us, but yeah a bunch of other tricky situations still pop up there...

PR #5457, opened 2025-06-21
ref: Convert SentryNSURLRequest to Swift

noahsmartin:
## :scroll: Description

Converts SentryNSURLRequest to Swift

Build + existing tests

## :pencil: Checklist


Sources/Swift/Tools/SentryURLRequestFactory.swift
```diff
@@ -0,0 +1,60 @@
...
+        string += newHeaderPart(key: "sentry_version", value: serverVersionString) + ","
+        string += newHeaderPart(key: "sentry_client", value: "\(SentryMeta.sdkName)/\(SentryMeta.versionString)") + ","
+        string += newHeaderPart(key: "sentry_key", value: url.user ?? "")
+        
+        if let password = url.password {
+            string += "," + newHeaderPart(key: "sentry_secret", value: password)
+        }
+        
+        return string
+    }
```
philprime, line 59: `l`: We could move this to it's own `SentryHTTPHeader` data type class with an `encode` method

noahsmartin: I think once we have a bit more in Swift we could do a lot with data, like making them real data types (struct/enum) and using built in encoding/decoding, at least for json. But for now just want to focus on getting it into Swift so we can have a better view of which of those changes make sense

Issue #5399, opened 2025-06-12
Investigate SDK crashes due dispatch_group_leave


philprime:
### Description

A few SDK crashes have surfaced with the following stack trace:

```
libdispatch +0x039e5c dispatch_group_leave.cold.1
libdispatch +0x004e1c dispatch_group_leave
REDACTED +0x50abf4 closure in SentryOnDemandReplay.createVideoWith In App
REDACTED +0x50aeec closure in SentryOnDemandReplay.renderVideo In App
REDACTED +0x50b7bc closure in SentryOnDemandReplay.renderVideo In App
REDACTED +0x512654 closure in SentryOnDemandReplay.renderVideo In App
REDACTED +0x02d9f4 thunk for closure (<compiler-generated>)
```

This is an indicator that a DispatchGroup is leaved multiple times, causing internal assertion to throw.

High chance this is related to the following synchronization code:

https://github.com/getsentry/sentry-cocoa/blob/6cbba5de48b704374bef6fb03e4e094f3a8ef5f6/Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift#L179-L207

Next steps:
- Investigate potential reasons for the `group.leave()` to be called multiple times
- Investigate potential reasons for the `group.leave()` to be called after a timeout has occured.
- Add safe-guards to handle `group.leave()` being called multiple times.
- Consider a wrapper around `DispatchGroup` similar to `SentryDispatchQueueWrapper`

itaybre:
Fixed in #5608

PR #5280, opened 2025-05-22
fix(session-replay): Use low priority queue for dispatch queue wrapper

antonis:
## :pencil: Checklist

You have to check all boxes before merging:

- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #5459, opened 2025-06-22
Convert dispatch queue wrapper

noahsmartin:
## :scroll: Description

Convert SentryDispatchQueueWrapper to Swift

dispatch_once doesn't have a Swift equivalent but the wrapper was only used in one place, so I just converted that to be used directly with a check for if tests are trying to override it with a stub.

Also the legacy queue priority is not replaced with QoS in Swift. I created new initializers for utility and userInitiated, which are the only QoS that was being used from the objc queue attributes.

#skip-changelog

Updated existing tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,11 +23,6 @@
 - Move SentryExperimentalOptions to a property defined in Swift (#5329)
 - Add redaction in session replay for `SFSafariView` used by `SFSafariViewController` and `ASWebAuthenticationSession` (#5408)
 - Convert SentryNSURLRequest to Swift (#5457)
-- Convert SentryDispatchQueueWrapper to Swift (#5459)
-
-### Fixes
-
-- Manually add `dyld` image to image cache to fix dyld symbols appearing as `unknown` (#5418)
 
 ## 8.52.1
 
--- a/Sources/Sentry/SentryTransportFactory.m
+++ b/Sources/Sentry/SentryTransportFactory.m
@@ -45,8 +45,10 @@ NS_ASSUME_NONNULL_BEGIN
     SentryEnvelopeRateLimit *envelopeRateLimit =
         [[SentryEnvelopeRateLimit alloc] initWithRateLimits:rateLimits];
 
-    SentryDispatchQueueWrapper *dispatchQueueWrapper =
-        [[SentryDispatchQueueWrapper alloc] initWithUtilityNamed:@"io.sentry.http-transport"];
+    dispatch_queue_attr_t attributes = dispatch_queue_attr_make_with_qos_class(
+        DISPATCH_QUEUE_SERIAL, DISPATCH_QUEUE_PRIORITY_LOW, 0);
+    SentryDispatchQueueWrapper *dispatchQueueWrapper = [[SentryDispatchQueueWrapper alloc]
+        initWithQueue:dispatch_queue_create("io.sentry.http-transport", attributes)];
 
     SentryNSURLRequestBuilder *requestBuilder = [[SentryNSURLRequestBuilder alloc] init];
 
--- a/Sources/Swift/Core/Helper/SentryDispatchQueueWrapper.swift
+++ b/Sources/Swift/Core/Helper/SentryDispatchQueueWrapper.swift
@@ -18,6 +18,11 @@ import Foundation
         super.init()
     }
     
+    @objc public init(queue: DispatchQueue) {
+        self.queue = queue
+        super.init()
+    }
+    
     @objc public init(utilityNamed name: String) {
         self.queue = DispatchQueue(label: name, qos: .utility)
         super.init()
@@ -30,11 +35,6 @@ import Foundation
     
     @objc public init(utilityNamed name: String, relativePriority: Int) {
         self.queue = DispatchQueue(label: name, qos: DispatchQoS(qosClass: .utility, relativePriority: relativePriority))
-        super.init()
-    }
-    
-    @objc public init(userInitiatedNamed name: String) {
-        self.queue = DispatchQueue(label: name, qos: .userInitiated)
         super.init()
     }
 
@@ -114,7 +114,7 @@ import Foundation
 
 // DispatchWorkItem is not visible to objc so this class wraps it
 // and is used in @objc APIs.
-@_spi(Private) @objc public class DispatchWorkItemWrapper: NSObject {
+@_spi(Private) @objc(SentryDispatchWorkItemWrapper) public class DispatchWorkItemWrapper: NSObject {
     let workItem: DispatchWorkItem
     
     init(workItem: DispatchWorkItem) {
--- a/Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
+++ b/Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
@@ -23,17 +23,22 @@
 
     // -- Arrange --
     NSString *queueName = @"sentry-dispatch-factory.test";
+    int relativePriority = -5;
 
     // -- Act --
     SentryDispatchQueueWrapper *wrappedQueue =
-        [[SentryDispatchQueueWrapper alloc] initWithUtilityNamed:queueName];
+        [[SentryDispatchQueueWrapper alloc] initWithUtilityNamed:queueName
+                                                relativePriority:relativePriority];
 
     // -- Assert --
     const char *actualName = dispatch_queue_get_label(wrappedQueue.queue);
     XCTAssertEqual(strcmp(actualName, [queueName cStringUsingEncoding:NSUTF8StringEncoding]), 0);
 
-    dispatch_qos_class_t actualQoSClass = dispatch_queue_get_qos_class(wrappedQueue.queue, nil);
+    int actualRelativePriority;
+    dispatch_qos_class_t actualQoSClass
+        = dispatch_queue_get_qos_class(wrappedQueue.queue, &actualRelativePriority);
     XCTAssertEqual(actualQoSClass, QOS_CLASS_UTILITY);
+    XCTAssertEqual(actualRelativePriority, relativePriority);
 }
 
 - (void)
@@ -43,7 +48,7 @@
     // making sure the factory sets the name and attributes correctly.
 
     // -- Arrange --
-    NSString *queueName = @"sentry-dispatch-factory.test";
+    const char *queueName = "sentry-dispatch-factory.test";
     int relativePriority = -5;
 
     // -- Act --
@@ -52,7 +57,7 @@
 
     // -- Assert --
     const char *actualName = dispatch_queue_get_label(wrappedQueue.queue);
-    XCTAssertEqual(strcmp(actualName, [queueName cStringUsingEncoding:NSUTF8StringEncoding]), 0);
+    XCTAssertEqual(strcmp(actualName, queueName), 0);
 
     int actualRelativePriority;
     dispatch_qos_class_t actualQoSClass
@@ -70,7 +75,9 @@
     XCTestExpectation *expectation = [self expectationWithDescription:@"Timer fired"];
     uint64_t interval = 100 * NSEC_PER_MSEC; // 100ms
     uint64_t leeway = 10 * NSEC_PER_MSEC;
-    NSString *queueName = @"sentry-dispatch-factory.timer";
+    const char *queueName = "sentry-dispatch-factory.timer";
+    dispatch_queue_attr_t attributes
+        = dispatch_queue_attr_make_with_qos_class(DISPATCH_QUEUE_SERIAL, QOS_CLASS_UTILITY, 0);
 
     __block int fireCount = 0;
     void (^eventHandler)(void) = ^{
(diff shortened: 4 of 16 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Crashes for uncaught NSExceptions will now report the stracktrace recorded within the exception (#5306)
 - Move SentryExperimentalOptions to a property defined in Swift (#5329)
 - Add redaction in session replay for `SFSafariView` used by `SFSafariViewController` and `ASWebAuthenticationSession` (#5408)
+- Convert SentryDispatchQueueWrapper to Swift (#5459)
```
philprime: `l`: I don't think this has a user impact and can therefore be skipped in the changelog

Sources/Sentry/SentryCrashIntegration.m
```diff
@@ -203,7 +202,11 @@ - (void)startCrashHandler:(NSString *)cacheDirectory
             [SentryCrashIntegration sendAllSentryCrashReports];
         }
     };
-    [self.dispatchQueueWrapper dispatchOnce:&installationToken block:block];
+    if (self.dispatchQueueWrapper.usesDispatchOnce) {
```
philprime, line 205: `m`: I do not fully understand why this is conditional is moved outside of `dispatchQueueWrapper`. I believe we should encapsulate it with a convenience method

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
```diff
@@ -91,8 +83,7 @@ - (void)testSourceWithInterval_providedEventHandlerIsCalled
     // -- Act --
     SentryDispatchSourceWrapper *source = [self.sut sourceWithInterval:interval
                                                                 leeway:leeway
-                                                             queueName:queueName
-                                                            attributes:attributes
+                                                   concurrentQueueName:queueName
```
philprime: `h`: The test previously created a serial queue with `QOS_CLASS_UTILITY`, not a concurrent queue.

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
```diff
@@ -109,27 +100,22 @@ - (void)testSource_queueUsesCorrectQoSAndPriority
...
-        = dispatch_queue_attr_make_with_qos_class(DISPATCH_QUEUE_SERIAL, QOS_CLASS_UTILITY, -10);
+    NSString *queueName = @"sentry-dispatch-factory.qos-check";
     void (^eventHandler)(void) = ^{ };
 
     // -- Act --
     SentryDispatchSourceWrapper *source = [self.sut sourceWithInterval:interval
                                                                 leeway:leeway
-                                                             queueName:queueName
-                                                            attributes:attributes
+                                                   concurrentQueueName:queueName
```
philprime: `h`: See other comment, regarding `QOS_CLASS_UTILITY` and concurrent

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
```diff
@@ -109,27 +100,22 @@ - (void)testSource_queueUsesCorrectQoSAndPriority
...
     // -- Assert --
     const char *actualLabel = dispatch_queue_get_label(queueWrapper.queue);
-    XCTAssertEqual(strcmp(actualLabel, queueName), 0);
+    XCTAssertEqual(strcmp(actualLabel, [queueName cStringUsingEncoding:NSUTF8StringEncoding]), 0);
 
-    int relativePriority;
-    dispatch_qos_class_t qos = dispatch_queue_get_qos_class(queueWrapper.queue, &relativePriority);
+    dispatch_qos_class_t qos = dispatch_queue_get_qos_class(queueWrapper.queue, nil);
     XCTAssertEqual(qos, QOS_CLASS_UTILITY);
-    XCTAssertEqual(relativePriority, -10);
```
philprime: `h`: As mentioned in the test name, the test also asserts the relative priority, so why remove it?

Tests/SentryTests/Networking/SentryDispatchQueueWrapperTests.m
```diff
@@ -12,70 +12,20 @@ @implementation SentryDispatchQueueWrapperTests
...
-    // -- Assert --
     int actualRelativePriority;
     dispatch_qos_class_t actualQoSClass
         = dispatch_queue_get_qos_class(wrappedQueue.queue, &actualRelativePriority);
     XCTAssertEqual(actualQoSClass, QOS_CLASS_UTILITY);
-    XCTAssertEqual(actualRelativePriority, -5);
-}
-
-- (void)
-    testInitWithNameAndAttributes_customAttributesWithPositiveRelativePriority_shouldCreateQueueWithUnspecifiedQoS
```
philprime: `m`: Is this test really not needed anymore?

Sources/Swift/Core/Helper/SentryDispatchQueueWrapper.swift
```diff
@@ -0,0 +1,123 @@
...
+
+    @objc(createDispatchBlock:)
+    public func createDispatchBlock(_ block: @escaping () -> Void) -> DispatchWorkItemWrapper? {
+        return .init(workItem: DispatchWorkItem(block: block))
+    }
+} 
+
+// DispatchWorkItem is not visible to objc so this class wraps it
+// and is used in @objc APIs.
+@_spi(Private) @objc public class DispatchWorkItemWrapper: NSObject {
```
philprime: `m`: We should keep using the Sentry type prefix

```suggestion
@_spi(Private) @objc public class SentryDispatchWorkItemWrapper: NSObject {
```

Sources/Sentry/SentryTransportFactory.m
```diff
@@ -44,11 +44,8 @@ @implementation SentryTransportFactory
     SentryEnvelopeRateLimit *envelopeRateLimit =
         [[SentryEnvelopeRateLimit alloc] initWithRateLimits:rateLimits];
 
-    dispatch_queue_attr_t attributes = dispatch_queue_attr_make_with_qos_class(
-        DISPATCH_QUEUE_SERIAL, DISPATCH_QUEUE_PRIORITY_LOW, 0);
     SentryDispatchQueueWrapper *dispatchQueueWrapper =
-        [[SentryDispatchQueueWrapper alloc] initWithName:"io.sentry.http-transport"
-                                              attributes:attributes];
+        [[SentryDispatchQueueWrapper alloc] initWithUtilityNamed:@"io.sentry.http-transport"];
```
philprime: `m`: I am not convinced that changing the naming from `low priority` to `utility` is making it clearer.

philprime (commented):
Found a couple of issues to look into

Sources/Sentry/SentryCrashIntegration.m
noahsmartin: This is the one I mentioned in the PR description about `dispatch_once`, it doesn't have a Swift equivalent and the `dispatch_once_t` type is not representable in Swift. So if it's wrapped it has to be wrapped in ObjC. While we could keep around an ObjC wrapper just for this and have the rest of the wrapper moved to Swift, I figured in a Swift only world we wouldn't need/can't use `dispatch_once` at all. Since this was the only use case we still have in the codebase, it seemed best to just special case it and avoid creating a new ObjC class

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
noahsmartin: The test did, but no SDK code ever called this method with anything other than concurrent, so seems like it was testing something that we didn't need tested. I updated the API to specifically make a concurrent queue. That seemed fine/better, but is there another way you'd suggest handling this? I didn't want to add a serial queue function that only ever got called from tests

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
noahsmartin: We no longer support passing attributes to the swift wrapper (not supported by GCD in swift) and since no SDK source code ever tried to use this method to create one with a relative priority, I didn't want to create new wrapper methods that would only be called from tests. I can rename this test though

Sources/Sentry/SentryTransportFactory.m
noahsmartin: We could call it "initWithLowPriorityNamed", it's just that low priority queues are deprecated, the docs here: https://developer.apple.com/documentation/dispatch/dispatch_queue_priority_low 
state: "Use quality-of-service constants instead. This constant maps to the QOS_CLASS_UTILITY class."

So IMO using utility will be more familiar to iOS devs. I only started iOS dev in ~2015 so I didn't actually know about low priority and was only familiar with QoS classes. Had to look up the priority system when I worked on this PR. Still happy to change it though if you'd rather have it use the priority naming in the function name

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
noahsmartin: Yeah same as that other one, this function, `sourceWithInterval:...`, is only used in one point in the SDK, and it's to create a concurrent utility queue, I updated the tests to reflect this use case

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
noahsmartin: Updated the name, so to be more clear about the name:

This was testing the priority before, but no code in the SDK set the priority, I removed the option to set the priority from the Swift API since it wasn't being used and that made this test unnecessary

Tests/SentryTests/Networking/SentryDispatchQueueWrapperTests.m
noahsmartin: Good catch, just added it back

Sources/Sentry/SentryCrashIntegration.m
philprime: Maybe it is cleaner to do with an Objective-C extension of the Swift version class `SentryDispatchWrapper` and moving it to Swift when the time comes, but I understand it better now, so proceed as you believe is best 👍

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
philprime: I had to invest quite some time to resolve the synchronization and thread inversion warnings for session replay with #5018 and #5280. 
In that case I heavily rely on the relative priority here:

https://github.com/getsentry/sentry-cocoa/blob/831372a3617378af22e25f2dbe612ae83b44e573/Sources/Sentry/SentrySessionReplayIntegration.m#L142-L153

These changes of mine probably also caused the issues in #5399, so I am just not sure if this isn't causing more issues than it fixes.


On a side-note: I am on limited time capacity now to fully review the impact of changing the type of queue, so you might be right with your proposed changes.

Sources/Sentry/SentryTransportFactory.m
philprime: I agree with you, I also had to look it up why the namings are different. We can proceed with the newer naming.

Please double-check the usages so we are not mixing "low priority" with "utility priority", to not cause any confusion with partial renaming.

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
noahsmartin: So the SentryDispatchQueueWrapper still supports relative priority, it's just SentryDispatchFactory that doesn't use it, and this test was for SentryDispatchFactory. There is still a test for the priority support in SentryDispatchQueueWrapper and I didn't change anything there (beyond moving it to swift)

Sources/Sentry/SentryTransportFactory.m
noahsmartin: Moved to https://github.com/getsentry/sentry-cocoa/pull/5466 to keep this PR focused on just the swift conversion

Tests/SentryTests/Networking/SentryDispatchFactoryTests.m
noahsmartin: Updated to keep the original behavior, although the SDK only uses concurrent queues with this API, there wasn't really any reason to change it as part of the Swift migration

philipphofmann (changes requested):
Thanks a lot for trying to convert more code to Swift. Unfortunately, I don't think we can do such a refactoring in one go. The risk is way too high to break things. The SentryDispatchQueueWrapper is heavily used in the SDK, and we don't have enough integration tests to ensure migrating this won't introduce any significant issues. If we mess this up, the whole SDK can be broken.

I would instead choose an opportunistic approach. I would create a duplicate class in Swift for the SentryDispatchQueueWrapper. Whenever we need parts from the SentryDispatchQueueWrapper in Swift, we add it. Once we know it's working correctly, we can let ObjC code use the new Swift class step by step. It might be OK, to keep a SentryDispatchQueueWrapper for specific parts of the SDK. 

Does that make sense, @noahsmartin?

philippsuess:
https://github.com/getsentry/sentry-cocoa/pull/5472 replaces this PR due to the concerns pointed out in https://github.com/getsentry/sentry-cocoa/pull/5459#pullrequestreview-2953504811. That's why we closed it.

PR #5460, opened 2025-06-23
ref: Inject dependencies for ViewHierarchyProvider

philipphofmann:
Instead of retrieving the dependencies in the
SentryViewHierarchyProvider via the SentryDependencyContainer, clearly specify them in the initializer, which is better for testability and readability. Another benefit is that the
SentryViewHierarchyProviderTests don't need to call SentryDependencyContainer.reset.

This came up while investigating `test_appViewHierarchy_usesMainThread` being flaky on MacCatalyst. This PR minimizes potential side effects of SentryDependencyContainer.reset.

Smoke test that ViewHierarchy still works: https://sentry-sdks.sentry.io/issues/6607086585/events/1623c22620294553a448e152fac70848/?project=5428557

Screenshot 2025-06-23 at 10 09 16

#skip-changelog

### DIFF
--- a/Tests/SentryTests/SentryViewHierarchyProviderTests.swift
+++ b/Tests/SentryTests/SentryViewHierarchyProviderTests.swift
@@ -7,7 +7,7 @@ class SentryViewHierarchyProviderTests: XCTestCase {
         let uiApplication = TestSentryUIApplication()
 
         var sut: SentryViewHierarchyProvider {
-            return SentryViewHierarchyProvider()
+            return SentryViewHierarchyProvider(dispatchQueueWrapper: SentryDispatchQueueWrapper(), sentryUIApplication: uiApplication)
         }
     }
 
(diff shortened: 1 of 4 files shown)


Tests/SentryTests/SentryViewHierarchyProviderTests.swift
```diff
@@ -175,7 +169,7 @@ class SentryViewHierarchyProviderTests: XCTestCase {
     }
 
     func test_invalidSerialization() {
-        let sut = TestSentryViewHierarchyProvider()
+        let sut = TestSentryViewHierarchyProvider(dispatchQueueWrapper: SentryDispatchQueueWrapper(), sentryUIApplication: fixture.uiApplication)
```
philprime, line 172: `l`: Could be moved to a `fixture.getSut()` helper method, like we do in other tests.
Not strong opinion here, so feel free to proceed at your discretion.

philippsuess: Nei, I don't like the fixtures. Test code can be repetitive and stupid. I'm going to keep it. But thanks for the hint.

PR #5461, opened 2025-06-23
ref: SessionTracker remove observers

philipphofmann:
Replace stop(gracefully) with removeObservers.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/5121#discussion_r2161035116.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
+++ b/Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
@@ -557,7 +557,10 @@ class SentrySessionTrackerTests: XCTestCase {
     }
 
     private func abnormalStopSut() {
+
+        // We remove the observers, to ensure future notifications are not handled.
         sut.removeObservers()
+
         #if os(iOS) || targetEnvironment(macCatalyst) || os(tvOS)
         // When the app stops, the app state is `inactive`.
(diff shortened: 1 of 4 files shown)


Tests/SentryTests/Integrations/Session/SentrySessionTrackerTests.swift
```diff
@@ -551,13 +551,13 @@ class SentrySessionTrackerTests: XCTestCase {
...
     private func stopSut() {
-        sut.stop(withGracefully: true)
+        sut.stop()
         // After stopping the session tracker, we need to set a new hub to the SDK.
         fixture.setNewHubToSDK()
     }
 
     private func abnormalStopSut() {
-        sut.stop(withGracefully: false)
+        sut.removeObservers()
```
philprime, line 562: `h`: We still need to reset the `wasStartSessionCalled`, that's why I didn't just remove the observer.

philippsuess: I don't think so. The tests are all green, cause they all create a new sut. You only need to set `wasStartSessionCalled = NO` in stop, when you're not getting a new instance.

philprime: ok if tests are green, LGTM.

PR #5463, opened 2025-06-23
chore: Remove unused singleton for SentryNSProcessInfoWrapper

philprime:
I could not find a use case of this, so we can remove it.

#skip-changelog


noahsmartin (approved):

itaybre (approved):
LGTM

philprime:
Moving back to draft because it seems to have issues I need to investigate.

Issue #5371, opened 2025-06-05
App extension API only doesn't need to be required


noahsmartin:
### Problem Statement

Right now we set APPLICATION_EXTENSION_API_ONLY=YES because of the side effect that it [makes internal swift public](https://stackoverflow.com/questions/42319892/checking-allow-app-extension-api-only-in-ios-framework-project-exposes-interna) There's no guarantee this will stay the case, and it's definitely not what this setting is meant for.

### Solution Brainstorm

Instead we could make the swift types that should be exposed to objc `public` which is the documented way to do it, and add `@_spi(Private)` to keep this code not visible in public swift code. We've already done some of this here: https://github.com/getsentry/sentry-cocoa/pull/5307 If we keep updating this swift code to use `@_spi` we'll be able to remove `APPLICATION_EXTENSION_API_ONLY=YES` from our config files.

PR #5472, opened 2025-06-24
ref: Wrap SentryDispatchQueueWrapper in Swift

noahsmartin:
This is a redo of https://github.com/getsentry/sentry-cocoa/pull/5459/ in a way that makes no changes to the implementation, just creates the Swift wrapper that replaces the ObjC version to allow further migration to Swift

#skip-changelog


Sources/Swift/Helper/SentryDispatchQueueWrapper.swift
```diff
@@ -0,0 +1,60 @@
...
+        internalWrapper = SentryDispatchQueueWrapperInternal()
+    }
+    
+    public init(name: UnsafePointer<CChar>, attributes: __OS_dispatch_queue_attr?) {
+        internalWrapper = SentryDispatchQueueWrapperInternal(name: name, attributes: attributes)
+    }
+    
+    public var queue: DispatchQueue {
+        internalWrapper.queue
+    }
```
philippsuess: `m`: Don't we need to use `@_spi(Private)` for these?

Sources/Swift/Helper/SentryDispatchQueueWrapper.swift
```diff
@@ -0,0 +1,60 @@
+@_implementationOnly import _SentryPrivate
+
+@objcMembers class SentryDispatchQueueWrapper: NSObject {
```
philippsuess: `m`: We should add a comment explaining why we need two implementations.

Sources/Sentry/include/SentryDispatchQueueWrapper.h
```diff
@@ -5,7 +5,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * A wrapper around DispatchQueue functions for testability.
  */
-@interface SentryDispatchQueueWrapper : NSObject
+@interface SentryDispatchQueueWrapperInternal : NSObject
```
philippsuess: `m`: I think we should also rename the file. Maybe we should give this a name that clearly points out you shouldn't use this class anymore or add this in the code docs. We could also write self-defined linter searching for imports of `SentryDispatchQueueWrapper.h` and then let CI fail if somebody still imports it by mistake. Up to you.

Sources/Swift/Helper/SentryDispatchQueueWrapper.swift
noahsmartin: I specifically left it out of this PR so avoid exploding the number of files it touches. (all the unit tests would have to add the spi import)

For now this works because we still use the app extension only flag: https://github.com/getsentry/sentry-cocoa/issues/5371

I do plan to convert this to @_spi in a separate PR, I just didn't want to make this one harder to review

Sources/Swift/Helper/SentryDispatchQueueWrapper.swift
noahsmartin: added

Sources/Sentry/include/SentryDispatchQueueWrapper.h
noahsmartin: Renamed, and prefixed with underscore. I think we should hold off on any extra linter/CI steps for now because I don't think it will be necessary. If it ends up becoming an issue though I'd be happy to add something

philippsuess: Also fine. We can then easily add something like 
```
! grep -rn --exclude='./Sources/Sentry/_SentryDispatchQueueWrapperInternal.m' --exclude='./Sources/Sentry/include/SentryPrivate.h' "SentryDispatchQueueWrapperInternal.h" ./Sources
```

Sources/Swift/Helper/SentryDispatchQueueWrapper.swift
philippsuess: But wouldn't this PR make then these methods public? If yes, I would wait with merging and open another PR to add `@_spi ` based on this one. Get a review for the other PR, merge it into this one and then merge these chances at once.

noahsmartin: It won't make it public, because the type itself is still `internal`. We can be confident in this now because the API stability check would fail if it was accidentally public. However, I can go ahead and make it SPI now

philippsuess: Then all good. Thanks

philipphofmann (commented):
LGTM, when CI is green.

PR #5475, opened 2025-06-24
chore: Add a few SPI

noahsmartin:
Going through and updating some of these to incrementally work towards https://github.com/getsentry/sentry-cocoa/issues/5371 

#skip-changelog

### DIFF
--- a/Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebEvent.swift
+++ b/Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebEvent.swift
@@ -14,7 +14,7 @@ protocol SentryRRWebEventProtocol: SentrySerializable {
 }
 
 @objcMembers
-class SentryRRWebEvent: NSObject, SentryRRWebEventProtocol {
+@_spi(Private) public class SentryRRWebEvent: NSObject, SentryRRWebEventProtocol {
     let type: SentryRRWebEventType
     let timestamp: Date
     let data: [String: Any]?
@@ -25,7 +25,7 @@ class SentryRRWebEvent: NSObject, SentryRRWebEventProtocol {
         self.data = data
     }
     
-    func serialize() -> [String: Any] {
+    public func serialize() -> [String: Any] {
         var result: [String: Any] = [
             "type": type.rawValue,
             "timestamp": SentryDateUtil.millisecondsSince1970(timestamp)
--- a/Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
+++ b/Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
@@ -46,7 +46,7 @@ class SentryTouchTracker: NSObject {
         self.dispatchQueue = dispatchQueue
     }
     
-    convenience init(dateProvider: SentryCurrentDateProvider, scale: Float) {
+    public convenience init(dateProvider: SentryCurrentDateProvider, scale: Float) {
         // SentryTouchTracker has it own dispatch queue instead of using the one
         // from Dependency container to avoid the bottleneck of sharing the same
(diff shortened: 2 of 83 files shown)


Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebEvent.swift
```diff
@@ -25,7 +25,7 @@ class SentryRRWebEvent: NSObject, SentryRRWebEventProtocol {
         self.data = data
     }
     
-    func serialize() -> [String: Any] {
+    public func serialize() -> [String: Any] {
```
armcknight, line 28: I assume the spi declaration applies to this too since you didn't add it, but just wanted to double check because that isn't the case with `public`.

Is there any way we can use the API stability check to check for non-SPI-guarded leakage as well?

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -46,7 +46,7 @@ class SentryTouchTracker: NSObject {
         self.dispatchQueue = dispatchQueue
     }
     
-    convenience init(dateProvider: SentryCurrentDateProvider, scale: Float) {
+    public convenience init(dateProvider: SentryCurrentDateProvider, scale: Float) {
```
armcknight, line 49: why no SPI here?

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
```diff
@@ -99,7 +99,7 @@ extension SentryUserFeedbackIntegrationDriver: SentryUserFeedbackWidgetDelegate
 // MARK: UIAdaptivePresentationControllerDelegate
 @available(iOS 13.0, *)
 extension SentryUserFeedbackIntegrationDriver: UIAdaptivePresentationControllerDelegate {
-    func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
+    public func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
```
armcknight, line 102: same question

Sources/Swift/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationContextProcessor.swift
```diff
@@ -19,7 +19,7 @@ class SentryWatchdogTerminationContextProcessor: NSObject {
         clear()
     }
 
-    func setContext(_ context: [String: [String: Any]]?) {
+    public func setContext(_ context: [String: [String: Any]]?) {
```
armcknight, line 22: same question about inherited public vs spi here and on line 38 below

Sources/Swift/Persistence/SentryScopeContextPersistentStore.swift
```diff
@@ -15,7 +15,7 @@ class SentryScopeContextPersistentStore: NSObject {
         self.fileManager.moveState(contextFileURL.path, toPreviousState: previousContextFileURL.path)
     }
 
-    func readPreviousContextFromDisk() -> [String: [String: Any]]? {
+    public func readPreviousContextFromDisk() -> [String: [String: Any]]? {
```
armcknight, line 18: same

Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebEvent.swift
noahsmartin: exactly since the containing type is already SPI. The API stability check covers us here already! Since we check in the full API any changes to the public API would fail the check and make us update the json file. As long as the check doesn't fail, this PR isn't changing public API

PR #5476, opened 2025-06-24
ci: cache unit test builds

armcknight:
This takes a little over a minute per run, but tests don't always change, so it's wasted time/compute. Instead, cache the build product so it can be reused.

I also realized i was misapplying the cache action in other recent changes, so removed the unnecessary second calls to it.

#skip-changelog

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -183,9 +183,17 @@ jobs:
       - name: Install Slather
         run: gem install slather
 
+      - name: Restore test build cache
+        id: test-build-cache
+        uses: actions/cache@v4
+        with:
+          key: ${{matrix.scheme}}-${{matrix.runs-on}}-${{matrix.platform}}-${{matrix.test-destination-os}}-xcode-${{matrix.xcode}}-${{hashFiles('Tests/**', 'Sentry.xcodeproj/**', 'Tests/Configuration/**', 'Sentry/Configuration/DeploymentTargets.xcconfig')}}
+          path: uikit-check-build
+
       # We split building and running tests in two steps so we know how long running the tests takes.
       - name: Build tests
         id: build_tests
+        if: steps.test-build-cache.outputs.cache-hit != 'true'
         run: |
           ./scripts/sentry-xcodebuild.sh \
             --platform ${{matrix.platform}} \
@@ -194,7 +202,8 @@ jobs:
             --command build-for-testing \
             --device "${{matrix.device}}" \
             --configuration TestCI \
-            --scheme ${{matrix.scheme}}
+            --scheme ${{matrix.scheme}} \
+            --derived-data unit-test-build
 
       - name: Run tests
         # We call a script with the platform so the destination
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -183,9 +183,17 @@ jobs:
       - name: Install Slather
         run: gem install slather
 
+      - name: Restore test build cache
+        id: test-build-cache
+        uses: actions/cache@v4
+        with:
+          key: ${{matrix.scheme}}-${{matrix.runs-on}}-${{matrix.platform}}-${{matrix.test-destination-os}}-xcode-${{matrix.xcode}}-${{hashFiles('Tests/**', 'Sentry.xcodeproj/**', 'Tests/Configuration/**', 'Sentry/Configuration/DeploymentTargets.xcconfig')}}
+          path: uikit-check-build
```
philippsuess, line 191: `l`: I wonder when the cache will get a hit. I guess for example when you only update the test.yml file but not the source code? Do you have any other good use cases? I'm curious. Maybe it makes sense to add the most important ones as a code comment so everybody understands why this cache makes sense.

philipphofmann (approved):
LGTM, worth a shot.

.github/workflows/test.yml
```diff
@@ -183,9 +183,17 @@ jobs:
       - name: Install Slather
         run: gem install slather
 
+      - name: Restore test build cache
+        id: test-build-cache
+        uses: actions/cache@v4
+        with:
+          key: ${{matrix.scheme}}-${{matrix.runs-on}}-${{matrix.platform}}-${{matrix.test-destination-os}}-xcode-${{matrix.xcode}}-${{hashFiles('Tests/**', 'Sentry.xcodeproj/**', 'Tests/Configuration/**', 'Sentry/Configuration/DeploymentTargets.xcconfig')}}
```
noahsmartin, line 190: Are we sure the `Sources/` file doesn't need to be included here? I thought the test code was statically linked to the sources so if the sources change the tests would be rebuilt?

.github/workflows/test.yml
armcknight: The idea was for cache hits on either workflow changes or source code changes that don't modify tests, like with refactors. This leads to a rabbit hole where we could further partition the tests by module, so that e.g. modifying the tests for SentryHTTPTransport doesn't recompile the tests for SentryUIViewControllerSwizzling, but I just wanted to get started somewhere and then we can drill down into more granular approaches like that later.

.github/workflows/test.yml
armcknight: It's possible that the Xcode build is structured that way, even though it shouldn't be: we should be able to modify source code and rerun the same test binary as long as no test code was changed. So, the cache key should not use a hash of the SDK source code, even if right now it technically would work that way (and we should decouple it if so)

armcknight: After thinking again I see what you're saying now. If we were to e.g. change the value of a static constant that's exposed to the tests, the tests would then fail.

armcknight:
I think there's too much risk involved here and the benefit would be small, so I'm going to decline to merge this.

PR #5456, opened 2025-06-21
SentrySDKPackage in Swift

noahsmartin:
## :pencil: Checklist

You have to check all boxes before merging:

- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #5478, opened 2025-06-24
ref: SentrySDKPackage in Swift

noahsmartin:
Converts SentrySDKPackage to Swift

This is a redo of https://github.com/getsentry/sentry-cocoa/pull/5456 which I accidentally merged into another feature branch rather than main 🤦‍♂️ 

#skip-changelog


armcknight (approved):
I diffed the patches of the two PRs and the only differences were in the pbxproj. As long as these look right to you and CI passes, ship it! (this pr on left, old pr on right)
image
image

PR #5480, opened 2025-06-24
chore: logging timestamp tweaks

armcknight:
Recently noticed the comment drifted from the implementation per https://github.com/getsentry/sentry-cocoa/pull/4781#discussion_r1938967646

Also, having this long prefix add a decent bit of noise, so I propose removing it. We know what that number is.

#skip-changelog


philipphofmann (approved):
Yes, I also found that annoying. LGTM, when tests are green.

PR #5481, opened 2025-06-24
ref: Convert SentrySdkInfo to Swift

noahsmartin:
Convert this file to Swift

#skip-changelog


Tests/SentryTests/Helper/EnvelopeComparison.swift
```diff
@@ -0,0 +1,69 @@
+import XCTest
+
+func compareEnvelopes(_ expectedEnvelope: Data?, _ actualEnvelope: Data?, message: String) throws {
```
philprime: `l`: Could be added as an extension of XCTest so it's limited in scope

Tests/SentryTests/Helper/EnvelopeComparison.swift
```diff
@@ -0,0 +1,69 @@
...
+        let dataAsString = try XCTUnwrap(String(data: data, encoding: .utf8))
+        return .string(dataAsString)
+    }
+    XCTAssertEqual(expectedData, actualData, message)
+}
+
+import Foundation
+
+extension Data {
+    func backwardCompatibleSplit(separator: Data) -> [Data] {
```
philprime, line 46: `m`: Please add documentation and inline-comments to this method so we know the purpose of it, and also know at which point in time we can get rid of the backwards compatilibity

Issue #5492, opened 2025-06-25
Release 8.53.0 seems broken


mman:
### Platform

iOS

### Environment

Production, TestFlight

### Installed

Swift Package Manager

### Version

8.53.0

### Xcode Version

16.4

### Did it work on previous versions?

8.53.1

### Steps to Reproduce

Update my Xcode project to latest package versions, pulls in sentry-cocoa 8.53.0. Push to GitHub and let Xcode Cloud rebuild it. Build for iOS and tvOS fails with:

`The value for key CFBundleVersion [8.53.0+f92cfa9b1199c75411a263d2d9bc2df8ea8029cf] in the Info.plist file must be no longer than 18 characters.`

Reverting to 8.51.0 works.

### Expected Result

Should work.

### Actual Result

`The value for key CFBundleVersion [8.53.0+f92cfa9b1199c75411a263d2d9bc2df8ea8029cf] in the Info.plist file must be no longer than 18 characters.`

ryanbrink:
We've noticed this too. Pinning our builds to 8.52.1 for now.

itaybre:
Thanks for reporting this, will take a look and fix this ASAP

itaybre:
@mman @ryanbrink Fixed on release 8.53.1: https://github.com/getsentry/sentry-cocoa/releases/tag/8.53.1

mman:
Thanks @itaybre for fast reaction. Can confirm the builds are running again. While I have you on the line, should I report the Xcode 26 beta crash in `SentrySDK.start` running iOS 26 beta in simulator or is this a known issue?

<img width="1233" alt="Image" src="https://github.com/user-attachments/assets/6541da44-4d7e-41e9-a8b0-cff4fb05ad50" />

philippsuess:
@mman, that looks like a different issue. Please open an extra issue for that @mman.

mman:
Thanks, filed as https://github.com/getsentry/sentry-cocoa/issues/5503

PR #5493, opened 2025-06-25
fix: Fix XCFramework version including commit sha on release

itaybre:
Release 8.53.0 job used a cache instead of building from scratch: https://github.com/getsentry/sentry-cocoa/actions/runs/15873193408/job/44760111068
The cache key did not include the version

Fixes: #5492

### DIFF
--- a/.github/workflows/assemble-xcframework-variant.yml
+++ b/.github/workflows/assemble-xcframework-variant.yml
@@ -45,6 +45,12 @@ on:
         type: string
         default: "iphoneos,iphonesimulator,macosx,maccatalyst,appletvos,appletvsimulator,watchos,watchsimulator,xros,xrsimulator"
 
+      release-version:
+        description: |-
+          For release workflows, the version to inject into the SDK.
+        required: false
+        type: string
+
 jobs:
   assemble-xcframework-variant:
     name: Assemble ${{inputs.name}}${{inputs.suffix}} XCFramework Variant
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 8.53.1
+
+### Fixes
+
+- Fix XCFramework version including commit sha on release.
```
```suggestion
## Unreleased

### Fixes

- Fix XCFramework version including commit sha on release. (#5493)
```

.github/workflows/assemble-xcframework-variant.yml
```diff
@@ -70,11 +76,21 @@ jobs:
...
+            echo "VERSION=$(grep MARKETING_VERSION Sources/Configuration/Versioning.xcconfig | cut -d ' ' -f 3)+${{ github.sha }}" >> $GITHUB_ENV
+          fi
+        shell: sh
+
       - name: Compute cache key
         run: |
           sdks_string=${{inputs.sdks}}
           sdks_string_slugified=${sdks_string//,/_}
-          echo "SENTRY_XCFRAMEWORK_CACHE_KEY=${{runner.os}}-xcframework-${{inputs.variant-id}}-$sdks_string_slugified-${{inputs.signed}}-${{hashFiles('Sources/**')}}-${{hashFiles('Sentry.xcodeproj/**')}}" >> $GITHUB_ENV
+          echo "SENTRY_XCFRAMEWORK_CACHE_KEY=${{runner.os}}-xcframework-${{inputs.variant-id}}-$sdks_string_slugified-${{inputs.signed}}-${{env.VERSION}}-${{hashFiles('Sources/**')}}-${{hashFiles('Sentry.xcodeproj/**')}}" >> $GITHUB_ENV
```
noahsmartin, line 93: Why not just include the GitHub.sha directly in the cache key, or the hash of the Versioning.xcconfig file ?

noahsmartin (approved):

.github/workflows/assemble-xcframework-variant.yml
armcknight: if you use the github sha, it'll effectively always be a cache miss for subsequent commits with no actual changes

itaybre: This key is to avoid building the framework from scratch if there are no code changes (se hash to `Sources/**` and `Sentry.xcodeproj/**`).
If we use the sha, it will never hit (unless we rerun a job). We want to be able to use the same cache across PR on simple PRs.

armcknight: versioning.xcconfig is better than sha, but would miss half the time, depending on which branch is taken in the previous step. i don't know exactly what we want there though 🤔

itaybre: Created an issue for this: https://github.com/getsentry/sentry-cocoa/issues/5494
The key can be improved, but this will work for now for the release

PR #5498, opened 2025-06-25
Add more SPI

noahsmartin:
Now that SentryDispatchQueueWrapper a few more of these are unblocked to be public and we can get a bit closer to not requiring the app extension API only flag

#skip-changelog


armcknight:
Just adding this to keep things connected: https://github.com/getsentry/sentry-cocoa/issues/5371

armcknight (approved):

Issue #5422, opened 2025-06-17
Fix flaky UI Tests Critical


philipphofmann:
### Description

The UI tests critical sometimes fail: https://github.com/getsentry/sentry-cocoa/actions/runs/15692735287/job/44211425971

```
[22:10:39]: ▸     ✔ testCrash (59.618 seconds)
[22:10:39]: ▸ Executed 2 tests, with 1 failure (0 unexpected) in 515.488 (515.494) seconds
[22:10:39]: ▸ Test Suite 'SwiftUITestSampleUITests.xctest' failed at 2025-06-16 22:10:39.114.
[22:10:39]: ▸ Executed 2 tests, with 1 failure (0 unexpected) in 515.488 (515.497) seconds
[22:10:39]: ▸ Test Suite 'All tests' failed at 2025-06-16 22:10:39.115.
[22:10:39]: ▸ Executed 2 tests, with 1 failure (0 unexpected) in 515.488 (515.499) seconds
[22:11:00]: ▸ 2025-06-16 22:11:00.651 xcodebuild[6665:36592] [MT] IDETestOperationsObserverDebug: 687.277 elapsed -- Testing started completed.
[22:11:00]: ▸ 2025-06-16 22:11:00.651 xcodebuild[6665:36592] [MT] IDETestOperationsObserverDebug: 0.000 sec, +0.000 sec -- start
[22:11:00]: ▸ 2025-06-16 22:11:00.652 xcodebuild[6665:36592] [MT] IDETestOperationsObserverDebug: 687.277 sec, +687.277 sec -- end
[22:11:01]: ▸ Testing failed:
[22:11:01]: ▸ 	Simulator device failed to launch io.sentry.tests.SwiftUITestSample.
[22:11:01]: ▸ Failing tests:
[22:11:01]: ▸ 	SwiftUITestSampleUITests.testCorruptedEnvelope()
[22:11:01]: ▸ ** TEST FAILED **
[22:11:04]: Exit status: 65
```

We need to identify the root cause and fix it, because if `testCorruptedEnvelope` is actually broken the damage would be fatal. This test validates that the SDK doesn't crash when reading a corrupted envelope.

One idea from @noahsmartin is to let the one for crashing the application run in its own GH action job, because crashing the app can mess up the other UI tests.

philprime:
From Cocoa Sync 06/18:<br><br>A potential fix for this issue could be switching the CI provider

philippsuess:
[...]

PR #5512, opened 2025-06-27
ci: Use xcrun for crash UI test

philipphofmann:
XCTest isn't built for crashing during tests. Instead of using XCTest to press a button and let a test app crash, we now use UserDefaults to tell the test app to crash during launch. We then simply launch the app again and wait to see if it keeps running. This is basically the same as the `testCrash` of the SwiftUITestSample without using XCTests.

The plan is to also replace the `ui_critical_tests_ios_swiftui_envelope` with this approach if this looks acceptable.

I still need to use Xcodegen for the test project.

### DIFF
--- a/.github/workflows/ui-tests-critical.yml
+++ b/.github/workflows/ui-tests-critical.yml
@@ -63,5 +63,4 @@ jobs:
 
       - name: Run SwiftUI Crash Test
         run: |
-          cd TestSamples/SwiftUICrashTest
-          ./test-crash-and-relaunch.sh
+          ./TestSamples/SwiftUICrashTest/test-crash-and-relaunch.sh
--- a/.github/workflows/ui-tests-critical.yml
+++ b/.github/workflows/ui-tests-critical.yml
@@ -58,9 +58,6 @@ jobs:
 
       - run: ./scripts/ci-select-xcode.sh 16.2
 
-      - name: Boot simulator
-        run: ./scripts/ci-boot-simulator.sh "iPhone 16"
-
       - name: Run SwiftUI Crash Test
         run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/ui-tests-critical.yml
```diff
@@ -48,4 +48,18 @@ jobs:
...
+      - uses: actions/checkout@v4
+
+      - run: ./scripts/ci-select-xcode.sh 16.2
+
+      - name: Boot simulator
+        run: ./scripts/ci-boot-simulator.sh
+
+      - name: Run SwiftUI Crash Test
+        run: |
+          ./TestSamples/SwiftUICrashTest/test-crash-and-relaunch.sh
```
itaybre: Should the test take a screenshot of the simulator and upload it to github just in case?

philippsuess: Yes, sure, that's a good idea.

noahsmartin (approved):
I think this looks great, exactly what I had in mind re: https://github.com/getsentry/sentry-cocoa/issues/5422 

If we wanted to have more control of the app we could start a HTTP server in the app being tested and make requests to it from the test host. This is what we do in [FaultOrdering](thttps://github.com/getsentry/FaultOrdering/blob/main/Sources/FaultOrderingTests/FaultOrderingTest.swift#L17) and in general is a pattern we've used a lot at emerge. The idea to use user defaults is great though! So much simpler for something like this. Thanks @philipphofmann !

.github/workflows/ui-tests-critical.yml
philippsuess: Done.

PR #5516, opened 2025-06-27
ref: Use SPI for SentryUserFeedbackIntegrationDriver

noahsmartin:
This makes SentryUserFeedbackIntegrationDriver's init public, another step towards https://github.com/getsentry/sentry-cocoa/issues/5371

This function used SentryScreenshot, which was internal in ObjC so couldn't be added to an SPI function. I converted it to Swift and marked it SPI so it could be used as a parameter to an SPI function

Checked this in the iOS-swift sample app by sending test feedback, the issue with associated screenshot is found here:
https://sentry-sdks.sentry.io/issues/2219152762/events/5b005262adcf449d9d0f26bda2af517d/?project=5428557&referrer=event-or-group-header

#skip-changelog


Sources/Sentry/include/SentryDependencyContainerSwiftHelper.h
```diff
@@ -0,0 +1,26 @@
...
+#if SENTRY_HAS_UIKIT
+#    import <UIKit/UIKit.h>
+#endif // SENTRY_HAS_UIKIT
+
+NS_ASSUME_NONNULL_BEGIN
+
+// Some Swift code needs to access SentryDependencyContainer. To
+// make that possible without requiring all of SentryDependencyContainer
+// to be exposed to Swift this class is exposed to Swift
+// and bridges some functionality from SentryDependencyContainer
```
philippsuess, line 13: Thanks for the comment 💯

Sources/Swift/Tools/SentryScreenshot.swift
```diff
@@ -0,0 +1,106 @@
...
+            let name = index == 0 ? "screenshot.png" : "screenshot-\(index + 1).png"
+            let fileName = (imagesDirectoryPath as NSString).appendingPathComponent(name)
+            try? data.write(to: URL(fileURLWithPath: fileName), options: .atomic)
+        }
+    }
+    
+    public func appScreenshots() -> [UIImage] {
+        let windows = SentryDependencyContainerSwiftHelper.windows() ?? []
+        var result: [UIImage] = []
+        result.reserveCapacity(windows.count)
```
philippsuess, line 65: `l`: I know the code already was like this, but we could do an early return here if windows are `nil` or empty.

Sources/Swift/Tools/SentryScreenshot.swift
```diff
@@ -0,0 +1,106 @@
...
+            // this shouldn't happen now that we discard windows with either 0 height or 0 width,
+            // but still, we shouldn't send any images with either one.
+            if img.size.width > 0 && img.size.height > 0 {
+                result.append(img)
+            }
+        }
+        return result
+    }
+    
+    @discardableResult public func appScreenshotsData() -> [Data] {
```
philippsuess: `m`: Why do we need `@discardableResult`? If we have some invocations just discarding the result, we should remove them IMO, cause why call `appScreenshotsData` when you don't use the data. We can also fix this in a follow-up PR. If only unit tests are impacted, maybe we should change these and remove the `@discardableResult` here so we get a warning if we call this method without using its return value. This comment also applies to `@discardableResult public func appScreenshotsFromMainThread() -> [UIImage] {`

philipphofmann (approved):
LGTM, when doing some manual test runs with the iOS-Swift sample app and validating that screenshots still work, because I don't think we have an e2e test coverage for the screenshots feature. Please post the URLs of the events, including some screenshots, to show that the feature still works.

Sources/Swift/Tools/SentryScreenshot.swift
noahsmartin: It was from unit tests, I updated the test to have `_ = ` so that we can remove the `@discardableResult`

Sources/Swift/Tools/SentryScreenshot.swift
noahsmartin: I don't expect that would make much of a difference since the loop wouldn't be entered at all, but I think it's best left for a future PR so this can stay as much as possible a 1:1 conversion

PR #5519, opened 2025-06-28
test: Fix flaky Tests for SwiftUI

noahsmartin:
This test flaked: https://github.com/getsentry/sentry-cocoa/actions/runs/15934135003/job/44950002764 

And I saw in the xcresult that the simulator crashed and was iOS 18.0. This PR updates it to 18.5 which we use for most other tests and we've seen to be less flaky

Screenshot 2025-06-28 at 6 03 02 PM

#skip-changelog


philipphofmann (approved):
LGTM

PR #5523, opened 2025-06-30
fix: Race condition when closing SDK

noahsmartin:
We have a race condition in the `SentryClient` property of `SentryHub` that is causing flaky unit tests. When the tests are torn down we call `bindClient:nil` (from `[SentrySDK close]`) This sets _client to nil, but the client property can also be accessed by background threads (in particular from `[SentryHub captureReplayEvent]`). This can lead to messaging a dangling pointer. This fix makes the client property atomic, since it is accessed/read from multiple threads.

Fixes https://github.com/getsentry/sentry-cocoa/issues/5506

#skip-changelog

### DIFF
--- a/Tests/SentryTests/SentrySDKThreadTests.swift
+++ b/Tests/SentryTests/SentrySDKThreadTests.swift
@@ -3,31 +3,32 @@ import XCTest
 
 final class SentrySDKThreadTests: XCTestCase {
     func testRaceWhenBindingClient() {
-        for _ in 0..<10_000 {
-            SentrySDK.start(options: .init())
+
+        let options = Options()
+        let sut = SentryHub(client: SentryClient(options: options), andScope: nil)
+
+        for _ in 0..<100 {
+
             let exp = expectation(description: "wait")
-            let warmupExpectation = expectation(description: "warmup")
-            let warmupCount = 100
-            warmupExpectation.expectedFulfillmentCount = warmupCount
+            exp.expectedFulfillmentCount = 100
+
             let queue = DispatchQueue(label: "com.sentry.test_client", qos: .userInteractive, attributes: [.concurrent, .initiallyInactive])
-            for i in 0..<1_000 {
-                exp.expectedFulfillmentCount += 1
+
+            for _ in 0..<100 {
                 queue.async {
-                    SentrySDK.currentHub().capture(event: .init())
+                    sut.bindClient(SentryClient(options: options))
+                    sut.capture(message: "Test message")
                     exp.fulfill()
-                    if i < warmupCount {
-                        warmupExpectation.fulfill()
-                    }
                 }
             }
-            exp.expectedFulfillmentCount -= 1
-            
+
             queue.activate()
-            
-            wait(for: [warmupExpectation])
-            SentrySDK.currentHub().bindClient(nil)
-            
-            wait(for: [exp])
+
+            for _ in 0..<100 {
+                sut.bindClient(nil)
+            }
+
+            wait(for: [exp], timeout: 1.0)
         }
     }
 }
--- a/Tests/SentryTests/SentrySDKThreadTests.swift
+++ b/Tests/SentryTests/SentrySDKThreadTests.swift
@@ -1,0 +1,33 @@
+@testable import Sentry
+import XCTest
+
+final class SentrySDKThreadTests: XCTestCase {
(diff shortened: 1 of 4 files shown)


Tests/SentryTests/SentrySDKThreadTests.swift
```diff
@@ -0,0 +1,33 @@
...
+                }
+            }
+            exp.expectedFulfillmentCount -= 1
+            
+            queue.activate()
+            
+            wait(for: [warmupExpectation])
+            SentrySDK.currentHub().bindClient(nil)
+            
+            wait(for: [exp])
```
philippsuess: `h`: Please always specify a timeout, cause otherwise this test could run for a long time in CI and also slow down our test suite.

Tests/SentryTests/SentrySDKThreadTests.swift
```diff
@@ -0,0 +1,33 @@
+@testable import Sentry
+import XCTest
+
+final class SentrySDKThreadTests: XCTestCase {
+    func testRaceWhenBindingClient() {
+        for _ in 0..<10_000 {
```
philippsuess: `m`: This test requires over a minute or so to run on my local computer. This will significantly increase the test duration. We don't need to use `SentrySDK.start` , which can be a heavy operation with side effects. Instead, we should only interact with the hub, cause that's where the problem lies.

philipphofmann (commented):
Thanks for adding the test, but it takes quite some time to run locally.

Tests/SentryTests/SentrySDKThreadTests.swift
```diff
@@ -0,0 +1,33 @@
+@testable import Sentry
+import XCTest
+
+final class SentrySDKThreadTests: XCTestCase {
+    func testRaceWhenBindingClient() {
```
philippsuess, line 5: So I couldn't stop thinking about this test, and I played a round with it. Iff you change the atomic for the property in the client back to nonatomic this test should fail, which it didn't for me locally. So here's my proposal, which is a bit faster and leads to a crash when the client property is nonatomic. I think it would be great to add a comment explaining the problem it reproduces.
```Swift
func testRaceWhenBindingClient() {

    let options = Options()
    let sut = SentryHub(client: SentryClient(options: options), andScope: nil)

    for _ in 0..<100 {

        let exp = expectation(description: "wait")
        exp.expectedFulfillmentCount = 100

        let queue = DispatchQueue(label: "com.sentry.test_client", qos: .userInteractive, attributes: [.concurrent, .initiallyInactive])

        for _ in 0..<100 {
            queue.async {
                sut.bindClient(SentryClient(options: options))
                sut.capture(message: "Test message")
                exp.fulfill()
            }
        }

        queue.activate()

        for _ in 0..<100 {
            sut.bindClient(nil)
        }

        wait(for: [exp], timeout: 1.0)
    }
}
```

Tests/SentryTests/SentrySDKThreadTests.swift
noahsmartin: Yeah, I brought this up in a slack thread as well. Couldn't find a way to test it any faster. Good point about not using start though, will try out your test below

Tests/SentryTests/SentrySDKThreadTests.swift
noahsmartin: It did fail for me locally when I changed that, but it required so many iterations I imagine it also had to do with my specific hardware setup

noahsmartin: @philipphofmann I just tried this test and it didn't crash the first time I ran it without atomic, but did crash the second time. Hopefully with the changes in this PR it will always succeed and won't be flaky, but that's what I always worry about with a test like this

itaybre (approved):
LGTM

Tests/SentryTests/SentrySDKThreadTests.swift
philippsuess: Yes, testing race conditions is always tricky. Sometimes you need to let it run in a tight loop for minutes so that the race condition occurs. I always try to come up with a repro for race conditions, cause it's so easy to get it wrong. I've messed this up already a couple of times in the past. Thinking that the race condition or a threading issue was fixed, but actually it wasn't. It then can take months or even years for it to surface.

PR #5524, opened 2025-06-30
Remove app extension only flag

noahsmartin:
Removes the `APPLICATION_EXTENSION_API_ONLY` flag which is not required anymore. Fixes https://github.com/getsentry/sentry-cocoa/issues/5371

#skip-changelog


philipphofmann (approved):
I think this deserves a changelog entry. As the SDK won't throw an error anymore if you set `APPLICATION_EXTENSION_API_ONLY_NO` in SentryDefines.h

Apart from that LGTM, thank you @noahsmartin

Samples/iOS-ObjectiveC/iOS-ObjectiveC/AppDelegate.m
```diff
@@ -122,7 +122,6 @@ - (BOOL)application:(UIApplication *)application
                     uiForm.useSentryUser = YES;
                 };
                 config.configureTheme = ^(SentryUserFeedbackThemeConfiguration *_Nonnull theme) {
-                    theme.font = [UIFont fontWithName:@"ChalkboardSE-Regular" size:25];
```
armcknight, line 125: Good catch finding the inconsistency on these, but actually all of these _should_ be `public` and should be tested from the objc sample app. I suppose the compiler wasn't able to help me notice this originally due to the thing you're fixing, but hopefully this doesn't become a blocker for this work. image

armcknight (approved):
Thank you for pushing this work forward, can't wait until this is behind us!

Samples/iOS-ObjectiveC/iOS-ObjectiveC/AppDelegate.m
noahsmartin: Confirmed offline that we don't need to make these public

PR #5525, opened 2025-06-30
Fix prewarmed swiftui tests

noahsmartin:
These tests flaked on me, and I noticed they didn't have our fixes for prewarmed app launches since those were in "SentrySwift" and these tests are for "SentrySwiftUI". I tried to move them to SentrySampleUITestShared, but it looks like that framework can't import XCTest so I just copied it here 

#skip-changelog


itaybre (approved):

philipphofmann (approved):
Thanks

PR #5531, opened 2025-07-01
feat: Record user for watchdog termination events

itaybre:
## :scroll: Description

Changes:
- Record user for watchdog termination events
- Added Base Scope and Processor

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
+++ b/Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
@@ -12,6 +12,7 @@
 #    import <SentryHub.h>
 #    import <SentryNSProcessInfoWrapper.h>
 #    import <SentryOptions+Private.h>
+#    import <SentryPropagationContext.h>
 #    import <SentrySDK+Private.h>
 #    import <SentrySwift.h>
 #    import <SentryWatchdogTerminationBreadcrumbProcessor.h>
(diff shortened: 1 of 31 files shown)


Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
```diff
@@ -96,6 +100,14 @@ - (BOOL)installWithOptions:(SentryOptions *)options
...
         // before installation.
         [scopeObserver setContext:outerScope.contextDictionary];
+        [scopeObserver setUser:outerScope.userObject];
+        [scopeObserver setTags:outerScope.tags];
+        [scopeObserver setLevel:outerScope.levelEnum];
+        [scopeObserver setExtras:outerScope.extraDictionary];
+        [scopeObserver setDist:outerScope.distString];
+        [scopeObserver setFingerprint:outerScope.fingerprintArray];
+        [scopeObserver setEnvironment:outerScope.environmentString];
+        [scopeObserver setTraceContext:[outerScope.propagationContext traceContextForEvent]];
```
philippsuess, line 110: `l`: I guess for now we would only need the user.
```suggestion
        [scopeObserver setUser:outerScope.userObject];
```

Sources/Sentry/include/SentryScope+Private.h
```diff
@@ -20,6 +20,26 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ * This distribution of the application.
+ */
+@property (atomic, copy) NSString *_Nullable distString;
+
+/**
+ * Set the fingerprint of an event to determine the grouping
+ */
+@property (atomic, strong) NSMutableArray<NSString *> *fingerprintArray;
+
```
philippsuess, line 42: `l`: I think we could ditch that for this PR. It's not required for the user.

Sources/Swift/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationUserProcessor.swift
```diff
@@ -0,0 +1,47 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+class SentryWatchdogTerminationUserProcessor: SentryWatchdogTerminationBaseProcessor<User> {
```
philippsuess, line 4: `h`: The biggest problem I see with this approach is that we will have multiple different ScopeProcessor classes for the implementation. So we end up with loads of code doing almost the same thing. The difference between all those persistence stores is only the file they're writing to. They all have to accept some object that can be serialized conforming to the protocol `SentrySerializable` and they must be decodable via the `Decodable` protocol. It would be great if we could find a functional approach with passing in some lambdas or something to avoid all this boilerplate code. We can also get together in a pair programming session and solve this. I'm pretty sure we can solve this and get rid of loads of code and tests.

itaybre:
Closing since this is implemented in #5558 and related PRs

PR #5535, opened 2025-07-01
Fix crash in file manager

noahsmartin:
Fixes a crash when SentryFileManager was null. The initializer for SentryFileManager is declared to return a nullable instance, so SentryScopeContextPersistanceStore should be the same. The crash looked like this:

Screenshot 2025-07-01 at 2 14 19 PM

From what I can tell this started crashing since https://github.com/getsentry/sentry-cocoa/pull/5242 so hasn't been in production for too long

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentrySessionReplayIntegration.m
+++ b/Sources/Sentry/SentrySessionReplayIntegration.m
@@ -428,7 +428,7 @@ static SentryTouchTracker *_touchTracker;
 {
     NSString *sentryPath = [SentryDependencyContainer.sharedInstance.fileManager sentryPath];
     if (!sentryPath) {
-        return NULL;
+        return nil;
     }
     NSURL *dir = [NSURL fileURLWithPath:sentryPath];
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -424,9 +424,12 @@ - (void)startWithOptions:(SentryReplayOptions *)replayOptions
                          options:replayOptions];
 }
 
-- (NSURL *)replayDirectory
+- (nullable NSURL *)replayDirectory
 {
     NSString *sentryPath = [SentryDependencyContainer.sharedInstance.fileManager sentryPath];
+    if (!sentryPath) {
+        return NULL;
```
armcknight: I doubt it matters functionally but I typically only use `NULL` for C/C++/ObjC++ and `nil` for plain ol objc

```suggestion
        return nil;
```

itaybre (approved):

PR #5540, opened 2025-07-01
Fix metrics CI

noahsmartin:
Pulling in the latest changes from this action to unblock CI which is failing on every PR due to a bug in the latest release of "action-app-sdk-overhead-metrics"

#skip-changelog

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -228,7 +228,7 @@ jobs:
           MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
           MATCH_USERNAME: ${{ secrets.MATCH_USERNAME }}
       - name: Collect app metrics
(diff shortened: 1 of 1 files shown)


itaybre (approved):

.github/workflows/release.yml
```diff
@@ -228,7 +228,7 @@ jobs:
           MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
           MATCH_USERNAME: ${{ secrets.MATCH_USERNAME }}
       - name: Collect app metrics
-        uses: getsentry/action-app-sdk-overhead-metrics@v1
+        uses: getsentry/action-app-sdk-overhead-metrics@c9eca50e02d180ee07a02952c062b2f3f545f735
```
philippsuess, line 231: When we use a hash, we usually add a version number or a URL so we know which version this refers to.

PR #5543, opened 2025-07-01
test: Fix simulator selection

noahsmartin:
Fixes the format of device type so fastlane picks up the intended one

#skip-changelog

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -50,7 +50,7 @@ jobs:
       xcode_version: 16.4
       build_with_make: true
       macos_version: macos-15
-      fastlane_command_extra_arguments: device:"iPhone 16 (18.5)"
+      fastlane_command_extra_arguments: device:iPhone 16 (18.5)
 
   ui-tests-swift:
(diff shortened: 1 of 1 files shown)


.github/workflows/ui-tests.yml
```diff
@@ -61,16 +61,16 @@ jobs:
...
-            device: "iPhone 15 (17.5)"
+            device: iPhone 15 (17.5)
           - runs-on: macos-15
             xcode: "16.4"
-            device: "iPhone 16 (18.5)"
+            device: iPhone 16 (18.5)
     with:
       fastlane_command: ui_tests_ios_swift
-      fastlane_command_extra_arguments: device:"${{matrix.device}}"
+      fastlane_command_extra_arguments: device:${{matrix.device}}
```
armcknight, line 73: I'm kinda surprised that it's not something like this, but let's see how CI likes it

```suggestion
      fastlane_command_extra_arguments: "device:${{matrix.device}}"
```

noahsmartin: Confirmed it works as is without the quotes

PR #5544, opened 2025-07-01
test: explicitly set code coverage to true in all test plans

armcknight:
I assume this is the default since it wasn't even present in the main SDK's test plan, but for some reason some of these had it set to false. Change those to true and add the explicit directive to all plans.

#skip-changelog

### DIFF
--- a/Plans/tvOS-Swift_Base.xctestplan
+++ b/Plans/tvOS-Swift_Base.xctestplan
@@ -9,7 +9,7 @@
     }
   ],
   "defaultOptions" : {
-    "codeCoverage" : false,
+    "codeCoverage" : true,
     "targetForVariableExpansion" : {
       "containerPath" : "container:tvOS-Swift.xcodeproj",
       "identifier" : "7BA61D61247FA32600C130A8",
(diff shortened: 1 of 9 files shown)


Plans/tvOS-Swift_Base.xctestplan
```diff
@@ -21,7 +21,7 @@
     {
       "target" : {
         "containerPath" : "container:tvOS-Swift.xcodeproj",
-        "identifier" : "7B64388526A6C71A000D0F65",
+        "identifier" : "1A9FD4374E53BEB8AC621B3D",
```
philprime, line 24: Why did this identifier change?

armcknight: My guess is that when I created the separate test plan for tvOS-SBSwift in https://github.com/getsentry/sentry-cocoa/commit/fb7eac99f7e336ec9c5b9cb74b28ff50b3aa409d, I simply copied it and only changed the name of the project, but not the identifier there, and since it's supposed to be unique, Xcode changed this one.

PR #5546, opened 2025-07-02
meta: fix changelog entry for #5223

armcknight:
I didn't notice that the changelog entry #5223 wasn't scoped down for what the PR wound up actually doing: it only correctly shows the widget when manually calling the display method; automatic injection will be worked on later.


philipphofmann (approved):
Thanks @armcknight, LGTM

PR #5549, opened 2025-07-02
test: Improve corrupted envelope UI test

philipphofmann:
Simplify the UI test by writing the corrupted envelope and then starting the SDK. This removes the need to relaunch the app, which often is problematic in CI. Furthermore, this PR now lets the UI test fail with an error message if the test can't write the corrupted envelope instead of using fatalError. So when this fails, we know clearly what went wrong.

Helps with https://github.com/getsentry/sentry-cocoa/issues/5422.

#skip-changelog


TestSamples/SwiftUITestSample/SwiftUITestSample/SwiftUITestSampleApp.swift
```diff
@@ -25,41 +12,72 @@ struct SwiftUITestSampleApp: App {
...
+    options.dsn = "https://6cc9bae94def43cab8444a99e0031c28@o447951.ingest.sentry.io/5428557"
+    options.debug = true
+    return options
+}()
+
+private func startSDK() {
+    SentrySDK.start(options: sentryOptions)
+}
+
+struct WriteCorruptedEnvelopeError: Error {
```
noahsmartin: `l`: I've never seen a `struct` used as a swift `Error`, usually it's an `enum` and you can conform the enum to `CustomStringConvertible` to get a custom  message. So something like:

```
enum Error: Swift.Error {
  case dsn
  case write(String)
}

extension Error: CustomStringConvertible {
  var description: String {
    switch self {
      case .dsn:
        return "DSN hash is not available"
      case .write(let s):
        return s
   }
  }
}
```

noahsmartin (approved):
Good idea!

TestSamples/SwiftUITestSample/SwiftUITestSample/SwiftUITestSampleApp.swift
philippsuess: Ups, yes, sorry, me neither. Struct doesn't make sense. Enums are the way to go, obviously 🤦

PR #5556, opened 2025-07-02
ci: update from xcode 16.2 to 16.4

armcknight:
I encountered an [error](https://github.com/getsentry/sentry-cocoa/actions/runs/16031579582/job/45233349395?pr=5318#step:8:69) on a check run:
```
xcodebuild: error: Unable to find a device matching the provided destination specifier
```

and thought maybe this is why. generally we should be on the latest version anyways. ideally we could centralize this in something like `//.xcode-version` which we could `cat` out into CI steps, but i'll tackle that separately.

#skip-changelog


armcknight:
I tried a few things in https://github.com/getsentry/sentry-cocoa/pull/5562 to attempt to fix the pod lint errors, but haven't had success yet.

PR #5093, opened 2025-04-15
test: Wait for Sim to boot and print its details

krystofwoldrich:
## :scroll: Description

Adds wait for simulator to boot and print details about the sim, so in case of a failure we can inspect the iOS version useds and other details

#skip-changelog

Example output:
```bash
➜ xcrun simctl list devices --json | jq '.devices | to_entries[] | select(.value[] | .state == "Booted")'
{
  "key": "com.apple.CoreSimulator.SimRuntime.iOS-18-2",
  "value": [
    {
      "lastBootedAt": "2025-04-15T11:03:03Z",
      "dataPath": "/Users/krystofwoldrich/Library/Developer/CoreSimulator/Devices/3BC9A608-0042-414F-B35C-7ADB869BC572/data",
      "dataPathSize": 3387273216,
      "logPath": "/Users/krystofwoldrich/Library/Logs/CoreSimulator/3BC9A608-0042-414F-B35C-7ADB869BC572",
      "udid": "3BC9A608-0042-414F-B35C-7ADB869BC572",
      "isAvailable": true,
      "logPathSize": 589824,
      "deviceTypeIdentifier": "com.apple.CoreSimulator.SimDeviceType.iPhone-16",
      "state": "Booted",
      "name": "iPhone 16"
    }
  ]
}
```

PR #5557, opened 2025-07-02
ci: remove simulator boot script

armcknight:
I noticed while preparing https://github.com/getsentry/sentry-cocoa/pull/5556 that this script is only called for a single UI test. It was originally introduced in https://github.com/getsentry/sentry-cocoa/pull/5031 and then mostly removed in https://github.com/getsentry/sentry-cocoa/pull/5093 (even though logic was added to the script, maybe this was some kind of mistake) and also one more instance was removed in https://github.com/getsentry/sentry-cocoa/pull/5398/files. 

Unless all the other UI tests have gotten worse since it was removed from them, this indicates to me that it isn't really necessary any longer.

#skip-changelog


armcknight:
Hmm, appears it actually is still needed. the ci workflow failed due to no booted devices: https://github.com/getsentry/sentry-cocoa/actions/runs/16033550566/job/45239750902#step:6:539

PR #5558, opened 2025-07-02
feat: Update Persistent store and processor to handle multiple data types

itaybre:
Update Persistent store and processor to handle multiple data types


Sources/Swift/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationAttributesProcessor.swift
```diff
@@ -0,0 +1,55 @@
...
+    // MARK: - Private
+    private func setData<T>(data: T?, field: SentryScopeField, save: @escaping (T) -> Void) {
+        SentrySDKLog.debug("Setting \(field.name) in background queue: \(String(describing: data))")
+        dispatchQueueWrapper.dispatchAsync { [weak self] in
+            guard let strongSelf = self else {
+                SentrySDKLog.debug("Can not set \(field.name), reason: reference to processor is nil")
+                return
+            }
+            guard let data = data else {
+                SentrySDKLog.debug("Data is nil, deleting active file.")
```
philippsuess: `l`: The log message could also mention the `field.name`.

Sources/Swift/Persistence/SentryScopePersistentStore.swift
```diff
@@ -0,0 +1,222 @@
...
+        }
+        
+        return deserialized as? [String: [String: Any]]
+    }
+}
+
+// MARK: - User
+extension SentryScopePersistentStore {
+    private func encode(user: User) -> Data? {
+        guard let sanitizedUser = sentry_sanitize(user.serialize()) else {
```
philippsuess: `m`: You shouldn't have to call `sentry_sanitize`, which is a bit expensive, cause it iterates recursively over the serialized JSON dictionary. `SentryUser.serialize` already calls sanitize for the properties it requires sanitize https://github.com/getsentry/sentry-cocoa/blob/ea5a59bc570a120182bd692e25f2216db020f858/Sources/Sentry/SentryUser.m#L101

Sources/Swift/Persistence/SentryScopePersistentStore.swift
```diff
@@ -0,0 +1,222 @@
...
+        guard let data = SentrySerialization.data(withJSONObject: sanitizedUser) else {
+            SentrySDKLog.error("Failed to serialize user, reason: user is not valid json: \(user)")
+            return nil
+        }
+        return data
+    }
+    
+    private func decodeUser(from data: Data) -> User? {
+        do {
+            let decoder = JSONDecoder()
```
philippsuess: `h`: Please use `decodeFromJSON` cause it has a custom `dateDecodingStrategy`.

https://github.com/getsentry/sentry-cocoa/blob/ea5a59bc570a120182bd692e25f2216db020f858/Sources/Swift/Protocol/Codable/SentryCodable.swift#L10-L47

Tests/SentryTests/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationAttributesProcessorTests.swift
```diff
@@ -0,0 +1,368 @@
...
+           user.username = "testuser"
+           user.name = "Test User"
+           user.ipAddress = "192.168.1.1"
+           user.data = ["custom_key": "custom_value"]
+           return user
+       }()
+       
+       let invalidUser: User = {
+           let user = User(userId: "test-user-id")
+           // Create an invalid user by setting data with non-serializable content
```
philippsuess, line 41: Thanks for the comment 🥇

Tests/SentryTests/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationAttributesProcessorTests.swift
```diff
@@ -0,0 +1,368 @@
...
+
+   func testSetUser_whenUserIsValid_shouldDispatchToQueue() {
+       // -- Act --
+       sut.setUser(fixture.user)
+
+       // -- Assert --
+       XCTAssertEqual(fixture.dispatchQueueWrapper.dispatchAsyncInvocations.count, 1)
+   }
+
+   func testSetUser_whenProcessorIsDeallocatedWhileDispatching_shouldNotCauseRetainCycle() {
```
philippsuess, line 198: `l`: I'm unsure if it makes sense to have two tests for this use case. We already test this with `testSetContext_whenProcessorIsDeallocatedWhileDispatching_shouldNotCauseRetainCycle`. It does no harm, but it's also code. Up to you. If you want to keep it, keep it.

philipphofmann (commented):
Almost, LGTM. @itaybre can you please also test this with a build from TestFlight or on a real device and double-check if the user now appears in watchdog termination events. The iOS-Swift sample app has a button called OOM crash, that you can use.

Sources/Swift/Persistence/SentryScopePersistentStore.swift
```diff
@@ -0,0 +1,222 @@
...
+        switch self {
+        case .context:
+            return "context"
+        case .user:
+            return "user"
+        }
+    }
+}
+
+@_spi(Private) @objc public protocol SentryFileManagerProtocol {
```
philprime: `m`: Please move this protocol to a different file, so if we reuse it for other classes unrelated to the `SentryScopePersistentStore` it's not referencing this file (modular file structure)

Sources/Swift/Persistence/SentryScopePersistentStore.swift
```diff
@@ -0,0 +1,222 @@
...
+    func getSentryPathAsURL() -> URL
+}
+
+@objc
+@_spi(Private) public class SentryScopePersistentStore: NSObject {
+    private let fileManager: SentryFileManagerProtocol
+    
+    @objc
+    public init?(fileManager: SentryFileManagerProtocol?) {
+        guard let fileManager else { return nil }
```
philprime, line 24: `h`: Why should we handle a null file manager here? I strongly believe that the initializer should be non-nullable and file manager also non-nullable, and whichever class uses the `SentryScopePersistentStore` is in charge of handling null.

Sources/Swift/Persistence/SentryScopePersistentStore.swift
```diff
@@ -0,0 +1,222 @@
...
+    func moveState(_ stateFilePath: String, toPreviousState previousStateFilePath: String)
+    func readData(fromPath path: String) throws -> Data
+    @objc(writeData:toPath:)
+    @discardableResult func write(_ data: Data, toPath path: String) -> Bool
+    func removeFile(atPath path: String)
+    func getSentryPathAsURL() -> URL
+}
+
+@objc
+@_spi(Private) public class SentryScopePersistentStore: NSObject {
```
philprime, line 19: `l`: To be honest, this file is already growing massively. Having multiple store types as subclasses or wrappers of the scope persistent store could be a way of sharing the generic persistence logic, while having simpler files/types.

No strong opinion here, so proceed as you believe best.

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -1083,6 +1100,25 @@ private extension SentrySDKTests {
         testProcessInfoWrapper.overrides.environment = ["XCODE_RUNNING_FOR_PREVIEWS": "1"]
         SentryDependencyContainer.sharedInstance().processInfoWrapper = testProcessInfoWrapper
     }
+    
+#if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
+    func createTestModels() throws -> (Options, TestSentryDispatchQueueWrapper, TestSentryScopePersistentStore, SentryWatchdogTerminationScopeObserver) {
```
philprime: `l`: This could also be a fixture class like we do in other existing tests.

Sources/Sentry/SentryWatchdogTerminationScopeObserver.m
```diff
@@ -49,7 +49,7 @@ - (void)clearBreadcrumbs
 
 - (void)setContext:(nullable NSDictionary<NSString *, id> *)context
 {
-    [self.contextProcessor setContext:context];
+    [self.attributesProcessor setContext:context];
```
philprime, line 52: `m`: Summarizing all processor into a single `attributesProcessor` kind of defeats the purpose of the processor classes spreading logic into multiple types. It's fine to stay this way.

philprime (commented):
Good progress! Left a couple of comments.

Sources/Swift/Persistence/SentryScopePersistentStore.swift
itaybre: While this is true, if we split this file into multiple subclasses and keep all of them in memory in the dependency container. Increasing the memory footprint + creating many initializations in the container.

itaybre: This is a trade off, there is no clear solution either, but I think this approach is easier to maintain.

Sources/Sentry/SentryWatchdogTerminationScopeObserver.m
itaybre: Technically not all of them, I left the breadcrumbs processor on a different class because it did have different logic. As said above, there is no clear better solution. But I think this solution will be easier to maintain later + less boilerplate for DI + less memory footprint.

Sources/Swift/Persistence/SentryScopePersistentStore.swift
itaybre: Since the initializer is in ObjC there is no way to guarantee the nullability.
The initialization might fail:
```
SENTRY_LAZY_INIT(_fileManager, ({
        NSError *error;
        SentryFileManager *manager = [[SentryFileManager alloc] initWithOptions:SentrySDK.options
                                                                          error:&error];
        if (manager == nil) {
            SENTRY_LOG_DEBUG(@"Could not create file manager - %@", error);
        }
        manager;
    }));
```

itaybre: Related to this PR: https://github.com/getsentry/sentry-cocoa/pull/5535

Sources/Swift/Persistence/SentryScopePersistentStore.swift
itaybre: Fair comment, moving it

itaybre:
> Almost, LGTM. @itaybre can you please also test this with a build from TestFlight or on a real device and double-check if the user now appears in watchdog termination events. The iOS-Swift sample app has a button called OOM crash, that you can use.

Confirmed working
image

Sources/Swift/Persistence/SentryScopePersistentStore.swift
philippsuess: We can move the extension classes to an extra file if it gets too big. For me, this is still OK. I prefer having it all in one place instead of having to jump around 10 different files and having 10 extra classes for simple logic. This all belongs together, so it's OK for me.

Sources/Swift/Persistence/SentryScopePersistentStore.swift
philippsuess: Thanks for bringing this up @philprime. The root cause of this is how we handle the `scopePersistentStore` dependency. We use the SentryDependencyContainer first in the SDK.m
https://github.com/getsentry/sentry-cocoa/blob/2ef82831b491f8b52459bf08a0b5b1c404d2220b/Sources/Sentry/SentrySDK.m#L243-L244

And again here 

https://github.com/getsentry/sentry-cocoa/blob/2ef82831b491f8b52459bf08a0b5b1c404d2220b/Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m#L69-L79

Ideally, if the fileManager is nil, we shouldn't even create a `SentryWatchdogTerminationTracker` in the `SentryWatchdogTerminationTrackingIntegration`, cause it won't work anyway. I think we don't have to fix this in this PR, as the problem already existed before. But I think we should follow up on this in an extra PR.

Sources/Swift/Persistence/SentryScopePersistentStore.swift
```diff
@@ -199,24 +190,19 @@ extension SentryScopePersistentStore {
...
-            return try decoder.decode(UserDecodable.self, from: data)
-        } catch {
-            SentrySDKLog.error("Failed to decode user, reason: \(error)")
-            return nil
-        }
+        return decoderAux(data)
+    }
+    
+    // Swift compiler can't infer T, even if I try to cast it
+    private func decoderAux(_ data: Data) -> UserDecodable? {
```
philippsuess: `m`: I'm a bit confused by the name `decoderAux`. What does it stand for?

philipphofmann (approved):
LGTM, but maybe it makes sense to also wait for the approval of @philprime, cause he came up with the initial new logic for the scope observer.

Sources/Swift/Persistence/SentryScopePersistentStore.swift
philprime: That was exactly my point, there is no reason why there should be a case the `SentryWatchdogTerminationTracker` can be created without a file manager. 

I am actually fixing the file manager handling with the nullability PR too, but I still propose we move the `nil` handling to the one calling the `SentryScopePersistentStore`.
https://github.com/getsentry/sentry-cocoa/pull/5572/files#diff-4c0fe64e40914fced7bf84b781ab1e553e128baf8ee1cf352e867f587b90a87e

philprime: To clarify, I believe the dependency manager should check if the file manager is nil, and if it is, it should not return the persistent store. This will require changing the accessor methods/types in the SentryDependencyContainer to be nullable, but this is necessary in any case, because the current approach could also cause the persistent store to be nil.

If the accessor `scopePersistentStore` is `null`, that must be handled in the termination tacking integration. We must not just swallow the potentially null case with bad nullability handling and instead be explicit.

philippsuess: All good points @philprime. Let's discuss this in a new issue or another PR, because these problems existed before. I don't think these problems should block this PR. Or do you think they should?

Sources/Swift/Persistence/SentryScopePersistentStore.swift
itaybre: `Auxiliary`, can be renamed to `Helper` if that is more clear

philippsuess: Ah OK, thanks. `decodeUserToWorkWithSwiftCompiler`, `decodeUserInternal`, `decodeUserHelper`, `decodeUserAuxiliary` all would work for me. `decoderAux` is a bit confusing. The comment above states it anyway. Pick whatever you prefer. I have no strong opinion.

Sources/Swift/Persistence/SentryScopePersistentStore.swift
philprime: While I dislike the workaround, I won't block this any further.

itaybre: Created an issue [here](https://github.com/getsentry/sentry-cocoa/issues/5576) to fix this

PR #5561, opened 2025-07-02
feat: Add support for tags and trace context fields for termination watch

itaybre:
feat: Add support for tags and context fields for termination watch


Sources/Sentry/SentryWatchdogTerminationTracker.m
```diff
@@ -60,6 +60,8 @@ - (void)start
             event.user = [self.scopePersistentStore readPreviousUserFromDisk];
             event.dist = [self.scopePersistentStore readPreviousDistFromDisk];
             event.environment = [self.scopePersistentStore readPreviousEnvironmentFromDisk];
+            event.tags = [self.scopePersistentStore readPreviousTagsFromDisk];
+            // TODO: See where to store trace context
```
philprime: @philipphofmann please correct me, but the trace context isn't relevant for watchdog termination events, right?

Sources/Swift/Persistence/SentryScopePersistentStore.swift
```diff
@@ -261,3 +291,75 @@ extension SentryScopePersistentStore {
         return String(data: data, encoding: .utf8)
     }
 }
+
+// MARK: - Tags
+extension SentryScopePersistentStore {
```
philprime: `m`: As mentioned in the other related PRs, I would propose we move these extensions to their own file.

Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
```diff
@@ -99,6 +100,8 @@ - (BOOL)installWithOptions:(SentryOptions *)options
         [scopeObserver setUser:outerScope.userObject];
         [scopeObserver setEnvironment:outerScope.environmentString];
         [scopeObserver setDist:outerScope.distString];
+        [scopeObserver setTags:outerScope.tags];
+        [scopeObserver setTraceContext:[outerScope.propagationContext traceContextForEvent]];
```
philprime: `m`: I don't think we need to persist the trace context here, but that could be wrong. cc @philipphofmann

Sources/Sentry/SentryWatchdogTerminationTracker.m
philippsuess: It would be, but the trace most likely will be lost, because the traceID is from the ongoing trace, which is in memory. When a watchdog termination happens, the SDK can't send the ongoing trace. So I think we can ignore it for now.

Sources/Sentry/SentryWatchdogTerminationTrackingIntegration.m
philippsuess: Same as https://github.com/getsentry/sentry-cocoa/pull/5561#discussion_r2189526990, so it's better to ditch this for now.

itaybre:
Confirmed working
image

Sources/Swift/Integrations/WatchdogTerminations/Processors/SentryWatchdogTerminationAttributesProcessor.swift
```diff
@@ -48,6 +48,18 @@ import Foundation
         }
     }
     
+    public func setTags(_ tags: [String: String]?) {
+        setData(data: tags, field: .tags) { [weak self] data in
+            self?.scopePersistentStore.writeTagsToDisk(tags: data)
+        }
+    }
+    
+    public func setTraceContext(_ traceContext: [String: Any]?) {
```
philprime: `m`: As we skip the trace context, is there any remaining value in this method?

Sources/Swift/Persistence/SentryScopePersistentStore+TraceContext.swift
```diff
@@ -0,0 +1,23 @@
+@_implementationOnly import _SentryPrivate
+
+extension SentryScopePersistentStore {
+    func encode(traceContext: [String: Any]) -> Data? {
```
philprime: `m`: As we are not persisting the trace context, do we still need this?

Sources/Swift/Persistence/SentryScopePersistentStore.swift
```diff
@@ -6,6 +6,8 @@ enum SentryScopeField: UInt, CaseIterable {
     case user
     case dist
     case environment
+    case traceContext
```
philprime: `m`: As we are not persisting the trace context, do we still need this?

Sources/Swift/Persistence/SentryScopePersistentStore.swift
```diff
@@ -112,6 +118,30 @@ enum SentryScopeField: UInt, CaseIterable {
...
+        }
+    }
+    
+    func writeTagsToDisk(tags: [String: String]) {
+        writeFieldToDisk(field: .tags, data: encode(tags: tags))
+    }
+    
+    // MARK: - Trace Context
+    @objc
+    public func readPreviousTraceContextFromDisk() -> [String: Any]? {
```
philprime: `m`: As we are not persisting the trace context, do we still need this?

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -1085,6 +1085,63 @@ class SentrySDKTests: XCTestCase {
...
+        // -- Arrange --
+        fixture.observer.setTags(["tag1": "value1", "tag2": "value2"])
+
+        // Wait for the observer to complete
+        let expectation = XCTestExpectation(description: "setTags completes")
+        fixture.dispatchQueueWrapper.dispatchAsync {
+            // Dispatching a block on the same queue will be run after the context processor.
+            expectation.fulfill()
+        }
+        wait(for: [expectation], timeout: 1.0)
```
philprime, line 1099: `h`: Shouldn't this wait condition be after the calling of `deleteAllPreviousState()`?This just waits for the `dispatchQueueWrapper` to be empty, and not actually waiting for moving to complete

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -1085,6 +1085,63 @@ class SentrySDKTests: XCTestCase {
...
+        // -- Arrange --
+        fixture.observer.setTraceContext(["trace_id": "1234567890", "span_id": "abcdef", "sampled": "true"])
+
+        // Wait for the observer to complete
+        let expectation = XCTestExpectation(description: "setTraceContext completes")
+        fixture.dispatchQueueWrapper.dispatchAsync {
+            // Dispatching a block on the same queue will be run after the context processor.
+            expectation.fulfill()
+        }
+        wait(for: [expectation], timeout: 1.0)
```
philprime: `h`: See comment above

Sources/Swift/Persistence/SentryScopePersistentStore+Tags.swift
```diff
@@ -0,0 +1,49 @@
+@_implementationOnly import _SentryPrivate
+
+extension SentryScopePersistentStore {
+    func encode(tags: [String: String]) -> Data? {
+        // We need to check if the Tags is a valid JSON object before encoding it.
+        // Otherwise it will throw an unhandled `NSInvalidArgumentException` exception.
+        // The error handler is required due but seems not to be executed.
+        guard let sanitizedTags = sentry_sanitize(tags) else {
```
philprime: `m`: Do we have a test case with an invalid tag value which requires the sanitization?

philprime (commented):
Good progress, almost LGTM.
Please take a look at the code coverage reported by Codecov.
You are missing test cases for the sanitization. Furthermore, please double-check that the expectation+wait in the tests actually make sense. Left comments for the remainders of the trace context.

Tests/SentryTests/SentrySDKTests.swift
itaybre: This ensures setTags completed its execution.
The moving of files does not use the queue, and is executed synchronously, so no need to wait there.

Tests/SentryTests/SentrySDKTests.swift
itaybre: See above

Sources/Swift/Persistence/SentryScopePersistentStore+Tags.swift
itaybre: Seems like I misread SentryEvent.m, only extras need sanitization here, not tags, fixing it

Sources/Sentry/SentryWatchdogTerminationScopeObserver.m
```diff
@@ -89,7 +89,7 @@ - (void)setUser:(nullable SentryUser *)user
 
 - (void)setTraceContext:(nullable NSDictionary<NSString *, id> *)traceContext
 {
-    // Left blank on purpose
+    // Nothing to do here, Trace Context is not persisted for watchdog termination events
```
noahsmartin, line 92: It might be nice to explain why, like in @philipphofmann's comment below that we can't persist it because we aren't given time when watchdog kills the app

Sources/Swift/Persistence/SentryScopePersistentStore+Context.swift
```diff
@@ -0,0 +1,48 @@
+@_implementationOnly import _SentryPrivate
+
+extension SentryScopePersistentStore {
+    func encode(context: [String: [String: Any]]) -> Data? {
+        // We need to check if the context is a valid JSON object before encoding it.
+        // Otherwise it will throw an unhandled `NSInvalidArgumentException` exception.
+        // The error handler is required due but seems not to be executed.
```
noahsmartin: typo?

Sources/Swift/Persistence/SentryScopePersistentStore+Tags.swift
```diff
@@ -0,0 +1,42 @@
+@_implementationOnly import _SentryPrivate
+
+extension SentryScopePersistentStore {
```
noahsmartin, line 3: Can you dedup both the methods in this extension with the `+Context` extension by using two generic functions instead? The only difference seems to be the type of the values of the dictionary, which I think you can just make a generic type

Sources/Swift/Persistence/SentryScopePersistentStore+User.swift
```diff
@@ -0,0 +1,20 @@
...
+            return nil
+        }
+        return data
+    }
+    
+    func decodeUser(from data: Data) -> User? {
+        return decoderUserHelper(data)
+    }
+    
+    // Swift compiler can't infer T, even if I try to cast it
```
noahsmartin, line 16: What do you mean T? This function isn't generic

noahsmartin: I see this is a copy/paste, feel free to ignore

itaybre: This is using a generic function, without the helper, the compiler can't infer it

Sources/Swift/Persistence/SentryScopePersistentStore+Context.swift
itaybre: Improved the message

Issue #5421, opened 2025-06-17
Fix or disable flaky SentryHTTPTransport flush tests


philipphofmann:
### Description

The SentryHTTPTransport tests for the flush implementation sometimes flake as here https://github.com/getsentry/sentry-cocoa/actions/runs/15700976571/job/44235777418

```
testFlush_CallingFlushDirectlyAfterCapture_Flushes, XCTAssertEqual failed: ("3") is not equal to ("0")

testFlush_CallingFlushDirectlyAfterCapture_Flushes, XCTAssertEqual failed: ("SentryFlushResult(rawValue: 2)") is not equal to ("SentryFlushResult(rawValue: 0)") - Flush should not time out.

testFlush_CallingFlushDirectlyAfterCapture_Flushes, XCTAssertEqual failed: ("2") is not equal to ("0")

testFlush_CallingFlushDirectlyAfterCapture_Flushes, XCTAssertEqual failed: ("SentryFlushResult(rawValue: 2)") is not equal to ("SentryFlushResult(rawValue: 0)") - Flush should not time out.

testFlush_CallingFlushDirectlyAfterCapture_Flushes, XCTAssertEqual failed: ("1") is not equal to ("0")

testFlush_CallingFlushDirectlyAfterCapture_Flushes, XCTAssertEqual failed: ("SentryFlushResult(rawValue: 1)") is not equal to ("SentryFlushResult(rawValue: 0)") - Flush should not time out.
```

Ideally, we should fix the flaky tests and not disable them, because they test the flush functionality. It's worth mentioning that the ANR tests were also flaky due to the TestCurrentDateProvider not being thread-safe. After we fixed this problem with https://github.com/getsentry/sentry-cocoa/pull/5299 the tests are more stable now. So when investigating this problem keep in mind that the test code could actually have bugs or maybe the flush implementation is buggy.

philprime:
From Cocoa Sync 06/18:<br><br>We need to look into the proposed changes of [https://github.com/getsentry/sentry-cocoa/pull/5220](https://github.com/getsentry/sentry-cocoa/pull/5220) again, but @philipphofmann believes that a test plan will not help and instead we need to evaluate the tests themselves again.

PR #5565, opened 2025-07-03
fix: Wrong flush timeout when called sequentially

philipphofmann:
## :scroll: Description

Fix an edge case leading to flush continuing to time out when it timed out once before the HTTP transport finished sending.

Came up while investigating https://github.com/getsentry/sentry-cocoa/issues/5421

Unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryHttpTransport.m
+++ b/Sources/Sentry/SentryHttpTransport.m
@@ -227,7 +227,7 @@
     // If nobody called `stopFlushing`, we must call `dispatch_group_leave` here, because
     // otherwise the dispatch group never leaves and the next time somebody calls flush,
     // flush will always time out.
-    [self stopFlusing];
+    [self stopFlushing];
 
     if (result == 0) {
         SENTRY_LOG_DEBUG(@"Finished flushing.");
@@ -238,7 +238,7 @@
     }
 }
 
-- (void)stopFlusing
+- (void)stopFlushing
 {
     @synchronized(self) {
         if (_isFlushing) {
@@ -434,7 +434,7 @@
 {
     @synchronized(self) {
         self.isSending = NO;
-        [self stopFlusing];
+        [self stopFlushing];
     }
 }
 


philippsuess:
Converting back to draft because the unit tests are failing.

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -223,6 +223,12 @@ - (SentryFlushResult)flush:(NSTimeInterval)timeout
 
     intptr_t result = dispatch_group_wait(self.dispatchGroup, dispatchTimeout);
 
+    // Every `dispatch_group_enter` must have a corresponding `dispatch_group_leave`.
+    // If nobody called `stopFlushing`, we must call `dispatch_group_leave` here, because
+    // otherwise the dispatch group never leaves and the next time somebody calls flush,
+    // flush will always time out.
+    [self stopFlusing];
```
armcknight:
```suggestion
    [self stopFlushing];
```

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -417,11 +434,7 @@ - (void)finishedSending
 {
     @synchronized(self) {
         self.isSending = NO;
```
armcknight, line 436: since `stopFlushing` also sets `_isFlushing = NO`, does it make sense to do it here as well? If we want it to be set before the `@synchronized` barrier, maybe we also set it before that in `stopFlushing` on line 243?

armcknight (approved):
LGTM with small feedback items.

Sources/Sentry/SentryHttpTransport.m
philippsuess: OMG, embarrassing 🤦. Thanks 

Somehow, I disabled spell checking
Screenshot 2025-07-16 at 09 50 40

Sources/Sentry/SentryHttpTransport.m
philippsuess: I'm unsure what you exactly mean. We don't want it to be set before the `@synchronized` barrier, but we want to update both `isSending` and `isFlushing` in the same synchronized block, which we already do here. We call `@synchronized` twice here, which is OK because `@synchronized` is reentrant. Am I missing something?

armcknight: I think I read `isSending` here and my brain interpreted it as `isFlushing` 🤦🏻

philippsuess: No worries, mixing things up happens to me too sometimes 😄

PR #5566, opened 2025-07-03
ref: Don't use functions for notification names

noahsmartin:
While looking at moving NotificationCenterWrapper to swift I noticed this separate cleanup opportunity. This moves the functions that bridged iOS/Mac notifications into a typedef, which is less lines of code (making it safer by reducing the chance for typos) and also more performant (less app size, fewer objc method dispatch)

It works even better in Swift (currently only used by tests) where you can just define a single typedef and get all the notifications bridged without having to write another line for each notification name

The main change is in `SentryNSNotificationCenterWrapper.h` the rest of the changes are just updating the callsites.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryAppStateManager.m
+++ b/Sources/Sentry/SentryAppStateManager.m
@@ -50,24 +50,21 @@
 - (void)start
 {
     if (self.startCount == 0) {
-        [self.notificationCenterWrapper
-            addObserver:self
-               selector:@selector(didBecomeActive)
-                   name:SentryNSNotificationCenterWrapper.didBecomeActiveNotificationName];
+        [self.notificationCenterWrapper addObserver:self
+                                           selector:@selector(didBecomeActive)
+                                               name:SentryDidBecomeActiveNotification];
 
         [self.notificationCenterWrapper addObserver:self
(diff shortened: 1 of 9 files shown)


Sources/Sentry/SentryAppStateManager.m
```diff
@@ -50,24 +50,21 @@ - (instancetype)initWithOptions:(SentryOptions *)options
 - (void)start
 {
     if (self.startCount == 0) {
-        [self.notificationCenterWrapper
-            addObserver:self
-               selector:@selector(didBecomeActive)
-                   name:SentryNSNotificationCenterWrapper.didBecomeActiveNotificationName];
+        [self.notificationCenterWrapper addObserver:self
+                                           selector:@selector(didBecomeActive)
+                                               name:SentryDidBecomeActiveNotification];
```
itaybre, line 55: This was the linter, right?

itaybre (approved):

Sources/Sentry/SentryAppStateManager.m
noahsmartin: yeah it moved the lines around

philippsuess:
Way better thanks @noahsmartin.

PR #5567, opened 2025-07-03
meta: Add objc to swift conversion analyzer

noahsmartin:
Adding the objc to swift conversion scripts to CI runs of main so it's easy to access the latest result

#skip-changelog

### DIFF
--- a/.github/workflows/objc-conversion-analysis.yml
+++ b/.github/workflows/objc-conversion-analysis.yml
@@ -22,9 +22,9 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-        
+
       - name: Setup Ruby
-        uses: ruby/setup-ruby@a4effe49ee8ee5b8b5091268c473a4628afb5651
+        uses: ruby/setup-ruby@a4effe49ee8ee5b8b5091268c473a4628afb5651 # pin@v5.4.3
         with:
           ruby-version: "3.2"
           bundler-cache: true
@@ -45,7 +45,7 @@ jobs:
       - name: Convert DOT to PNG
         working-directory: SwiftConversion
         run: |
-          dot -Tpng objc_dependencies_topo.dot -o objc_dependencies_topo.png
+          dot -Tsvg objc_dependencies_topo.dot -o objc_dependencies_topo.svg
 
       - name: Upload analysis artifacts
         uses: actions/upload-artifact@v4
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,6 +98,10 @@ To make a header public follow these steps:
 - Add it to the Umbrella Header [Sentry.h](/Sources/Sentry/Public/Sentry.h).
 - Set the target membership to public.
 
+## Converting to Swift
+
+Converting internal ObjC classes to Swift needs to be done following a partial order. The scripts in `SwiftConversion` generate this ordering and can be used to guide ObjC to Swift conversions.
+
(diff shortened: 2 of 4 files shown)


.github/workflows/objc-conversion-analysis.yml
```diff
@@ -0,0 +1,58 @@
...
+    name: Analyze Objective-C to Swift Conversion
+    runs-on: macos-15
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@a4effe49ee8ee5b8b5091268c473a4628afb5651
```
philippsuess: `l`: Please add the setup-ruby version this points to as a comment as we do for example here https://github.com/getsentry/sentry-cocoa/blob/ea5a59bc570a120182bd692e25f2216db020f858/.github/workflows/test.yml#L256

SwiftConversion/objc_conversion_analyzer.rb
```diff
@@ -0,0 +1,211 @@
+#!/usr/bin/env ruby
+
+require 'set'
+require 'json'
+
+class ObjCConversionAnalyzer
+  def initialize(include_dir = '../Sources/Sentry/include')
+    @include_dir = include_dir
```
philippsuess, line 8: `m`: What about other folders such as `/Sources/Sentry/Public/` or `Sources/SentryCrash/`? We can also do these in follow up PRs.

.github/workflows/objc-conversion-analysis.yml
```diff
@@ -0,0 +1,58 @@
...
+          dot -Tpng objc_dependencies_topo.dot -o objc_dependencies_topo.png
+
+      - name: Upload analysis artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: objc-conversion-analysis
+          path: |
+            SwiftConversion/objc_conversion_analysis.json
+            SwiftConversion/objc_dependencies_topo.dot
+            SwiftConversion/objc_dependencies_topo.png
```
philippsuess: `m`: Does .svg work? If yes that would be great, cause the png resolution is annoying once you zoom in.
```suggestion
          dot -Tpng objc_dependencies_topo.dot -o objc_dependencies_topo.png
          dot -Tpng objc_dependencies_topo.dot -o objc_dependencies_topo.svg

      - name: Upload analysis artifacts
        uses: actions/upload-artifact@v4
        with:
          name: objc-conversion-analysis
          path: |
            SwiftConversion/objc_conversion_analysis.json
            SwiftConversion/objc_dependencies_topo.dot
            SwiftConversion/objc_dependencies_topo.png
            objc_dependencies_topo.svg
```

philprime:
Do we have an example to view the output?

SwiftConversion/objc_conversion_analyzer.rb
noahsmartin: This is specifically looking at conversion for internal types, not public. That's because internal types sometimes have blockers on other pieces that need to be converted. I'll make it more clear in a readme

noahsmartin:
@philprime there's an example that I posted in slack a few days ago, but you can also take a look at the one in the status check on this PR

.github/workflows/objc-conversion-analysis.yml
noahsmartin: yep updated

PR #5572, opened 2025-07-04
refactor: Enable strict nullability for Objective-C

### LINKED ISSUE #5577: Enforce stricter compiler rules for Objective-C
### Description

Objective-C can be configured to be stricter at compile-time, therefore surfacing potential runtime issues early.

Potential `OTHER_CFLAGS` candidates to explore:

### Nullability & Typing

| Flag | Description |
|------|-------------|
| `-Wnullability-completeness` | Warns if `_Nonnull` or `_Nullable` annotations are missing. |
| `-Wnullable-to-nonnull-conversion` | Warns on unsafe conversion from nullable to nonnull. |
| `-Wobjc-generics` | Warns when using collection types without generic type parameters. |
| `-Wstrict-prototypes` | Warns on functions without complete prototype declarations. |
| `-Wstrict-selector-match` | Warns if a selector passed to `performSelector:` doesn't exactly match method signature. |

---

### ARC & Memory Safety

| Flag | Description |
|------|-------------|
| `-fobjc-arc` | Enables Automatic Reference Counting. |
| `-fobjc-arc-unsafe-retained` | Warns on unsafe `__unsafe_unretained` usage. |
| `-Wimplicit-retain-self` | Warns if `self` is retained inside a block without explicit capture. |
| `-Wdirect-ivar-access` | Warns when accessing ivars directly instead of using properties. |
| `-Wdeprecated-objc-isa-usage` | Warns about direct access to the `isa` pointer. |

---

### Best Practices & API Safety

| Flag | Description |
|------|-------------|
| `-Wunguarded-availability` | Warns when using APIs that might not be available on all platforms. |
| `-Wundeclared-selector` | Warns on undeclared selectors (e.g., `@selector(foo:)` if `foo:` isn't declared). |
| `-Wprotocol` | Warns if a class doesn’t implement all required methods of a protocol. |
| `-Wselector` | Warns on potentially unsafe or unused selectors. |
| `-Wdocumentation` | Warns on malformed or missing API documentation comments. |

### Description

Resolve the nullability warnings showing up when compiling Sentry.
Warnings look like these:

```
Implicit conversion from nullable pointer 'SentryScopePersistentStore * _Nullable' to non-nullable pointer type 'SentryScopePersistentStore * _Nonnull'
```

Some warnings already have valid null checks before them, therefore we only need to silence the warning using `SENTRY_UNWRAP_NULLABLE(<type>, <var>)` as defined in `SentryInternalDefines.h`.

Example:

```
NSString *_Nullable nullableValue = @"some value or null";
if (nullableValue == nil) {
    return;
}
NSString *_Nonnull value = SENTRY_UNWRAP_NULLABLE(NSString, nullableValue);
```

In other cases we need actual handling of the null case, e.g. `SentryRateLimitParser.parseRateLimitSeconds(_:)` should return a nullable value because parsing of a string to a number can fail. These sorts of changes will require additional work and callers need to be updated, including unit tests.

 - [ ] SentryDependencyContainer.m
 - [ ] SentrySwizzle.m
 - [ ] SentryRateLimitParser.m
 - [ ] SentryHttpTransport.m
 - [ ] SentryDsn.m
 - [ ] SentryMetricKitIntegration.m
 - [ ] SentryANRTrackingIntegration.m
 - [ ] SentryAutoBreadcrumbTrackingIntegration.m
 - [ ] SentryUIEventTrackerTransactionMode.m
 - [ ] SentryAppStartTracker.m
 - [ ] SentryNetworkTracker.m
 - [ ] SentryNSURLSessionTaskSearch.m
 - [ ] SentryTimeToDisplayTracker.m
 - [ ] SentryCrashIntegration.m
 - [ ] SentrySessionReplayIntegration.m
 - [ ] SentryHub.m
 - [ ] SentryStacktrace.m
 - [ ] SentrySampling.m
 - [ ] SentrySpan.m
 - [ ] SentryTraceContext.m
 - [ ] SentryTracer.m
 - [ ] PrivateSentrySDKOnly.m
 - [ ] SentryClient.m

philprime:
## :scroll: Description

This PR enables the following nullability flags in the Sentry build settings and fixes all cases of invalid null handling.

**`CLANG_UNDEFINED_BEHAVIOR_SANITIZER_NULLABILITY`:**

[Documentation](https://developer.apple.com/documentation/xcode/build-settings-reference#Enable-Nullability-Annotation-Checks)

Check for violations of nullability annotations in function calls, return statements, and assignments.

**`-Werror=nullable-to-nonnull-conversion` in `OTHER_CFLAGS`:**

[Documentation](https://clang.llvm.org/docs/DiagnosticsReference.html#wnullable-to-nonnull-conversion)

Enables (error) warnings for converting between nullable to non-nullable pointers.

**`-Werror=nullability-completeness` in `OTHER_CFLAGS`:**

[Documentation](https://clang.llvm.org/docs/DiagnosticsReference.html#wnullability-completeness)

Triggers a (error) warning if a type has no nullability defined. This might be enabled by default, so I'll double-check it.

We want to have stricter null checking at compile-time to reduce amount of issues that appear at run-time.

Existing and new unit tests

## :pencil: Checklist

- [ ] Replace silencing of errors with actual null handling


Sources/Sentry/SentryTimeToDisplayTracker.m
```diff
@@ -199,9 +201,14 @@ - (void)framesTrackerHasNewFrame:(NSDate *)newFrameDate
     }
 }
 
-- (void)addTimeToDisplayMeasurement:(SentrySpan *)span name:(NSString *)name
+- (void)addTimeToDisplayMeasurement:(SentrySpan *_Nonnull)span name:(NSString *_Nonnull)name
 {
-    NSTimeInterval duration = [span.timestamp timeIntervalSinceDate:span.startTimestamp] * 1000;
+    NSDate *_Nullable startTimestamp = span.startTimestamp;
+    if (startTimestamp == nil) {
+        // TODO: Handle this case
```
philprime, line 208: Not sure how to handle this case.

Sources/Sentry/SentrySwizzle.m
```diff
@@ -160,7 +160,10 @@ @implementation SentrySwizzle
             // If the class does not implement the method
             // we need to find an implementation in one of the superclasses.
             Class superclass = class_getSuperclass(classToSwizzle);
-            imp = method_getImplementation(class_getInstanceMethod(superclass, selector));
+            Method superMethod = class_getInstanceMethod(superclass, selector);
```
philprime, line 163: Maybe we should just ignore nullability handling here and silence the warning instead

Sources/Sentry/SentryAppStartTracker.m
```diff
@@ -206,6 +206,16 @@ - (void)buildAppStartMeasurement:(NSDate *)appStartEnd
...
         }
 
+        NSDate *_Nonnull sdkStartTimestamp;
+        if (SentrySDK.startTimestamp != nil) {
+            sdkStartTimestamp = (NSDate *_Nonnull)SentrySDK.startTimestamp;
+        } else {
+            SENTRY_LOG_WARN(@"SentrySDK.startTimestamp is nil. Not measuring app start duration. "
+                            @"This is expected if the SDK was not initialized in "
+                            @"application:didFinishLaunchingWithOptions:.");
+            sdkStartTimestamp = [SentryDependencyContainer.sharedInstance.dateProvider date];
```
philprime, line 217: Is this a valid fallback?

Sources/Sentry/SentryAutoBreadcrumbTrackingIntegration.m
```diff
@@ -28,14 +28,22 @@ - (BOOL)installWithOptions:(SentryOptions *)options
     }
 
 #if TARGET_OS_IOS && SENTRY_HAS_UIKIT
+    SentryFileManager *fileManager = [SentryDependencyContainer sharedInstance].fileManager;
+    if (fileManager == nil) {
+        SENTRY_LOG_ERROR(@"File manager is nil. Cannot install auto breadcrumb tracking.");
+        return NO;
```
philprime, line 34: I disable the integraiton here if no file manager is available, which shouldn't be the case here anyways.

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -67,7 +67,10 @@ - (void)binaryImageAdded:(const SentryCrashBinaryImage *)image
 
     SentryBinaryImageInfo *newImage = [[SentryBinaryImageInfo alloc] init];
     newImage.name = imageName;
-    newImage.UUID = [SentryBinaryImageCache convertUUID:image->uuid];
+    NSString *convertedUUID = [SentryBinaryImageCache convertUUID:image->uuid];
+    if (convertedUUID != nil) {
+        newImage.UUID = convertedUUID;
```
philprime, line 72: How should we handle if the conversion fails?

Sources/Sentry/SentryBinaryImageCache.m
```diff
@@ -176,7 +179,11 @@ - (NSInteger)indexOfImage:(uint64_t)address
 - (NSArray<SentryBinaryImageInfo *> *)getAllBinaryImages
 {
     @synchronized(self) {
-        return _cache.copy;
+        if (_cache != nil) {
+            return (NSMutableArray<SentryBinaryImageInfo *> *_Nonnull)_cache.copy;
+        } else {
+            return @[];
```
philprime, line 183: We need to decide if we should just go with no binary images or if we want to indicate an issue by returning `nil`

Sources/Sentry/SentryBreadcrumb.m
```diff
@@ -59,7 +59,10 @@ - (instancetype)init
     NSMutableDictionary *serializedData = [NSMutableDictionary new];
 
     [serializedData setValue:nameForSentryLevel(self.level) forKey:@"level"];
-    [serializedData setValue:sentry_toIso8601String(self.timestamp) forKey:@"timestamp"];
+    if (self.timestamp != nil) {
+        [serializedData setValue:sentry_toIso8601String((NSDate *_Nonnull)self.timestamp)
```
philprime, line 63: This could cause breadcrumbs without timestamps, but only if the breadcrumb was created from an dictionary without a timestamp set.

Sources/Sentry/SentryCrashIntegrationSessionHandler.m
```diff
@@ -80,8 +84,12 @@ - (void)endCurrentSessionIfRequired
                 @"App hang event deleted between check and read. Cannot end current session.");
             return;
         }
-
-        [session endSessionAbnormalWithTimestamp:appHangEvent.timestamp];
+        if (appHangEvent.timestamp != nil) {
+            [session endSessionAbnormalWithTimestamp:(NSDate *_Nonnull)appHangEvent.timestamp];
+        } else {
+            SENTRY_LOG_DEBUG(@"App hang event timestamp is nil. Using current date.");
+            [session endSessionAbnormalWithTimestamp:self.currentDateProvider.date];
```
philprime, line 91: Is using the current timestamp a valid fallback?

Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -256,7 +270,12 @@ - (SentryThread *_Nullable)threadAtIndex:(NSInteger)threadIndex
     }
     NSDictionary *threadDictionary = self.threads[threadIndex];
 
-    SentryThread *thread = [[SentryThread alloc] initWithThreadId:threadDictionary[@"index"]];
+    id threadId = threadDictionary[@"index"];
+    if (![threadId isKindOfClass:[NSNumber class]]) {
+        threadId = @(0);
```
philprime, line 275: We could either use this approach and fallback to the main thread, or just return `nil`. As returning `nil` would cause threads to be removed, we could also just use an arbitrary high thread id to indicate that something went wrong.

Sources/Sentry/SentryDevice.m
```diff
@@ -236,7 +236,7 @@
     size_t size = sizeof(str);
     int cmd[2] = { CTL_KERN, KERN_OSVERSION };
     if (SENTRY_LOG_ERRNO(sysctl(cmd, sizeof(cmd) / sizeof(*cmd), str, &size, NULL, 0)) == 0) {
-        return [NSString stringWithUTF8String:str];
+        return [NSString stringWithUTF8String:str] ?: @"<nil>";
```
philprime, line 239: What's the best fallback here?

Sources/Sentry/SentryNSURLSessionTaskSearch.m
```diff
@@ -52,10 +53,10 @@ @implementation SentryNSURLSessionTaskSearch
 
     while (class_getInstanceMethod(currentClass, setStateSelector)) {
         Class superClass = [currentClass superclass];
-        IMP classResumeIMP
-            = method_getImplementation(class_getInstanceMethod(currentClass, setStateSelector));
-        IMP superclassResumeIMP
-            = method_getImplementation(class_getInstanceMethod(superClass, setStateSelector));
+        IMP classResumeIMP = method_getImplementation(
```
philprime, line 56: I am uncertain if we need to have a null handler here.

Sources/Sentry/SentryRateLimitParser.m
```diff
@@ -78,7 +78,12 @@ - (NSNumber *)parseRateLimitSeconds:(NSString *)string
 {
     NSNumberFormatter *numberFormatter = [[NSNumberFormatter alloc] init];
     numberFormatter.numberStyle = NSNumberFormatterNoStyle;
-    return [numberFormatter numberFromString:string];
+    NSNumber *_Nullable result = [numberFormatter numberFromString:string];
+    if (result != nil) {
+        return (NSNumber *_Nonnull)result;
+    } else {
+        return @0; // Default to 0 if parsing fails
```
philprime, line 83: Not sure if we should return `nil`

Sources/Sentry/SentryRateLimitParser.m
```diff
@@ -100,12 +105,17 @@ - (NSNumber *)parseRateLimitSeconds:(NSString *)string
...
          andRateLimitInSeconds:(NSNumber *)newRateLimitInSeconds
 {
     NSDate *newDate = [self.currentDateProvider.date
         dateByAddingTimeInterval:[newRateLimitInSeconds doubleValue]];
-    return [SentryDateUtil getMaximumDate:newDate andOther:existingRateLimit];
+    NSDate *_Nullable result = [SentryDateUtil getMaximumDate:newDate andOther:existingRateLimit];
+    if (result != nil) {
+        return (NSDate *_Nonnull)result;
+    } else {
+        return newDate; // Default to newDate if getMaximumDate returns nil
```
philprime, line 114: Not sure if this default is the correct approach

Sources/Sentry/SentryScope.m
```diff
@@ -626,7 +641,12 @@ - (void)addObserver:(id<SentryScopeObserver>)observer
...
-        return [span serialize];
+        NSDictionary *_Nullable serialized = [span serialize];
+        if (serialized != nil) {
+            return (NSDictionary *_Nonnull)serialized;
+        } else {
+            return @{};
+        }
     } else {
         return [self.propagationContext traceContextForEvent];
     }
```
philprime, line 638: It might make sense to fallback to the `[self.propagationContext traceContextForEvent];`

Sources/Sentry/SentryTracer.m
```diff
@@ -587,8 +590,12 @@ - (void)finishInternal
     SentryTransaction *transaction = [self toTransaction];
 
 #if SENTRY_TARGET_PROFILING_SUPPORTED
-    sentry_stopProfilerDueToFinishedTransaction(
-        _hub, _dispatchQueue, transaction, _isProfiling, self.startTimestamp, _startSystemTime
+    NSDate *_Nullable startTimestamp = self.startTimestamp;
+    if (startTimestamp == nil) {
+        // TODO: Handle this case
```
philprime, line 595: I am not sure how we need to handle this case.

Sources/Sentry/SentryTracer.m
```diff
@@ -647,7 +654,12 @@ - (BOOL)finishTracer:(SentrySpanStatus)unfinishedSpansFinishStatus shouldCleanUp
     // Prewarming can execute code up to viewDidLoad of a UIViewController, and keep the app in the
     // background. This can lead to auto-generated transactions lasting for minutes or even hours.
     // Therefore, we drop transactions lasting longer than SENTRY_AUTO_TRANSACTION_MAX_DURATION.
-    NSTimeInterval transactionDuration = [self.timestamp timeIntervalSinceDate:self.startTimestamp];
+    NSDate *_Nullable startTimestamp = self.startTimestamp;
+    if (startTimestamp == nil) {
+        // TODO: Handle this case
```
philprime, line 659: Also not sure how to handle this case

Sources/Sentry/SentryTracer.m
```diff
@@ -825,7 +840,12 @@ - (nullable SentryAppStartMeasurement *)getAppStartMeasurement SENTRY_DISABLE_TH
     NSDate *appStartEndTimestamp =
         [appStartTimestamp dateByAddingTimeInterval:measurement.duration];
 
-    NSTimeInterval difference = [appStartEndTimestamp timeIntervalSinceDate:self.startTimestamp];
+    NSDate *_Nullable spanStartTimestamp = self.startTimestamp;
+    if (spanStartTimestamp == nil) {
+        // TODO: Handle this case
```
philprime, line 845: Also not sure how to handle this case.

Sources/Sentry/SentryFileManager.m
```diff
@@ -158,7 +158,12 @@ - (void)createPathsWithOptions:(SentryOptions *)options
     SENTRY_LOG_DEBUG(@"SentryFileManager.cachePath: %@", cachePath);
 
     self.basePath = [cachePath stringByAppendingPathComponent:@"io.sentry"];
-    self.sentryPath = [self.basePath stringByAppendingPathComponent:[options.parsedDsn getHash]];
+    NSString *dsnHash = [options.parsedDsn getHash];
+    if (dsnHash != nil) {
+        self.sentryPath = [self.basePath stringByAppendingPathComponent:dsnHash];
+    } else {
+        self.sentryPath = [self.basePath stringByAppendingPathComponent:@"unknown"];
```
philprime, line 165: We need some sort of fallback, `unknown` might not be clear enough, but any string could work

Sources/Sentry/SentrySDKInternal.m
```diff
@@ -121,22 +121,22 @@ + (SentryLogger *)logger
...
-                    initWithClient:currentHub.client
+                    initWithClient:(SentryClient *_Nonnull)currentHub.client
                      dispatchQueue:SentryDependencyContainer.sharedInstance.dispatchQueueWrapper];
             } else {
                 batcher = nil;
             }
 
             currentLogger = [[SentryLogger alloc]
-                 initWithHub:currentHub
+                 initWithHub:(SentryHub *_Nonnull)currentHub // TODO: Remove this force-cast from
```
philprime: We need to properly handle the fallback when the hub is `null`. I would go with a force-cast for now and fix this in a dependency container refactoring

Sources/Sentry/SentryOptions.m
```diff
@@ -216,9 +216,13 @@ - (instancetype)init
...
         self.failedRequestStatusCodes = @[ defaultHttpStatusCodeRange ];
-        self.cacheDirectoryPath
-            = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES)
-                  .firstObject;
+        NSArray<NSString *> *cachePaths
+            = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
+        if (cachePaths.count > 0 && cachePaths.firstObject != nil) {
+            self.cacheDirectoryPath = (NSString *_Nonnull)cachePaths.firstObject;
+        } else {
+            self.cacheDirectoryPath = @""; // Default to empty string if no cache path found
```
philprime, line 176: We might want to change the `cacheDirectoryPath` to be nullable, so that we can do a null fallback

Sources/Sentry/SentryNetworkTracker.m
```diff
@@ -511,7 +544,12 @@ - (void)addBreadcrumbForSessionTask:(NSURLSessionTask *)sessionTask
     SentryBreadcrumb *breadcrumb = [[SentryBreadcrumb alloc] initWithLevel:breadcrumbLevel
                                                                   category:@"http"];
 
-    UrlSanitized *urlComponents = [[UrlSanitized alloc] initWithURL:sessionTask.currentRequest.URL];
+    NSURL *requestURL = sessionTask.currentRequest.URL;
+    if (requestURL == nil) {
+        // Use a fallback URL if the request URL is nil
+        requestURL = [NSURL URLWithString:@"unknown"];
```
philprime, line 550: Instead of falling back to a dummy URL, we might want to drop the breadcrumb. Thoughts?

Sources/Sentry/Profiling/SentryProfilerState.mm
```diff
@@ -28,7 +28,7 @@
                                  options:0
                                    error:nil];
     });
-    const auto symbolNSStr = [NSString stringWithUTF8String:symbol];
+    const auto symbolNSStr = [NSString stringWithUTF8String:symbol] ?: @"<nil>";
```
philprime, line 31: I am uncertain if we should prefer an empty string or `<nil>` to indicate that the symbol could not be converted

Sources/Sentry/SentryHub.m
```diff
@@ -457,17 +477,24 @@ - (SentryTracer *)startTransactionWithContext:(SentryTransactionContext *)transa
...
+        // Create a default sampler decision when no options are available
+        tracesSamplerDecision = [[SentrySamplerDecision alloc] init];
+        [tracesSamplerDecision setValue:@(kSentrySampleDecisionYes) forKey:@"decision"];
+    }
     transactionContext = [self transactionContext:transactionContext
                                       withSampled:tracesSamplerDecision.decision
-                                       sampleRate:tracesSamplerDecision.sampleRate
-                                       sampleRand:tracesSamplerDecision.sampleRand];
+                                       sampleRate:tracesSamplerDecision.sampleRate ?: @(1.0)
+                                       sampleRand:tracesSamplerDecision.sampleRand ?: @(0.0)];
```
philprime, line 479: I am uncertain about this

philprime:
This PR will be split into smaller ones.

philprime:
Closing this because the compiler already shows nullability warnings on `main` and this branch is too far diverted.

PR #5578, opened 2025-07-07
ref: SentryNSNotificationCenterWrapper in Swift

noahsmartin:
Converts SentryNSNotificationCenterWrapper to Swift, which is one of the highest dependency objc classes so it helps future classes get moved to Swift.

There is a notable difference between the Swift and ObjC versions that makes the codebase a bit more "Swifty". Instead of a class that wraps NSNotificationCenter, the dependency container now vends a type conforming to the `SentryNSNotificationCenterWrapper` protocol. It's more idiomatic Swift because Swift prefers protocols to inheritance, and now the tests can use protocol conformance rather than inheritance. It also makes the SDK more performant because we can directly conform NotificationCenter to the protocol rather than defining new types.

In a future where more of the codebase is in Swift it also enables us to do things like this:
```
#if TEST
protocol SentryNSNotificationCenterWrapper {}
#else
typealias SentryNSNotificationCenterWrapper = NotificationCenter
#endif
```

That would mean we could make injecting Notification Center fully testable without any runtime overhead for release builds. This only works if all the call-sites are Swift so still a while out, but we are more prepared for that now with a protocol instead of a base class.

I had to update many callosities to pass the null `object` param since it is required by the Notification Center API. In a Swift world we could have defined it as an extension on the protocol like this:

```
extension SentryNSNotificationCenterWrapper {
  func removeObserver(_ observer: Any, name aName: NSNotification.Name?) {
    removeObserver(observer, name: aName, object: nil)
  }
}
```
However, extensions of protocols are not visible to ObjC, so had to add in all the object params for now. I think it's a small price to pay for having the protocol based injection.

#skip-changelog


Tests/SentryTests/Helper/SentryNSNotificationCenterWrapperTests.swift
```diff
@@ -1,97 +0,0 @@
-import SentryTestUtils
-import XCTest
-
-class SentryNSNotificationCenterWrapperTests: XCTestCase {
```
noahsmartin, line 4: These tests are not needed anymore since `SentryNSNotificationCenterWrapper` is no longer a class with an implementation to test, it's just a protocol and the conforming type (NotificationCenter) is provided by the OS.

Sources/Sentry/include/SentryNSNotificationCenterWrapper.h
```diff
@@ -9,46 +9,3 @@
...
-
-/**
- * A wrapper around @c NSNotificationCenter functions for testability.
- * @discussion Testing with @c NSNotificationCenter in CI leads to flaky tests for some classes.
- * Therefore, we can use a wrapper around @c NSNotificationCenter to not depend on it. Instead, we
- * call the methods
- * @c NSNotificationCenter would call with Dynamic and ensure that sut properly subscribes to
- * @c NSNotificationCenter.
- */
-@interface SentryNSNotificationCenterWrapper : NSObject
```
itaybre, line 23: Not a big issue, but maybe we need to rename this file to better represent it only contains some notifications

And also remove the import

noahsmartin: I believe the import is still used because it's how the file gets `SENTRY_HAS_UIKIT`. Will double check and rename though

PR #5591, opened 2025-07-08
V9 config

noahsmartin:
WIP

#skip-changelog


philipphofmann (approved):
I like this approach. I think we should also document this in our https://github.com/getsentry/sentry-cocoa/blob/main/develop-docs/README.md. You could also parts of https://www.notion.so/sentry/sentry-cocoa-V9-2228b10e4b5d80a1bbaadfcaee3b44ca .

Let's give this a shot. Thanks @noahsmartin

PR #5598, opened 2025-07-09
feat(v9): Remove deprecated SentryDebugImageProvider in V9

noahsmartin:
Moves SentryDebugImageProvider to be an internal class for v9, visible only to the hybrid SDK. The only methods on this class that are currently public are deprecated and now removed behind the V9 flag.

#skip-changelog


philipphofmann (approved):
@noahsmartin, perhaps it makes sense to keep an internal deprecation changelog in `/develop-docs`, making it easier to create a proper changelog for V9. Finding all the PRs later could be challenging. 

LGTM

noahsmartin:
Yeah I was thinking the same thing @philipphofmann I'll do that today

PR #4946, opened 2025-03-06
meta: don't treat swift warnings as errors

armcknight:
Due to the conversation here: https://github.com/getsentry/sentry-cocoa/pull/4854#discussion_r1959308210

We need to turn off treating warnings as errors in Swift code. To retain confidence in our release integrity, this PR (in addition to setting that setting to `NO`) introduces a script that checks a build log for any warnings, and compares them to a list of known warnings, failing if any warnings were present in the build log that weren't in the known warning list.

It also provides a facility to incorporate new warnings into the list of known warnings. The envisioned workflow is that this script fails CI due to new warnings being found, then we fix the warnings we deem unacceptable, then rerun the build, and rerun the script with the incorporation option set to update the known warnings list.

#skip-changelog

PR #5599, opened 2025-07-09
feat(v9): Mark deprecated APIs as deprecated

noahsmartin:
These are all deprecated but were not using the deprecated attribute

#skip-changelog


armcknight:
The CI failures here are due to things we previously just punted on here: https://github.com/getsentry/sentry-cocoa/pull/4854#discussion_r1959308210 and here: https://github.com/getsentry/sentry-cocoa/pull/4946

armcknight (approved):

PR #5600, opened 2025-07-09
feat(SPM): Support full build with SPM

noahsmartin:
These are the remaining changes necessary to get SPM building the entire codebase, previously it didn't support profiling code which is why CI built for watchOS. Now it's building iOS in CI.

The main change here is to pull out code that uses Swift from ObjectiveC++ files (.mm) and into regular .m files. I added a few helper functions in `SentryProfilingSwiftHelpers.m` to act as this ObjectiveC++ to Swift bridge.

#skip-changelog


itaybre (commented):
Could you add some tests for `SentryProfilingSwiftHelpers.m`?
just to ensure we don't break those in future changes

noahsmartin:
@itaybre good call, just added a few tests

itaybre (approved):

Sources/Sentry/include/SentryProfilingSwiftHelpers.h
```diff
@@ -0,0 +1,49 @@
...
+
+// The functions in this file exist to bridge ObjectiveC++ to Swift. When building with Swift
+// Package Manager you can’t import Swift into ObjectiveC++ so instead that code calls plain C
+// functions in this file which then uses Swift in their implementation.
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+BOOL isContinuousProfilingEnabled(SentryClient *client);
```
armcknight: if these are globally visible functions, we may want to prefix them with `sentry_`

armcknight (approved):
Yep, the basic idea here is sound. Just the one question about C function prefixes.

Sources/Sentry/include/SentryProfilingSwiftHelpers.h
noahsmartin: They are in internal header files so I don't think they need to be, but it seems like it can't hurt, I'll add that

PR #5604, opened 2025-07-10
ref: Extract video processing to a new class

itaybre:
Extract video generation to an independent class to enable testing


Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
```diff
@@ -0,0 +1,170 @@
...
+            case .failed:
+                SentrySDKLog.warning("[Session Replay] Finish writing video failed, reason: \(String(describing: videoWriter.error))")
+                completion(.failure(videoWriter.error ?? SentryOnDemandReplayError.errorRenderingVideo))
+            case .unknown:
+                SentrySDKLog.warning("[Session Replay] Finish writing video with unknown status, reason: \(String(describing: videoWriter.error))")
+                completion(.failure(videoWriter.error ?? SentryOnDemandReplayError.errorRenderingVideo))
+            @unknown default:
+                SentrySDKLog.warning("[Session Replay] Finish writing video in unknown state, reason: \(String(describing: videoWriter.error))")
+                completion(.failure(SentryOnDemandReplayError.errorRenderingVideo))
+            }
```
philippsuess, line 135: `m`: If I'm not mistaken, you copied this from `SentryOnDemandReplay.finishVideo`, but the PR doesn't remove `SentryOnDemandReplay.finishVideo`. Now `SentryOnDemandReplay.finishVideo` is duplicate and unused, if I'm not mistaken.

I think the same applies to `SentryOnDemandReplay.getVideoInfo`

Tests/SentryTests/Integrations/SessionReplay/SentryVideoFrameProcessorTests.swift
```diff
@@ -0,0 +1,480 @@
...
+            requestMediaDataWhenReadyQueue = queue
+            requestMediaDataWhenReadyBlock = block
+        }
+
+        override func markAsFinished() {
+            // No-op for testing
+        }
+    }
+
+    private class Fixture {
```
philippsuess, line 96: `m`: I know we use the Fixture pattern heavily in the code base, but we kind of informally agreed that we don't have to use it. It's a bit verbose when using it, and if often ends up being a state mess. Meaning we have tons of different configurations in it and we use different flags to change behavior. Also you need to jump from the test to the fixture to know what the test is doing. Instead, we try to do as much as possible in the test directly to keep each test understandable. Basically by just reading the test you should know what it's testing. If you have common logic to setup the sut we simply now create top level functions. You can also keep it as it is.

philipphofmann (approved):
LGTM, when removing the duplicate code.

Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
itaybre: Missed that? Oops

philippsuess: That's why I asked for an extra refactoring PR. Then it's easier to catch such mistakes in the diff 😃 . Can happen no worries.

PR #5607, opened 2025-07-10
fix: Ensure completion callback is invoked only once in SentryVideoFrameProcessor

itaybre:
Avoid calling the completion block twice and crashing the app

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Extract video processing to a new class (#5604)
 
+### Fixes
+
+- Ensure completion callback is invoked only once in SentryVideoFrameProcessor (#5607)
+
 ## 8.53.2
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Extract video processing to a new class (#5604)
 
+### Fixes
+
+- Ensure completion callback is invoked only once in SentryVideoFrameProcessor (#5607)
```
noahsmartin, line 11: This is not part of the public API right? So doesn't seem like it should be part of the changeling. If it fixes a bug then maybe we should say which bug it's fixing here?

itaybre: Fair point, will update it to mention the bug

Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
```diff
@@ -46,15 +46,22 @@ class SentryVideoFrameProcessor {
     }
 
     func processFrames(videoWriterInput: AVAssetWriterInput, onCompletion: @escaping (Result<SentryRenderVideoResult, Error>) -> Void) {
+        let completionBlock: (Result<SentryRenderVideoResult, Error>) -> Void = { [weak self] result in
```
noahsmartin, line 49: I'm not sure I'm seeing how it would have been called more than once before this change. I wonder if a more specific fix, like filling in a missing return or whatever the issue was causing multiple calls, would be better? I'm worried this fixes the specific issue of completion called multiple times, but other side effects could still happen from whatever code path it went through twice.


Also, is it possible the block would never be called now because isFinished gets set to true somewhere else?

itaybre:
Real issue found in #5608, closing this PR

PR #5608, opened 2025-07-10
fix: Ensure video writer inputs are marked as finished on cancellation in SentryVideoFrameProcessor

itaybre:


philippsuess:
Thank you @itaybre

Issue #5609, opened 2025-07-10
Hang detection deadlocks when debug is enabled


noahsmartin:
### Platform

iOS

### Environment

Develop

### Installed

Manually

### Version

8.53.2

### Xcode Version

16.5

### Steps to Reproduce

In https://github.com/getsentry/sentry-cocoa/pull/4063 we fixed app hang deadlocks when SentryOptions.debug was set to false. However there are still known deadlocks when debug is set to true.

I just ran into one here (note I had made some changes to the code working on some hang detection updates, but the cause of the bug is still present in the latest code on main):

<img width="4964" height="3034" alt="Image" src="https://github.com/user-attachments/assets/59a158d0-1d51-48e0-aa56-838ed42c715a" />

<img width="4964" height="3034" alt="Image" src="https://github.com/user-attachments/assets/46c26d25-0027-4913-91c8-e22078d9ea43" />

I imagine it's also possible this is the source of some flaky tests

### Expected Result

Should not be attempting to symbolicate while threads are suspended, we should instead only do that after retrieving the stack traces and resuming the threads

### Actual Result

The app hangs

philippsuess:
@noahsmartin thanks for reporting this. We disable certain features, such as app hang detection, because it can lead to deadlocks when the debugger is attached. See

https://github.com/getsentry/sentry-cocoa/blob/43597ba400daa59034f323b22bbc6695c185e2fa/Sources/Sentry/SentryANRTrackingIntegration.m#L77-L80

and 

https://github.com/getsentry/sentry-cocoa/blob/43597ba400daa59034f323b22bbc6695c185e2fa/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitorType.h#L62-L79

itaybre:
We are closing this issue because attaching the debugger should not be used with app hangs.

There is a related issue for the symbolication in [https://github.com/getsentry/sentry-cocoa/issues/2996](https://github.com/getsentry/sentry-cocoa/issues/2996) that we will follow up

noahsmartin:
I don't think this is related to the debugger being attached right? It's just anytime you run the SDK with `debug: true`. If the debugger is also connected you won't run into it, but you can run with `debug: true` and the debugger disconnected. @itaybre please let me know if that's not the case and we can re-close, but the way I understand it this is still an issue because it's caused by our "debug" flag, not the debugger.

philippsuess:
[...]

PR #5610, opened 2025-07-10
Make async unsafe more clear

noahsmartin:
Noticed this when I hit deadlock caused by this code: https://github.com/getsentry/sentry-cocoa/issues/5609

Since there were two methods "sentrycrashsymbolicator_symbolicate_async_unsafe" and "sentrycrashsymbolicator_symbolicate" but both are async unsafe, I figured the second one should be renamed

#skip-changelog


armcknight (approved):

philipphofmann (approved):
Thanks, LGMT 💯 

Nitpick: You could prefix refactoring PR titles with `ref:` so it's clear from reading the title it's a refactoring.

PR #5612, opened 2025-07-10
ref: Update frame processing logic in SentryVideoFrameProcessor to handle multiple frames when input is ready

itaybre:
#skip-changelog

### DIFF
--- a/Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
+++ b/Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
@@ -46,46 +46,48 @@ class SentryVideoFrameProcessor {
     }
 
     func processFrames(videoWriterInput: AVAssetWriterInput, onCompletion: @escaping (Result<SentryRenderVideoResult, Error>) -> Void) {
-        SentrySDKLog.debug("[Session Replay] Video writer input is ready, status: \(videoWriter.status)")
-        guard videoWriter.status == .writing else {
-            SentrySDKLog.error("[Session Replay] Video writer is not writing anymore, cancelling the writing session, reason: \(videoWriter.error?.localizedDescription ?? "Unknown error")")
-            videoWriter.inputs.forEach { $0.markAsFinished() }
-            videoWriter.cancelWriting()
-            return onCompletion(.failure(videoWriter.error ?? SentryOnDemandReplayError.errorRenderingVideo))
+        while videoWriterInput.isReadyForMoreMediaData {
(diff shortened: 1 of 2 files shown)


Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
```diff
@@ -46,46 +46,49 @@ class SentryVideoFrameProcessor {
...
-        let presentTime = SentryOnDemandReplay.calculatePresentationTime(
-            forFrameAtIndex: frameIndex,
-            frameRate: frameRate
-        ).timeValue
-        guard currentPixelBuffer.append(image: image, presentationTime: presentTime) else {
-            SentrySDKLog.error("[Session Replay] Failed to append image to pixel buffer, cancelling the writing session, reason: \(String(describing: videoWriter.error))")
-            videoWriter.inputs.forEach { $0.markAsFinished() }
-            videoWriter.cancelWriting()
-            return onCompletion(.failure(videoWriter.error ?? SentryOnDemandReplayError.errorRenderingVideo))
+        while videoWriterInput.isReadyForMoreMediaData {
```
philippsuess, line 52: `m:` At first glance, this looks like something that could easily lead to an endless loop. It would be great to add a comment that calling `isReadyForMoreMediaData` in while is the recommended approach of https://developer.apple.com/documentation/avfoundation/avassetwriterinput/requestmediadatawhenready(on:using:)

After looking at the tests and the code docs, it seems like not doing this while loop will lead to missing frames. If this fixes a bug, it would be great to document this here as well.

philipphofmann (approved):
Shouldn't we add a changelog entry for this? This is basically using the recommended approach from the [Apple docs](https://developer.apple.com/documentation/avfoundation/avassetwriterinput/requestmediadatawhenready(on:using:)). We previously didn't check for `isReadyForMoreMediaData`, which we should according to the docs. This change could fix potential problems or at least speed up the proccess as pointed out in the PR description. If it's only speeding up, it's an improvement worthy to be on the changelog.

LGTM, thanks.

itaybre:
> Shouldn't we add a changelog entry for this? This is basically using the recommended approach from the [Apple docs](https://developer.apple.com/documentation/avfoundation/avassetwriterinput/requestmediadatawhenready(on:using:)). We previously didn't check for `isReadyForMoreMediaData`, which we should according to the docs. This change could fix potential problems or at least speed up the proccess as pointed out in the PR description. If it's only speeding up, it's an improvement worthy to be on the changelog.
> 
> LGTM, thanks.

I wasn't sure this was worthy enough for a changelog entry

PR #5613, opened 2025-07-10
Armcknight/impr/profile serialization off main thread

armcknight:
It was pointed out that we're performing JSON serialization of continuous profiling chunks on the main thread. Move it off to a background thread similar to how it was done for transaction profiling in https://github.com/getsentry/sentry-cocoa/pull/4377.

### DIFF
--- a/Sources/Sentry/Profiling/SentryContinuousProfiler.mm
+++ b/Sources/Sentry/Profiling/SentryContinuousProfiler.mm
@@ -61,8 +61,8 @@ _sentry_threadUnsafe_transmitChunkEnvelope(void)
     [profiler.state clear]; // !!!: profile this to see if it takes longer than one sample duration
                             // length: ~9ms
 
-    const auto metricProfilerState = [profiler.metricProfiler serializeContinuousProfileMetrics];
-    [profiler.metricProfiler clear];
+    // Capture the metric profiler reference for background serialization
+    const auto metricProfiler = profiler.metricProfiler;
 
 #    if SENTRY_HAS_UIKIT
     const auto framesTracker = SentryDependencyContainer.sharedInstance.framesTracker;
@@ -70,14 +70,25 @@ _sentry_threadUnsafe_transmitChunkEnvelope(void)
     [framesTracker resetProfilingTimestamps];
 #    endif // SENTRY_HAS_UIKIT
 
-    const auto envelope = sentry_continuousProfileChunkEnvelope(
-        profiler.profilerId, profilerState, metricProfilerState
+    // Capture profiler ID on main thread since we need it for the background work
+    const auto profilerID = profiler.profilerId;
+
+    // Move the serialization work to a background queue to avoid potentially
+    // blocking the main thread. The serialization can take several milliseconds.
+    [SentryDependencyContainer.sharedInstance.dispatchQueueWrapper dispatchAsyncWithBlock:^{
+        // Perform metric serialization on background queue, then clear
+        const auto metricProfilerState = [metricProfiler serializeContinuousProfileMetrics];
+        [metricProfiler clear];
+
+        const auto envelope
+            = sentry_continuousProfileChunkEnvelope(profilerID, profilerState, metricProfilerState
 #    if SENTRY_HAS_UIKIT
-        ,
-        screenFrameData
+                ,
+                screenFrameData
 #    endif // SENTRY_HAS_UIKIT
-    );
-    [SentrySDK captureEnvelope:envelope];
+            );
+        [SentrySDK captureEnvelope:envelope];
+    }];
 }
 
 void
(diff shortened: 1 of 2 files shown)


Sources/Sentry/Profiling/SentryContinuousProfiler.mm
```diff
@@ -61,23 +61,34 @@
...
 
-    const auto envelope = sentry_continuousProfileChunkEnvelope(
-        profiler.profilerId, profilerState, metricProfilerState
+    // Capture profiler ID on main thread since we need it for the background work
+    const auto profilerID = profiler.profilerId;
+
+    // Move the serialization work to a background queue to avoid potentially
+    // blocking the main thread. The serialization can take several milliseconds.
+    [SentryDependencyContainer.sharedInstance.dispatchQueueWrapper dispatchAsyncWithBlock:^{
+        // Perform metric serialization on background queue, then clear
```
philippsuess: `l`: Kind of redundant, this comment. We could ditch it.
```suggestion
```

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
```diff
@@ -61,23 +61,34 @@
...
+        const auto metricProfilerState = [metricProfiler serializeContinuousProfileMetrics];
+        [metricProfiler clear];
```
philippsuess: `m`: Clearing now happens on a bg thread and therefore a bit later than previously on the calling thread. Is this acceptable? Can this lead to race conditions? The `serializeContinuousProfileMetrics` seems not so expensive, so it might be acceptable doing this on the callig thread.

philipphofmann (commented):
Thanks @armcknight, LGTM with one comment.

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
armcknight: Yeah I went back and forth on that, but if we immediately serialize the metrics, and then wait arbitrary time to serialize the profile data, then you wind up with start/end time mismatches.

noahsmartin (approved):

Sources/Sentry/Profiling/SentryContinuousProfiler.mm
armcknight: Refactored to copy the metrics data and clear before the dispatch, the same as the normal profiling data, in https://github.com/getsentry/sentry-cocoa/pull/5613/commits/b8586c96752283c5c3ded6f14862f9e566229f33. passed all unit tests locally and verified manually at https://sentry-sdks.sentry.io/explore/profiling/profile/sentry-cocoa/flamegraph/?colorCoding=by%20system%20vs%20application%20frame&end=2025-07-11T19%3A08%3A50.671Z&eventId=14234a17660e4127a17b2dc84120a396&profilerId=e5056a830709480aba041de4fe2b5e07&query=&sorting=call%20order&start=2025-07-11T19%3A08%3A39.020Z&tid=259&view=top%20down
image

Sources/Sentry/include/SentryMetricProfiler.h
```diff
@@ -45,23 +63,10 @@ SENTRY_NO_INIT
...
- *          // ... more dictionaries like that ...
- *      ]
- * }
- * @endcode
+ * Copy the data structures in a thread safe way so they can be used later to serialize into a
+ * payload for transmission.
  */
-- (NSMutableDictionary<NSString *, SentrySerializedMetricEntry *> *)
-    serializeContinuousProfileMetrics;
+- (NSDictionary *)copyMetricProfilerData;
```
itaybre, line 69: 👏

Sources/Sentry/SentryMetricProfiler.m
```diff
@@ -177,36 +206,15 @@ - (void)stop
...
-                cpuEnergyUsage, kSentryMetricProfilerSerializationUnitNanoJoules);
-    }
-
-    if (cpuUsage.count > 0) {
-        dict[kSentryMetricProfilerSerializationKeyCPUUsage]
-            = serializeContinuousProfileMetricReadings(
-                cpuUsage, kSentryMetricProfilerSerializationUnitPercentage);
+        copy[kSentryMetricProfilerSerializationKeyMemoryFootprint] = _memoryFootprint;
+        copy[kSentryMetricProfilerSerializationKeyCPUUsage] = _cpuUsage;
+        copy[kSentryMetricProfilerSerializationKeyCPUEnergyUsage] = _cpuEnergyUsage;
```
itaybre: Should these be deep copies?

armcknight: Yes, you're right. I made these all `= [theThing copy]` 🤦🏻 this ain't swift!

noahsmartin (approved):

PR #5614, opened 2025-07-11
meta: claude init

armcknight:
I've started using Claude Code, and wanted to introduce a CLAUDE.md into our repo to help with it.

For future reference, we have a couple other repos at Sentry that already have a CLAUDE.md file: 
- https://github.com/getsentry/sentry-javascript/blob/develop/CLAUDE.md
- https://github.com/getsentry/getsentry/blob/master/CLAUDE.md
- https://github.com/getsentry/sentry-mcp/blob/main/CLAUDE.md
- https://github.com/getsentry/vanguard/blob/main/CLAUDE.md

And there is also a general collection of rules at https://github.com/steipete/agent-rules. 

And then there are the anthropic docs: https://www.anthropic.com/engineering/claude-code-best-practices and https://docs.anthropic.com/en/docs/claude-code/memory were two I read while going through this.

Keeping those in mind, let's iterate on this going forward, step by step.

#skip-changelog

PR #5620, opened 2025-07-11
docs: reorganize maintainer markdown documentation

armcknight:
based on feedback in https://github.com/getsentry/sentry-cocoa/pull/5614, moving maintainer doc reorg to a separate PR

I'd previously remarked that we should consider consolidating our CONTRIBUTING.md, README.md and develop-docs files into one. I took this opportunity to instead reorganize them into smaller, more digestible, topical documents, which I then reference from CLAUDE.md because that had started simply reiterating what we'd already written in those after I ran `/init` in Claude Code and iterated on it a few times, and I don't think we should have duplication.

Nothing was removed (unless it referenced something that didn't even exist anymore, like `scripts/upload-dsyms-with-xcode-build-phase.sh`), I only moved things to the topical markdown docs or to other preexisting markdown docs where they made more sense, like Samples/README.md.

#skip-changelog


develop-docs/TEST.md
```diff
@@ -0,0 +1,144 @@
+# Testing
+
+## Sample Apps
+
+See //Samples/README.md for more information about how to use the sample apps to test SDK functionality.
```
itaybre, line 5: This is just moved content, right?

develop-docs/RELEASE.md
```diff
@@ -0,0 +1,18 @@
...
+
+1. Checkout a new branch from the GH tag of the beta release: `git checkout -b publish/x.x.x x.x.x-beta.1`. You can't use `release/x.x.x` or `x.x.x` as the branch name as craft will fail, as it creates a `release/x.x.x` branch for updating the changelog and it will create a tag `x.x.x` for the release.
+2. Duplicate the Changelog.md entry of the beta release and change header of the version number to unreleased.
+3. Commit and push the changes.
+4. Trigger the release workflow with use workflow from the `publish/x.x.x` branch and set the target branch to merge into to `publish/x.x.x`, cause per default craft will merge into the main branch and this could lead to merge conflicts in the changelog.
+5. After the successful release, validate that craft merged the changes back into `publish/x.x.x` branch and deleted the release branch.
+6. Manually open a PR from the `publish/x.x.x` branch into the main branch and merge it.
```
itaybre, line 18: This could be removed, right?
We use the action now

itaybre (approved):
LGTM, just added a few questions

develop-docs/TEST.md
armcknight: There was already a README.md in //Samples, but I moved more stuff there, like codesigning stuff that only applied to them

develop-docs/RELEASE.md
armcknight: I don't really know, I've never done the beta promotion process, @philipphofmann would need to answer this. but,  we should probably do that in a different PR anyways, to minimize the actual semantic content changes in this one

PR #5622, opened 2025-07-12
feat(v9): Add API stability for V9

noahsmartin:
Added a CI check for API changes in v9, also made the existing API check faster by only building for iphoneos (previously it was also building for simulator but only using the device build)

#skip-changelog


Makefile
```diff
@@ -87,7 +87,7 @@ format-yaml:
 	dprint fmt "**/*.{yaml,yml}"
 
 generate-public-api:
-	./scripts/update-api.sh
+	./scripts/update-api.sh $(CONFIG)
```
itaybre: Wouldn't generating both json be easier?
```suggestion
	./scripts/update-api.sh
	./scripts/update-api.sh V9
```

noahsmartin: Sure we can do that, makes sense

PR #5363, opened 2025-06-05
tests: Change fatal test assert to fatal error log for multiple span finish calls

philipphofmann:
During an investigation of the flaky UI tests @philprime and myself discovered that the `SentryFramesTracker` and `SentryTimeToDisplayTracker` both try to finish the same span, therefore causing the `SENTRY_TEST_FATAL` to trigger a crashing test.

https://github.com/getsentry/sentry-cocoa/blob/49e2bd0ee860b000d1ed31b9c628c636dfec8b9c/Sources/Sentry/Profiling/SentryProfiledTracerConcurrency.mm#L80-L87

As this is blocking our release pipeline, we need to change it to a non-fatal test and figure out a solution to solve the double-finish in another PR.

We currently assume that the changes introduced in #5063 could be the cause of the issue.

Fixes https://github.com/getsentry/sentry-cocoa/issues/5356 and https://github.com/getsentry/sentry-cocoa/issues/5345.

#skip-changelog

PR #5624, opened 2025-07-12
test: remove remaining SENTRY_TEST_FATAL macro usages

armcknight:
It was brought up that these are disruptive to the test processes, and I actually agree with that as of previously tried to remove as many crashes as possible from the tests. Remove these and replace them with error logs, as was done in #5363 

#skip-changelog


itaybre (commented):
To be fair, I am not sure if this is something I would like to remove.
The real reason behind why these tests fail is because we can't setup the context properly.

armcknight:
@itaybre that is true, but again see my point about taking down the test suite with a crash due to abort. 

the test should still fail if we hit these situations, but we should be able to run the whole suite and see all test failures, instead of having to rerun every time we fix a crasher. that is much more time (and morale) consuming. especially in CI, if for some reason something happens there but not locally.

philipphofmann (approved):
Thanks for following up and removing these @armcknight 💯

PR #5632, opened 2025-07-14
New runloop observer

noahsmartin:
WIP because this has no tests yet, just opening it to let CI run

#skip-changelog


Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,302 @@
...
+        self.debugImageCache = debugImageCache
+        self.fileManager = fileManager
+        self.crashWrapper = crashWrapper
+        self.lastFrameTime = 0
+        self.minHangTime = minHangTime
+#if canImport(UIKit) && !SENTRY_NO_UIKIT && !os(visionOS) && !os(watchOS)
+        var maxFPS = 60.0
+            if #available(iOS 13.0, tvOS 13.0, *) {
+            let window = UIApplication.shared.connectedScenes.flatMap { ($0 as? UIWindowScene)?.windows ?? [] }.first { $0.isKeyWindow }
+            maxFPS = Double(window?.screen.maximumFramesPerSecond ?? 60)
```
trevor-e, line 48: I believe the display rate can change dynamically in a few cases like when displaying video. Not seeing any way to subscribe to that change though.

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,302 @@
...
+    
+    // MARK: Background queue
+    
+    private var blockingDuration: TimeInterval?
+    private var threads: [SentryThread]?
+    
+    private func waitForHang(semaphore: DispatchSemaphore, started: TimeInterval, isStarting: Bool) {
+        dispatchPrecondition(condition: .onQueue(queue))
+        
+        let timeout = DispatchTime.now() + DispatchTimeInterval.milliseconds(Int((expectedFrameDuration + thresholdForFrameStacktrace) * 1_000))
```
trevor-e, line 206: nit: for easier readability I'd expect `thresholdForFrameStacktraceDurationMs = expectedFrameDuration *1.5 * 1000` on assignment instead of calculating down here.

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,302 @@
...
+    private func continueHang(started: TimeInterval, isStarting: Bool) {
+        dispatchPrecondition(condition: .onQueue(queue))
+
+        if isStarting {
+            // A hang lasts a while, we show the stacktrace when it was first detected
+            threads = threadInspector.getCurrentThreadsWithStackTrace()
+            threads?.forEach { $0.current = false }
+            threads?[0].current = true
+        }
+        let duration = dateProvider.systemUptime() - started
```
trevor-e, line 228: Assuming this works fine for the first frame edge case?

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,302 @@
...
+            semaphore.signal()
+            print("[HANG] Timeout, hang detected")
+            continueHang(started: started, isStarting: isStarting)
+            waitForHang(semaphore: semaphore, started: started, isStarting: false)
+        case .success:
+            break
+        }
+    }
+
+    private func continueHang(started: TimeInterval, isStarting: Bool) {
```
trevor-e, line 219: How does this work if the app is backgrounded?

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,302 @@
...
+    private var semaphore = DispatchSemaphore(value: 0)
+    private var lastFrameTime: TimeInterval
+    private var running = false
+    private var frameStatistics = [(startTime: TimeInterval, delayTime: TimeInterval)]()
+    // Keeps track of how long the current hang has been running for
+    // Set to nil after the current hang ends
+    private var maxHangTime: TimeInterval?
+    
+    func start() {
+        let observer = CFRunLoopObserverCreateWithHandler(nil, CFRunLoopActivity.beforeWaiting.rawValue | CFRunLoopActivity.afterWaiting.rawValue | CFRunLoopActivity.beforeSources.rawValue, true, CFIndex(INT_MAX)) { [weak self] _, activity in
```
trevor-e, line 79: One weird edge-case that ChatGPT claims is possible is re-entry into the current runloop via `CFRunLoopPerformBlock` that could result in false negatives, I guess as part of keyboard/modal interactions? I'm not very familiar with this API though so not sure if that's a real or hallucination.

Sources/Swift/HangTracker.swift
noahsmartin: yeah I didn't see it either. I figured using maximum at least would mean we would detect any time slow code is preventing the app from running at fastest possible. I think other things can cause the OS to decide to slow it down like low power mode. But you probably want to still know when frames are going to slow even if the OS chooses not to render them all

Sources/Swift/HangTracker.swift
noahsmartin: I've never seen a case where the runloop doesn't get to finish an iteration. Even if the app is backgrounded your code isn't suspended instantly (for example NSNotifications get called on the main thread). If the main thread was suspended and never got a chance to resume then we would use the event stored to disk and report a fatal exception

Sources/Swift/HangTracker.swift
noahsmartin: I'll test out keyboards specifically

Sources/Swift/HangTracker.swift
trevor-e: Gotcha, I was more thinking about whether the `lastFrameTime` value could get out of sync at all and report a false positive since that seems like it might have happened with the current implementation, but I'm guessing you're right that the runloop is allowed to finish its iteration.

Sources/Swift/HangTracker.swift
philippsuess: The CADisplayLink docs actually mention this a bit https://developer.apple.com/documentation/quartzcore/cadisplaylink#Preferred-and-Actual-Frame-Rates.

You run into a risk of falsely reporting slow frames if the frame rate changes. That's why we calculate the actual frame rate here: https://github.com/getsentry/sentry-cocoa/blob/9a17767807611439455149cb24fdfc9e723c41f4/Sources/Sentry/SentryFramesTracker.m#L186-L197

philprime:
@noahsmartin is this superseded by #6237? If yes, let's close this PR.

PR #5638, opened 2025-07-15
feat(v9): Remove segment property

noahsmartin:
Remove deprecated SentryUser segment property in v9

#skip-changelog


Sources/Sentry/SentryNetworkTracker.m
```diff
@@ -230,14 +230,18 @@ - (void)urlSessionTaskResume:(NSURLSessionTask *)sessionTask
...
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    segment = SentrySDK.currentHub.scope.userObject.segment;
+#    pragma clang diagnostic pop
+#endif // !SDK_V9
     SentryTraceContext *traceContext =
         [[SentryTraceContext alloc] initWithTraceId:propagationContext.traceId
                                             options:SentrySDK.currentHub.client.options
-                                        userSegment:SentrySDK.currentHub.scope.userObject.segment
+                                        userSegment:segment
```
itaybre, line 246: Shouldn't we remove the parameter from `SentryTraceContext` too?

noahsmartin: Left that out of this PR intentionally because SentryTraceContext is not part of the public api so it wouldn’t need to be a v9 change

noahsmartin: Ended up putting cursor on this and it found a bunch of extra stuff to remove!

CHANGELOG-v9.md
```diff
@@ -14,3 +14,4 @@ Removes deprecated getStoreEndpoint (#5591)
 Removes deprecated useSpan function (#5591)
 Removes deprecated SentryDebugImageProvider class (#5598)
 Makes app hang tracking V2 the default and removes the option to enable/disable it (#5615)
+Removes segment property on SentryUser, SentryBaggage, and SentryTraceContext (#5638)
```
armcknight, line 11: I think anything in this section starting with "Removes" should be under "breaking changes".

The format of this changelog is slightly out of spec. We follow keepachangelog, which specifies the h1 header be named "Changelog", the h2 headers are versions, and h3 headers are things like "breaking changes", "features", "fixes", "improvements". See our current shipping changelog: https://github.com/getsentry/sentry-cocoa/blob/a7655b5e00da79dd91e17bc09eeaf229d556d68a/CHANGELOG.md?plain=1#L1-L21

Sources/Sentry/Public/SentryTraceContext.h
```diff
@@ -111,16 +117,28 @@ NS_SWIFT_NAME(TraceContext)
...
  *  @param options The current active options.
+ */
+#else
+/**
+ * Initializes a SentryTraceContext with data from a traceID, options and userSegment.
+ *
+ *  @param traceId The current tracer.
+ *  @param options The current active options.
+ *
  *  @param userSegment You can retrieve this usually from the `scope.userObject.segment`.
```
armcknight: A couple fixes here: 
- `ID` -> `Id` (I don't like that capitalization, but that's what we currently have)
- need to remove the headerdoc about segments from the v9 version
- need to add headerdoc about session replay to both (i know this isn't due to your changes but we might as well do it

```suggestion
 * Initializes a SentryTraceContext with data from a traceId, options and replayId.
 *
 *  @param traceId The current tracer.
 *  @param options The current active options.
 *  @param replayId The current session replay.
 */
#else
/**
 * Initializes a SentryTraceContext with data from a traceId, options, userSegment and replayId.
 *
 *  @param traceId The current tracer.
 *  @param options The current active options.
 *  @param userSegment You can retrieve this usually from the `scope.userObject.segment`.
 *  @param replayId The current session replay.
```

armcknight (approved):
Nice 👍🏻

CHANGELOG-v9.md
noahsmartin: my plan was just to copy the Fixes/Improvements sections into the real changelog when we were ready to release, but didn't realize Breaking Changes is something that would get copied too, will update

Sources/Sentry/Public/SentryTraceContext.h
noahsmartin: Happy to make the change, just fwiw this was just a copy/paste of the old comment with the one deleted param removed, was trying to avoid changing things like capitalization from the hold header, but we can make it different in v9

PR #5641, opened 2025-07-15
fix: Delete undefined SentryMetricsAPI

noahsmartin:
This type was only forward declared and never used, we can delete it

#skip-changelog

### DIFF
--- a/Sources/Sentry/Public/SentryHub.h
+++ b/Sources/Sentry/Public/SentryHub.h
@@ -12,7 +12,6 @@
 @class SentryEvent;
 @class SentryFeedback;
 @class SentryId;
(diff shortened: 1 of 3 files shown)


noahsmartin:
@itaybre I don't think it's public API, you can see in the api stability check. If you tried to access this property you'd get a compiler errors because the type is not defined (only forward declared) so we can remove it now with no impact on public API

itaybre (approved):
You are right, it is only a forward declaration.

LGTM.
But we should update the docs: https://github.com/getsentry/sentry-docs/blob/master/docs/platforms/apple/common/metrics/index.mdx

PR #5654, opened 2025-07-17
Use latest version of appium

itaybre:
Test latest version of appium


philippsuess:
@itaybre, please add a comment why you closed this. Did you have problems? If yes, which ones?

itaybre:
This was my first test when checking SauceLabs incident.
Since this wasn't the issue, it made no sense changing this

PR #5663, opened 2025-07-18
fix: Support V9 in SampleShared

noahsmartin:
Adding the compiler flag for SentrySampleShared target

#skip-changelog


armcknight:
See also https://github.com/getsentry/sentry-cocoa/pull/5664

PR #5666, opened 2025-07-18
feat(SPM): Build full SDK with SPM

noahsmartin:
Support building the full SDK with SPM, including all features available on iOS.

This just required fixing one more issue related to `sessionReplay` being defined in Swift but `SentryOptions` in ObjC so it wasn't visible to Swift code in SPM. I'll have a longer term fix for this by making a Swift interface for SentryOptions, but that is blocked by a few other PRs right now so this change is just a smaller patch to get CI building for iOS, ensuring we don't accidentally break anything with the SPM build.

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,8 +112,9 @@ jobs:
     if: startsWith(github.ref, 'refs/heads/release/') == false
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


noahsmartin:
Although it builds, the public API isn't quite finished yet (some things are missing/changed) So I was thinking I would fix that first before updating the changelog

PR #5671, opened 2025-07-21
meta: log complete raw envelope contents in debug

armcknight:
I had a need to debug the complete raw contents of an envelope. AFAICT we logged a few things like headers, but not the envelope in its entirety. It produces large output so I also constrained it to just DEBUG builds. In the future it'd be nice to be able to selectively enable/disable this type of logging like I proposed in https://github.com/getsentry/sentry-cocoa/issues/4020

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryEnvelope.m
+++ b/Sources/Sentry/SentryEnvelope.m
@@ -1,6 +1,7 @@
 #import "SentryAttachment.h"
 #import "SentryBreadcrumb.h"
 #import "SentryClientReport.h"
+#import "SentryDateUtils.h"
 #import "SentryEnvelope+Private.h"
 #import "SentryEnvelopeAttachmentHeader.h"
 #import "SentryEnvelopeItemHeader.h"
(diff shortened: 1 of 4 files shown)


philipphofmann (commented):
We store envelopes to disk before sending them. I usually get the envelope cache path and directly look at the envelope contents on the file system with Cursor. This has the advantage of looking at the envelope close to how it will be sent. As the envelope contents can be long, I don't really find printing the on the console useful, as I anyways would copy and paste the data into Cursor.

If we want to have some way to log the envelope contents, I would rather hook into the HTTPTransport and simply print its content here https://github.com/getsentry/sentry-cocoa/blob/7c7ac56f381bb42b528ad2126860309bd98f0803/Sources/Sentry/SentryHttpTransport.m#L341

or here 

https://github.com/getsentry/sentry-cocoa/blob/7c7ac56f381bb42b528ad2126860309bd98f0803/Sources/Sentry/SentryNSURLRequestBuilder.m#L15-L16

Wouldn't that be sufficient? It would also be less code. The logEnvelopeContents implementation can easily deviate from the actual envelope serialization.

armcknight:
@philipphofmann indeed, you are right about implementation drift. i see that we do log the path to a cached envelope, so i could look at the file, that's even better anyways. 

i see we have an alternative path for envelopes that can't be cashed to disk for whatever reason that are sent directly from an in memory data structure… It wouldn't work for these, but I suppose if I see this happening in the future when I need to look at the file on disc, I can deal with it then!

philippsuess:
> i see we have an alternative path for envelopes that can't be cashed to disk for whatever reason that are sent directly from an in memory data structure…

Where do we have such envelopes? I'm only aware that the SpoltlightHttpTransport doesn't use the disk cache, but apart from that every envelope going through the normal SentryHttpTransport should get stored to disk first. If not, it's a bug or something new that I missed.

armcknight:
It just looks like a piece of defensive programming, since the call to the file manager can technically return an error result: https://github.com/getsentry/sentry-cocoa/blob/6cd7f860febec8af51fc19071f83da4dfc86a284/Sources/Sentry/SentryHttpTransport.m#L139-L145

PR #5672, opened 2025-07-21
test: reorganize networking feature flags and add another one

armcknight:
I needed to disable stuff for a test and noticed I had never provided a feature flag and override enum case for SentryOptions.enableCaptureFailedRequests, so i added that and reorganized the networking options under their own enum

#skip-changelog


philipphofmann (approved):
LGTM

PR #5674, opened 2025-07-21
ref(SPM): Create SentrySDKInternal

noahsmartin:
This renamed SentrySDK to SentrySDKInternal, and then creates a new copy of SentrySDK.h that isn't renamed. The implementation of this new SentrySDK is just a wrapper on SentrySDKInternal.

The reason for this change is to prepare to implement SentrySDK in Swift, which can be done with very little changes by having the swift version just call the ObjC version (SentrySDKInternal). I'll do this in another PR (draft here https://github.com/getsentry/sentry-cocoa/pull/5653) I just wanted the routine renaming in 100 files to be done separate from the actual language conversion so it can be more carefully reviewed.

This is done for supporting the SPM API

#skip-changelog

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift/Tools/SentryExposure.h
+++ b/Samples/iOS-Swift/iOS-Swift/Tools/SentryExposure.h
@@ -1,9 +1,23 @@
 #import <Foundation/Foundation.h>
 #import <UIKit/UIKit.h>
 
+@class SentryHub;
+
+NS_ASSUME_NONNULL_BEGIN
+
 @interface SentryBreadcrumbTracker : NSObject
 
-+ (NSDictionary *)extractDataFromView:(UIView *)view
-          withAccessibilityIdentifier:(BOOL)includeIdentifier;
++ (nullable NSDictionary *)extractDataFromView:(UIView *)view
+                   withAccessibilityIdentifier:(BOOL)includeIdentifier;
 
 @end
+
+@interface SentrySDKInternal : NSObject
+
++ (nullable NSArray<NSString *> *)relevantViewControllersNames;
+
++ (SentryHub *)currentHub;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Sources/Sentry/SentrySDKWrapper.m
+++ b/Sources/Sentry/SentrySDKWrapper.m
@@ -1,0 +1,242 @@
+#import "SentryProfilingConditionals.h"
+#import "SentrySDK.h"
+#import "SentrySDKInternal.h"
+
+@implementation SentrySDK
+
++ (id<SentrySpan>)span
+{
+    return [SentrySDKInternal span];
+}
+
++ (BOOL)isEnabled
+{
+    return [SentrySDKInternal isEnabled];
+}
+
+#if SENTRY_TARGET_REPLAY_SUPPORTED
++ (SentryReplayApi *)replay
+{
+    return [SentrySDKInternal replay];
+}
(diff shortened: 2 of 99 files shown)


Samples/iOS-Swift/iOS-Swift/Tools/SentryExposure.h
```diff
@@ -1,9 +1,23 @@
...
 @interface SentryBreadcrumbTracker : NSObject
 
-+ (NSDictionary *)extractDataFromView:(UIView *)view
-          withAccessibilityIdentifier:(BOOL)includeIdentifier;
++ (nullable NSDictionary *)extractDataFromView:(UIView *)view
+                   withAccessibilityIdentifier:(BOOL)includeIdentifier;
+
+@end
+
+@interface SentrySDKInternal : NSObject
```
itaybre, line 15: I would prefer to have this class declared in an independent file

itaybre: In fact, this seems to be part of the test app, why is this needed?

Sources/Sentry/SentrySDKWrapper.m
```diff
@@ -0,0 +1,242 @@
+#import "SentryProfilingConditionals.h"
+#import "SentrySDK.h"
+#import "SentrySDKInternal.h"
+
+@implementation SentrySDK
```
itaybre, line 5: Seems like this file name is `SentrySDKWrapper`, should it be `SentrySDK.h`?

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -95,7 +95,7 @@ class SentrySDKTests: XCTestCase {
 
         givenSdkWithHubButNoClient()
 
-        if let autoSessionTracking = SentrySDK.currentHub().installedIntegrations().first(where: { it in
+        if let autoSessionTracking = SentrySDKInternal.currentHub().installedIntegrations().first(where: { it in
```
itaybre, line 97: Somewhat redundant, but we should test SentrySDKInternal and SentrySDK (public API) in different tests

Samples/iOS-Swift/iOS-Swift/Tools/SentryExposure.h
noahsmartin: Because the test app calls some private functions that are not part of the public API, this file "SentryExposure" seemed to be the place to "expose" these private classes to the SDK, that's what the other classes in this file were doing

Sources/Sentry/SentrySDKWrapper.m
noahsmartin: I only made it named this so that the git diff would pick up the other file that went from SentrySDK.m to SentrySDKInternal.m as a rename, otherwise it would think I changed every line in the file and this PR would look way bigger than it is. I can rename this file in a follow up PR, but I think it should stay named something other than `SentrySDK.m` for now or this PR will be really hard to review

Tests/SentryTests/SentrySDK+Tests.h
```diff
@@ -1,16 +1,15 @@
 #import "SentryDefines.h"
-#import "SentrySDK.h"
 
 @class SentryEnvelope;
 @class SentryHub;
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface SentrySDK ()
+@interface SentrySDKInternal ()
```
itaybre, line 8: You should rename this file to `SentrySDKInternal+Tests.h`

itaybre (approved):
LGTM, just a small comment

PR #5676, opened 2025-07-21
ref(v9): Make SentryHub private

noahsmartin:
Testing this out

#skip-changelog


philippsuess:
Out of curiosity, what drives you to make the hub private?

The hub is part of the now [deprecated unified API](https://develop.sentry.dev/sdk/miscellaneous/unified-api/#hub) and it's supposed to be public. We're working on merging the hub and the scope https://develop.sentry.dev/sdk/miscellaneous/hub_and_scope_refactoring/. The plan there is to basically get rid of the hub and only have scope, client, and transport.

noahsmartin:
@philipphofmann just because it seemed like it was unused, I didn't see what API users of the SDK would call that would use `SentryHub` so was trying to make it private to see what (if any) of the build would fail which would guide me to where it's used

philippsuess:
@noahsmartin it would be great to add a short comment explaining why we closed this PR.

PR #5679, opened 2025-07-22
Fix v9 typo

noahsmartin:
Fix typo in V9 check. At the time this code was added we didn't have the API stability check so this was missed, now that we have the check we can confirm the API is removed because the sdk_api_v9.json file updated

#skip-changelog


philipphofmann (approved):
LGTM

Issue #5554, opened 2025-07-02
Delayed AppHang Reports When App Moves to the Background


philipphofmann:
### Description

The ANRTrackingIntegration ignores app hangs when the trackers report them while the app is in the background https://github.com/getsentry/sentry-cocoa/blob/bce9765eba9af4af7ad81ff7feb3684aed3ed684/Sources/Sentry/SentryANRTrackingIntegration.m#L116-L121

Also the ANRTrackerV2 ignores app hangs when the app is in the background: https://github.com/getsentry/sentry-cocoa/blob/bce9765eba9af4af7ad81ff7feb3684aed3ed684/Sources/Sentry/SentryANRTrackerV2.m#L108C33-L111

There might be some edge cases when the app is hanging while moving to the background, and it moves to the foreground again, for which the SDK then reports a delayed app hang. So it looks like the SDK reported an app hang for when the app was in the background.

A customer reported this

![Image](https://github.com/user-attachments/assets/a858dff1-496b-4e96-a1e8-38500a010c68)

philippsuess:
We could solve this by still reporting apps in the background when they happened while the app was in the foreground, but we also need to ensure to ditch the move to background breadcrumb and maybe we need to set a different timestamp for the event.

hovox:
hi @philipphofmann 

When app goes into background `pause` method is called inside `SentryFramesTracker.m` but `previousFrameTimestamp` property is not being reset there and as I understood frame duration is calculated by this line - [`CFTimeInterval frameDuration = thisFrameTimestamp - self.previousFrameTimestamp;`](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentryFramesTracker.m#L219).  When app goes into foreground after some time `thisFrameTimestamp - self.previousFrameTimestamp` value shows the time app was in the background. I have not tested my assumptions, sorry if I'm totally tally wrong.

philippsuess:
Thanks for pointing that out. Yes, it could be that we miss an edge case there and add wrong frame statistics. We should double-check your hint when tackling this issue.

PR #5681, opened 2025-07-22
feat(ANR): Ignore frame delays when going into background

itaybre:
The SDK was reporting hangs when coming back to foreground.

Fixes: #5554


Sources/Sentry/include/SentryFramesTracker+Test.h
```diff
@@ -0,0 +1,12 @@
+#if SENTRY_HAS_UIKIT
+
+#    import "SentryFramesTracker.h"
+
+@interface SentryFramesTracker ()
+
+- (void)pause;
+- (void)unpause;
+
+@end
```
philippsuess: `m`: You should be able to test this functionality by using the `TestNSNotificationCenterWrapper` and using the notifications `didBecomeActive` and `willResignActive` which call pause and unpause. So I think we can remove this or am I missing something?

https://github.com/getsentry/sentry-cocoa/blob/6b57184fb4aea6c8c6ad2765b3fe5f2a87f73710/Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift#L682-L689

Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
```diff
@@ -710,6 +710,156 @@ class SentryFramesTrackerTests: XCTestCase {
         XCTAssertEqual(sut.isRunning, true)
     }
     
+    func testUnpause_ResetsPreviousFrameTimestamp_ToAvoidFalseAppHangDetection() throws {
```
philippsuess: `l`: The frames tracker doesn't know the app hang detection exists. We validate that we don't record wrong frame statistics. The test names should reflect that.

Sources/Sentry/SentryFramesTracker.m
```diff
@@ -166,7 +166,9 @@ - (void)unpause
     }
 
     _isRunning = YES;
-
+    // Reset the previous frame timestamp to avoid false app hang detection
+    // This also means we will not detect a hang if the app stuck while the tracker was paused.
```
philippsuess: `m`: The frames tracker doesn't know about app hangs. This here is solely about wrong frame statistics, which has the positive side effect of getting rid of wrong app hangs as well, because app hangs v2 uses the frame statistics.

Tests/SentryTests/Integrations/Performance/FramesTracking/SentryFramesTrackerTests.swift
```diff
@@ -710,6 +710,156 @@ class SentryFramesTrackerTests: XCTestCase {
...
+        // Simulate app going to background
+        fixture.notificationCenter.post(Notification(name: CrossPlatformApplication.willResignActiveNotification))
+        XCTAssertFalse(sut.isRunning)
+        
+        // Simulate a long time in background
+        fixture.dateProvider.advance(by: 10.0)
+        
+        // Simulate app coming to foreground
+        fixture.notificationCenter.post(Notification(name: CrossPlatformApplication.didBecomeActiveNotification))
+        XCTAssertTrue(sut.isRunning)
```
philippsuess, line 778: `l`: Yes this is how we indirectly call pause and unpause. We can merge this test case with `testUnpause_AfterLongPause_DoesNotDetectHang`. We only need one. They are basically testing the same thing.

CHANGELOG.md
```diff
@@ -17,8 +17,8 @@
 
 - Fix video replay crashes due to video writer inputs not marked as finished on cancellation (#5608)
 - Fix wrong flush timeout (#5565). When flush timed out before the SDK finished sending data, it always blocked the full flush timeout the next time being called. This is fixed now.
-
 - Launch profiling now respects original configured options if they change on the next launch (#5417)
+- Ignore frame delays when going into background (#5681)
```
philippsuess: `m`: This PR fixes getting wrong frame statistics when the app goes to the background and again to the foreground which has the positive side effect of fixing wrongly reported app hangs when the app was in the background only for app hangs v2, because app hangs v2 uses the frame statistics.
The changelog entry and the PR title should reflect that impact.

philipphofmann (approved):
Good catch @itaybre. I can reproduce wrong frozen frames with the iOS-Swift app from a testflight build by simply putting the app into background for a second, opening it again and then checking the frozen frame number. It would be great if you also did such a test to validate that this problem is now gone. If it is, we can merge this. Or how did you validate that the problem doesn't persist any more with manual tests?

itaybre:
> Good catch @itaybre. I can reproduce wrong frozen frames with the iOS-Swift app from a testflight build by simply putting the app into background for a second, opening it again and then checking the frozen frame number. It would be great if you also did such a test to validate that this problem is now gone. If it is, we can merge this. Or how did you validate that the problem doesn't persist any more with manual tests?
> 
> In case you didn't know yet, you can trigger test flights build manually here https://github.com/getsentry/sentry-cocoa/actions/workflows/testflight.yml
> 
> Screenshot 2025-07-23 at 16 31 52
> And then you need to find the build in testflight with this number Screenshot 2025-07-23 at 16 33 16

Yes, I validated it in testflight

itaybre:
Prior to the change:

https://github.com/user-attachments/assets/8c667a68-834f-4639-b83f-784156e3f271

After the change:

https://github.com/user-attachments/assets/44a16523-b008-4706-9fea-5812fe3dabe9

Issue #5648, opened 2025-07-17
SentryUserFeedbackWidget prevents usage of SDK in app extensions


AgentFeeble:
### Platform

iOS

### Environment

Other

### Installed

CocoaPods

### Version

8.53.2

### Xcode Version

16.4

### Did it work on previous versions?

Yes, it was working on version 8.53.1

### Steps to Reproduce

1. Add the sentry-cocoa SDK to an iOS project that includes an app extension target.
2. Attempt to import and use the SDK from the app extension.
3. Build the extension target.

### Expected Result

The app extension should be able to import and use parts of the Sentry SDK that are safe for extensions, while any incompatible APIs (e.g. those using `UIApplication.shared`) should be unavailable.

### Actual Result

The build fails with the following compile-time error:

`'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.`

This occurs because some SDK classes use `UIApplication.shared` without being marked as unavailable for app extensions.

Specifically, the issue stems from:
- `SentryUserFeedbackWidget` (`SentryUserFeedbackWidget.swift`)
- `SentryUserFeedbackIntegrationDriver` and all its extensions (`SentryUserFeedbackIntegrationDriver.swift`)

**Suggested Fix**

Adding this annotation to the affected classes and extensions should resolve the issue:

```swift
@available(iOS 13.0, *)
@available(iOSApplicationExtension, unavailable)
```

This prevents incompatible classes from being used in app extensions, while allowing the rest of the SDK to compile and function normally.

Thanks for your work on Sentry!

### Are you willing to submit a PR?

I can submit a pull request if needed

philippsuess:
@armcknight, we released https://github.com/getsentry/sentry-cocoa/pull/5223 in 8.53.2, which touched SentryUserFeedbackWidget. Could you maybe have a look?

armcknight:
Hi @AgentFeeble , thanks for reporting. Could you let me know what kind of extension you're building? We currently only have an AppClip extension in our test harnesses. I wonder if this only happens in certain ones.

armcknight:
Actually, nevermind, this is extension API based, not UIKit based like I was initially thinking. We used to have this annotation but thought we could remove it. We'll need to investigate what we can do.

liamjones:
Sorry, ignore my deleted comment if you had it emailed to you. I had a workaround for https://github.com/getsentry/sentry-react-native/issues/3908 I hadn't removed.

[...]

PR #5682, opened 2025-07-22
fix: mark feedback API as unavailable to app extensions

armcknight:
For https://github.com/getsentry/sentry-cocoa/issues/5648, fix compilation in extensions where we use UIApplication API, which is unavailable in certain extension contexts.

### DIFF
--- a/Sources/Sentry/SentryOptions.m
+++ b/Sources/Sentry/SentryOptions.m
@@ -78,8 +78,10 @@ NSString *const kSentryDefaultEnvironment = @"production";
         [SentrySwiftAsyncIntegration class], nil];
 
 #if TARGET_OS_IOS && SENTRY_HAS_UIKIT
-    if (@available(iOS 13.0, *)) {
-        [defaultIntegrations addObject:[SentryUserFeedbackIntegration class]];
+    if (@available(iOSApplicationExtension 13.0, *)) {
+        if (@available(iOS 13.0, *)) {
+            [defaultIntegrations addObject:NSClassFromString(@"SentryUserFeedbackIntegration")];
+        }
     }
 #endif // TARGET_OS_IOS && SENTRY_HAS_UIKIT
 
--- a/Sources/Sentry/Public/SentryFeedbackAPI.h
+++ b/Sources/Sentry/Public/SentryFeedbackAPI.h
@@ -12,6 +12,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+API_UNAVAILABLE(iOSApplicationExtension)
 @interface SentryFeedbackAPI : NSObject
 
 /**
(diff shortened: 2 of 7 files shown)


Sources/Sentry/SentryOptions.m
```diff
@@ -78,8 +78,10 @@ @implementation SentryOptions {
         [SentrySwiftAsyncIntegration class], nil];
 
 #if TARGET_OS_IOS && SENTRY_HAS_UIKIT
-    if (@available(iOS 13.0, *)) {
-        [defaultIntegrations addObject:[SentryUserFeedbackIntegration class]];
+    if (@available(iOSApplicationExtension 13.0, *)) {
```
noahsmartin: I think these two can be combined into one if statement:

```
if (@available( iOS 13.0, iOSApplicationExtension 13.0, *) { ... }
```

noahsmartin (approved):

CHANGELOG.md
```diff
@@ -17,8 +17,8 @@
 
 - Fix video replay crashes due to video writer inputs not marked as finished on cancellation (#5608)
 - Fix wrong flush timeout (#5565). When flush timed out before the SDK finished sending data, it always blocked the full flush timeout the next time being called. This is fixed now.
-
 - Launch profiling now respects original configured options if they change on the next launch (#5417)
+- Compilation in app extensions (#5682)
```
philippsuess: `m`: The changelog entry could be a bit more descriptive so it's clearer what exactly it fixes.
```suggestion
- Build error in app extensions (#5682)
```

or also `compilation error` would work.

Sources/Sentry/SentryOptions.m
```diff
@@ -78,8 +78,10 @@ @implementation SentryOptions {
...
+        if (@available(iOS 13.0, *)) {
+            [defaultIntegrations addObject:NSClassFromString(@"SentryUserFeedbackIntegration")];
```
philippsuess: `m`: As the `SentryUserFeedbackIntegration` is marked as `API_AVAILABLE(ios(13.0))`, we should be able to use
```suggestion
            [defaultIntegrations addObject:[SentryUserFeedbackIntegration class]];
```

armcknight:
@philipphofmann I'm working on the validation step. ETA: I just pushed up what I'd already done so far to https://github.com/getsentry/sentry-cocoa/pull/5696, but there are still some challenges associated with that.

Sources/Sentry/SentryOptions.m
armcknight: You're right, this only failed when I still had APPLICATION_EXTENSION_API_ONLY = YES for testing. Thanks!

balavor:
Thank you for adding this. I believe that in the long term, it is entirely valid to continue using flags to mark features as unavailable for app extensions where they don't make sense or rely on `UIApplication`.

PR #5683, opened 2025-07-23
feat(SPM) Convert SentrySDK to Swift

noahsmartin:
This converts SentrySDK to Swift. It's a pretty straightforward conversion since I had already refactored SentrySDKInternal to contain all the implementation, the biggest changes here are around SPM support. In SPM you can't have Swift code that calls an ObjC API that uses a Swift type. This means the Swift/ObjC bridge needs to only contains ObjC types. That leads to 3 changes in this PR:

1. SentryLogger is handled directly in SentrySDK.swift and removed from SentrySDKInternal. SentryLogger is defined in Swift so the Swift code could not access it through SentrySDKInternal making it necessary to convert to Swift.
2. Methods returning SentryId in SentrySDKInternal have been changed to return NSString. The Swift code creates the SentryId from the String. This is necessary because SentryId is written in Swift.
3. The capture(userFeedback:) method of SentrySDKInternal has been changed to take the already serialized user feedback data. This is necessary because the UserFeedback type is written in Swift.

Other than those 3 changes this PR is just copying the interface to Swift. The number of lines changes is big because the sdk_api json files change but the changes are just internal to how the file represents Swift vs. ObjC definitions, not breaking changes.

#skip-changelog


armcknight (approved):
This looks great, thanks for taking it on! Feel free to merge when CI is happy 🤖

PR #5684, opened 2025-07-23
test: Extract flush integration tests to own file

philipphofmann:
Put all the flush integration tests into an extra file / XCTestCase to minimize potential side effects and reduce flakiness.

This will make it easier to run them in an extra test plan.

The unit tests that are failing are not related to this PR and should be fixed with https://github.com/getsentry/sentry-cocoa/pull/5687.

#skip-changelog


philippsuess:
Converted to draft because many tests are failing.

Tests/SentryTests/Networking/TestRequestManager.swift
```diff
@@ -9,7 +9,7 @@ public class TestRequestManager: NSObject, RequestManager {
     
     var requests = Invocations<URLRequest>()
     
-    private let queue = DispatchQueue(label: "TestRequestManager", qos: .background, attributes: [])
+    private let queue = DispatchQueue(label: "TestRequestManager")
```
noahsmartin, line 12: Just wanted to make sure this was intentional since it doesn't seem related to moving the tests into a separate file

philippsuess: Thanks for double checking. Yes it's intentional, should have added a comment to this PR. With the `.background` qos we have more flakiness in the tests due to timeouts. Using the default qos fixes that.

PR #5687, opened 2025-07-23
test: Remove state in MainThreadTestIntegration

philipphofmann:
Remove the static global state in MainThreadTestIntegration. Also make installedInTheMainThread readonly for public access.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/SentrySDKInternalTests.swift
+++ b/Tests/SentryTests/SentrySDKInternalTests.swift
@@ -509,16 +509,21 @@ class SentrySDKInternalTests: XCTestCase {
 
     /// Ensure to start the UIDeviceWrapper before initializing the hub, so enrich scope sets the correct OS version.
     func testStartSDK_ScopeContextContainsOSVersion() throws {
-        let expectation = expectation(description: "MainThreadTestIntegration install called")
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/SentrySDKInternalTests.swift
```diff
@@ -1045,19 +1055,26 @@ class SentrySDKWithSetupTests: XCTestCase {
...
-    static var expectation: XCTestExpectation?
+    public let expectation = XCTestExpectation(description: "MainThreadTestIntegration installed")
 
-    public var installedInTheMainThread = false
+    private var installedInTheMainThread = false
 
     public func install(with options: Options) -> Bool {
         print("[Sentry] [TEST] [\(#file):\(#line) starting install.")
+
         installedInTheMainThread = Thread.isMainThread
```
noahsmartin: I would write this as `dispatchPrecondition(condition: .onQueue(.main))` technically I think it could be on the main thread but not on the main queue, such as if you `dispatch_sync` from the main queue. If you used this you also wouldn't need the `wasInstalledOnTheMainThread()` function at all

noahsmartin (approved):

Issue #5159, opened 2025-04-30
iOS User Feedback is sampled by sampleRate / beforeSend causing JSON conversion error


kerenkhatiwada:
### Platform

iOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

8.49.0

### Xcode Version

16.3

### Steps to Reproduce

1. Set sampleRate to a value that samples out some events (e.g., 0.5), or configure beforeSend to return nil 
2. Send a user feedback event
```
struct UserInfo {
    var email: String?
}

let userInfo = UserInfo(email: nil)

let emailString = String(userInfo.email ?? "newanonymous@example.com")

let feedback = SentryFeedback(
    message: "messageString",
    name: "nameString",
    email: emailString,
    source: .widget
)

SentrySDK.capture(feedback: feedback)
```

3. Observe that the feedback event is sampled and not sent, and the following logs are printed:
```
[Sentry] [debug] [timeIntervalSince1970:1746034783.3588839] [SentryClient:686] Event got sampled, will not send the event
[Sentry] [error] [timeIntervalSince1970:1746034783.358952] [SentrySerialization:22] Dictionary is not a valid JSON object.
[Sentry] [debug] [timeIntervalSince1970:1746034783.359066] [SentryEvent:76] Serialized event: {
"event_id" = 604741ad8cb9447e8008b165a7f9d3ca;
level = warning;
message = {
formatted = "JSON conversion error for event with message: '(null)'";
};
platform = cocoa;
timestamp = "1746034783.358967";
}
```
as a result, [a JSON conversion error](https://org2-w7.sentry.io/issues/6578315483/?project=4509170522193920&query=is%3Aunresolved&referrer=issue-stream&sort=date&stream_index=0) is reported in Sentry 

[ZD ticket](https://sentry.zendesk.com/agent/tickets/151406)

### Expected Result

User feedback should not be subject to sampling or beforeSend filtering

### Actual Result

User feedback is subject to sampling and beforeSend filtering, which causes a JSON conversion error

armcknight:
Thanks for reporting @kerenkhatiwada , I'm investigating. Just preliminarily and for future reference by myself, here are some pertinent dev docs:

- [there should be no sample rate](https://develop.sentry.dev/sdk/telemetry/feedbacks/#dropped-feedback-reports)
[...]

PR #5692, opened 2025-07-23
fix: user feedback events should not be sampled or subject to `beforeSend`

armcknight:
Fixes https://github.com/getsentry/sentry-cocoa/issues/5159#issuecomment-3070392706

Added unit tests that failed before the fix and passed afterwards. Also added a new SDK override for configuring the events sample rate (and reorganized that and the preexisting override to reject events in `beforeSend` under a new `Events` override enum), and set that as a tag on the scope, and verified that with both overrides configured to reproduce the setup, the user feedback was still received: https://sentry-sdks.sentry.io/issues/feedback/?feedbackSlug=sentry-cocoa%3A6765771740&mailbox=unresolved&project=5428557&referrer=feedback_list_page&statsPeriod=1h
image

### DIFF
--- a/Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
+++ b/Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
@@ -226,11 +226,8 @@ class SentryFeedbackTests: XCTestCase {
         // Verify that the feedback was captured and sent despite the 0.0 sample rate
         let lastSentEventArguments = try XCTUnwrap(transportAdapter.sendEventWithTraceStateInvocations.last)
         let capturedFeedback = try XCTUnwrap(lastSentEventArguments.event)
-        print(capturedFeedback)
-//
-//        XCTAssertEqual(capturedFeedback.message, "Test feedback message")
-//        XCTAssertEqual(capturedFeedback.name, "Test User")
-//        XCTAssertEqual(capturedFeedback.email, "test@example.com")
+
+        XCTAssertEqual(capturedFeedback.type, SentryEnvelopeItemTypeFeedback)
     }
     
     func testFeedbackNotSubjectToBeforeSendFiltering() throws {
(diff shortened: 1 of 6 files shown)


Tests/SentryTests/Integrations/Feedback/SentryFeedbackTests.swift
```diff
@@ -190,6 +190,130 @@ class SentryFeedbackTests: XCTestCase {
...
+        SentrySDK.capture(feedback: feedback)
+        
+        // Verify that the feedback was captured and sent despite the 0.0 sample rate
+        let lastSentEventArguments = try XCTUnwrap(transportAdapter.sendEventWithTraceStateInvocations.last)
+        let capturedFeedback = try XCTUnwrap(lastSentEventArguments.event)
+        print(capturedFeedback)
+//
+//        XCTAssertEqual(capturedFeedback.message, "Test feedback message")
+//        XCTAssertEqual(capturedFeedback.name, "Test User")
+//        XCTAssertEqual(capturedFeedback.email, "test@example.com")
```
itaybre: This asserts should be enabled, right?

armcknight: whoops, no, they should be removed. there's currently no way to inspect the event for these particular attributes, since SentryFeedback (like SentryUserFeedback before it) don't actually inherit from SentryEvent.

armcknight: Good catch, I can't check these specific attributes since SentryFeedback (like SentryUserFeedback before it) doesn't inherit from SentryEvent, but, I will see if I can check something in the event metadata that indicates it's actually a user feedback.

armcknight: added assertions on event types in https://github.com/getsentry/sentry-cocoa/pull/5692/commits/b4e4c51c78f7872284b782fdfc26434fdcd4b42c

PR #5693, opened 2025-07-23
ci: Bump macos-13 to macos-14 runners to use arm64 runners

itaybre:
According to github docs: https://github.com/actions/runner-images?tab=readme-ov-file
macos-13 does not use arm64 machines

There are a few macos-13 remaining but used for Xcode 14.
When runners are deprecated in september, we will update those

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -139,6 +139,12 @@ jobs:
             --device "iPhone 14" \
             --configuration DebugWithoutUIKit \
             --derived-data uikit-check-build
+      - run: ls uikit-check-build
+      - run: ls uikit-check-build/Build
+      - run: ls uikit-check-build/Build/Products
+      - run: ls uikit-check-build/Build/Products/DebugWithoutUIKit-iphonesimulator
+      - run: ls uikit-check-build/Build/Products/DebugWithoutUIKit-iphonesimulator/SentryWithoutUIKit.framework/
+      - run: ls uikit-check-build/Build/Products/DebugWithoutUIKit-iphonesimulator/SentryWithoutUIKit.framework/SentryWithoutUIKit
       - name: Ensure UIKit is not linked
         run: ./scripts/check-uikit-linkage.sh DebugWithoutUIKit uikit-check-build unlinked SentryWithoutUIKit
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -126,7 +126,7 @@ jobs:
 
   check-debug-without-UIKit:
     name: Check no UIKit linkage (DebugWithoutUIKit)
-    runs-on: macos-14
+    runs-on: macos-15
     steps:
       - uses: actions/checkout@v4
       - name: Build for Debug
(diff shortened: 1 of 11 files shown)


Sources/Swift/Integrations/Performance/IO/Data+SentryTracing.swift
```diff
@@ -18,9 +18,9 @@ public extension Data {
     ///   - url: The location on disk of the data to read.
     ///   - options: The mask specifying the options to use when reading the data. For more information, see ``NSData.ReadingOptions``.
     /// - Note: See ``Data.init(contentsOf:options:)`` for more information.
-    init(contentsOfWithSentryTracing url: URL, options: Data.ReadingOptions = []) throws {
+    init(contentsOfWithSentryTracing url: URL, options: NSData.ReadingOptions = []) throws {
```
armcknight: Is this required for this PR? What was going on with it?

.github/workflows/lint.yml
```diff
@@ -31,9 +31,11 @@ concurrency:
 jobs:
   swift-lint:
     name: Swift Lint
-    runs-on: macos-13
+    runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
+      - name: Install tooling
```
armcknight, line 37: How was this working before?

Sources/Swift/Integrations/Performance/IO/Data+SentryTracing.swift
armcknight: This reminded me of something similar: https://github.com/getsentry/sentry-cocoa/pull/5059

Could we try the same fix here, prefixing as `Foundation.Data` or `Swift.Data`? I don't actually see that string anywhere already in the codebase so maybe they backed out the change but forgot to update the PR description. It's kind of hard to follow :/

armcknight (approved):
Thanks for doing this, although I just wanted to be sure all the changes here are required for the macOS image version upgrade, or are there other unrelated changes here?

Sources/Swift/Integrations/Performance/IO/Data+SentryTracing.swift
itaybre: This shouldn't be here, I moved to https://github.com/getsentry/sentry-cocoa/pull/5695
Will move this PR back to draft

.github/workflows/lint.yml
itaybre: Swiftlint was preinstalled on macos13

itaybre:
Moving this back to draft, there are some changes I am aplying in others PR which should unlock more runners to be upgraded

.github/workflows/build.yml
```diff
@@ -126,7 +126,7 @@ jobs:
 
   check-debug-without-UIKit:
     name: Check no UIKit linkage (DebugWithoutUIKit)
-    runs-on: macos-13
+    runs-on: macos-15
```
noahsmartin: Will we need to Xcode select an older version of Xcode here to make sure the SDK still builds with Xcode 15? Or are we ok with using new Swift features only available in the latest Xcode?

philippsuess: Unfortunately, we will only be able testing with Xcode 15.0.1 / iOS 17 and above when using macos-14 https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md. But yes we should bump to macos 14 and still use the lowest Xcode version possible.

itaybre: For this check we only run otool, so there is no need to use an old version of macos.
I haven't made any change regarding the xcode version used to build the xcframework

PR #5698, opened 2025-07-23
ref: Create SentryOptionsInternal

noahsmartin:
This pulls all the internal code that doesn't really need to be part of SentryOptions into its own helper class. With this change the hybrid SDK options initialization logic goes through a static method on SentryOptionsInternal, so SentryOptions can be written in Swift without affecting the API that hybrid SDKs use.

While this doesn't change the public API, it does change the hybrid SDK from using `[[SentryOptions alloc] initWithDict:dictionary didFailWithError:&error]` to using `[SentryOptionsInternal initWithDict:dictionary didFailWithError:&error]` I will follow up with hybrid SDK PRs to make this change.

#skip-changelog


Tests/SentryTests/SentryTests.m
```diff
@@ -34,7 +34,7 @@ - (void)setUp
 - (void)testSharedClient
 {
     NSError *error = nil;
-    SentryOptions *options = [[SentryOptions alloc]
+    SentryOptions *options = [[SentryOptionsInternal alloc]
```
itaybre: Be careful this is broken

itaybre: Also, should it be `SentryOptionsInternal *options`?

itaybre: I think you need to add it to `Sentry.modulemap` https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Resources/Sentry.modulemap#L7

PR #5701, opened 2025-07-23
fix: some versioning issues in SDK

armcknight:
Also from #5139, along with #5700, this fixes some incorrect assignments of kinds of versions to settings and infoplist entries.

#skip-changelog


armcknight:
There is actually more work that would need to be done to avoid breaking the release flow, in VersionBump, so I'm not going to merge this as-is. I resolved the most recent conflicts, but the same kind will crop up again if another release is done.

Issue #5705, opened 2025-07-24
SwiftUI Crash Tests Not Crashing


philipphofmann:
### Description

The SwiftUI Crash Tests didn't crash, although they should, see https://github.com/getsentry/sentry-cocoa/actions/runs/16480337748/job/46592679402

```
[20:00:42] Installing app on simulator.
[20:01:43] Terminating app if running.
[20:01:47] Setting crash flag.
[20:01:50] Launching app with expected crash.
io.sentry.tests.SwiftUICrashTest: 11099
[20:01:59] ⏳ App is still running
[20:01:59] ❌ App is still running after 5 seconds, but it should have crashed instead.
Error: Process completed with exit code 1.
```

Maybe we need to increase the timeout for waiting for the app to crash.

PR #5706, opened 2025-07-24
ci: SwiftUI Crash Test Increase Timeouts

philipphofmann:
Increase the timeout to 20 seconds to check if the app has crashed and if it's still running. Also add more log messages.

Fixes GH-5705

#skip-changelog

### DIFF
--- a/TestSamples/SwiftUICrashTest/test-crash-and-relaunch.sh
+++ b/TestSamples/SwiftUICrashTest/test-crash-and-relaunch.sh
@@ -100,7 +100,9 @@ xcrun simctl spawn $DEVICE_ID defaults write $BUNDLE_ID $USER_DEFAULT_KEY -bool 
 log "Launching app with expected crash."
 xcrun simctl launch $DEVICE_ID $BUNDLE_ID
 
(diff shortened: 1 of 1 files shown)


TestSamples/SwiftUICrashTest/test-crash-and-relaunch.sh
```diff
@@ -100,7 +100,9 @@ xcrun simctl spawn $DEVICE_ID defaults write $BUNDLE_ID $USER_DEFAULT_KEY -bool
 log "Launching app with expected crash."
 xcrun simctl launch $DEVICE_ID $BUNDLE_ID
 
-# Check every 100ms for 5 seconds if the app is still running.
+log "Starting to check if app crashed as expected."
```
noahsmartin, line 103: Something I recently learned about is `xcrun simctl runtime dyld_shared_cache update --all` That will block until the dyld shared cache is created for the simulators. This could potentially be happening and the app won't launch until it's done, so adding that before we start waiting for a crash may help to avoid timeouts

noahsmartin (approved):

PR #5716, opened 2025-07-24
test: remove unused plan and add more to workspace for editing

armcknight:
Noticed a couple things with test plans:
- some weren't added to the workspace, so i added them so it's easier to edit them
- there was one that was unused `macOS-SwiftUI` has no tests

#skip-changelog


philipphofmann (approved):
Thanks for the cleanup.

PR #5718, opened 2025-07-24
docs: swiftify the headerdocs

armcknight:
A suggestion to make the headerdocs idiomatically swift:
- use `///` instead of `/** */`
- use e.g. `- parameter paramName:` instead of `@param paramName`
- use backticks where we used to have `@c`

A couple very small tweaks to improve things, like marking a few more things with `- note:` and `- seealso:`

#skip-changelog


noahsmartin (approved):

philippsuess:
Thanks, @armcknight. It would be great to have some linter for this to ensure it doesn't happen again.

PR #5719, opened 2025-07-24
ci: separate slather step

armcknight:
I noticed that sometimes unit test jobs still report failures even though no tests fail, with the script exiting nonzero. See https://github.com/getsentry/sentry-cocoa/actions/runs/16505144160/job/46673775222

I thought maybe this is actually an error coming from slather. Separate that into its own step so we can isolate potential failures.

#skip-changelog

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -254,6 +254,9 @@ jobs:
           path: |
             ~/Library/Logs/DiagnosticReports/**
 
+      - name: Gather code coverage information
+        run: slather coverage --configuration TestCI --scheme Sentry
+
       # We can upload all coverage reports, because codecov merges them.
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -254,6 +254,9 @@ jobs:
           path: |
             ~/Library/Logs/DiagnosticReports/**
 
+      - name: Gather code coverage information
```
philippsuess: `l`: So we know we're running slather when checking the jobs in CI
```suggestion
      - name: Gather code coverage information via slather
        run: slather coverage --configuration TestCI --scheme Sentry
```

itaybre:
Looks like this still failed with the same error here.
Might be related to the teardown of xcodebuild, not the slather step

PR #5658, opened 2025-07-18
test: try codecov test analytics

armcknight:
Trying the new codecov test analytics as mentioned in #3897 (docs at https://docs.codecov.com/docs/test-analytics)

#skip-changelog

PR #5720, opened 2025-07-24
test: remove obsoleted slowest test script

armcknight:
After adding codecov test analytics in #5658, we can now see aggregated analytics, including which tests take the most time to run. Now we don't have to use CI time to do this 🎉 See https://app.codecov.io/github/getsentry/sentry-cocoa/tests/main?parameter=SLOWEST_TESTS

#skip-changelog


philipphofmann (approved):
Good catch thanks 🥇

PR #5721, opened 2025-07-24
Use optimize for size in size tests

noahsmartin:
#skip-changelog


armcknight (approved):
interesting. I guess -O is better in niche cases where certain hot paths can be optimized separately, as a next step before dropping down to assembly.

PR #5723, opened 2025-07-25
Add test for null cache path

noahsmartin:
#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -156,6 +156,10 @@ _non_thread_safe_removeFileAtPath(NSString *path)
     NSString *cachePath = options.cacheDirectoryPath;
 
     SENTRY_LOG_DEBUG(@"SentryFileManager.cachePath: %@", cachePath);
+    if (cachePath == NULL) {
+        SENTRY_LOG_FATAL(
+            @"SentryFileManager attempting to be created with a NULL cachePath %@", cachePath);
+    }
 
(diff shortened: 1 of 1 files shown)


Sources/Sentry/SentryFileManager.m
```diff
@@ -156,6 +156,10 @@ - (void)createPathsWithOptions:(SentryOptions *)options
     NSString *cachePath = options.cacheDirectoryPath;
 
     SENTRY_LOG_DEBUG(@"SentryFileManager.cachePath: %@", cachePath);
+    if (cachePath == NULL) {
+        SENTRY_LOG_FATAL(
+            @"SentryFileManager attempting to be created with a NULL cachePath %@", cachePath);
```
philippsuess, line 161: `m`: I think it would be great to explain what could cause this and the problems it causes. Otherwise it's confusing for users.

Sources/Sentry/SentryFileManager.m
```diff
@@ -156,6 +156,10 @@ - (void)createPathsWithOptions:(SentryOptions *)options
     NSString *cachePath = options.cacheDirectoryPath;
 
     SENTRY_LOG_DEBUG(@"SentryFileManager.cachePath: %@", cachePath);
+    if (cachePath == NULL) {
```
philippsuess, line 159: `m`: I think it should be nil?
```suggestion
    if (cachePath == nil) {
```

philprime:
@noahsmartin why did you close this issue after the PR was approved? If it's not needed anymore a PR comment would be great for future reference

PR #5724, opened 2025-07-25
test: Fix flaky FlushCalledMultipleTimes

philipphofmann:
Increase the flush timeout to 5 seconds and use XCTestExpectations to let the test fail with proper assert messages.

#skip-changelog


philippsuess:
Back to draft cause some unit tests are failing.

philippsuess:
The flaky test validates an unlikely edge case of calling flush in a tight loop. I already spent a lot of time to fix this tests. I would rather close this PR and remove the test than spending more time fixing the flaky test.

PR #5729, opened 2025-07-25
ci: Set xcodeVersion to 14.3.1 in xcodegen configurations

itaybre:
Set xcodeVersion to 14.3.1 so CI can open xcode projects

#skip-changelog


itaybre:
The issue stems from xcodegen using Xcode 16's object version.
I will fix the the xcodegen version to the previous one

Issue #5731, opened 2025-07-25
Fix flaky test `testFlush_CallingFlushDirectlyAfterCapture_Flushes`


itaybre:
### Description

The test `testFlush_CallingFlushDirectlyAfterCapture_Flushes` is experiencing some flakiness.

philippsuess:
Might be fixed with #5724.

itaybre:
Fixed in #5754

PR #5732, opened 2025-07-25
test: Use test file manager to track envelopes created and deleted

itaybre:
Fixes: #5731

### DIFF
--- a/Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
+++ b/Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
@@ -139,8 +139,7 @@ final class SentryHttpTransportFlushIntegrationTests: XCTestCase {
         options.debug = true
         options.dsn = TestConstants.dsnAsString(username: "SentryHttpTransportFlushIntegrationTests.\(testName)")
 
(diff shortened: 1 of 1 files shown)


philprime:
@itaybre can you please add more information how this fixes the problem?

philippsuess:
> @itaybre can you please add more information how this fixes the problem?

Yes, I'm also curious.

itaybre:
The previous approach used `SentryFileManager` which checks for envelopes in a directory, files there may be created in the middle of the test by other processes (not ideal, but outside of the scope of this PR) but this will let us ignore that problem.

philippsuess:
> The previous approach used `SentryFileManager` which checks for envelopes in a directory, files there may be created in the middle of the test by other processes (not ideal, but outside of the scope of this PR) but this will let us ignore that problem.

The getSut already uses a different DSN per test method, so each file manager uses a different directory anyway. The changes in this PR shouldn't solve the problem you described ⬆️ . Using an in-memory file manager could help at first glance, but if we can't use a per-test-scoped file manager for the tests, something's wrong with the file manager, and we should fix it, because our users would also have problems with it in production.

itaybre:
Closing since this approach is not going to fix this issue

PR #5736, opened 2025-07-28
chore: Enable strict nullability warnings

philprime:
_This PR is derived from #5572 in an effort to make the large amount of changes easier to review._

This PR enables strict nullability handling in Objective-C as warnings in Xcode.
By adding the flags twice, once with `no-error` and once without, the flags are treated as warnings instead of errors.

```
OTHER_CFLAGS = (
    "-Wnullable-to-nonnull-conversion",
    "-Wnullability-completeness",
    "-Wno-error=nullable-to-nonnull-conversion",
    "-Wno-error=nullability-completeness",
);
```

After all warnings have resolved we'll remove the `no-error` ones, to change the warnings to compile-time errors.

#skip-changelog

PR #5737, opened 2025-07-28
refactor: Add full nullability handling to SentryFileManager

philprime:
_This PR is derived from #5572 in an effort to make the large amount of changes easier to review for #5577._

**Should be merged after #5736**

Fixes nullability handling in the `SentryFileManager`.

Most notably this PR adds a new macro `SENTRY_UNWRAP_NULLABLE` which is just a convenience macro for force-casting a nullable to non-null pointer. I introduced it, so we can easily find all locations where we force-cast.

#skip-changelog

### DIFF
--- a/Sources/Sentry/Public/SentryDefines.h
+++ b/Sources/Sentry/Public/SentryDefines.h
@@ -203,3 +203,20 @@ typedef void (^SentryUserFeedbackConfigurationBlock)(
     SentryUserFeedbackConfiguration *_Nonnull configuration);
 
 #endif // TARGET_OS_IOS && SENTRY_HAS_UIKIT
+
+/**
+ * `SENTRY_UNWRAP_NULLABLE` is used to unwrap a nullable pointer type to a non-nullable pointer type
+ *  It should be used after the pointer has been checked for nullability.
+ *
+ *  For example:
+ *  ```objc
+ *  id _Nullable nullablePointer = ...;
+ *  if (nullablePointer != nil) {
+ *      MyClass *_Nonnull nonNullPointer = SENTRY_UNWRAP_NULLABLE(MyClass, nullablePointer);
+ *  }
+ *  ```
+ *
+ *  We use this macro instead of directly casting to be able to find all usages of this
+ *  pattern in the codebase.
+ */
+#define SENTRY_UNWRAP_NULLABLE(type, nullable_var) (type *_Nonnull)(nullable_var)
--- a/Sources/Sentry/SentryFileManager.m
+++ b/Sources/Sentry/SentryFileManager.m
@@ -161,9 +161,15 @@ _non_thread_safe_removeFileAtPath(NSString *path)
 
     NSString *_Nullable nullableDsnHash = [options.parsedDsn getHash];
     if (nullableDsnHash == nil) {
-        SENTRY_LOG_DEBUG(@"No DSN provided, using base path for envelopes: %@", self.basePath);
+        SENTRY_LOG_FATAL(@"No DSN provided, using base path for envelopes: %@", self.basePath);
     }
-    self.sentryPath = [self.basePath stringByAppendingPathComponent:nullableDsnHash ?: @"default"];
+    // We decided against changing the `sentryPath` and use a null fallback instead, because the
+    // impact of changing the base path can result in critical issues.
+    //
+    // Instead we silence the nullability warning and let `stringByAppendingPathComponent` handle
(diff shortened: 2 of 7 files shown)


Sources/Sentry/Public/SentryDefines.h
```diff
@@ -203,3 +203,20 @@ typedef void (^SentryUserFeedbackConfigurationBlock)(
...
+ *  For example:
+ *  ```objc
+ *  id _Nullable nullablePointer = ...;
+ *  if (nullablePointer != nil) {
+ *      MyClass *_Nonnull nonNullPointer = SENTRY_UNWRAP_NULLABLE(MyClass, nullablePointer);
+ *  }
+ *  ```
+ *
+ *  We use this macro instead of directly casting to be able to find all usages of this
+ *  pattern in the codebase.
```
philippsuess, line 220: Thanks for the explanation

Sources/Sentry/SentryFileManager.m
```diff
@@ -158,7 +158,13 @@ - (void)createPathsWithOptions:(SentryOptions *)options
     SENTRY_LOG_DEBUG(@"SentryFileManager.cachePath: %@", cachePath);
 
     self.basePath = [cachePath stringByAppendingPathComponent:@"io.sentry"];
-    self.sentryPath = [self.basePath stringByAppendingPathComponent:[options.parsedDsn getHash]];
+
+    NSString *_Nullable nullableDsnHash = [options.parsedDsn getHash];
+    if (nullableDsnHash == nil) {
+        SENTRY_LOG_DEBUG(@"No DSN provided, using base path for envelopes: %@", self.basePath);
+    }
+    self.sentryPath = [self.basePath stringByAppendingPathComponent:nullableDsnHash ?: @"default"];
```
philippsuess: `h`: Now it's obvious that we have a significant problem here. It shouldn't be possible to initialize a SentryFilemanager without a DSN. We previously silently wrote the envelopes to the `basePath`. We shouldn't add a new default path. The envelopes in the `default` path will never be sent because the SDK doesn't have a DSN and doesn't know where to send them. 
Instead, I think we should keep the wrong current behaviour and not add a new folder called default and we should follow up with ensuring you can't create an SentryFileManager without a DSN.

philprime: Are we certain that `stringByAppendingPathComponent` with a `nil` argument returns the original base string? I agree that changing the behavior shouldn't be taken lightly, but I also don't think we should use the `basePath` without any suffix. 

I don't have a strong opinion here. I can keep the original approach, silent the warning and add a comment if you would like me to.

philippsuess: I would maintain the existing broken approach and refrain from introducing new logic that's also flawed and create an issue to follow up to ensure you can't even use a SentryFileManager when not having a DSN. Fixing that could be a bit more work.

philprime: I double-checked it:

```objc
NSLog(@"%@", [self.basePath stringByAppendingPathComponent:x]);
NSLog(@"%@", [self.basePath stringByAppendingPathComponent:@""]);
NSLog(@"%@", [self.basePath stringByAppendingPathComponent:@"foo"]);
```

results in:

```
.../Library/Caches/io.sentry
.../Library/Caches/io.sentry
.../Library/Caches/io.sentry/foo
```

So while the method signature of `stringByAppendingPathComponent` is not defining it as null-safe, it seems to work that way.

Sources/Sentry/SentryFileManager.m
```diff
@@ -158,7 +158,19 @@ - (void)createPathsWithOptions:(SentryOptions *)options
...
 
     self.basePath = [cachePath stringByAppendingPathComponent:@"io.sentry"];
-    self.sentryPath = [self.basePath stringByAppendingPathComponent:[options.parsedDsn getHash]];
+
+    NSString *_Nullable nullableDsnHash = [options.parsedDsn getHash];
+    if (nullableDsnHash == nil) {
+        SENTRY_LOG_FATAL(@"No DSN provided, using base path for envelopes: %@", self.basePath);
+    }
+    // We decided against changing the `sentryPath` and use a null fallback instead, because the
+    // impact of changing the base path can result in critical issues.
```
philippsuess: `m`: I would add that this is broken for a very long time already and we plan on fixing this in an GH issue. You could link the issue.

philipphofmann (approved):
LGTM.

PR #5741, opened 2025-07-28
ci: Increase test timeout

noahsmartin:
The shorter timeout was still sometimes to short, to simplify let's just use one timeout. 

I saw the timeout error here: https://github.com/getsentry/sentry-cocoa/actions/runs/16528521520/job/46747914086

#skip-changelog


philprime:
@philipphofmann @itaybre I believe you mentioned we should not increase timeouts earlier today in the team sync. Is this related to that, or were you talking about CI job timeouts?

PR #5742, opened 2025-07-28
refactor: Add nullability handling for data serialization

philprime:
_This PR is derived from #5572 in an effort to make the large amount of changes easier to review for #5577._

**Should be merged after #5737**

Fixes nullability handling and tests for the `SentrySerialization` and `SentryEnvelope`

#skip-changelog


Sources/Sentry/SentrySerialization.m
```diff
@@ -260,13 +269,28 @@ + (SentrySession *_Nullable)sessionWithData:(NSData *)sessionData
     return session;
 }
 
-+ (NSData *)dataWithReplayRecording:(SentryReplayRecording *)replayRecording
++ (NSData *_Nullable)dataWithReplayRecording:(SentryReplayRecording *)replayRecording
```
philprime, line 272: I am not 100% certain if an early return approach makes sense here, or if we should return empty data instead.

I decided to go for the early return, as it does not make sense to me to return data e.g. without the header.

WDYT?

Sources/Sentry/SentryEnvelope.m
```diff
@@ -231,15 +232,17 @@ - (nullable instancetype)initWithReplayEvent:(SentryReplayEvent *)replayEvent
...
     NSURL *envelopeContentUrl =
         [[videoURL URLByDeletingPathExtension] URLByAppendingPathExtension:@"dat"];
 
-    BOOL success = [SentryMsgPackSerializer serializeDictionaryToMessagePack:@{
-        @"replay_event" : replayEventData,
-        @"replay_recording" : recording,
-        @"replay_video" : videoURL
+    NSMutableDictionary *envelopeContent = [NSMutableDictionary dictionary];
+    envelopeContent[@"replay_event"] = replayEventData;
+    if (nil != recording) {
```
philprime: Instead of skipping the recording data, it might make sense to abort the envelope initialization instead. WDYT?

philippsuess: If a replay event without a recording is still somewhat useful, then I would say keep the envelope with only the replay event. If Relay or the backend discards such envelopes, then we could throw away the envelope here already.

Sources/Sentry/SentrySerialization.m
philippsuess: In this case, I think an envelope item without a header is invalid, and we should do an early return. So please keep it as it is.

philipphofmann (approved):
LGTM

Sources/Sentry/SentryEnvelope.m
philprime: I asked internally to find the best approach for this case.

philprime:
Result from internal discussion:

When encoding of replay data fails, it should be propagated to the caller, to then drop the envelope and emit a client report.

philippsuess:
> When encoding of replay data fails, it should be propagated to the caller, to then drop the envelope and emit a client report.

@philprime, we could also move the relay data serialization to its own PR. So basically, remove it from this PR and follow up. Then we can already merge this. Up to you if it's worth the effort.

philprime:
@philipphofmann I looked into the propagation of the encoding result to report the dropped event. The PR is a work-in-progress, but the main problem I encountered is, that right now we if we return `nil`, the event is dropped as we want, but I can not use that as a trigger to record the lost event.

https://github.com/getsentry/sentry-cocoa/pull/5742/files#diff-47eb3eb3ea905acecc12679b7d5eac4688cb64f396c5e02f3213aefc5d0a639aR547-R550

If I leave the changes of the `SentryClient.m` like this, it would also record events which where filtered using `beforeSend`.

My idea would be changing the `prepareEvent:withScope:alwaysAttachStacktrace` to throw an error if encoding fails, but that would be quite a large change.

WDYT?

Sources/Sentry/SentryEnvelope.m
```diff
@@ -231,15 +233,23 @@ - (nullable instancetype)initWithReplayEvent:(SentryReplayEvent *)replayEvent
...
-    BOOL success = [SentryMsgPackSerializer serializeDictionaryToMessagePack:@{
-        @"replay_event" : replayEventData,
-        @"replay_recording" : recording,
-        @"replay_video" : videoURL
-    }
+    NSMutableDictionary *envelopeContent = [NSMutableDictionary dictionary];
+    envelopeContent[@"replay_event"] = replayEventData;
+    envelopeContent[@"replay_recording"] = recording;
+    envelopeContent[@"replay_video"] = videoURL;
+    BOOL success = [SentryMsgPackSerializer serializeDictionaryToMessagePack:envelopeContent
```
philprime: This can now be reverted:

```suggestion
    BOOL success = [SentryMsgPackSerializer serializeDictionaryToMessagePack:@{
        @"replay_event" : replayEventData,
        @"replay_recording" : recording,
        @"replay_video" : videoURL
    }
```

philprime:
cursor review

Sources/Sentry/SentryClient.m
```diff
@@ -554,8 +562,12 @@ - (void)captureReplayEvent:(SentryReplayEvent *)replayEvent
                                                   video:videoURL];
 
     if (videoEnvelopeItem == nil) {
-        SENTRY_LOG_DEBUG(@"The Session Replay segment will not be sent to Sentry because an "
+        SENTRY_LOG_ERROR(@"The Session Replay segment will not be sent to Sentry because an "
                          @"Envelope Item could not be created.");
+        // Record a lost event in case the preparing of the event (e.g. encoding the event) failed.
+        // This is used to determin if replay events are missing due to an error in the SDK.
```
philprime:
```suggestion
        // This is used to determine if replay events are missing due to an error in the SDK.
```

Sources/Sentry/SentryEnvelope.m
```diff
@@ -201,10 +203,10 @@ - (_Nullable instancetype)initWithAttachment:(SentryAttachment *)attachment
...
                 containsObject:@"--io.sentry.other.base64-attachment-data"]) {
-            data = [[[[NSFileManager defaultManager] contentsAtPath:attachment.path]
+            data = [[[[NSFileManager defaultManager] contentsAtPath:attachmentPath]
                 base64EncodedStringWithOptions:0] dataUsingEncoding:NSUTF8StringEncoding];
         } else {
-            data = [[NSFileManager defaultManager] contentsAtPath:attachment.path];
+            data = [[NSFileManager defaultManager] contentsAtPath:attachmentPath];
         }
 #else
         data = [[NSFileManager defaultManager] contentsAtPath:attachment.path];
```
philprime:
```suggestion
        data = [[NSFileManager defaultManager] contentsAtPath:attachmentPath];
```

Tests/SentryTests/SentryClientTests.swift
```diff
@@ -2073,21 +2073,59 @@ class SentryClientTest: XCTestCase {
...
+        // -- Act --
+        sut.capture(replayEvent, replayRecording: replayRecording, video: movieUrl, with: scope)
+
+        // -- Assert --
+        XCTAssertEqual(fixture.transport.sentEnvelopes.count, 0)
+
+        XCTAssertEqual(fixture.transport.recordLostEventsWithCount.count, 1)
+        let lostEvent = try XCTUnwrap(fixture.transport.recordLostEventsWithCount.first)
+        XCTAssertEqual(lostEvent.category, .replay)
+        XCTAssertEqual(lostEvent.reason, SentryDiscardReason.eventProcessor)
```
philprime:
```suggestion
        XCTAssertEqual(lostEvent.reason, SentryDiscardReason.insufficientData)
```

PR #5744, opened 2025-07-28
ci: fix broken bump version util due to missing path

itaybre:


philprime:
How was this tested/verified? Every time we touch the version bump utility, we must ensure it's still working properly! This has been broken way to many times.

Also please start writing PR descriptions explaining how/why this PR fixes something. It's hard to understand why it was changed when not working on the utility myself.

cc @philipphofmann

philippsuess:
> Also please start writing PR descriptions explaining how/why this PR fixes something. It's hard to understand why it was changed when not working on the utility myself.

Yes, please do ⬆️

itaybre:
The title of the PR is the same content I would add in the description.
A file was removed and no longer existed so the utility crashed.

How I tested it:
- I ran it -> No crash
- CI confirmed this

philprime:
For future reference and additional context for this PR:

The file was removed in #5700 because it now uses the shared file located at `Sources/Configuration/Versioning.xcconfig`

PR #5746, opened 2025-07-29
The test testFlush_CalledMultipleTimes_ImmediatelyReturnsFalse tests an

philipphofmann:
unlikely edge case of calling flush in a tight loop. Instead of investing more time to fix the flaky tests, I think it's acceptable to simply delete the test. We hardly ever change the flush implementation, and it's not worth the effort to keep a test that causes so much maintenance effort.

Fixes GH-5731

#skip-changelog

### DIFF
--- a/SentryTestUtils/TestTransport.swift
+++ b/SentryTestUtils/TestTransport.swift
@@ -28,8 +28,4 @@ public class TestTransport: NSObject, Transport {
         flushInvocations.record(timeout)
         return .success
     }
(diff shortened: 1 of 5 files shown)


philippsuess:
It seems like https://github.com/getsentry/sentry-cocoa/pull/5754 actually fixed this flaky tests. I'm going to close this. We can add the `MissingTestCase` approach as soon as we need it in any other PR.

PR #5751, opened 2025-07-29
ci: Assert a UIWindowScene is available on sceneDidBecomeActive

noahsmartin:
Sometimes I see a UI test fail to find the feedback button in the SwiftUI sample app. One codepath I saw that would not present the button is if there isn't a connected UIWindowScene. Adding this assertion so if this is the reason it fails it fails early so we can tell this is happening


philipphofmann (approved):
LGTM.

PR #5754, opened 2025-07-29
test: Update SentryHttpTransportFlushIntegrationTests to wait for the dispatch queue to complete before flushing

itaybre:
I believe this should fix the test in #5746 instead of removing it too
#skip-changelog

### DIFF
--- a/Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
+++ b/Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
@@ -8,7 +8,7 @@ final class SentryHttpTransportFlushIntegrationTests: XCTestCase {
 
     func testFlush_WhenNoEnvelopes_BlocksAndFinishes() throws {
 
-        let (sut, _, _) = try getSut()
+        let (sut, _, _, _) = try getSut()
 
         var blockingDurationSum: TimeInterval = 0.0
         let flushInvocations = 100
@@ -26,7 +26,7 @@ final class SentryHttpTransportFlushIntegrationTests: XCTestCase {
     }
 
     func testFlush_WhenNoInternet_BlocksAndFinishes() throws {
-        let (sut, requestManager, _) = try getSut()
+        let (sut, requestManager, _, _) = try getSut()
 
         requestManager.returnResponse(response: nil)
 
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
```diff
@@ -49,12 +49,18 @@ final class SentryHttpTransportFlushIntegrationTests: XCTestCase {
...
         defer { fileManager.deleteAllEnvelopes() }
 
         for _ in 0..<10 {
             sut.send(envelope: SentryEnvelope(event: Event()))
+            // Wait until the dispath queue drains to confirm the envelope is stored
+            let expectation = XCTestExpectation(description: "Envelope sent")
+            dispatchQueueWrapper.dispatchAsync {
+                expectation.fulfill()
+            }
+            wait(for: [expectation], timeout: 0.5)
```
noahsmartin: I think we should use a much higher timeout, like 10s. CI can be very slow and threads might not get a chance to run for a while. Worst case it adds 10s to a CI run and then fails. But best case it prevents and entire 10+ minute re-run on CI

Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
```diff
@@ -80,11 +86,18 @@ final class SentryHttpTransportFlushIntegrationTests: XCTestCase {
...
         requestManager.returnResponse(response: nil)
         for _ in 0..<30 {
             sut.send(envelope: SentryEnvelope(event: Event()))
+            
+            // Ensure envelope is stored
+            let expectation = XCTestExpectation(description: "Envelope sent")
+            dispatchQueueWrapper.dispatchAsync {
+                expectation.fulfill()
+            }
+            wait(for: [expectation], timeout: 0.5)
```
noahsmartin: Same here for 10s timeout

Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
philippsuess: Yes, 10 or 5 seconds is what we need.

philippsuess: When the file manager uses the same dispatch queue we should only need to call this once after the loop cause the dispatch queue is serial.

Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
itaybre: Bumped

noahsmartin (approved):

PR #5759, opened 2025-07-30
fix: remove internal macro SENTRY_UNWRAP_NULLABLE from public header

philprime:
Moves the utility macro introduced in #5737 from the public `SentryDefines.h` to the internal `SentryInternalDefines.h`, as it does not need to be publicly exposed.

#skip-changelog


itaybre (approved):

philipphofmann (approved):
LGTM

PR #5747, opened 2025-07-29
test: Fix ConcurrentTransactions_OnlyOneGetsMeasurement

philipphofmann:
Fix the failing `SentryTracer.testConcurrentTransactions_OnlyOneGetsMeasurement` by removing the `qos: .background` for the DispatchQueue, which led to timeouts in CI, because the DisptachQueue simply didn't start the transactions. 

Also, replace DispatchGroup with XCTestExpectation so we get a test failure with an assertion message instead of a infinite timeout or a crash if the async code doesn't run as expected.

I will follow up with PR to remove other usages of `qos: .background` in tests and a linter to warn us that we must not use this in tests.

#skip-changelog

PR #5760, opened 2025-07-30
ci: Use expectation for test

noahsmartin:
This has been failing in tests with this message:

```
Error:     Restarting after unexpected exit, crash, or test timeout in SentryTracerTests.testConcurrentTransactions_OnlyOneGetsMeasurement(); summary will include totals from previous launches.
```

Let's use an expectation so if it times out we get a clear error message

#skip-changelog


itaybre:
Isn't this the same as #5747?

noahsmartin:
@itaybre looks like it, we should merge that other one

PR #5763, opened 2025-07-31
ref: Remove SentryApplication forward declaration

philipphofmann:
The same header file declares the protocol below, no need to use a forward declaration.

#skip-changelog


philprime (approved):
LGTM, thanks.

PR #5768, opened 2025-07-31
Fix: Energy reporting should report energy not time

noahsmartin:
The data being reported here was CPU time, not energy (in joules) despite the API request claiming it was reporting joules. This led to invalid data being displayed on the frontend, the UI would say it was graphing watts, but it wasn't actually. You could see this by writing CPU intensive code (I used a simd operation) on a userInitiated queue and comparing it to the same operation on a background queue. Here's the result from my test app before this change on a userInitiatedQueue:

Screenshot 2025-07-31 at 9 42 47 AM

and on a background queue:
Screenshot 2025-07-31 at 9 43 48 AM

The shape is the same, and neither show reasonable numbers for watts (< 0.1W). The shape is the same because about the same amount of CPU time is used, one was just expected to draw more power.

With this fix the graph looks much better and you can clearly see the difference that a userInitiated and background queue make on energy consumption:
Screenshot 2025-07-31 at 9 46 20 AM

Screenshot 2025-07-31 at 9 46 43 AM

#skip-changelog


Sources/Sentry/SentryMetricProfiler.m
```diff
@@ -283,6 +285,8 @@ - (void)recordCPUsage
     }
 }
 
+// Only some architectures support reading energy.
+#if defined(__arm__) || defined(__arm64__)
```
philippsuess: `m`: It would be great to add some link or ressource how we know that these are the architectures support reading energy

Sources/Sentry/SentrySystemWrapper.mm
```diff
@@ -99,7 +100,8 @@ - (NSNumber *)cpuEnergyUsageWithError:(NSError **)error
             ;
         }
     }
-    return @(powerInfo.cpu_energy.total_system + powerInfo.cpu_energy.total_user);
+    return @(powerInfo.task_energy);
```
philippsuess, line 107: `m`: Can we somehow add a test for this? It's always a bit of a smell when we change something, and we don't have to adapt the tests. As far as I know there aren't any tests for this, but it would be great to have some.

philipphofmann (approved):
Thank you @noahsmartin.

Sources/Sentry/SentryMetricProfiler.m
noahsmartin: The code won't compile without these, because the task_energy property only exists  there:

```
struct task_power_info_v2 {
	task_power_info_data_t  cpu_energy;
	gpu_energy_data gpu_energy;
#if defined(__arm__) || defined(__arm64__)
	uint64_t                task_energy;
#endif /* defined(__arm__) || defined(__arm64__) */
	uint64_t                task_ptime;
	uint64_t                task_pset_switches;
};
```

I will clarify in the comment

Sources/Sentry/SentrySystemWrapper.mm
noahsmartin: Unfortunately this isn't really a logic error, just a misunderstanding of the API, so I'm not aware of any kind of test that would test for this kind of mistake. Also these functions are undocumented so the only way to really tell is to dig into the source code and see what Apple is setting this to/using it for

noahsmartin: The one thing I thought of was a test that actually tries to do something high energy and makes sure it measure it here, but then that is more testing the device/OS efficiency not really our own logic, so I'd rather keep something like that out of CI. 


Somewhat related: I do think we could have better tests to answer things like, What is the overhead of running Sentry in my app? And I'm looking into how we could do that

And: I could commit the tests I used to generate the images in the PR description to the sample app so anyone else can test it that way, I will open a new PR with that

philippsuess: Yes, testing these types of things is always challenging. We don't have the setup for things like these currently. 

A somewhat unconventional idea would be to run validation tests on the payload of incoming profiles. We could look at the data during processing and raise alerts if we detect anomalies. The major downside is that we would only know that something is broken after releasing, but it's better than not knowing at all. 
We could use that setup for validating various things.

 > Somewhat related: I do think we could have better tests to answer things like, What is the overhead of running Sentry in my app? And I'm looking into how we could do that
 
 We some tests testing the startup time here https://github.com/getsentry/sentry-cocoa/pull/5768#issuecomment-3140390884

But yes, I agree we could do a better job and this topic is challenging.

> And: I could commit the tests I used to generate the images in the PR description to the sample app so anyone else can test it that way, I will open a new PR with that

Thanks

PR #5770, opened 2025-07-31
fix: Memory leak in SentrySessionReplayIntegration

philipphofmann:
## :scroll: Description

The global event processor used a strong reference to self which lead to strong reference cycle, meaning that the ARC couldn't dealloc SentrySessionReplayIntegration. This also led to having multiple SentrySessionReplayIntegrations running in the tests after running the SentrySessionReplayIntegrationTests, making the tests flaky. This is fixed now by using a weak reference.

This came up while investigating https://github.com/getsentry/sentry-cocoa/pull/5747.

Unit test. The testSessionReplayIntegration_DoesNotLeakMemory fails without the weak ref.

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentrySessionReplayIntegration.m
+++ b/Sources/Sentry/SentrySessionReplayIntegration.m
@@ -159,12 +159,19 @@ static SentryTouchTracker *_touchTracker;
     [self cleanUp];
 
     [SentrySDKInternal.currentHub registerSessionListener:self];
+
+    __weak SentrySessionReplayIntegration *weakSelf = self;
     [SentryDependencyContainer.sharedInstance.globalEventProcessor
         addEventProcessor:^SentryEvent *_Nullable(SentryEvent *_Nonnull event) {
+            if (weakSelf == nil) {
+                SENTRY_LOG_DEBUG(@"WeakSelf is nil. Not doing anything.");
+                return event;
+            }
+
             if (event.isFatalEvent) {
-                [self resumePreviousSessionReplay:event];
+                [weakSelf resumePreviousSessionReplay:event];
             } else {
-                [self.sessionReplay captureReplayForEvent:event];
+                [weakSelf.sessionReplay captureReplayForEvent:event];
             }
             return event;
         }];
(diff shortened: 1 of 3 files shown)


Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -668,6 +668,30 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
         XCTAssertLessThan(processingQueue.queue.qos.relativePriority, assetWorkerQueue.queue.qos.relativePriority)
     }
 
+    func testSessionReplayIntegration_DoesNotLeakMemory() throws {
+
+        // -- Arrange --
+        weak var weakSut: SentrySessionReplayIntegration?
```
philprime, line 677: `m`: Please add a comment explaining that this is caused by the event processor having a reference to the integration.

Sources/Sentry/SentrySessionReplayIntegration.m
```diff
@@ -159,12 +159,19 @@ - (void)setupWith:(SentryReplayOptions *)replayOptions
     [self cleanUp];
 
     [SentrySDKInternal.currentHub registerSessionListener:self];
+
+    __weak SentrySessionReplayIntegration *weakSelf = self;
```
philprime, line 163: `m`: We should on-top of having a weak reference also add a `removeEventProcessor` when uninstalling the integration, or when de-initalizing.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
philippsuess: This would detect any memory leak you would add to the SessionReplayIntegration, and it's not specific to the GlobalEventProcessor. So if you would do something like this in the SessionReplayIntegration
```swift
dispatchQueue.async(after: .now() + 10) {
    self.wathever

}
```

this test would also now start to fail. I thought about making it specific to the event processor but decided against that to catch more potential future memory leaks.

Still, I can add a comment explaining what caused this in the past.

Sources/Sentry/SentrySessionReplayIntegration.m
philippsuess: Yes, I agree, but the event processor doesn't offer that at the moment. I created an issue to follow up https://github.com/getsentry/sentry-cocoa/issues/5771

The thing is that unistall of this integration is usually only called when somebody calls SentrySDK.close and this will also throw away the GlobalEventProcessor. So I don't think it's a huge deal, but we should do it to keep the state clean.

PR #5776, opened 2025-07-31
ci: Run Cocoapods integration test on iOS 18.5

itaybre:
iOS18.2 simulators sometimes crash like here: https://github.com/getsentry/sentry-cocoa/actions/runs/16647631298/job/47112011249
This PR selects iOS18.5 to run the integration test and avoid this issues

#skip-changelog

### DIFF
--- a/Samples/iOS-Cocoapods-Swift6/fastlane/Fastfile
+++ b/Samples/iOS-Cocoapods-Swift6/fastlane/Fastfile
@@ -8,14 +8,16 @@ platform :ios do
     run_tests(
       workspace: "iOS-Cocoapods-Swift6.xcworkspace",
       scheme: "App",
-      build_for_testing: true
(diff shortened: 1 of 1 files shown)


philprime:
This didn't help and now it create simulator clones which are also crashing:

https://github.com/getsentry/sentry-cocoa/actions/runs/16825449784/job/47660723001#step:5:1055

itaybre:
> This didn't help and now it create simulator clones which are also crashing:
> 
> https://github.com/getsentry/sentry-cocoa/actions/runs/16825449784/job/47660723001#step:5:1055

They are always called clones, I think that's on how fastlane runs the tests.
Regarding the error itself, it is a different one. I think `Failed to launch app with identifier: com.getsentry.iOS-Cocoapods-Swift6-UITests.xctrunner ` just happens sometimes.

PR #5752, opened 2025-07-29
refactor: Add null unwrapping to silence handled nullability warnings

philprime:
_This PR is derived from #5572 in an effort to make the large amount of changes easier to review for #5577._

**This PR must be merged after https://github.com/getsentry/sentry-cocoa/pull/5777**

Adds the macro `SENTRY_UNWRAP_NULLABLE` for already handled cases to silence warnings

#skip-changelog

PR #5777, opened 2025-08-01
refactor: Add dummy classes for SPM build system with beautified CI

philprime:
## :scroll: Description

- Changes the SPM build command in GitHub Actions to use xcbeautify for readibility
- Adds raw build log uploading as artifact
- Adds two dummy files for the SPM targets to be compilable without touching actual production code

Based on internal discussion with @noahsmartin:

- SPM uses the file system structure to determine which files are in which target.
- The file `SentryDsn.m` is included in the SentryHeaders SPM target, because SPM requires an objc target to have at least one implementation file. 
- This file was a good candidate since it only used public headers. 
- My PR #5752 adds a private header to it's imports, so it can't be used in that target anymore

Required to merge #5752

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,10 +112,21 @@ jobs:
     if: startsWith(github.ref, 'refs/heads/release/') == false
     steps:
       - uses: actions/checkout@v4
-      - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace && EXPERIMENTAL_SPM_BUILDS=1 xcodebuild build -scheme SentrySPM -sdk watchos -destination 'generic/platform=watchOS'
+
+      - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace
+      - run: set -o pipefail && NSUnbufferedIO=YES EXPERIMENTAL_SPM_BUILDS=1 xcodebuild build -scheme SentrySPM -sdk watchos -destination 'generic/platform=watchOS' | tee raw-build-output-spm-watchos.log | xcbeautify
         shell: sh
-      - run: EXPERIMENTAL_SPM_BUILDS=1 xcodebuild build -scheme SentrySPM -sdk iphoneos -destination 'generic/platform=iphoneos'
+      - run: set -o pipefail && NSUnbufferedIO=YES EXPERIMENTAL_SPM_BUILDS=1 xcodebuild build -scheme SentrySPM -sdk iphoneos -destination 'generic/platform=iphoneos' | tee raw-build-output-spm-iphoneos.log | xcbeautify
         shell: sh
+
+      - name: Upload SPM Build Logs
+        uses: actions/upload-artifact@v4
+        if: ${{ failure() || cancelled() }}
+        with:
+          name: raw-build-output-spm
+          path: |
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -70,7 +75,7 @@ if let env = env, String(cString: env, encoding: .utf8) == "1" {
             name: "SentryObjc",
             dependencies: ["SentrySwift"],
             path: "Sources",
-            exclude: ["Sentry/SentryDsn.m", "Sentry/NSLocale+Sentry.m", "Swift", "SentrySwiftUI", "Resources", "Configuration"],
```
philippsuess, line 73: `h`: Is it really OK to remove `"Sentry/SentryDsn.m", "Sentry/NSLocale+Sentry.m"`?

philprime: @noahsmartin can you please review and confirm this change?

.github/workflows/build.yml
```diff
@@ -112,11 +112,22 @@ jobs:
     if: startsWith(github.ref, 'refs/heads/release/') == false
     steps:
       - uses: actions/checkout@v4
-      - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace && EXPERIMENTAL_SPM_BUILDS=1 xcodebuild build -scheme SentrySPM -sdk watchos -destination 'generic/platform=watchOS'
+
+      - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace
+      - run: set -o pipefail && NSUnbufferedIO=YES EXPERIMENTAL_SPM_BUILDS=1 xcodebuild build -scheme SentrySPM -sdk watchos -destination 'generic/platform=watchOS' | tee raw-build-output-spm-watchos.log | xcbeautify
```
noahsmartin, line 117: Thanks for adding these!

Sentry.xcodeproj/project.pbxproj
```diff
@@ -883,6 +883,8 @@
 		D4CD2A812DE9F91900DA9F59 /* SentryRedactRegionType.swift in Sources */ = {isa = PBXBuildFile; fileRef = D4CD2A7D2DE9F91900DA9F59 /* SentryRedactRegionType.swift */; };
 		D4E3F35D2D4A864600F79E2B /* SentryNSDictionarySanitizeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D42E48582D48FC8F00D251BC /* SentryNSDictionarySanitizeTests.swift */; };
 		D4E3F35E2D4A877300F79E2B /* SentryNSDictionarySanitize+Tests.m in Sources */ = {isa = PBXBuildFile; fileRef = D41909942D490006002B83D0 /* SentryNSDictionarySanitize+Tests.m */; };
+		D4ECA4012E3CBEDE00C757EA /* SentryDummyPublicEmptyClass.m in Sources */ = {isa = PBXBuildFile; fileRef = D4ECA4002E3CBEDE00C757EA /* SentryDummyPublicEmptyClass.m */; };
```
noahsmartin, line 886: You don't technically need to add this to the Xcode project, it doesn't hurt to add it, but we don't need to

Package.swift
philprime: Noah reviewed the PR and approved the changes.

PR #5779, opened 2025-08-01
test: Remove background QoS usages

philipphofmann:
Remove usages of DispatchQueues with qos .background as these can cause flaky tests. Also add a linter to check for such usages in unit tests.

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Networking/RateLimits/SentryConcurrentRateLimitsDictionaryTests.swift
+++ b/Tests/SentryTests/Networking/RateLimits/SentryConcurrentRateLimitsDictionaryTests.swift
@@ -33,8 +33,8 @@ class SentryConcurrentRateLimitsDictionaryTests: XCTestCase {
     }
 
     func testConcurrentReadWrite() {
-        let queue1 = DispatchQueue(label: "SentryConcurrentRateLimitsStorageTests1", qos: .background, attributes: [.concurrent, .initiallyInactive])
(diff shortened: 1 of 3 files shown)


Tests/SentryTests/Networking/RateLimits/SentryConcurrentRateLimitsDictionaryTests.swift
```diff
@@ -33,8 +33,8 @@ class SentryConcurrentRateLimitsDictionaryTests: XCTestCase {
     }
 
     func testConcurrentReadWrite() {
-        let queue1 = DispatchQueue(label: "SentryConcurrentRateLimitsStorageTests1", qos: .background, attributes: [.concurrent, .initiallyInactive])
-        let queue2 = DispatchQueue(label: "SentryConcurrentRateLimitsStorageTests2", qos: .utility, attributes: [.concurrent, .initiallyInactive])
+        let queue1 = DispatchQueue(label: "SentryConcurrentRateLimitsStorageTests1", attributes: [.concurrent, .initiallyInactive])
```
philprime, line 36: `m`: I believe we should have two queues with different priorities, rather than the same one.

philprime (approved):
LGTM, not sure if the test behaviour changes, but won't block it.

Tests/SentryTests/Networking/RateLimits/SentryConcurrentRateLimitsDictionaryTests.swift
philippsuess: Thanks for double-checking. I looked at the test and I don't think different priorities are required to check if the implementation is thread safe.

PR #5781, opened 2025-08-01
ci: Add full screen capture in addition to simulator as backup

itaybre:
Some jobs fail to capture screenshots ([example](https://github.com/getsentry/sentry-cocoa/actions/runs/16675184003/job/47200028195))
Taking a screenshot should help here

#skip-changelog


philipphofmann (approved):
Useful addition. Thanks

PR #5782, opened 2025-08-01
refactor: Add nullability property for screenName

philprime:
This PR is derived from https://github.com/getsentry/sentry-cocoa/pull/5572 in an effort to make the large amount of changes easier to review for https://github.com/getsentry/sentry-cocoa/issues/5577.

Changes the parameter `screenName` to be nullable, as the method implementation allows it.


philipphofmann (approved):
Thanks for keeping the PRs small.

PR #5785, opened 2025-08-01
ref: Make SentrySession have a swift interface

noahsmartin:
Creates a Swift wrapper for SentrySession, and updates all callosities to using the wrapper (by changing the name of the ObjC type so it's now just an implementation detail)

This is because I'm trying to write a runloop observer feature all in Swift, and it requires SentryFileManager, which is blocked from being used in Swift APIs by SentrySession according to the latest ObjC to Swift conversion dependency tree: https://github.com/getsentry/sentry-cocoa/actions/runs/16677150074/job/47206480213

#skip-changelog


Sources/Swift/SentrySession.swift
```diff
@@ -0,0 +1,127 @@
...
+    }
+    
+    @objc public func setFlagInit() {
+        session.setFlagInit()
+    }
+
+    public func copy(with zone: NSZone? = nil) -> Any {
+        let copy = session.copy(with: zone)
+        guard let newSession = copy as? SentrySessionInternal else {
+            return copy
```
philippsuess: `h`: Is it really safe to return `copy` here? We have no clue what its type is. This would mean we're violating the NSCopying protocol, and we could get extremely weird bugs, if I'm not mistaken. I think it's actually safer to do a force unwrap here. I mean what can we actually do if that happens? I guess it's better to terminate the program than to violate the NSCopying protocol.

Sources/Swift/SentrySession.swift
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @objc public var sessionId: UUID {
+        session.sessionId
+    }
+    @objc public var started: Date {
+        session.started
+    }
+    @objc public var status: SentrySessionStatus {
+        SentrySessionStatus(rawValue: session.status) ?? .abnormal
```
philippsuess: `m`: Returning abnormal is new logic, if I'm not mistaken and we should add a test for it. I think it's correct to mark it as abnormal.

philipphofmann (commented):
I found one potential issue. Apart from that LGTM.

Sources/Swift/SentrySession.swift
noahsmartin: Updated this so the compiler enforces the type and we don't need an unwrap

Sources/Swift/SentrySession.swift
noahsmartin: Changed it to have an enum on the objc side too so we don't have to have this be optional

Sources/Swift/SentrySession.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    case ok       = 0
+    case exited   = 1
+    case crashed  = 2
+    case abnormal = 3
+}
+
+/// The SDK uses SentrySession to inform Sentry about release and project associated project health.
+@objc @_spi(Private) public class SentrySession: NSObject, NSCopying {
```
philprime, line 27: `m`: Just a double-check, could this change cause issues with Hybrid SDKs?

philprime (approved):
LGTM

Sources/Swift/SentrySession.swift
noahsmartin: It wasn't a part of the `_Hybrid` module so I don't think so

noahsmartin: And I kept the interface of the class the same

PR #5788, opened 2025-08-01
feat: Add a new dynamic framework with arm64e and remove it from the …

itaybre:
Still WIP

### DIFF
--- a/.github/workflows/assemble-xcframework-variant.yml
+++ b/.github/workflows/assemble-xcframework-variant.yml
@@ -3,7 +3,7 @@ name: "Assemble Sentry Cocoa XCFramework variant"
 on:
   workflow_call:
     inputs:
-      name:
+      scheme:
         description: |-
           The Sentry project target to build an XCFramework slice for.
           Possible values: Sentry, SentrySwiftUI.
@@ -51,9 +51,21 @@ on:
         required: false
         type: string
 
+      excluded-archs:
+        description: |-
+          The archs to exclude from the XCFramework.
+        required: false
+        type: string
+
+      override-name:
+        description: |-
+          The name to use for the XCFramework. If not provided, the default name will be used.
+        required: false
+        type: string
+
 jobs:
   assemble-xcframework-variant:
-    name: Assemble ${{inputs.name}}${{inputs.suffix}} XCFramework Variant
+    name: Assemble ${{ inputs.override-name || format('{0}{1}', inputs.scheme, inputs.suffix) }} XCFramework Variant
 
     runs-on: macos-14
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/assemble-xcframework-variant.yml
```diff
@@ -112,26 +133,44 @@ jobs:
...
+            binary_path="$framework_path/$(basename "$framework_path" .framework)"
+            if [ -f "$binary_path" ]; then
+              echo "Processing binary: $binary_path"
+              echo "Removing architectures: ${{ inputs.excluded-archs }}"
+              temp_binary="${binary_path}.tmp"
+              lipo -remove ${{ inputs.excluded-archs }} "$binary_path" -output "$temp_binary"
+              mv "$temp_binary" "$binary_path"
+              echo "Updated binary: $binary_path"
+            fi
+          done
```
philprime: `m`: We should move this to a script file so we can execute the same logic on our local machines too.

.github/workflows/release.yml
```diff
@@ -73,26 +73,34 @@ jobs:
...
     strategy:
       matrix:
         variant:
-          - name: Sentry
+          - scheme: Sentry
             macho-type: mh_dylib
             suffix: "-Dynamic"
             id: sentry-dynamic
-          - name: Sentry
+            excluded-archs: arm64e
```
philprime, line 93: `h`: Do we have documentation somewhere why we remove the arm64e from the default `Sentry.xcframework` and only add it in `Sentry-Dynamic-WithARM64e`? This seems like a major breaking change.

itaybre: I took this decision because:
- Most apps don't have the arm64e slice, Xcode doesn't build it by default
- This is a breaking change from Xcode anyway

philprime: > This is a breaking change from Xcode anyway

I remember you mentioning this, but we should definitely communicate it. 
Please add a note to the CHANGELOG.md

cc @philipphofmann

philippsuess: Yes, please add a note to the Changelog to clearly communicate this and explain which Xcode versions are impacted. Can it break for customers using old Xcode versions?

itaybre: Added the note to the changelog.
Regarding Xcode versions, it shouldn't affect anyone regarding that specifically. Only whoever is using arm64e , will need to use the new variant

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+> [!Important]
+> Because of Xcode's 26 changes not allowing individual frameworks to contain arm64e slices if the main binary doesn't contain it.
+> We have decided to split the Dynamic variant of Sentry into 2:
```
```suggestion
> Xcode 26 is not allowing individual frameworks to contain arm64e slices anymore if the main binary doesn't contain it.
> We have decided to split the Dynamic variant of Sentry into two variants:
```

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+>
+> - `Sentry-Dynamic`: Without arm64e
+> - `Sentry-Dynamic-WithARM64e`: _With_ ARM64e slice
+>
+> If your app does not need arm64e, you don't need to do any change.
+> But if your app _does need arm64e_ please use `Sentry-Dynamic-WithARM64e` from 8.55.0 so you don't have issues uploading to the AppStore.
```
```suggestion
> If your app does not need arm64e, you don't need to do any changes.
> But if your app _needs arm64e_ please use `Sentry-Dynamic-WithARM64e` from 8.55.0 so you don't have issues uploading to the AppStore.
```

philprime (approved):
LGTM with a nitpick

itaybre:
Cursor review

Issue #5789, opened 2025-08-04
Remove DispatchGroups from UnitTests


philipphofmann:
### Description

The unit tests have a couple of usages of DispatchGroup to wait for async code to finish. This has the major downside that the tests don't fail with a clear assertion if async code doesn't finish within the specified timeout. Therefore, we want to remove all usages of DispatchGroup and add a linter to prevent future usages.

PR #5790, opened 2025-08-04
test: SentryClient use TestExpectations

philipphofmann:
Replace DispatchGroup with TestExpectation and add a comment for one DispatchGroup on why we can't use a TestExpectation.

This is part of https://github.com/getsentry/sentry-cocoa/issues/5789 to replace DispatchGroup with in TestExpectation where possible.

#skip-changelog


philprime (approved):
LGTM

PR #5792, opened 2025-08-04
fix: SentrySDK.feedback unavailable for Extensions

itaybre:
Fixes building on iOS extensions

### DIFF
--- a/Sources/Sentry/Public/SentryFeedbackAPI.h
+++ b/Sources/Sentry/Public/SentryFeedbackAPI.h
@@ -13,7 +13,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 @interface SentryFeedbackAPI : NSObject
-#    if !SDK_V9
+#    if SDK_V9
 SENTRY_NO_INIT
 #    endif
 
--- a/sdk_api.json
+++ b/sdk_api.json
@@ -26950,6 +26950,31 @@
               "Dynamic"
             ],
             "funcSelfKind": "NonMutating"
+          },
+          {
+            "kind": "Constructor",
+            "name": "init",
+            "printedName": "init()",
+            "children": [
+              {
+                "kind": "TypeNominal",
+                "name": "SentryFeedbackAPI",
+                "printedName": "Sentry.SentryFeedbackAPI",
+                "usr": "c:objc(cs)SentryFeedbackAPI"
+              }
+            ],
+            "declKind": "Constructor",
+            "usr": "c:objc(cs)NSObject(im)init",
+            "moduleName": "Sentry",
+            "overriding": true,
+            "implicit": true,
+            "objc_name": "init",
+            "declAttributes": [
+              "Override",
+              "ObjC",
+              "Dynamic"
+            ],
+            "init_kind": "Designated"
           }
         ],
         "declKind": "Class",


Sources/Sentry/Public/SentryFeedbackAPI.h
```diff
@@ -12,9 +12,10 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-NS_EXTENSION_UNAVAILABLE("Sentry User Feedback UI cannot be used from app extensions.")
 @interface SentryFeedbackAPI : NSObject
 
+SENTRY_NO_INIT
```
noahsmartin, line 17: I think this is a good change, but technically API breaking so it might need to be gated behind V9

Sources/Swift/Helper/SentrySDK.swift
```diff
@@ -259,8 +259,15 @@ import Foundation
     
     #if os(iOS) && !SENTRY_NO_UIKIT
     @available(iOS 13.0, *)
+    @available(iOSApplicationExtension, unavailable)
     @objc public static let feedback = {
-      return SentryFeedbackAPI()
+        // When building the SDK, the compiler doesn't add `NS_EXTENSION_UNAVAILABLE` to the header so ObjC,
```
noahsmartin: typo?

Sources/Swift/Helper/SentrySDK.swift
```diff
@@ -259,8 +259,15 @@ import Foundation
...
+        // this ensures that users will at least get a warning that `SentryFeedbackAPI` is not available.
+        if #available(iOSApplicationExtension 13.0, *) {
+            SentrySDKLog.warning("Sentry User Feedback is only available on iOS 13 or later.")
```
noahsmartin: In my testing this evaluates to true even in non-extensions, so everyone would get this warning log?

Sources/Sentry/SentryFeedbackAPI.m
```diff
@@ -16,6 +17,11 @@
 
 @implementation SentryFeedbackAPI
 
++ (instancetype)newInstance
```
philprime, line 20: `l`: Why this convenience initializer?

Sources/Sentry/include/SentryFeedbackAPI+Private.h
```diff
@@ -0,0 +1,17 @@
...
+
+#if TARGET_OS_IOS && SENTRY_HAS_UIKIT
+
+NS_ASSUME_NONNULL_BEGIN
+
+// Helper category to allow us to create an instance of SentryFeedbackAPI` within the SDK.
+
+@interface SentryFeedbackAPI ()
+
++ (instancetype)newInstance;
```
philprime, line 11: `l`: I don't really see the value in having a static method to use `[SentryFeedbackAPI newInstance]` instead of `[[SentryFeedbackAPI alloc] init]`. It doesn't seem like a pattern we use a lot in the SDK.

Sources/Sentry/Public/SentryFeedbackAPI.h
```diff
@@ -12,9 +12,10 @@
...
+
```
philprime, line 19: `m`: We should still mark the methods `showWidget` and `hideWidget` as unavailable in app extensions if UIKit is not available (not always the case, e.g. share extensions can have an UI).

Sources/Sentry/include/SentrySDKInternal.h
```diff
@@ -282,7 +282,8 @@ SENTRY_NO_INIT
 
 #if TARGET_OS_IOS && SENTRY_HAS_UIKIT
 
-@property (nonatomic, class, readonly) SentryFeedbackAPI *feedback;
+@property (nonatomic, class, readonly) SentryFeedbackAPI *feedback NS_EXTENSION_UNAVAILABLE(
```
philprime, line 285: `m`: Why is this marked unavailable? The `SentryFeedbackAPI`?

itaybre: It should not be used from an extension.
Even if it not public, it will help us avoid mistakes

Sources/Sentry/Public/SentryFeedbackAPI.h
itaybre: Fair point

Sources/Sentry/include/SentryFeedbackAPI+Private.h
itaybre: This is to ensure no one creates their own instance of `SentryFeedbackAPI` outside de SDK.
If we add `SENTRY_NO_INIT`, we cant use it either, so I just added a private initializer.

sdk_api.json
```diff
@@ -26950,31 +26950,6 @@
               "Dynamic"
             ],
             "funcSelfKind": "NonMutating"
-          },
-          {
-            "kind": "Constructor",
```
noahsmartin: Should we make this constructor change only in V9? I worry about introducing this since it's a breaking change

Sources/Swift/Helper/SentrySDK.swift
```diff
@@ -259,8 +259,15 @@ import Foundation
     
     #if os(iOS) && !SENTRY_NO_UIKIT
     @available(iOS 13.0, *)
+    @available(iOSApplicationExtension, unavailable)
     @objc public static let feedback = {
-      return SentryFeedbackAPI()
+        // When building the SDK, the compiler doesn't add `NS_EXTENSION_UNAVAILABLE` to the header so
+        // this ensures that users will at least get a warning that `SentryFeedbackAPI` is not available.
+        if Bundle.main.bundlePath.hasSuffix(".appex") {
```
noahsmartin, line 266: Is this necessary? We will already disallow use of the two functions show/hide, so IMO this is redundant and is extra complexity. Basically the compiler will already fail anytime you try to use SentryFeedbackAPI, I think it's ok to let people access the `feedback` param and do nothing with it in an extension if they want

itaybre: I think it is good to let users know about a mistake, but I can remove this

Sources/Sentry/include/SentrySDKInternal.h
philprime: I just noticed my feedback got cut off. My follow-up would have been that `SentryFeedbackAPI` will be extended with additional options down the road, so maybe we should make it accessible so we are not blocking ourselves.

itaybre:
Closing since this is already fixed via https://github.com/getsentry/sentry-cocoa/pull/5813 (released in 8.55.0)

PR #5796, opened 2025-08-05
Remove unnecessary dependency property

noahsmartin:
This property was redundant, and only had one use. Updated it to the new property 

#skip-changelog


philprime (approved):
Nice catch, would say LGTM, but you need to fix the unit tests too.

PR #5798, opened 2025-08-05
fix: Change conditional import of MetricKit

### LINKED ISSUE #5793: Additional dependencies when adding the SDK to `main.mm`
### Platform

iOS

### Environment

Production

### Installed

Manually

### Version

8.54.0

### Xcode Version

16.4

_No response_

### Steps to Reproduce

Adding the Cocoa SDK through the Unity SDK. The resulting `main.mm` looks like this

```
#import <UIKit/UIKit.h> <-- Error when commenting this
#import <MetricKit/MetricKit.h> <-- Error when commenting this
#import <Sentry/Sentry-Swift.h>
#include "SentryOptions.m"
#include <UnityFramework/UnityFramework.h>

int main(int argc, char* argv[])
{
    @autoreleasepool
    {
        SentryOptions* options = getSentryOptions();
        if(options != nil)
        {
            [SentrySDK startWithOptions:options];
        }

        id ufw = UnityFrameworkLoad();
        [ufw runUIApplicationMainWithArgc: argc argv: argv];
        return 0;
    }
}
```

Note the 
```
#import <UIKit/UIKit.h>
#import <MetricKit/MetricKit.h>
```
at the top.

### Expected Result

No additional dependencies to be added.

### Actual Result

Without the additional `imports` the build fails with

```
/Sentry.framework/Headers/Sentry-Swift.h:2985:40 No type or protocol named 'MXMetricManagerSubscriber'
in file included from /main.mm:3

/Sentry.framework/Headers/Sentry-Swift.h:3012:39 Attempting to use the forward class 'UIView' as superclass of 'SentryMaskingPreviewView'
in file included from /main.mm:3

Forward declaration of class here
/Sentry.framework/Headers/Sentry-Swift.h:4139:75 Cannot find protocol declaration for 'UIAdaptivePresentationControllerDelegate'
in file included from /main.mm:3

/Sentry.framework/Headers/Sentry-Swift.h:4233:23 Unknown type name 'UIWindowLevel'
in file included from /main.mm:3

/Sentry.framework/Headers/Sentry-Swift.h:4237:23 Unknown type name 'NSDirectionalRectEdge'
in file included from /main.mm:3

/Sentry.framework/Headers/Sentry-Swift.h:4241:23 Unknown type name 'UIOffset'
in file included from /main.mm:3
```

Maybe :)

philprime:
## :scroll: Description

The import of MetricKit is wrapped using a `#if canImport(MetricKit)` conditional compiler check. If `MetricKit` is not available, we are still using classes of it, causing compiler issues. This PR changes the conditional compilation move the `#endif` to cover classes using MetricKit.

**We need to verify compilation and SDK usage still works when using the SDK on iOS or macOS without MetricKit available**! I do not think we have a sample project for that.

Fixes #5793

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Change conditional import of MetricKit (#5798)
+
 ### Fixes
 
 - Add support for PDFKit views in session replay (#5750)
--- a/Sources/Swift/Core/MetricKit/SentryMXManager.swift
+++ b/Sources/Swift/Core/MetricKit/SentryMXManager.swift
@@ -7,7 +7,6 @@ import Foundation
  */
 #if canImport(MetricKit)
 import MetricKit
-#endif
 
 @available(iOS 15.0, macOS 12.0, macCatalyst 15.0, *)
 @available(tvOS, unavailable)
@@ -88,4 +87,5 @@ import MetricKit
     }
 }
 
+#endif
 #endif


noahsmartin (commented):
I don't think we should change this, the entire file is already wrapped in `#if os(iOS) || os(macOS)` and metric kit is available on these platforms. We should not need the `#if canImport` at all. The only reason it would be necessary is if we wanted to support compiling on versions of Xcode that don't include metric kit in the search path, but it has been included since Xcode 11.

Alternatively, I think we should remove the check entirely: https://github.com/getsentry/sentry-cocoa/pull/5809

philprime:
@bitsandfoxes Removing the `#canImport(MetricKit)` would still require you to import MetricKit in sentry-unity correct?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Change conditional import of MetricKit (#5798)
```
philippsuess, line 5: `m`: I think this shoulb be under the heading ### Fixes.

philipphofmann (commented):
LGTM. Yes,  we should follow up on this and add a sample project
> We need to verify compilation and SDK usage still works when using the SDK on iOS or macOS without MetricKit available! I do not think we have a sample project for that.

philipphofmann (approved):
LGTM

noahsmartin:
Just to double check my earlier comment that this change isn't necessary and the lines can in fact be removed, I downloaded the header file from this PR and [my PR](https://github.com/getsentry/sentry-cocoa/pull/5809) and diffed them, and confirmed they are identical. So we should be fine to merge my PR to remove the unnecessary lines, and close this PR, the `#canImport` here has no effect:

Job for framework built on my PR: https://github.com/getsentry/sentry-cocoa/actions/runs/16749353669/job/47416775698?pr=5809
Job for framework built on Phil's PR: https://github.com/getsentry/sentry-cocoa/actions/runs/16743836918/job/47398738944?pr=5798

```
noahmartin@L4F00WF7KT iOS-Cocoapods-Swift6 % diff /Users/noahmartin/Downloads/Sentry-Dynamic-2.xcframework/ios-arm64_arm64e/Sentry.framework/Headers/Sentry-Swift.h /Users/noahmartin/Downloads/Sentry-Dynamic-3.xcframework/ios-arm64_arm64e/Sentry.framework/Headers/Sentry-Swift.h 
noahmartin@L4F00WF7KT iOS-Cocoapods-Swift6 
```

Maybe I am missing something though? Is there some other part of the build these lines have an effect on?

philprime:
After a sync call with @bitsandfoxes and @noahsmartin we decided to close this PR in favor of #5809.

See https://github.com/getsentry/sentry-cocoa/pull/5809#issuecomment-3160681727

PR #5799, opened 2025-08-05
test: Add MissingTestCase struct

philipphofmann:
Add a struct for documenting the removal of test cases to fix flaky CI.

#skip-changelog


SentryTestUtils/MissingTestCase.swift
```diff
@@ -0,0 +1,20 @@
+/// A struct to represent uncovered test cases that are not implemented.
+public struct MissingTestCase {
```
philprime, line 2: `m`: What do we use for Objective-C tests?

philippsuess:
As discussed today in our call with you @philprime and @kahest, we hopefully don't need this. We'll add it once we do.

PR #5800, opened 2025-08-05
fix: Add null-handling for parsed DSN in SentryHTTPTransport

philprime:

Adds null-handling for the `SentryOptions#parsedDsn` used in the SentryHTTPTransport.

This should actually be considered an invalid SDK state, therefore I added fatal logs.

Open for discussion if deleting the envelope and sending the next one is too drastic.


Sources/Sentry/SentryHttpTransport.m
```diff
@@ -356,14 +357,21 @@ - (void)sendAllCachedEnvelopes
     // We must set sentAt as close as possible to the transmission of the envelope to Sentry.
     rateLimitedEnvelope.header.sentAt = [self.dateProvider date];
 
-    NSError *requestError = nil;
-    NSURLRequest *request = [self.requestBuilder createEnvelopeRequest:rateLimitedEnvelope
-                                                                   dsn:self.options.parsedDsn
-                                                      didFailWithError:&requestError];
+    if (nil == self.options.parsedDsn) {
+        SENTRY_LOG_FATAL(@"No DSN configured, not sending envelope.");
+        [self deleteEnvelopeAndSendNext:envelopeFilePath];
```
philippsuess: `h`: I'm unsure if this change makes sense. If we don't have a valid parsed DSN, this will start deleting the current envelope without sending it, which is new behaviour. I think we should log an error or a warning and return here. Then the logic stays the same.

philipphofmann (commented):
I'm a bit unsure about the approach. I think it changes the logic.

Sources/Sentry/SentryHttpTransport.m
philprime: Do you think this could this cause it trying to process the same envelopes over and over again? If not I would agree with just logging the error and not deleting anything.

philippsuess: Actually, when we don't have a valid DSN you can't use the transport at all. While we could store the envelopes to a fallback path, they will be deleted anyway. I think we could already do an early exit. Ideally, you can only initialize a transport when having a valid DSN. The transport only uses the `self.options.sendClientReports` from the options. Instead of passing in the options in the init method, we could ask for `sendClientReports` and the parsedDSN, but then the class initializing the Transport needs to handle this edge case. The SentryTransportFactory could create a NoOpTransport or simply not add the HTTPTransport to the list here https://github.com/getsentry/sentry-cocoa/blob/51b7dd37d987674400569385cc69e55aaefa1d86/Sources/Sentry/SentryTransportFactory.m#L56-L75

and it could log an error that the DSN is valid. 

But if this is too complicated, we can also come up with an easier solution right now and follow up.

philprime: That would've also been my suggestion, reducing to the minimum required options and ensuring they are non null on init.
I'll update the PR

philippsuess: Thanks 🥳

philprime:
cursor review

Sources/Sentry/SentryTransportFactory.m
```diff
@@ -27,6 +29,13 @@ @implementation SentryTransportFactory
                                sentryFileManager:(SentryFileManager *)sentryFileManager
                                       rateLimits:(id<SentryRateLimits>)rateLimits
 {
+    if (!options.parsedDsn) {
+        SENTRY_LOG_FATAL(@"Failed to create transports because the SentryOptions does not contain "
+                         @"a parsed DSN.");
+        return @[];
+    }
+    SentryDsn *_Nonnull dsn = SENTRY_UNWRAP_NULLABLE(SentryDsn, options.parsedDsn);
```
philippsuess: `m`: The Spotlight transport doesn't require a DSN. You could use it without a DSN. Currently, the client drops everything if no DSN is set. So ideally, we would allow this in the client, but we don't need to change this now. Instead, we should only log a warning or even only a debug. I also think some users set a wrong DSN on purpose to avoid sending any data. The fatal error message could be confusing to some users I believe. Also we already log an error message somewhere else if we couldn't parse the DSN.

philipphofmann (approved):
LGTM, with a comment on the log message.

Sources/Sentry/SentryTransportFactory.m
philprime: I refactored the entire factory and added additional tests. It now returns the spotlight transport if enabled and the http transport if successfully created.

Sources/Sentry/SentryTransportFactory.m
```diff
@@ -64,18 +69,39 @@ @implementation SentryTransportFactory
...
-        return @[ httpTransport, spotlightTransport ];
+        [transports addObject:httpTransport];
     } else {
-        return @[ httpTransport ];
+        SENTRY_LOG_WARN(
+            @"Failed to create HTTP transport because the SentryOptions does not contain "
+            @"a parsed DSN.");
     }
+
+    return transports;
```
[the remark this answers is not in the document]
philprime: @philipphofmann do you think this could actually become a problem?

philprime:
Reposting because Cursor Bot marks comments as outdated:
https://github.com/getsentry/sentry-cocoa/pull/5800#discussion_r2269292292

@philipphofmann do you think this is valid feedback?

> The SentryTransportFactory no longer creates an SentryHttpTransport when options.parsedDsn is nil. This removes the mechanism that previously eagerly processed and deleted cached envelopes on startup. As a result, cached envelopes are never processed or purged, leading to unbounded cache accumulation on disk, particularly when the DSN is temporarily disabled or misconfigured.

philippsuess:
> Reposting because Cursor Bot marks comments as outdated: [#5800 (comment)](https://github.com/getsentry/sentry-cocoa/pull/5800#discussion_r2269292292)
> 
> @philipphofmann do you think this is valid feedback?
> 
> > The SentryTransportFactory no longer creates an SentryHttpTransport when options.parsedDsn is nil. This removes the mechanism that previously eagerly processed and deleted cached envelopes on startup. As a result, cached envelopes are never processed or purged, leading to unbounded cache accumulation on disk, particularly when the DSN is temporarily disabled or misconfigured.

Yes, this is valid feedback. If you previously had `options.parsedDsn` being nil, it helped cleaning up some old envelopes. But the client discards most data anyways when you don't have a dsn set https://github.com/getsentry/sentry-cocoa/blob/ed85746bba3b980b01a324ebc48f9cc8a7c77340/Sources/Sentry/SentryClient.m#L923-L926

We could add a cleanup call to the file manager or anywhere else that deletes envelopes in the fallback folder when the DSN is nil, but it's also some overhead for everybody checking if the folder exists and has files. What would you suggest.

philprime:
But isn't that what you wanted me to change [in this PR comment](https://github.com/getsentry/sentry-cocoa/pull/5800/commits/0f0443d105abad3180f8da8dc3e6f68a47bb7f0d#r2254581106) when you gave the first round of feedback?

philippsuess:
We can keep this. We have a safeguard that I forgot about here https://github.com/getsentry/sentry-cocoa/blob/32584c4cb5ea7c3d6d9251f386cc0f05413a739d/Sources/Sentry/SentryClient.m#L166-L168

I thought that we had the safeguard in the transport, but it's in the client. Sorry my bad.

Before this PR, when you didn't set a DSN, the client should not have stored any envelopes to disk via the transport. I thought that when we remove the transport, we stop cleaning up old envelopes at some point when not setting a DSN, because I thought the logic was in the transport. But it's in the client so all good. The logic stays the same.

philprime:
Alright in that case I'll merge it as-is.

PR #5803, opened 2025-08-05
fix: Add null-handling for sentry level parsing

philprime:

The function `sentryLevelForString` is used to map a `NSString` to a `SentryLevel` and expects a non-null input value. 

This is an important fix, because the implementation uses `SentryLevelHelper` which is a Swift file, and while the Objective-C code can deal with null values, in Swift it would actually be considered force-unwrapped and crash.

https://github.com/getsentry/sentry-cocoa/blob/1936411584fc75124f877a9698f1fbb66989b6c3/Sources/Sentry/SentryLevelMapper.m#L5-L9

https://github.com/getsentry/sentry-cocoa/blob/1936411584fc75124f877a9698f1fbb66989b6c3/Sources/Swift/Core/Helper/Log/SentryLevel.swift#L38-L47

This is something we must keep in mind when bridging Objective-C files to Swift (cc @noahsmartin as you are currently most active in this field).

#skip-changelog


Sources/Sentry/SentryHub.m
```diff
@@ -774,13 +775,17 @@ - (BOOL)envelopeContainsEventWithErrorOrHigher:(NSArray<SentryEnvelopeItem *> *)
...
+            if (nullableEventJson == nil) {
                 return NO;
             }
+            NSDictionary *_Nonnull eventJson
+                = SENTRY_UNWRAP_NULLABLE(NSDictionary, nullableEventJson);
 
-            SentryLevel level = sentryLevelForString(eventJson[@"level"]);
+            // Fallback to error level if no level is set. This ensures that the parameter is a
+            // non-null value, while delegating the fallback to the function `sentryLevelForString`.
+            SentryLevel level = sentryLevelForString(eventJson[@"level"] ?: @"unknown");
```
philippsuess: `l`: Maybe we should even change the string we're passing to something like this, so it's clearer. I usually read the code before reading the comment.
```suggestion
            SentryLevel level = sentryLevelForString(eventJson[@"level"] ?: @"use_default_of_ sentryLevelForString");
```

Sources/Sentry/SentrySerialization.m
```diff
@@ -308,7 +309,9 @@ + (SentryLevel)levelFromData:(NSData *)eventEnvelopeItemData
         return kSentryLevelError;
     }
 
-    return sentryLevelForString(eventDictionary[@"level"]);
+    // Fallback to error level if no level is set. This ensures that the parameter is a non-null
+    // value, while delegating the fallback to the function `sentryLevelForString`.
+    return sentryLevelForString(eventDictionary[@"level"] ?: @"unknown");
```
philippsuess: `m`: Now that I see this in two places, maybe we should allow passing nil to `sentryLevelForString` which then uses the default of error when receiving nil. Then we don't have to unwrap in two places with a comment explaining it defaults to error.

philipphofmann (approved):
LGTM

Sources/Sentry/SentrySerialization.m
philprime: I agree, allowing `nil` as a parameter seems cleaner, I'll update the PR

Sources/Sentry/SentrySerialization.m
```diff
@@ -5,6 +5,7 @@
 #import "SentryEnvelopeAttachmentHeader.h"
 #import "SentryEnvelopeItemType.h"
 #import "SentryError.h"
+#import "SentryInternalDefines.h"
```
philprime:
```suggestion
```

Tests/SentryTests/Helper/SentryLevelMapperTests.swift
```diff
@@ -0,0 +1,25 @@
+@testable import Sentry
+import XCTest
+
+class SentryLevelMapperTests: XCTestCase {
```
philippsuess, line 4: Thanks for the extra tests 🥇

philipphofmann (approved):
Again LGTM

PR #5804, opened 2025-08-05
fix: Add null-handling for debug meta

philprime:

Adds explicit null-handling for debug metadata. This is necessary so we do not use `nil` as a key. While this most likely will not cause issues in Objective-C, we should see it as a best practice instead.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryMetricKitIntegration.m
+++ b/Sources/Sentry/SentryMetricKitIntegration.m
@@ -509,7 +509,9 @@ NS_ASSUME_NONNULL_BEGIN
         uint64_t imageAddress = mxFrame.address - mxFrame.offsetIntoBinaryTextSegment;
         debugMeta.imageAddress = sentry_formatHexAddressUInt64(imageAddress);
 
-        debugMetas[debugMeta.debugID] = debugMeta;
+        if (nil != debugMeta.debugID) {
+            debugMetas[SENTRY_UNWRAP_NULLABLE(NSString, debugMeta.debugID)] = debugMeta;
+        }
     }
 
     return [debugMetas allValues];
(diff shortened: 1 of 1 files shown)


Sources/Sentry/SentryMetricKitIntegration.m
```diff
@@ -509,7 +509,9 @@ - (SentryStacktrace *)convertMXFramesToSentryStacktrace:(NSEnumerator<SentryMXFr
         uint64_t imageAddress = mxFrame.address - mxFrame.offsetIntoBinaryTextSegment;
         debugMeta.imageAddress = sentry_formatHexAddressUInt64(imageAddress);
 
-        debugMetas[debugMeta.debugID] = debugMeta;
+        if (nil != debugMeta.debugID) {
+            debugMetas[SENTRY_UNWRAP_NULLABLE(NSString, debugMeta.debugID)] = debugMeta;
+        }
```
philippsuess: `m`: We should be able to simplify this to
```suggestion
        debugMetas[binaryUUID] = debugMeta;
```

As the `debugMeta.debugID` is basically the binaryUUID.
https://github.com/getsentry/sentry-cocoa/blob/b2c7769cc38bd692aee5c6ea101589600b97a1a7/Sources/Sentry/SentryMetricKitIntegration.m#L499-L512

philipphofmann (approved):
LGTM, but I think we should be able to simplify this.

Sources/Sentry/SentryMetricKitIntegration.m
philprime: Thanks for pointing this out, I'll update the PR.

PR #5806, opened 2025-08-05
fix: Add null-handling for value in profile timeseries

philprime:

Adds silencing of a null-handling warning by ignoring it, because Objective-C dictionaries should be able to handle null.

#skip-changelog


Sources/Sentry/SentryProfileTimeseries.m
```diff
@@ -114,7 +115,9 @@
 
         [slicedGPUEntries addObject:@ {
             @"elapsed_since_start_ns" : sentry_stringForUInt64(relativeTimestamp),
-            @"value" : obj[@"value"],
+            // We unwrap the nullable to silence the compiler warning, as Objective-C will just
+            // ignore the null value.
+            @"value" : SENTRY_UNWRAP_NULLABLE(NSNumber, obj[@"value"]),
```
philippsuess: `h`: You actually found a bug here. When `obj[@"value"]` is `nil`, the code crashes with something like
```
 attempt to insert nil object from objects[0] (NSInvalidArgumentException)
```

I think we should not add the value when `obj[@"value"]` is `nil` to avoid the crash.

philprime: Oh really? Thanks for double-checking, as you can see in my comment I expected it to behave differently.

philippsuess: I know this because other SDK maintainers and I already caused multiple SDK crashes by that 😅

philprime:
cursor review

Sources/Sentry/SentryProfileTimeseries.m
```diff
@@ -112,15 +113,20 @@
...
-            @"value" : obj[@"value"],
-        }];
+        NSMutableDictionary *entry =
+            [@ { @"elapsed_since_start_ns" : sentry_stringForUInt64(relativeTimestamp) }
+                mutableCopy];
+        NSNumber *value = obj[@"value"];
+        if (value != nil && ![value isKindOfClass:[NSNull class]]) {
+            entry[@"value"] = value;
+        }
+        [slicedGPUEntries addObject:entry];
```
philippsuess, line 123: `l`: I would maybe double-check if an entry without a value actually makes sense. If not, we could already drop it. But I guess there should also be no big harm if we send it.

Tests/SentryProfilerTests/SentryProfileTimeseriesTests.m
```diff
@@ -0,0 +1,289 @@
...
+
+    // Verify first entry
+    XCTAssertEqualObjects(result[0][@"elapsed_since_start_ns"], @"500000000");
+    XCTAssertEqualObjects(result[0][@"value"], @(60.0));
+
+    // Verify second entry
+    XCTAssertEqualObjects(result[1][@"elapsed_since_start_ns"], @"1000000000");
+    XCTAssertEqualObjects(result[1][@"value"], @(30.0));
+
+    // Verify third entry
```
philippsuess: `l`: A bit verbose
```suggestion
    XCTAssertEqualObjects(result[0][@"elapsed_since_start_ns"], @"500000000");
    XCTAssertEqualObjects(result[0][@"value"], @(60.0));

    XCTAssertEqualObjects(result[1][@"elapsed_since_start_ns"], @"1000000000");
    XCTAssertEqualObjects(result[1][@"value"], @(30.0));

```

Sources/Sentry/SentryProfileTimeseries.m
philprime: I double-checked and found that only the `SentryFramesTracker` is recording new entries, with the value set to `SentryFramesTracker.currentFrameRate`, which is a non-null value.

Therefore I believe this might never be `nil` or `NSNull`, but because it is technically possible, I'll keep the implementation as is.

Ideally we wouldn't use a Dictionary for the entires but instead a data structure, but that's out of scope for this PR.

philprime:
@codecov-ai-reviewer test

ghost:
On it! Codecov is generating unit tests for this PR.

PR #5808, opened 2025-08-05
ref: Make SentryEventDecodable internal in V9

noahsmartin:
Implements the note in this comment for V9. The initializer has to be duplicated because a required initializer cannot be declared in an extension

#skip-changelog


Sources/Swift/Protocol/Codable/SentryEventCodable.swift
```diff
@@ -18,39 +16,73 @@ import Foundation
...
+        }
+        
+        self.stacktrace = try container.decodeIfPresent(SentryStacktraceDecodable.self, forKey: .stacktrace)
+        
+        if let rawDebugMeta = try container.decodeIfPresent([String: [DebugMetaDecodable]].self, forKey: .debugMeta) {
+            self.debugMeta = rawDebugMeta["images"]
+        }
+        
+        self.breadcrumbs = try container.decodeIfPresent([BreadcrumbDecodable].self, forKey: .breadcrumbs)
+        self.request = try container.decodeIfPresent(SentryRequestDecodable.self, forKey: .request)
```
philippsuess: `m`: This code is now duplicate. Can't we create a common private init method or something else to ditch the duplication?

philipphofmann (approved):
LGTM, but if you can wait after August 13th to merge this.

Sources/Swift/Protocol/Codable/SentryEventCodable.swift
noahsmartin: unfortunately no, because a `required` initializer cannot call other initializers. We could define an entire new private type that does the decoding and have this type wrap it. But I worry about creating too many wrappers (complexity and SDK size)

philippsuess: No, a private type is an overkill. But the event properties are all mutable I believe. We should be able to use a helper method that we call from the init methods.

noahsmartin: Good idea, updated

Sources/Swift/Protocol/Codable/SentryEventCodable.swift
```diff
@@ -51,11 +66,7 @@ open class SentryEventDecodable: Event, Decodable {
         case request
     }
     
-    required convenience public init(from decoder: any Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        
-        self.init()
-
+    private func writePropertiesFrom(container: KeyedDecodingContainer<CodingKeys>) throws {
```
philippsuess, line 69: Way better thanks

PR #5809, opened 2025-08-05
fix: Remove unnecessary metric kit check

noahsmartin:
As an alternative to https://github.com/getsentry/sentry-cocoa/pull/5798 I think we should just remove this conditional check because the whole file is already wrapped in `#if os(iOS) || os(macOS)`. MetricKit can only not be imported on watchOS (according to the [docs](https://developer.apple.com/documentation/metrickit)) so this check was redundant

#skip-changelog


noahsmartin:
Synced with @philprime and @bitsandfoxes about this and decided to merge this one. We can also change metric kit to be `@_implementationOnly` if we decide that's necessary later

PR #5811, opened 2025-08-05
test: Fix flaky Flush_BlocksCallingThread_TimesOut

philipphofmann:
Make the test requirements less strict by removing the assert of the blocking duration and adding a test expectation timing out after 10 seconds to prevent hanging CI if the flush blocks longer.

The test failed here https://github.com/getsentry/sentry-cocoa/actions/runs/16743553141/job/47396825598


philprime (approved):
LGTM

philippsuess:
Unrelated unit and ui tests failed. I'm still merging this as a rerun would be a waste of time.

PR #5813, opened 2025-08-05
ci: Add test for building in keyboard extension with cocoapods

noahsmartin:
This adds an app extension to the cocoa pods test app. It fails to build without the other changes in this PR. Mainly, it sets the `APPLICATION_EXTENSION_API_ONLY` config to NO in the podfile. Without this cocoapods would automatically add it and set it to YES when the framework was linked with app extensions. We can set it to NO and instead add app extension availability flags in the public API, to document to our users that some of our API cannot be called from an app extension.

The main reason we need `APPLICATION_EXTENSION_API_ONLY=NO` is because the feedback button can be configured to show automatically on SDK start, and that requires `UIApplication.shared`. Since that codepath goes through SentrySDK.start() we cannot annotate it with `NS_EXTENSION_UNAVAILABLE`. We could make another API like `SentrySDK.configureUserFeedbackButton()` which is extension unavailable, but the fix in this PR should also work and keep the same public API.

#skip-changelog

### DIFF
--- a/Samples/iOS-Cocoapods-Swift6/TestKeyboardExtension/KeyboardViewController.swift
+++ b/Samples/iOS-Cocoapods-Swift6/TestKeyboardExtension/KeyboardViewController.swift
@@ -1,61 +1,6 @@
-//
-//  KeyboardViewController.swift
-//  TestKeyboardExtension
-//
-//  Created by Noah Martin on 8/5/25.
-//
-
 import Sentry
 import UIKit
 
 class KeyboardViewController: UIInputViewController {
-
-    @IBOutlet var nextKeyboardButton: UIButton!
-    
-    override func updateViewConstraints() {
-        super.updateViewConstraints()
-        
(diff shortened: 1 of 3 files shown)


Samples/iOS-Cocoapods-Swift6/TestKeyboardExtension/KeyboardViewController.swift
```diff
@@ -0,0 +1,61 @@
+//
+//  KeyboardViewController.swift
+//  TestKeyboardExtension
+//
+//  Created by Noah Martin on 8/5/25.
```
noahsmartin: I will clean up this header and the rest of the file if we go with this approach, this is just the default file Xcode gave me when I created a new target. The only important part of this file is the `import Sentry`

philippsuess:
I leave this PR to @itaybre or @philprime, because you're more involved in the app extension topic and you should be able to give a more meaningful review.

PR #5814, opened 2025-08-05
style: Run pre-commit-hooks on all files

philprime:
I ran `pre-commit run --all-files` and found these fails were not formatted properly.

#skip-changelog


philprime:
I agree this is unexpected. I actually noticed it because I was able to commit a check with local verification, which then failed in CI. I'll use this PR to investigate the formatting setup.

philprime:
So I believe most of these weren't format checked because our CI checks are selective based on the path, and they weren't touched in a while.

philipphofmann (approved):
Again LGTM

PR #5817, opened 2025-08-05
feat: Add subspec for AppExtensions

itaybre:
Added a new subpec so we can set a precompiler macro for App Extensions


Samples/iOS-Cocoapods-Swift6/Podfile
```diff
@@ -6,5 +6,5 @@ target 'App' do
 end
 
 target 'TestKeyboardExtension' do
-  pod 'Sentry', :path => "../../"
+  pod 'Sentry/AppExtension', :path => "../../"
```
noahsmartin, line 9: I don't think we should do this, it's an API breaking change that would require anyone using Sentry to change their pod files after we release it

noahsmartin: It would also increase the size of apps and create risks of classes being implemented in two places. Because the extension would need to link to a different dylib than the main app

itaybre: By default that's the case for CocoaPods, you have the SDK twice within your bundle

itaybre:
Closing since this is already fixed

PR #5819, opened 2025-08-06
test: Fix flaky SentryThreadInspectorTests

philipphofmann:
Change the assertion approach of
testGetCurrentThreadsWithStacktrace_WithSymbolication and testGetCurrentThreadsWithStacktrace_WithoutSymbolication to validate the percentage of threads having a stracktrace, which is more robust than the previous method which failed if more than one thread had no stacktrace frames. Especially during the test when launching multiple threads for the concurrent DispatchQueue it can occur that more than one thread have no stacktrace frames yet, because they just started.

The tests failed here https://github.com/getsentry/sentry-cocoa/actions/runs/16760984375/job/47456025143

#skip-changelog

### DIFF
--- a/Tests/SentryTests/SentryCrash/SentryThreadInspectorTests.swift
+++ b/Tests/SentryTests/SentryCrash/SentryThreadInspectorTests.swift
@@ -1,3 +1,4 @@
+@_spi(Private) @testable import Sentry
 import SentryTestUtils
 import XCTest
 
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/SentryCrash/SentryThreadInspectorTests.swift
```diff
@@ -54,32 +55,43 @@ class SentryThreadInspectorTests: XCTestCase {
...
+                    threadsWithStackTraceFrames += 1
                 }
-                
-                XCTAssertTrue(threadsWithFrames > 1, "Not enough threads with frames")
-                
+
+                let percantageWithStacktraceFrames = Double(threadsWithStackTraceFrames) / Double(threads.count)
+
+                // During testing we usually have around 90% to 100%
+                // We choose a bit lower threshold to avoid flaky tests in CI
```
philprime, line 88: `m`: It might be worth the effort to add the reasoning why there are threads with absent stack traces, as you described in the PR documentation, or at least a link to this PR so we can find it again.

philippsuess: Yes totally makes sense. Thanks for the hint.

Issue #5667, opened 2025-07-21
Unable to build with Xcode 26 for visionOS


levinli303:
### Platform

visionOS

### Environment

Other

### Installed

Swift Package Manager

### Version

8.53.2

### Xcode Version

26.0 beta 3

### Did it work on previous versions?

16.4

### Steps to Reproduce

1. create an Xcode project with iOS target and add visionOS support
2. add code that starts Sentry
3. build the project

### Expected Result

Should build fine

### Actual Result

The build fails in `Compiling Swift module Sentry`

```
<unknown>:0: error: C99 was enabled in PCH file but is currently disabled
<unknown>:0: error: module file /Users/[redacted]/Library/Developer/Xcode/DerivedData/ModuleCache.noindex/SwiftShims-8E9KY877QCTXDJCE0XIKBPGI7.pcm cannot be loaded due to a configuration mismatch with the current compilation
<unknown>:0: error: missing required module 'SwiftShims'
```

philippsuess:
I just tried compiling a sample project including sentry-cocoa 8.53.2 via SPM on Xcode 26 beta 3 and it worked.

You can often resolve errors with `SwiftShims` by doing a clean build or clearing the `~/Library/Developer/Xcode/DerivedData` folder. Please let us know if that works.

vedlai:
I have the same issue, 8.53.2, Xcode 26, Tahoe,  Swift 6. Cleaned Derived Data and Clean Build.

This might be of help, seems similar environment.

https://github.com/firebase/firebase-ios-sdk/issues/14950

philippsuess:
@vedlai, thanks for the hint. I tried using Sentry-Cocoa 8.53.2 with Xcode 26 beta.3 and Swift 6 on macOS 15.5, and was unable to reproduce the issue. It could be related to macOS Tahoe. Could you maybe provide us with a small sample app, so we can reproduce it?

vedlai:
@philipphofmann I think it is Tahoe and I think it is any visionOS app. I started a blank app and got the same error.

I think it is related to Tahoe because the Firebase example was Tahoe as well.

It seems to be isolated to Sentry and not Sentry SwiftUI. 

Clarification: Sentry SwiftUI does not work either but I think it is because Sentry doesn't work.

philippsuess:
> I started a blank app and got the same error.

@vedlai, do you mean a blank app with or without Sentry?

vedlai:
[...]

PR #5823, opened 2025-08-06
meta: Enable interoperability mode for Swift 5.9

itaybre:
Changes the GCC C language to compiler default.

Aims to fix: #5667


philipphofmann (commented):
We must do the same for the new Package file
https://github.com/getsentry/sentry-cocoa/blob/48e5c8a5aeeb44c7fa4f348b6db6be53f1a19099/.github/workflows/release.yml#L134-L139

We must trigger an alpha release after merging this to ensure this doesn't break the release, please.

Also, I see a high risk that both Package files get out of sync. Do you have any ideas how we can avoid this?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Fix Infinite Session Replay Processing Loop (#5765)
 - Fix memory leak in SessionReplayIntegration (#5770)
 - Fix reporting of energy used while profiling (#5768)
+- Fix linking with SentrySwiftUI on Xcode 26 for visionOS (#5823)
```
philprime: `m`: We should add a callout to the top of the changelog so a user knows how to use the swift-5.9 version. WDYT?

Package@swift-5.9.swift
```diff
@@ -0,0 +1,102 @@
...
+    .library(name: "Sentry", targets: ["Sentry"]),
+    .library(name: "Sentry-Dynamic", targets: ["Sentry-Dynamic"]),
+    .library(name: "SentrySwiftUI", targets: ["Sentry", "SentrySwiftUI"])
+]
+
+var targets: [Target] = [
+    .binaryTarget(
+        name: "Sentry",
+        url: "https://github.com/getsentry/sentry-cocoa/releases/download/8.54.0/Sentry.xcframework.zip",
+        checksum: "f0f4b5582f7fef029ddedd22b6b9894a6d2d31ea9c7b52a2133b1ec7c4cff206" //Sentry-Static
```
philprime, line 25: `h`: We must look into the bump-version-util to also update this Package swift accordingly

itaybre: Yes, but I wanted to do this on a different PR

CHANGELOG.md
itaybre: It is automatic, the users don't have to change anything

itaybre:
> We must do the same for the new Package file
> 
> https://github.com/getsentry/sentry-cocoa/blob/48e5c8a5aeeb44c7fa4f348b6db6be53f1a19099/.github/workflows/release.yml#L134-L139
> 
> We must trigger an alpha release after merging this to ensure this doesn't break the release, please.
> 
> Also, I see a high risk that both Package files get out of sync. Do you have any ideas how we can avoid this?

Unfortunately not many, I tried extracting the code to a common file, but there is no way to import / use them :(

Package@swift-5.9.swift
itaybre: Here: https://github.com/getsentry/sentry-cocoa/pull/5839

Package@swift-5.9.swift
```diff
@@ -0,0 +1,102 @@
+// swift-tools-version:5.9
```
philippsuess, line 1: `m`: I think it would be good to point out in a code comment why we need this extra Package.swift file.

philipphofmann (commented):
> Unfortunately not many, I tried extracting the code to a common file, but there is no way to import / use them :(

@itay, what we could do is to store the current git diff between those two files in the repo by using 
```
git diff --no-index Package.swift Package@swift-5.9.swift > expected.patch
```

And then we could validate that the files don't change with something like in CI
```
git diff --no-index Package.swift Package@swift-5.9.swift > current.patch
diff expected.patch current.patch
```

If we do this, I think we should also add a make command that updates the patch for you and the error message in CI should reference this command.

Up to you if you believe it's worth the effort. Or we find some way to validate both package files by using SPM. That would actually be better cause it's an integration test. Or we could also do both. Patch and integration test. Up to you to decice.

itaybre:
> > Unfortunately not many, I tried extracting the code to a common file, but there is no way to import / use them :(
> 
> @itay, what we could do is to store the current git diff between those two files in the repo by using
> 
> ```
> git diff --no-index Package.swift Package@swift-5.9.swift > expected.patch
> ```
> 
> And then we could validate that the files don't change with something like in CI
> 
> ```
> git diff --no-index Package.swift Package@swift-5.9.swift > current.patch
> diff expected.patch current.patch
> ```
> 
> If we do this, I think we should also add a make command that updates the patch for you and the error message in CI should reference this command.
> 
> Up to you if you believe it's worth the effort. Or we find some way to validate both package files by using SPM. That would actually be better cause it's an integration test. Or we could also do both. Patch and integration test. Up to you to decice.

Implemented the linter here: https://github.com/getsentry/sentry-cocoa/pull/5854

Package@swift-5.9.swift
```diff
@@ -79,7 +82,7 @@ if let env = env, String(cString: env, encoding: .utf8) == "1" {
             dependencies: ["SentrySwift"],
             path: "Sources",
             exclude: ["Sentry/SentryDummyPublicEmptyClass.m", "Sentry/SentryDummyPrivateEmptyClass.m", "Swift", "SentrySwiftUI", "Resources", "Configuration"],
-            publicHeadersPath: "Sentry/Public",
+            publicHeadersPath: "",
```
philprime, line 85: `m`: Is there a difference to removing this public headers path completely? Why is it not needed here anymore?

itaybre: If I remove it, it fails to build with this error:
```
public headers ("include") directory path for 'SentryObjc' is invalid or not contained in the target
```
Regarding `Sentry/Public`, it works.
I just decided to keep as similar to Package.swift as possible (`publicHeadersPath` is not in there)

PR #5824, opened 2025-08-06
ci: Fix flaky flush test

noahsmartin:
Fixed this flaky test. It had two flakes:

1. All messages could be sent before we even called flush, fixed that by moving the waitForResponseDispatchGroup = true further up.
2. The flush call could have timed out before the following concurrent calls to flush got scheduled. Fixed this by making the responseDispatchGroup only exit after all the other calls have been made. I changed the original call to expect success not timeout because otherwise the timeout would have to be very large due to it needing to block long enough for these other calls to flush to finish, and I didn't want to make this test take a long time. What that first call returned wasn't important to the test, what is important is that subsequent calls return `alreadyFlushing` which is still what is tested

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
+++ b/Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
@@ -96,11 +96,13 @@ final class SentryHttpTransportFlushIntegrationTests: XCTestCase {
         waitForEnvelopeToBeStored(dispatchQueueWrapper)
         requestManager.returnResponse(response: HTTPURLResponse())
 
-        let flushTimeout = 0.1
+        // This must be long enough that all the threads we start below get to run
+        // while the first call to flush is still blocking
+        let flushTimeout = 10.0
         requestManager.waitForResponseDispatchGroup = true
         requestManager.responseDispatchGroup.enter()
 
-        let allFlushCallsGroup = DispatchGroup()
+        let initialFlushCallGroup = DispatchGroup()
         let ensureFlushingGroup = DispatchGroup()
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
```diff
@@ -96,24 +102,19 @@ final class SentryHttpTransportFlushIntegrationTests: XCTestCase {
         waitForEnvelopeToBeStored(dispatchQueueWrapper)
         requestManager.returnResponse(response: HTTPURLResponse())
 
-        let flushTimeout = 0.1
-        requestManager.waitForResponseDispatchGroup = true
-        requestManager.responseDispatchGroup.enter()
-
-        let allFlushCallsGroup = DispatchGroup()
+        let initialFlushCallGroup = DispatchGroup()
```
philippsuess, line 105: `m`: I currently remove all DispatchGroups from the tests and replace them with XCTestExpectations cause these fail when they time out and we can also configure them to fail when they get called too often. This is part of https://github.com/getsentry/sentry-cocoa/issues/5789. But we can also follow up on this.

philipphofmann (approved):
Thanks @noahsmartin. I'm already too blind to see the bugs in my test code. I spent too much time with it and it's great to get some fresh perspective.

Tests/SentryTests/Networking/SentryHttpTransportFlushIntegrationTests.swift
noahsmartin: Good point that would be better, I'll go ahead and merge this since it's passing CI and will help get other things passing faster, but agree we should switch these

PR #5829, opened 2025-08-07
test: Use XCTestExpectation in HttpTransportTests

philipphofmann:
Replace the dispatch group with an XCTestExpectation to get a failure message when the wait times out.

Contributes to GH-5789

#skip-changelog


philprime (approved):
LGTM

PR #5838, opened 2025-08-07
chore(ci): Update concurrency configuration across workflows

philprime:
This pull request refactors concurrency management across all GitHub Actions workflow files to improve CI efficiency and resource usage. 

- All workflow files now use `concurrency.group: ${{ github.workflow }}-${{ github.ref }}` to prevent duplicate runs of the same workflow on the same branch/commit, replacing the previous fallback logic.
- Cancellation logic is now customized per workflow: for pull requests, in-progress runs are always cancelled; for main branch pushes and scheduled runs, cancellation is disabled to ensure complete baselines and release integrity. 
- The benchmarking workflow now includes a step to forcibly cancel SauceLabs jobs if the workflow is cancelled, preventing resource leaks and unnecessary external charges.
- Each workflow now contains explanatory comments detailing the concurrency strategy and rationale for cancellation behavior.
- Security (`codeql-analysis.yml`) and release (`release.yml`) workflows now have logic to never cancel scheduled or main branch runs, ensuring complete security and release validation.

cursor review

#skip-changelog


philprime:
cursor review

.github/workflows/api-stability.yml
```diff
@@ -16,6 +16,15 @@ on:
       - "sdk_api.json"
       - "sdk_api_v9.json"
 
+# Concurrency configuration:
```
philippsuess, line 19: Thanks for the comment

.github/workflows/benchmarking.yml
```diff
@@ -24,10 +24,16 @@ on:
       - ".github/workflows/build-xcframework-variant-slices.yml"
       - ".github/workflows/assemble-xcframework-variant.yml"
 
-# https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value
+# Concurrency configuration:
+# - We use workflow-specific concurrency groups to prevent multiple benchmark runs on the same code,
+#   as benchmarks are extremely resource-intensive and require dedicated device time on SauceLabs.
+# - For pull requests, we cancel in-progress runs when new commits are pushed to avoid wasting
+#   expensive external testing resources and provide timely performance feedback.
+# - For main branch pushes, we never cancel benchmarks to ensure we have complete performance
```
philippsuess, line 32: `l`: Thanks for the explanation. The strategy is basically cancel everything which isn't a push to the main branch. I think we could simplify the comment or mention this in the comment. The same applies also to other `.yml` files

.github/workflows/benchmarking.yml
```diff
@@ -171,8 +177,40 @@ jobs:
...
+            --verbose \
+            2>&1 | tee retry-output.log
+
+      - name: Force Cancel SauceLabs Job on Workflow Cancellation
+        if: ${{ cancelled() }}
+        env:
+          SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
+        run: |
+          echo "::warning Workflow was cancelled, attempting to cancel SauceLabs jobs"
```
philippsuess: If this really works, it will save a lot of SauceLabs time and should speed up our CI. Thanks 👍

.github/workflows/benchmarking.yml
philprime: That wouldn't be entirely true, because for some cases we also do not want to cancel scheduled or workflow_dispatch runs.

philippsuess: Sorry, I meant the strategy in the code below is basically everything which isn't a push to the main branch.

```
cancel-in-progress: ${{ !(github.event_name == 'push' && github.ref == 'refs/heads/main') }}
```

Or am I mistaken

.github/workflows/benchmarking.yml
philprime: I just double-checked, and it looks like this approach isn't working as expected.

https://github.com/getsentry/sentry-cocoa/actions/runs/16824685121/job/47663071992

I'll iterate it or remove it if I can't find a good way to cancel them.

.github/workflows/benchmarking.yml
philprime: Ah yes, in this case you are right, that's also what the Cursor bot raised. I'll change that to `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` where applicable

.github/workflows/benchmarking.yml
```diff
@@ -186,31 +186,101 @@ jobs:
...
-          }
-
-          # Cancel jobs from both possible log files
-          cancel_jobs_from_log "output.log"
-          cancel_jobs_from_log "retry-output.log"
-
-          echo "SauceLabs job cancellation attempts completed"
+        with:
+          script: |
+            const fs = require('fs');
```
philippsuess: `m`: Wouldn't it be better to extract this into the `scripts/` folder, so you can also run it manually?

philipphofmann (approved):
LGTM

.github/workflows/benchmarking.yml
philprime: I agree, especially because it is not working properly right now, so local debugging is easier.

philprime:
The saucelabs helper grew quite in size, I'll move it to a different PR.

philprime:
Moved SauceLabs cancelling to https://github.com/getsentry/sentry-cocoa/pull/5873

PR #5842, opened 2025-08-07
ci: Open Simulator GUI on CI

itaybre:
After reviewing screenshots on https://github.com/getsentry/sentry-cocoa/actions/runs/16805606320/job/47597097673, I see that the simulator is launched in background.

Having it in the foreground will help the screenshots to capture relevant data.

### DIFF
--- a/scripts/ci-boot-simulator.sh
+++ b/scripts/ci-boot-simulator.sh
@@ -59,6 +59,8 @@ head -n1)
 
 echo "Booting simulator $SIMULATOR - iOS $IOS_VERSION: $UDID"
 xcrun simctl boot "$UDID"
+
+# We use `open -a Simulator` because there's no lower-level CLI like `simctl` to display the simulator UI available.
 open -a Simulator
 
 # Wait for the simulator to boot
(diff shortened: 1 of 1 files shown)


scripts/ci-boot-simulator.sh
```diff
@@ -59,6 +59,7 @@ head -n1)
 
 echo "Booting simulator $SIMULATOR - iOS $IOS_VERSION: $UDID"
 xcrun simctl boot "$UDID"
+open -a Simulator
```
philprime, line 64: `l`: I researched the approach too and added a comment + some arguments I found online. Feel free to add or drop this suggestion as you believe is best.

```suggestion
# We use `open -a Simulator` because there's no lower-level CLI like `simctl` to display the simulator UI available.
open -a Simulator --args -CurrentDeviceUDID "$UDID"
```

philprime (approved):
LGTM

scripts/ci-boot-simulator.sh
itaybre: I like the comment, but the argument is unnecessary since simulator always opens to the booted simulator

PR #5843, opened 2025-08-07
ci: Allow bash to take snapshots on private windows

itaybre:
After reviewing screenshots on https://github.com/getsentry/sentry-cocoa/actions/runs/16805606320/job/47597097673, I see that the screenshots sometimes fail with this message:

132223_app-did-not-crash_full_screen

Having this should help capture more helpful screenshots.

### DIFF
--- a/.github/workflows/ui-tests-critical.yml
+++ b/.github/workflows/ui-tests-critical.yml
@@ -65,6 +65,9 @@ jobs:
       - run: make init-ci-build
       - run: make xcode-ci
 
+      - name: Enable screenshots permissions
+        run: ./scripts/ci-enable-permissions.sh
+
       - name: Boot simulator
         run: ./scripts/ci-boot-simulator.sh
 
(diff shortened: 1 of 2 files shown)


philprime (commented):
I would like to raise the question again, if patching the system permissions and taking a screenshot is the correct approach here. 

Is the prompt caused by us taking screenshots on error or is it caused by UI testing? If the latter one, that should IMO not happen.

philippsuess:
> I would like to raise the question again, if patching the system permissions and taking a screenshot is the correct approach here.
> 
> Is the prompt caused by us taking screenshots on error or is it caused by UI testing? If the latter one, that should IMO not happen.

It's because we want to take screenshots. I'm already starting to believe that maybe we should ditch the screenshot taking and screen recording cause it would reduce complexity. Allowing bash to record videos could again lead to flakiness. The more complex the script, the higher the chance of failure.

philprime:
> Allowing bash to record videos could again lead to flakiness

That's what I am thinking too. While I believe the scripts to be helpful, they could also turn on us.

philippsuess:
> they could also turn on us.

yes indeed.

itaybre:
I don't think the screenshots have added any flakiness, if the permissions are not there, they just fail silently.

Having the screenshots have helped finding some strange issues before

PR #5844, opened 2025-08-08
test: Use XCTestExpectation for NetworkTracker

philipphofmann:
Replace the dispatch group with an XCTestExpectation to get a failure message when the wait times out.

Contributes to GH-5789

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Integrations/Performance/Network/SentryNetworkTrackerTests.swift
+++ b/Tests/SentryTests/Integrations/Performance/Network/SentryNetworkTrackerTests.swift
@@ -806,19 +806,25 @@ class SentryNetworkTrackerTests: XCTestCase {
         let transaction = startTransaction()
 
         let queue = DispatchQueue(label: "SentryNetworkTrackerTests", qos: .userInteractive, attributes: [.concurrent, .initiallyInactive])
-        let group = DispatchGroup()
 
-        for _ in 0...100_000 {
-            group.enter()
+        let loopCount = 1_000
+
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/Integrations/Performance/Network/SentryNetworkTrackerTests.swift
```diff
@@ -806,19 +806,25 @@ class SentryNetworkTrackerTests: XCTestCase {
         let transaction = startTransaction()
 
         let queue = DispatchQueue(label: "SentryNetworkTrackerTests", qos: .userInteractive, attributes: [.concurrent, .initiallyInactive])
-        let group = DispatchGroup()
 
-        for _ in 0...100_000 {
-            group.enter()
+        let loopCount = 1_000
```
philprime, line 810: `m`: Why changing the loop count too?

philprime (approved):
LGTM

Tests/SentryTests/Integrations/Performance/Network/SentryNetworkTrackerTests.swift
philippsuess: 100_000 is a bit much for CI and not really required. If you really want to test thread-safety you anyways increase the loop count to even higher value and you must run the test repeatedly.

Sorry, I should have added a comment for this in the PR.

philippsuess:
Failing unit tests are fixed on main with https://github.com/getsentry/sentry-cocoa/pull/5851. No need to rerun these.

PR #5850, opened 2025-08-08
test: Use XCTestExpectation for concurrent modifications

philipphofmann:
Replace the dispatch group with an XCTestExpectation to get a failure message when the wait times out. To make this work, this PR changes the method to an extension so it can call XCTestCase.wait. Furthermore, this PR reduces the timeout from 500s to 10s, because 500s is way too high and risks slowing down CI.

Contributes to GH-5789

#skip-changelog


philprime (approved):
LGTM, thank you.

PR #5851, opened 2025-08-08
test: Fix flaky non file url for data tracing

philipphofmann:
The web request to the .gitignore file on GitHub sometimes failed in CI. We can test the same approach with a data URL to avoid flakiness.

The test `testInitContentsOfWithSentryTracing_nonFileUrl_shouldNotTraceManually` using `nonFileUrl` failed here https://github.com/getsentry/sentry-cocoa/actions/runs/16823204267/job/47654500560

```
swift:226: error: -[SentryTests.DataSentryTracingIntegrationTests testInitContentsOfWithSentryTracing_nonFileUrl_shouldNotTraceManually] : failed: caught error: "Error Domain=NSCocoaErrorDomain Code=256 "The file “.gitignore” couldn’t be opened." UserInfo={NSURL=https://raw.githubusercontent.com/getsentry/sentry-cocoa/refs/heads/main/.gitignore}"
```

I double checked the URL and it works fine. Maybe our CI hits some rate limiting or something similar. Therefore, we use a local data URL

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Integrations/Performance/IO/DataSentryTracingIntegrationTests.swift
+++ b/Tests/SentryTests/Integrations/Performance/IO/DataSentryTracingIntegrationTests.swift
@@ -93,8 +93,10 @@ class DataSentryTracingIntegrationTests: XCTestCase {
         }
 
         var nonFileUrl: URL {
-            // URL to a file that is not a file but should exist at all times
-            URL(string: "https://raw.githubusercontent.com/getsentry/sentry-cocoa/refs/heads/main/.gitignore")!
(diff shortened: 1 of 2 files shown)


philprime (approved):
Good catch, GitHub is definitely enforcing rate-limits for their files. Didn't think of this one.

LGTM

philippsuess:
> Good catch, GitHub is definitely enforcing rate-limits for their files. Didn't think of this one.

Happened to me in the past as well 😅 . What's weird to me is that we didn't notice for so long, or why it took so long to surface.

PR #5854, opened 2025-08-08
ci: Check Package.swift differences on CI

itaybre:
Added a CI lint to validate the Package.swift files don't get out of sync.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,6 @@
 - Fix memory leak in SessionReplayIntegration (#5770)
 - Fix reporting of energy used while profiling (#5768)
 - Fixed a build error in `SentryFeedback.swift` when building with cocoapods on Xcode 14.2 (#5917)
-- Fix linking with SentrySwiftUI on Xcode 26 for visionOS (#5823)
 
 ### Internal
 


philprime:
Do I understand correctly that this script is creating a diff between the two Package.swift files and then compares it to the expected diff file, because there will always be a diff between the two files (e.g. comments)? 

What do you think about actually comparing the abstract syntax tree instead using e.g. [SourceKitten](https://github.com/jpsim/SourceKitten)? I am afraid the more complex the Package.swift files get over time, the more error-prone the diff will be.

Also, please be more verbose with your documentation in code an PR descriptions, so others don't need to assume the reasoning for the approach.

philprime (commented):
See comment in PR

philprime:
We could maybe also do it using `swift package describe --type json` in combination with `jq`

philippsuess:
> What do you think about actually comparing the abstract syntax tree instead using e.g. [SourceKitten](https://github.com/jpsim/SourceKitten)? I am afraid the more complex the Package.swift files get over time, the more error-prone the diff will be.

@philprime, I don't think it's a good idea to add an extra dependency for this simple check. I also don't think that we will actually look at the diff. We only use it to ensure we don't update one file wihout updating the other. It's just a safety guard that you have to update whenever you touch these files. When the files deviate a lot and our current approach doesn't work, we can always revisit our approach and add something more advanced. I think this approach is good enough for now.

philprime:
> I don't think it's a good idea to add an extra dependency for this simple check. I also don't think that we will actually look at the diff. 

If the purpose is just checking that both files are changed, then I am fine with the proposed approach. If we want to lint the actual Package.swift, I am advocating for a full validation.

philippsuess:
> > I don't think it's a good idea to add an extra dependency for this simple check. I also don't think that we will actually look at the diff.
> 
> If the purpose is just checking that both files are changed, then I am fine with the proposed approach. If we want to lint the actual Package.swift, I am advocating for a full validation.

The first one, no linting. I guess we could change the workflow name to reflect this.

itaybre:
The job name might have been a bad choice, the idea is to confirm the files are in sync.
They have small differences, but for now they are the only changes needed.

Makefile
```diff
@@ -183,3 +183,13 @@ xcode-ci:
 	xcodegen --spec Samples/watchOS-Swift/watchOS-Swift.yml
 	xcodegen --spec TestSamples/SwiftUITestSample/SwiftUITestSample.yml
 	xcodegen --spec TestSamples/SwiftUICrashTest/SwiftUICrashTest.yml
+
+.PHONY: check-package-diff update-package-diff
+# Diff will return an error if the files are different, so we need to ignore it with `|| true`.
+check-package-diff:
+	diff Package.swift Package@swift-5.9.swift > current_package_diff.patch || true
+	diff ./Utils/expected_package_diff.patch current_package_diff.patch
```
philippsuess: `m`: I think we should add an error message explaining what the problem is. This will almost always fail when you modify one of the two files, I believe. So we could explain that hey you updated one file of these files or maybe two. Please ensure that you apply the same changes to both files. Then update the diff with the command below.

Othewise, it can be quite confusing what the problem is and you have to ask.

philipphofmann (approved):
Thanks, LGTM with a comment.

Makefile
itaybre: Message added

PR #5855, opened 2025-08-08
ci: Crash test app onAppear instead of init

itaybre:
Even though the app has crashes the simulator seems stuck with the app on an intermediary state.

This change attempts to let the app fully launch but crash after a slight delay.

090908_app-did-not-crash

### DIFF
--- a/TestSamples/SwiftUICrashTest/SwiftUICrashTest/SwiftUICrashTestApp.swift
+++ b/TestSamples/SwiftUICrashTest/SwiftUICrashTest/SwiftUICrashTestApp.swift
@@ -8,18 +8,20 @@ struct SwiftUICrashTestApp: App {
         SentrySDK.start { options in
             options.dsn = "https://6cc9bae94def43cab8444a99e0031c28@o447951.ingest.sentry.io/5428557"
         }
-
-        let userDefaultsKey = "crash-on-launch"
-        if UserDefaults.standard.bool(forKey: userDefaultsKey) {
-
-            UserDefaults.standard.removeObject(forKey: userDefaultsKey)
-            SentrySDK.crash()
-        }
     }
 
     var body: some Scene {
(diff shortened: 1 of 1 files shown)


noahsmartin (approved):

philprime:
Just because it's flaky we should not just change the test suite.
Before we were testing what happens when the app crashes at startup.
Now we are testing what happens when the app crashes on appear.

That's two completely different points in the UI life cycle.

@itaybre and @noahsmartin did you consider that in this PR? If yes, shouldn't we document it for future reference?

philippsuess:
> Just because it's flaky we should not just change the test suite! Before we were testing what happens when the app crashes at startup. Now we are testing what happens when the app crashes on appear.
> 
> That's two completely different points in the UI life cycle.
> 
> @itaybre and @noahsmartin did you consider that in this PR? If yes, shouldn't we document it for future reference?

@philprime, we didn't document this anywhere, but the main goal of this test is to ensure that the app can launch again after a crash. 

https://github.com/getsentry/sentry-cocoa/blob/f262e101660a2e489a414e4d231e26e8de7a0b05/TestSamples/SwiftUICrashTest/test-crash-and-relaunch.sh#L5

The app crashing on startup would be an extra test case that we could cover in an extra test. This one a more common one.

philippsuess:
Anyways, it would make sense to add a comment why we moved the crash to onAppear @itaybre. Just stating that dirctly after the sdk start didn't work reliably in CI and moving this to onAppear is an attempt to make it more reliable.

itaybre:
> Anyways, it would make sense to add a comment why we moved the crash to onAppear @itaybre. Just stating that dirctly after the sdk start didn't work reliably in CI and moving this to onAppear is an attempt to make it more reliable.

Sure, I will add a comment

PR #5859, opened 2025-08-08
ci: Allow up to 0.05% decrease in coverage before marking the build as failed

itaybre:
We have some variance on what code is executed on our tests.
Because of this, the lines covered may change a little bit, with some builds changing by 0.017%.
I believe this is within as margin of error and can be safely ingored

#skip-changelog

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,8 +1,8 @@
 codecov:
   branch: main
-  require_ci_to_pass: true
+  require_ci_to_pass: yes
   notify:
-    wait_for_ci: true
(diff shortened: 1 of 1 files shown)


philipphofmann (approved):
LGTM, I almost always ignore the report when it fails.

itaybre:
> LGTM, thank you, I believe this will also fix the red checks on `main`

That's the idea, we can play with the number to find our ideal value.

> LGTM, I almost always ignore the report when it fails.

Same for me, but I would like to be able to see when it is relevant

PR #5862, opened 2025-08-08
ci: Fixes codecov config

itaybre:
Fixes yaml using `yes` instead of `true`

#skip-changelog

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,8 +1,8 @@
 codecov:
   branch: main
-  require_ci_to_pass: yes
+  require_ci_to_pass: false # Check coverage even if the tests fail
   notify:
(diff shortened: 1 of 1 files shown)


philprime (approved):
LGTM. Do we have a reference to documentation explaining the difference between `yes` and `true`?

itaybre:
> LGTM. Do we have a reference to documentation explaining the difference between `yes` and `true`?

YAML 1.1 accepts `yes` as true.
But YAML 1.2 doesn't accept it anymore, seems like codecov plugin for cursor uses that version.

So in theory it is supported but `true` is preferred

philprime:
Thanks for the note, lgtm 👍

PR #5864, opened 2025-08-10
feat(v9): Remove deprecated setExtraValue

noahsmartin:
This removes a deprecated function from V9

#skip-changelog

### DIFF
--- a/Sources/Sentry/Public/SentryDefines.h
+++ b/Sources/Sentry/Public/SentryDefines.h
@@ -152,6 +152,7 @@ typedef BOOL (^SentryShouldQueueEvent)(
 typedef NSNumber *_Nullable (^SentryTracesSamplerCallback)(
     SentrySamplingContext *_Nonnull samplingContext);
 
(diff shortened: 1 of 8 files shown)


Sources/Sentry/Public/SentryDefines.h
```diff
@@ -143,14 +143,14 @@ typedef BOOL (^SentryShouldQueueEvent)(
 typedef NSNumber *_Nullable (^SentryTracesSamplerCallback)(
     SentrySamplingContext *_Nonnull samplingContext);
 
+#if !SDK_V9
 /**
  * Function pointer for span manipulation.
  * @param span The span to be used.
  */
 typedef void (^SentrySpanCallback)(id<SentrySpan> _Nullable span DEPRECATED_MSG_ATTRIBUTE(
```
noahsmartin, line 160: I had previously removed the only use of this typedef from the API in V9 but forgot to remove the typedef itself, so included it in the preprocessor check here.

PR #5872, opened 2025-08-11
Convert SentryAppState to Swift

noahsmartin:
Migrate SentryAppState to Swift

#skip-changelog


Sources/Swift/SentryAppState.swift
```diff
@@ -0,0 +1,132 @@
...
+        self.wasTerminated = wasTerminated
+        
+        // Validate and extract isANROngoing
+        guard let isANROngoing = jsonObject["is_anr_ongoing"] as? Bool else {
+            return nil
+        }
+        self.isANROngoing = isANROngoing
+        
+        // Validate and extract isSDKRunning
+        if let isSDKRunning = jsonObject["is_sdk_running"] as? Bool {
```
philprime, line 103: `m`: As mentioned by Cursor Bot, this value is not a boolean but an NSNumber and needs to be casted to boolean.

We should have or add a unit test for this.

philprime (approved):
LGTM but there is one change in the deserialization you need to fix. I'll approve so you can proceed with the PR when fixed.

Sources/Swift/SentryAppState.swift
philippsuess: Please add a test to reproduce the problem. Since no unit test is failing due to this, it means we don't have all the edge cases covered. I would prefer adding more tests covering all the edge cases and running them against the old implementation, and then use them to validate we don't break anything, please.

noahsmartin: I don't think cursor is right about this, you can cast the value to either Bool or NSNumber, but Bool is more idiomatic Swift:

```
        print("NSNumber value \(String(describing: isOngoing as? NSNumber))")
        print("Bool value \(String(describing: isOngoing as? Bool))")
```
will print:
```
NSNumber value Optional(1)
Bool value Optional(true)
```

Also there are tests for this and they're passing. Here is a [Swift PR](https://github.com/swiftlang/swift-corelibs-foundation/pull/1642) where they say you can cast NSNumber to bool

itaybre: I double checked and found no issues casting NSNumbers to Bool in a playground.

Sources/Swift/SentryAppState.swift
```diff
@@ -0,0 +1,132 @@
...
+            data["release_name"] = releaseName
+        }
+        data["os_version"] = self.osVersion
+        data["vendor_id"] = self.vendorId
+        data["is_debugging"] = self.isDebugging
+        data["system_boot_timestamp"] = sentry_toIso8601String(self.systemBootTimestamp)
+        data["is_active"] = self.isActive
+        data["was_terminated"] = self.wasTerminated
+        data["is_anr_ongoing"] = self.isANROngoing
+        data["is_sdk_running"] = self.isSDKRunning
```
itaybre: Is the output the same as before? Seems to be using booleans instead of NSNumbers.
I worry this is read by some other class

noahsmartin: I think internally the difference here is with a boolean it's `true` and with nsnumber it's `1` so it could be possible (though unlikely) someone uses this and only checks for `1`, I'll update

Issue #5219, opened 2025-05-12
Use `settings.infer_ip` instead of `ip_address: {{auto}}`


kahest:
### Description

Today, in the Cocoa SDK we set `ip_address: {{auto}}` to infer the ip address on the server.

There is now a new way to do this, documented here: https://develop.sentry.dev/sdk/data-model/event-payloads/sdk/#attributes

So we should:

* Stop setting `ip_address: {{auto}}`
* And instead set `settings.infer_ip: 'auto'` in the event metadata (if default PII is enabled)

This should make it easier to do that, as we can control this in a single place in the event processing pipeline instead of in multiple places (which is necessary today, as we also set this for standalone spans etc).

Note that this is specifically relevant for Cocoa (and JS) because we have legacy workarounds in place in Relay for these SDKs.

kahest:
PR that implements this on sentry-javascript: [https://github.com/getsentry/sentry-javascript/pull/17364](https://github.com/getsentry/sentry-javascript/pull/17364)

PR #5877, opened 2025-08-11
feat: Replace `user.ipAddress = {{auto}}` with a new settings.infer_ip property in the sdk metadata

itaybre:
Added a new class to handle SDK settings, for now only `infer_ip` is needed but given this name, it will probably be used for more attributes.

Updates all tests using SentrySDKInfotoo

Implements: #5219

Given that no changes are made in the public API: #skip-changelog


Tests/SentryTests/Protocol/SentrySDKSettings+Equality.h
```diff
@@ -0,0 +1,13 @@
+#import "SentrySDKSettings.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SentrySDKSettings (Equality)
```
philprime, line 5: `l`: Is there a reason why this is not part of the SDK?

Tests/SentryTests/Protocol/SentrySDKSettings+Equality.m
```diff
@@ -0,0 +1,30 @@
...
+@implementation SentrySDKSettings (Equality)
+
+- (BOOL)isEqual:(id _Nullable)object
+{
+    if (object == self)
+        return YES;
+    if ([self class] != [object class])
+        return NO;
+
+    SentrySDKSettings *otherSDKSettings = (SentrySDKSettings *)object;
```
philprime: `m`: Consider using the macro for unwrapping nullable. But you will have to include "SentryInternalDefines.h".

```suggestion
    SentrySDKSettings *_Nonnull otherSDKSettings = SENTRY_UNWRAP_NULLABLE(SentrySDKSettings, object);
```

Tests/SentryTests/Protocol/SentrySDKSettings+Equality.m
```diff
@@ -0,0 +1,30 @@
...
+    if (!self.autoInferIP == otherSDKSettings.autoInferIP) {
+        return NO;
+    }
+
+    return YES;
+}
+
+- (NSUInteger)hash
+{
+    NSUInteger hash = 18;
```
philprime: `l`: I might be wrong here, but shouldn't hashing functions use prime numbers to reduce the chance of collisions?

```suggestion
    NSUInteger hash = 17;
```

philprime (approved):
LGTM, left some comments to consider.

Tests/SentryTests/Protocol/SentrySDKSettings+Equality.m
itaybre: Yeah, I don't know what crossed my mind here, will fix it

Tests/SentryTests/Protocol/SentrySDKSettings+Equality.h
itaybre: This is only needed for testing.

PR #5883, opened 2025-08-12
feat: Proposal for flushing logs

philipphofmann:
Proposal for chaning flush for logs https://github.com/getsentry/sentry-cocoa/pull/5834.


philippsuess:
@denrase already included the feedback in https://github.com/getsentry/sentry-cocoa/pull/5834, so we can close this.

PR #5889, opened 2025-08-12
fix: Add null-handling for static cache path in async logger

philprime:
This PR is derived from https://github.com/getsentry/sentry-cocoa/pull/5572 in an effort to make the large amount of changes easier to review for https://github.com/getsentry/sentry-cocoa/issues/5577.

The async logger relies on a nullable static caches path, so we need to unwrap it before we check for existence. As we are already logging an error and exiting the method, when directory creation fails, I use the same approach when the cache directory is not defined.

#skip-changelog


philipphofmann (approved):
LGTM

PR #5893, opened 2025-08-12
chore(ci): Add api stability check to workflows

philprime:
## :scroll: Description

This PR implements a conditional enforcement mechanism for the API stability workflow, mirroring the existing `unit-tests-required-check` pattern.

## :bulb: Motivation and Context

This change is required to:
*   **Improve CI efficiency**: API stability checks will now only run when relevant files are modified, saving CI resources.
*   **Enable required checks**: A new `api-stability-required-check` job allows enforcing that API stability checks either pass or are explicitly skipped, similar to unit tests.
*   **Maintain consistency**: Aligns the API stability workflow with the established pattern for `unit-tests-required-check`.

## :green_heart: How did you test it?

The changes were implemented by following the established and proven pattern of the `unit-tests-required-check` workflow and file filters.

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

---
[Slack Thread](https://sentry.slack.com/archives/D0970KPGH1R/p1755005654311149?thread_ts=1755005654.311149&cid=D0970KPGH1R)

<a href="https://cursor.com/background-agent?bcId=bc-bed9475a-11bb-46df-8541-08b55516503b">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">
    Open in Cursor
  </picture>
</a>
<a href="https://cursor.com/agents?id=bc-bed9475a-11bb-46df-8541-08b55516503b">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">
    Open in Web
  </picture>
</a>

#skip-changelog

PR #5896, opened 2025-08-12
fix(ci): Add files-changed to needs of unit-tests-required-check

philprime:
During review of #5893 the Cursor Review Bot and Sentry Seer mentioned [reported](https://github.com/getsentry/sentry-cocoa/pull/5893#discussion_r2269955609) the following issue:

The `api-stability-required-check` job only depends on `api-stability`, not `files-changed`. If the `files-changed` job fails, the `api-stability` job will be skipped. Since `api-stability-required-check` only checks for 'failure' or 'cancelled' results, it incorrectly passes, allowing the workflow to complete without validating if API stability checks were needed. This could lead to undetected API changes. The `api-stability-required-check` job should depend on both `files-changed` and `api-stability`.

The same applies ot the `unit-tests-required-check`, therefore I am adding the `files-changed` to the list of needs.

#skip-changelog

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -319,6 +319,7 @@ jobs:
   unit-tests-required-check:
     needs:
       [
+        files-changed,
         build-test-server,
         unit-tests,
       ]


philippsuess:
Thanks @philprime

Issue #5876, opened 2025-08-11
Sentry-WithoutUIKitOrAppKit should omit arm64e or be delivered in both flavors


jboulter11:
### Platform

iOS

### Environment

Other

### Installed

Manually

### Version

N/A

### Xcode Version

N/A

### Steps to Reproduce

https://github.com/getsentry/sentry-cocoa/releases/tag/8.54.1-alpha.0

This releasee mentions that Sentry is now being shipped with and without arm64e, which is good because we don't want arm64e for our framework.  However, it seems we also need the `Sentry-WithoutUIKitOrAppKit` to be delivered without arm64e.  I'm not sure if we (Dropbox) are the only customer using that framework or not, so I leave the call on whether to issue the UI-less framework in both flavors, or just without arm64e, up to you.

Thank you!

### Expected Result

`Sentry-WithoutUIKitOrAppKit` should not contain arm64e arch

### Actual Result

`Sentry-WithoutUIKitOrAppKit` does contain arm64e arch

### Are you willing to submit a PR?

No

philprime:
@itaybre I would assume this to be a rather easy change based on the new scripts. What do you think?

itaybre:
Yes, should be fairly easy to do. But it would be a good idea to confirm if they need arm64e or not.

itaybre:
> Yes, should be fairly easy to do. But it would be a good idea to confirm if they need arm64e or not.

I misread the issue, I can do this change

PR #5897, opened 2025-08-12
feat: Exclude arm64e slice from Sentry-WithoutUIKitOrAppKit

itaybre:
Excludes ARM64e from Sentry-WithoutUIKitOrAppKit and adds a new variant with ARM64e.
Also updates local script to build these new variants.

Fixes: #5876

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -115,6 +115,12 @@ jobs:
             configuration-suffix: WithoutUIKit
             id: sentry-withoutuikit-dynamic
             excluded-archs: arm64e
+          - scheme: Sentry
+            macho-type: mh_dylib
+            suffix: "-WithoutUIKitOrAppKit"
+            configuration-suffix: WithoutUIKit
+            id: sentry-withoutuikit-dynamic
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -114,6 +114,13 @@ jobs:
             suffix: "-WithoutUIKitOrAppKit"
             configuration-suffix: WithoutUIKit
             id: sentry-withoutuikit-dynamic
+            excluded-archs: arm64e
+          - scheme: Sentry
+            macho-type: mh_dylib
+            suffix: "-WithoutUIKitOrAppKit"
+            configuration-suffix: WithoutUIKit
+            id: sentry-withoutuikit-dynamic
```
philprime, line 122: `m`: @itaybre I believe the Cursor feedback is valid and the id is not unique

philprime (approved):
LGTM, except one comment. I'll approve this now, so you can fix + merge it.

.github/workflows/release.yml
itaybre: This id is used to download the previous step archives, so it needs to be the same

PR #5899, opened 2025-08-12
meta: Add comment explaining why the crash was moved to onAppear

itaybre:
Adds a some more context to the code for testing crash on launch which was changed in #5855

#skip-changelog


philprime (approved):
Thank you for adding this

PR #5900, opened 2025-08-12
fix: Race condition in SentryApplication and Swift implementation

noahsmartin:
WIP Working on a race condition in SentryApplication handling of app state

#skip-changelog


ohwhen:
🔥

Sources/Sentry/SentryCrashWrapper.m
```diff
@@ -185,7 +185,8 @@ - (void)enrichScope:(SentryScope *)scope
 // The UIWindowScene is unavailable on visionOS
 #if SENTRY_TARGET_REPLAY_SUPPORTED
 
-    NSArray<UIWindow *> *appWindows = SentryDependencyContainer.sharedInstance.application.windows;
+    NSArray<UIWindow *> *appWindows =
```
philprime, line 186: `m`: Is there a good reason why we shouldn't use the thread-safe version here?

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -166,14 +160,15 @@ - (instancetype)init
...
-        _application = [[SentryUIApplication alloc]
-            initWithNotificationCenterWrapper:_notificationCenterWrapper
-                         dispatchQueueWrapper:_dispatchQueueWrapper];
+        _application = UIApplication.sharedApplication;
+        _threadsafeApplication = [[SentryThreadsafeApplication alloc]
+            initWithInitialState:_application.unsafeApplicationState
+              notificationCenter:_notificationCenterWrapper];
 #elif TARGET_OS_OSX
-        _application = [[SentryNSApplication alloc] init];
+        _application = NSApplication.sharedApplication;
```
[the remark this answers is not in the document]
philprime: I believe this comment is bogus because the container will always be initialized on the main thread.

Sources/Swift/Helper/SentryApplication.swift
```diff
@@ -0,0 +1,45 @@
...
+import AppKit
+#endif
+#if canImport(UIKit) && !SENTRY_NO_UIKIT
+import UIKit
+#endif
+
+@objc @_spi(Private) public protocol SentryApplication {
+    
+    // This can only be accessed on the main thread
+    var mainThread_isActive: Bool { get }
```
philprime, line 11: `m`: Is there a good reason for us to not fully synchronize the application onto the main thread? The app life cycle will always live of main

Sources/Swift/Helper/ThreadSafeApplication.swift
```diff
@@ -0,0 +1,49 @@
+#if !os(macOS) && !os(watchOS) && !SENTRY_NO_UIKIT
+import UIKit
+
+@objc @_spi(Private) public final class SentryThreadsafeApplication: NSObject {
```
philprime, line 4: `m`: We should replace the `application` in the dependency container with this wrapper, so we can not use the wrong one by accident.

Tests/SentryTests/SentryUIApplication+Private.h
```diff
@@ -1,21 +1,21 @@
...
-
-- (NSArray<UIViewController *> *)relevantViewControllers;
-
-@end
-
-NS_ASSUME_NONNULL_END
-
-#endif /* SentryUIApplication_Private_h */
-#endif
+//
```
philprime: `l`: If we don't need it, remove it please.

philprime (commented):
Good progress, let some comments

Sources/Sentry/SentryCrashWrapper.m
noahsmartin: `getWindows` is only defined on the non-thread safe version. AFAIK it's never safe to have a UIWindow instance off the main thread

noahsmartin: Or in Swift terms, UIWindow is not Sendable

Sources/Swift/Helper/SentryApplication.swift
noahsmartin: not sure what you mean, this is conformed to by NSApplication and NSApplication can only be used on the main thread. We could write a helper/extension that is a function designed to be called from any context that then dispatches sync to the main thread and check "isActive" but there is no use case in the codebase that requires that right now so I would be hesitant to add extra code, particularly code that uses "dispatch_sync" since that's easy to cause deadlock with and I try to avoid it

Sources/Swift/Helper/ThreadSafeApplication.swift
noahsmartin: The dependency container has both, that's the idea behind this PR because previously some of SentryApplication was designed to be called form anywhere and some was not. The one that can only be used form the main thread is still useful to have in the dependency container IMO since it allows us to mock those calls in tests

Sources/Swift/Helper/SentryApplication.swift
noahsmartin: I think I see what you mean, like by doing something similar to what we do with UIApplication. The reason is that before this PR we were not doing anything to try to make this accessible to a background thread, and it was only called from the main thread. Still only called from the main thread and I wanted to avoid any extra changes here other than cleaning up the code to make it more clear what each piece does and fixing the one race condition bug

Sources/Sentry/SentryCrashWrapper.m
philprime: Ah makes sense, also this method `enrichScope(scope:)` should never be called from a background thread anyways.

philprime (approved):
LGTM

PR #5905, opened 2025-08-13
test: Increase timeout to 5 seconds for ANRV1

philipphofmann:
A waitTimeout of 2 seconds is usually not enough. Let's increase it to 5 seconds.

Failed here on main https://github.com/getsentry/sentry-cocoa/actions/runs/16913124095/job/47921390079

#skip-changelog


philprime (approved):
LGTM

Issue #5915, opened 2025-08-13
Improve GlobalEventProcessor


romtsn:
### Description

At the moment, `GlobalEventProcessor` is only used internally in one place (Session Replay) and not exposed to customers. Therefore there's two potential ways to improve that:
- Align with other SDKs ([JS](https://github.com/getsentry/sentry-javascript/blob/develop/packages/core/src/types-hoist/eventprocessor.ts), [Java](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/EventProcessor.java)) and properly expose it to customers:
  - Allow adding event processors via options
  - Allow removing event processors
  - Document it and showcase how to use it
-  Ditch the processor completely and change how we trigger replay capture for errors (e.g. by retrieving the SessionReplay integration from hub/client and calling `captureReplay` directly)

philprime:
@romtsn is this a follow-up issue on your changes for the order processing in beforeSend or can this be closed as completed?

philippsuess:
@philprime, this is a follow-up issue that remains outstanding. Correct me if I'm wrong @romtsn.

romtsn:
Yes, this is a follow up that is yet to be tackled

PR #5916, opened 2025-08-13
fix(replay): Don't capture replays for events dropped in beforeSend

### LINKED ISSUE #5735: Change order of event processing to filter replay events with `beforeSend`
### Description

Session replay events are created and send by using the event processor called from the `SentryClient`:

https://github.com/getsentry/sentry-cocoa/blob/6b57184fb4aea6c8c6ad2765b3fe5f2a87f73710/Sources/Sentry/SentryClient.m#L826

https://github.com/getsentry/sentry-cocoa/blob/22b6996c73cfaa36ff18f44e779c9d7ef826622a/Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift#L141

The event processor is called before the option `beforeSend` is evaluated, resulting in replay events being sent while the related error event is actually filtered using `beforeSend`. 

https://github.com/getsentry/sentry-cocoa/blob/6b57184fb4aea6c8c6ad2765b3fe5f2a87f73710/Sources/Sentry/SentryClient.m#L860

We must not send replay events for errors which were filtered out.

romtsn:
## :scroll: Description
- Reorder eventProcessors with beforeSend so if the event is dropped in beforeSend the processors are not called
  - This is currently only used in one place for SessionReplay and we don't expose it to customers, so it shouldn't break any existing behavior
  -  I created #5915 to improve this, and depending on the decision made there we might have to ditch the event processors or document the order.

Closes #5735

Manually + automated

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryClient.m
+++ b/Sources/Sentry/SentryClient.m
@@ -834,22 +834,6 @@ NSString *const DropSessionLogMessage = @"Session has no release name. Won't sen
         currentSpanCount = 0;
     }
 
-    event = [self callEventProcessors:event];
-    if (event == nil) {
-        [self recordLost:eventIsNotATransaction reason:kSentryDiscardReasonEventProcessor];
-        if (eventIsATransaction) {
-            // We dropped the whole transaction, the dropped count includes all child spans + 1 root
-            // span
-            [self recordLostSpanWithReason:kSentryDiscardReasonEventProcessor
-                                  quantity:currentSpanCount + 1];
-        }
-    } else {
-        if (eventIsATransactionClass) {
-            [self recordPartiallyDroppedSpans:(SentryTransaction *)event
-                                   withReason:kSentryDiscardReasonEventProcessor
-                         withCurrentSpanCount:&currentSpanCount];
-        }
-    }
     if (event != nil && eventIsATransaction && self.options.beforeSendSpan != nil) {
         SentryTransaction *transaction = (SentryTransaction *)event;
         NSMutableArray<id<SentrySpan>> *processedSpans = [NSMutableArray array];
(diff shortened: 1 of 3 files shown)


ohwhen:
👀

Sources/Sentry/SentryClient.m
```diff
@@ -887,6 +871,25 @@ - (SentryEvent *_Nullable)prepareEvent:(SentryEvent *)event
         }
     }
 
+    if (event != nil) {
```
philippsuess, line 874: `m`: I think it would be good to add a short comment explaining why we need to call this code at this location.

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -294,6 +294,34 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
...
+            options.beforeSend = { _ in
+                return nil
+            }
+        })
+        
+        let client = SentryClient(options: try XCTUnwrap(SentrySDKInternal.options))
+        let scope = Scope()
+        let hub = TestHub(client: client, andScope: scope)
+        SentrySDKInternal.setCurrentHub(hub)
+        let expectation = expectation(description: "Replay to be capture")
```
philippsuess: `l`: Typo
```suggestion
        let expectation = expectation(description: "Replay to be captured")
```

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -294,6 +294,34 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
...
+        expectation.isInverted = true
+        hub.onReplayCapture = {
+            expectation.fulfill()
+        }
+        
+        let crash = Event(error: NSError(domain: "Error", code: 1))
+        crash.context = [:]
+        crash.isFatalEvent = true
+        client?.capture(event: crash)
```
philippsuess: `m`: The client must not be nullable. I would use XCTUnwrap for the client as well, because if it's nil the test will always succeed cause we don't even capture an event.

philprime (approved):
LGTM, same feedback as @philipphofmann

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
romtsn: to be honest I just copied the setup from the test above but I can change it ofc 👍

philippsuess: Thanks 😃

philipphofmann (approved):
Thanks @romtsn

PR #5942, opened 2025-08-13
ci: Bump iOS and tvOS versions to 18.6

itaybre:
Bump iOS and tvOS versions used for testing since Github removed older versions on https://github.com/actions/runner-images/pull/12734

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -124,7 +124,7 @@ jobs:
             runs-on: macos-15
             platform: "iOS"
             xcode: "16.4"
-            test-destination-os: "18.2"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -191,7 +191,7 @@ jobs:
             runs-on: macos-15
             platform: "tvOS"
             xcode: "16.4"
-            test-destination-os: "18.1"
+            test-destination-os: "18.6"
```
itaybre:
```suggestion
            test-destination-os: "18.5"
```
tvOS 18.6 is not available

PR #5943, opened 2025-08-13
ref: Convert SentryUIDevice to Swift

noahsmartin:
Fixes https://github.com/getsentry/sentry-cocoa/issues/5921

#skip-changelog


Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
```diff
@@ -0,0 +1,96 @@
...
+    var orientation: UIDeviceOrientation { get }
+    var isBatteryMonitoringEnabled: Bool { get }
+    var batteryState: UIDevice.BatteryState { get }
+    var batteryLevel: Float { get }
+#endif
+}
+
+@_spi(Private) @objc public final class DefaultSentryUIDeviceWrapper: NSObject, SentryUIDeviceWrapper {
+    
+    let queueWrapper: SentryDispatchQueueWrapper
```
philippsuess: `m`: Shouldn't that be?
```suggestion
    private let queueWrapper: SentryDispatchQueueWrapper
```

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
```diff
@@ -0,0 +1,96 @@
...
+
+#if os(iOS)
+    var orientation: UIDeviceOrientation { get }
+    var isBatteryMonitoringEnabled: Bool { get }
+    var batteryState: UIDevice.BatteryState { get }
+    var batteryLevel: Float { get }
+#endif
+}
+
+@_spi(Private) @objc public final class DefaultSentryUIDeviceWrapper: NSObject, SentryUIDeviceWrapper {
```
philippsuess: `l`: We usually always start class names with Sentry
```suggestion
@_spi(Private) @objc public final class SentryDefaultUIDeviceWrapper: NSObject, SentryUIDeviceWrapper {
```

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
```diff
@@ -0,0 +1,96 @@
...
+    
+    let queueWrapper: SentryDispatchQueueWrapper
+    
+    override convenience init() {
+        self.init(queueWrapper: Dependencies.dispatchQueueWrapper)
+    }
```
philippsuess, line 29: `m`: I would really prefer removing these convenience initializers not asking for all the required dependencies, if possible. It already was like this, so we can also delay this to another PR.

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
```diff
@@ -0,0 +1,96 @@
...
+
+    #endif // os(iOS)
+    
+    @objc public func getSystemVersion() -> String {
+        systemVersion
+    }
+    
+    private var systemVersion = ""
+    private var cleanupBatteryMonitoring = false
+    private var cleanupDeviceOrientationNotifications = false
```
philippsuess: `l`: I prefer having the private properties at the top of classes. Having them at the bottom feels a bit weird to me.

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
```diff
@@ -0,0 +1,96 @@
...
+                device.endGeneratingDeviceOrientationNotifications()
+            }
+            if needsDisablingBattery {
+                device.isBatteryMonitoringEnabled = false
+            }
+        }
+        #endif
+    }
+    
+    func dealloc() {
```
philippsuess: `h`: Yes, what cursors said https://github.com/getsentry/sentry-cocoa/pull/5943/files#r2274679589.
```suggestion
    deinit {
```

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
philprime: This should be at least `m` because we use the naming of classes to detect stack traces in crash detection.

noahsmartin: Ah I didn't think about the detecting stack traces that is interesting. It would be great if we could move that to use the module name as we adopt swift, since technically the name is now `Sentry.SentryDefaultUIDeviceWrapper`, a bit redundant. Will change for this PR, just making the note for the future

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
noahsmartin: I agree, I think unfortunately it cannot be removed because we are subclassing NSObject, but I will update it with `@available(*, unavailable)` so at least it's clear not to use it from Swift

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
noahsmartin: 🤦 I will also add a test for this

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
noahsmartin: I'm used to the opposite: https://github.com/airbnb/swift?tab=readme-ov-file#marks-within-types but I have no preference, will update it

noahsmartin:
Updated and added 2 new tests for this

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
philprime: Could we use something like `fatalError("Not implemented")` in Swift, or do you think that would be too dangerous?

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
```diff
@@ -0,0 +1,99 @@
...
+@_spi(Private) @objc public protocol SentryUIDeviceWrapper {
+    func start()
+    func stop()
+    func getSystemVersion() -> String
+
+#if os(iOS)
+    var orientation: UIDeviceOrientation { get }
+    var isBatteryMonitoringEnabled: Bool { get }
+    var batteryState: UIDevice.BatteryState { get }
+    var batteryLevel: Float { get }
```
philprime, line 13: `m`: Might also be relevant for macOS? Probably out of scope for this PR though

philprime (approved):
LGTM, but it might be worth to add the additional two test cases to get the [code coverage](https://app.codecov.io/gh/getsentry/sentry-cocoa/blob/sentryUIDeviceWrapperSwift/Sources%2FSwift%2FCore%2FHelper%2FSentryUIDeviceWrapper.swift) of `SentryUIDeviceWrapper.swift` to 100%

Sources/Swift/Core/Helper/SentryUIDeviceWrapper.swift
noahsmartin: It's not something we use in macOS right now, so want to keep the behavior the same as much as possible and just focus on the bug fix for this PR

PR #5950, opened 2025-08-14
chore(ci): Add CI debug script

philprime:
Adds a debug diagnostics script to log all relevant information for each workflow. This should help to debug the reason for failing checks like this one:

https://github.com/getsentry/sentry-cocoa/actions/runs/16962270797/job/48077729227?pr=5949

I created a utility script to add logging in the GitHub Actions style, so we can use grouping and log levels to improve visualization.

EDIT:
The script is now executed as the last step of the workflows with a failure-only condition to reduce the amount of overhead introduced.

#skip-changelog


.github/workflows/ui-tests-common.yml
```diff
@@ -49,15 +49,11 @@ jobs:
     name: UI Tests Common
     runs-on: ${{ inputs.macos_version }}
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout code
+        uses: actions/checkout@v4
```
philprime:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/codeql-analysis.yml
```diff
@@ -33,12 +33,17 @@ jobs:
...
+        run: ./scripts/ci-debug-xcode-env.sh
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v3
         with:
           languages: ${{ matrix.language }}
 
+      - name: Select Xcode version
+        run: ./scripts/ci-select-xcode.sh 16.4
+
```
philprime:
```suggestion
```

.github/workflows/release.yml
```diff
@@ -148,7 +150,8 @@ jobs:
     runs-on: macos-14
     needs: assemble-xcframework-variant
     steps:
-      - uses: actions/checkout@v5
+      - name: Debug Xcode environment
+        run: ./scripts/ci-debug-xcode-env.sh
```
philprime:
```suggestion
      - uses: actions/checkout@v5
      - name: Debug Xcode environment
        run: ./scripts/ci-debug-xcode-env.sh
```

.github/workflows/api-stability.yml
```diff
@@ -46,6 +46,8 @@ jobs:
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
+      - name: Debug Xcode environment
+        run: ./scripts/ci-debug-xcode-env.sh
```
philippsuess: `m`: Running the script takes around 20 to 30s. I would prefer running this only when a job failed to avoid slowing down CI. We can also call this script at the end of every workflow and do a lightweight version if the job succeeds and a more detailed one when it fails. 
In most cases we don't need the info.

philprime: If it fails, we don't have the information. Then we re-run it and it doesn't fail anymore, we have information which is unnecessary.

I agree this adds overhead, but we need it right now for all jobs.

philprime: Okay but I agree we could instead run this script in case of failure at the end of the workflow. That way it is only triggered when something went wrong, while still running anytime we need it.

philprime:
Works as expected here:
https://github.com/getsentry/sentry-cocoa/actions/runs/16966362222/job/48091613171?pr=5950

Skipped as expected here:
https://github.com/getsentry/sentry-cocoa/actions/runs/16966362201/job/48091256954?pr=5950

PR #5989, opened 2025-08-25
fix(client): Fix attachment processor dropping additional attachments

philprime:
## :scroll: Description

Changes the attachment processors to reuse the output of previous integrations, to not drop attachments.

When enabling the options `attachScreenshot` and `attachViewHierarchy`, the two respective integrations `SentryScreenshotIntegration` and `SentryViewHierarchyIntegration` are installed respectively.

The current implementation of the `[SentryClient processAttachmentsForEvent:attachments:]` is simply replacing the `processedAttachments` containing the output of the previous integration, instead of merging the result, therefore only one integration can be used at a time.

This PR changes the loop to call each processor with the output of previous ones. This allows processors to add **and remove** attachments of previous processors. This also introduces an implicit order of integrations based on the installation order in [SentryOptionsInternal.m](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentyOptionsInternal.m#L35-L72).

We are ignoring this order for now, as we only have two integrations processing attachments, and they are both only adding attachments to the result. I added a note in the code for this.

TBD

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryClient.m
+++ b/Sources/Sentry/SentryClient.m
@@ -541,8 +541,16 @@ NSString *const DropSessionLogMessage = @"Session has no release name. Won't sen
                                                 withScope:scope
                                    alwaysAttachStacktrace:NO];
 
+    if (replayEvent == nil) {
+        SENTRY_LOG_DEBUG(@"The replay event was filtered out in prepare event. "
+                         @"The replay was discarded.");
+        return;
+    }
+
+    // Only check the type of the returned event, as the instance could be changed in the event
+    // preprocessor and before-send handlers.
     if (![replayEvent isKindOfClass:SentryReplayEvent.class]) {
-        SENTRY_LOG_DEBUG(@"The event preprocessor didn't update the replay event in place. The "
+        SENTRY_LOG_ERROR(@"The event preprocessor didn't update the replay event in place. The "
                          @"replay was discarded.");
         return;
     }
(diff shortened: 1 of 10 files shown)


Sources/Sentry/SentryClient.m
```diff
@@ -1134,7 +1134,13 @@ - (void)removeAttachmentProcessor:(id<SentryClientAttachmentProcessor>)attachmen
     NSArray<SentryAttachment *> *processedAttachments = attachments;
 
     for (id<SentryClientAttachmentProcessor> attachmentProcessor in self.attachmentProcessors) {
-        processedAttachments = [attachmentProcessor processAttachments:attachments forEvent:event];
+        // Keep chaining the processed attachments so each processor works on the output of the
+        // previous one. This is necessary so each processor can add and remove attachments.
+        //
+        // Important: This means the order of adding processors matters and relies on the initialization order of
+        // the integrations. At this point in time the attachment processors are only adding attachments, therefore
```
kahest: `h`: this sounds like a breaking change, i.e. that it can alter existing behaviour ppl are used to, even though it might not be fully correct. am I wrong?

philprime: I don't think this is a breaking change, because before this PR it would only use the result of the last processor in the list, ignoring any previous output.

As the two existing integrations are only additive, it will add more (missing) data, than changing existing ones:

https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentryViewHierarchyIntegration.m#L70-L110

https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentryScreenshotIntegration.m#L63-L110

kahest: got it, sgtm then 👍

Sources/Sentry/SentryClient.m
```diff
@@ -1134,7 +1134,14 @@ - (void)removeAttachmentProcessor:(id<SentryClientAttachmentProcessor>)attachmen
     NSArray<SentryAttachment *> *processedAttachments = attachments;
 
     for (id<SentryClientAttachmentProcessor> attachmentProcessor in self.attachmentProcessors) {
-        processedAttachments = [attachmentProcessor processAttachments:attachments forEvent:event];
+        // Keep chaining the processed attachments so each processor works on the output of the
+        // previous one. This is necessary so each processor can add and remove attachments.
+        //
+        // Important: This means the order of adding processors matters and relies on the
+        // initialization order of the integrations. At this point in time the attachment processors
+        // are only adding attachments, therefore we can ignore this restriction for now.
```
itaybre, line 1155: Thanks for adding this 🎉

itaybre (approved):
LGTM

PR #5995, opened 2025-08-27
ref: Migrate `SentryBinaryImageCache` to Swift

itaybre:
RN are failing due to an import, but I fixed than on https://github.com/getsentry/sentry-react-native/pull/5109

#skip-changelog


Sources/Sentry/SentryBinaryImageCacheCallbacks.m
```diff
@@ -0,0 +1,30 @@
+#import "SentryBinaryImageCacheCallbacks.h"
+#import "SentryDependencyContainer.h"
+#import "SentryInternalDefines.h"
+#import "SentrySwift.h"
+
+// Used to call SentryDependencyContainer, since it isn't available to swift yet
```
noahsmartin: I have been using `SentryDependencyContainerSwiftHelper` for this

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,182 @@
...
+    private var isDebug: Bool = false
+    // Use a recursive lock to allow the same thread to enter again
+    private let lock = NSRecursiveLock()
+    private var dispatchQueueWrapper: SentryDispatchQueueWrapper!
+    
+    @objc public override init() {
+        self.isDebug = false
+    }
+    
+    @_spi(Private) @objc public func start(_ isDebug: Bool, dispatchQueueWrapper: SentryDispatchQueueWrapper) {
```
noahsmartin: I see you changed the initializer here by adding the dispatchQueueWrapper param. While I think that's good (makes it more testable) Just wanted to give you a heads up that you could also have made this PR smaller by not making this improvement in the same change and instead accessed it form Swift through `Dependencies.dispatchQueueWrapper`

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,182 @@
...
+ */
+@objc(SentryBinaryImageCache)
+public class SentryBinaryImageCache: NSObject {
+    @objc public internal(set) var cache: [SentryBinaryImageInfo]?
+    private var isDebug: Bool = false
+    // Use a recursive lock to allow the same thread to enter again
+    private let lock = NSRecursiveLock()
+    private var dispatchQueueWrapper: SentryDispatchQueueWrapper!
+    
+    @objc public override init() {
```
noahsmartin: Can this be removed since you already set `isDebug` to false by default in the property declaration?

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,182 @@
...
+ * information in a sorted collection that will be used to symbolicate frames with better
+ * performance.
+ */
+@objc(SentryBinaryImageCache)
+public class SentryBinaryImageCache: NSObject {
+    @objc public internal(set) var cache: [SentryBinaryImageInfo]?
+    private var isDebug: Bool = false
+    // Use a recursive lock to allow the same thread to enter again
+    private let lock = NSRecursiveLock()
+    private var dispatchQueueWrapper: SentryDispatchQueueWrapper!
```
noahsmartin: The force unwrap worries me, maybe we should use `Dependencies.dispatchQueueWrapper` to match the original behavior and be sure no-one is every relying on this being present without start being called

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
itaybre: Removed it, and will do a follow up PR with moving `dispatchQueueWrapper` to a parameter in start

Sources/Sentry/SentryBinaryImageCacheCallbacks.m
itaybre: After trying it, I remembered why I can't use `SentryDependencyContainerSwiftHelper`.
Since `SentryBinaryImageCache` is a swift class, the class forwarding doesn't work with SPM; joys of workign with ObjC and Swift

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+/**
+ * This class listens to `SentryCrashBinaryImageCache` to keep a copy of the loaded binaries
+ * information in a sorted collection that will be used to symbolicate frames with better
+ * performance.
+ */
+@objc(SentryBinaryImageCache)
+public class SentryBinaryImageCache: NSObject {
```
noahsmartin: This whole class should be marked @_spi(Private) so that it doesn't get added to sdk_api.json, then you should be able to undo your changes to that file

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,176 @@
...
+ * performance.
+ */
+@objc(SentryBinaryImageCache)
+@_spi(Private) public final class SentryBinaryImageCache: NSObject {
+    @objc public internal(set) var cache: [SentryBinaryImageInfo]?
+    private var isDebug: Bool = false
+    // Use a recursive lock to allow the same thread to enter again
+    private let lock = NSRecursiveLock()
+    
+    @_spi(Private) @objc public func start(_ isDebug: Bool) {
```
noahsmartin: No need for the spi annotation here since it's on the class now

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,176 @@
...
+                let mid = (left + right) / 2
+                let compareImage = cache[mid]
+                if newImage.address < compareImage.address {
+                    right = mid
+                } else {
+                    left = mid + 1
+                }
+            }
+            
+            self.cache?.insert(newImage, at: left)
```
noahsmartin, line 92: I think we use `cache` here instead of `self.cache?` to remove the optional?

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+    
+    @objc
+    public static func convertUUID(_ value: UnsafePointer<UInt8>?) -> String? {
+        guard let value = value else { return nil }
+        
+        var uuidBuffer = [CChar](repeating: 0, count: 37)
+        sentrycrashdl_convertBinaryImageUUID(value, &uuidBuffer)
+        return String(cString: uuidBuffer)
```
noahsmartin: Maybe we should pass `encoding` here since it was set to `NSASCIIStringEncoding` before

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,176 @@
...
+        lock.synchronized {
+            guard let index = indexOfImage(address: imageAddress), index >= 0 else { return }
+            self.cache?.remove(at: index)
+        }
+    }
+    
+    @objc
+    public func imageByAddress(_ address: UInt64) -> SentryBinaryImageInfo? {
+        lock.synchronized {
+            guard let index = indexOfImage(address: address) else { return nil }
```
noahsmartin, line 125: I think this should have the same `index >= 0` check that the above method has

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,176 @@
...
+        
+        imagePaths = lock.synchronized {
+            guard let cache = self.cache else { return imagePaths }
+            
+            for info in cache {
+                if SentryInAppLogic.isImageNameInApp(info.name, inAppInclude: inAppInclude) {
+                    imagePaths.insert(info.name)
+                }
+            }
+            return imagePaths
```
noahsmartin, line 164: Don't think you need the return here or the variable assignment outside of this synchronized block. Can just return imagePaths directly

noahsmartin (approved):

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
itaybre: Forgot to remove if from the other method, it returns nil if the item was not found, so no need to check for negative numbers

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -0,0 +1,174 @@
...
+                let compareImage = cache[mid]
+                if newImage.address < compareImage.address {
+                    right = mid
+                } else {
+                    left = mid + 1
+                }
+            }
+            
+            cache.insert(newImage, at: left)
+        }
```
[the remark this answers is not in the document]
itaybre: Good catch!

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
philprime: @noahsmartin it should be an implicit return, as the `lock.synchronized` is a single statement and therefore automatically returned.

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
philprime: +1

Sources/SentryCrash/Recording/SentryBinaryImageCacheCallbacks.m
```diff
@@ -0,0 +1,30 @@
...
+
+// Used to call SentryDependencyContainer, since it isn't available to swift yet
+void
+binaryImageWasAdded(const SentryCrashBinaryImage *_Nullable image)
+{
+    if (image == NULL) {
+        SENTRY_LOG_WARN(@"The image is NULL. Can't add NULL to cache.");
+        return;
+    }
+    [SentryDependencyContainer.sharedInstance.binaryImageCache binaryImageAdded:image->name
```
philprime, line 14: Note for future reference: We really need to find a way to get rid of SentryDependencyContainer.sharedInstance and instead configure a reference somehow else. Out of scope for this PR.

philprime (approved):
LGTM, only a couple of comments left to consider

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
itaybre: Arrays in swift are value types, so it won't work if I use `cache`.

PR #6008, opened 2025-08-27
feat(V9): Make performance V2 the default in V9

noahsmartin:
As the comment on this property already stated, it should be the default in V9

#skip-changelog

### DIFF
--- a/Sources/Sentry/Public/SentryOptions.h
+++ b/Sources/Sentry/Public/SentryOptions.h
@@ -288,6 +288,7 @@ NS_SWIFT_NAME(Options)
  */
 @property (nonatomic, assign) BOOL enableAutoPerformanceTracing;
 
+#if !SDK_V9
 /**
  * We're working to update our Performance product offering in order to be able to provide better
  * insights and highlight specific actions you can take to improve your mobile app's overall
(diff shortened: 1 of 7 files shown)


Sources/Sentry/Public/SentryOptions.h
```diff
@@ -287,6 +287,7 @@ NS_SWIFT_NAME(Options)
  */
 @property (nonatomic, assign) BOOL enableAutoPerformanceTracing;
 
+#if !SDK_V9
```
philprime, line 291: `h`: Not sure if I understand this PR's intention. Do you want to make `enablePerformanceV2` be enabled by default or do you want to remove the options? If the latter one why would we want to remove the option?

noahsmartin: @philprime I assumed if we had a V2 of something we'd eventually want to remove the V1 since V2 is better/newer, but maybe the intent here was to have two definitions of app start, and then I'd recommend having something like an enum for "AppStartType" so it's clear what is changing rather than a v1/v2

philippsuess: We should remove the option and enable everything by default that depends on performanceV2 in the next major. `performanceV2`. Currently that's only 
> The app start duration will now finish when the first frame is drawn instead of when the OS posts the

We didn't change this in a minor, because it impacts the app start duration, and it could cause alerts. 

I thought we would make more changes with this feature flag, but then our performance work improvements got deprioritized, and we never made further improvements. This option adds complexity, and we should ditch it.

philprime: I understand, in that case yes let's change this for V9.

We need to start a v8 -> v9 migration guide!

Issue #5440, opened 2025-06-20
Remove deprecated SCNetworkReachability


noahsmartin:
### Problem Statement

While poking around the codebase a bit to understand the networking side, I noticed use of SCNetworkReachability, which is deprecated with a detailed message from Apple in their docs:
```
All SCNetworkReachability functions are deprecated, because network conditions change too frequently for preflight reachability tests to be accurate and useful. Instead of trying to perform a preflight check, just attempt your connection. The connection may not be available immediately, but you can use [waitsForConnectivity](https://developer.apple.com/documentation/Foundation/URLSessionConfiguration/waitsForConnectivity) in your [URLSessionConfiguration](https://developer.apple.com/documentation/Foundation/URLSessionConfiguration) to automatically connect as soon as it’s possible to do so. If you’re using the [Network](https://developer.apple.com/documentation/Network) framework, you can monitor the state of an [NWConnection](https://developer.apple.com/documentation/Network/NWConnection) ([nw_connection_t](https://developer.apple.com/documentation/Network/nw_connection_t) in Objective-C), which lets you know when the connection is waiting for a change in connectivity before automatically proceeding. In limited cases, if you need to change a network request based on the network you’re using, use [NWPathMonitor](https://developer.apple.com/documentation/Network/NWPathMonitor) to observe the system networking state.
```

### Solution Brainstorm

It would be nice if we didn't need reachability at all as Apple says in the docs, but I see it's used for re-sending failed requests when a connection is made, which makes sense to me. So I think we should switch it to NWPathMonitor. I've used this API before and it should be pretty straightforward, and it would support all platforms so we won't need `SENTRY_HAS_REACHABILITY`. Only catch is it supports iOS 12+. We technically still support iOS 11 but Xcode does not support iOS 11 anymore, so we might be able to just easily drop iOS 11 without a major SDK version increase.

philippsuess:
Yes, we should do this. This would also fix https://github.com/getsentry/sentry-cocoa/issues/5426. I think we should use `NWPathMonitor`.

> Only catch is it supports iOS 12+

I think now with Xcode 26 and that you must use [Xcode 16 to submit apps](https://developer.apple.com/news/upcoming-requirements/?id=02212025a), we can even bump to iOS 12 in a non-major version, cause like you said, you can't submit apps anymore with min iOS 11.

philippsuess:
[...]

PR #6019, opened 2025-08-29
Convert `SentryReachability` to Swift

itaybre:
#skip-changelog

### DIFF
--- a/Sources/Swift/Networking/SentryReachability.swift
+++ b/Sources/Swift/Networking/SentryReachability.swift
@@ -36,6 +36,10 @@ public class SentryReachability: NSObject {
 #if DEBUG || SENTRY_TEST || SENTRY_TEST_CI
     @objc public var skipRegisteringActualCallbacks = false
     private var ignoreActualCallback = false
+    
+    public var pathMonitorIsNil: Bool {
+        return pathMonitor == nil
+    }
 #endif // DEBUG || SENTRY_TEST || SENTRY_TEST_CI
     
     @objc(addObserver:)
--- a/Tests/SentryTests/Networking/SentryReachabilitySwiftTests.swift
+++ b/Tests/SentryTests/Networking/SentryReachabilitySwiftTests.swift
@@ -116,11 +116,8 @@ final class SentryReachabilitySwiftTests: XCTestCase {
     
     func testAddRemoveFromMultipleThreads() throws {
         let sut = SentryReachability()
-        // Calling the methods of SCNetworkReachability in a tight loop from
-        // multiple threads is not an actual use case, and it leads to flaky test
-        // results. With this test, we want to test if the adding and removing
-        // observers are adequately synchronized and not if we call
-        // SCNetworkReachability correctly.
+        // With this test, we want to test if the adding and removing
+        // observers are adequately synchronized.
         sut.skipRegisteringActualCallbacks = true
         testConcurrentModifications(writeWork: { _ in
             sut.add(TestSentryReachabilityObserver())
@@ -129,20 +126,27 @@ final class SentryReachabilitySwiftTests: XCTestCase {
         })
     }
 
-    /// This tests actually test NWPathMonitor response, if it becomes blaky we can disable it
-    func testRegisteringActualCallbacks_CallbackIsCalled() {
+    func testAddingAndRemovingObserversCleanTheMonitor() {
         reachability.skipRegisteringActualCallbacks = false
         reachability.setReachabilityIgnoreActualCallback(false)
-        
-        let expectation = XCTestExpectation(description: "Callback should be called")
-        
         let observer = TestSentryReachabilityObserver()
-        observer.onReachabilityChanged = { _, _ in
-            expectation.fulfill()
-        }
         
+        // Ensure starting scenario
+        XCTAssertTrue(reachability.pathMonitorIsNil)
+        
+        // Do
         reachability.add(observer)
         
-        wait(for: [expectation], timeout: 5)
+        // Verify
+        // Monitor should not be nil when at least one observer is added
+        XCTAssertFalse(reachability.pathMonitorIsNil)
+        
+        // Do again
+        sleep(1)
+        reachability.remove(observer)
+        
+        // Verify
+        // Ensure when all observers are removed, the monitor is set to nil
+        XCTAssertTrue(reachability.pathMonitorIsNil)
     }
 }


Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,298 @@
...
+#endif // DEBUG || SENTRY_TEST || SENTRY_TEST_CI
+        
+        sentryCurrentReachabilityState = kSCNetworkReachabilityFlagsUninitialized
+        
+        if let reachabilityRef = sentryReachabilityRef {
+            SentrySDKLog.debug("removing callback for reachability ref \(reachabilityRef)")
+            SCNetworkReachabilitySetCallback(reachabilityRef, nil, nil)
+            SCNetworkReachabilitySetDispatchQueue(reachabilityRef, nil)
+            sentryReachabilityRef = nil
+        }
```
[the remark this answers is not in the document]
itaybre: No need to release CoreFoundation object in Swift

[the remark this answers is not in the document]
itaybre: No need to release CoreFoundation object in Swift

philippsuess:
@itaybre we actually want to rewrite SentryReachability because it uses `SCNetworkReachability` see https://github.com/getsentry/sentry-cocoa/issues/5440. I'm unsure if it's worth converting it to Swift. If you still think we should do it, I can give this PR a review.

itaybre:
> @itaybre we actually want to rewrite SentryReachability because it uses `SCNetworkReachability` see #5440. I'm unsure if it's worth converting it to Swift. If you still think we should do it, I can give this PR a review.

I remember that change proposal, but while we still support iOS 11, we need to use this reachability since this will be a blocker for the swift conversion project

philippsuess:
> I remember that change proposal, but while we still support iOS 11, we need to use this reachability since this will be a blocker for the swift conversion project

We can also consider removing support for iOS 11 for SentryReachability. We only use it for network breadcrumbs and sending envelopes when a device gets online again. The impact of the removed functionality for iOS 11 users is minimal, and the market share is close to 0%.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,232 @@
...
+#if DEBUG || SENTRY_TEST || SENTRY_TEST_CI
+    @objc public var skipRegisteringActualCallbacks = false
+    private var ignoreActualCallback = false
+#endif // DEBUG || SENTRY_TEST || SENTRY_TEST_CI
+    
+    @objc(addObserver:)
+    public func add(_ observer: SentryReachabilityObserver) {
+        SentrySDKLog.debug("Adding observer: \(observer)")
+        
+        objc_sync_enter(reachabilityObservers)
```
philprime: `m`: I am seeing this sync solution for the first time in the repository. Is there an advantage of `objc_sync_enter` over other sync methods like locks?

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,232 @@
...
+        }
+        
+#if DEBUG || SENTRY_TEST || SENTRY_TEST_CI
+        if skipRegisteringActualCallbacks {
+            SentrySDKLog.debug("Skip registering actual callbacks")
+            return
+        }
+#endif // DEBUG || SENTRY_TEST || SENTRY_TEST_CI
+        
+        let reachabilityQueue = DispatchQueue(label: "io.sentry.cocoa.connectivity")
```
philprime: `m`: We could consider `SentryDispatchQueueWrapper` even if it is a 1:1 conversion

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,232 @@
...
+        if #available(iOS 12.0, macOS 10.14, tvOS 12.0, visionOS 1.0, *) {
+            let monitor = NWPathMonitor()
+            pathMonitor = monitor
+            monitor.pathUpdateHandler = pathUpdateHandler
+            monitor.start(queue: reachabilityQueue)
+            
+            SentrySDKLog.debug("Started NWPathMonitor")
+        } else {
+            // For iOS 11 and earlier, simulate always being connected via WiFi
+            SentrySDKLog.debug("NWPathMonitor not available. Using fallback: always connected via WiFi")
```
philprime: `m`: This should be a user-visible warning, not only in the debug logs.

```suggestion
            SentrySDKLog.warn("NWPathMonitor not available. Using fallback: always connected via WiFi")
```

philippsuess:
> @philipphofmann I would like us to stick to the plan and convert to Swift as much as possible, instead of rewriting parts of the SDK now.

@philprime, the conversion wasn't looking so great because the APIs are mostly ObjC low-level. As we want to ditch SentryReachability with `SCNetworkReachability` anyways, I raised the question if it wouldn't be better to replace the implementation with `NWPathMonitor` already as we would ditch the Swift `SCNetworkReachability` version soon.

Sources/Swift/Networking/SentryReachability.swift
itaybre: `NWPathMonitor` requires a queue to start

Sources/Swift/Networking/SentryReachability.swift
itaybre: It is the same behavior

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,233 @@
+import Foundation
+import Network
+
+#if !os(watchOS) && !((swift(>=5.9) && os(visionOS)) && SENTRY_NO_UIKIT)
```
philippsuess: `h`: Shouldn't we be able to make this available on all platforms now, as we're using `NWPathMonitor`, which works on 
```
iOS 12.0+
iPadOS 12.0+
Mac Catalyst 12.0+
macOS 10.14+
tvOS 12.0+
visionOS 1.0+
watchOS 5.0+
```

If yes, we should be able to remove `SENTRY_HAS_REACHABILITY`.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,233 @@
...
+
+// MARK: - SentryConectivity
+@objc
+public enum SentryConnectivity: Int {
```
philippsuess: `h`: Aren't we missing `@_spi(Private)` here? Why does this need to be public?

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,233 @@
...
+                monitor.pathUpdateHandler = self?.pathUpdateHandler
+                monitor.start(queue: self?.reachabilityQueue ?? DispatchQueue.global(qos: .background))
+            }
+        } else {
+            // For iOS 11 and earlier, simulate always being connected via WiFi
+            SentrySDKLog.warning("NWPathMonitor not available. Using fallback: always connected via WiFi")
+            reachabilityQueue.async { [weak self] in
+                self?.connectivityCallback(.wiFi)
+            }
+        }
```
philippsuess: `m`: I would rather make the whole SentryReachability class only available on the specific platforms. We cann add a NoOp `SentryReachability` for the platforms below, or we simply wait until we bump the OS versions with V9. That would reduce the complexity.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,233 @@
...
+        
+#if DEBUG || SENTRY_TEST || SENTRY_TEST_CI
+        if skipRegisteringActualCallbacks {
+            SentrySDKLog.debug("Skip registering actual callbacks")
+            return
+        }
+#endif // DEBUG || SENTRY_TEST || SENTRY_TEST_CI
+        
+        if #available(iOS 12.0, macOS 10.14, tvOS 12.0, *) {
```
philippsuess: `h`: This also works on visionOS 1.0+ and watchOS 5.0+ ; see https://developer.apple.com/documentation/network/nwpathmonitor

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,233 @@
...
+            // If we don't use the main queue to start the monitor, the app seems to freeze on iOS 14.8 (SauceLabs)
+            // Also do it async to avoid blocking the main thread
+            // Right now `SentryDispatchQueueWrapper.dispatchAsyncOnMainQueue` is not used because the SDK starts on the main thread,
+            // thus the block is executed in the main thread, not async.
+            DispatchQueue.main.async { [weak self] in
+                let monitor = NWPathMonitor()
+                self?.pathMonitor = monitor
+                monitor.pathUpdateHandler = self?.pathUpdateHandler
+                monitor.start(queue: self?.reachabilityQueue ?? DispatchQueue.global(qos: .background))
+            }
```
philippsuess: `h`: That looks a bit weird. The [docs](https://developer.apple.com/documentation/network/nwpathmonitor) don't mention anything about starting this on the main thread. Maybe the problem is something else. Have you tried another dispatch queue apart from main? If it also works on the `.background` queue maybe we have a deadlock and we're simply avoing that by now dispatch this to the main thread async.

Furthermore, we now have a potential race condidion because we set the `self.pathMonitor` to `nil` below in `stopMonitoring`, which uses the `observersLock`, but this in here doesn't. If we could remove dispatching this on the main thread, we're in the lock context and the race condition would go away.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// MARK: - SentryReachability
+@_spi(Private) @objc
+public class SentryReachability: NSObject {
+    private var reachabilityObservers = NSHashTable<SentryReachabilityObserver>.weakObjects()
+    private var currentConnectivity: SentryConnectivity = .none
+    private var pathMonitor: Any? // NWPathMonitor for iOS 12+
+    private let reachabilityQueue: DispatchQueue = DispatchQueue(label: "io.sentry.cocoa.connectivity", qos: .background, attributes: [])
+    private let observersLock = NSLock()
```
philippsuess: `m`: I think using a NSRecursiveLock here might be a good idea, because NSLock isn't reentrant. I think we don't have a risk of a deadlock right now, but it's very easy to add one in the future, if you're not extremely careful.

philipphofmann (commented):
Thanks for converting this to using NWPathMonitor. I maybe found a couple of problems.

Sources/Swift/Networking/SentryReachability.swift
itaybre: Currently it should never re entry, this is why I used just NSLock

Sources/Swift/Networking/SentryReachability.swift
itaybre: Definitely weird, but happened 100% of the times on SauceLabs (see old commits).
Either they are somehow causing this behavior by injecting the code. I tested other versions and didn't seem to happen there.
Nothing happened on a sample app just printing NWPathMonitor, so I am unsure about the real cause

Sources/Swift/Networking/SentryReachability.swift
itaybre: I used `SENTRY_HAS_REACHABILITY` values:
```
#if !TARGET_OS_WATCH && !(TARGET_OS_VISION && SENTRY_NO_UIKIT == 1)
#    define SENTRY_HAS_REACHABILITY 1
#else
#    define SENTRY_HAS_REACHABILITY 0
#endif
```

Sources/Swift/Networking/SentryReachability.swift
itaybre: Public for testing purposes with ObjC.
Regarding `@_spi(Private)`, I just added it

Sources/Swift/Networking/SentryReachability.swift
itaybre: Same as before, I was replicating `SENTRY_HAS_REACHABILITY` for a 1:1 conversion.
I can enable SentryReachability for every platform

Sources/Swift/Networking/SentryReachability.swift
itaybre: This if was intended to make it available for every platform unless NWPathMonitor was unavailable, not exclude specific platforms

Sources/Swift/Networking/SentryReachability.swift
philippsuess: Yes, currently. But like I said, it's so easy to mess this up if you're not careful. This is not a high threading environment and NSRecursiveLock is almost as fast as NSLock. I would prefer using NSRecursiveLock cause it shouldn't make a difference except it helps us to avoid deadlocks.

Sources/Swift/Networking/SentryReachability.swift
philippsuess: OK, you didn't want to change it. Thanks 👍 

But now it's better thanks cause it works on all platforms 😃 except Linux of course.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,240 @@
...
+    case none
+}
+#else
+@objc
+enum SentryConnectivity: Int {
+    case cellular
+    case wiFi
+    case none
+}
+#endif // DEBUG || SENTRY_TEST || SENTRY_TEST_CI
```
philippsuess: `m`: I think this should be enough. When you add `@_spi(Private)` it's not actually public. I think we can remove the `#if ... #else`. Or am I missing something?

```suggestion
@objc @_spi(Private)
public enum SentryConnectivity: Int {
    case cellular
    case wiFi
    case none
}
```

Sources/Swift/Networking/SentryReachability.swift
philippsuess: Can we wait until V9 with this? Then we can ditch this extra functionality.

Sources/Swift/Networking/SentryReachability.swift
philippsuess: I would like to investigate this further. Not knowing why something works is usually a risk of getting bitten later. Can we try `DispatchQueue.global.async` and see if the problem disappears? We can also investigate it together to find the root cause.

itaybre: > I would like to investigate this further. Not knowing why something works is usually a risk of getting bitten later. Can we try `DispatchQueue.global.async` and see if the problem disappears? We can also investigate it together to find the root cause.

Already tried without success, spent a while night on this issue. Without a device to properly test myself locally it is very hard to debug.
Feel free to build and debug this if you have an idea

Sources/Swift/Networking/SentryReachability.swift
itaybre: We can wait until then, but remember there are exactly 0 users from iOS 11

Sources/Swift/Networking/SentryReachability.swift
itaybre: `@_spi(Private)` doesn't make it private either.
Since this was only used for testing, I see no benefit on making not making it private.
I will just migrate the tests to Swift so we can avoid this discussion

Sources/Swift/Networking/SentryReachability.swift
itaybre: `almost` as fast.

But if you want to future proof, I can add it.

itaybre:
Looks like I was able to find the cause: https://github.com/Iterable/iterable-swift-sdk/issues/688
NWPathMonitor crashed if the min iOS as is lower than iOS 12, but SauceLabs was not reporting any crash.

This happens on iOS 14 and iOS15.

philippsuess:
> Looks like I was able to find the cause: [Iterable/iterable-swift-sdk#688](https://github.com/Iterable/iterable-swift-sdk/issues/688) NWPathMonitor crashed if the min iOS as is lower than iOS 12, but SauceLabs was not reporting any crash.
> 
> This happens on iOS 14 and iOS15.

Excellent, so this means we should remove the async main thread dispatch and wait until we bump the minOS versions before merging this?

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -14,14 +14,7 @@
...
 #import "SentrySwift.h"
 
-#if !TARGET_OS_WATCH
-#    import "SentryReachability.h"
-#endif // !TARGET_OS_WATCH
-
-@interface SentryHttpTransport ()
-#if SENTRY_HAS_REACHABILITY
-    <SentryReachabilityObserver>
-#endif // !TARGET_OS_WATCH
```
philippsuess, line 24: Much better thanks

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// MARK: - SentryReachability
+@_spi(Private) @objc
+public class SentryReachability: NSObject {
+    private var reachabilityObservers = NSHashTable<SentryReachabilityObserver>.weakObjects()
+    private var currentConnectivity: SentryConnectivity = .none
+    private var pathMonitor: NWPathMonitor?
+    private let reachabilityQueue: DispatchQueue = DispatchQueue(label: "io.sentry.cocoa.connectivity", qos: .background, attributes: [])
+    private let observersLock = NSRecursiveLock()
```
philippsuess, line 34: Thanks 👍

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,209 @@
...
+        
+        observersLock.lock()
+        defer { observersLock.unlock() }
+        
+        SentrySDKLog.debug("Synchronized to remove observer: \(observer)")
+        reachabilityObservers.remove(observer)
+        
+        if reachabilityObservers.count == 0 {
+            stopMonitoring()
+        }
```
philippsuess: `l`: You could also use the extension for NSRecursiveLock if you want.

```swift
         observersLock.synchronized {
            SentrySDKLog.debug("Synchronized to remove observer: \(observer)")
            reachabilityObservers.remove(observer)

            if reachabilityObservers.count == 0 {
                stopMonitoring()
            }
        }
```

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,209 @@
...
+        
+        if reachabilityObservers.contains(observer) {
+            SentrySDKLog.debug("Observer already added. Doing nothing.")
+            return
+        }
+        
+        reachabilityObservers.add(observer)
+        
+        if reachabilityObservers.count > 1 {
+            return
```
philippsuess, line 63: `l`: We could add an extra debug log here.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,209 @@
...
+        if skipRegisteringActualCallbacks {
+            SentrySDKLog.debug("Skip registering actual callbacks")
+            return
+        }
+#endif // DEBUG || SENTRY_TEST || SENTRY_TEST_CI
+        
+        let monitor = NWPathMonitor()
+        self.pathMonitor = monitor
+        monitor.pathUpdateHandler = self.pathUpdateHandler
+        monitor.start(queue: self.reachabilityQueue)
```
philippsuess: `l`: A bit shorter
```suggestion
        self.pathMonitor = NWPathMonitor()
        self.pathMonitor?.pathUpdateHandler = self.pathUpdateHandler
        self.pathMonitor?.start(queue: self.reachabilityQueue)
```

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,209 @@
...
+        currentConnectivity = .none
+        
+        // Clean up NWPathMonitor
+        if let monitor = pathMonitor {
+            SentrySDKLog.debug("Stopping NWPathMonitor")
+            monitor.cancel()
+            pathMonitor = nil
+        }
+        
+        SentrySDKLog.debug("Cleaning up reachability queue.")
```
philippsuess: `l`: That log message is a bit confusing to me. It states that we are about to clean something up, but there's no code after it in this method.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,209 @@
...
+    
+    private func connectivityShouldReportChange(_ connectivity: SentryConnectivity) -> Bool {
+        if connectivity == currentConnectivity {
+            SentrySDKLog.debug("No change in reachability state. ConnectivityShouldReportChange will return false for connectivity \(connectivity.toString()), currentConnectivity \(currentConnectivity.toString())")
+            return false
+        }
+        
+        currentConnectivity = connectivity
+        return true
+    }
```
philippsuess: `l`: I would put this method below the `connectivityCallback` as the `connectivityCallback` is the only method using it and it's a bit confusing to me to see this method before it gets used below.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,209 @@
...
+            return .cellular
+        } else {
+            return .wiFi
+        }
+#else
+        return .wiFi
+#endif // canImport(UIKit)
+    }
+    
+    private func connectivityShouldReportChange(_ connectivity: SentryConnectivity) -> Bool {
```
philippsuess: `l`: The method name doesn't suggest that we also update the `currentConnectivity`. It's `connectivityShouldReportChange`, so it indicates we only want to know if we should report the change. The method name doesn't reflect that it has a side effect. We can move the `currentConnectivity = connectivity` out of this method and call it below in `connectivityCallback` after we call this method here or we change the method name.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,209 @@
...
+        SentrySDKLog.debug("Entered synchronized region of SentryConnectivityCallback with connectivity: \(connectivity.toString())")
+        
+        guard reachabilityObservers.count > 0 else {
+            SentrySDKLog.debug("No reachability observers registered. Nothing to do.")
+            return
+        }
+        
+        guard connectivityShouldReportChange(connectivity) else {
+            SentrySDKLog.debug("ConnectivityShouldReportChange returned false for connectivity \(connectivity.toString()), will not report change to observers.")
+            return
```
philippsuess: `l`: We already log a similar log message in `connectivityShouldReportChange` when it returns false. We don't need both.

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,209 @@
...
+        }
+        
+        let connected = connectivity != .none
+        
+        SentrySDKLog.debug("Notifying observers...")
```
philippsuess: `l`: This could contain what we're notifing them with. So nofitying them with the connected and typeDescription.

Tests/SentryTests/Networking/SentryReachabilitySwiftTests.swift
```diff
@@ -1,7 +1,120 @@
...
+        XCTAssertEqual(1, observer.connectivityChangedInvocations)
+        
+        reachability.remove(observer)
+    }
+    
+    /**
+     * We only want to make sure running the actual registering and unregistering callbacks doesn't
+     * crash.
+     */
+    func testRegisteringActualCallbacks() {
```
philippsuess, line 108: Ups forgot one comment.

`m`: I think it would be great to have one test not using the `skipRegisteringActualCallbacks` and actually running the code for the `NWPathMonitor` and validating with a test expectation or something that we at least get one update via the `pathUpdateHandler`. I don't think we should change the connectivity somehow manually in the unit test. This test would ensure we glue everything correctly together, because now we don't have an integration test.

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@
 - Fixes warnings about minimum OS version being lower than Xcode supported version (#5591)
 - Fix rendering method for fast view rendering (#6360)
 
+### Improvements
+
+- Replaces the use of SCNetworkReachabilityFlags with NWPathMonitor (#6019)
```
philippsuess:
```suggestion
- Replace deprecated SCNetworkReachability with NWPathMonitor (#6019)
```

Sources/Swift/Networking/SentryReachability.swift
```diff
@@ -0,0 +1,205 @@
...
+        guard reachabilityObservers.count > 0 else {
+            SentrySDKLog.debug("No reachability observers registered. Nothing to do.")
+            return
+        }
+        
+        let previousConnectivity = currentConnectivity
+        currentConnectivity = connectivity
+        guard connectivityShouldReportChange(previousConnectivity, currentConnectivity) else {
+            return
+        }
```
philippsuess, line 165: Cleaner thanks

Tests/SentryTests/Networking/SentryReachabilitySwiftTests.swift
```diff
@@ -10,16 +122,27 @@ final class SentryReachabilitySwiftTests: XCTestCase {
...
+        
+        let expectation = XCTestExpectation(description: "Callback should be called")
+        
+        let observer = TestSentryReachabilityObserver()
+        observer.onReachabilityChanged = { _, _ in
+            expectation.fulfill()
+        }
+        
+        reachability.add(observer)
+
```
philippsuess, line 139: `m`: I think that doesn't work as we only get an update when the connectivity changes. I think it's OK to remove the XCTestExpectation and only validate that we can add and remove an observer. I guess here it would be OK to put a small delay of a second or so in between those two.

philipphofmann (approved):
Thank you @itaybre

Tests/SentryTests/Networking/SentryReachabilitySwiftTests.swift
itaybre: NWPathMonitor does seem to trigger the callback (almost) immediately, but not on all devices.
I will remove the expectation, just in case

PR #6021, opened 2025-08-29
fix(ci): Fix brew not installing local formulas

itaybre:
Homebrew made some changes recently which broke our ci, see discussion: https://github.com/orgs/Homebrew/discussions/6351

#skip-changelog


philprime (approved):
IIUC this is an unexpected change of Homebrew and should be fixed in a future release. Maybe it makes sense to create an issue to revert this in a couple of weeks. LGTM

philprime:
Merging this now because it's blocking other PRs

PR #6029, opened 2025-09-01
chore(ci): Update lint-cocoapods-specs.yml to remove paths

philprime:
The workflow was already migrated to use a file-change detection job, but forgot to remove the paths for pushes

#skip-changelog


philipphofmann (approved):
LGTM

Issue #6009, opened 2025-08-27
Adhoc builds are reported as enterprise build types


devioustree:
The check for an [enterprise build](https://github.com/getsentry/sentry-cocoa/blob/130c6ba8d92bc4cd14a4f40e39137ecfd2221f2d/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m#L482) is not sufficient to be able to distinguish Enterprise program distributed apps from Adhoc built and distributed apps.

If it's not possible to disambiguate these, then I think we should acknowledge that and update this to something like `"enterprise / adhoc"` or `"enterprise or adhoc"`

PR #6044, opened 2025-09-01
feat: Add build type for adhoc

itaybre:
## :scroll: Description

Adds a new build type `adhoc` based on the embedded mobileprovision profile content.

Fixes: #6009

## :pencil: Checklist


Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
```diff
@@ -469,6 +470,19 @@
...
 
+/**
+ * Check if the embdded.mobileprovision provisions all devices
+ * If true the profile is for `enterprise`, if not it is `adhoc`
+ */
+static bool
+mobileProvisionProfileProvisionsAllDevices(void)
+{
+    NSString *path = [[NSBundle mainBundle] pathForResource:@"embedded" ofType:@"mobileprovision"];
+    SentryMobileProvisionParser *parser = [[SentryMobileProvisionParser alloc] initWithPath:path];
```
philippsuess: `m`: We now get the path for the `mobileprovision` in two different places. Once here and once in hasEmbeddedMobileProvision https://github.com/getsentry/sentry-cocoa/blob/34ca22f558f73d31ce49e777fa6a28c088ec1ded/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m#L467C1-L470

I think we could move the logic from `hasEmbeddedMobileProvision` into the `SentryMobileProvisionParser` to only have it in one place.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
```diff
@@ -469,6 +470,19 @@
     return [[NSBundle mainBundle] pathForResource:@"embedded" ofType:@"mobileprovision"] != nil;
 }
 
+/**
+ * Check if the embdded.mobileprovision provisions all devices
+ * If true the profile is for `enterprise`, if not it is `adhoc`
+ */
+static bool
+mobileProvisionProfileProvisionsAllDevices(void)
```
philippsuess: `m`: I think this method could be part of the `SentryMobileProvisionParser`. All the `SentryMobileProvisionParser` needs to do is to check if there is an embedded mobile provision and if yes which one.

Sources/Swift/Helper/SentryMobileProvisionParser.swift
```diff
@@ -0,0 +1,63 @@
...
+        let options: PropertyListSerialization.ReadOptions = []
+#endif
+        
+        guard let obj = try? PropertyListSerialization.propertyList(from: plistData,
+                                                                    options: options,
+                                                                    format: &format),
+              let dict = obj as? [String: Any] else {
+            return
+        }
+        profileContent = dict
```
philippsuess: `m`: I don't think we need to store the whole dict. We only need to know if it contains `ProvisionsAllDevices` or not.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 - Lazily CharacterSet only once in SentryBaggageSerialization (#5871)
 - Structured Logging: Log `SentrySDK.logger` calls to `SentrySDKLog` (#5991)
+- Build type metadata now makes difference between `enterprise` and `adhoc` (#6044)
```
philippsuess: `m`: We can be a bit more specific so the user knows exactly where this surfaces.
```suggestion
- The build type in the app context now differentiates between `enterprise` and `adhoc` (#6044)
```

philipphofmann (commented):
Thanks for doing this @itaybre. I think we can improve the code a bit, so I added a few comments. I think this is a great opportunity to do some pair programming, so we both learn from each other. Otherwise, I see a bit of a risk of ping pong comments.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
```diff
@@ -478,8 +470,9 @@
     if (isDebugBuild()) {
         return "debug";
     }
-    if (hasEmbeddedMobileProvision()) {
-        return "enterprise";
+    SentryMobileProvisionParser *parser = [[SentryMobileProvisionParser alloc] init];
+    if ([parser hasEmbeddedMobileProvisionProfile]) {
+        return [parser mobileProvisionProfileProvisionsAllDevices] ? "enterprise" : "adhoc";
```
philippsuess, line 475: `m`: It would be great to add test for this, but I know we don't have tests for this and it could require refactoring.

Sources/Swift/Helper/SentryMobileProvisionParser.swift
```diff
@@ -0,0 +1,63 @@
...
+        return provisionsAllDevices
+    }
+    
+    @objc
+    convenience override public init() {
+        self.init(nil)
+    }
+    
+    @objc
+    public init(_ path: String?) {
```
philippsuess: `l`: Maybe you can add a comment why we make this nullable for testability and not having to pass in the path from the crash monitor system

Sources/Swift/Helper/SentryMobileProvisionParser.swift
```diff
@@ -0,0 +1,63 @@
...
+        super.init()
+        embeddedProfilePath = path ?? Bundle.main.path(forResource: "embedded", ofType: "mobileprovision")
+        guard let embeddedProfilePath else { return }
```
philippsuess: `m`: Maybe we can add a log message for debugging in the return here.

Sources/Swift/Helper/SentryMobileProvisionParser.swift
```diff
@@ -0,0 +1,63 @@
...
+            return
+        }
+        
+        // The file is a CMS (PKCS#7) container with a plist embedded inside as text.
+        // Convert to Latin-1 to preserve the bytes -> string even if not UTF-8.
+        guard let payload = String(data: fileData, encoding: .isoLatin1),
+              let startRange = payload.range(of: "<plist"),
+              let endRange = payload.range(of: "</plist>") else {
+            return
+        }
```
philippsuess: `l`: We could add a few debug log messages here in the return statements.

philipphofmann (approved):
LGTM with a few suggestions.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
itaybre: We don't have tests for this right now, and it *would* require some refactoring since the logic to get the build type depends on some precompiler macros

Issue #3798, opened 2024-03-27
Crash in [SentrySubClassFinder actOnSubclassesOfViewControllerInImage] (SentrySubClassFinder.m:61)


dimacorp:
### Platform

iOS

### Environment

Production, Develop

### Installed

CocoaPods

### Version

8.22.4

### Steps to Reproduce

1. Add Swift conditional class for example
import RoomPlan
@available(iOS 17.0, *)
class RoomCaptureTools {
    private var finalResults: CapturedStructure?
}
This class is a helper class to a feature available on latest iOS while the app itself supports iOS 13
2. Integrate Sentry
3. Start the app on older iOS (for example 15.5)

### Expected Result

no crash

### Actual Result

The app crashes in SentrySDK startWithConfigureOptions

Thread 3 Queue : sentry-ui-view-controller-swizzling (serial)
#0	0x0000000000000000 in 0x00000000 ()
#1	0x00000001043eb498 in type metadata accessor for CapturedStructure? ()
#2	0x00000001043eb3ac in type metadata completion function for RoomCaptureTools ()
#3	0x000000018f9c7930 in swift::MetadataCacheEntryBase<(anonymous namespace)::SingletonMetadataCacheEntry, int>::doInitialization(swift::MetadataWaitQueue::Worker&, swift::MetadataRequest) ()
#4	0x000000018f9b8cf8 in swift_getSingletonMetadata ()
#5	0x00000001043eb360 in type metadata accessor for RoomCaptureTools ()
#6	0x00000001043eb314 in ObjC metadata update function for RoomCaptureTools ()
#7	0x000000018017d938 in realizeClassMaybeSwiftMaybeRelock(objc_class*, mutex_tt<false>&, bool) ()
#8	0x0000000180183548 in look_up_class ()
#9	0x00000001807e7cb8 in NSClassFromString ()
#10	0x000000010514b4c0 in __69-[SentrySubClassFinder actOnSubclassesOfViewControllerInImage:block:]_block_invoke at /Pods/Sentry/Sources/Sentry/SentrySubClassFinder.m:61
#11	0x00000001050e0f04 in __53-[SentryDispatchQueueWrapper dispatchAsyncWithBlock:]_block_invoke at /Pods/Sentry/Sources/Sentry/SentryDispatchQueueWrapper.m:31
#12	0x0000000104db433c in _dispatch_call_block_and_release ()
#13	0x0000000104db5b94 in _dispatch_client_callout ()
[...]

PR #6045, opened 2025-09-02
fix(session-replay): Add exemption for CameraUI traversal for iOS 26.0

### LINKED ISSUE #5647: SessionReplay crashes when launching the camera in an app in iOS 26
### Platform

iOS

### Environment

Production

### Installed

CocoaPods

### Version

8.53.2

### Xcode Version

16.4

No

### Steps to Reproduce

If options.sessionReplay.sessionSampleRate = Float(1.0) or
options.sessionReplay.onErrorSampleRate = Float(1.0) is set when initializing sentry, the app will crash when the following code is used to start the camera in the app. This only happens in iOS 26. It is normal in earlier versions of iOS.

> This problem exists whether it is initialized in OC or in Swift.

```kotlin
actual suspend fun FileKit.openCameraPicker(
    type: FileKitCameraType
): PlatformFile? = withContext(Dispatchers.Main) {
    suspendCoroutine { continuation ->
        cameraControllerDelegate = CameraControllerDelegate(
            onImagePicked = { image ->
                if (image != null) {
                    // Convert UIImage to NSData (JPEG format with compression quality 1.0)
                    val imageData = UIImageJPEGRepresentation(image, 1.0)

                    // Get the path for the temporary directory
                    val tempDir = NSTemporaryDirectory()
                    val fileName = "image_${NSUUID().UUIDString}.jpg"
                    val filePath = tempDir + fileName

                    // Create an NSURL for the file path
                    val fileUrl = NSURL.fileURLWithPath(filePath)

                    // Write the NSData to the file
                    if (imageData?.writeToURL(fileUrl, true) == true) {
                        // Return the NSURL of the saved image file
                        continuation.resume(PlatformFile(fileUrl))
                    } else {
                        // If saving fails, return null
                        continuation.resume(null)
                    }
                } else {
                    continuation.resume(null)
                }
            }
        )

        val pickerController = UIImagePickerController()
        pickerController.sourceType =
            UIImagePickerControllerSourceType.UIImagePickerControllerSourceTypeCamera
        pickerController.delegate = cameraControllerDelegate

        UIApplication.sharedApplication.firstKeyWindow?.rootViewController?.presentViewController(
            pickerController,
            animated = true,
            completion = null
        )
    }
}
```

This is a screenshot of xcode：

Image

Image

Image

This is a screenshot of Sentry：

Image

Image

### Expected Result

Don't Crash

### Actual Result

App Crash

_No response_

philprime:
## :scroll: Description

This pull request introduces a targeted workaround in the `SentryUIRedactBuilder` to handle a crash related to `CameraUI.ChromeSwiftUIView` on iOS 26 when built with Xcode 16. The workaround ensures that the redaction process skips subtrees of this specific internal camera view to prevent the crash, while maintaining normal redaction behavior for other views. 

* Added a runtime check in `SentryUIRedactBuilder` to detect and skip subtrees of `CameraUI.ChromeSwiftUIView` when running on iOS 26+ to prevent crashes from unimplemented initializers. (`cameraSwiftUIViewClassObjectId`, `isViewSubtreeIgnored`, `redactRegionsFor(view:)`) [[1]](diffhunk://#diff-a8f2efc99fe1ea5939407dec83583a1dc4981e500e0df9112ee39b554d4924baR12-R32) [[2]](diffhunk://#diff-a8f2efc99fe1ea5939407dec83583a1dc4981e500e0df9112ee39b554d4924baL286-R330) [[3]](diffhunk://#diff-a8f2efc99fe1ea5939407dec83583a1dc4981e500e0df9112ee39b554d4924baL259-R285)
* Updated redaction region naming to use `view.debugDescription` instead of `layer.name` for more consistent identification. [[1]](diffhunk://#diff-a8f2efc99fe1ea5939407dec83583a1dc4981e500e0df9112ee39b554d4924baL242-R261) [[2]](diffhunk://#diff-a8f2efc99fe1ea5939407dec83583a1dc4981e500e0df9112ee39b554d4924baL275-R297)
* Modified `ExtrasViewController.swift` to present the camera UI using `UIImagePickerController` for testing camera permission flows when pressing the button `Show Camera UI`.
* Added an UI test triggering the camera UI in CI without crashing to detect regression.

Fixes #5647

- Added additional unit tests.
- Compiled the iOS-Swift sample using Xcode 16 and ran the app in iOS 26 Beta 7 on simulator and device. It crashed without the changes but does not crash with the changes.

## :pencil: Checklist

## Full Investigation

The issue could be reproduced when using Xcode 16.4 to run on iOS 26.0 Beta 6 or later, where accessing the `layer.subLayers` during the redaction geometry calculations while displaying the standard Camera UI caused a crash:

```
Fatal error: Use of unimplemented initializer 'init(layer:)' for class 'CameraUI.ModeLoupeLayer'
```

Screenshot 2025-09-09 at 09 41 28

During my investigation I found the class to be part of the private framework `CameraUI`, located at `/Library/Developer/CoreSimulator/Volumes/iOS_23A5326a/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 26.0.simruntime/Contents/Resources/RuntimeRoot/System/Library/PrivateFrameworks/CameraUI.framework` when running on iOS 26.0 Beta 6.

My approach to resolve the issue is excluding the view hierarchy of the camera UI from traversal and instead fully redact the UI, as it can be considered personal information anyways. I found the `CameraUI.ModeLoupeLayer` to be part of the view hierarchy of the `CameraUI.ChromeSwiftUIView`, so I use it to decide if the subview hierarchy should be ignored from traversal.

While working on the unit tests to verify the changes redaction algorithm, I encountered issues in creating an instance of the `CameraUI.ChromeSwiftUIView` as it is a private framework which can not be imported directly into the unit tests.

Instead I tried to find the class using `NSClassFromString("CameraUI.ChromeSwiftUIView")` which also returned `nil` because the framework was not yet loaded. I was able to resolve this by creating an instance of `UIImagePickerController` (the original cause of the crash) which then transitively imports the framework `CameraUI`.

```swift
let cameraUIViewClass: AnyClass = try XCTUnwrap(NSClassFromString("CameraUI.ChromeSwiftUIView"))
```

To analyse the class, I created the following dump tool to access the available Objective-C runtime information:

```swift
func dumpObjC(_ cls: AnyClass) {
    func listMethods(of c: AnyClass, title: String) {
        var count: UInt32 = 0
        guard let list = class_copyMethodList(c, &count) else {
            print("No methods for \(title)")
            return
        }
        print("=== \(title) (\(count)) ===")
        for i in 0..<Int(count) {
            let m = list[i]
            let sel = method_getName(m)
            if let name = NSStringFromSelector(sel) as String? {
                let numArgs = method_getNumberOfArguments(m)
                let typeEnc = String(cString: method_getTypeEncoding(m)!)
                print("- \(name)  args:\(numArgs)  types:\(typeEnc)")
            }
        }
        free(list)
    }
    func listProps(of c: AnyClass, title: String) {
        var count: UInt32 = 0
        if let props = class_copyPropertyList(c, &count) {
            print("=== \(title) properties (\(count)) ===")
            for i in 0..<Int(count) {
                let p = props[i]
                let name = String(cString: property_getName(p))
                let attrs = String(cString: property_getAttributes(p)!)
                print("- \(name)  attrs:\(attrs)")
            }
            free(props)
        }
    }
    func listIvars(of c: AnyClass, title: String) {
        var count: UInt32 = 0
        if let ivars = class_copyIvarList(c, &count) {
            print("=== \(title) ivars (\(count)) ===")
            for i in 0..<Int(count) {
                let iv = ivars[i]
                let name = String(cString: ivar_getName(iv)!)
                let type = String(cString: ivar_getTypeEncoding(iv)!)
                print("- \(name)  type:\(type)  offset:\(ivar_getOffset(iv))")
            }
            free(ivars)
        }
    }

    let meta: AnyClass = object_getClass(cls)! // metaclass (holds class methods)
    print("Class: \(NSStringFromClass(cls))")
    print("Superclass: \(NSStringFromClass(class_getSuperclass(cls) ?? NSObject.self))")
    print("Is meta: \(class_isMetaClass(cls))")
    listMethods(of: cls, title: "Instance methods")
    listMethods(of: meta, title: "Class methods")
    listProps(of: cls, title: "Instance")
    listIvars(of: cls, title: "Instance")
}
```

This resulted in the following output:

```
Class: CameraUI.ChromeSwiftUIView
Superclass: UIView
Is meta: false
=== Instance methods (6) ===
- .cxx_destruct  args:2  types:v16@0:8
- initWithCoder:  args:3  types:@24@0:8@16
- initWithFrame:  args:3  types:@48@0:8{CGRect={CGPoint=dd}{CGSize=dd}}16
- hitTest:withEvent:  args:4  types:@40@0:8{CGPoint=dd}16@32
- layoutSubviews  args:2  types:v16@0:8
- pointInside:withEvent:  args:4  types:B40@0:8{CGPoint=dd}16@32
No methods for Class methods
=== Instance ivars (2) ===
- viewModel  type:  offset:408
- scenePhaseView  type:  offset:416
```

This shows that the `initWithCoder` and `initWithFrame` exist as expected, but as shown in the crash message might not be implemented.

My initial approach to get an instance of the view was using the instance of the `UIImagePickerController` and its subviews for the redaction testing. Unfortunately the image picker controller seems to be empty at the beginning and loads additional views after being displayed - most likely also due to permissions checking.

❌ Instance of UIImagePickerController

My second attempt was using the `NSClassFromString` to create the reference of the class type to then call it's init method with or without a frame:

```swift
let cls: AnyClass = try XCTUnwrap(NSClassFromString("CameraUI.ChromeSwiftUIView"))
let CameraViewType = try XCTUnwrap(cls as? UIView.Type)
print(CameraViewType)

let cameraUIView = CameraViewType.init(frame: .init(origin: .zero, size: .init(width: 40, height: 40)))
print(cameraUIView) // ❌ Fatal error: Use of unimplemented initializer 'init(frame:)' for class 'CameraUI.ChromeSwiftUIView'

let cameraUIView2 = CameraViewType.init()
print(cameraUIView2) // ❌ Fatal error: Use of unimplemented initializer 'init(frame:)' for class 'CameraUI.ChromeSwiftUIView'

let data = try! NSKeyedArchiver.archivedData(withRootObject: [:], requiringSecureCoding: false)
let coder = try! NSKeyedUnarchiver(forReadingFrom: data)
coder.requiresSecureCoding = false
let cameraUIView3 = try XCTUnwrap(CameraViewType.init(coder: coder))
print(cameraUIView3) ❌ CameraUI/ChromeSwiftUIView.swift:37: Fatal error: init(coder:) has not been implemented
```

I concluded that it’s a Swift class that likely has a Swift-only designated initializer (internal/private, not @objc) and the ObjC inits (-init, -initWithFrame:, -initWithCoder:) are intentionally trapped to fatalError.

- Your ObjC-runtime dump only shows @objc methods. Swift-only initializers won’t appear there.
- initWithCoder: is explicitly unimplemented, so nib/storyboard is out.
- There is almost certainly a factory somewhere in CameraUI that returns this view.

❌ Invoking init methods of `CameraUI.ChromeSwiftUIView` via Objective-C runtime

My next approach was circumventing the traps by swizzling the init methods of the `ChromeSwiftUIView` with the init methods of it's superclass `UIView`:

```swift
// Replace -init with UIView's -init
let initSel = NSSelectorFromString("init")
let initMethod = try XCTUnwrap(class_getInstanceMethod(UIView.self, initSel))
let initImp = method_getImplementation(initMethod)
let initTypes = method_getTypeEncoding(initMethod)
class_replaceMethod(cls, initSel, initImp, initTypes)
defer {
    // TODO: revert the swizzling
}

// Replace -initWithFrame: with UIView's -initWithFrame:
let initWithFrameSel = NSSelectorFromString("initWithFrame:")
let initWithFrameMethod = try XCTUnwrap(class_getInstanceMethod(UIView.self, initWithFrameSel))
let initWithFrameImp = method_getImplementation(initWithFrameMethod)
let initWithFrameTypes = method_getTypeEncoding(initWithFrameMethod)
class_replaceMethod(cls, initWithFrameSel, initWithFrameImp, initWithFrameTypes)
defer {
    // TODO: revert the swizzling
}

let cameraUIView = try XCTUnwrap((cls as? UIView.Type)?.init(frame: .zero))
print(cameraUIView) // ✅ <CameraUI.ChromeSwiftUIView: 0x103c2ce90; frame = (0 0; 0 0); layer = <CALayer: 0x600000cf4b70>>
```

✅ Swizzling the type initializer

As the solution above is replacing class type methods which can have side effects, I also explored the option of raw object initialization using the Objective-C runtime

```swift
// 1) Allocate object memory without calling subclass init
let cameraView = try XCTUnwrap(class_createInstance(cls, 0) as? UIView)

// 2) Reinitialize storage using UIView’s initWithFrame:
typealias InitWithFrame = @convention(c) (AnyObject, Selector, CGRect) -> AnyObject
let sel = NSSelectorFromString("initWithFrame:")
let m = try XCTUnwrap(class_getInstanceMethod(UIView.self, sel))
let f = unsafeBitCast(method_getImplementation(m), to: InitWithFrame.self)
_ = f(cameraView, sel, .zero)

print(cameraView) // ✅ <CameraUI.ChromeSwiftUIView: 0x106e05270; frame = (0 0; 0 0); layer = <CALayer: 0x600000cf5110>>
```

✅ Swizzling the instance initializer

This implementation still relies on Objective-C, but I wanted to check if I can directly call the private Swift initializer too. I started to look into the LLVM symbol table dumper of the `CameraUI` framework to see if I can find any 

```bash
$ export FRAMEWORK="/Library/Developer/CoreSimulator/Volumes/iOS_23A5326a/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 26.0.simruntime/Contents/Resources/RuntimeRoot/System/Library/PrivateFrameworks/CameraUI.framework/CameraUI"
$ xcrun nm -m "$FRAMEWORK" | grep ChromeSwiftUIView
0000000000260cb8 (__TEXT,__text) non-external -[_TtC8CameraUI17ChromeSwiftUIView .cxx_destruct]
0000000000260474 (__TEXT,__text) non-external -[_TtC8CameraUI17ChromeSwiftUIView hitTest:withEvent:]
000000000025fe64 (__TEXT,__text) non-external -[_TtC8CameraUI17ChromeSwiftUIView initWithCoder:]
0000000000260c5c (__TEXT,__text) non-external -[_TtC8CameraUI17ChromeSwiftUIView initWithFrame:]
0000000000260140 (__TEXT,__text) non-external -[_TtC8CameraUI17ChromeSwiftUIView layoutSubviews]
00000000002604dc (__TEXT,__text) non-external -[_TtC8CameraUI17ChromeSwiftUIView pointInside:withEvent:]
000000000059e9f0 (__DATA,__objc_data) non-external (was a private external) _OBJC_CLASS_$__TtC8CameraUI17ChromeSwiftUIView
00100000005a8a48 (__TEXT,__text) non-external _OBJC_IVAR_$__TtC8CameraUI17ChromeSwiftUIView.scenePhaseView
00100000005a8a40 (__TEXT,__text) non-external _OBJC_IVAR_$__TtC8CameraUI17ChromeSwiftUIView.viewModel
00000000005a8a50 (__DATA,__data) non-external (was a private external) _OBJC_METACLASS_$__TtC8CameraUI17ChromeSwiftUIView
000000000059e990 (__DATA,__objc_data) non-external __DATA__TtC8CameraUI17ChromeSwiftUIView
0000000000327fa8 (__TEXT,__objc_methlist) non-external __INSTANCE_METHODS__TtC8CameraUI17ChromeSwiftUIView
000000000056c8c0 (__DATA_CONST,__objc_const) non-external __IVARS__TtC8CameraUI17ChromeSwiftUIView
000000000056c878 (__DATA_CONST,__objc_const) non-external __METACLASS_DATA__TtC8CameraUI17ChromeSwiftUIView
00000000003b46ce (__TEXT,__swift5_typeref) non-external (was a private external) _symbolic _____ 8CameraUI17ChromeSwiftUIViewC
00000000003cbd96 (__TEXT,__swift5_typeref) non-external (was a private external) _symbolic _____Sg 8CameraUI17ChromeSwiftUIViewC
```

To summarize this output, the available symbols only include the `_TtC8CameraUI17ChromeSwiftUIView` demangles to `CameraUI.ChromeSwiftUIView` (Swift class bridged to ObjC) and class and metaclass exist in the ObjC runtime (so `NSClassFromString` works), but symbol lookup by name (`(lldb) image lookup -rn ChromeSwiftUIView`) finds nothing because the relevant symbols aren’t exported.

**❌ This means we can not `dlsym` a Swift initializer to invoke it directly and instance swizzling is the best approach**.

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift-UITests/SessionReplayUITests.swift
+++ b/Samples/iOS-Swift/iOS-Swift-UITests/SessionReplayUITests.swift
@@ -19,7 +19,7 @@ class SessionReplayUITests: BaseUITest {
 
         // After the Camera UI is shown, we keep it open for 6 seconds to trigger at least one full
         // video segment captured (segments are 5 seconds long).
-        wait(6)
+        wait(10)
 
         // We know the test succeeded if we reach this point without the app crashing.
     }


Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,8 +9,27 @@ import WebKit
...
+    /// causes a crash due to unimplemented init(layer:) initializer.
+    private static let cameraSwiftUIViewClassObjectId: ObjectIdentifier? = {
+#if compiler(>=6.0) // Xcode 16+ uses Swift 6.0 compiler
+        guard let classType = NSClassFromString("CameraUI.ChromeSwiftUIView") else {
+            return nil
+        }
+        return ObjectIdentifier(classType)
+#else
+        return nil
+#endif
```
philprime:
```suggestion
        guard let classType = NSClassFromString("CameraUI.ChromeSwiftUIView") else {
            return nil
        }
        return ObjectIdentifier(classType)
```

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,8 +9,23 @@ import WebKit
...
 final class SentryUIRedactBuilder {
+    // MARK: - Constants
+
+    /// Class object identifier for ``CameraUI.ChromeSwiftUIView``, if it exists.
+    ///
+    /// This object identifier is used to identify views of this class type during the redaction process.
+    /// This workaround is specifically for Xcode 16 building for iOS 26 where accessing CameraUI.ModeLoupeLayer
+    /// causes a crash due to unimplemented init(layer:) initializer.
+    private static let cameraSwiftUIViewClassObjectId: ObjectIdentifier? = {
+        guard let classType = NSClassFromString("CameraUI.ChromeSwiftUIView") else {
```
philippsuess: `h`: When looking at the crash in https://github.com/getsentry/sentry-cocoa/issues/5647 it could be that simply calling `NSClassFromString` could still lead to crashes. We had a weird issue in the past when swizzling UIViewControllers (https://github.com/getsentry/sentry-cocoa/issues/3798) and therefore we don't call it in the SentrySubClassFinder for excluded classes https://github.com/getsentry/sentry-cocoa/blob/15a6325c9249637592cd76d09c4c1cc45489b504/Sources/Sentry/SentrySubClassFinder.m#L72-L77

I even opened a bug in the Swift repo but it never got fixed: https://github.com/swiftlang/swift/issues/72657. I'm unsure if it's worth the effort to do this right now. Simply doing a check if we're running on iOS 26 could be safer.

philprime:
Using `NSClassFromString` will load the type into the Objective-C runtime and call it's `+initialize()` method, which could cause side-effects. Instead I will refactor it to convert the type of the current view and compare by their string name instead.

philprime:
@cursor review

philprime:
@philipphofmann we can wait with this PR until after #6075 so we have verification in CI that the tests work as expected.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Fix macOS's frameworks not following the versioned framework structure (#6049)
 - Add warning to addBreadcrumb when used before SDK init (#6083)
 - Add null-handling for parsed DSN in SentryHTTPTransport (#5800)
+- Add exemption for CameraUI traversal for iOS 26.0 (#6045)
```
philippsuess: `m`: I would focus more on the user impact in the changelog. We fixed a SR crash releate to `CameraUI` if I'm not mistaken. I would simply state that, cause that's what the users care about. The downside is now that we redact more parts of the screen, if yes, I think we can also point this out here.

Samples/iOS-Swift/iOS-Swift-UITests/SessionReplayUITests.swift
```diff
@@ -0,0 +1,32 @@
...
+        // We need to verify the camera UI is shown by checking for the existence of a UI element.
+        // This can be any element that is part of the camera UI and can be found reliably.
+        // The "PhotoCapture" button is a good candidate as it is always present when the
+        // camera UI is shown.
+        let cameraUIElement = app.buttons["PhotoCapture"]
+        XCTAssertTrue(cameraUIElement.waitForExistence(timeout: 5))
+
+        // After the Camera UI is shown, we keep it open for 10 seconds to trigger at least one full
+        // video segment captured (segments are 5 seconds long).
+        wait(10)
```
philippsuess, line 22: `m`: If I'm not mistaken, we don't have a beforeSendReplay option yet. Could we use beforeSend and wait until we get an event if an SR context instead? Or could we force a full video segment with some SR APIs? Doing timeouts in tests easily leads to flakiness. We can also keep it as is, if we don't find a better solution.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -283,11 +307,27 @@ final class SentryUIRedactBuilder {
...
-                name: layer.name ?? layer.debugDescription
+                name: view.debugDescription
             ))
         }
     }
 
+    private func isViewSubtreeIgnored(_ view: UIView) -> Bool {
+        // We are using the string description of the type instead of converting it to ObjectIdentifier, because
+        // the conversion would require an to use `NSClassFromString` which can lead to crashes in some cases, as it
+        // calls the `+initialize` methods of the class.
```
philippsuess: `m`: We avoid using `NSClassFromString` because we had bad experiences in the past. I'm not sure if we have proof that it also crashes in that scenario. I think it would be great to reference SubClassFinder, which explains a crash we experienced in detail. It would be great to add the info Noah posted about another crash they experienced. It gives us more context around the decision of not using `NSClassFromString`.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -283,11 +307,27 @@ final class SentryUIRedactBuilder {
                 size: layer.bounds.size,
                 transform: newTransform,
                 type: .clipBegin,
-                name: layer.name ?? layer.debugDescription
+                name: view.debugDescription
             ))
         }
     }
 
+    private func isViewSubtreeIgnored(_ view: UIView) -> Bool {
```
philippsuess, line 315: `l`: I think the function name could reflect that we only check if the view is a `CameraUI.ChromeSwiftUIView`. We can change the function name later, but now it only checks for `CameraUI.ChromeSwiftUIView`.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -283,11 +307,27 @@ final class SentryUIRedactBuilder {
...
+        // We are using the string description of the type instead of converting it to ObjectIdentifier, because
+        // the conversion would require an to use `NSClassFromString` which can lead to crashes in some cases, as it
+        // calls the `+initialize` methods of the class.
+        let viewTypeId = type(of: view).description()
+        if #available(iOS 26.0, *), viewTypeId == Self.cameraSwiftUIViewClassId {
+            // CameraUI.ChromeSwiftUIView is a special case because it contains layers which can not be iterated due to this error:
```
philippsuess, line 334: `h`: Are there any other `CameraUI` views we should ignore potentially? We could do a more conservative approach of ignoring all views that start with `CameraUI.`.

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests.swift
```diff
@@ -768,6 +768,215 @@ class SentryUIRedactBuilderTests: XCTestCase {
...
+        // This works, because we don't actually use any of the logic of the `CameraUI.ChromeSwiftUIView` and only need
+        // an instance with the expected type.
+        typealias InitWithFrame = @convention(c) (AnyObject, Selector, CGRect) -> AnyObject
+        let sel = NSSelectorFromString("initWithFrame:")
+        let m = try XCTUnwrap(class_getInstanceMethod(UIView.self, sel))
+        let f = unsafeBitCast(method_getImplementation(m), to: InitWithFrame.self)
+        _ = f(cameraView, sel, .zero)
+
+        // Assert that the initialization worked but the type is still the expected one
+        XCTAssertEqual(type(of: cameraView).description(), "CameraUI.ChromeSwiftUIView")
```
philippsuess, line 837: `m`: I'm a bit unsure about the testing approach here. After also reading the PR description, I'm unsure if we should try to initialize a class of a private framework to test our redaction functionality. This test seems to know too much about the implementation and migh tbreak easily. Was your intention to have a test failing when the `UIImagePickerController` stops using the `CameraUI.ChromeSwiftUIView`? 

Wouldn't a test only using the `UIImagePickerController` be sufficient, because our implementation already checks for the `CameraUI.ChromeSwiftUIView`, so if the `UIImagePickerController` stops using the `CameraUI.ChromeSwiftUIView` this test would already fail?

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests.swift
```diff
@@ -768,6 +768,215 @@ class SentryUIRedactBuilderTests: XCTestCase {
...
+        cameraView.addSubview(label)
+
+        // -- Act --
+        let sut = getSut()
+        let result = sut.redactRegionsFor(view: rootView)
+
+        // -- Assert --
+        // With the fix, we should get exactly ONE redaction region for the CameraUI view
+        // The label inside should NOT create a separate redaction region because the
+        // CameraUI view is handled with early return in isViewSubtreeIgnored
```
philippsuess, line 935: Thanks for testing these edge cases and adding comments.

philipphofmann (commented):
Thanks a lot for tackling this complex crash. I have a few questions.

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests.swift
philippsuess: Another thought is to rely on a UI test to validate this.

CHANGELOG.md
philprime: I'll update the changelog to be more user-focused 👍

Samples/iOS-Swift/iOS-Swift-UITests/SessionReplayUITests.swift
philprime: This is an UI test, so we do not have any access to beforeSend etc.
We are also not testing if the session replay is triggered/sent etc.
This is just a smoke test making sure it doesn't crash.
Without my fix this test would not succeed on iOS 26.0

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I'll look up the Slack conversation and add more context.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I am uncertain about this approach, as our current solution tries to keep a minimal impact and masking everything related to the camera by default could make issues like #6104 even worse.

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests.swift
philprime: The only way to verify the adapted redaction logic is with an instance of `CameraUI.ChromeSwiftUIView`. I was not able to get the instance by creating an `UIImagePickerController`, so I had to do it manually. This test doesn't really care much about the `UIImagePickerController`. 

How would you test this using an UI test? We don't have any setup to verify redaction from UI tests and we also do not write any information to disk which we could use (e.g. geometry metadata in addition to screenshots)

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I am going to keep it this way right now, because when reading the implementation of `mapRedactRegion` the readability is better, with the "special views" being abstracted away.

```swift
    private func mapRedactRegion(fromLayer layer: CALayer, relativeTo parentLayer: CALayer?, redacting: inout [SentryRedactRegion], rootFrame: CGRect, transform: CGAffineTransform, forceRedact: Bool = false) {
        guard !redactClassesIdentifiers.isEmpty && !layer.isHidden && layer.opacity != 0, let view = layer.delegate as? UIView else {
            return
        }
        let newTransform = concatenateTranform(transform, from: layer, withParent: parentLayer)
        
        // Check if the subtree should be ignored to avoid crashes with some special views.
        // If a subtree is ignored, it will be fully redacted and we return early to prevent duplicates.
        if isViewSubtreeIgnored(view) {
            redacting.append(SentryRedactRegion(
                size: layer.bounds.size,
                transform: newTransform,
                type: .redact,
                color: self.color(for: view),
                name: view.debugDescription
            ))
            return
        }
        ...
```

philprime:
@philipphofmann is the new UI test failing on your machine too? It passes on my machine but fails in CI.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Fix crash in Session Replay when opening the camera UI on iOS 26+ by skipping redaction of internal views.
```
philippsuess, line 7: Way better thanks

Samples/iOS-Swift/iOS-Swift-UITests/SessionReplayUITests.swift
philippsuess: OK fair enough. I would keep the 10 seconds then. With 6 seconds, I see a risk of this becoming flaky.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -283,11 +307,40 @@ final class SentryUIRedactBuilder {
...
+        // Instead, we use the string description of the type (i.e., `type(of: view).description()`)
+        // for comparison. This is a safer, more "Swifty" approach that avoids the pitfalls of
+        // class initialization side effects.
+        //
+        // We have previously encountered related issues:
+        // - In EmergeTools' snapshotting code where using `NSClassFromString` led to crashes [1]
+        // - In Sentry's own SubClassFinder where storing or accessing class objects on a background thread caused crashes due to `+initialize` being called on UIKit classes [2]
+        //
+        // [1] https://github.com/EmergeTools/SnapshotPreviews/blob/main/Sources/SnapshotPreviewsCore/View%2BSnapshot.swift#L248
+        // [2] https://github.com/getsentry/sentry-cocoa/blob/00d97404946a37e983eabb21cc64bd3d5d2cb474/Sources/Sentry/SentrySubClassFinder.m#L58-L84
```
philippsuess, line 330: Golden thanks 🥇

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philippsuess: OK, step by step. I just wanted to double check

philippsuess:
> @philipphofmann is the new UI test failing on your machine too? It passes on my machine but fails in CI.

It runs successfully on my machine. I tested it with Xcode 26 RC and iPhone 17 Pro iOS 26 simulator.

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests.swift
philippsuess: Ok, thanks for the explanation. Doing that with a UI test would be ideal, but that isn't a bit hard to do with our current testing setup and would be too much work.

Can we maybe add a public view class to the subview instead of the `CameraUI.ChromeSwiftUIView`? So basically, trying to redact the views of the `UIImagePickerController`, which uses the `CameraUI.ChromeSwiftUIView` internally? The advantage would be that we don't test the implementation details, but are closer to the problem we're trying to solve. When the internals of the `UIImagePickerController` change and it stops using the `CameraUI.ChromeSwiftUIView` or the name of `CameraUI.ChromeSwiftUIView` is changed, we would stop redacting it and possibly crash again. 

Ah, but wait, we have the UI test that ensures we're not suddenly crashing. And it could be that if they change the `UIImagePickerController` to use something different, we don't need to redact anymore because it doesn't crash anymore.

Anyway, you have all the info I had in my head. When you think your testing strategy is good enough, please ignore me.

Accessing private classes always raises an alarm bell for me, because apps got rejected in the past, because we did this. As we're doing that only in test code, it's OK.

philipphofmann (approved):
LGTM

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests.swift
philprime: > Can we maybe add a public view class to the subview instead of the `CameraUI.ChromeSwiftUIView`? So basically, trying to redact the views of the UIImagePickerController, which uses the CameraUI.ChromeSwiftUIView internally? The advantage would be that we don't test the implementation details, but are closer to the problem we're trying to solve. 

I agree we shouldn't test implementation details, but then the only real use case would be setting up an sample view similar to the `UIImagePickerController` in production, which is a couple of nested views with navigation bars etc. which would require even more hacky inits of private types.

As this test is also used to verify the current behaviour and we really need to get the fix out, I'll keep it as-is.

> When the internals of the UIImagePickerController change and it stops using the CameraUI.ChromeSwiftUIView or the name of CameraUI.ChromeSwiftUIView is changed, we would stop redacting it and possibly crash again.

If the internals are changed and cause crashes again, we'll have to investigate it in general.

> Ah, but wait, we have the UI test that ensures we're not suddenly crashing. And it could be that if they change the UIImagePickerController to use something different, we don't need to redact anymore because it doesn't crash anymore.

We can't detect it if the `UIImagePickerController` isn't using the `CameraUI.ChromeSwiftUIView` unless I create an actual UI test, display the camera view (which is absent in simulators), then somehow extract the view hierarchy and assert it.

If they stop using the `ChromeSwiftUIView` we basically just have unused code in SDK, but that's not a massive issue.

> Accessing private classes always raises an alarm bell for me, because apps got rejected in the past, because we did this. As we're doing that only in test code, it's OK.

I 100% agree with you, but as you mentioned it's only tests.

PR #6047, opened 2025-09-02
ref: Convert `SentryCrashWrapper` to Swift

itaybre:
#skip-changelog


Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -0,0 +1,262 @@
...
+        // This would be iOS. Set it to macOS instead.
+        return "macOS"
+#else
+        return family
+#endif // targetEnvironment(macCatalyst)
+    }
+    
+    private func setScreenDimensions(_ deviceData: inout [String: Any]) {
+        // The UIWindowScene is unavailable on visionOS
+#if (os(iOS) || os(tvOS) || (swift(>=5.9) && os(visionOS))) && !SENTRY_NO_UIKIT && !(swift(>=5.9) && os(visionOS))
```
noahsmartin: You did the vision OS check twice here, I think you can remove them both and just say iOS || tvOS

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -0,0 +1,262 @@
...
+        }
+        
+        if kerr == KERN_SUCCESS {
+            return info.internal + info.compressed
+        }
+        
+        return 0
+    }
+    
+    @objc public func enrichScope(_ scope: Scope) {
```
philippsuess: `h`: If I'm not mistaken, the test coverage for all the scope-enriching logic is very limited. I don't feel comfortable changing this without adding tests first for the existing implementation and then using these tests to avoid regressions. Or how can we be sure we don't introduce any new bugs?

philipphofmann (commented):
I'm a bit afraid that we introduce new bugs with this refactoring cause the test coverage for the SentryCrashWrapper is quite limited.

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
itaybre: There is an integration test here: SentryCrashIntegrationTests
Not the same, but I will add individual UT

philippsuess: SentryCrashIntegrationTests covers some parts of it yes, but I think not all.

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+#if SENTRY_TEST || SENTRY_TEST_CI
+    // This var and initializer are used to inject system info during tests
+    var testSystemInfo: [String: Any]?
+    public init(processInfoWrapper: SentryProcessInfoSource, systemInfo: [String: Any]) {
+        self.processInfoWrapper = processInfoWrapper
+        testSystemInfo = systemInfo
+    }
+#endif // SENTRY_TEST && SENTRY_TEST_CI
```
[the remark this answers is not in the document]
itaybre: This is intentional

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
itaybre: I added more tests, let me know if you think they are not confident on those

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -0,0 +1,291 @@
...
+#if (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT
+        if let appWindows = SentryDependencyContainerSwiftHelper.windows(),
+           let appScreen = appWindows.first?.screen {
+            deviceData["screen_height_pixels"] = appScreen.bounds.size.height
+            deviceData["screen_width_pixels"] = appScreen.bounds.size.width
+        }
+#endif // (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT
+    }
+    
+    // MARK: - Private Methods
```
philippsuess: `l`: That needs to move up before ` private func enrichScopeWithOSData(_ scope: Scope, systemInfo: [String: Any])`.

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -0,0 +1,291 @@
...
+        // Static singleton pattern for systemInfo
+        struct Holder {
+            static let sharedInfo: [String: Any] = {
+                guard let systemInfo = SentryDependencyContainerSwiftHelper.crashReporter().systemInfo as? [String: Any] else {
+                    return [:]
+                }
+                return systemInfo
+            }()
+        }
+        return Holder.sharedInfo
```
philippsuess: `m`: Instead of the weird lazy singleton pattern, we could simply initialize the system info in the init method already, because we usually always use this in `enrichScope` which gets called during SDK start anyways. We could simplify this.

Tests/SentryTests/SentryCrash/SentryCrashWrapperTests.swift
```diff
@@ -0,0 +1,125 @@
...
+    override func tearDown() {
+        crashWrapper = nil
+        scope = nil
+        super.tearDown()
+    }
+    
+    func testEnrichScope_WithSystemInfo_SetsOSContext() {
+        crashWrapper.enrichScope(scope)
+        
+        let osContext = scope.contextDictionary["os"] as? [String: Any]
```
philippsuess: `m`: We can use XCTUnwrap here so we don't have to assert it's not nil below and we can ditch all the `?`. This applies to all tests.

Tests/SentryTests/SentryCrash/SentryCrashWrapperTests.swift
```diff
@@ -0,0 +1,125 @@
...
+        XCTAssertNotNil(osContext)
+        XCTAssertNotNil(osContext?["name"])
+        XCTAssertNotNil(osContext?["version"])
```
philippsuess: `h`: This returns an empty string, because the `SentryUIDeviceWrapper.getSystemVersion()` returns an empty string. It would be great to do proper assertions here and not simply assert it's not `nil`. Either the test setup is wrong or we have a bug.

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -0,0 +1,291 @@
...
+#elseif os(iOS)
+        return "iOS"
+#elseif os(tvOS)
+        return "tvOS"
+#elseif os(watchOS)
+        return "watchOS"
+#elseif (swift(>=5.9) && os(visionOS))
+        return "visionOS"
+#else
+        return "Unknown"
```
philippsuess: `m`: Unknown is a new functionality. Previously, we didn't set the OS if we couldn't detect it. Ist that by intention?

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -0,0 +1,291 @@
...
+    
+    private func getOSVersion() -> String {
+        // For MacCatalyst the UIDevice returns the current version of MacCatalyst and not the
+        // macOSVersion. Therefore we have to use NSProcessInfo.
+#if (os(iOS) || os(tvOS) || (swift(>=5.9) && os(visionOS))) && !SENTRY_NO_UIKIT && !targetEnvironment(macCatalyst)
+        return Dependencies.uiDeviceWrapper.getSystemVersion()
+#else
+        let version = ProcessInfo.processInfo.operatingSystemVersion
+        return "\(version.majorVersion).\(version.minorVersion).\(version.patchVersion)"
+#endif // (os(iOS) || os(tvOS) || (swift(>=5.9) && os(visionOS))) && !SENTRY_NO_UIKIT && !targetEnvironment(macCatalyst)
```
philippsuess, line 246: `h`: The `#if` condition here changed. Previously, we had only used the UIDeviceWrapper if `SENTRY_HAS_UIKIT && !TARGET_OS_MACCATALYST`. What drove you to change this?

philipphofmann (commented):
I maybe found a few issues and a bug. I think we must capture messages before this refactoring on different platforms and compare the event context properties to ensure we didn't mess anything up about the enrichScope method. Otherwise, I'm afraid we introduce bugs.

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
itaybre: Good eye

Tests/SentryTests/SentryCrash/SentryCrashWrapperTests.swift
itaybre: When running this tests only , it seems like it was empty, fixing that

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
itaybre: Yes, I can roll back to the previous behavior

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
itaybre: `SENTRY_HAS_UIKIT` is a macro only available to Objective C.
This is the expansion for that:
```
// SENTRY_UIKIT_AVAILABLE basically means: are we on a platform where we can link UIKit?
#if TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_VISION
#    define SENTRY_UIKIT_AVAILABLE 1
#else
#    define SENTRY_UIKIT_AVAILABLE 0
#endif

// SENTRY_HAS_UIKIT means we're on a platform that can link UIKit and we're building a configuration
// that will allow it to be autolinked. SENTRY_NO_UIKIT is set in GCC_PREPROCESSOR_DEFINITIONS
// for configurations that we will not allow to link UIKit by setting CLANG_MODULES_AUTOLINK to NO.
#if SENTRY_UIKIT_AVAILABLE && !SENTRY_NO_UIKIT
#    define SENTRY_HAS_UIKIT 1
#else
#    define SENTRY_HAS_UIKIT 0
#endif
```

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -0,0 +1,289 @@
...
+        guard let systemInfo = SentryDependencyContainerSwiftHelper.crashReporter().systemInfo as? [String: Any] else {
+            return [:]
+        }
+        return systemInfo
+    }()
+    
+    @objc
+    public init(processInfoWrapper: SentryProcessInfoSource) {
+        self.processInfoWrapper = processInfoWrapper
+        super.init()
```
philippsuess, line 52: `l`: calling super.init after assigning properties looks a bit weird.
```suggestion
        super.init()
        self.processInfoWrapper = processInfoWrapper
```

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -0,0 +1,289 @@
...
+#else
+@objc @_spi(Private)
+public final class SentryCrashWrapper: NSObject {
+    let processInfoWrapper: SentryProcessInfoSource
+    private static var systemInfo: [String: Any] = {
+        guard let systemInfo = SentryDependencyContainerSwiftHelper.crashReporter().systemInfo as? [String: Any] else {
+            return [:]
+        }
+        return systemInfo
+    }()
```
philippsuess: `m`: Thanks for moving this up here. I think it would be better to make this non static to remove global state and initialize the `systemInfo` in the init method. It's also better for testability. Ideally we would even inject the `crashReporter` and retrieve the system info from it. Then we don't need the `SentryDependencyContainerSwiftHelper` here.

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
philippsuess: Ups sorry, you have to initialize the property before calling super 🤦 . Ignore me

itaybre: You can't call super.init before initializing all properties

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
itaybre: I don't expect systemInfo to change values within a session, also right now we can't inject crashReporter because it is written in ObjC: `Cannot use class 'SentryCrash' here; '_SentryPrivate' has been imported as implementation-only`
That will be a TODO for later

itaybre:
macOS:
|Before|After|
|-----|-----|
|image|image|

iOS:
|Before|After|
|-----|-----|
| image |image|

PR #6053, opened 2025-09-03
chore(ci): Replace latest OS with deterministic versions

philprime:
This pull request is an attempt to make our CI more deterministic, as the the tag `latest` sometimes resolves to different versions, causing flakiness.

Furthermore we want to ensure that the selected simulator runtime is available and due to constant changes of the runner image by the GitHub team we need to lock it down.

#skip-changelog

### DIFF
--- a/scripts/ci-boot-simulator.sh
+++ b/scripts/ci-boot-simulator.sh
@@ -11,6 +11,10 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


philprime:
@cursor review

philipphofmann (approved):
Thanks LGTM

PR #6072, opened 2025-09-04
ref: Convert SentryProcessInfoWrapper to Swift

noahsmartin:
#skip-changelog


Sources/Swift/Helper/SentryProcessInfo.swift
```diff
@@ -0,0 +1,23 @@
+@_spi(Private) @objc public protocol SentryProcessInfo {
```
philprime: `m`: While I understand that the naming originates from `ProcessInfo` and `NSProcessInfo`, I would like to propose that we use a naming scheme where the protocol indicates that the implementation is the provider/source of information, e.g. `SentryProcessInfoSource` (provider might be unsuitable due to dependency injection).

This is not a strong request, so I won't block this. I just want to point out that it's not clear by our naming if a type is a logical type or a data structure.

philprime (commented):
LGTM, left one comment to discuss.

Sources/Swift/Helper/SentryProcessInfo.swift
noahsmartin: No preference from me so I'll change it since I have to rebase for the project file conflicts anyways

itaybre (approved):

Issue #6059, opened 2025-09-03
Make SentryDsn.getHash internal


philipphofmann:
### Description

There is no strong reason why this method needs to be public. We can make it internal. Also see [https://github.com/getsentry/sentry-cocoa/discussions/2946#discussioncomment-13867959](https://github.com/getsentry/sentry-cocoa/discussions/2946#discussioncomment-13867959).

In SentryDsn.h the function getHash is declared. This makes it part of the SDKs public API, but I think it doesn't need to be a public facing ability, it can be made internally only

philippsuess:
Will be fixed with https://github.com/getsentry/sentry-cocoa/pull/6080.

PR #6080, opened 2025-09-04
feat(V9): Private functions in SentryDsn

noahsmartin:
#skip-changelog

Fixes https://github.com/getsentry/sentry-cocoa/issues/6059

### DIFF
--- a/CHANGELOG-v9.md
+++ b/CHANGELOG-v9.md
@@ -16,6 +16,7 @@ Removes deprecated `setExtraValue` from SentrySpan (#5864)
 Removes `integrations` property from `SentryOptions` (#5749)
 Makes `SentryEventDecodable` internal (#5808)
 The `span` property on `SentryScope` is now readonly (#5866)
+Removes `getHash` function on `SentryDsn` (#6080)
 
 ### Fixes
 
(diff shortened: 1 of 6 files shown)


philippsuess:
We use the getHash method in two different places in our sample apps to determine the folder for the stored envelopes

https://github.com/getsentry/sentry-cocoa/blob/c174e5eea476f2c659ba06d86c3d4deb68640831/Samples/iOS-Swift/iOS-Swift/ExtraViewController.swift#L318-L331

https://github.com/getsentry/sentry-cocoa/blob/c174e5eea476f2c659ba06d86c3d4deb68640831/TestSamples/SwiftUITestSample/SwiftUITestSample/SwiftUITestSampleApp.swift#L70-L79

I can think of three different options right now: 

1. The option `cacheDirectoryPath` is public. We can determine the envelope path without knowing the actual path. The envelopes are be stored in `"\(sentryOptions.cacheDirectoryPath)/io.sentry/\(dsnHash)/envelopes"`. Usually, there should be only one dsnHash, so we could find the stored envelopes even without the dsnHash. We can use the `cacheDirectoryPath` for tests, so we are sure we don't have multiple DSNS in the same `cacheDirectoryPath`.

2. We could add a property to expose the envelopes' path, which is not ideal, because our users shouldn't even know that envelopes exist. Once we add it, people might use it. Yes, they can also get to the envelope path as pointed out in option 1, but that's hacky and you have to know internal parts of the SDK.

3. As we only use the getHash method for retrieving the envelopes' path for testing, we could also change our tests so we don't need it anymore, but that's some work.

I think option 1 is the easiest one. Another option is to simply delay this until v 10.0.0.

noahsmartin:
I considered forward declaring this function in the tests, but I think that's a bad practice and it would be better to do something like convert it to Swift and use @testable or @_spi, but I think those are a bit beyond the scope of what I wanted to do for this PR, so going to close it for now and we can revisit after V9

PR #6092, opened 2025-09-05
Test Job to validate visionOS integration job

itaybre:
This PR is expected to fail at the visionOS integration Job because it is missing the libc++ link.


philippsuess:
@itaybre if it's expected to fail, do we want to merge it? Shouldn't this be a draft PR?

itaybre:
> @itaybre if it's expected to fail, do we want to merge it? Shouldn't this be a draft PR?

my bad, it already confirmed my check, so I can close this

PR #6094, opened 2025-09-05
Convert `SentryScreenFrames` to Swift

### LINKED ISSUE #6119: Convert `SentryScreenFrames` to Swift
> [!NOTE]
> The pull request "[Convert `SentryScreenFrames` to Swift](https://github.com/getsentry/sentry-cocoa/pull/6094)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

#skip-changelog

itaybre:
#skip-changelog

Closes #6119


itaybre:
Waiting for RN integration fix here: https://github.com/getsentry/sentry-react-native/pull/5153

.github/workflows/test-cross-platform.yml
```diff
@@ -55,6 +55,7 @@ jobs:
         with:
           repository: getsentry/sentry-react-native
           path: sentry-react-native
+          ref: itay/fix_sentreyscreenframes
```
philprime, line 58: `h`: I believe this is a left-over, so please remove it.

Sources/Sentry/Profiling/SentryProfilingScreenFramesHelper.h
```diff
@@ -0,0 +1,17 @@
+#import "SentryDefines.h"
+#import <Foundation/Foundation.h>
+
+#if SENTRY_HAS_UIKIT
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class SentryScreenFrames;
+
+// Helper to use SentryScreenFrames without importing Swift on ObjC++ files.
```
philprime, line 10: `m`: Please add more documentation or reference existing documentation what exactly this means and why it is necessary. Knowledge of contributors on swift-objc++ interop could be limited

Sources/Swift/Core/Integrations/FramesTracking/SentryScreenFrames.swift
```diff
@@ -0,0 +1,182 @@
...
+        self.total = total
+        self.frozen = frozen
+        self.slow = slow
+        self.slowFrameTimestamps = slowFrameTimestamps
+        self.frozenFrameTimestamps = frozenFrameTimestamps
+        self.frameRateTimestamps = frameRateTimestamps
+        
+        super.init()
+    #else
+        fatalError("SentryScreenFrames only works with UIKit enabled. Ensure you're using the right configuration of Sentry that links UIKit.")
```
philprime: `h`: Using `fatalError` is not the same as `SENTRY_GRACEFUL_FATAL` we used before. `fatalError` will always crash the host app, while the graceful fatal just crashed in debug. Consider [assertionFailure](https://developer.apple.com/documentation/swift/assertionfailure(_:file:line:)) instead

Tests/SentryTests/Swift/Profiling/SentryScreenFramesTests.swift
```diff
@@ -0,0 +1,359 @@
...
+        XCTAssertTrue(screenFrames.frameRateTimestamps.isEmpty)
+        #endif // os(iOS)
+    }
+    
+    func testInitWithZeroValues() {
+        let screenFrames = SentryScreenFrames(total: 0, frozen: 0, slow: 0)
+        
+        XCTAssertEqual(screenFrames.total, 0)
+        XCTAssertEqual(screenFrames.frozen, 0)
+        XCTAssertEqual(screenFrames.slow, 0)
```
philprime, line 29: `m`: No checks for the timestamp arrays?

.github/workflows/test-cross-platform.yml
itaybre: Not a leftover, since SentryScreenFrames is no longer available, it will break RN builds.
This branch contains the fix, which cannot be merged until we release a new Sentry version

Sources/Swift/Core/Integrations/FramesTracking/SentryScreenFrames.swift
```diff
@@ -0,0 +1,187 @@
+import Foundation
+
+#if (os(iOS) || os(tvOS) || (swift(>=5.9) && os(visionOS)))
+
+/// An array of dictionaries that each contain a start and end timestamp for a rendered frame.
+#if os(iOS)
```
philprime: `l`: I think the docs should only be added if the type is compiled.

```suggestion
#if os(iOS)
/// An array of dictionaries that each contain a start and end timestamp for a rendered frame.
```

Sources/Swift/Core/Integrations/FramesTracking/SentryScreenFrames.swift
```diff
@@ -0,0 +1,187 @@
...
+    
+    // MARK: - Initialization
+    
+    /// Creates a `SentryScreenFrames` instance with basic frame metrics.
+    /// - Parameters:
+    ///   - total: Total number of frames rendered
+    ///   - frozen: Number of frozen frames
+    ///   - slow: Number of slow frames
+    @objc public init(total: UInt, frozen: UInt, slow: UInt) {
+#if SENTRY_NO_UIKIT
```
philprime, line 59: `h`: In the Objective-C implementation we checked at the top level if UIKit is available, here we check in the initializer. This changes from a compile-time check to a runtime-check and could introduce problems for users, if they use this class without UIKit

itaybre: The top level macro `SENTRY_UIKIT_AVAILABLE` checks if the platform supports UIKIT (iOS / tvOS / visionOS)
The swift version for this is `#if (os(iOS) || os(tvOS) || (swift(>=5.9) && os(visionOS)))`

And then inside the init we also had `SENTRY_HAS_UIKIT` which just checked if we were building without UIKit on one of this platforms.

philprime (approved):
LGTM

PR #6095, opened 2025-09-05
ref: Convert SentryClientReport to Swift

noahsmartin:
#skip-changelog


Sources/Sentry/SentryHttpTransport.m
```diff
@@ -293,7 +292,9 @@ - (SentryEnvelope *)addClientReportTo:(SentryEnvelope *)envelope
         [self.discardedEvents removeAllObjects];
     }
 
-    SentryClientReport *clientReport = [[SentryClientReport alloc] initWithDiscardedEvents:events];
+    SentryClientReport *clientReport = [[SentryClientReport alloc]
+        initWithDiscardedEvents:events
+                   dateProvider:SentryDependencyContainer.sharedInstance.dateProvider];
```
philippsuess: `m`: We already have a property for the date provider. No need to use the `SentryDependencyContainer` here.
```suggestion
                   dateProvider:self.dateProvider];
```

Sources/Swift/Tools/SentryClientReport.swift
```diff
@@ -0,0 +1,25 @@
...
+            "timestamp": NSNumber(value: self.timestamp.timeIntervalSince1970),
+            "discarded_events": events
+        ]
+    }
+    
+    /**
+     * The timestamp of when the client report was created.
+     */
+    private let timestamp: Date
+    let discardedEvents: [SentryDiscardedEvent]
```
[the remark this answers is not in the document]
philippsuess: `m`: If nobody uses the timestamp we can ignore this. Maybe we could even make the `discardedEvents` private as well.

Sources/Swift/Tools/SentryClientReport.swift
```diff
@@ -0,0 +1,25 @@
+@objc
+@_spi(Private) public final class SentryClientReport: NSObject, SentrySerializable {
+    
+    @objc(initWithDiscardedEvents:dateProvider:)
+    public init(discardedEvents: [SentryDiscardedEvent], dateProvider: SentryCurrentDateProvider) {
+        timestamp = dateProvider.date()
```
philippsuess, line 6: `l`: So we consistently use self in the init method.
```suggestion
        self.timestamp = dateProvider.date()
```

Sources/Swift/Tools/SentryClientReport.swift
```diff
@@ -0,0 +1,25 @@
+@objc
+@_spi(Private) public final class SentryClientReport: NSObject, SentrySerializable {
+    
+    @objc(initWithDiscardedEvents:dateProvider:)
+    public init(discardedEvents: [SentryDiscardedEvent], dateProvider: SentryCurrentDateProvider) {
```
philippsuess, line 5: `l`: Maybe we could even remove the dependency to the `SentryCurrentDateProvider` here and only ask for the current date, which has the downside that everybody now needs to add `.date()` in the tests, and the HTTPTransport. I doesn't really matter cause both are OK.

philipphofmann (approved):
Thanks with a few comments.

Sources/Swift/Tools/SentryClientReport.swift
noahsmartin: I thought about this as well but figured since the goal of this change was just to convert it to Swift I'd start with this to make minimal changes everywhere else

Sources/Swift/Tools/SentryClientReport.swift
noahsmartin: I'm used to the airbnb swift style guide here: https://github.com/airbnb/swift?tab=readme-ov-file#omit-self

No strong preference other than that if we have a preference we should make it automatically format that way (either auto add or auto remove, or no preference)

noahsmartin:
Github wouldn't let me directly comment but re: `m: If nobody uses the timestamp we can ignore this. Maybe we could even make the discardedEvents private as well.`

I tried making discardedEvents private as well but unfortunately there is some test code that accesses it

Sources/Swift/Tools/SentryClientReport.swift
philippsuess: Yep, keep it as is then.

Sources/Swift/Tools/SentryClientReport.swift
philippsuess: I care mostly about consistency. You can add a linter and enforce omitting self. I would be fine with that.

PR #6097, opened 2025-09-08
Fix Changelog: Move PR 5800 to unreleased

philipphofmann:
We merged #5800 after releasing 8.55.0. We need to move it up to the unreleased section.

Screenshot 2025-09-08 at 08 56 34

#skip-changelog


philprime (approved):
LGTM

PR #6125, opened 2025-09-10
fix: Fix stripped categories in static frameworks

noahsmartin:
This fixes a crash when using Sentry as a static framework. The linker will try to remove these @objc extensions in static frameworks because they are in a file with no other used classes. To avoid this we need to put something else in these files that get referenced from another part of the codebase. This isn't a problem for Swift, but is for objc that uses `performSelector`

This made me realize we don't have a CI test for sentry being used as a static framework. We should add one, but it's a bit tricky to do without SPM, that's why this is still a WIP

#skip-changelog


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Fix crash in Session Replay when opening the camera UI on iOS 26+ by skipping redaction of internal views.
   This may result in more of the camera screen being redacted. (#6045)
+- Fix crash when using the SDK as a static framework (#6125)
```
philippsuess: `m`: I would be a bit more specific, which crash this fixes, so everybody knows this by just looking at the changelog.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -144,6 +144,10 @@ - (instancetype)init
     if (self = [super init]) {
         isInitialializingDependencyContainer = YES;
 
+        // These classes must be referenced somewhere for their files to not be stripped.
+        id _ = [[PlaceholderSentryApplication alloc] init];
+        _ = [[PlaceholderProcessInfoClass alloc] init];
```
philippsuess: `m`: We already have similar code here https://github.com/getsentry/sentry-cocoa/blob/dba2de8a2edfdd5ee8d08d5937baac59b8036727/Sources/Sentry/SentrySDKInternal.m#L253-L256

I think we should add it there.

philippsuess:
> This made me realize we don't have a CI test for sentry being used as a static framework. We should add one, but it's a bit tricky to do without SPM, that's why this is still a WIP

Yes indeed.

Sources/Swift/Helper/SentryApplicationExtensions.swift
```diff
@@ -1,3 +1,9 @@
+// This is needed because a file that only contains an @objc extension will get automatically stripped out
```
philippsuess, line 1: Thanks for the comment.

noahsmartin:
Thanks @philipphofmann - Since it's only in an alpha and reported by just 1 user I'm thinking it would add more confusion that it solves to put it in the troubleshooting

PR #6131, opened 2025-09-10
fix: Add null checks for envelope item data in session migration and serialization

philprime:
See #5577

* Updated `SentryMigrateSessionInit.m` to check for nil `item.data` before deserialization in both `migrateSessionInit` and `setInitFlagIfContainsSameSessionId`, logging a warning and skipping items with no data to prevent crashes. [[1]](diffhunk://#diff-6f3027a312ac25d26cc2c67b1a6094d78ffcd75e380c4143023b7e8959457bc3L21-R28) [[2]](diffhunk://#diff-6f3027a312ac25d26cc2c67b1a6094d78ffcd75e380c4143023b7e8959457bc3L75-R88)
* Modified `SentrySerialization.m` to return nil and log an error if any envelope item has nil data during serialization, ensuring envelopes are not serialized with missing data.
* Replaced the old Objective-C test file `SentryMigrateSessionInitTests.m` with a new, comprehensive Swift test file `SentryMigrateSessionInitTests.swift`, adding multiple tests to validate behavior with nil and mixed data in envelopes, and confirming correct migration logic. [[1]](diffhunk://#diff-541c5b36da9e990aaa1d7b1c07d4f4f993b0f53ae17e983e6ceb47ba1664fccbL1-L22) [[2]](diffhunk://#diff-68928cc09503dae0631c1f7c76d7817dfb4643b38187890352bdc1c77f40f9b3R1-R153)
* Added new tests to `SentrySerializationTests.swift` to ensure serialization returns nil when envelope items have nil data, covering both single and mixed cases.

#skip-changelog


Sources/Sentry/SentryMigrateSessionInit.m
```diff
@@ -18,7 +19,13 @@ + (BOOL)migrateSessionInit:(SentryEnvelope *)envelope
 
     for (SentryEnvelopeItem *item in envelope.items) {
         if ([item.header.type isEqualToString:SentryEnvelopeItemTypes.session]) {
-            SentrySession *session = [SentrySerialization sessionWithData:item.data];
+            if (nil == item.data) {
```
philippsuess, line 22: `l`: I think the yoda style checks a bit confusing.
```suggestion
            if (item.data == nil) {
```

We even have a linter for this in Swift https://github.com/getsentry/sentry-cocoa/blob/2372f11029ab027ad82b80af0aa583aa8ccb68a1/.swiftlint.yml#L79

Tests/SentryTests/Helper/SentryMigrateSessionInitTests.swift
```diff
@@ -0,0 +1,153 @@
...
+    
+    override func setUp() {
+        super.setUp()
+        clearTestState()
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        clearTestState()
+    }
```
philippsuess, line 15: `m`: Do we really need to call `clearTestState` for all the tests? `clearTestState` has some side effects and it's good avoid these and extra work.

Tests/SentryTests/Helper/SentryMigrateSessionInitTests.swift
```diff
@@ -0,0 +1,153 @@
...
+                                                       length: UInt(testSessionData.count))
+        let testSessionItem = SentryEnvelopeItem(header: testSessionHeader, data: testSessionData)
+        let testEnvelope = SentryEnvelope(header: SentryEnvelopeHeader(id: nil), singleItem: testSessionItem)
+        let testEnvelopeData = try XCTUnwrap(SentrySerialization.data(with: testEnvelope))
+        
+        try! testEnvelopeData.write(to: URL(fileURLWithPath: tempPath))
+        
+        defer {
+            try? FileManager.default.removeItem(atPath: tempPath)
+        }
```
philippsuess, line 132: `m`: I would move that up a bit so it's directly below to ensure we always remove the tempPath.
```
let tempPath = (tempDir as NSString).appendingPathComponent(tempFile)
```

Tests/SentryTests/Helper/SentryMigrateSessionInitTests.swift
```diff
@@ -0,0 +1,153 @@
...
+        let sessionHeader = SentryEnvelopeItemHeader(type: SentryEnvelopeItemTypes.session, 
+                                                   length: UInt(sessionData.count))
+        let sessionItem = SentryEnvelopeItem(header: sessionHeader, data: sessionData)
+        let envelope = SentryEnvelope(header: SentryEnvelopeHeader(id: nil), singleItem: sessionItem)
+        
+        let tempDir = NSTemporaryDirectory()
+        let tempFile = "test_envelope.json"
+        let tempPath = (tempDir as NSString).appendingPathComponent(tempFile)
+        
+        // Create a test envelope file with the same session ID but without init flag
```
philippsuess, line 118: `l`: That comment doesn't make sense. I think we can remove it.
```suggestion
```

Tests/SentryTests/Helper/SentryMigrateSessionInitTests.swift
```diff
@@ -0,0 +1,153 @@
...
+        let testSession = SentrySession(releaseName: "1.0.0", distinctId: "test-id")
+        // Don't set init flag on test session
+        let testSessionData = try XCTUnwrap(SentrySerialization.data(with: testSession))
+        let testSessionHeader = SentryEnvelopeItemHeader(type: SentryEnvelopeItemTypes.session, 
+                                                       length: UInt(testSessionData.count))
+        let testSessionItem = SentryEnvelopeItem(header: testSessionHeader, data: testSessionData)
+        let testEnvelope = SentryEnvelope(header: SentryEnvelopeHeader(id: nil), singleItem: testSessionItem)
+        let testEnvelopeData = try XCTUnwrap(SentrySerialization.data(with: testEnvelope))
+        
+        try! testEnvelopeData.write(to: URL(fileURLWithPath: tempPath))
```
philippsuess, line 128: `l`: I'm a bit confused with the naming of the variables here. What drove you to prefix this with test? It refers to the same session as above, but the name doesn't reflect that. It's quite confusing.

Tests/SentryTests/Helper/SentryMigrateSessionInitTests.swift
```diff
@@ -0,0 +1,153 @@
...
+        let result = SentryMigrateSessionInit.setInitFlagIfContainsSameSessionId(session.sessionId, 
+                                                                                envelope: envelope, 
+                                                                                envelopeFilePath: "/tmp/test")
+        
+        // Assert
+        XCTAssertTrue(result, "Should return true when matching session is found and init flag is set")
+    }
+
+    @available(*, deprecated, message: "This is only marked as deprecated because enableAppLaunchProfiling is marked as deprecated. Once that is removed this can be removed.")
+    func testMigrateSessionInit_WithValidSessionInit_ReturnsTrue() throws {
```
philippsuess, line 104: `m`: I don't really understand why we need this test. The `testMigrateSessionInit**` tests in the SentryFileManager cover this use case already. What drove you to add this tests?

philipphofmann (commented):
CI is completely broken. Something is off here.

philprime:
Closing in favor of #6162

Issue #6129, opened 2025-09-10
`_SentryDispatchQueueWrapperInternal.dispatchAsyncOnMainQueue` may execute in a synchronous way


itaybre:
### Description

A current bug in our logic might execute immediately and synchronously if we call `_SentryDispatchQueueWrapperInternal.dispatchAsyncOnMainQueue` from the main thread

PR #6132, opened 2025-09-10
chore: `dispatchAsyncOnMainQueue` to `dispatchAsyncOnMainQueueIfNotMainThread`

itaybre:
Fixes: #6129

When `dispatchAsyncOnMainQueue` is called from the main thread, the block ends up being executed immediately instead of async. I am renaming the method to make the behavior clearer

#skip-changelog


itaybre:
Looks like there are several places relying on the previous behavior, I will rename the method and create a new one then

philipphofmann (approved):
LGTM, merge it quickly please to avoid conflicts 🚀

PR #6143, opened 2025-09-11
refactor: Migrate SentryMsgPackSerializer from Objective-C to Swift

### LINKED ISSUE #6140: Fix nullability in SentryMsgPackSerializer.m

philprime:
## Summary

This PR migrates the SentryMsgPackSerializer from Objective-C to Swift while maintaining 100% behavioral compatibility.

### Complete Migration
- Converted `SentryMsgPackSerializer.m/.h` to modular Swift implementation 
- Reorganized code into separate files in `Sources/Swift/Tools/MsgPack/` directory:
  - `SentryMsgPackSerializer.swift` - Main serialization logic
  - `SentryStreamable.swift` - Protocol definition
  - `SentryMsgPackSerializerError.swift` - Error types
  - `Data+SentryStreamable.swift` - Data extension
  - `URL+SentryStreamable.swift` - URL extension  
  - `NSData+SentryStreamable.swift` - NSData extension
  - `NSURL+SentryStreamable.swift` - NSURL extension

### Behavioral Compatibility
- Maintained exact behavior matching original Objective-C implementation
- Preserved edge cases like silent key length truncation for >255 byte keys
- Kept error handling patterns including -1 return values for file size errors
- Maintained logging levels (error vs debug) matching original code

### Enhanced Test Coverage
- Added comprehensive test suite with 14 test cases covering all code paths
- Added edge case testing for large dictionaries, long keys, stream errors
- Improved error validation for nil input streams and invalid file paths
- AAA test pattern with proper setup/teardown and temp file cleanup

### Implementation Improvements
- Modern Swift patterns with proper error throwing instead of boolean returns
- Type safety with explicit error types via `SentryMsgPackSerializerError`
- Memory safety improvements while maintaining compatibility
- Cleaner byte operations using modern Swift APIs

### **Key Technical Details**
- Protocol signature: Uses `streamSize() -> Int` to support -1 error values from Objective-C
- Truncating conversion: Uses `UInt8(truncatingIfNeeded:)` to match Objective-C silent truncation
- Error propagation: Swift errors are caught and converted to boolean returns for Objective-C compatibility
- File I/O: Improved path validation through Swift's `Data.write(to:)` method

## Testing

All existing functionality verified through comprehensive test suite:
- 14 tests covering all code paths and edge cases
- 100% backward compatibility with existing behavior
- Proper error handling for all failure scenarios

Closes #6140

#skip-changelog

### DIFF
--- a/.github/workflows/build-xcframework-variant-slices.yml
+++ b/.github/workflows/build-xcframework-variant-slices.yml
@@ -56,9 +56,9 @@ jobs:
   build-xcframework-variant-slices:
     name: ${{matrix.sdk}}
 
-    # We must compile this on an arm64 runner, cause it's required for visionOS. macos-14 uses arm64.
+    # We must compile this on an arm64 runner, cause it's required for visionOS. macos-15 uses arm64.
     # To see the available runners see https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories.
-    runs-on: macos-14
+    runs-on: macos-15
 
     strategy:
       matrix:
@@ -67,12 +67,8 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      # We have to compile on Xcode 15.2 because compiling on Xcode 15.4 fails with
-      # Data+SentryTracing.swift:21:62: error: 'ReadingOptions' aliases 'Foundation.ReadingOptions'
-      # and cannot be used here because C++ types from imported module 'Foundation' do not support
-      # library evolution; this is an error in the Swift 6 language mode
-      # We also can't use Xcode 16.x because validating the XCFramework then fails with Xcode 15.x.
-      - run: ./scripts/ci-select-xcode.sh 15.2
+      # We use Xcode 16.4 for building XCFrameworks as it's the standard version available on macos-15 runners.
+      - run: ./scripts/ci-select-xcode.sh 16.4
         shell: bash
 
       - name: Get version
@@ -109,7 +105,7 @@ jobs:
         env:
           VERSION: ${{ env.VERSION }}
         run: |
-          ./scripts/ci-select-xcode.sh 15.2
+          ./scripts/ci-select-xcode.sh 16.4
           make bump-version TO=$VERSION
 
       - name: Build ${{inputs.name}}${{inputs.suffix}} XCFramework slice for ${{matrix.sdk}}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -106,7 +106,7 @@ jobs:
 
   validate-xcframework:
     name: Validate XCFramework
-    runs-on: macos-14
+    runs-on: macos-15
     needs: [files-changed, assemble-xcframework-variant]
     # Run the job only for PRs with related changes or non-PR events.
     if: github.event_name != 'pull_request' || needs.files-changed.outputs.run_release_for_prs == 'true'
@@ -120,7 +120,7 @@ jobs:
         with:
           pattern: xcframework-${{github.sha}}-sentry-swiftui
           path: Carthage/
-      - run: ./scripts/ci-select-xcode.sh 15.4
+      - run: ./scripts/ci-select-xcode.sh 16.4
       - run: make build-xcframework-sample
       - name: Run CI Diagnostics
         if: failure()
@@ -131,7 +131,7 @@ jobs:
   # See https://github.community/t/github-sha-isnt-the-value-expected/17903/17906.
   validate-spm:
     name: Validate SPM Static
-    runs-on: macos-14
+    runs-on: macos-15
     needs: [files-changed, assemble-xcframework-variant]
     # Run the job only for PRs with related changes or non-PR events.
     if: github.event_name != 'pull_request' || needs.files-changed.outputs.run_release_for_prs == 'true'
@@ -145,6 +145,7 @@ jobs:
         uses: ./.github/actions/prepare-package.swift
         with:
           is-pr: ${{ github.event_name == 'pull_request' }}
+      - run: ./scripts/ci-select-xcode.sh 16.4
       - run: swift build
         working-directory: Samples/macOS-SPM-CommandLine
       - name: Run CI Diagnostics
@@ -153,7 +154,7 @@ jobs:
 
   validate-spm-dynamic:
     name: Validate SPM Dynamic
-    runs-on: macos-14
+    runs-on: macos-15
     needs: [files-changed, assemble-xcframework-variant]
     # Run the job only for PRs with related changes or non-PR events.
     if: github.event_name != 'pull_request' || needs.files-changed.outputs.run_release_for_prs == 'true'
@@ -167,6 +168,7 @@ jobs:
         uses: ./.github/actions/prepare-package.swift
         with:
           is-pr: ${{ github.event_name == 'pull_request' }}
+      - run: ./scripts/ci-select-xcode.sh 16.4
       - run: swift build
         working-directory: Samples/SPM-Dynamic
       - name: Run CI Diagnostics
@@ -175,7 +177,7 @@ jobs:
 
   swift-build:
     name: Build Swift Static
-    runs-on: macos-14
+    runs-on: macos-15
     needs: [files-changed, assemble-xcframework-variant]
     # Run the job only for PRs with related changes or non-PR events.
     if: github.event_name != 'pull_request' || needs.files-changed.outputs.run_release_for_prs == 'true'
@@ -189,6 +191,7 @@ jobs:
         uses: ./.github/actions/prepare-package.swift
         with:
           is-pr: ${{ github.event_name == 'pull_request' }}
+      - run: ./scripts/ci-select-xcode.sh 16.4
       - run: swift build
       - name: Run CI Diagnostics
         if: failure()
@@ -196,7 +199,7 @@ jobs:
 
   validate-spm-visionos:
     name: Validate SPM Static visionOS
-    runs-on: macos-14
+    runs-on: macos-15
     needs: [files-changed, assemble-xcframework-variant]
     # Run the job only for PRs with related changes or non-PR events.
     if: github.event_name != 'pull_request' || needs.files-changed.outputs.run_release_for_prs == 'true'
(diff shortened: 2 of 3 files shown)


philprime:
@cursor review

philprime:
@cursor review
@seer review

Sources/Swift/Tools/MsgPack/NSData+SentryStreamable.swift
```diff
@@ -0,0 +1,9 @@
+extension NSData: SentryStreamable {
```
noahsmartin: I don't think this is necessary, NSData and Data are toll free bridged, the Data conformance should be enough, can we remove this file?

Sources/Swift/Tools/MsgPack/NSURL+SentryStreamable.swift
```diff
@@ -0,0 +1,27 @@
+extension NSURL: SentryStreamable {
```
noahsmartin: Can we make this conformance on `URL`?

Sources/Swift/Tools/MsgPack/NSData+SentryStreamable.swift
```diff
@@ -0,0 +1,9 @@
...
+    func asInputStream() -> InputStream? {
+        return InputStream(data: self as Data)
```
noahsmartin: Or if we can't remove this file for some reason, I think we can change the implementations to be like:

```
(self as Data).asInputStream()
```

Sources/Swift/Tools/MsgPack/NSURL+SentryStreamable.swift
```diff
@@ -0,0 +1,27 @@
...
+        } catch {
+            SentrySDKLog.error("Could not read file attributes - File: \(self) - \(error)")
+            return -1
+        }
+        guard let fileSize = attributes[.size] as? NSNumber else {
+            SentrySDKLog.error("Could not read file size attribute - File: \(self)")
+            return -1
+        }
+        let unsignedSize = fileSize.uintValue
+        // Handle potential overflow for very large files
```
noahsmartin: Should we change this function to return NSInteger or Uint to match the objc behavior for large file sizes?

Sources/Swift/Tools/MsgPack/NSURL+SentryStreamable.swift
noahsmartin: I see you already have a `URL` one in another file, as with the other case would be great to delete this file. At least I don't think we should have a separate implementation here, the function could be 

```
(self as URL).streamSize()
```

That would avoid all the code duplication

Sources/Swift/Tools/MsgPack/NSData+SentryStreamable.swift
philprime: Are we 100% certain this won't cause an infinite recursion?

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
```diff
@@ -0,0 +1,99 @@
...
+            return false
+        }
+    }
+
+    static func serializeDictionaryToMessagePack(_ dictionary: [String: Any]) throws -> Data { // swiftlint:disable:this function_body_length
+        let outputStream = OutputStream.toMemory()
+        outputStream.open()
+        defer { outputStream.close() }
+        
+        let mapHeader = UInt8(0x80 | dictionary.count) // Map up to 15 elements
```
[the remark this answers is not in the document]
noahsmartin: It looks like this is true? At least the first part about accumulating everything to memory. Any reason we can't use the ObjC behavior here of having the output stream write to a file?

Sources/Swift/Tools/MsgPack/NSData+SentryStreamable.swift
noahsmartin: I don't think it would, if my understanding is wrong though could always have something like a static method that takes the Data as a parameter

Sources/Swift/Tools/MsgPack/NSURL+SentryStreamable.swift
philprime: I was thinking that too, but I can't do that because we already returned `-1` before, which can't be represented as an UInt → I'll add a comment

philprime: Nevermind, it's only used in one spot so I'll change it to be a nullable unsigned integer and return null for invalid values

Sources/Swift/Tools/MsgPack/NSData+SentryStreamable.swift
philprime: I removed the NSData extensions, because the class is only used from Swift

philprime:
@sentry review

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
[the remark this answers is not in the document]
philprime: I'll double-check why I changed it, it's been a while

Sources/Swift/Tools/MsgPack/SentryStreamable.swift
```diff
@@ -0,0 +1,4 @@
+protocol SentryStreamable {
+    func asInputStream() -> InputStream?
+    func streamSize() -> UInt?
+}
```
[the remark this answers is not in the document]
philprime: It didn't break so not needed

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
```diff
@@ -0,0 +1,98 @@
...
+        _ = outputStream.write([mapHeader], maxLength: 1)
+
```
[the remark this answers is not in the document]
philprime: This is the same implementation as before

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
```diff
@@ -0,0 +1,98 @@
...
+            }
+            guard let keyData = key.data(using: .utf8) else {
+                throw SentryMsgPackSerializerError.invalidInput("Could not encode key as UTF-8: \(key)")
+            }
+            
+            let str8Header: UInt8 = 0xD9 // String up to 255 characters
+            let keyLength = UInt8(truncatingIfNeeded: keyData.count) // Truncates if > 255, matching Objective-C behavior
+            _ = outputStream.write([str8Header], maxLength: 1)
+            _ = outputStream.write([keyLength], maxLength: 1)
+
```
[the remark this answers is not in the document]
philprime: Won't do because this is a 1:1 conversion

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
```diff
@@ -0,0 +1,104 @@
+/**
+ * This is a partial implementation of the MessagePack format.
+ * We only need to concatenate a list of NSData into an envelope item.
+ */
+class SentryMsgPackSerializer {
+    @objc
```
noahsmartin: Since the class itself is not marked with @objc I guess this is only called from Swift? Can we remove this objc?

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
```diff
@@ -0,0 +1,104 @@
+/**
+ * This is a partial implementation of the MessagePack format.
+ * We only need to concatenate a list of NSData into an envelope item.
+ */
+class SentryMsgPackSerializer {
```
noahsmartin: `final`?

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
```diff
@@ -0,0 +1,104 @@
...
+    @objc
+    static func serializeDictionary(toMessagePack dictionary: [String: Any], intoFile fileURL: URL) -> Bool {
```
noahsmartin: Can we make this input type `[String: SentryStreamable]` I think we can since the function doesn't need to be exposed to objc.

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
```diff
@@ -0,0 +1,104 @@
...
+                try FileManager.default.removeItem(at: fileURL)
+            } catch {
+                // Ignore cleanup errors - file might not exist
+            }
+            return false
+        }
+    }
+    
+    // swiftlint:disable:next function_body_length cyclomatic_complexity
+    private static func serializeToFile(dictionary: [String: Any], fileURL: URL) throws {
```
noahsmartin: Assuming we change the above not to use Any, can change this as well

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
```diff
@@ -0,0 +1,104 @@
...
+        }
+        
+        let mapHeader = UInt8(truncatingIfNeeded: 0x80 | dictionary.count) // Map up to 15 elements
+        _ = outputStream.write([mapHeader], maxLength: 1)
+
+        for (key, anyValue) in dictionary {
+            guard let value = anyValue as? SentryStreamable else {
+                throw SentryMsgPackSerializerError.invalidValue("Value does not conform to SentryStreamable: \(anyValue)")
+            }
+            guard let keyData = key.data(using: .utf8) else {
```
noahsmartin: I think we can use `Data(key.utf8)` and then it won't be optional

Sources/Swift/Tools/MsgPack/SentryMsgPackSerializer.swift
```diff
@@ -0,0 +1,104 @@
...
+
+            let valueLength = UInt32(truncatingIfNeeded: dataLength)
+            // We will always use the 4 bytes data length for simplicity.
+            // Worst case we're losing 3 bytes.
+            let bin32Header: UInt8 = 0xC6
+            _ = outputStream.write([bin32Header], maxLength: 1)
+            
+            // Write UInt32 as big endian bytes
+            let lengthBytes = [
+                UInt8((valueLength >> 24) & 0xFF),
```
noahsmartin: Can we still use `NSSwapHostIntToBig` like the objc code did? That function should be available to Swift

philprime:
Closing this due to sunken cost fallacy. We should revisit this another time.

PR #6149, opened 2025-09-11
feat: Add SentryDistribution

### LINKED ISSUE #6173: feat: Add SentryDistribution
> [!NOTE]
> The pull request "[feat: Add SentryDistribution](https://github.com/getsentry/sentry-cocoa/pull/6149)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Adds the ETDistribution model to the package.swift

Intentionally not supporting cocoapods or usage from ObjC for now, that can come in a future version.

There are a few ways we could expose this API:

1. Using the `SentrySDK` naming so you would do something like `SentrySDK.checkForUpdates()`
  - a: As an extension from the existing `SentrySDK`. This is not preferred because it means users would have to depend on both the Sentry target, and the distribution target.
  - b: As a new target containing a new object with the same name. So there would be `Sentry.SentrySDK` and `SentryDistribution.SentrySDK` but most of the time you wouldn't need the module prefixes if they are being used in different files.
2. With a new type name like `Distribution`. This is what this PR does. The API is:
```
import SentryDistribution

Updater.checkForUpdates()
```

#skip-changelog

noahsmartin:
Adds the ETDistribution model to the package.swift

Intentionally not supporting cocoapods or usage from ObjC for now, that can come in a future version.

There are a few ways we could expose this API:

1. Using the `SentrySDK` naming so you would do something like `SentrySDK.checkForUpdates()`
  - a: As an extension from the existing `SentrySDK`. This is not preferred because it means users would have to depend on both the Sentry target, and the distribution target.
  - b: As a new target containing a new object with the same name. So there would be `Sentry.SentrySDK` and `SentryDistribution.SentrySDK` but most of the time you wouldn't need the module prefixes if they are being used in different files.
2. With a new type name like `Distribution`. This is what this PR does. The API is:
```
import SentryDistribution

Updater.checkForUpdates()
```

Closes #6173

---

> [!NOTE]
> Adds new `SentryDistribution` SwiftPM target with update-checking API, includes sample app and CI build/test coverage.
> 
> - **SDK / SPM**:
>   - Add `SentryDistribution` target and `SentryDistributionTests` in `Package.swift`.
>   - Implement update-checking API (`Updater`, models, networking, Mach-O UUID parsing, utils) under `Sources/SentryDistribution`.
> - **Samples**:
>   - Add `Samples/DistributionSample` app (XcodeGen spec, SwiftUI app, update flow) and include it in workspace and Makefile.
> - **CI**:
>   - Extend `.github/actions/prepare-package.swift` with `remove-duplicate` and `change-path` inputs and conditional steps.
>   - Add `build-distribution` job to build `SentryDistribution` framework.
>   - Add `distribution-tests` job to run SPM package tests (with duplicate target removal), and wire into required checks.
>   - Build matrix: include `DistributionSample` scheme.
> - **Changelog**:
>   - Unreleased: feature entry for `SentryDistribution` as a SwiftPM target.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6e87db3c1fc82f1bf8d081f6a81054c7ec39bb4c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,6 +92,29 @@ jobs:
         if: failure()
         run: ./scripts/ci-diagnostics.sh
 
+  distribution-tests:
+    name: Distribution Tests
+    runs-on: macos-15
+    needs: files-changed
+    steps:
+      - uses: actions/checkout@v5
+      - name: Prepare Package.swift
+        uses: ./.github/actions/prepare-package.swift
+        with:
+          is-pr: ${{ github.event_name == 'pull_request' }}
+          change-path: false
+          remove-duplicate: true
+      - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace
+      - run: set -o pipefail && NSUnbufferedIO=YES SKIP_BINARIES=1 xcodebuild test -scheme Sentry-Package -sdk iphonesimulator -destination 'platform=iOS Simulator,OS=18.4,name=iPhone 16 Pro' | tee raw-test-output-distribution.log | xcbeautify --preserve-unbeautified
+        shell: sh
+      - name: Upload Distribution Test Logs
+        uses: actions/upload-artifact@v4
+        if: ${{ failure() || cancelled() }}
+        with:
+          name: raw-test-output-distribution
+          path: |
+            raw-test-output-distribution.log
+
   # This matrix runs only the unit tests requiring the test server.
   # We do this to speed up the other unit test jobs and to avoid running the
   # test server with potential side effects in GH actions for all unit tests.
@@ -453,6 +476,7 @@ jobs:
       [
         files-changed,
         build-test-server,
+        distribution-tests,
         unit-tests,
         unit-tests-with-test-server,
       ]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add SentryDistribution as Swift Package Manager target (#6149)
+
 ## 8.56.2
 
 ### Fixes
--- a/Package.swift
+++ b/Package.swift
@@ -15,7 +15,8 @@ var products: [Product] = [
     .library(name: "Sentry-Dynamic-WithARM64e", targets: ["Sentry-Dynamic-WithARM64e"]),
     .library(name: "Sentry-WithoutUIKitOrAppKit", targets: ["Sentry-WithoutUIKitOrAppKit", "SentryCppHelper"]),
     .library(name: "Sentry-WithoutUIKitOrAppKit-WithARM64e", targets: ["Sentry-WithoutUIKitOrAppKit-WithARM64e", "SentryCppHelper"]),
-    .library(name: "SentrySwiftUI", targets: ["Sentry", "SentrySwiftUI", "SentryCppHelper"])
+    .library(name: "SentrySwiftUI", targets: ["Sentry", "SentrySwiftUI", "SentryCppHelper"]),
+    .library(name: "SentryDistribution", targets: ["SentryDistribution"])
 ]
 
 var targets: [Target] = [
@@ -66,7 +67,9 @@ var targets: [Target] = [
         linkerSettings: [
          .linkedLibrary("c++")
         ]
-    )
+    ),
+    .target(name: "SentryDistribution", path: "Sources/SentryDistribution"),
+    .testTarget(name: "SentryDistributionTests", dependencies: ["SentryDistribution"], path: "Sources/SentryDistributionTests")
 ]
 
 let env = getenv("EXPERIMENTAL_SPM_BUILDS")
@@ -101,7 +104,7 @@ if let env = env, String(cString: env, encoding: .utf8) == "1" {
             name: "SentryObjc",
             dependencies: ["SentrySwift"],
             path: "Sources",
-            exclude: ["Sentry/SentryDummyPublicEmptyClass.m", "Sentry/SentryDummyPrivateEmptyClass.m", "Swift", "SentrySwiftUI", "Resources", "Configuration", "SentryCppHelper"],
+            exclude: ["Sentry/SentryDummyPublicEmptyClass.m", "Sentry/SentryDummyPrivateEmptyClass.m", "Swift", "SentrySwiftUI", "Resources", "Configuration", "SentryCppHelper", "SentryDistribution", "SentryDistributionTests"],
             cSettings: [
                 .headerSearchPath("Sentry/include/HybridPublic"),
                 .headerSearchPath("Sentry"),
--- a/Samples/DistributionSample/DistributionSample.xcconfig
+++ b/Samples/DistributionSample/DistributionSample.xcconfig
@@ -1,0 +1,4 @@
+PRODUCT_BUNDLE_IDENTIFIER = io.sentry.sample.DistributionSample
+INFOPLIST_FILE = DistributionSample/Info.plist
+SUPPORTED_PLATFORMS = iphoneos iphonesimulator
+IPHONEOS_DEPLOYMENT_TARGET = 15.0
--- a/Samples/DistributionSample/DistributionSample/UpdateUtil.swift
+++ b/Samples/DistributionSample/DistributionSample/UpdateUtil.swift
@@ -1,0 +1,38 @@
+import Foundation
+import SentryDistribution
+import UIKit
+
+enum UpdateUtil {
+  @MainActor
+  static func checkForUpdates() {
+    let params = CheckForUpdateParams(accessToken: Constants.accessToken, organization: Constants.organization, project: Constants.project)
+    Updater.checkForUpdate(params: params) { result in
+      handleUpdateResult(result: result)
+    }
+  }
+  
+  static func handleUpdateResult(result: Result<UpdateCheckResponse, Error>) {
+    guard case let .success(releaseInfo) = result else {
+      if case let .failure(error) = result {
+        print("Error checking for update: \(error)")
+      }
+      return
+    }
+    
+    guard let releaseInfo = releaseInfo.update else {
+      print("Already up to date")
+      return
+    }
+    
+    UpdateUtil.installRelease(releaseInfo: releaseInfo)
+  }
+
+  private static func installRelease(releaseInfo: ReleaseInfo) {
+    guard let url = Updater.buildUrlForInstall(releaseInfo.downloadUrl) else {
+      return
+    }
+    DispatchQueue.main.async {
+      Updater.install(url: url)
+    }
+  }
+}
--- a/Sources/SentryDistribution/Mach-O/BinaryParser.swift
+++ b/Sources/SentryDistribution/Mach-O/BinaryParser.swift
@@ -1,0 +1,32 @@
+import Darwin
+import Foundation
+import MachO
+
+typealias GetUUIDFuncType = (@convention(c) (UnsafePointer<mach_header>, UnsafeRawPointer) -> Bool)
+fileprivate let _dyld_get_image_uuid = unsafeBitCast(dlsym(dlopen(nil, RTLD_LAZY), "_dyld_get_image_uuid"), to: GetUUIDFuncType.self)
+
+enum BinaryParser {
+  
+  static func getMainBinaryUUID(getUUID: GetUUIDFuncType = _dyld_get_image_uuid) -> String? {
+    guard let executableName = Bundle.main.infoDictionary?["CFBundleExecutable"] as? String else {
+      return nil
+    }
+    let executablePath = "\(Bundle.main.bundlePath)/\(executableName)"
+    
+    for i in 0..<_dyld_image_count() {
+      guard let header = _dyld_get_image_header(i) else { continue }
+      let imagePath = String(cString: _dyld_get_image_name(i))
+      
+      guard imagePath == executablePath else {
+        continue
+      }
+
+      var _uuid: uuid_t = (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
+      let _ = withUnsafeMutablePointer(to: &_uuid) {
+        getUUID(header, $0)
+      }
+      return UUID(uuid: _uuid).uuidString
+    }
+    return nil
+  }
+}
--- a/Sources/SentryDistribution/Models/CheckForUpdateParams.swift
+++ b/Sources/SentryDistribution/Models/CheckForUpdateParams.swift
@@ -1,0 +1,28 @@
+import Foundation
+
+/// A model for configuring parameters needed to check for app updates.
+public struct CheckForUpdateParams {
+  /// Creates a new instance with the required parameters for update checking.
+  /// - Parameters:
+  ///   - accessToken: Bearer token for authentication with Sentry API
+  ///   - organization: Sentry organization slug
+  ///   - project: Sentry project slug
+  ///   - hostname: Sentry hostname (defaults to "us.sentry.io")
+  ///   - binaryIdentifierOverride: Optional UUID override for the main binary
+  ///   - appIdOverride: Optional bundle identifier override
+  public init(accessToken: String, organization: String, project: String, hostname: String = "us.sentry.io", binaryIdentifierOverride: String? = nil, appIdOverride: String? = nil) {
+    self.accessToken = accessToken
+    self.organization = organization
+    self.project = project
+    self.hostname = hostname
+    self.binaryIdentifierOverride = binaryIdentifierOverride
+    self.appIdOverride = appIdOverride
+  }
+
+  let accessToken: String
+  let organization: String
+  let project: String
+  let hostname: String
+  let binaryIdentifierOverride: String?
+  let appIdOverride: String?
+}
--- a/Sources/SentryDistribution/Models/ReleaseInfo.swift
+++ b/Sources/SentryDistribution/Models/ReleaseInfo.swift
@@ -1,0 +1,25 @@
+import Foundation
+
+/// Information about a release version returned from the update check API.
+public struct ReleaseInfo: Decodable, Sendable {
+  /// Unique identifier for the release
+  public let id: String
+  /// Build version string (e.g., "1.2.3")
+  public let buildVersion: String
+  /// Build number (e.g., 123)
+  public let buildNumber: Int
+  /// Optional release notes describing changes
+  public let releaseNotes: String?
+  /// URL to download the IPA file
+  public let downloadUrl: String
+  /// Optional URL to the app icon
+  public let iconUrl: String?
+  /// Display name of the app
+  public let appName: String
+  private let createdDate: String
+
+  /// Parsed creation date from the server response
+  public var created: Date? {
+    Date.fromString(createdDate)
+  }
+}
--- a/Sources/SentryDistribution/Models/UpdateCheckResponse.swift
+++ b/Sources/SentryDistribution/Models/UpdateCheckResponse.swift
@@ -1,0 +1,9 @@
+import Foundation
+
+/// Response from the update check API containing current and available release information.
+public struct UpdateCheckResponse: Decodable, Sendable {
+  /// Information about the currently installed release
+  public let current: ReleaseInfo?
+  /// Information about the available update (nil if no update available)
+  public let update: ReleaseInfo?
+}
--- a/Sources/SentryDistribution/Network/URLSession+Distribute.swift
+++ b/Sources/SentryDistribution/Network/URLSession+Distribute.swift
@@ -1,0 +1,38 @@
+import Foundation
+
+extension URLSession {
+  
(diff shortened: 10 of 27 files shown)


Sources/SentryDistribution/Distribution.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+import UIKit
+
+@MainActor
```
noahsmartin: Should probably try to remove all the @MainActor from the public API, it doesn't seem necessary

.github/workflows/test.yml
```diff
@@ -92,6 +92,23 @@ jobs:
...
         run: ./scripts/ci-diagnostics.sh
 
+  distribution-tests:
+    name: Distribution Tests
+    runs-on: macos-15
+    needs: files-changed
+    steps:
+      - uses: actions/checkout@v5
+      - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace
+      - run: set -o pipefail && NSUnbufferedIO=YES xcodebuild test -scheme SentryDistributionTests -sdk iphoneos -destination 'generic/platform=iphoneos' | tee raw-test-output-spm.log | xcbeautify
```
philprime: `m`: We should start using `--preserve-unbeautified` so we can see all logs

```suggestion
      - run: set -o pipefail && NSUnbufferedIO=YES xcodebuild test -scheme SentryDistributionTests -sdk iphoneos -destination 'generic/platform=iphoneos' | tee raw-test-output-spm.log | xcbeautify --preserve-unbeautified
```

.github/workflows/test.yml
```diff
@@ -92,6 +92,23 @@ jobs:
...
+        shell: sh
+      - name: Upload SPM Test Logs
+        uses: actions/upload-artifact@v4
+        if: ${{ failure() || cancelled() }}
+        with:
+          name: raw-test-output-spm
+          path: |
+            raw-test-output-spm.log
```
philprime: `m`: Could cause naming conflicts

```suggestion
          name: raw-test-output-distribution-tests
          path: |
            raw-test-output-distribution-tests.log
```

Samples/DistributionSample/DistributionSample.xcconfig
```diff
@@ -0,0 +1,4 @@
+PRODUCT_BUNDLE_IDENTIFIER = io.sentry.DistributionSample
```
philprime: `m`: We should keep using the `io.sentry.sample.*` pattern for sample apps

```suggestion
PRODUCT_BUNDLE_IDENTIFIER = io.sentry.DistributionSample
```

Samples/DistributionSample/DistributionSample/UpdateUtil.swift
```diff
@@ -0,0 +1,49 @@
...
+        // Post notification event before closing the app
+        NotificationCenter.default.post(name: UIApplication.willTerminateNotification, object: nil)
+
+        // Close the app after a slight delay so it has time to execute code for the notification
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
+          // We need to exit since iOS doesn't start the install until the app exits
+          exit(0)
+        }
+      }
+    }
```
philprime, line 36: `h`: Why isn't this part of the SDK? It would seem more user-friendly if they can just call a single method in their apps to install updates.

Sources/SentryDistribution/Models/CheckForUpdateParams.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/// A model for configuring parameters needed to check for app updates.
+public struct CheckForUpdateParams {
+  public init(accessToken: String, organization: String, project: String, hostname: String = "us.sentry.io", binaryIdentifierOverride: String? = nil, appIdOverride: String? = nil) {
```
philprime, line 13: `h`: All public methods should be documented. Please add docs.

Sources/SentryDistribution/Models/ReleaseInfo.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+public struct ReleaseInfo: Decodable {
```
philprime: `h`: All public types and properties should be documented. Please add docs

Sources/SentryDistribution/Models/UpdateCheckResponse.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public struct UpdateCheckResponse: Decodable {
```
philprime: `h`: All public types and properties should be documented. Please add docs.

Sources/SentryDistribution/Network/URLSession+Distribute.swift
```diff
@@ -0,0 +1,46 @@
...
+extension URLSession {
+  func checkForUpdate(_ request: URLRequest, completion: @escaping (Result<UpdateCheckResponse, Error>) -> Void) {
+    self.perform(request, decode: UpdateCheckResponse.self, completion: completion)
+  }
+  
+  private func perform<T: Sendable & Decodable>(
+    _ request: URLRequest,
+    decode decodable: T.Type,
+    completion: @escaping (Result<T, Error>) -> Void) {
+    URLSession.shared.dataTask(with: request) { (data, response, error) in
```
philprime: `h`: As reported by Cursor BugBot, this should not use `URLSession.shared` but `self` instead

Sources/SentryDistribution/Network/URLSession+Distribute.swift
```diff
@@ -0,0 +1,46 @@
...
+      }
+      
+      do {
+        let jsonDecoder = JSONDecoder()
+        jsonDecoder.keyDecodingStrategy = .convertFromSnakeCase
+        result = .success(try jsonDecoder.decode(decodable, from: data))
+      } catch {
+        result = .failure(error)
+      }
+    }.resume()
```
philprime, line 35: `m`: IIUC the data task is resumed and executed, while not returning any reference. We should consider adding a deduplication logic so that multiple calls do not stack up, e.g. triggered multiple times before response is fetched due to bad network connection

Samples/DistributionSample/DistributionSample/UpdateUtil.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+import SentryDistribution
+import UIKit
+
+enum UpdateUtil {
+  @MainActor
+  static func checkForUpdates() {
+    let params = CheckForUpdateParams(accessToken: Constants.accessToken, organization: Constants.organization, project: Constants.project)
+    Updater.shared.checkForUpdate(params: params) { result in
```
philprime: `h`: The SentrySDK's interface is using `SentrySDK.<method>` without having to access a singleton first. We should keep that pattern.

Sources/SentryDistribution/Updater.swift
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  // MARK: - Private
+  private lazy var session = URLSession(configuration: URLSessionConfiguration.ephemeral)
+  private lazy var uuid = BinaryParser.getMainBinaryUUID()
+  
+  private func getUpdatesFromBackend(params: CheckForUpdateParams,
+                                     completion: @escaping ((Result<UpdateCheckResponse, Error>) -> Void)) {
+    guard var components = URLComponents(string: "http://\(params.hostname)/api/0/projects/\(params.organization)/\(params.project)/preprodartifacts/check-for-updates/") else {
+      fatalError("Invalid URL")
```
philprime: `h`: We must not crash the user app, especially when they can accidentally build a malformed URL by injecting the wrong params

Sources/SentryDistribution/Updater.swift
```diff
@@ -0,0 +1,78 @@
...
+    
+    components.queryItems = [
+      URLQueryItem(name: "main_binary_identifier", value: params.binaryIdentifierOverride ?? uuid),
+      URLQueryItem(name: "app_id", value: params.appIdOverride ?? Bundle.main.bundleIdentifier),
+      URLQueryItem(name: "platform", value: "ios"),
+      URLQueryItem(name: "version", value: "1.0")
+    ]
+    
+    guard let url = components.url else {
+      fatalError("Invalid URL")
```
philprime: `h`: We must not crash the user app

Sources/SentryDistribution/Updater.swift
```diff
@@ -0,0 +1,78 @@
...
+  
+  private func getUpdatesFromBackend(params: CheckForUpdateParams,
+                                     completion: @escaping ((Result<UpdateCheckResponse, Error>) -> Void)) {
+    guard var components = URLComponents(string: "http://\(params.hostname)/api/0/projects/\(params.organization)/\(params.project)/preprodartifacts/check-for-updates/") else {
+      fatalError("Invalid URL")
+    }
+    
+    components.queryItems = [
+      URLQueryItem(name: "main_binary_identifier", value: params.binaryIdentifierOverride ?? uuid),
+      URLQueryItem(name: "app_id", value: params.appIdOverride ?? Bundle.main.bundleIdentifier),
```
philprime: `m`: The `Bundle.main.bundleIdentifier` is an nullable option. We should handle that properly and abort fetching updates if it's not available

Sources/SentryDistribution/Utils/Date+Parse.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+extension Date {
+  static func fromString(_ input: String) -> Date? {
+    let formatter = ISO8601DateFormatter()
+    formatter.formatOptions = [ .withInternetDateTime, .withFractionalSeconds ]
+    return formatter.date(from: input)
+  }
```
philprime, line 14: `l`: IIRC it's best practice to reuse date formatters. Probably not possible if this stays an extension of Date.

Rough idea:
```suggestion
  private static var iso8601Formatter: ISO8601DateFormatter = {
    let formatter = ISO8601DateFormatter()
    formatter.formatOptions = [ .withInternetDateTime, .withFractionalSeconds ]
    return formatter
  }()
  
  static func fromString(_ input: String) -> Date? {
    iso8601Formatter.date(from: input)
  }
```

Sources/SentryDistribution/Mach-O/BinaryParser.swift
```diff
@@ -0,0 +1,42 @@
+import Darwin
+import Foundation
+import MachO
+
+fileprivate let _dyld_get_image_uuid = unsafeBitCast(dlsym(dlopen(nil, RTLD_LAZY), "_dyld_get_image_uuid"), to: (@convention(c) (UnsafePointer<mach_header>, UnsafeRawPointer) -> Bool).self)
+
+struct BinaryParser {
```
philprime: `h`: This logic is quite complex. Please add unit tests.

Sources/SentryDistribution/Models/CheckForUpdateParams.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/// A model for configuring parameters needed to check for app updates.
+public struct CheckForUpdateParams {
```
philprime, line 4: `m`: It can easily happen that the default values are changed by accident. Please add unit tests so we can detect unwanted changes

Sources/SentryDistribution/Models/ReleaseInfo.swift
```diff
@@ -0,0 +1,16 @@
...
+  public let buildNumber: Int
+  public let releaseNotes: String?
+  public let downloadUrl: String
+  public let iconUrl: String?
+  public let appName: String
+  private let createdDate: String
+
+  public var created: Date? {
+    Date.fromString(createdDate)
+  }
```
philprime, line 24: `m`: The `createdDate` is a constant, so we could parse the `created` Date once when initializing and assign it as another constant. But it would require implementing the decoding logic which is compiler generated. I'll leave it to you.

Sources/SentryDistribution/Updater.swift
```diff
@@ -0,0 +1,78 @@
+import Foundation
+import UIKit
+
+public final class Updater {
```
philprime: `h`: We should add unit tests for this class and try to mock the network calls.

Sources/SentryDistribution/Updater.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+    components.queryItems = [
+      URLQueryItem(name: "action", value: "download-manifest"),
+      URLQueryItem(name: "url", value: plistUrl)
+    ]
+    return components.url
+  }
+
+  // MARK: - Private
+  private lazy var session = URLSession(configuration: URLSessionConfiguration.ephemeral)
```
philprime: `m`: Add an initializer with an ephemeral session as the default value. We want this to be injectable for unit tests

philprime:
Why do you want to skip the changelog when introducing such a big new feature?

Samples/DistributionSample/DistributionSample/UpdateUtil.swift
noahsmartin: It wasn't in the SDK when @itaybre made the emerge version. I guess the idea was to keep it as light as possible. Although I can definitely see the benefits of including it. So if @itaybre doesn't have any objections I'll move it in

Sources/SentryDistribution/Network/URLSession+Distribute.swift
noahsmartin: I think that is handled by URLSession automatically, but will check

noahsmartin:
I don't think we would actually skip the changelog, I just want to align on the API before writing a changelog, and don't want the bot to post that big comment on the PR

Sources/SentryDistribution/Models/ReleaseInfo.swift
noahsmartin: I was able to make the date formatter only initialize once, so don't think it's necessary to change this

noahsmartin:
Thanks @philprime for the review, just updated it. PTAL whenever you get a chance

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add SentryDistribution Swift package target (#6149)
```
philprime: `l`: nitpick

```suggestion
- Add SentryDistribution as Swift Package Manager target (#6149)
```

Package.swift
```diff
@@ -66,9 +43,44 @@ var targets: [Target] = [
...
          .linkedLibrary("c++")
         ]
-    )
+    ),
+    .target(name: "SentryDistribution", path: "Sources/SentryDistribution"),
+    .testTarget(name: "SentryDistributionTests", dependencies: ["SentryDistribution"], path: "Sources/SentryDistributionTests")
 ]
 
+let skipBinaries = getenv("SKIP_BINARIES").map { String(cString: $0, encoding: .utf8) }
+if skipBinaries != "1" {
```
philprime: `h`: Please add some sort of verification check so we know that this change is not breaking our release process.

Sources/SentryDistribution/Updater.swift
```diff
@@ -0,0 +1,113 @@
...
+  ///
+  ///
+  /// - Parameters:
+  ///   - params: A `CheckForUpdateParams` object.
+  ///   - completion: A closure that is called with the result of the update check.
+  ///
+  /// - Example:
+  /// ```
+  /// let params = CheckForUpdateParams(accessToken: "your_access_token")
+  /// checkForUpdate(params: params) { result in
```
philprime:
```suggestion
  /// Updater.checkForUpdate(params: params) { result in
```

Sources/SentryDistribution/Updater.swift
```diff
@@ -0,0 +1,113 @@
...
+    shared.getUpdatesFromBackend(params: params) { result in
+      completion(result.mapError { $0 })
+    }
+  }
+
+  #if canImport(UIKit)
+  /// Install an update from the provided URL
+  /// - Parameter url: The itms-services URL
+  @MainActor public static func install(url: URL) {
+    UIApplication.shared.open(url) { _ in
```
philprime, line 45: `m`: I believe it's best practice to use [UIApplication.canOpenUrl(_:)](https://developer.apple.com/documentation/uikit/uiapplication/canopenurl(_:)) first and you should check the callback `success` parameter to check if the URL was opened successfully and do an early-exit otherwise.

Sources/SentryDistribution/Updater.swift
```diff
@@ -0,0 +1,113 @@
...
+  }
+  #endif
+
+  /// Obtain a URL to install an IPA
+  /// - Parameter plistUrl: The URL to the plist containing the IPA information
+  /// - Returns: a URL ready to install the IPA using itms-services
+  public static func buildUrlForInstall(_ plistUrl: String) -> URL? {
+    guard var components = URLComponents(string: "itms-services://") else {
+      return nil
+    }
```
philprime: `m`: IIUC we will never get into this case because we expect the scheme to work. We could create the object without parsing a string:

```suggestion
    var components = URLComponents()
    components.scheme= "itms-services"
```

Sources/SentryDistribution/Updater.swift
```diff
@@ -0,0 +1,113 @@
...
+  
+  // MARK: - Internal
+  init(session: URLSession = URLSession(configuration: URLSessionConfiguration.ephemeral)) {
+    self.session = session
+  }
+  
+  func getUpdatesFromBackend(
+    params: CheckForUpdateParams,
+    completion: @escaping @Sendable (Result<UpdateCheckResponse, Error>) -> Void) {
+    guard var components = URLComponents(string: "http://\(params.hostname)/api/0/projects/\(params.organization)/\(params.project)/preprodartifacts/check-for-updates/") else {
```
philprime: `h`: We should use `https`

```suggestion
    guard var components = URLComponents(string: "https://\(params.hostname)/api/0/projects/\(params.organization)/\(params.project)/preprodartifacts/check-for-updates/") else {
```

Sources/SentryDistributionTests/BinaryParserTests.swift
```diff
@@ -0,0 +1,13 @@
...
+
+@Test func testBinaryParserGetsCorrectUUID() throws {
+  let uuid = BinaryParser.getMainBinaryUUID { _, ptr in
+    let mutable = UnsafeMutableRawPointer(mutating: ptr)
+    for i in 0..<16 {
+      mutable.storeBytes(of: 10, toByteOffset: i, as: CChar.self)
+    }
+    return true
+  }
+  #expect(uuid == "0A0A0A0A-0A0A-0A0A-0A0A-0A0A0A0A0A0A")
```
philprime, line 14: `m`: Where is this defined? Will this ever change? Please add a comment for future reference.

Sources/SentryDistribution/Updater.swift
noahsmartin: I don't think that's best practice, IMO it's discouraged. If you use that method you have to add it to the Info.plist 

```
If you link your app on or after iOS 9.0, you must declare the URL schemes you pass to this method by adding the LSApplicationQueriesSchemes key to your app’s Info.plist file. This method always returns [false](https://developer.apple.com/documentation/swift/false) for undeclared schemes, even if the device doesn’t have a registered app installed. Apps linked on or after iOS 15 are limited to a maximum of 50 entries in the LSApplicationQueriesSchemes key. To learn more about the key, see [LSApplicationQueriesSchemes](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/LaunchServicesKeys.html#//apple_ref/doc/plist/info/LSApplicationQueriesSchemes).
```

From https://developer.apple.com/documentation/uikit/uiapplication/canopenurl(_:)

That doc also mentions using the method I am using here as an alternative that doesn't require calling canOpen. The reason it's not preferred is it's commonly used to exfiltrate data about users. For example you can call `canOpenUrl("shatsapp://")` to silently check if the user has WhatsApp installed on their phone. Best to avoid this method entirely and use this new callback based approach that was added in iOS 10 like we're doing here

Sources/SentryDistributionTests/BinaryParserTests.swift
noahsmartin: This is just a test, so I had to use something, I just chose to use this as my test id, we can pick anything just needs to be something. I can comment

Package.swift
noahsmartin: Don't think this is actually needed anymore, removed it

noahsmartin:
Thanks @philprime it's ready for another pass whenever you get a chance!

Sources/SentryDistribution/Updater.swift
philprime: TIL, not sure why `canOpenUrl` still exists and is not deprecated then.

philprime:
@noahsmartin I did not expect the bot to change the PR, you might want to revert the last commit

PR #6159, opened 2025-09-12
ref: SentryFileManager to Swift

### LINKED ISSUE #6238: ref: SentryFileManager to Swift
> [!NOTE]
> The pull request "[ref: SentryFileManager to Swift](https://github.com/getsentry/sentry-cocoa/pull/6159)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

To make this safe most of the conversion is just wrapping the old SentryFileManager which I renamed to SentryFileManagerHelper. There were two changes actually made:

1 - Many functions in SentryFileManagerHelper that previously used an SDK internal type like AppState or Session, now use NSData. The serializing/deserializing is done in the Swift version but is just one function call so very little code change.

2 - The callback for pruning the maximum number of events has been moved to Swift, it's just one function and had existing tests.

#skip-changelog

noahsmartin:
To make this safe most of the conversion is just wrapping the old SentryFileManager which I renamed to SentryFileManagerHelper. There were two changes actually made:

1 - Many functions in SentryFileManagerHelper that previously used an SDK internal type like AppState or Session, now use NSData. The serializing/deserializing is done in the Swift version but is just one function call so very little code change.

2 - The callback for pruning the maximum number of envelopes has been moved to Swift, it's just one function and had existing tests.

#skip-changelog

Closes #6238

### DIFF
--- a/SentryTestUtils/SentryFileManager+Test.h
+++ b/SentryTestUtils/SentryFileManager+Test.h
@@ -1,4 +1,4 @@
-#import "SentryFileManager.h"
+#import "SentryFileManagerHelper.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -11,11 +11,7 @@ NSString *_Nullable sentryBuildScopedCachesDirectoryPath(NSString *cachesDirecto
 SENTRY_EXTERN NSURL *_Nullable launchProfileConfigFileURL(void);
 SENTRY_EXTERN NSURL *_Nullable sentryLaunchConfigFileURL;
 
-@interface SentryFileManager ()
-
-@property (nonatomic, copy) NSString *eventsPath;
-@property (nonatomic, copy) NSString *envelopesPath;
-@property (nonatomic, copy) NSString *timezoneOffsetFilePath;
+@interface SentryFileManagerHelper ()
 
 - (void)clearDiskState;
 
--- a/Sources/Sentry/SentryAsyncLog.m
+++ b/Sources/Sentry/SentryAsyncLog.m
@@ -1,6 +1,6 @@
 #import "SentryAsyncLog.h"
 #import "SentryAsyncSafeLog.h"
-#import "SentryFileManager.h"
+#import "SentryFileManagerHelper.h"
 #import "SentryInternalCDefines.h"
 #import "SentryInternalDefines.h"
 #import "SentryLogC.h"
--- a/Sources/Sentry/include/SentryFileManagerHelper.h
+++ b/Sources/Sentry/include/SentryFileManagerHelper.h
@@ -1,4 +1,3 @@
-#import "SentryDataCategory.h"
 #import "SentryDefines.h"
 #import "SentryProfilingConditionals.h"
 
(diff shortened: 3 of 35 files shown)


SentryTestUtils/SentryFileManager+Test.h
```diff
@@ -11,11 +11,7 @@ NSString *_Nullable sentryBuildScopedCachesDirectoryPath(NSString *cachesDirecto
 SENTRY_EXTERN NSURL *_Nullable launchProfileConfigFileURL(void);
 SENTRY_EXTERN NSURL *_Nullable sentryLaunchConfigFileURL;
 
-@interface SentryFileManager ()
-
-@property (nonatomic, copy) NSString *eventsPath;
-@property (nonatomic, copy) NSString *envelopesPath;
-@property (nonatomic, copy) NSString *timezoneOffsetFilePath;
+@interface SentryFileManagerHelper ()
```
itaybre, line 14: Maybe `SentryFileManagerTestHelper`?

Sources/Sentry/SentryAsyncLog.m
```diff
@@ -1,6 +1,6 @@
 #import "SentryAsyncLog.h"
 #import "SentryAsyncSafeLog.h"
-#import "SentryFileManager.h"
+#import "SentryFileManagerHelper.h"
```
itaybre, line 3: This import is only being used during tests, right?

Sources/Sentry/include/SentryFileManagerHelper.h
```diff
@@ -9,14 +8,12 @@ NS_ASSUME_NONNULL_BEGIN
...
 @class SentryEnvelopeItem;
-@class SentryFileContents;
 @class SentryOptions;
 @class SentrySession;
 
 @protocol SentryCurrentDateProvider;
 
-NS_SWIFT_NAME(SentryFileManager)
-@interface SentryFileManager : NSObject
+@interface SentryFileManagerHelper : NSObject
```
itaybre, line 16: I see sometimes we use Helper or Internal for this classes, not blocking but maybe we should align on naming

Sources/Sentry/SentryAsyncLog.m
itaybre: I just noticed this was the internal class, so maybe we should use different names for the files

noahsmartin: I'm not sure what you mean, the file name is changed?

Sources/Sentry/include/SentryFileManagerHelper.h
noahsmartin: No preference from me, happy to change it, let me know what you'd prefer

SentryTestUtils/SentryFileManager+Test.h
noahsmartin: This isn't a separate class, it's a category on the existing one. I kept this the same as before just update the name to match the updated name in the other file

itaybre: Ohh, you are right, looks good then

Sources/Sentry/SentryAsyncLog.m
itaybre: This is on me, I misunderstood.
Disregard this comment

PR #6182, opened 2025-09-17
ref: ANRIntegration depContainer thread inspector

### LINKED ISSUE #6183: ref: ANRIntegration depContainer thread inspector
> [!NOTE]
> The pull request "[ref: ANRIntegration depContainer thread inspector](https://github.com/getsentry/sentry-cocoa/pull/6182)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Use the thread inspector from the dependency container instead of getting it via the client to align with the other dependencies.

This came up while investigating flaky tests.

#skip-changelog

philipphofmann:
Use the thread inspector from the dependency container instead of getting it via the client to align with the other dependencies.

This came up while investigating flaky tests.

#skip-changelog 

Closes #6183

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -186,6 +186,8 @@ jobs:
             test-destination-os: "latest"
             scheme: "Sentry"
 
+          ## We don't run unit tests on macCatalyst 26 yet because of https://github.com/getsentry/sentry-cocoa/issues/6165.
+
           # We don't run the unit tests on tvOS 16 cause we run them on all on GH actions available iOS versions.
           # The chance of missing a bug solely on tvOS 16 that doesn't occur on iOS, tvOS 15 or tvOS 16 is minimal.
           # We are running tests on tvOS 17 and latest, as there were OS-internal changes introduced in succeeding versions.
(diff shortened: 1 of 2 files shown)


Tests/SentryTests/Integrations/ANR/SentryANRTrackingIntegrationTests.swift
```diff
@@ -773,7 +780,7 @@ class SentryANRTrackingIntegrationTests: SentrySDKIntegrationTestsBase {
         } else {
             threadInspector.allThreads = []
         }
-        
-        SentrySDKInternal.currentHub().getClient()?.threadInspector = threadInspector
+
+        SentryDependencyContainer.sharedInstance().threadInspector = threadInspector
```
noahsmartin, line 786: Ideally we would be passing this through the init not setting/changing it here, so that in Swift it could be a `let`. But this PR seems like an improvement either way, no need to make all the changes at once

noahsmartin (approved):

Tests/SentryTests/Integrations/ANR/SentryANRTrackingIntegrationTests.swift
philippsuess: Yes, indeed. The problem here is that the entry point is the SentryANRTrackingIntegration, which gets initialized from the SentrySDKInternal, so we can't entirely get rid of the dependency container, but we can ditch the dependency container in the tests by adding an initializer just for the tests. I'm going to follow up. Thanks for the hint.

PR #6186, opened 2025-09-17
ref: Envelope deserialization to Swift

noahsmartin:
#skip-changelog


Sources/Swift/Helper/DataDeserialization.swift
```diff
@@ -1,3 +1,5 @@
+@_implementationOnly import _SentryPrivate
+
 @_spi(Private) @objc public final class DataDeserialization: NSObject {
```
philippsuess: `h`: Don't we have to prefix this class with Sentry to avoid naming collisions cause it's exposed to ObjC?

Sources/Swift/Helper/DataDeserialization.swift
```diff
@@ -29,4 +31,154 @@
...
+                envelopeHeaderIndex = i
+                // Envelope header end
+                let headerData = data.subdata(in: 0..<i)
+                #if DEBUG
+                  if let headerString = String(data: headerData, encoding: .utf8) {
+                    SentrySDKLog.debug("Header \(headerString)")
+                  }
+                #endif
+                do {
+                    let headerDictionary = try JSONSerialization.jsonObject(with: headerData) as? [String: Any]
```
philippsuess: `m`: If I'm not mistaken, we can move the catch block up here, cause nothing below uses try. This would be better for readability, because the logging of the error is closer to the try here.

Sources/Swift/Helper/DataDeserialization.swift
```diff
@@ -29,4 +31,154 @@
...
+        var itemHeaderStart = envelopeHeaderIndex + 1
+        var items: [SentryEnvelopeItem] = []
+        let endOfEnvelope = data.count - 1
+
+        var i = itemHeaderStart
+        while i <= endOfEnvelope {
+            defer {
+                i += 1
+            }
+            if bytes[i] == UInt8(ascii: "\n") || i == endOfEnvelope {
```
philippsuess: `l`: We could define a variable or `UInt8(ascii: "\n")` at the top of this method, so we don't have it twice.

Sources/Swift/Helper/DataDeserialization.swift
```diff
@@ -29,4 +31,154 @@
...
+                    let itemBody = (data as NSData).subdata(with: NSRange(location: i + 1, length: Int(bodyLength)))
+                    let envelopeItem = SentryEnvelopeItem(header: itemHeader, data: itemBody)
+                    items.append(envelopeItem)
+
+                    i = i + 1 + Int(bodyLength)
+                    itemHeaderStart = i
+                } catch {
+                    SentrySDKLog.error("Failed to parse envelope item header \(error)")
+                    return nil
+                }
```
philippsuess: `l`: Again I think we should be able to move this catch statement up a bit so it's closer to the try.

philipphofmann (approved):
Woah, this code is complex 🤯 . Thanks for converting it to Swift. It looks way better now.

Sources/Swift/Helper/DataDeserialization.swift
noahsmartin: For some reason I thought it would be `Sentry.DataSerialization` but it of course isn't because I use it in objc from other places in this PR. So yes, I'll update it

Sources/Swift/Helper/DataDeserialization.swift
noahsmartin: The problem is we need to use the `headerDictionary` variable in the below code, if we moved the code that uses it out of the do/catch I think we would need to use an extra `guard` Even though we would know if should be impossible to reach that point with headerDictionary being null, Swift would not guarantee it. That could lead to more bugs in the future + a codepath that would be untestable/unreachable

Sources/Swift/Helper/DataDeserialization.swift
noahsmartin: Same as above

PR #6187, opened 2025-09-17
chore: Refine sed command for binaryTarget removal

### LINKED ISSUE #6188: chore: Refine sed command for binaryTarget removal
> [!NOTE]
> The pull request "[chore: Refine sed command for binaryTarget removal](https://github.com/getsentry/sentry-cocoa/pull/6187)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Also removed step to delete newer package files from action, we don't need it any more since there is only one Package.swift

#skip-changelog

noahsmartin:
Also removed step to delete newer package files from action, we don't need it any more since there is only one Package.swift

#skip-changelog

Closes #6188


philipphofmann (approved):
Thanks, LGTM

PR #6208, opened 2025-09-19
ref: Update SentryCrashReportConverter for improved nullability handling

philprime:
## :scroll: Description

- Renamed SentryKSCrashReportConverterTests to SentryCrashReportConverterTests for consistency.
- Enhanced nullability checks in SentryCrashReportConverter to prevent potential crashes.
- Added new test cases to validate handling of various edge cases in crash report conversion.

Closes [#6207](https://github.com/getsentry/sentry-cocoa/issues/6207)

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryCrashReportConverter.m
+++ b/Sources/Sentry/SentryCrashReportConverter.m
@@ -51,7 +51,7 @@
         // here. For more details please check out SentryCrashScopeObserver.
         NSMutableDictionary *userContextMerged =
             [[NSMutableDictionary alloc] initWithDictionary:userContextUnMerged];
-        [userContextMerged addEntriesFromDictionary:report[@"sentry_sdk_scope"]];
+        [userContextMerged addEntriesFromDictionary:report[@"sentry_sdk_scope"] ?: @{}];
         [userContextMerged removeObjectForKey:@"sentry_sdk_scope"];
         self.userContext = userContextMerged;
 
@@ -108,8 +108,9 @@
         if ([self.report[@"report"][@"timestamp"] isKindOfClass:NSNumber.class]) {
             event.timestamp = [NSDate
                 dateWithTimeIntervalSince1970:[self.report[@"report"][@"timestamp"] integerValue]];
-        } else {
-            event.timestamp = sentry_fromIso8601String(self.report[@"report"][@"timestamp"]);
+        } else if ([self.report[@"report"][@"timestamp"] isKindOfClass:NSString.class]) {
+            event.timestamp = sentry_fromIso8601String(
+                SENTRY_UNWRAP_NULLABLE(NSString, self.report[@"report"][@"timestamp"]));
         }
         event.threads = [self convertThreads];
         event.debugMeta = [self debugMetaForThreads:event.threads];
@@ -119,7 +120,7 @@
         event.environment = self.userContext[@"environment"];
 
         NSMutableDictionary *mutableContext =
-            [[NSMutableDictionary alloc] initWithDictionary:self.userContext[@"context"]];
+            [[NSMutableDictionary alloc] initWithDictionary:self.userContext[@"context"] ?: @{}];
         if (self.userContext[@"traceContext"]) {
             mutableContext[@"trace"] = self.userContext[@"traceContext"];
         }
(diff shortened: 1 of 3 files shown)


philprime:
@cursor review

philprime:
@cursor review
@sentry review

Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -186,13 +187,19 @@ - (SentryUser *_Nullable)convertUser
     if (nil != self.userContext[@"breadcrumbs"]) {
         NSArray *storedBreadcrumbs = self.userContext[@"breadcrumbs"];
         for (NSDictionary *storedCrumb in storedBreadcrumbs) {
+            if (!storedCrumb[@"category"]) {
```
philprime: We have two options:

1. Drop the incomplete breadcrumb
2. Fallback to the default category `@"default"`

Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -399,19 +428,20 @@ - (SentryDebugMeta *)debugMetaFromBinaryImageDictionary:(NSDictionary *)sourceIm
                               self.exceptionContext[@"mach"][@"exception"],
                               self.exceptionContext[@"mach"][@"code"],
                               self.exceptionContext[@"mach"][@"subcode"]]
-                     type:self.exceptionContext[@"mach"][@"exception_name"]];
+                     type:self.exceptionContext[@"mach"][@"exception_name"] ?: @"Mach Exception"];
```
philprime: This fallback is best-attempt, so if we have a better fitting one I am open for feedback

Sources/Sentry/Public/SentryThread.h
```diff
@@ -21,9 +21,11 @@ NS_ASSUME_NONNULL_BEGIN
 SENTRY_NO_INIT
 
 /**
- * Number of the thread
+ * Number of the thread.
+ *
+ * Can be nil for threads in recrash reports where the thread index information is not available.
  */
-@property (nonatomic, copy) NSNumber *threadId;
+@property (nullable, nonatomic, copy) NSNumber *threadId;
```
philprime, line 29: The changes are public API but they are loosening the API contract so the changes are non-breaking

itaybre: Are you sure? On Swift that could be a breaking change

Sources/Sentry/SentryCrashReportConverter.m
itaybre: Could we add this as a comment?

Sources/Sentry/Public/SentryThread.h
philippsuess: Yep, this is a breaking change. Making something non-nullable nullable is a breaking change. People could rely on this not being nullable and now it's nullable. Then the compiler complains and bam breaking change.

philprime: I did not consider that this would require Swift users to add null-handling if they use this `threadId`, apologies for that.

But still, right now the field is working similar as if we do a force-unwrapping, therefore if a user actually uses the `threadId` and rely on it to never be `nil`, it might actually happen that it is `nil` and their app will crash.
Therefore even though it is a breaking change, we should actually break customers to make them aware and push them to adapting their usage of the `threadId`.

I will check again if this field is ever `nil`. Just because it could have been a null value before, it might not be one after this PR anyways.

philippsuess: Yep thanks. Anyways we can make it nullable in V9. So then it wouldn't be a problem to change it.

philprime: I reverted the changes with compiler flags for non-v9

Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -428,20 +428,25 @@ - (SentryDebugMeta *)debugMetaFromBinaryImageDictionary:(NSDictionary *)sourceIm
                               self.exceptionContext[@"mach"][@"exception"],
                               self.exceptionContext[@"mach"][@"code"],
                               self.exceptionContext[@"mach"][@"subcode"]]
-                     type:self.exceptionContext[@"mach"][@"exception_name"] ?: @"Mach Exception"];
+                     type:self.exceptionContext[@"mach"][@"exception_name"]
+                ?: @"Mach Exception"]; // The fallback value is best-attempt in case the exception
+                                       // name is not available
```
philippsuess, line 433: Thanks for the comment.

philipphofmann (approved):
LGTM

PR #6209, opened 2025-09-19
ref: Convert SentrySysctl to Swift

### LINKED ISSUE #6210: ref: Convert SentrySysctl to Swift
> [!NOTE]
> The pull request "[ref: Convert SentrySysctl to Swift](https://github.com/getsentry/sentry-cocoa/pull/6209)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Convert this to swift by wrapping the ObjC implementation

#skip-changelog

noahsmartin:
Convert this to swift by wrapping the ObjC implementation

#skip-changelog

Closes #6210

### DIFF
--- a/Sources/Swift/Helper/SentrySysctl.swift
+++ b/Sources/Swift/Helper/SentrySysctl.swift
@@ -1,0 +1,37 @@
+@_implementationOnly import _SentryPrivate
+
+/// A wrapper around sysctl for testability.
+@_spi(Private) @objc public class SentrySysctl: NSObject {
+    
+    private let objcHelper = SentrySysctlObjC()
+    
+    /// Returns the time the system was booted with a precision of microseconds.
+    ///
+    /// @warning We must not send this information off device because Apple forbids that.
+    /// We are allowed send the amount of time that has elapsed between events that occurred within the
+    /// app though. For more information see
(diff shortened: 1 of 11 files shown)


Sources/Swift/Helper/SentrySysctl.swift
```diff
@@ -0,0 +1,37 @@
+@_implementationOnly import _SentryPrivate
+
+/// A wrapper around sysctl for testability.
+@_spi(Private) @objc public class SentrySysctl: NSObject {
+    
+    private let objcHelper = SentrySysctlObjC()
+    
+    /// Returns the time the system was booted with a precision of microseconds.
+    ///
+    /// @warning We must not send this information off device because Apple forbids that.
```
philprime, line 10: `m`: Should we consider making it impossible to access the system boot timestamp and only allow accessing elapsed time instead to enforce this warning? If yes, let's do it in a follow-up PR.

noahsmartin: Well it would always be possible for someone to accidentally not use this class and access the system boot timestamp directly. Unless we maybe combined it with some lint rule? I will create a follow up GitHub issue to discuss

noahsmartin: https://github.com/getsentry/sentry-cocoa/issues/6233

PR #6237, opened 2025-09-22
Hang tracker abstraction

noahsmartin:
WIP

#skip-changelog


philprime:
What's the status of this PR? I see it is outdated and has merge conflicts due to being open for a while now.

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,216 @@
...
+        removeObserver: @escaping RemoveObserverFunc<T>,
+        queue: DispatchQueue = DispatchQueue(label: "io.sentry.runloop-observer-checker")
+    ) {
+        self.dateProvider = dateProvider
+        self.createObserver = createObserver
+        self.addObserver = addObserver
+        self.removeObserver = removeObserver
+        self.queue = queue
+#if canImport(UIKit) && !SENTRY_NO_UIKIT && (!swift(>=5.9) || !os(visionOS)) && !os(watchOS)
+        var maxFPS = 60.0
```
philprime: `m`: On ProMotion displays we have higher frame rates than 60, e.g. when switching apps it can go up to 120. Do we need to consider that here?

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+        return id
+    }
+    
+    func removeLateRunLoopObserver(id: UUID) {
+        queue.async { [weak self] in
+            guard let self else { return }
+            lateRunLoop.removeValue(forKey: id)
+            if lateRunLoop.isEmpty {
```
philprime: `h`: Seer already mentioned it, I believe the two nested `async` statements could allow a race-condition

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,216 @@
...
+        let id = UUID()
+        finishedRunLoop[id] = handler
+        startIfNecessary()
+        return id
+    }
+    
+    func removeFinishedRunLoopObserver(id: UUID) {
+        finishedRunLoop.removeValue(forKey: id)
+        if finishedRunLoop.isEmpty {
+            queue.async { [weak self] in
```
philprime: h: Seer already mentioned it, I believe the two nested async statements could allow a race-condition

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegrationSwift.swift
```diff
@@ -0,0 +1,50 @@
...
+
+     init(hangTracker: HangTracker, timeoutInterval: TimeInterval, hangStarted: @escaping () -> Void, hangStopped: @escaping () -> Void) {
+         self.tracker = hangTracker
+        self.timeoutInterval = timeoutInterval
+        self.hangStarted = hangStarted
+        self.hangStopped = hangStopped
+    }
+    
+    @objc public func start() {
+        dispatchPrecondition(condition: .onQueue(.main))
```
philprime: `h`: We do not use `dispatchPrecondition` in the SDK because it causes runtime crashes if used incorrectly.  Instead we should do a main thread check and fail softly.

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegrationSwift.swift
```diff
@@ -0,0 +1,50 @@
...
+        }
+        
+        let finished = tracker.addFinishedRunLoopObserver { [weak self] _ in
+            self?.hangStopped()
+        }
+        callbackId = (late, finished)
+    }
+    
+    @objc public func stop() {
+        dispatchPrecondition(condition: .onQueue(.main))
```
philprime: `h`: See above wrt. not using `dispatchPrecondition`

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,216 @@
...
+        self.observer = nil
+    }
+    
+    // MARK: Background queue
+    
+    private var lateRunLoop = [UUID: (UUID, TimeInterval) -> Void]()
+    private var hangId = UUID()
+    
+    private func waitForHang(semaphore: DispatchSemaphore, started: TimeInterval, isStarting: Bool) {
+        dispatchPrecondition(condition: .onQueue(queue))
```
philprime: `h`: We do not use `dispatchPrecondition` in the SDK because it causes runtime crashes. Please use a soft or dispatching to main

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,216 @@
...
+            default:
+                fatalError()
+            }
+        }
+        self.observer = observer
+        addObserver(CFRunLoopGetMain(), observer, .commonModes)
+    }
+    
+    private func stop() {
+        dispatchPrecondition(condition: .onQueue(.main))
```
philprime: `h`: We do not use `dispatchPrecondition` in the SDK because it causes runtime crashes. Please use a soft or dispatching to main

Tests/SentryTests/Integrations/ANR/HangTrackerTests.swift
```diff
@@ -0,0 +1,118 @@
+@_spi(Private) @testable import Sentry
+@_spi(Private) import SentryTestUtils
+import XCTest
+
+struct TestRunLoopObserver: RunLoopObserver { }
```
philprime: `m`: While I believe this to be a valid unit test, I am not 100% sure if I understand the mocking here. Do we have a test case using a real runloop to test this feature?

philprime (commented):
Almost LGTM, left some high priority comments

philprime:
FYI @noahsmartin thanks for creating this draft and doing the heavy lifting.
I am going to take over this PR to get it finished.

philprime:
@sentry review
@cursor review

philprime:
**Update on this PR:**

I have completed the implementation of the app hang tracker, according to our internal architecture. The testing has been harder to implement than hoped for, so this will now require me to manually fix the unit tests and integration tests of the app hang tracker. After that it should be possible to review this.

philprime:
## Progress update

### SentryHangTracker concurrency fixes

Audited all dispatch patterns in `SentryHangTracker.swift` and fixed several issues:

- **Deadlock fix**: `waitForHangIterative` was calling `queue.dispatchSync` back to itself on a serial queue. Removed it — `lateRunLoop` is accessed directly since the method already runs on the serial queue.
- **Premature stop fix**: `removeFinishedRunLoopObserver` could call `stop()` even when `lateRunLoop` still had active observers. Added a cross-dictionary check.
- **Dedicated queue**: The hang tracker now gets its own serial queue (`io.sentry.hang-tracker`) in `SentryDependencyContainer` instead of sharing the default `dispatchQueueWrapper`. This prevents `waitForHangIterative` from blocking unrelated SDK operations.
- **Documentation**: Added thread model docs explaining why the queue must be serial and dedicated, and why `dispatchSync` must not be used from within the blocked thread.

### Unit tests (`SentryHangTrackerTests.swift`)

Simplified from ~1000 lines / 26 tests down to ~500 lines / 15 tests:

- Removed 11 tests that had no practical value (crash-only tests on safe operations, concurrent tests with mock queues, async semantics tests with no assertions, etc.)
- Removed all `NSLock`s and thread-safe wrappers that were only needed for the deleted concurrent tests.
- Fixed a race condition test (`testRemoveFinishedRunLoopObserver_whenNewObserverAddedBeforeAsyncCleanup_shouldNotStop`) to properly defer async execution so the scenario is actually reproduced.

### Integration tests (`SentryHangTrackerIntegrationTests.swift`)

The previous integration tests hung indefinitely because XCTest runs on the main thread — the same thread the hang tracker monitors. `Thread.sleep` inside `waitForHangIterative` blocked the test thread, preventing `beforeWaiting` from ever signaling the semaphore.

Rewrote them using a secondary RunLoop on a dedicated thread (each thread has its own RunLoop per [Apple's CFRunLoop docs](https://developer.apple.com/documentation/corefoundation/cfrunloop)). Everything is real — no mocks:

- Real `CFRunLoopObserverCreateWithHandler`
- Real `SentryDispatchQueueWrapper` (real serial GCD queue)
- Real `DispatchSemaphore`
- Real `SentryDefaultCurrentDateProvider`

The only injection redirects `addObserver`/`removeObserver` to the test thread's RunLoop instead of `CFRunLoopGetMain()`.

3 integration tests: hang detection, no-hang (fast completion), and finished observer callback.

philprime:
@sentry review
@cursor review

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
 - Transport logs error message for HTTP 413 responses (#7261)
 - Add `installGroupsOverride` parameter to `CheckForUpdateParams` and `installGroups` property to `ReleaseInfo` for SentryDistribution (#7278)
 
+### Improvements
```
itaybre: `h`: The changes are in an old release now

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationHangTracker.swift
```diff
@@ -0,0 +1,68 @@
...
+                    return
+                }
+
+                currentHangId = id
+                self.hangStarted()
+            }
+            let finished = tracker.addFinishedRunLoopObserver { [weak self] _ in
+                self?.hangStopped()
+            }
+            callbackId = (late, finished)
```
itaybre: `l`: This is a personal preference, but I can't say I love tuples (even if named)

Sources/Swift/Core/Integrations/ANR/SentryHangTracker.swift
```diff
@@ -0,0 +1,505 @@
+// swiftlint:disable file_length
+
+#if (os(iOS) || os(tvOS) || targetEnvironment(macCatalyst) || os(visionOS)) && !SENTRY_NO_UIKIT
```
itaybre: `h`: SENTRY_NO_UIKIT was renamed to SENTRY_NO_UI_FRAMEWORK

Sources/Swift/Core/Integrations/ANR/SentryHangTracker.swift
```diff
@@ -0,0 +1,505 @@
...
+
+@_implementationOnly import _SentryPrivate
+#if canImport(UIKit) && !SENTRY_NO_UIKIT
```
itaybre: Same here

Sources/Swift/Core/Integrations/ANR/SentryHangTracker.swift
```diff
@@ -0,0 +1,505 @@
...
+        self.createObserver = createObserver
+        self.addObserver = addObserver
+        self.removeObserver = removeObserver
+        self.createSemaphore = createSemaphore
+
+        // Derive the target frames-per-second ratio from the active scene's key window with a ratio used to correct slow frames
+        //
+        // Platform-specific initialization:
+        // - On UIKit platforms (iOS, tvOS): Query the key window's maximum FPS to adapt
+        //   to ProMotion displays (120Hz) or standard displays (60Hz)
```
itaybre, line 197: `l`: Out of curiosity, does this handle automatically dynamic frame rate? 
I think iOS supports that

Sources/Swift/Core/Integrations/ANR/SentryHangTracker.swift
```diff
@@ -0,0 +1,505 @@
...
+        // - On non-UIKit platforms (macOS): Default to 60 FPS
+        //
+        // The hang threshold is calculated as: (1.0 / maxFPS) * 1.5
+        // This means a hang is detected if a run loop iteration takes longer than 1.5 frame durations.
+        // For example: 60 FPS = 16.67ms per frame, so hang threshold = 25ms
+        #if (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT
```
itaybre:
```suggestion
        #if (os(iOS) || os(tvOS)) && !SENTRY_NO_UI_FRAMEWORK
```

Sources/Swift/Core/Integrations/ANR/SentryHangTracker.swift
itaybre:
```suggestion
#if canImport(UIKit) && !SENTRY_NO_UI_FRAMEWORK
```

Sources/Swift/Core/Integrations/ANR/SentryHangTracker.swift
itaybre:
```suggestion
#if (os(iOS) || os(tvOS) || os(visionOS)) && !SENTRY_NO_UI_FRAMEWORK
```

Also `targetEnvironment(macCatalyst)` is a subset of `os(iOS)`

Sources/Swift/Core/Integrations/ANR/SentryHangTracker.swift
```diff
@@ -0,0 +1,505 @@
...
+        let application = applicationProvider.application()
+        let windows = application?.connectedScenes.flatMap { ($0 as? UIWindowScene)?.windows ?? [] } ?? []
```
itaybre:
```suggestion
        let windows = application?.connectedScenes.compactMap { $0 as? UIWindowScene }.flatMap { $0.windows } ?? []
```
`l` very low priority, but this feels more readable

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationHangTracker.swift
```diff
@@ -0,0 +1,68 @@
...
+
+final class SentryWatchdogTerminationHangTracker {
+    private let queue: SentryDispatchQueueWrapperProtocol
+    private let tracker: SentryHangTracker
+    private let timeoutInterval: TimeInterval
+    private let hangStarted: () -> Void
+    private let hangStopped: () -> Void
+
+    private var callbackId: (late: UUID, finishedRunLoop: UUID)?
+    private var finishedRunLoopId: UUID?
```
itaybre: `m`: I don't see it being used
```suggestion
```

Tests/SentryTests/Integrations/ANR/SentryHangTrackerIntegrationTests.swift
```diff
@@ -0,0 +1,252 @@
...
+        CFRunLoopStop(runLoop)
+        thread?.cancel()
+        _ = runLoopStopped.wait(timeout: .now() + 2.0)
+    }
+    
+    /// Returns the RunLoop for this thread.
+    /// Must be called after `start()`.
+    func getRunLoop() -> CFRunLoop {
+        guard let runLoop = runLoop else {
+            fatalError("RunLoop not ready. Call start() first.")
```
itaybre: `h`: Even tough it shouldn't happen, I don't believe we should have a fatalError in production

philprime: This is a test file, I can probably replace it with a XCTUnwrap

Sources/Swift/Core/Integrations/ANR/SentryHangTracker.swift
philprime: At this point it does not, because the interval is hard set on init

philprime:
@itaybre one thing I am unsure is if we want this to be behind a new config flag an mark it as experimental. What do you think?

itaybre:
> @itaybre one thing I am unsure is if we want this to be behind a new config flag an mark it as experimental. What do you think?

It is a big change, it refactors how we track hangs, releasing it with an experimental flag is a good idea IMO

philprime:
Moving this back to draft as @noahsmartin is going to take a look at the potential race-conditions now that the complexity of the implementation increased

philprime:
Superseded by https://github.com/getsentry/sentry-cocoa/pull/7464

PR #6242, opened 2025-09-23
test: Skip AppHangTracking when debugger attached

### LINKED ISSUE #6243: test: Skip AppHangTracking when debugger attached
> [!NOTE]
> The pull request "[test: Skip AppHangTracking when debugger attached](https://github.com/getsentry/sentry-cocoa/pull/6242)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Skip the testResumeAndPauseAppHangTracking when a debugger is attached to avoid confusion.

#skip-changelog

philipphofmann:
Skip the testResumeAndPauseAppHangTracking when a debugger is attached to avoid confusion.

#skip-changelog 

Closes #6243


Tests/SentryTests/SentrySDKInternalTests.swift
```diff
@@ -609,6 +609,10 @@ class SentrySDKInternalTests: XCTestCase {
 
     @available(*, deprecated, message: "This is deprecated because SentryOptions integrations is deprecated")
     func testResumeAndPauseAppHangTracking() throws {
+        if SentryDependencyContainer.sharedInstance().crashWrapper.isBeingTraced {
+            throw XCTSkip("This test only works when the debugger is NOR attached, because it requires the SentryANRTrackingIntegration being installed, which the SDK only installs if the debugger is not attached.")
```
philprime: `l`: Typo?

```suggestion
            throw XCTSkip("This test only works when the debugger is NOT attached, because it requires the SentryANRTrackingIntegration being installed, which the SDK only installs if the debugger is not attached.")
```

PR #6264, opened 2025-09-24
fix: Lazily access UIApplication.shared

### LINKED ISSUE #6270: fix: Lazily access UIApplication.shared
> [!NOTE]
> The pull request "[fix: Lazily access UIApplication.shared](https://github.com/getsentry/sentry-cocoa/pull/6264)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Fixes null UIApplication

noahsmartin:
Fixes null UIApplication by accessing it lazily instead of storing it in an instance variable. Also updated the Swift nullability to properly reflect that this value is nullable.

This does not fix a related issue: https://github.com/getsentry/sentry-cocoa/issues/6275 Since that issue existed in the ObjC version I wanted to focus on getting parity with the old version first, then fixing that issue in a follow up.

Closes #6270

Fixes GH-6265


Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -61,7 +61,7 @@
 @interface SentryFileManager () <SentryFileManagerProtocol>
 @end
 
-@interface SentryDependencyContainer ()
+@interface SentryDependencyContainer () <SentryApplicationProvider>
```
philippsuess: `h`: We pass the SentryDependencyContainer as a reference to the references it holds below. This could easily lead to cyclic references and memory leaks, if I'm not mistaken.

What about creating an extra class for the `SentryApplicationProvider` ? The dependency container should only manage depenencies, IMO. I don't think it's a good idea to mix that type of logic for the `SentryApplicationProvider`.  Also, it's better for testability. We can swap out the `SentryApplicationProvider` in the container if we want.

Sources/Swift/Helper/ThreadSafeApplication.swift
```diff
@@ -4,11 +4,15 @@ import UIKit
...
     
-    @objc public init(initialState: UIApplication.State, notificationCenter: SentryNSNotificationCenterWrapper) {
+    @objc public init(applicationProvider: SentryApplicationProvider, notificationCenter: SentryNSNotificationCenterWrapper, dispatchQueue: SentryDispatchQueueWrapper) {
         self.notificationCenter = notificationCenter
-        _internalState = initialState
+        _internalState = .background
         super.init()
         
+        dispatchQueue.dispatchAsyncOnMainQueue { [weak self] in
+            self?._internalState = applicationProvider.application?.unsafeApplicationState ?? .background
```
philippsuess: `m`: I think it would be good to log an error or a warning here if the `applicationProvider.application` is `nil`. This will help us to investigate problems in the future.

Sources/Swift/Helper/ThreadSafeApplication.swift
```diff
@@ -4,11 +4,15 @@ import UIKit
...
     private let notificationCenter: SentryNSNotificationCenterWrapper
     
-    @objc public init(initialState: UIApplication.State, notificationCenter: SentryNSNotificationCenterWrapper) {
+    @objc public init(applicationProvider: SentryApplicationProvider, notificationCenter: SentryNSNotificationCenterWrapper, dispatchQueue: SentryDispatchQueueWrapper) {
         self.notificationCenter = notificationCenter
-        _internalState = initialState
+        _internalState = .background
         super.init()
         
+        dispatchQueue.dispatchAsyncOnMainQueue { [weak self] in
```
philippsuess: `m`: This is new logic. It would be great to add some unit tests for this, please.

Sources/Swift/Helper/ThreadSafeApplication.swift
```diff
@@ -4,11 +4,15 @@ import UIKit
 @objc @_spi(Private) public final class SentryThreadsafeApplication: NSObject {
     private let notificationCenter: SentryNSNotificationCenterWrapper
     
-    @objc public init(initialState: UIApplication.State, notificationCenter: SentryNSNotificationCenterWrapper) {
+    @objc public init(applicationProvider: SentryApplicationProvider, notificationCenter: SentryNSNotificationCenterWrapper, dispatchQueue: SentryDispatchQueueWrapper) {
         self.notificationCenter = notificationCenter
-        _internalState = initialState
+        _internalState = .background
```
philippsuess: `m`: I'm curious what drove you to choose `.background` as the default state.

Sources/Swift/Helper/ThreadSafeApplication.swift
```diff
@@ -4,11 +4,15 @@ import UIKit
...
         super.init()
         
+        dispatchQueue.dispatchAsyncOnMainQueue { [weak self] in
+            self?._internalState = applicationProvider.application?.unsafeApplicationState ?? .background
+        }
```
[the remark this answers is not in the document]
philippsuess: Yep, I think this is true. We use locks for writing to the `_internalState` everywhere in this class except here.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Lazily access UIApplication.shared to fix null application in SwiftUI apps (#6264)
```
philippsuess: `m`: I think we don't need to mention the lazy access here. Just the crash(es) we're fixing. That's what users need to know.

Sources/Swift/Helper/ThreadSafeApplication.swift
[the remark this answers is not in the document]
noahsmartin: I was trying to copy the behavior exactly from the objc version: https://github.com/getsentry/sentry-cocoa/pull/5900/files#diff-43142e7f18c96843ccf8ebac84d1e04ab4ca45ead281010af2d39baf06dac93cL42 But the whole point of that previous PR was to fix this race condition that existed in ObjC, so yes I should fix it here too

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -189,6 +187,15 @@
...
 }
 
+- (nullable id<SentryApplication>)application
+{
+#if SENTRY_HAS_UIKIT
+    return UIApplication.sharedApplication;
+#elif TARGET_OS_OSX
+    return NSApplication.sharedApplication;
+#endif
+}
```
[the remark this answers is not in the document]
noahsmartin: I don't think this is right, what would happen is it would fail to compile if we ever tried to compile for a new platform. Maybe we want it to actually compile, but it isn't undefined behavior

Sources/Swift/Helper/ThreadSafeApplication.swift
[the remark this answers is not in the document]
noahsmartin: This is correct, and matches the old objc logic. It's kind of the whole point, we need to start out with this being a default value (in objc the default was `0` but swift is safer than that)

Sources/Swift/Helper/ThreadSafeApplication.swift
noahsmartin: It's restoring the logic from: https://github.com/getsentry/sentry-cocoa/pull/5900/files#diff-43142e7f18c96843ccf8ebac84d1e04ab4ca45ead281010af2d39baf06dac93cL42 not to say it shouldn't be tested, just FYI that it's not anything I came up with I just copied from before

Sources/Swift/Helper/ThreadSafeApplication.swift
noahsmartin: The intention is to match the objc version. I'll have to double check what that used

noahsmartin: The intention is to match the objc version. I'll have to double check what that used

noahsmartin: Turns out `0` is `active`

Sources/Sentry/SentryDependencyContainer.m
noahsmartin: I've seen this "provider" pattern is commonly for swift DI because it allows easily protocol composition. Sometimes it is combined with a scope that could help with the cycles, but I also think for something like this we never expect the dependency container, and any the dependencies it holds, to ever be deallocated.

That being said, I think that pattern is not really the focus of this PR and refactored it to not use it. I don't want to add another type, wrapper types always seem like code smells to me and unnecessarily bloat the app (hurts app size and perf) so I don't want to get into the habit of using it for things like this. I update it to a function, which is I think more in line with what we want anyways. It's not the most pretty in ObjC, but works, and makes a lot of sense for Swift IMO

Sources/Swift/Helper/ThreadSafeApplication.swift
noahsmartin: Realized it was doing nothing because internally that `dispatchAsyncOnMainQueue` actually called the closure right away since it was invoked on the main thread

Sources/Sentry/SentryDependencyContainer.m
philippsuess: > but I also think for something like this we never expect the dependency container, and any the dependencies it holds, to ever be deallocated.

When we close the SDK, we call reset of the dependency container, which creates a new instance of it. https://github.com/getsentry/sentry-cocoa/blob/4be5cd8ec957763bd50e7be61fd78c472434d318/Sources/Sentry/SentryDependencyContainer.m#L111-L135

So in that case, the old instance could stay in memory. I didn't test it, but it's likely.

> I don't want to add another type, wrapper types always seem like code smells to me and unnecessarily bloat the app (hurts app size and perf) so I don't want to get into the habit of using it for things like this.

Let's discuss this later and not in this PR. Our entire dependency injection / management strategy has significant room for improvement. Most important is to avoid memory leaks.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -189,6 +186,30 @@ - (instancetype)init
...
+#endif
+    return ^id<SentryApplication> _Nullable()
+    {
+#if SENTRY_HAS_UIKIT
+        return UIApplication.sharedApplication;
+#elif TARGET_OS_OSX
+        return NSApplication.sharedApplication;
+#endif
+        return nil;
+    };
```
philippsuess: `m`: I'm a bit concerned that somebody will add self to this block by accident and then we have a memory leak. It's so easy to overlook. What about moving this block before the dependency container class and declare it as static to prevent this:

```
// Declare the application provider block at the top level to prevent capturing 'self' 
// from the dependency container, which would create cyclic dependencies and memory leaks.
static id<SentryApplication> _Nullable (^defaultApplicationProvider)(void) = ^id<SentryApplication> _Nullable()
{
#if SENTRY_HAS_UIKIT
    return UIApplication.sharedApplication;
#elif TARGET_OS_OSX
    return NSApplication.sharedApplication;
#endif
    return nil;
};
```

Sources/Sentry/SentryDependencyContainer.m
philippsuess: I also prefer anonymous functions over extra wrapper classes, cause wrappers are so heavy. I'm still trying to undo all the Java / .NET mentality that I grabbed up in my first 8 years of coding. 

I did so many of JavaManagerOfManagersStrategyPatternWrapperOfWhateverClasses in the past 😅

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -189,6 +186,30 @@ - (instancetype)init
     return self;
 }
 
+- (id<SentryApplication> _Nullable (^)(void))applicationProvider
```
philippsuess: `m`: I think we could do a typedef for this
```
typedef id<SentryApplication> _Nullable (^SentryApplicationProviderBlock)(void);
```

Sources/Swift/Helper/ThreadSafeApplication.swift
```diff
@@ -4,11 +4,19 @@ import UIKit
...
-        _internalState = initialState
+        // This matches the ObjC behavior which did not initialize the state when the UIApplication was null
+        // so it kept a default value of 0 which happens to be defined to be `active`.
+        // Acquiring the lock is not necessary here since the instance has not been initialized yet.
+        if let application = applicationProvider() {
+            _internalState = application.unsafeApplicationState
+            SentrySDKLog.warning("Application is null in SentryThreadsafeApplication")
+        } else {
+            _internalState = .active
+        }
```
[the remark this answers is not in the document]
philippsuess: Yes, I think the log statement should be in the else block.

Sources/Swift/Helper/ThreadSafeApplication.swift
philippsuess: Not having test in the past isn't always a good reason to skip adding tests when we can. This topic has already caused problems, so thanks for adding some tests 💯

Sources/Swift/Helper/ThreadSafeApplication.swift
```diff
@@ -4,11 +4,19 @@ import UIKit
 @objc @_spi(Private) public final class SentryThreadsafeApplication: NSObject {
     private let notificationCenter: SentryNSNotificationCenterWrapper
     
-    @objc public init(initialState: UIApplication.State, notificationCenter: SentryNSNotificationCenterWrapper) {
+    @objc public init(applicationProvider: () -> SentryApplication?, notificationCenter: SentryNSNotificationCenterWrapper) {
         self.notificationCenter = notificationCenter
-        _internalState = initialState
+        // This matches the ObjC behavior which did not initialize the state when the UIApplication was null
+        // so it kept a default value of 0 which happens to be defined to be `active`.
+        // Acquiring the lock is not necessary here since the instance has not been initialized yet.
```
philippsuess, line 11: Ah thanks, we're in the initializer, so no locks needed. Thanks for the comment 🤦

philipphofmann (approved):
Thanks for tackling this @noahsmartin. LGTM with a few suggestions.

philippsuess:
FYI, @noahsmartin, I merged main back into this branch cause this should fix a few flaky unit tests.

Sources/Swift/Helper/ThreadSafeApplication.swift
noahsmartin: oh yeah agreed, I more meant that I thought it was already tested. But yes should all be covered now

Sources/Sentry/SentryDependencyContainer.m
noahsmartin: Works for me 👍

Sources/Sentry/SentryDependencyContainer.m
noahsmartin: I actually had exactly that, but I removed it because it was only ever used in this one place so I figured it was more concise to have inline. Will see how it looks with the move to a global function

noahsmartin: With the new global it doesn't look like there's any need for the typedef, but lmk if you think it would help somewhere

philippsuess: We can put it in the header file of the dependency container. Definitely not global. If that's tricky, we can also skip it.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -58,6 +58,19 @@
 #define SENTRY_THREAD_SANITIZER_DOUBLE_CHECKED_LOCK                                                \
     SENTRY_DISABLE_THREAD_SANITIZER("Double-checked locks produce false alarms.")
 
+typedef id<SentryApplication> _Nullable (^SentryApplicationProviderBlock)(void);
+// Declare the application provider block at the top level to prevent capturing 'self'
+// from the dependency container, which would create cyclic dependencies and memory leaks.
+SentryApplicationProviderBlock defaultApplicationProvider = ^id<SentryApplication> _Nullable()
```
philippsuess, line 64: Thanks 💯

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -189,6 +199,17 @@ - (instancetype)init
...
+- (nullable id<SentryApplication>)application
+{
+#if defined(SENTRY_TEST) || defined(SENTRY_TEST_CI)
+    id<SentryApplication> override = self.applicationOverride;
+    if (override) {
+        return override;
+    }
+#endif
+    return defaultApplicationProvider();
+}
```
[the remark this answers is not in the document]
philippsuess: The `SentrySessionTracker` won't crash; it will simply not start the session here

https://github.com/getsentry/sentry-cocoa/blob/4be5cd8ec957763bd50e7be61fd78c472434d318/Sources/Sentry/SentrySessionTracker.m#L96-L98

@noahsmartin, we could add a check for nil to the code above to clearly point out that this could be the case with a comment, but it doesn't change the functionality of the code. Not required, so I leave it up to you if you want to do that.

philipphofmann (approved):
Again, thanks.

Sources/Sentry/SentryDependencyContainer.m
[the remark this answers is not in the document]
philippsuess: We can follow up with a PR.

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -445,7 +467,7 @@ - (SentryGlobalEventProcessor *)globalEventProcessor SENTRY_THREAD_SANITIZER_DOU
 - (SentrySessionTracker *)getSessionTrackerWithOptions:(SentryOptions *)options
 {
     return [[SentrySessionTracker alloc] initWithOptions:options
-                                             application:self.application
+                                     applicationProvider:defaultApplicationProvider
```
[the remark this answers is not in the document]
philippsuess: That's OK, we don't want to use the test provider for these instances.

philippsuess:
UI tests are flaky because of https://github.com/getsentry/sentry-cocoa/issues/6291 and will be disabled with https://github.com/getsentry/sentry-cocoa/pull/6289.

PR #6268, opened 2025-09-24
test: Add SwiftUI test to reproduce SwiftUI lifecycle only bugs

### LINKED ISSUE #6269: test: Add SwiftUI test
> [!NOTE]
> The pull request "[test: Add SwiftUI test](https://github.com/getsentry/sentry-cocoa/pull/6268)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Add a test to capture error in SwiftUI test app

#skip-changelog

noahsmartin:
Add a test to capture error in SwiftUI test app, reproduces the error fixed by https://github.com/getsentry/sentry-cocoa/pull/6264

#skip-changelog

Closes #6269

---

> [!NOTE]
> Adds a UITest to assert `capture(error:)` returns a non-empty Sentry ID and updates the SwiftUI sample to surface that ID for testing.
> 
> - **UITests (`Samples/iOS-SwiftUI/iOS-SwiftUI-UITests/LaunchUITests.swift`)**:
>   - Import `Sentry`.
>   - Add `testCaptureErrorReturnsValidId` that taps "Capture Error", reads `errorId`, builds `SentryId`, and asserts it’s not empty.
> - **SwiftUI app (`Samples/iOS-SwiftUI/iOS-SwiftUI/ContentView.swift`)**:
>   - Add `@State var errorId: SentryId?` and display it with `accessibilityIdentifier` `errorId`.
>   - Change `captureErrorAction` to a function that captures the error and stores the returned `SentryId` in state.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit be6cc0e2a7f9d00e94751d1ca4fc01bf40ff3c58. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Samples/iOS-SwiftUI/iOS-SwiftUI-UITests/LaunchUITests.swift
+++ b/Samples/iOS-SwiftUI/iOS-SwiftUI-UITests/LaunchUITests.swift
@@ -1,3 +1,4 @@
+import Sentry
 import SentrySampleShared
 import SentrySampleUITestShared
 import XCTest
@@ -49,11 +50,13 @@ class LaunchUITests: XCTestCase {
         XCTAssertEqual(app.staticTexts["TTDInfo"].label, "TTID and TTFD found")
     }
   
-  func testSendsError() {
+  func testCaptureErrorReturnsValidId() {
     let app = newAppSession()
     app.safelyLaunch()
     app.buttons["Capture Error"].tap()
-    XCTAssertFalse(app.staticTexts["errorId"].label.isEmpty)
+    let errorId = app.staticTexts["errorId"].label
+    let sentryErrorId = SentryId(uuidString: errorId)
+    XCTAssertNotEqual(sentryErrorId.sentryIdString, SentryId.empty.sentryIdString, "The id should not be empty, since empty is used to indicate the capture did not work.")
   }
 
     func newAppSession() -> XCUIApplication {


Samples/iOS-SwiftUI/iOS-SwiftUI-UITests/LaunchUITests.swift
```diff
@@ -48,6 +48,13 @@ class LaunchUITests: XCTestCase {
 
         XCTAssertEqual(app.staticTexts["TTDInfo"].label, "TTID and TTFD found")
     }
+  
+  func testSendsError() {
```
philippsuess: `m`: This test doesn't really validate that we send an error. It's a smoke test that capturing error doesn't crash and the SDK returns an ID. It would be great to put that info in the test name.

Samples/iOS-SwiftUI/iOS-SwiftUI-UITests/LaunchUITests.swift
```diff
@@ -48,6 +48,13 @@ class LaunchUITests: XCTestCase {
...
+    let app = newAppSession()
+    app.safelyLaunch()
+    app.buttons["Capture Error"].tap()
+    XCTAssertFalse(app.staticTexts["errorId"].label.isEmpty)
```
philippsuess: `m`: We could assert that the ID is a valid UUID. Then at least we know it's not complete garbage. If the string is `foo-bar` the test still succeeds. Also the UUID shouldn't be `SentryId.empty`, because the hub and client return this when they can't send an event.

https://github.com/getsentry/sentry-cocoa/blob/d8ceea3a0ce99c0dd499d3b9472c87aaf0713d3c/Sources/Swift/Core/Protocol/SentryId.swift#L6-L9

philipphofmann (approved):
Thanks. It would be great to make the PR title a bit more specific and mention the related issues in the PR description please.

Samples/iOS-SwiftUI/iOS-SwiftUI-UITests/LaunchUITests.swift
```diff
@@ -48,6 +49,15 @@ class LaunchUITests: XCTestCase {
...
         XCTAssertEqual(app.staticTexts["TTDInfo"].label, "TTID and TTFD found")
     }
+  
+  func testCaptureErrorReturnsValidId() {
+    let app = newAppSession()
+    app.safelyLaunch()
+    app.buttons["Capture Error"].tap()
+    let errorId = app.staticTexts["errorId"].label
+    let sentryErrorId = SentryId(uuidString: errorId)
+    XCTAssertNotEqual(sentryErrorId.sentryIdString, SentryId.empty.sentryIdString)
```
philippsuess: `l`: I would add an assertion message to explain that we expect the ID to be anything but an empty ID, because the SDK returns an empty ID if the capture method didn't work.

PR #6271, opened 2025-09-24
ref: SentryPerformanceTracker in Swift

noahsmartin:
Converts this class to Swift by just wrapping the ObjC implementation 

#skip-changelog


Sources/Sentry/include/SentryPerformanceTrackerHelper.h
```diff
@@ -25,7 +25,7 @@ NS_ASSUME_NONNULL_BEGIN
  * @return The span id.
  */
 - (SentrySpanId *)startSpanWithName:(NSString *)name
-                         nameSource:(SentryTransactionNameSource)source
+                         nameSource:(NSInteger)source
```
itaybre, line 30: Do you mind adding that the source is expected to be a value of `SentryTransactionNameSource`?

philippsuess: And also maybe explain why we can't use it here in the comment please.

Sources/Swift/Integrations/Performance/UIViewController/SentryPerformanceTracker.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    /// A static instance of performance tracker.
+    @objc public static let shared = SentryPerformanceTracker()
+    
+    /// Starts a new span if no span is active, then bind it to the scope if no span is bound.
+    /// @note If there's an active span, starts a child of the active span.
+    /// @param name Span name.
+    /// @param source the transaction name source.
+    /// @param operation Span operation.
+    /// @return The span id.
```
philippsuess, line 19: `l`: I don't think we need these comments here and also for the `SentryPerformanceTrackerHelper.h`. Maybe we could remove them at one place, so we don't have to maintain both.

philipphofmann (approved):
Thank you, @noahsmartin

noahsmartin:
I don't think this will be necessary for the dependency container to be in Swift, so going to close it for now

PR #6273, opened 2025-09-24
ref: Create thread inspector Swift protocol

noahsmartin:
This removes `SentryThreadInspector` from the dependency container public interface, in favor of a protocol defined in Swift. This means we don't need to convert the entire thing to Swift before the dependency container

Resolves https://linear.app/getsentry/issue/COCOA-608/convert-sentrycrashstackentrymapper-to-swift and https://linear.app/getsentry/issue/COCOA-708/convert-sentrythreadinspector and https://linear.app/getsentry/issue/COCOA-707/convert-sentrystacktracebuilder

#skip-changelog

---

> [!NOTE]
> Replaces public SentryThreadInspector exposure with Swift protocols and a SentryDefaultThreadInspector implementation, updating dependencies, integrations, and tests accordingly.
> 
> - **Thread Inspection**:
>   - Add Swift protocols `SentryThreadInspector` and `SentryThreadInspecting` for thread inspection.
>   - Rename Obj-C class `SentryThreadInspector` → `SentryDefaultThreadInspector` and migrate call sites.
> - **Core/DI**:
>   - `SentryDependencyContainer`: expose `threadInspector` as `id<SentryThreadInspector>` and lazily instantiate `SentryDefaultThreadInspector`.
>   - `SentryClient`: ctor now takes/creates `SentryDefaultThreadInspector` and stores it; uses it for attaching threads/stacktraces.
> - **Integrations**:
>   - Update ANR, Core Data, File I/O, Network tracking to depend on protocol or `SentryDefaultThreadInspector` and use `getCurrentThreads*`/`stacktraceForCurrentThreadAsyncUnsafe`.
> - **Swift/Obj-C bridges & Tests**:
>   - Add new Swift files: `SentryThreadInspector.swift`, `SentryThreadInspecting.swift`.
>   - Rename tests to `SentryDefaultThreadInspectorTests`; adapt helpers (`TestThreadInspector`) and bridging headers/imports.
> - **Misc**:
>   - `SentryProfileTimeseries.h`: switch UIKit import to `SentryProfilerDefines`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit da10f3bbe9cb5bbf860cbe3110ae5d3cd56a1029. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


Sources/Sentry/include/HybridPublic/SentryDependencyContainer.h
```diff
@@ -108,7 +108,7 @@ SENTRY_NO_INIT
 
 @property (nonatomic, strong, nullable) SentryFileManager *fileManager;
 @property (nonatomic, strong) SentryAppStateManager *appStateManager;
-@property (nonatomic, strong) SentryThreadInspector *threadInspector;
+- (id<SentryThreadInspecting>)threadInspecting;
```
philippsuess: `m`: Why don't we keep this as a property? We can do
```suggestion
@property (nonatomic, strong) id<SentryThreadInspecting> threadInspecting;
```

Sources/Swift/SentryCrash/SentryThreadInspecting.swift
```diff
@@ -0,0 +1,4 @@
+@_spi(Private) @objc public protocol SentryThreadInspecting {
```
philippsuess, line 1: `m`: This protocol name is a bit confusing to me. Now we have `SentryThreadInspecting` and `SentryThreadInspector`. I would rather name the protocol `SentryThreadInspector` and rename the objc `SentryThreadInspector` to somehting like `SentryDefaultThreadInspector`.

philipphofmann (commented):
I'm a bit confused. Can't the `SentryThreadInspector` implement the `SentryThreadInspecting` protocol? Then, the `SentryDependencyContainer` doesn't need to have a private property `@property (nonatomic, strong) SentryThreadInspector *threadInspector;`. We can have one public property for the new interface `SentryThreadInspecting` and it whould work.

Sources/Swift/SentryCrash/SentryThreadInspecting.swift
noahsmartin: Sure I can change the names

noahsmartin: FWIW I picked this based on the Swift API design guidelines: 
https://www.swift.org/documentation/api-design-guidelines/

```
Protocols that describe a capability should be named using the suffixes able, ible, or ing (e.g. Equatable, ProgressReporting).
```

Sources/Sentry/include/HybridPublic/SentryDependencyContainer.h
noahsmartin: I had done it this way just to reduce the amount of changes in the PR, but I changed it to be like this and it's not that much more, updated!

noahsmartin:
Thanks @philipphofmann just made those changes

PR #6274, opened 2025-09-24
ref: Create file io tracker Swift interface

noahsmartin:
Resolves https://linear.app/getsentry/issue/COCOA-709/convert-sentryfileiotracker

#skip-changelog


noahsmartin:
I'm not sure why this only fails to compile for static builds, converting it to draft while I investigate

Sources/Sentry/include/SentryFileIOTracker.h
```diff
@@ -4,6 +4,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @protocol SentryProcessInfoSource;
 @protocol SentryThreadInspector;
+@protocol SentryFileIOTracking;
```
itaybre: Where is this being used?

noahsmartin: Was supposed to be used below, just updated

itaybre: I was wondering if that was supposed to be using the protocol.

Sources/Swift/Helper/Dependencies.swift
```diff
@@ -8,7 +8,7 @@
 #if !os(watchOS) && !os(macOS) && !SENTRY_NO_UIKIT
     @objc public static let uiDeviceWrapper = SentryDefaultUIDeviceWrapper(queueWrapper: Dependencies.dispatchQueueWrapper)
 #endif // !os(watchOS) && !os(macOS) && !SENTRY_NO_UIKIT
-
-    @objc public static var threadInspector: SentryThreadInspector = SentryThreadInspector()
+    @objc public static var threadInspector = SentryThreadInspector()
+    @objc public static var fileIOTracker = SentryFileIOTracker(threadInspector: threadInspector, processInfoWrapper: processInfoWrapper)
```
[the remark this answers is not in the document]
noahsmartin: This is the same behavior as before

PR #6289, opened 2025-09-26
test: Skip flaky user feedback UITests

philipphofmann:
Disable the user feedback tests that assert the latest envelope content, as this approach is unreliable. The assertion relies on retrieving the latest envelope, which the transport often sends, causing it to be missing and resulting in a test failure.

#skip-changelog


itaybre (approved):
LGTM

Issue #5155, opened 2025-04-30
Investigate and improve performance on high-end devices


philprime:
### Description

We need to investigate potential reasons for performance drops even with the `experimentalViewRenderer` enabled. This also seems to visible on high-end devices such as an iPhone 14 Pro or iPhone 16 Pro Max.

Project to reproduce: https://github.com/AlirezaHadjar/sentry-session-replay-bug

Based on the report of [this comment](https://github.com/getsentry/sentry-cocoa/issues/4000#issuecomment-2839020095) in #4000.

Potential ideas to start off:
- Use instruments and CPU profiling to find slow methods
- Find alternative render methods to the current view renderer (i.e. finding a way to get data directly from GPU).

philprime:
Potential way of improving could be replacing the CADisplayLink and instead use a run loop observer.

https://developer.apple.com/documentation/corefoundation/cfrunloopobserver

pachuho:
We ran into the Session Replay scroll cost in a production IoT app and profiled it with Instruments on 9.24.0. We've since turned Replay off in production because of the scroll impact, but since this issue is about frame-capture performance specifically, here's a real-app A/B in case it's a useful complement to the reproducer already in the thread.

Short version: it's still very visible on 9.24.0, with the default view renderer and `quality = .low`. The run-loop scheduler change in 9.19.0 (#7851) helped but didn't remove the scroll-time cost.

Setup:
- sentry-cocoa 9.24.0, iPhone 16 / iOS 18.6.2 (so not a low-end-device effect)
- Release build; we only start Replay in Release
- `sessionReplay.onErrorSampleRate = 1.0` vs `0.0`, `maskAllImages = true`, `maskAllText = false`, `quality = .low`, default renderer — everything else identical
- `onErrorSampleRate` runs in buffer mode, so capture is continuous even when no error fires, which is the scroll-time cost shown below
- 40 s of continuous scrolling on a list-heavy screen, "Animation Hitches" template attached to the running process

| Metric | Replay ON | Replay OFF |
|---|---|---|
[...]

PR #6292, opened 2025-09-26
fix(session-replay): Extend masking and focus masking on sensitive information

philprime:
## :scroll: Description

This PR fixes several session replay masking issues and adds comprehensive test coverage for the redaction builder to prevent regressions.

### Fixed Masking Issues

**SwiftUI.List Background Clipping Issue (#6292)**
One screen in the dogfooding app [Flinky](https://github.com/techprimate/Flinky) had significant masking problems where the entire top half of a screen with SwiftUI.List was unmasked. Analysis revealed that `_UICollectionViewListLayoutSectionBackgroundColorDecorationView` has an extremely large frame (`-20 -1135.33; 442 2336`) that extends far beyond the visible list bounds, causing incorrect clipping calculations.

### Implementation Improvements

**String-Based Class Comparison**
- Introduced `ExtendedClassIdentifier` to compare view classes by string description instead of direct class references
- Prevents triggering Objective-C `+initialize` on private UIKit classes when running off the main thread
- Supports optional layer class filtering for multi-purpose views (e.g., `SwiftUI._UIGraphicsView`)

**Rendering & Layer Improvements**
- Fixed sublayer rendering order by sorting layers by `zPosition` (with insertion order as tie-breaker)
- Switched from view-based to layer-based traversal for accurate presentation state
- Properly handle presentation layers during animations vs model layers

**Edge Case Handling**
- UISwitch subtree ignoring to prevent internal UIImageView from being incorrectly redacted
- UITextField placeholder handling via `UITextFieldLabel` detection
- SwiftUI.List background decoration view treated as special case (direct redact, no clip-out)
- Axis-aligned transform detection to optimize opaque view clipping
- Custom anchor point support in transform calculations

### Test Coverage

Added 49 new tests to systematically verify:
- ExtendedClassIdentifier matching with layer filtering
- UIImageView boundary conditions (nil, 9x9, 10x10, 11x11 pixels)
- Hidden views and zero opacity handling
- Transform calculations with custom anchor points
- Region ordering and SwiftUI region prioritization
- Nested clipping regions
- Class hierarchy inheritance patterns
- iOS version-specific behavior (iOS 26 UISlider visual element)

**Note:** SwiftUI-specific integration tests (Text, Image, List, Label, Button) will be added in a follow-up PR.

### Manual Testing
- Verified fix in Flinky app - SwiftUI.List screens now mask correctly
- Tested across iOS 15.5, 16.4, 17.5, 18.6, and 26.0 simulators

### Automated Testing
- 49 new test cases added
- All 118+ tests passing (0 failures)
- Tests follow Arrange-Act-Assert pattern
- Standardized snapshot naming (`masked`/`unmasked`)

### Files Changed
- `Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift` - Core fixes
- `Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+EdgeCases.swift` - New (32 tests)
- `Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+Common.swift` - Enhanced (8 new tests)
- `Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+UIKit.swift` - Enhanced (5 new tests, removed 2 duplicates)
- `Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+SpecialViews.swift` - Updated naming

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
+++ b/Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
@@ -501,7 +501,7 @@ final class SentryUIRedactBuilder {
                         size: layer.bounds.size,
                         transform: newTransform,
                         type: .clipOut,
-                        name: layer.debugDescription
+                        name: view.debugDescription
                     ))
                 }
             }


Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
...
+    // MARK: - Types
+
+    /// Type used to represented a view that needs to be redacted
+    struct ExtendedClassIdentifier: Hashable {
+        /// String representation of the class
+        ///
+        /// We deliberately store class identities as strings (e.g. "SwiftUI._UIGraphicsView")
+        /// instead of `AnyClass` to avoid triggering Objective‑C `+initialize` on UIKit internals
+        /// or private classes when running off the main thread. The string is obtained via
+        /// `type(of: someObject).description()`.
```
philippsuess, line 22: Thanks a lot for this valuable comment.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
...
+        ///
+        /// Some view types are reused for multiple purposes. For example, `SwiftUI._UIGraphicsView`
+        /// is used both as a structural background (should not be redacted) and as a drawing surface
+        /// for images when paired with `SwiftUI.ImageLayer` (should be redacted). When `layerId` is
+        /// provided we only match a view if its backing layer’s type description equals the filter.
+        let layerId: String?
+
+        /// Initializes a new instance of the extended class identifier using a class ID.
+        ///
+        /// - parameter classId: The class name.
```
philippsuess, line 35: `l`: When the classId is the class name most of the time, maybe we should rename the classId to className, the same applies for layerId.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
 #endif
 
 final class SentryUIRedactBuilder {
+    // MARK: - Types
+
+    /// Type used to represented a view that needs to be redacted
+    struct ExtendedClassIdentifier: Hashable {
```
philippsuess: `l`: This is an inner struct, so maybe we can ditch the `extended` to make the code a bit shorter. No strong opinion here.
```suggestion
    struct ClassIdentifier: Hashable {
```

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
...
+
+            // Used by React Native to render long text
+            redactClasses.insert(ExtendedClassIdentifier(classId: "RCTParagraphComponentView"))
+
+            // Used by SwiftUI to render text without UIKit, e.g. `Text("Hello World")`.
+            // We include the class name without a layer filter because it is specifically
+            // used to draw text glyphs in this context.
+            redactClasses.insert(ExtendedClassIdentifier(classId: "SwiftUI.CGDrawingView"))
+
+            // Used to render SwiftUI.Text on iOS versions prior to iOS 18
```
philippsuess, line 144: Thanks for the comments 💯

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
...
+        // region for the decoration view instead of clip-outs.
+        redactClasses.insert(ExtendedClassIdentifier(classId: "_UICollectionViewListLayoutSectionBackgroundColorDecorationView"))
+
+        ignoreClassesIdentifiers = [
+            ExtendedClassIdentifier(objcType: UISlider.self),
+            ExtendedClassIdentifier(objcType: UISwitch.self)
+        ]
 #else
         ignoreClassesIdentifiers = []
 #endif
```
philippsuess, line 200: `l`: Would be great to briefly explain here with a comment why we ignore these and why they are empty on non iOS.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
...
+        redactClassesIdentifiers = redactClasses
+
         for type in options.unmaskedViewClasses {
-            self.ignoreClassesIdentifiers.insert(ObjectIdentifier(type))
+            self.ignoreClassesIdentifiers.insert(ExtendedClassIdentifier(class: type))
         }
         
         for type in options.maskedViewClasses {
-            self.redactClassesIdentifiers.insert(ObjectIdentifier(type))
+            self.redactClassesIdentifiers.insert(ExtendedClassIdentifier(class: type))
```
philippsuess: `l`: Nothing to do with this PR, but as we already touch it we don't use self anywhere above, so we can make this consistent.
```suggestion
            ignoreClassesIdentifiers.insert(ExtendedClassIdentifier(class: type))
}

for type in options.maskedViewClasses {
            redactClassesIdentifiers.insert(ExtendedClassIdentifier(class: type))
```

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
...
+    /// Returns `true` if the provided class type is contained in the ignore list.
+    ///
+    /// This compares by string description to avoid touching Objective‑C class objects directly.
     func containsIgnoreClass(_ ignoreClass: AnyClass) -> Bool {
-        return ignoreClassesIdentifiers.contains(ObjectIdentifier(ignoreClass))
+        /// Edge case: ``UITextField`` uses an internal type of ``UITextFieldLabel`` for the placeholder, which should also be ignored
+        if ignoreClass.description() == "UITextFieldLabel" {
+            return ignoreClassesIdentifiers.contains(where: { $0.classId == "UITextField" })
+        }
+        return ignoreClassesIdentifiers.contains(where: { $0.classId == ignoreClass.description() })
```
philippsuess: `m`: Can't we do this instead? It would make this O(1) instead of O(n).
```suggestion
        return ignoreClassesIdentifiers.contains(ExtendedClassIdentifier(class: ignoreClass))
```

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
...
+    ///   `containsRedactClass(viewClass: MyTitleLabel.self, layerClass: CALayer.self) == true`.
+    /// - SwiftUI image drawing: `viewClass == SwiftUI._UIGraphicsView` and `layerClass == SwiftUI.ImageLayer`
+    ///   will match because we register `("SwiftUI._UIGraphicsView", layerId: "SwiftUI.ImageLayer")`.
+    /// - SwiftUI structural background: `viewClass == SwiftUI._UIGraphicsView` with a generic `CALayer`
+    ///   will NOT match (no `ImageLayer`), so we don’t redact background fills.
+    /// - `UIImageView` will match the class rule; the final decision is refined by `shouldRedact(imageView:)`.
+    func containsRedactClass(viewClass: AnyClass, layerClass: AnyClass) -> Bool {
+        var currentClass: AnyClass? = viewClass
+        while let iteratorClass = currentClass {
+            if redactClassesIdentifiers.contains(where: { $0.matches(viewClass: iteratorClass, layerClass: layerClass) }) {
```
philippsuess: `m`: This is again O(n). Maybe we can somehow change this using an instance of `ExtendedClassIdentifier`, so it becomes O(1).

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
...
         
         for type in options.maskedViewClasses {
-            self.redactClassesIdentifiers.insert(ObjectIdentifier(type))
+            self.redactClassesIdentifiers.insert(ExtendedClassIdentifier(class: type))
         }
     }
     
+    /// Returns `true` if the provided class type is contained in the ignore list.
+    ///
+    /// This compares by string description to avoid touching Objective‑C class objects directly.
```
philippsuess, line 251: `m`: I'm curious why we don't need to check superclasses here as we do in `containsRedactClass`.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -9,138 +10,270 @@ import WebKit
...
+    /// - A custom label `class MyTitleLabel: UILabel {}` will match because `UILabel` is in the redact set:
+    ///   `containsRedactClass(viewClass: MyTitleLabel.self, layerClass: CALayer.self) == true`.
+    /// - SwiftUI image drawing: `viewClass == SwiftUI._UIGraphicsView` and `layerClass == SwiftUI.ImageLayer`
+    ///   will match because we register `("SwiftUI._UIGraphicsView", layerId: "SwiftUI.ImageLayer")`.
+    /// - SwiftUI structural background: `viewClass == SwiftUI._UIGraphicsView` with a generic `CALayer`
+    ///   will NOT match (no `ImageLayer`), so we don’t redact background fills.
+    /// - `UIImageView` will match the class rule; the final decision is refined by `shouldRedact(imageView:)`.
+    func containsRedactClass(viewClass: AnyClass, layerClass: AnyClass) -> Bool {
+        var currentClass: AnyClass? = viewClass
+        while let iteratorClass = currentClass {
```
philippsuess, line 287: `m`: An idea for speeding this up maybe for later: When there is a class that we should redact and it's not in the `redactClassesIdentifiers` we could add it to the `redactClassesIdentifiers`, so we don't have to traverse the classes up every time. I'm not sure if it's worth the effort, and I'm also unsure how often this code gets called.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -214,100 +349,141 @@ final class SentryUIRedactBuilder {
...
     
+    /// Special handling for `UIImageView` to avoid masking tiny gradient strips and
+    /// bundle‑provided assets (e.g. SF Symbols or app assets), which are unlikely to contain PII.
     private func shouldRedact(imageView: UIImageView) -> Bool {
-        // Checking the size is to avoid redact gradient background that
-        // are usually small lines repeating
-        guard let image = imageView.image, image.size.width > 10 && image.size.height > 10  else { return false }
+        // Checking the size is to avoid redacting gradient backgrounds that are usually
+        // implemented as very thin repeating images.
+        guard let image = imageView.image, image.size.width > 10 && image.size.height > 10  else {
```
philippsuess, line 447: `l`: If you have any clue why we use `10` for the size, please add the info as a comment.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -214,100 +349,141 @@ final class SentryUIRedactBuilder {
...
                 ))
+
+                guard !view.clipsToBounds else {
+                    return
+                }
+                enforceRedact = true
+            } else if isOpaque(view) {
+                let finalViewFrame = CGRect(origin: .zero, size: layer.bounds.size).applying(newTransform)
+                if isAxisAligned(newTransform) && finalViewFrame == rootFrame {
+                    //Because the current view is covering everything we found so far we can clear `redacting` list
```
philippsuess: `l`:
```suggestion
                    // Because the current view is covering everything we found so far we can clear `redacting` list
```

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -330,20 +506,31 @@ final class SentryUIRedactBuilder {
...
         }
+
+        #if os(iOS)
+        // UISwitch uses UIImageView internally, which can be in the list of redacted views.
+        // But UISwitch is in the list of ignored class identifiers by default, because it uses
+        // non-sensitive images. Therefore we want to ignore the subtree of UISwitch, unless
+        // it was removed from the list of ignored classes
+        if viewTypeId == UISwitch.self.description(), ignoreClassesIdentifiers.contains(
+            where: { $0.classId == UISwitch.self.description()
+            }) {
```
philippsuess: `m`: We should be able to use `ExtendedClassIdentifier(classId: UISwitch.self.description())` to make this O(1), if I'm not mistaken.

Sources/Swift/Core/Tools/ViewCapture/SentryRedactRegion.swift
```diff
@@ -22,6 +22,25 @@ final class SentryRedactRegion {
     func canReplace(as other: SentryRedactRegion) -> Bool {
         size == other.size && transform == other.transform && type == other.type
     }
+
+    static func == (lhs: SentryRedactRegion, rhs: SentryRedactRegion) -> Bool {
```
philippsuess: `m`: Can't we turn this class into a struct? Then we don't need to do this manually.

CHANGELOG.md
```diff
@@ -24,6 +24,15 @@
 - Fix rendering method for fast view rendering (#6360)
 - Fix issue where the thread that generated an event could be missing when more than 100 threads are running (#6377)
 - Fix wrong Frame Delay when becoming active, which lead to false reported app hangs when the app moves to the foreground after being in the background (#6381)
+- Session Replay masking improvements (#6292)
```
philippsuess, line 37: Thanks for the detailed changelog entry

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+SpecialViews.swift
```diff
@@ -0,0 +1,404 @@
...
+        //    |    |    |    | <UIView: 0x105c3cf60; frame = (0 0; 0 0); backgroundColor = UIExtendedGrayColorSpace 0 0.5; layer = <CALayer: 0x600000cf1770>>
+    }
+
+    func testRedact_withWKWebView_withMaskingEnabled_shouldRedactView() throws {
+        // -- Arrange --
+        setupWKWebViewFixture()
+
+        // -- Act --
+        let sut = getSut(maskAllText: true, maskAllImages: true)
+        let result = sut.redactRegionsFor(view: rootView)
```
philippsuess: `l`: I would name this `redactedRegions`. Result doesn't really say what it is.

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+SpecialViews.swift
```diff
@@ -0,0 +1,404 @@
...
+        //    |    |    | <WKContentView: 0x121822800; frame = (0 0; 40 40); anchorPoint = (0, 0); layer = <CALayer: 0x600000ce8750>>
+        //    |    |    |    | <UIView: 0x121012580; frame = (0 0; 0 0); anchorPoint = (0, 0); clipsToBounds = YES; layer = <CALayer: 0x600000cf8960>>
+        //    |    |    |    |    | <UIView: 0x1210123e0; frame = (0 0; 0 0); autoresize = W+H; layer = <CALayer: 0x600000cf8930>>
+        //    |    |    |    | <WKVisibilityPropagationView: 0x105c21580; frame = (0 0; 0 0); layer = <CALayer: 0x600000cf01b0>>
+        //    |    |    | <_UIScrollViewScrollIndicator: 0x105c33170; frame = (34 30; 3 7); alpha = 0; autoresize = LM; layer = <CALayer: 0x600000cf0c30>>
+        //    |    |    |    | <UIView: 0x105c2b5f0; frame = (0 0; 0 0); backgroundColor = UIExtendedGrayColorSpace 0 0.5; layer = <CALayer: 0x600000cf1830>>
+        //    |    |    | <_UIScrollViewScrollIndicator: 0x105c3eb90; frame = (30 34; 7 3); alpha = 0; autoresize = TM; layer = <CALayer: 0x600000cf1620>>
+        //    |    |    |    | <UIView: 0x105c3cf60; frame = (0 0; 0 0); backgroundColor = UIExtendedGrayColorSpace 0 0.5; layer = <CALayer: 0x600000cf1770>>
```
philippsuess: `l`: How do you get this `View Hierarchy` info here? What do I have to do if I want to update this?

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+SpecialViews.swift
```diff
@@ -0,0 +1,404 @@
...
+    private func getSut(maskAllText: Bool, maskAllImages: Bool) -> SentryUIRedactBuilder {
+        return SentryUIRedactBuilder(options: TestRedactOptions(
+            maskAllText: maskAllText,
+            maskAllImages: maskAllImages
+        ))
+    }
+
+    override func setUp() {
+        rootView = UIView(frame: CGRect(x: 0, y: 0, width: 100, height: 100))
+    }
```
philippsuess: `l`: I like to avoid global test variables to minimize side effects or shared state for the tests. I know we have them all over our test code, but I don't really like it. What about having a helper function `getRootView`, removing the `rootView` and let every test have its own local variable of `rootView`?

philipphofmann (commented):
Thanks, this looks way better than before. I gave the tests just a quick pass; they look good, but I found a couple of minor things to improve. Thanks @philprime for cleaning this up.

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+SpecialViews.swift
philprime: The original test used a global root view, I also don't like it, will remove it.

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+SpecialViews.swift
philprime: I had a note at the top of class, but extended it now with a step-by-step guide in `SentryUIRedactBuilderTests.swift`

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I don't know why they are ignored, it's not documented. I'll a comment explaining that.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I don't know where it comes from. I added a comment explaining it's an undocumented magic number

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: That's a fair question and I do not fully know why, this logic was not written by me.
But I could argue that for redaction we want to be more selective when ignoring redaction, so we don't accidentally unmask all children too.

Example:

class MyLabel: UILabel {}
class SuperSensitiveLabel: UILabel {}

if we ignore `UILabel` it would also expose `MyLabel` and `SuperSensitiveLabel`, which might not be what the user wants.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I don't think we can do this because `.contains(ExtendedClassIdentifier(class: ignoreClass))` uses the hash of the class identifier which includes the layerId too, but my current logic only checks if the class matches and ignores the layer. This applies to all of the O(n) to O(1) suggestions. One approach could be keeping additional arrays/sets only for class comparison. That would allow O(1) but require more memory. WDYT?

philippsuess: I don't know how often this code gets actually called. We don't have that many elements in these data structures, so it seems to me that changing the data structures and potentially slightly increasing the memory footprint could be worth it. Or we're just trying to fix something that's not an issue. I let you make the call, cause you know the code better than I do.

Sources/Swift/Core/Tools/ViewCapture/SentryRedactRegion.swift
```diff
@@ -4,7 +4,7 @@ import Foundation
 import ObjectiveC.NSObjCRuntime
 import UIKit
 
-final class SentryRedactRegion: Equatable {
+struct SentryRedactRegion: Equatable {
```
philippsuess, line 7: Yes 👍

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philippsuess: Yes, that makes sense. Maybe worth double checking with Roman.

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests.swift
```diff
@@ -6,6 +6,23 @@ import SwiftUI
 import UIKit
 import XCTest
 
+/// To print the internal view hierarchy of a view in the test cases, follow these steps:
+///
+/// 1. Set a breakpoint in the test case after creating the view instance.
+/// 2. Run the test case.
+/// 3. In the debugger, print the view hierarchy by evaluating the following expression in `lldb`:
```
philippsuess, line 13: `m`: We can delay this comment to a follow-up PR. What about storing the output of `view.value(forKey: "recursiveDescription")` as an assertion? So basically, let the description of the views in the test assert against the stored output. Then we know when it changes. Or is this a bad idea?

philprime: I wouldn't do this because it could result in failures when Apple changes the implementation of `recursiveDescription`, causing the string output to change while the internal view hierarchy didn't change.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: @romtsn wdyt?

romtsn: yes exactly, we wanted to be super cautious with unmasking and rather unmask too little than too much.

philprime: Thanks for confirming, I'll add a comment.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I'll leave it out of this PR for now. We can revisit it when we do #5155

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I just double-checked it and so far we only use the `ClassIdentifier` for ignored classes without layer filtering, so changing it to your proposed changes works and all tests pass.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: Replaced it to use `containsIgnoreClass` so it's reusing the same lookup/

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests+SpecialViews.swift
```diff
@@ -0,0 +1,415 @@
...
+            XCTAssertNil(navigationBarRegion.color)
+            XCTAssertEqual(navigationBarRegion.size, CGSize(width: 0, height: 0))
+            XCTAssertEqual(navigationBarRegion.type, .redact)
+            XCTAssertEqual(navigationBarRegion.transform, CGAffineTransform(a: 1, b: 0, c: 0, d: 1, tx: 20, ty: 20))
+
+            let placeholderRegion = try XCTUnwrap(regions.element(at: 2)) // SFSafariLaunchPlaceholderView
+            XCTAssertNil(placeholderRegion.color)
+            XCTAssertEqual(placeholderRegion.size, CGSize(width: 40, height: 40))
+            XCTAssertEqual(placeholderRegion.type, .redact)
+            XCTAssertEqual(toolbarRegion.transform, CGAffineTransform(a: 1, b: 0, c: 0, d: 1, tx: 20, ty: 20))
```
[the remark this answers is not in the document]
philprime: Good Bot

Tests/SentryTests/ViewCapture/SentryUIRedactBuilderTests.swift
philippsuess: > could result in failures when Apple changes the implementation of recursiveDescription

Maybe that's a good thing, cause it could tell us that they changed something that could impact masking.

philprime: We are handling that by asserting the exact regions in combinations with snapshots

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I added a lookup set which should be O(1)

philprime:
FYI @philipphofmann @itaybre @noahsmartin this PR is done, but the CI is failing because snapshot tests of standard components are flaky, especially as we are using a Beta of Xcode 26.1

philippsuess:
> FYI @philipphofmann @itaybre @noahsmartin this PR is done, but the CI is failing because snapshot tests of standard components are flaky, especially as we are using a Beta of Xcode 26.1

@philprime what do you suggest to solve this problem?

philprime:
I am going to split this up into a PR stack with a PR per test-file so we can merge them individually instead of keeping this open forever

philprime:
* **#6571** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cocoa/6571?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6570** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cocoa/6570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6569** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cocoa/6569?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6568** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cocoa/6568?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6567** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cocoa/6567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6292** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cocoa/6292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cocoa/6292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

PR #6300, opened 2025-09-29
test: Ensure test server running

philipphofmann:
Ensure that the test server is running with a retry mechanism to avoid flakiness in CI.

#skip-changelog


philippsuess:
Back to draft cause tests are failing

PR #6309, opened 2025-09-30
ref: Migrate AppStateManager to Swift

noahsmartin:
Resolves https://linear.app/getsentry/issue/COCOA-714/convert-sentryappstatemanager

#skip-changelog


Sources/Swift/SenryAppStateManager.swift
itaybre, line 1: `h`: I guess this file was a mistake and can be deleted

PR #6312, opened 2025-10-01
chore: Cleanup interface for SentryScreenFrames without UIKit

### LINKED ISSUE #6313: chore: Cleanup interface for SentryScreenFrames without UIKit
> [!NOTE]
> The pull request "[chore: Cleanup interface for SentryScreenFrames without UIKit](https://github.com/getsentry/sentry-cocoa/pull/6312)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

I just noticed that this class had a strange pattern of logging errors when used without UIKit instead of using the macros we had defined. Updated it so it wouldn't accidentally get used, luckily it looks like there isn't a call site using it when it shouldn't

#skip-changelog

noahsmartin:
I just noticed that this class had a strange pattern of logging errors when used without UIKit instead of using the macros we had defined. Updated it so it wouldn't accidentally get used, luckily it looks like there isn't a call site using it when it shouldn't

#skip-changelog

Closes #6313

### DIFF
--- a/Sources/Sentry/SentryScreenFrames.m
+++ b/Sources/Sentry/SentryScreenFrames.m
@@ -1,21 +1,20 @@
 #import <SentryScreenFrames.h>
 
-#if SENTRY_UIKIT_AVAILABLE
+#if SENTRY_HAS_UIKIT
 #    import "SentryInternalDefines.h"
 
 @implementation SentryScreenFrames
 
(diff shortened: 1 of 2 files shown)


itaybre (commented):
I am converting this to swift in here: https://github.com/getsentry/sentry-cocoa/pull/6094

PR #6343, opened 2025-10-03
Remove enable tracing flag from V9

noahsmartin:
#skip-changelog


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 Removes deprecated user feedback API, this is replaced with the new feedback API (#5591)
 Removes `enablePerformanceV2` option and makes this the default. The app start duration will now finish when the first frame is drawn instead of when the OS posts the UIWindowDidBecomeVisibleNotification. (#6008)
+Removes enableTracing property from SentryOptions (#5694)
```
philprime: `m`: Please use the list format as we did before v9

Sources/Sentry/Public/SentryOptions.h
```diff
@@ -407,17 +407,6 @@ NS_SWIFT_NAME(Options)
...
 
-#if !SDK_V9
-/**
- * Indicates whether tracing should be enabled.
- * @discussion Enabling this sets @c tracesSampleRate to @c 1 if both @c tracesSampleRate and
- * @c tracesSampler are @c nil. Changing either @c tracesSampleRate or @c tracesSampler to a value
- * other then @c nil will enable this in case this was never changed before.
- */
-@property (nonatomic)
-    BOOL enableTracing DEPRECATED_MSG_ATTRIBUTE("Use tracesSampleRate or tracesSampler instead");
```
philprime, line 418: `h`: With the deprecation notice gone, we need to add a `MIGRATION.md` explaining how to migrate from v8 to v9. It could also be a page in the sentry-docs

noahsmartin: It looks like the Migration.md file points to the docs page already, so seems like we should put it there? Since it's a separate repo I'm thinking we wait to merge that change until we are ready to release V9. We can keep adding to the changelog in this repo, and then for every line in the changelog we can put a more detailed entry in migration.md, how does that sound?

philippsuess: > We can keep adding to the changelog in this repo, and then for every line in the changelog we can put a more detailed entry in migration.md, how does that sound?

That works for me. It's usually easier to do this before we release the next major.

PR #6346, opened 2025-10-04
chore: Remove more deprecated APIs in V9

noahsmartin:
Removes more code that is not included in V9

#skip-changelog

### DIFF
--- a/Sources/Sentry/Public/SentryScope.h
+++ b/Sources/Sentry/Public/SentryScope.h
@@ -165,18 +165,6 @@ NS_SWIFT_NAME(Scope)
  */
 - (void)clear;
 
-#if !SDK_V9
-/**
- * Mutates the current transaction atomically.
- * @param callback the SentrySpanCallback.
- */
-- (void)useSpan:(SentrySpanCallback)callback
-    DEPRECATED_MSG_ATTRIBUTE(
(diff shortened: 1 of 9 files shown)


Sources/Sentry/Public/SentryScope.h
```diff
@@ -165,18 +165,6 @@ NS_SWIFT_NAME(Scope)
...
-#if !SDK_V9
-/**
- * Mutates the current transaction atomically.
- * @param callback the SentrySpanCallback.
- */
-- (void)useSpan:(SentrySpanCallback)callback
-    DEPRECATED_MSG_ATTRIBUTE(
-        "This method was used to create an atomic block that could be used to mutate the current "
-        "span. It is not atomic anymore and due to issues with memory safety in `NSBlock` it is "
-        "now considered unsafe and deprecated. Use `span` instead.");
```
philprime, line 177: `h`: We should add the note to a migration guide, either `MIGRATION.md` or a page in sentry-docs

noahsmartin: Yep as discussed in the other PR we can accumulate the diffs in changelog and then make a PR in the docs repo to add more detailed explanations in the migration guide

PR #6350, opened 2025-10-06
ci: Only run test server on macOS

### LINKED ISSUE #6351: ci: Only run test server on macOS
> [!NOTE]
> The pull request "[ci: Only run test server on macOS](https://github.com/getsentry/sentry-cocoa/pull/6350)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

The test server tests on iOS are very flaky. Let's only run these on macOS.

#skip-changelog

philipphofmann:
The test server tests on iOS are very flaky. Let's only run these on macOS, as they didn't fail there in the last two weeks.
Perhaps that's because the simulator has difficulty communicating with a running test server in CI.

Screenshot 2025-10-06 at 07 10 25

#skip-changelog

Closes #6351

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -130,9 +130,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # Running the tests with simulators is incredibly flaky. As we don't touch that code often, the risk is small
-          # that the unit tests break on iOS but not on macOS, and it's better to have rock-solid unit tests, we only run
-          #these on macOS.
+          # Running the tests with simulators is incredibly flaky. Our assumption is that simulators might have difficulties
+          # with communicating with the test server in CI.
+          # As we don't touch that code often, the risk is small that the unit tests break on iOS but not on macOS, and it's
+          # better to have rock-solid unit tests, we only run these on macOS.
           - name: macOS 15
             runs-on: macos-15
             platform: "macOS"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -130,13 +130,10 @@ jobs:
...
       matrix:
         include:
-          - name: iOS 18
-            runs-on: macos-15
-            platform: "iOS"
-            xcode: "16.4"
-            device: "iPhone 16 Pro"
-            test-destination-os: "18.5"
-
+          # Running the tests with simulators is incredibly flaky. Our assumption is that simulators might have difficulties
```
philprime, line 133: `h`: If I understand this proposal correctly, you want to stop testing functionality requiring the test server like the network tracker to be tested on iOS, due to flakiness? If yes, I don't think this is a valid approach, we must test all functionality on all devices and even though we "don't touch that code often" the risk is high because the environment constantly changes with every iOS release

noahsmartin: I think instead of deleting you can comment out the flaky test with a link to an issue to bring it back or fix it. A flaky test hurts more than it helps so it shouldn't be kept around, but if we believe the underlying idea was useful we can track it and prioritize brining it back, similar to behavior in other codebases that will auto-disable flaky tests and track them

philippsuess: Okay, then let's create an issue to follow up with this: https://github.com/getsentry/sentry-cocoa/issues/6361

PR #6373, opened 2025-10-07
chore: Add tests for SentryCrashContext

itaybre:
Adds a couple of tests before adding some modifications

#skip-changelog

### DIFF
--- a/Tests/SentryTests/SentryCrash/SentryCrashMachineContextTests.m
+++ b/Tests/SentryTests/SentryCrash/SentryCrashMachineContextTests.m
@@ -1,0 +1,122 @@
+#import <XCTest/XCTest.h>
+
+#import "SentryCrashMachineContext.h"
+#import "SentryCrashMachineContext_Apple.h"
+#import "TestThread.h"
+#import <mach/mach.h>
+
+@interface SentryCrashMachineContextTests : XCTestCase
+@end
+
+@implementation SentryCrashMachineContextTests
+
+- (void)testGetContextForThread_NonCrashedContext_DoesNotPopulateThreadList
+{
+    // Create a test thread
+    NSObject *notificationObject = [[NSObject alloc] init];
+    TestThread *thread = [[TestThread alloc] init];
+    thread.notificationObject = notificationObject;
+
+    XCTestExpectation *exp = [self expectationWithDescription:@"thread started"];
+    [NSNotificationCenter.defaultCenter
+        addObserverForName:@"io.sentry.test.TestThread.main"
+                    object:notificationObject
+                     queue:nil
+                usingBlock:^(NSNotification *_Nonnull __unused notification) {
+                    [NSNotificationCenter.defaultCenter
+                        removeObserver:self
+                                  name:@"io.sentry.test.TestThread.main"
+                                object:notificationObject];
+                    [exp fulfill];
+                }];
+
+    [thread start];
+    [self waitForExpectationsWithTimeout:1 handler:nil];
+
+    kern_return_t kr;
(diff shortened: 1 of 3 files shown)


Tests/SentryTests/SentryCrash/SentryCrashMachineContextTests.m
```diff
@@ -0,0 +1,122 @@
...
+    int threadCount = sentrycrashmc_getThreadCount(machineContext);
+    XCTAssertTrue(
+        threadCount >= 10, @"Thread count should include all test threads, got %d", threadCount);
+    XCTAssertTrue(threadCount <= SENTRY_CRASH_MAX_THREADS,
+        @"Thread count should not exceed maximum of SENTRY_CRASH_MAX_THREADS, got %d", threadCount);
+
+    // Verify that all our threads are in the list
+    for (TestThread *thread in threads) {
+        int threadIndex = sentrycrashmc_indexOfThread(machineContext, thread.thread);
+        XCTAssertTrue(threadIndex >= 0, @"Thread should be found in the list");
```
philippsuess, line 116: `h`: When I run this test repeatedly in Xcode, it always fails for me here. This is because the test doesn't wait until all threads are finished, and then I have over 100 threads running. I think we should add some cleanup loop that iterates over all threads and waits until all of them are finished. Ideally, we use XCTestExpectations for this. The same applies to the test above.

philipphofmann (commented):
The tests don't cleanup the threads. We need to fix that. Otherwise, we might end up with flaky tests.

Tests/SentryTests/SentryCrash/SentryCrashMachineContextTests.m
```diff
@@ -0,0 +1,134 @@
...
+
+    // Clean up
+    thread_resume(firstThread.thread);
+    NSMutableArray<XCTestExpectation *> *finishExpectations =
+        [NSMutableArray arrayWithCapacity:threads.count];
+    for (TestThread *thread in threads) {
+        thread.endExpectation =
+            [[XCTestExpectation alloc] initWithDescription:@"Wait for thread to cancel"];
+        [finishExpectations addObject:thread.endExpectation];
+        [thread cancel];
```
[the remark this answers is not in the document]
itaybre: Can be discarded, the thread wont end until we call `cancel`

Tests/SentryTests/SentryCrash/SentryCrashMachineContextTests.m
```diff
@@ -0,0 +1,134 @@
...
+                usingBlock:^(NSNotification *_Nonnull __unused notification) {
+                    [NSNotificationCenter.defaultCenter
+                        removeObserver:self
+                                  name:@"io.sentry.test.TestThread.main"
+                                object:notificationObject];
+                    [exp fulfill];
+                }];
+
+    [thread start];
+    [self waitForExpectationsWithTimeout:1 handler:nil];
```
noahsmartin: Might want to increase this timeout since CI machines can run slow

Tests/SentryTests/SentryCrash/SentryCrashMachineContextTests.m
```diff
@@ -0,0 +1,134 @@
...
+                                      name:@"io.sentry.test.TestThread.main"
+                                    object:notificationObject];
+                        [exp fulfill];
+                    }];
+
+        [threads addObject:thread];
+        [thread start];
+    }
+
+    [self waitForExpectations:expectations timeout:2];
```
noahsmartin: Same here, I recommend 5 for CI

PR #6375, opened 2025-10-07
fix: Fix thread deallocation

itaybre:
Thread deallocation logic is only deallocating thread in the context

#skip-changelog

### DIFF
--- a/Sources/SentryCrash/Recording/Tools/SentryCrashMachineContext.c
+++ b/Sources/SentryCrash/Recording/Tools/SentryCrashMachineContext.c
@@ -70,7 +70,7 @@ getThreadList(SentryCrashMachineContext *context)
         SENTRY_ASYNC_SAFE_LOG_ERROR("task_threads: %s", mach_error_string(kr));
         return false;
     }
-    SENTRY_ASYNC_SAFE_LOG_TRACE("Got %d threads", context->threadCount);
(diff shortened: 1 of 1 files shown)


Sources/SentryCrash/Recording/Tools/SentryCrashMachineContext.c
```diff
@@ -84,7 +84,7 @@ getThreadList(SentryCrashMachineContext *context)
     context->threadCount = threadCount;
 
     for (mach_msg_type_number_t i = 0; i < actualThreadCount; i++) {
-        mach_port_deallocate(thisTask, context->allThreads[i]);
+        mach_port_deallocate(thisTask, threads);
```
philippsuess: `h`:
```suggestion
        mach_port_deallocate(thisTask, threads[i]);
```

philipphofmann (approved):
When fixing the h issue. LGTM

Sources/SentryCrash/Recording/Tools/SentryCrashMachineContext.c
itaybre: Wow, huge thanks for catching that

philippsuess: Seer also caught it 😄 https://github.com/getsentry/sentry-cocoa/pull/6375#discussion_r2411931563

PR #6383, opened 2025-10-08
fix: Improve flaky tests at SentryDispatchQueueWrapperSwiftTests

itaybre:
Tests at `SentryDispatchQueueWrapperSwiftTests` are being flaky.
Adding some active wait to improve this

image

#skip-changelog

### DIFF
--- a/Tests/SentryTests/Networking/SentryDispatchQueueWrapperTests.swift
+++ b/Tests/SentryTests/Networking/SentryDispatchQueueWrapperTests.swift
@@ -30,46 +30,51 @@ class SentryDispatchQueueWrapperSwiftTests: XCTestCase {
     }
     
     func testDispatchAsyncOnMainQueueIfNotMainThreadOnMain() {
-        let expectation = XCTestExpectation()
+        let blockExpectation = XCTestExpectation(description: "Block Expectation")
         
-        var flag = false
+        let outerExpectation = XCTestExpectation(description: "This exepctation should execute last")
+        outerExpectation.isInverted = true
         
         let sut = SentryDispatchQueueWrapper()
         sut.dispatchAsyncOnMainQueueIfNotMainThread {
-            // Add a sleep to ensure this executes before the flag is set even if delayed
-            sleep(1)
             XCTAssertTrue(Thread.isMainThread)
-            XCTAssertFalse(flag, "Block did not run synchronously")
+            self.wait(for: [outerExpectation], timeout: 1)
             
-            expectation.fulfill()
+            blockExpectation.fulfill()
         }
-        flag = true
+        outerExpectation.fulfill()
         
-        wait(for: [expectation], timeout: 2)
+        wait(for: [blockExpectation], timeout: 2)
     }
     
     func testDispatchAsyncOnMainQueueIfNotMainThreadFromBackground() {
         let expectation = XCTestExpectation()
         
         DispatchQueue.global().async {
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/Networking/SentryDispatchQueueWrapperTests.swift
```diff
@@ -36,6 +36,8 @@ class SentryDispatchQueueWrapperSwiftTests: XCTestCase {
         
         let sut = SentryDispatchQueueWrapper()
         sut.dispatchAsyncOnMainQueueIfNotMainThread {
+            // Add a sleep to ensure this executes before the flag is set even if delayed
+            sleep(1)
```
philippsuess: `h`: I think this sleep shouldn't be required. Everything should run on the main thread. If we need a sleep here, something is broken.

Tests/SentryTests/Networking/SentryDispatchQueueWrapperTests.swift
```diff
@@ -55,6 +57,8 @@ class SentryDispatchQueueWrapperSwiftTests: XCTestCase {
             var flag = false
             let sut = SentryDispatchQueueWrapper()
             sut.dispatchAsyncOnMainQueueIfNotMainThread {
+                // Add a sleep to ensure this executes after the flag is set
+                sleep(1)
```
philippsuess: `m`: This won't help much. We have a race condition here. Both the main thread and the background thread write the `flag`. When we schedule work here from a bg thread to the main thread, the order of execution isn't deterministic, meaning if `XCTAssertTrue(flag)` or `flag = true` runs first is somewhat random. Even adding a thread sleep(1) could maybe in rare cases lead to a test failure.

I would recommend something like this, which would deadlock for 5 seconds if the two expectations run on the main thread, which is basically removing `DispatchQueue.global().async {` from the test.
```swift
func testDispatchAsyncOnMainQueueIfNotMainThreadFromBackground() {
    let expectation = XCTestExpectation()
    
        let mainThreadExpectation = XCTestExpectation(description: "Main Thread Expectation")
        let bgThreadExpectation = XCTestExpectation(description: "BG Thread Expectation")

        let sut = SentryDispatchQueueWrapper()

        sut.dispatchAsyncOnMainQueueIfNotMainThread {
            XCTAssertTrue(Thread.isMainThread, "The block didn't run on the main thread, but it should have")

            // Wait for the background thread to fulfill its expectation
            // If this code runs on the same thread as the bgThreadExpectation, the expectation times out.
            self.wait(for: [bgThreadExpectation], timeout: 5.0)

            // Unblock the BG thread
            mainThreadExpectation.fulfill()
        }

        // Unblock the main thread
        bgThreadExpectation.fulfill()

        self.wait(for: [mainThreadExpectation], timeout: 5.0)

        expectation.fulfill()
    }
    
    wait(for: [expectation], timeout: 10.0)
}
```

Tests/SentryTests/Networking/SentryDispatchQueueWrapperTests.swift
itaybre: Nono, the sleep is not required. Added it as an additional verification.
I can use an inverted expectation too to make this clear

philippsuess: Yes, please ditch the sleep. Everything runs on the main thread. We shouldn't have to use expectations.

PR #6389, opened 2025-10-09
fix(session-replay): Add detection for potential PII leaks disabling session replay

philprime:
## :scroll: Description

- Adds a check using the iOS runtime version and the `DTXcode` from the app's `Info.plist` to detect if the environment should be considered `dangerous` for PII.
- Apps built with Xcode 26 and running on iOS 26 will not start Session Replay due to PII concerns
- Adds a new option (should be deprecated as soon as fixed) to override the behavior.

While working on #6292 I noticed that standard SwiftUI components like `Text`, `Image`, `Label` are not properly masked. Further research has shown that this is caused by the masking/redaction logic not being able to detect sensitive views due to Apple changing the internal rendering logic with their release of _Liquid Glass_ on iOS 26.0.

Liquid Glass is only used by an app running on iOS 26.0 if the app was built with Xcode 26.0, therefore we can rule out apps built with older Xcode versions. This can be checked by looking at the `DT*` fields in the app's `Info.plist`.

The approach taken in this PR is defensive, meaning that we assume the runtime environment is dangerous for PII unless there are strong indicators it's safe. This approach should reduce the risk of PII being leaked by changes outside of our control, but it could cause session replay not being started even though it's safe ("better safe than sorry").

I am going to create a follow-up issue with more insight into the issue and link it here afterwards.

- I exported a sample app using Xcode 16.4 and using Xcode 26, then compared the `Info.plist` of the two apps to confirm the values of the field `DTXcode`. Using Xcode 16.4

image
[Info-iOS18.plist.xml](https://github.com/user-attachments/files/22799590/Info-iOS18.plist.xml)
[Info-iOS26.plist.xml](https://github.com/user-attachments/files/22799591/Info-iOS26.plist.xml)

- I ran the iOS-SwiftUI sample using Xcode 16.4 on iOS 26.0 simulator → session replay is enabled
- I ran the iOS-SwiftUI sample using Xcode 26.0 on iOS 26.0 simulator:
  - with `options.sessionReplay.disableInDangerousEnvironment = true` (default) → session replay is disabled
  - with `options.sessionReplay.disableInDangerousEnvironment = false` → session replay was enabled

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift
+++ b/Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift
@@ -2,8 +2,15 @@
 
     private let bundle: Bundle
 
-    public init(bundle: Bundle = Bundle.main) {
+    public override init() {
+        // We can not use defaults in the initializer because this class is used from Objective-C
+        self.bundle = Bundle.main
+        super.init()
+    }
+
+    public init(bundle: Bundle) {
         self.bundle = bundle
+        super.init()
     }
 
     // MARK: - Bridge to ObjC


Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -215,6 +216,17 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+    /**
+     * Due to internal changes with the release of Liquid Glass on iOS 26.0, the masking of text and images can not be reliably guaranteed.
+     *
+     * Therefore the session replay integration is disabled by default starting with `8.57.0` as a defensive mechanism.
+     *
+     * - Important: This flag allows to re-enable the session replay integration on iOS 26.0 and later, but please be aware that text and images may not be masked as expected.
+     *
+     * - Note: See [GitHub issues #1234](https://github.com/getsentry/sentry-cocoa/issue/1234) for more information.
+     */
+    public var disableInDangerousEnvironment: Bool
```
philippsuess: `h`: I am a bit unsure about the name of this property. I wouldn't add dangerous in the wording. I would rephrase it to reflect that we can't guarantee masking is reliable. Maybe something with reliable. Maybe we should also consider starting the naming with `enable...`, because that's how we usually do all the other options.

What about `enableInNonReliableEnvironment`? The environment isn't dangerous, but we can't reliable mask everything.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -73,6 +74,18 @@ import UIKit
             SentrySDKLog.debug("[Session Replay] Session replay is already running, not starting again")
             return 
         }
+        
+        // Detect if we are running on iOS 26.0 with Liquid Glass and disable session replay.
+        // This needs to be done until masking for session replay is properly supported, as it can lead
+        // to PII leaks otherwise.
+        if isRunningInDangerousEnvironment() {
+            if replayOptions.disableInDangerousEnvironment {
+                SentrySDKLog.fatal("[Session Replay] Detected environment potentially causing PII leaks, disabling Session Replay. To override this mechanism, set `options.disableInDangerousEnvironment` to `false`")
```
philippsuess: `m: I would change the wording to communicate that we can't reliably mask in this environment. Again, I would remove the word dangerous.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +382,51 @@ import UIKit
             replayMaker.addFrameAsync(timestamp: timestamp, maskedViewImage: maskedViewImage, forScreen: screen)
         }
     }
+    
+    private func isRunningInDangerousEnvironment() -> Bool {
```
philippsuess: `h`: This could turn into a nightmare to maintain, because we might have to change the condition in here. If we suddenly can fix masking on liquid class, but detected somehting else that's `non-reliable / dangerous` we can't change the logic here without a breaking change. What about changing the logic to only detecting if we're running liquid glass?

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +382,51 @@ import UIKit
...
+        // 2. The app was built with Xcode < 26.0 (DTXcode < 2600)
+        
+        // First check: Are we even on iOS 26.0+?
+        guard #available(iOS 26.0, *) else {
+            // Not on iOS 26.0+ - safe to use Session Replay
+            return false
+        }
+        
+        // We're on iOS 26.0+ - assume dangerous unless proven otherwise
+        guard let infoDictionary = Bundle.main.infoDictionary else {
```
philippsuess: `m`: We could inject the infoDict somehow, so we can write proper unit tests for the logic here.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +382,51 @@ import UIKit
...
+            // Can't read Info.plist - stay defensive
```
philippsuess: That makes sense 👍

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +382,51 @@ import UIKit
...
+            SentrySDKLog.debug("[Session Replay] Running on iOS 26.0+ but cannot read Info.plist - treating as dangerous")
```
philippsuess: `m`: I guess it's worth to know
```suggestion
            SentrySDKLog.warning("[Session Replay] Running on iOS 26.0+ but cannot read Info.plist - treating as dangerous")
```

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +382,51 @@ import UIKit
...
+        // Safety check 1: Is compatibility mode explicitly enabled?
+        if let requiresCompatibility = infoDictionary["UIDesignRequiresCompatibility"] as? Bool,
+           requiresCompatibility == true {
+            SentrySDKLog.debug("[Session Replay] Running on iOS 26.0+ with UIDesignRequiresCompatibility=YES - safe to use")
+            return false
+        }
+        
+        // Safety check 2: Was the app built with an older Xcode version?
+        // DTXcode format: Xcode 16.4 = "1640", Xcode 26.0 = "2600"
+        if let xcodeVersionString = infoDictionary["DTXcode"] as? String,
```
philippsuess: `h`: I'm unsure if the `DTXcode` ends up in all release builds and we can rely on it.

philprime:
From a sync between @philipphofmann and myself on Oct 10th 2025:

- Our main concern is user PII therefore we prefer to disable session replay in cases where masking is not reliable over session replay being available. This might cause SDK users to question why they are not receiving session replays, but we believe this is in their interest too.
- We move the `options.sessionReplay.enableInUnreliableEnvironment` to `options.experimental.enableInUnreliableEnvironment`
- The option is there to allow SDK users to overrule our detection of unreliable environment for the case they know it better than us
- This gives us more time to build a proper API because we want to ship a fix as soon as possible without having to consider all aspects of future-proofing this change (incl. backwards and forwards compatibility)
- We will remove the option most likely in a **future minor version**
- If an SDK user doesn't notice that we disabled session replay in a minor version for them, this can cause disruption. After consideration we did not find a intuitive way without causing too much friction to inform the developer at build time. Therefore we will only add warnings to these:
  - Release Notes
  - Changelog (in `v8.x` branch and link in v9 branch in `main`)
  - Troubleshooting Docs
- We will also add a link to v8 branch + v8 Changelog in the v9/`main` branch README.md
- We are going to release this hotfix as `8.57.0` after resetting the `v8.x` branch to the `8.56.2` commit + cherry-picking CI commits to be able to release the hotfix (@philipphofmann will take care of this).
- The changes of this PR will also be applied to v9 in a follow-up PR

kahest:
> * options.experimental.enableInUnreliableEnvironment

I think we should still include SR in the option name, maybe something like `options.experimental.enableSessionReplayInUnreliableEnvironment` because I assume/suggest we only use this for SR

kahest:
> * We are going to release this hotfix as `8.57.0`

why not `8.56.3` which would be an actual hotfix rather than a minor bump?

kahest:
> * If an SDK user doesn't notice that we disabled session replay in a minor version for them, this can cause disruption. After consideration we did not find a intuitive way without causing too much friction to inform the developer at build time.

because it's not mentioned in your comment - we'll still show a log message right?

philprime:
> I think we should still include SR in the option name, maybe something like options.experimental.enableSessionReplayInUnreliableEnvironment because I assume/suggest we only use this for SR

Fine for me, as it is going to be an experimental option we can reduce the usage scope to session replay. In case we want to repurpose it at some point we should keep the name broader.

> * We are going to release this hotfix as `8.57.0`

I would propose we go for a minor because it has quite some impact on the SDK user.

> because it's not mentioned in your comment - we'll still show a log message right?

Yes

philprime:
@cursor review
@sentry review

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -30,7 +30,24 @@ public class SentryExperimentalOptions: NSObject {
...
+     * - The app was built with Xcode < 26.0 (DTXcode < 2600)
+     *
+     * - Important: This flag allows to re-enable the session replay integration on iOS 26.0 and later, but please be aware that text and images may not be masked as expected.
+     *
+     * - Note: See [GitHub issues #6389](https://github.com/getsentry/sentry-cocoa/issues/6389) for more information.
+     */
+    public var enableSessionReplayInUnreliableEnvironment = false
+
     /**
      * Logs are considered beta.
```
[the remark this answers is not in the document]
philprime: While this is true and we are not adhering to semantic versioning, as this is a experimental option used as a workaround it's fine.

CHANGELOG.md
```diff
@@ -1,7 +1,40 @@
 # Changelog
 
+## Unreleased
```
philprime, line 3: @kahest would appreciate your feedback on the wording of this warning at the unreleased changes and at the ones at previous releases.

I will also add these warnings to the Releases and to the v9/main branch

CHANGELOG.md
```diff
@@ -1,7 +1,40 @@
...
+- Session replay is now automatically disabled in environments with unreliable masking to prevent PII leaks (#6389)
+  - Detects iOS 26.0+ runtime with Xcode 26.0+ builds (DTXcode >= 2600)
+  - Detects missing or disabled `UIDesignRequiresCompatibility`
+  - Uses defensive approach: assumes unsafe unless proven safe
+- Add `options.experimental.enableSessionReplayInUnreliableEnvironment` to allow overriding the automatic disabling (#6389)
+
 ## 8.56.2
 
+> [!Warning]
+> Session Replay in this version does not correctly mask views when built with Xcode 26 and running on iOS 26 with Liquid Glass, which may lead to PII leaks. Please upgrade to 8.57.0 or later.
```
kahest: > Please upgrade to 8.57.0 or later

this suggests to me that it's working correctly in 8.57.0 or later - but it's only automatically deactivated there. I think it'd be better to refer to the GH issue for now and update with a fixed version later, OR change it to "Please refer to [release 8.57.0](add link here) for more information"

CHANGELOG.md
```diff
@@ -1,7 +1,40 @@
 # Changelog
 
+## Unreleased
+
+> [!Warning]
+> **Session Replay is disabled by default on iOS 26.0+ with Xcode 26.0+ to prevent PII leaks**
+>
+> Due to masking issues introduced by Apple's Liquid Glass rendering changes in iOS 26.0, session replay is now **automatically disabled** on apps running iOS 26.0+ when built with Xcode 26.0 or later. This is a defensive measure to protect user privacy and prevent potential PII leaks until masking is reliably supported.
```
kahest:
```suggestion
> Due to potential masking issues introduced by Apple's Liquid Glass rendering changes in iOS 26.0, Session Replay is now **automatically disabled** on apps running iOS 26.0+ when built with Xcode 26.0 or later. This is a defensive measure to protect user privacy and prevent potential PII leaks until masking is reliably supported.
```

CHANGELOG.md
```diff
@@ -1,7 +1,40 @@
...
+>
+> Session replay will work normally if:
```
kahest, line 10: did we try to also keep it enabled if `maskAllImages` AND `maskAllText` are both turned off?

philprime: Not yet, at this point we are disabling it in any case. Do you think making an exception for `maskAllImages = false` and `masksAllText = false` is reasonable? SDK users could still have special configurations with masking modifiers etc.

CHANGELOG.md
philprime: I agree it makes sense to mention the GH issue, but upgrading to v8.57.0 will be the solution we should propose to customers so they don't stop using session replay all together but only when the environment is unreliable. wdyt?

kahest: fair point - we can always go back and update the version once we have an actual fix

CHANGELOG.md
kahest: can custom masking rules also affected, meaning could liquid glass also break custom masking rules depending on how they're defined? if yes, then let's keep the way it is. if we can be reasonably sure that only auto-masking is affected, we could think about my suggestion

CHANGELOG.md
itaybre: I would mention that 8.57.0 does not fix the issue, but disabled the feature if there is a risk of PII leak

SentryTestUtils/TestInfoPlistWrapper.swift
```diff
@@ -0,0 +1,55 @@
...
+        switch result {
+        case .success(let value):
+            return value
+        case .failure(let error):
+            throw error
+        }
+    }
+
+    public var getAppValueBooleanInvocations = Invocations<(String, NSErrorPointer)>()
+    private var mockedGetAppValueBooleanReturnValue: [String: Result<Bool, NSError>] = [:]
```
itaybre: `l`: Personal opinion, but I prefer to keep the variable at the top of the class

SentryTestUtilsTests/TestInfoPlistWrapperTests.swift
```diff
@@ -0,0 +1,295 @@
...
+
+    // MARK: - getAppValueString(for:)
+
+    func testGetAppValueString_withoutMockedValue_shouldFailWithPreconditionFailure() throws {
+        // -- Arrange --
+        let sut = TestInfoPlistWrapper()
+        // Don't mock any value for this key
+
+        // -- Act --
+        let e = catchBadInstruction {
```
itaybre: `h`: Looks like this line fails on some tests

SentryTestUtilsTests/TestInfoPlistWrapperTests.swift
```diff
@@ -0,0 +1,295 @@
...
+
+    // MARK: - getAppValueBoolean(for:errorPtr:)
+
+    func testGetAppValueBoolean_withoutMockedValue_shouldFailWithPreconditionFailure() throws {
+        // -- Arrange --
+        let sut = TestInfoPlistWrapper()
+        // Don't mock any value for this key
+
+        // -- Act --
+        let e = catchBadInstruction {
```
itaybre: Same

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +388,86 @@ import UIKit
...
+        }
+
+        // Safety check 2: Was the app built with an older Xcode version?
+        // DTXcode format: Xcode 16.4 = "1640", Xcode 26.0 = "2600"
+        do {
+            let xcodeVersionString = try infoPlistWrapper.getAppValueString(
+                for: SentryInfoPlistKey.xcodeVersion.rawValue
+            )
+            if let xcodeVersion = Int(xcodeVersionString) {
+                if xcodeVersion < 2_600 {
```
itaybre: `l`: Feels like a magic number, can you add a name to the variable?

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +388,86 @@ import UIKit
...
+                }
+            } else {
+                SentrySDKLog.warning("[Session Replay] Found xcode version key but could not parse as Int: \(xcodeVersionString)")
+            }
+        } catch SentryInfoPlistError.mainInfoPlistNotFound {
+            // Can't read Info.plist - stay defensive
+            SentrySDKLog.warning("[Session Replay] Running on iOS 26.0+ but cannot read Info.plist - detected as unreliable")
+            return true
+        } catch SentryInfoPlistError.keyNotFound {
+            // Due to Objective-C using a return value of `nil` as an indicator for an error,
```
itaybre: A possible fallback is reading `LC_BUILD_VERSION` command from the main binary.

Tests/SentryTests/Helper/SentryInfoPlistWrapperTests.swift
```diff
@@ -0,0 +1,194 @@
...
+        // Assert
+        // If the key exists and is a boolean, it should work without error
+        // If the key doesn't exist, error should be set
+        if error == nil {
+            // Success case - value is valid
+            XCTAssertTrue(value == true || value == false, "Boolean value should be true or false")
+        } else {
+            // Key not found is acceptable for this test setup
+            XCTAssertTrue(error?.domain == "SentryInfoPlistError" || error != nil)
+        }
```
itaybre: Wouldn't it be better to inject a custom info.plist instead of relying on an error to know what to test?

It could break at some point and we won't notice

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayTests.swift
```diff
@@ -551,6 +568,145 @@ class SentrySessionReplayTests: XCTestCase {
...
     }
+    
+    // MARK: - iOS 26 Liquid Glass Detection Tests
+    
+    func testStart_withIOS26WithLiquidGlass_withDefaultConfiguration_shouldNotStartSessionReplay() throws {
+        // This test will only run on iOS 26.0+
+        // It tests that session replay is blocked when Liquid Glass is detected
+        guard #available(iOS 26.0, *) else {
+            throw XCTSkip("Test is disabled for this OS version")
+        }
```
itaybre: `l`: Do we have a test that ensures it will be running if the OS < iOS26?

SentryTestUtils/TestInfoPlistWrapper.swift
philippsuess: I agree

SentryTestUtilsTests/TestInfoPlistWrapperTests.swift
```diff
@@ -0,0 +1,295 @@
...
+        let result1 = sut.getAppValueBoolean(for: "key1", errorPtr: &error1)
+
+        var error2: NSError?
+        let result2 = sut.getAppValueBoolean(for: "key1", errorPtr: &error2)
+
+        // -- Assert --
+        XCTAssertTrue(result1)
+        XCTAssertNil(error1)
+        XCTAssertTrue(result2)
+        XCTAssertNil(error2)
```
philippsuess: `l`: It would be nice to add a failure message for these asserts, as when they fail you have no clue what the problem is.

SentryTestUtilsTests/TestInfoPlistWrapperTests.swift
```diff
@@ -0,0 +1,295 @@
+import CwlPreconditionTesting
+@_spi(Private) @testable import Sentry
+@_spi(Private) @testable import SentryTestUtils
+import XCTest
+
+class TestInfoPlistWrapperTests: XCTestCase {
```
philippsuess, line 5: Thanks for adding tests for a test class 🥇

Sources/Sentry/SentryDependencyContainer.m
```diff
@@ -189,6 +189,7 @@ - (instancetype)init
             [[SentryDefaultRateLimits alloc] initWithRetryAfterHeaderParser:retryAfterHeaderParser
                                                          andRateLimitParser:rateLimitParser
                                                         currentDateProvider:_dateProvider];
+        _infoPlistWrapper = [[SentryInfoPlistWrapper alloc] init];
```
philippsuess, line 192: Nothing to worry about right now, but this has the downside of allocating a `SentryInfoPlistWrapper` even for users not using SR. I'm only adding this comment, because you want to fix the DependencyContainer at some point and this should be considered when you do this.

Sources/Swift/Helper/InfoPlist/SentryInfoPlistError.swift
```diff
@@ -0,0 +1,5 @@
+enum SentryInfoPlistError: Error {
```
philippsuess, line 1: `m`: I think we could put all the plist classes into a PlistInfo or something folder instead of the Helper folder.

Sources/Swift/Helper/InfoPlist/SentryInfoPlistKey.swift
```diff
@@ -0,0 +1,15 @@
+public enum SentryInfoPlistKey: String {
```
philippsuess: `h`: We miss `@objc @_spi(Private)` here. Is that intentional?

Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift
```diff
@@ -0,0 +1,40 @@
+@objc @_spi(Private) public class SentryInfoPlistWrapper: NSObject, SentryInfoPlistWrapperProvider {
+    // MARK: - Bridge to ObjC
+
+    public func getAppValueBoolean(for key: String, errorPtr errPtr: NSErrorPointer) -> Bool {
```
philippsuess, line 18: `m`: If I'm not mistaken we don't call this from ObjC anywhere anymore. Can't we remove the `NSErrorPointer`?

Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift
```diff
@@ -0,0 +1,40 @@
...
+        }
+        return value
+    }
+
+    // MARK: - Swift Implementation
+
+    private func getAppValue<T>(for key: String, type: T.Type) throws -> T? {
+        // As soon as this class is not consumed from Objective-C anymore, we can use this method directly to reduce
+        // unnecessary duplicate code. In addition this method can be adapted to use `SentryInfoPlistKey` as the type
+        // of the parameter `key`
```
philippsuess, line 42: `m`: I think we can already do that, as we're not using this class in ObjC, except for the dependency container.

Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift
```diff
@@ -0,0 +1,27 @@
+@_spi(Private) @objc public protocol SentryInfoPlistWrapperProvider {
```
philippsuess, line 1: `m`: We only have one real implementation of this protocol, and of course the test wrapper.  Can't we subclass the real implementation with the test wrapper? Then we can ditch this one extra protocol.

Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Retrieves a value from the app's `Info.plist` file for the given key and trys to cast it to a ``String``.
+     *
+     * - Parameter key: The key for which to retrieve the value from the `Info.plist`.
+     * - Throws: An error if the value cannot be cast to type ``String`` or ``SentryInfoPlistError.keyNotFound`` if the key was not found or the value is `nil`
+     * - Returns: The value associated with the specified key cast to type ``String``
+     * - Note: The return value can not be nullable, because a throwing function in Objective-C uses `nil` to indicate an error:
+     *
+     *       Throwing method cannot be a member of an '@objc' protocol because it returns a value of optional type 'String?'; 'nil' indicates failure to Objective-C
+     */
```
philippsuess, line 11: `l`: This is an internal class. IMO, we don't have to add detailed docs for this, because you can checkout the code and tests, and these docs tend to get out of date when the implementation changes

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +388,86 @@ import UIKit
             replayMaker.addFrameAsync(timestamp: timestamp, maskedViewImage: maskedViewImage, forScreen: screen)
         }
     }
+    
+    // swiftlint:disable:next cyclomatic_complexity function_body_length
+    private func isEnvironmentUnreliable() -> Bool {
```
philippsuess: `m`: I think we could move this method to an extra file to keep this class small. We can either create an extra class or (what I would prefer) create a static method in an extra file and only pass the `infoPlistWrapper` as a method parameter. This would also make it way easier to test this functionality in isolation.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +388,86 @@ import UIKit
...
+                errorPtr: &error
+            )
+            if let error = error as Error? {
+                throw error
+            } else if requiresCompatibility {
+                SentrySDKLog.debug("[Session Replay] Running with UIDesignRequiresCompatibility set to YES - detected as reliable")
+                return false
+            } else {
+                SentrySDKLog.debug("[Session Replay] Running with UIDesignRequiresCompatibility is set to NO")
+            }
```
philippsuess: `m`: I think we could remove the NSError part and rely on Swift error handling to simplify the code.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +388,86 @@ import UIKit
...
+        } catch SentryInfoPlistError.mainInfoPlistNotFound {
+            // Can't read Info.plist - stay defensive
+            SentrySDKLog.warning("[Session Replay] Running on iOS 26.0+ but cannot read Info.plist - detected as unreliable")
+            return true
+        } catch SentryInfoPlistError.keyNotFound {
+            // Due to Objective-C using a return value of `nil` as an indicator for an error,
+            // we need to throw an error when the key was not found
+            SentrySDKLog.debug("[Session Replay] No UIDesignRequiresCompatibility found in Info.plist")
+        } catch {
+            SentrySDKLog.error("[Session Replay] Failed to read Info.plist: \(error)")
```
philippsuess: `h`: Shouldn't we also return `true` here and disable SR.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +388,86 @@ import UIKit
...
+            )
+            if let xcodeVersion = Int(xcodeVersionString) {
+                if xcodeVersion < 2_600 {
+                    SentrySDKLog.debug("[Session Replay] Running on iOS 26.0+ but built with Xcode \(xcodeVersionString) (< 26.0) - detected as reliable")
+                    return false
+                } else {
+                    SentrySDKLog.debug("[Session Replay] Detected built with Xcode version: \(xcodeVersionString)")
+                }
+            } else {
+                SentrySDKLog.warning("[Session Replay] Found xcode version key but could not parse as Int: \(xcodeVersionString)")
```
philippsuess: `l`
```suggestion
                SentrySDKLog.warning("[Session Replay] Found Xcode version key but could not parse as Int: \(xcodeVersionString)")
```

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +388,86 @@ import UIKit
...
+            }
+        } catch SentryInfoPlistError.mainInfoPlistNotFound {
+            // Can't read Info.plist - stay defensive
+            SentrySDKLog.warning("[Session Replay] Running on iOS 26.0+ but cannot read Info.plist - detected as unreliable")
+            return true
+        } catch SentryInfoPlistError.keyNotFound {
+            // Due to Objective-C using a return value of `nil` as an indicator for an error,
+            // we need to throw an error when the key was not found.
+            SentrySDKLog.warning("[Session Replay] Could not find xcode version key in Info.plist")
```
philippsuess: `l`:
```suggestion
            SentrySDKLog.warning("[Session Replay] Could not find Xcode version key in Info.plist")
```

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayTests.swift
philprime: Yes, the test in line 623

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +388,86 @@ import UIKit
...
+            let xcodeVersionString = try infoPlistWrapper.getAppValueString(
+                for: SentryInfoPlistKey.xcodeVersion.rawValue
+            )
+            if let xcodeVersion = Int(xcodeVersionString) {
+                if xcodeVersion < 2_600 {
+                    SentrySDKLog.debug("[Session Replay] Running on iOS 26.0+ but built with Xcode \(xcodeVersionString) (< 26.0) - detected as reliable")
+                    return false
+                } else {
+                    SentrySDKLog.debug("[Session Replay] Detected built with Xcode version: \(xcodeVersionString)")
+                }
```
philippsuess: `l`: We could do an early return and invert the `if let xcodeVersion < 2_600` to reduce nesting and simply the code.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -369,7 +388,86 @@ import UIKit
...
+        } catch SentryInfoPlistError.keyNotFound {
+            // Due to Objective-C using a return value of `nil` as an indicator for an error,
+            // we need to throw an error when the key was not found.
+            SentrySDKLog.warning("[Session Replay] Could not find xcode version key in Info.plist")
+        } catch {
+            SentrySDKLog.error("[Session Replay] Failed to read Info.plist: \(error)")
+        }
+
+        // No safety conditions met - treat as unreliable
+        SentrySDKLog.warning("[Session Replay] Detected environment as unreliable")
```
philippsuess: `m`: I would add the `no safety condidtions met` info in this log message.

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -30,7 +30,24 @@ public class SentryExperimentalOptions: NSObject {
...
+     *
+     * Indicators for reliable environments include:
+     * - Running on an older version of iOS that doesn't have Liquid Glass (iOS 18 or earlier)
+     * - UIDesignRequiresCompatibility is explicitly set to YES in Info.plist
+     * - The app was built with Xcode < 26.0 (DTXcode < 2600)
+     *
+     * - Important: This flag allows to re-enable the session replay integration on iOS 26.0 and later, but please be aware that text and images may not be masked as expected.
+     *
+     * - Note: See [GitHub issues #6389](https://github.com/getsentry/sentry-cocoa/issues/6389) for more information.
+     */
```
philippsuess, line 48: `l`: I know the rest of the class already has this wrong, but we should use Swift style code comments in here. We should add a linter for this at some point.

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -30,7 +30,24 @@ public class SentryExperimentalOptions: NSObject {
      * - Experiment: This is an experimental feature and is therefore disabled by default. We'll enable it by default in a future major release.
      */
     public var enableUnhandledCPPExceptionsV2 = false
-    
+
+    /**
+     * Forces enabling of session replay in unreliable environments.
+     *
+     * Due to internal changes with the release of Liquid Glass on iOS 26.0, the masking of text and images can not be reliably guaranteed.
+     * Therefore the SDK uses a defensive programming approach to disable the session replay integration by default, unless the environment is detected as reliable.
```
philippsuess, line 38: `m`: I think we should mention that we currently can't guarantee masking, and this might change in a future version and that's why this flag is eperimental.

SentryTestUtils/TestInfoPlistWrapper.swift
philprime: I agree, in this case I tried to keep them close to the method they mock/track.
I want to refactor this mocking/refactoring soon anyways by encapsulating it.
Will keep it like this for now.

Sources/Sentry/SentryDependencyContainer.m
philprime: So your point being that this should be lazy-loaded?

Tests/SentryTests/Helper/SentryInfoPlistWrapperTests.swift
philprime: That's good feedback, I'll adapt this test

Sources/Swift/SentryExperimentalOptions.swift
philprime: Please create a suggestion, because we already mention that masking is not reliable/guaranteed, so I am not sure what to change.

Sources/Swift/SentryExperimentalOptions.swift
philprime: Do you mean the `///` instead of `/**`? Because both are valid Swift style code comments/documentation.
https://www.swift.org/documentation/docc/writing-symbol-documentation-in-your-source-files#Describe-the-Parameters-of-a-Method

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
philprime: But we do that at the end of the function?

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
philprime: No we can not. Please see the docs of `getAppValueBoolean`.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
philprime: Good idea, I can do that

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
philprime: Good idea, I'll more details as a full PR comment for future reference.

philprime:
Based on [this PR comment](https://github.com/getsentry/sentry-cocoa/pull/6389#discussion_r2420200154) I did an analysis of the `LC_BUILD_VERSION` and will add it as another check:


```
# Built with Xcode 26.0 for iOS Simulator 18.6
$ otool -l iOS-SwiftUI | grep -A 7 LC_BUILD_VERSION
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 7
    minos 14.0
      sdk 18.5
   ntools 1
     tool 3
  version 1167.5
  
# Built with Xcode 26.0 for iOS Simulator 26.0
$ otool -l iOS-SwiftUI | grep -A 7 LC_BUILD_VERSION
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 7
    minos 14.0
      sdk 18.5
   ntools 1
     tool 3
  version 1167.5

# Built with Xcode 26.0 for iOS Simulator 18.6
$ otool -l iOS-SwiftUI | grep -A 7 LC_BUILD_VERSION
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 7
    minos 14.0
      sdk 26.0
   ntools 1
     tool 3
  version 1221.4

# Built with Xcode 26.0 for iOS Simulator 26.0
$ otool -l iOS-SwiftUI | grep -A 7 LC_BUILD_VERSION
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 7
    minos 14.0
      sdk 26.0
   ntools 1
     tool 3
  version 1221.4

File: Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift (hunk 25 above)
Comment (philprime) [reply Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift:18]: We are using this from `SentrySessionReplayIntegration.m` via a the protocol `SentryInfoPlistWrapperProvider` which is the reason why it must be Objective-C compatible.

File: Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift (hunk 26 above)
Comment (philprime) [reply Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift:42]: We are using this from SentrySessionReplayIntegration.m via a the protocol SentryInfoPlistWrapperProvider which is the reason why it must be Objective-C compatible.

File: Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift (hunk 27 above)
Comment (philprime) [reply Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift:1]: Clean architectures use protocols so the implementation can be replaced. We should also become a best-practice SDK with clean interfaces. So no, we should not remove this protocol and keep it in so we can easily replace the implementation in tests

File: Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift (hunk 28 above)
Comment (philprime) [reply Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift:11]: This docs are for you and other future maintainers. Just because it's not consumed by an SDK user, it doesn't mean it won't be consumed by us. The protocol defines the contract, and the implementation must adhere to the contract. Any other class using a dependency with this protocol expect the implementation to work as defined here, not as defined in the implementation. 
If the implementation is changed and doesn't work as defined in the protocol anymore, than the contract is broken.

Also in the age of LLMs this is also context for the agents.

File: Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift (hunk 34 above)
Comment (philprime) [reply Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift]: Which part should be the early return?

File: Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift (hunk 35 above)
Comment (philprime) [reply Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift]: not sure what you mean, please give me a suggestion

File: Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift (hunk 25 above)
Comment (philippsuess) [reply Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift:18]: Yes, but we don't call any methods. We only pass the dependency around. Or am I missing something?

Comment (philprime) [reply Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift:18]: The protocol is not compiling if you switch it to `throws`

File: Sources/Swift/Helper/InfoPlist/SentryInfoPlistKey.swift (hunk 24 above)
Comment (philprime) [reply Sources/Swift/Helper/InfoPlist/SentryInfoPlistKey.swift]: It's not used from Objective-C nor can Swift Enums be exposed to ObjC, so we don't need it

File: Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift (hunk 27 above)
Comment (philippsuess) [reply Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift:1]: I don't want to start an argument around YAGNI, interface overuse, and clean architecture right now. You can just keep it, if you like. Let's discuss that later.

File: Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift (hunk 28 above)
Comment (philippsuess) [reply Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapperProvider.swift:11]: Again, discussion around YAGNI, code style, and other stuff. IMO not required, because people forget to update docs when they change the code, and then it isn't very clear after all, but you can keep it if you want. We can discuss that later.

File: Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift (hunk 31 above)
Comment (philippsuess) [reply Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift]: Sorry missed that.

File: Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift (hunk 35 above)
Comment (philippsuess) [reply Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift]: Sorry, I could have been more specific.
```suggestion
        SentrySDKLog.warning("[Session Replay] Detected environment as unreliable, because no safety conditions met.")
```

Sources/Swift/SentryExperimentalOptions.swift
philippsuess: > Do you mean the /// instead of /**?

Yes, the link doesn't uses `///` all over the place, and I think that's what most people usually use. But this shouldn't block the PR.

philprime: Yeah, will align this in another PR

Sources/Sentry/SentryDependencyContainer.m
philippsuess: Yes, but please not right now.

Sources/Swift/Helper/InfoPlist/SentryInfoPlistKey.swift
philippsuess: But can't you use that enum them from Swift? If not, please add a comment why you can't use that from Swift. My alarm bells are rining when I see public code, that doesn't need to be public.

CHANGELOG.md
philprime: I added a suffix to the warning

Sources/Swift/Helper/InfoPlist/SentryInfoPlistWrapper.swift
philippsuess: Super weird, because the protocol has already
```
 func getAppValueString(for key: String) throws -> String
```

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
philippsuess: Nothing to worry about right now. I can follow up later.

Sources/Swift/Helper/InfoPlist/SentryInfoPlistKey.swift
philprime: Ah you might be correct that the `public` shouldn't be there anymore, or at least `@_spi(Private)` should be used. Checking...

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayEnvironmentChecker.swift
```diff
@@ -0,0 +1,182 @@
...
+    private func checkSDKVersion() -> Reliability {
+        // Check the LC_BUILD_VERSION load command in the main executable's Mach-O header
+        // to determine the SDK version used to build the app.
+        // SDK versions < 26.0 don't have Liquid Glass, so they're reliable.
+
+        // Get the main executable's Mach-O header
+        guard let header = _dyld_get_image_header(0) else {
+            SentrySDKLog.warning("[Session Replay] Failed to get Mach-O header - unclear")
+            return .unclear
+        }
```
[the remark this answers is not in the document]
philprime: @itaybre looks to me like we shouldn't use the `LC_BUILD_VERSION` to check the SDK version as a fallback.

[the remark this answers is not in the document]
philippsuess: `h`: Sorry I missed this one. Yes, please don't do this on the main thread. I recently fixed a bug that led to app hangs on start up https://github.com/getsentry/sentry-cocoa/pull/6181

philprime:
As mentioned in [this PR comment](https://github.com/getsentry/sentry-cocoa/pull/6389#discussion_r2426536565) we should not use `_dyld_get_image_header(0)` on the main thread because we already had to fix app hangs with https://github.com/getsentry/sentry-cocoa/pull/6181.

Therefore I am going to remove it from the PR now.

CHANGELOG.md
philprime: I am not sure how to validate this and how to cover all cases, so in interest of having a solution without further leaking issues, I will not exempt the case with both options turned off for now.

PR #6429, opened 2025-10-15
ref: Convert SentrySwizzleWrapper to Swift

noahsmartin:
Closes https://linear.app/getsentry/issue/COCOA-539/convert-sentryswizzlewrapper-to-swift

#skip-changelog


Sources/Swift/Helper/SentrySwizzleWrapper.swift
```diff
@@ -0,0 +1,41 @@
...
+import UIKit
+
+@_spi(Private) public typealias SentrySwizzleSendActionCallback = (String, Any?, Any?, UIEvent?) -> Void
+
+@_spi(Private) @objc public final class SentrySwizzleWrapper: NSObject {
+    
+    static var sentrySwizzleSendActionCallbacks = [String: SentrySwizzleSendActionCallback]()
+    
+    @objc public func swizzleSendAction(_ callback: @escaping SentrySwizzleSendActionCallback, forKey key: String) {
+        Self.sentrySwizzleSendActionCallbacks[key] = callback
```
itaybre, line 13: The behavior is a little different here (obj was retaining a copy), but I don't think it will make any difference, right?

noahsmartin: Right, in Swift you aren't even able to make a copy, IIRC it's a value type so automatically passed as a copy

PR #6430, opened 2025-10-15
ref: Remove SentrySystemWrapper from the dependency container

noahsmartin:
As @philipphofmann noticed, we don't need to have this be a property in the dependency container, which removes the need to convert it to Swift and closes https://linear.app/getsentry/issue/COCOA-537/convert-sentrysystemwrapper-to-swift

#skip-changelog


itaybre (approved):
LGTM

PR #6438, opened 2025-10-15
ref: Convert SentryCrash to swift

noahsmartin:
Closes https://linear.app/getsentry/issue/COCOA-703/convert-sentrycrash-to-swift

#skip-changelog

### DIFF
--- a/Sources/SentryCrash/Installations/SentryCrashInstallation.m
+++ b/Sources/SentryCrash/Installations/SentryCrashInstallation.m
@@ -33,6 +33,7 @@
 #import "SentryCrashNSErrorUtil.h"
 #import "SentryCrashReportFilterBasic.h"
 #import "SentryDependencyContainer.h"
+#import "SentrySwift.h"
 #import <objc/runtime.h>
 
 /** Max number of properties that can be defined for writing to the report */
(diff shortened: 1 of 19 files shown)


Sources/SentryCrash/Installations/SentryCrashInstallation.m
```diff
@@ -33,15 +33,16 @@
...
 
 /** Max number of properties that can be defined for writing to the report */
 #define kMaxProperties 500
 
 static CrashHandlerData *g_crashHandlerData;
 
-static void
-crashCallback(const SentryCrashReportWriter *writer)
+void
+sentry_crashCallback(const SentryCrashReportWriter *writer)
```
itaybre, line 45: Can't we keep this as static and pass it as an argument?

noahsmartin: @itaybre we can't, it needs to be visible to Swift code. Otherwise the type we pass in from the Swift API would need to be `Any` because Swift code can't have `SentryCrashReportWriter` in it's public API (It is imported as implementationOnly)

itaybre: Thanks for the clarification

PR #6439, opened 2025-10-16
feat: Add SPM sample

### LINKED ISSUE #6762: feat: Add SPM sample
> [!NOTE]
> The pull request "[feat: Add SPM sample](https://github.com/getsentry/sentry-cocoa/pull/6439)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

This adds a sample app that builds natively with SPM. I don't want to call this GA yet, but once we merge this it will be possible for anyone to build with SPM and I would update the docs to call it a pre-release. With a future major version we can update all the public API to be in Swift and then call it GA

#skip-changelog

noahsmartin:
This adds a sample app that builds natively with SPM. I don't want to call this GA yet, but once we merge this it will be possible for anyone to build with SPM and I would update the docs to call it a pre-release. With a future major version we can update all the public API to be in Swift and then call it GA

#skip-changelog

Closes #6762

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,18 +9,17 @@ import MSVCRT
 
 import PackageDescription
 
-let products: [Product] = [
+var products: [Product] = [
     .library(name: "Sentry", targets: ["Sentry", "SentryCppHelper"]),
     .library(name: "Sentry-Dynamic", targets: ["Sentry-Dynamic"]),
     .library(name: "Sentry-Dynamic-WithARM64e", targets: ["Sentry-Dynamic-WithARM64e"]),
     .library(name: "Sentry-WithoutUIKitOrAppKit", targets: ["Sentry-WithoutUIKitOrAppKit", "SentryCppHelper"]),
     .library(name: "Sentry-WithoutUIKitOrAppKit-WithARM64e", targets: ["Sentry-WithoutUIKitOrAppKit-WithARM64e", "SentryCppHelper"]),
     .library(name: "SentrySwiftUI", targets: ["Sentry", "SentrySwiftUI", "SentryCppHelper"]),
-    .library(name: "SentryDistribution", targets: ["SentryDistribution"]),
-    .library(name: "SentrySPM", type: .dynamic, targets: ["SentryObjc"])
+    .library(name: "SentryDistribution", targets: ["SentryDistribution"])
 ]
 
-let targets: [Target] = [
+var targets: [Target] = [
     .binaryTarget(
         name: "Sentry",
         url: "https://github.com/getsentry/sentry-cocoa/releases/download/9.0.0-alpha.0/Sentry.xcframework.zip",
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -114,7 +112,6 @@ if let env = env, String(cString: env, encoding: .utf8) == "1" {
                 .headerSearchPath("SentryCrash/Reporting/Filters"),
                 .headerSearchPath("SentryCrash/Reporting/Filters/Tools")])
     ])
```
itaybre: `l`: Can we move them up together with the other targets and products?

Sentry.xcodeproj/project.pbxproj
```diff
@@ -2856,6 +2856,7 @@
 				6304360C1EC05CEF00C4D3FA /* Frameworks */,
 				6327C5D41EB8A783004E799B /* Products */,
 				7D826E3C2390840E00EED93D /* Utils */,
+				FA0C83812EC3BE3A003271F8 /* Recovered References */,
```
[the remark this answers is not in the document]
itaybre: @noahsmartin seems worth looking at

itaybre (approved):
LGTM, with a comment

Samples/SPM/SPM/ContentView.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  ContentView.swift
+//  SPMTest
+//
+//  Created by Noah Martin on 10/15/25.
+//
+
```
philprime:
```suggestion
```

Samples/SPM/SPM/SPMTestApp.swift
```diff
@@ -0,0 +1,28 @@
+//
+//  SPMTestApp.swift
+//  SPMTest
+//
+//  Created by Noah Martin on 10/15/25.
+//
+
```
philprime:
```suggestion
```

Samples/SPM/SPM/SPMTestApp.swift
```diff
@@ -0,0 +1,28 @@
...
+import SentrySwift
+import SwiftUI
+
+@main
+struct SPMTestApp: App {
+  init() {
+    let options = Options()
+    options.enableAppHangTracking = true
+    options.dsn = "testing"
```
philprime: `m`: I believe we should set up all samples with the sentry-sdks DSN, you can use this for reference:

https://github.com/getsentry/sentry-cocoa/blob/4606e3591f4ad082b2dfb2dc876cdf4f6ea9bd2b/Samples/iOS-SwiftUI-Widgets/iOS-SwiftUI-Widgets/Sources/MainApp.swift#L10-L19

Package.swift
noahsmartin: I tried this but it breaks the script that modifies the package file, I think best to leave it like this for now to keep that script simpler

PR #6444, opened 2025-10-17
ref: Dependency container in Swift

noahsmartin:
Wraps up the Swift migration by making the dependency container in Swift.

There are a huge number of files that #import "SentryDependencyContainer.h" So to make this PR easier to review I've left that file there, just empty, and will remove it in a follow up PR.

#skip-changelog

### DIFF
--- a/Sources/Swift/SentryDependencyContainer.swift
+++ b/Sources/Swift/SentryDependencyContainer.swift
@@ -104,7 +104,7 @@ extension SentryFileManager: SentryFileManagerProtocol { }
         instanceLock.synchronized {
             instance.reachability.removeAllObservers()
 #if (os(iOS) || os(tvOS) || (swift(>=5.9) && os(visionOS))) && !SENTRY_NO_UIKIT
-            instance.framesTracker.stop()
+            instance._framesTracker?.stop()
 #endif
             instance = SentryDependencyContainer()
         }


Sources/Sentry/include/SentryDelayedFramesTracker.h
```diff
@@ -59,10 +65,10 @@ SENTRY_NO_INIT
  * @param isRunning Wether the frames tracker is running or not.
  * @param slowFrameThreshold The threshold for a slow frame. For 60 fps this is roughly 16.67 ms.
  */
-- (SentryFramesDelayResult *)getFramesDelay:(uint64_t)startSystemTimestamp
-                         endSystemTimestamp:(uint64_t)endSystemTimestamp
-                                  isRunning:(BOOL)isRunning
-                         slowFrameThreshold:(CFTimeInterval)slowFrameThreshold;
+- (SentryFramesDelayResultObjC *)getFramesDelayObjC:(uint64_t)startSystemTimestamp
```
itaybre: Do we need to rename it to `getFramesDelayObjC`?

Sources/Swift/Core/Integrations/ANR/SentryANRTracker.swift
```diff
@@ -1,29 +1,80 @@
...
+            value.helper != nil
+        }
+        let wrapped = DelegateWrapper(helper: listener)
+        mapping[ObjectIdentifier(listener)] = wrapped
+        helper.addListener(wrapped)
     }
     
     @objc(removeListener:) public func remove(listener: SentryANRTrackerDelegate) {
-        helper.removeListener(listener)
+        guard let mapped = mapping[ObjectIdentifier(listener)] else {
```
itaybre, line 26: I see you are removing entried without references in `add`, but why not in `remove`?

noahsmartin: My thought was that I didn't want it to grow unbounded, so as long as we clear out anything leftover anytime we add it will never get too big, if you also clear out when you remove it won't really make a difference on keeping it from getting too big

PR #6447, opened 2025-10-17
chore: Convert SentryFramesTracker to Swift

itaybre:
Converts SentryFramesTracker to Swift as part of our Swift Conversion

### DIFF
--- a/Sources/Sentry/SentryDependencyContainer.m
+++ b/Sources/Sentry/SentryDependencyContainer.m
@@ -1,6 +1,7 @@
 #import "SentryANRTrackerV1.h"
 
 #import "SentryDefaultThreadInspector.h"
+#import "SentryDelayedFramesTracker.h"
 #import "SentryInternalCDefines.h"
 #import "SentryInternalDefines.h"
 #import "SentryLogC.h"


Sources/Sentry/Profiling/SentryContinuousProfiler.mm
```diff
@@ -16,12 +16,15 @@
...
 
 #    if SENTRY_HAS_UIKIT
-#        import "SentryFramesTracker.h"
+#        import "SentrySwift.h"
 #        import <UIKit/UIKit.h>
 #    endif // SENTRY_HAS_UIKIT
 
 #    pragma mark - Private
 
+NSString *const kSentryNotificationContinuousProfileStarted
```
itaybre, line 24: Had to make this change otherwise tests weren't building, even if I imported `SentryContinousProfiler.h`

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -0,0 +1,380 @@
...
+        )
+
+        lock.synchronized {
+            listeners.removeAllObjects()
+        }
+    }
+
+    @objc
+    public func currentFrames() -> SentryScreenFrames {
+#if os(iOS)
```
noahsmartin, line 125: Shouldn't this be iOS || macOS?

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -0,0 +1,380 @@
...
+        dispatchQueueWrapper.dispatchAsyncOnMainQueueIfNotMainThread {
+            self.listeners.remove(listener)
+        }
+    }
+    
+    deinit {
+        stop()
+    }
+
+#if os(iOS)
```
noahsmartin, line 171: Same here, I think profiling is supported on macos

PR #6452, opened 2025-10-17
ref: Refactor app state manager to not use protocol

### LINKED ISSUE #6453: ref: Refactor app state manager to not use protocol
> [!NOTE]
> The pull request "[ref: Refactor app state manager to not use protocol](https://github.com/getsentry/sentry-cocoa/pull/6452)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

This also needs to use a class instead of a protocol to finish the swift conversion

#skip-changelog

noahsmartin:
This also needs to use a class instead of a protocol to finish the swift conversion

#skip-changelog

Closes #6453


itaybre (approved):
LGTM

PR #6454, opened 2025-10-18
ref: Remove v9 checks for debug image provider

### LINKED ISSUE #6455: ref: Remove v9 checks for debug image provider
> [!NOTE]
> The pull request "[ref: Remove v9 checks for debug image provider](https://github.com/getsentry/sentry-cocoa/pull/6454)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

The V9 migration makes this class internal instead of public, which means it needs to be migrated to Swift to be part of the Swift dependency container. This removes the V9 checks and migrates it to Swift

#skip-changelog

noahsmartin:
The V9 migration makes this class internal instead of public, which means it needs to be migrated to Swift to be part of the Swift dependency container. This removes the V9 checks and migrates it to Swift

#skip-changelog

Closes #6455


.github/workflows/test-cross-platform.yml
```diff
@@ -55,7 +55,7 @@ jobs:
         with:
           repository: getsentry/sentry-react-native
           path: sentry-react-native
-          ref: itay/frames_tracker_swift
+          ref: debugImageProviderSwift
```
philprime: `m`: IIUC this must be reverted before merging it, correct?

Sources/Resources/Sentry.modulemap
```diff
@@ -10,7 +10,6 @@ framework module Sentry {
         header "PrivatesHeader.h"
         header "SentryAppStartMeasurement.h"
         header "SentryBreadcrumb+Private.h"
-        header "SentryDebugImageProvider+HybridSDKs.h"
```
philprime, line 13: `m`: Did you check that this doesn't break hybrid SDKs?

.github/workflows/test-cross-platform.yml
noahsmartin: Just reverted it 👍

Sources/Resources/Sentry.modulemap
noahsmartin: Yep, the API is the same with @objc now, verified it works in the RN sample

PR #6460, opened 2025-10-20
ref: Remove deprecated app start option

### LINKED ISSUE #6461: ref: Remove deprecated app start option
> [!NOTE]
> The pull request "[ref: Remove deprecated app start option](https://github.com/getsentry/sentry-cocoa/pull/6460)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Fully removes "enableAppLaunchProfiling" which is removed in V9

#skip-changelog

noahsmartin:
Fully removes "enableAppLaunchProfiling" which is removed in V9

#skip-changelog

Closes #6461


noahsmartin:
@philprime I was surprised as well, it seems like it was only retained to keep the public API from breaking

PR #6465, opened 2025-10-20
ref: Convert SentryANRTracker to class

### LINKED ISSUE #6466: ref: Convert SentryANRTracker to class
> [!NOTE]
> The pull request "[ref: Convert SentryANRTracker to class](https://github.com/getsentry/sentry-cocoa/pull/6465)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

I pulled this change out of the overall dependency container Swift refactor to make that PR easier to review

#skip-changelog

noahsmartin:
I pulled this change out of the overall dependency container Swift refactor to make that PR easier to review

#skip-changelog

Closes #6466

### DIFF
--- a/Sources/Sentry/SentryANRTrackerV1.m
+++ b/Sources/Sentry/SentryANRTrackerV1.m
@@ -13,7 +13,7 @@ typedef NS_ENUM(NSInteger, SentryANRTrackerState) {
     kSentryANRTrackerStopping
 };
 
-@interface SentryANRTrackerV1 () <SentryANRTracker>
+@interface SentryANRTrackerV1 ()
 
 @property (nonatomic, strong) SentryCrashWrapper *crashWrapper;
 @property (nonatomic, strong) SentryDispatchQueueWrapper *dispatchQueueWrapper;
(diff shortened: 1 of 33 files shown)


Sources/Sentry/SentryANRTrackerV1.m
```diff
@@ -28,6 +28,15 @@ @implementation SentryANRTrackerV1 {
     SentryANRTrackerState state;
 }
 
+- (instancetype)initWithTimeoutInterval:(NSTimeInterval)timeoutInterval
+{
+    return
+        [self initWithTimeoutInterval:timeoutInterval
+                         crashWrapper:SentryDependencyContainer.sharedInstance.crashWrapper
+                 dispatchQueueWrapper:SentryDependencyContainer.sharedInstance.dispatchQueueWrapper
+                        threadWrapper:SentryDependencyContainer.sharedInstance.threadWrapper];
```
itaybre, line 37: Is this safe?
I know we can't use the DependencyContainer too early in the initialization

Sources/Sentry/SentryANRTrackerV1.m
```diff
@@ -28,6 +28,15 @@ @implementation SentryANRTrackerV1 {
     SentryANRTrackerState state;
 }
 
+- (instancetype)initWithTimeoutInterval:(NSTimeInterval)timeoutInterval
```
itaybre, line 31: Where is this being used?

noahsmartin: No where yet, but the intention of this PR is to prepare for the swift conversion PR without adding too much boilerplate there

noahsmartin: It’s because when this is initialized from swift you can’t pass swift types through the objc header

Sources/Sentry/SentryANRTrackerV1.m
noahsmartin: Yes, it’s not called during the dependency container init, it’s called in a function on the dep container. If it wasn’t safe the code would assert right away (that assert is in the dep container)

Issue #3535, opened 2024-01-04
Turn on enablePreWarmedAppStartTracing by default


philipphofmann:
### Description

Turning on `enablePreWarmedAppStartTracing` impacts the app start duration, which is a breaking change. Once the app start module of Mobile Starfish handles different app types properly we can enable this by default.

PR #6508, opened 2025-10-23
feat: enablePreWarmedAppStartTracing by default

philipphofmann:
## :scroll: Description

Set the default value of enablePreWarmedAppStartTracing to true.

Docs PR https://github.com/getsentry/sentry-docs/pull/15295.

Fixes GH-3535

Unit tests.

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Helper/SentryEnabledFeaturesBuilder.swift
+++ b/Sources/Swift/Helper/SentryEnabledFeaturesBuilder.swift
@@ -17,14 +17,6 @@ import Foundation
         if options.enableTimeToFullDisplayTracing {
             features.append("timeToFullDisplayTracing")
         }
-
-#if os(iOS) || os(tvOS)
-#if canImport(UIKit) && !SENTRY_NO_UIKIT
-        if options.enablePreWarmedAppStartTracing {
-            features.append("preWarmedAppStartTracing")
-        }
-#endif // canImport(UIKit)
-#endif // os(iOS) || os(tvOS)
         
         if options.swiftAsyncStacktraces {
             features.append("swiftAsyncStacktraces")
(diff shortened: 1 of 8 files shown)


Sources/Swift/Helper/SentryEnabledFeaturesBuilder.swift
```diff
@@ -17,14 +17,6 @@ import Foundation
         if options.enableTimeToFullDisplayTracing {
             features.append("timeToFullDisplayTracing")
         }
-
-#if os(iOS) || os(tvOS)
-#if canImport(UIKit) && !SENTRY_NO_UIKIT
-        if options.enablePreWarmedAppStartTracing {
-            features.append("preWarmedAppStartTracing")
-        }
```
philippsuess, line 25: I removed it here, as when it's enabled by default, it doesn't really make sense anymore to track adoption. This only increases the payload size without providing any real benefit.

Sources/Swift/Helper/SentrySdkInfo.swift
```diff
@@ -54,12 +54,7 @@ import Foundation
     
     @objc public convenience init(withOptions options: Options?) {
         let features = SentryEnabledFeaturesBuilder.getEnabledFeatures(options: options)
-        var integrations = SentrySDKInternal.currentHub().trimmedInstalledIntegrationNames()
-        #if (os(iOS) || os(tvOS) || (swift(>=5.9) && os(visionOS))) && !SENTRY_NO_UIKIT
-            if options?.enablePreWarmedAppStartTracing ?? false {
-                integrations.append("PreWarmedAppStartTracing")
-            }
```
philippsuess, line 61: I removed it here, as when it's enabled by default, it doesn't really make sense anymore to track adoption. This only increases the payload size without providing any real benefit.

Sources/Swift/Helper/SentryEnabledFeaturesBuilder.swift
philprime: We could leave it in to see if we see a difference in adoption to v9. No strong opinion here.

philippsuess: It's enabled by default. What we would track is people disabling the feature, which I hope not many people do. We can always add it again if we see it's valuable. I would rather ditch it to remove some code and reduce the payload size.

PR #6517, opened 2025-10-23
fix: Prevent duplicate issue creation in PR automation workflow

philprime:
The workflow was creating duplicate issues when multiple trigger events (opened, edited, synchronize, etc.) fired in quick succession. Both workflow runs would check the PR simultaneously, find no issue reference, and each create a new issue.

Added a check to search for existing issues created by this automation before creating a new one. The workflow now searches for open issues that contain the PR URL and were created by github-actions bot, preventing duplicate issue creation.

Fixes the issue where PR #6386 had two issues created (#6499 and #6500).

#skip-changelog


philprime:
@sentry review

Issue #6491, opened 2025-10-21
Expose SentryAttachment's attachmentType for Internal SDKs


itaybre:
### Description

Some SDKs rely on sentry-cocoa internally but implement their own logic to retrieve the view hierarchy.
However, this approach requires using a specific attachmentType that is currently not exposed, preventing these SDKs from fully integrating their custom implementations.

## Proposed Solution
Expose the required attachmentType so that SDKs with custom view hierarchy implementations can properly attach and send this data.

PR #6521, opened 2025-10-23
feat: Expose attachment type on SentryAttachment

itaybre:
This PR adds a new method in PrivateSentrySDKOnly for SDKs which use our SDK can add a custom view hierarchy.

Fixes: #6491

### DIFF
--- a/Sources/Sentry/Public/SentryAttachment.h
+++ b/Sources/Sentry/Public/SentryAttachment.h
@@ -10,9 +10,24 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  * Attachment Type
+ *
+ * This enum specifies the type of attachment. The attachment type is primarily used by downstream
+ * SDKs (such as sentry-godot) to distinguish between different attachment categories.
+ * Most applications using the SDK directly do not need to specify this, as the default
+ * @c kSentryAttachmentTypeEventAttachment is used for regular file attachments.
+ *
+ *
+ * See also: https://develop.sentry.dev/sdk/data-model/envelope-items/#attachment
  */
 typedef NS_ENUM(NSInteger, SentryAttachmentType) {
+    /**
+     * Standard event attachment. This is the default type for user-provided attachments.
+     */
     kSentryAttachmentTypeEventAttachment,
+    /**
+     * View hierarchy attachment. Automatically set by the SDK when capturing view hierarchy data.
+     * This type is primarily used by downstream SDKs.
+     */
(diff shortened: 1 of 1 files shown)


noahsmartin (commented):
Is it not entirely possible to achieve the same behavior today without adding to this private API? You can call SentrySDK.configureScope { ... } Wouldn't that allow the Godot SDK to do the same thing?

IMO we should try to have fewer hybrid SDK specific APIs, and move towards a version where anything that needs to be visible to an SDK should be part of the public API. So my suggestion is just to double check, do we really need to add an API for this?

itaybre:
> Is it not entirely possible to achieve the same behavior today without adding to this private API? You can call SentrySDK.configureScope { ... } Wouldn't that allow the Godot SDK to do the same thing?
> 
> IMO we should try to have fewer hybrid SDK specific APIs, and move towards a version where anything that needs to be visible to an SDK should be part of the public API. So my suggestion is just to double check, do we really need to add an API for this?

The current blocker is SentryAttachment's attachmentType being [private only](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/include/SentryAttachment%2BPrivate.h#L46)

noahsmartin:
Makes sense thanks for explaining @itaybre Maybe we should make the attachment type property visible to hybrid SDKs, or visible to everyone? That way the extra logic that would never get called from regular users of the cocoa SDK doesn't have to be included in every native app and can instead be in the Godot repo, seems more scalable/better abstracted that way

philippsuess:
> Maybe we should make the attachment type property visible to hybrid SDKs, or visible to everyone?

We didn't decide on this yet, but I think we should come up with proper APIs also for hybrid. This visible only to hybrid only causes problems. We quite often break them. We should treat hybrid as any other SDK user. So IMO we should make the things they need public with a proper API and only change these APIs when doing a major.

Sources/Sentry/Public/SentryAttachment.h
```diff
@@ -8,6 +8,14 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ * Attachment Type
+ */
```
philippsuess, line 21: `m`: I think it would make sense to explain a bit what these attachment types are and to explain clearly that the typical user doesn't need to worry about this as mostly HybridSDKs use it. The same applies for the init methods.

philipphofmann (approved):
Thanks @itaybre.

PR #6528, opened 2025-10-24
ref: Remove SentrySerializable on user

### LINKED ISSUE #6529: ref: Remove SentrySerializable on user
> [!NOTE]
> The pull request "[ref: Remove SentrySerializable on user](https://github.com/getsentry/sentry-cocoa/pull/6528)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

#skip-changelog

noahsmartin:
I couldn't quite keep the existing V9 code because they failed the cocoapods build. But with these changes to use the function name instead of the protocol it compiles in all cases and makes this internal implementation detail not part of the public API

#skip-changelog

Closes #6529


philprime (approved):

noahsmartin:
This isn't working with the cocoapods build so I'm going to convert it back to draft for now

PR #6538, opened 2025-10-24
fix: Check SessionReplay json's path before reading it

itaybre:
Fixes a crash when SessionReplay json file does not contain a path field


philprime (approved):
LGTM, we should never assume fields to be set in a dictionary

itaybre:
Closing this since it was merged to branch v8

PR #6539, opened 2025-10-24
fix: Disable SessionSentryReplayIntegration if the environment is unsafe

itaybre:
This PR disable the SessionReplayIntegration instead of disabling SentrySessionReplay to avoid running any SessionReplay logic if not needed.


Sources/Swift/Helper/Dependencies.swift
```diff
@@ -5,7 +5,7 @@
     @objc public static let threadWrapper = SentryThreadWrapper()
     @objc public static let processInfoWrapper: SentryProcessInfoSource = ProcessInfo.processInfo
     static let infoPlistWrapper: SentryInfoPlistWrapperProvider = SentryInfoPlistWrapper()
-    @objc public static let sessionReplayEnvironmentChecker: SentrySessionReplayEnvironmentChecker = {
+    @objc public static var sessionReplayEnvironmentChecker: SentrySessionReplayEnvironmentCheckerProvider = {
```
noahsmartin, line 8: Only a `let` is thread safe. If you need it to override the value for testing I think it's better to pass the dependencies as a parameter to the class you're testing, rather than having it used from here.

Or if that's too hard to refactor just move it to SentryDependencyContainer and use the lock with a lazy var

itaybre: This is only for testing. 

Moving this to `SentryDependencyContainer` won't work on the v8 branch, which also needs this fix

philprime:
We need to do this fix from the branch `8.57.0`, not from `main`

itaybre:
Closing this since it was merged to branch v8

PR #6549, opened 2025-10-28
test: Adjust data count range in SpotlightTests

### LINKED ISSUE #6550: test: Adjust data count range in SpotlightTests
> [!NOTE]
> The pull request "[test: Adjust data count range in SpotlightTests](https://github.com/getsentry/sentry-cocoa/pull/6549)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Updated the expected data count range in the
SentrySpotlightTransportTests to allow for a wider variance in GZip compression results, changing the lower and upper bounds from ±20 to ±40 to avoid flakiness.

Test failed here https://github.com/getsentry/sentry-cocoa/actions/runs/18778698663/job/53579222897.

#skip-changelog

philipphofmann:
Updated the expected data count range in the
SentrySpotlightTransportTests to allow for a wider variance in GZip compression results, changing the lower and upper bounds from ±20 to ±40 to avoid flakiness.

Test failed here https://github.com/getsentry/sentry-cocoa/actions/runs/18778698663/job/53579222897.

#skip-changelog 

Closes #6550


philippsuess:
> LGTM, but to be honest this is a wide range and it seems that this test highly unstable. We might want to explore if GZip can be configured to be deterministic

True, but honestly, I don't want to invest that time right now for Spotlight. I don't think many people are using it. At least I can't remember a user opening an issue or asking questions about it.

philprime:
I was even thinking about suggesting we remove the tests of Spotlight instead of fixing them.

philippsuess:
> I was even thinking about suggesting we remove the tests of Spotlight instead of fixing them.

Yes, me too. We can ditch this single test case if it still breaks in the future.

Issue #6560, opened 2025-10-28
Deadlock when handling crashes


noahsmartin:
### Platform

iOS

### Environment

Production

### Installed

Swift Package Manager

### Version

Current

### Xcode Version

All

### Did it work on previous versions?

No

### Steps to Reproduce

There is a deadlock when the SDK attempts to write out a crash report if any other thread holds the dyld lock. This means when a crash occurs the SDK sometimes freezes and never writes the crash report, causing crashes to be dropped.

To reproduce, run this function:
```
  func hammerDyldImageCount(workers: Int = 10) {
    let q = DispatchQueue(label: "test", attributes: [.concurrent, .initiallyInactive])

      for _ in 0..<workers {
        q.async {
          var checksum: UInt64 = 0
          while true {
            let count = _dyld_image_count()
            for j in 0..<count {
              let header = _dyld_get_image_header(j)
              checksum &+= UInt64(UInt(bitPattern: header))
            }
          }
        }
      }
    q.activate()
  }
```
And then crash the app. I did this by adding this code into the iOS-Swift sample app and pressing "Crash the app". The debugger must be disconnected when you run this.

This is the same root cause as here: https://github.com/getsentry/sentry-cocoa/issues/1892#issuecomment-1171254043 and here: https://github.com/getsentry/sentry-cocoa/issues/5609 although in that issue I thought it was debug only. It's also the same root cause as: https://github.com/getsentry/sentry-cocoa/issues/4056 that specific cause was fixed (for non-debug) but the codepath is still reachable with the steps described here.

The cleanest solution might be to delete all symbolication from the SDK, since it's rarely safe to use or beneficial (because it requires dsyms to really work). That could be done in the upcoming major version release and if in-process symbolication needs to be added back for some reason we can take the time to do it in a single more robust way.

### Expected Result

The app crash should trigger the SDK to write a payload to disk, and then the app should exit and the system crash dialog gets presented.

### Actual Result

The app deadlocks and has to be force quit. It happens about 1/3 times for me.

philippsuess:
> The cleanest solution might be to delete all symbolication from the SDK, since it's rarely safe to use or beneficial (because it requires dsyms to really work).

We already discussed this multiple times in the past. Now that this issue is present, it's time to do it. Local symbolication causes more trouble. The only thing we still could discuss if we want to keep it behing an optional opt in feature flag or when having debug = true, but it should be off everywhere by default.

PR #6562, opened 2025-10-28
fix: Remove symbolication that causes deadlocks

### LINKED ISSUE #6698: fix: Remove symbolication that causes deadlocks
> [!NOTE]
> The pull request "[fix: Remove symbolication that causes deadlocks](https://github.com/getsentry/sentry-cocoa/pull/6562)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Resolves https://github.com/getsentry/sentry-cocoa/issues/6560 (and all the other related issues)

This also has the benefit of unifying the debug and non-debug codepaths to reduce the chances that a bug affects one environment but not the other. For example, the debug behavior of buildStacktraceForCurrentThread would remove the frame for that function itself, but this wouldn't happen in non-debug (since it relied on symbolication). This PR preserves the non-debug behavior, now in all configurations

#skip-changelog

noahsmartin:
Resolves https://github.com/getsentry/sentry-cocoa/issues/6560 (and all the other related issues)

This also has the benefit of unifying the debug and non-debug codepaths to reduce the chances that a bug affects one environment but not the other. For example, the debug behavior of buildStacktraceForCurrentThread would remove the frame for that function itself, but this wouldn't happen in non-debug (since it relied on symbolication). This PR preserves the non-debug behavior, now in all configurations

#skip-changelog

Closes #6698

### DIFF
--- a/Tests/SentryTests/SentryCrash/SentryStacktraceBuilderTests.swift
+++ b/Tests/SentryTests/SentryCrash/SentryStacktraceBuilderTests.swift
@@ -58,6 +58,7 @@ class SentryStacktraceBuilderTests: XCTestCase {
         // Make sure the first 4 frames contain an address close to the main function
         let isMainInFirstFrames = actual.frames[...3].contains(where: { frame in
             let inst = Int(frame.instructionAddress?.replacingOccurrences(of: "0x", with: "") ?? "", radix: 16) ?? 0
+            // Use the symbol name for the first stack entry in dyld
             #if targetEnvironment(simulator)
             return name(for: inst) == "start_sim"
             #else


Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -307,15 +307,13 @@ - (SentryFrame *)stackFrameAtIndex:(NSInteger)frameIndex inThreadIndex:(NSIntege
         = (uintptr_t)[frameDictionary[@"instruction_addr"] unsignedLongLongValue];
     NSDictionary *binaryImage = [self binaryImageForAddress:instructionAddress];
     SentryFrame *frame = [[SentryFrame alloc] init];
-    frame.symbolAddress = sentry_formatHexAddress(frameDictionary[@"symbol_addr"]);
+    // The symbol address is not known, so it defaults to 0
+    frame.symbolAddress = @"0x0000000000000000";
```
philippsuess: `m`: We usually then set these things to `nil` and let the backend set default values, because it gives us more flexibility. Or do we absolutely need to set this to `@"0x0000000000000000"`?

Sources/SentryCrash/Recording/SentryCrashReport.c
```diff
@@ -799,20 +799,6 @@ writeBacktrace(const SentryCrashReportWriter *const writer, const char *const ke
...
-                            writer->addStringElement(writer, SentryCrashField_ObjectName,
-                                sentrycrashfu_lastPathEntry(stackCursor->stackEntry.imageName));
-                        }
-                        writer->addUIntegerElement(writer, SentryCrashField_ObjectAddr,
-                            stackCursor->stackEntry.imageAddress);
-                        if (stackCursor->stackEntry.symbolName != NULL) {
-                            writer->addStringElement(writer, SentryCrashField_SymbolName,
-                                stackCursor->stackEntry.symbolName);
-                        }
-                        writer->addUIntegerElement(writer, SentryCrashField_SymbolAddr,
```
philippsuess, line 813: `m`: We should be able to remove also `SentryCrashField_ObjectName`, `SentryCrashField_ObjectAddr`, `SentryCrashField_SymbolName`, `SentryCrashField_SymbolAddr` completely from the codebase.

philipphofmann (commented):
We definitely need a clear changelog entry to explain that local symbolication is gone now, and we should add a decision log entry here https://github.com/getsentry/sentry-cocoa/blob/main/develop-docs/DECISIONS.md to explain why we removed local symbolication.

Also, we must check our onboarding here https://docs.sentry.io/platforms/apple/guides/ios/ and the wizard if we shouldn't mention that you need to upload dSYMs to guarantee a great experience. Otherwise, some users might be frustrated during the onboarding, because it usuallally sets `options.debug = true`.

And huge thanks for doing this @noahsmartin

Sources/Sentry/SentryCrashReportConverter.m
noahsmartin: I was trying to leave behavior completely unchanged for when debug = False, which was to set this to this string

Sources/SentryCrash/Recording/SentryCrashReport.c
noahsmartin: I actually did that and then undid it because it made the PR even larger, I will re-do it though

noahsmartin:
Thanks @philipphofmann just made those changes?

develop-docs/DECISIONS.md
```diff
@@ -1,5 +1,13 @@
 # Decision Log
 
+## No local symbolication of crashes
+
+Date: Nov 7th, 2025
+Contributors: @noahsmartin, @philipphofmann
+
+We decided to remove local symbolication when capturing a crash. The existing local symbolication was not signal-safe and caused deadlocks (https://github.com/getsentry/sentry-cocoa/issues/6560).
```
philippsuess: `m`: It's not just about crashes. It's completely removing it, if I'm not mistaken.
```suggestion
We decided to remove local symbolication. The existing local symbolication was not signal-safe and caused deadlocks (https://github.com/getsentry/sentry-cocoa/issues/6560).
```

Sources/Sentry/SentryCrashReportConverter.m
philippsuess: Please remove it and let it `nil`, I don't think that's required. At least I didn't find anything in our develop docs or in our repos https://github.com/search?q=repo%3Agetsentry%2Fsentry%200x0000000000000000&type=code. I would try sending a frame with the symbol address manually setting to `nil` in `beforeSend` and check what happens. If it makes a difference, we can consider adding it back.

develop-docs/DECISIONS.md
noahsmartin: I thought of the removal of the rest of it as more of an implementation detail that is more likely to change in the future. Like one day we may add back symbolication on the next launch of the app (assuming it didn't update and change the binary) but maybe I overcomplicated it, will change it

Tests/SentryTests/SentryCrash/SentryDefaultThreadInspectorTests.swift
```diff
@@ -48,74 +46,14 @@ class SentryDefaultThreadInspectorTests: XCTestCase {
...
-    func testGetCurrentThreadsWithStacktrace_WithSymbolication() {
+    func testGetCurrentThreadsWithStacktrace() {
         let queue = DispatchQueue(label: "test-queue", attributes: [.concurrent, .initiallyInactive])
         
         let expect = expectation(description: "Read every thread")
         expect.expectedFulfillmentCount = 10
         
         let sut = self.fixture.getSut(testWithRealMachineContextWrapper: true)
 
-        for _ in 0..<10 {
```
philippsuess, line 59: I removed this test as we had two of these. One withSymbolication and one without. As we don't have symbolication anymore, we can delete one.

Tests/SentryTests/SentryCrash/SentryStacktraceBuilderTests.swift
```diff
@@ -46,33 +45,28 @@ class SentryStacktraceBuilderTests: XCTestCase {
...
         let actual = fixture.sut.buildStacktraceForCurrentThread()
 
         // -- Assert --
-        // Make sure the first 4 frames contain main
-        let isMainInFirstFrames = actual.frames[...3].contains(where: { $0.function == "main" })
+        // Make sure the first 4 frames contain an address close to the main function
+        let isMainInFirstFrames = actual.frames[...3].contains(where: { frame in
+            let inst = Int(frame.instructionAddress?.replacingOccurrences(of: "0x", with: "") ?? "", radix: 16) ?? 0
+            #if targetEnvironment(simulator)
+            return name(for: inst) == "start_sim"
```
philprime, line 63: `m`: Couldn't this break in future simulator versions? If yes, we should add a comment.

philprime (approved):
LGTM

PR #6563, opened 2025-10-28
fix: Resolve remaining nullability warnings; enabling them as errors

### LINKED ISSUE #4012: Make SentryException value and type nullable
### Description

Only one of `SentryException.value` or `SentryException.type` is required to be nonnull; see [Relay](https://github.com/getsentry/relay/blob/5942007922cac572002ed0c89ee47a13b4c8e959/relay-event-schema/src/protocol/exception.rs#L35-L46). The `value` can already be `nil` when converting an NSException to a SentryException in the client as NSException.reason is nullable. https://github.com/getsentry/sentry-cocoa/blob/cf972098baef6f2042eee18549c33451fd983731/Sources/Sentry/SentryClient.m#L208-L210

Therefore, we should make both `value` and `type` nullable readonly and add constructors to ensure either one of the two is set. Consider converting this class to Swift when doing these changes.

### LINKED ISSUE #6572: fix: Resolve remaining nullability warnings; enabling them as errors
> [!NOTE]
> The pull request "[fix: Resolve remaining nullability warnings; enabling them as errors](https://github.com/getsentry/sentry-cocoa/pull/6563)" was created by @philprime but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

### Description

- Uses the `SENTRY_UNWRAP_NULLABLE` and `SENTRY_UNWRAP_NULLABLE_VALUE` macros to resolve null handling warnings.
- Where possible I added proper handling
- For files which existed and worked properly for a long time (e.g. `SentryNetworkTracker.m`) I only silenced the warnings.
- Enables null handling errors by removing the `no-error` C flags

### Testing

- Tested using unit tests

#skip-changelog

philprime:
### Description

- Uses the `SENTRY_UNWRAP_NULLABLE` and `SENTRY_UNWRAP_NULLABLE_VALUE` macros to resolve null handling warnings.
- Where possible I added proper handling
- For files which existed and worked properly for a long time (e.g. `SentryNetworkTracker.m`) I only silenced the warnings.
- Enables null handling errors by removing the `no-error` C flags

### Testing

- Tested using unit tests

Closes #6572
Closes #4012

### DIFF
--- a/Sources/Sentry/SentryClient.m
+++ b/Sources/Sentry/SentryClient.m
@@ -202,8 +202,8 @@ NSString *const DropSessionLogMessage = @"Session has no release name. Won't sen
 - (SentryEvent *)buildExceptionEvent:(NSException *)exception
 {
     SentryEvent *event = [[SentryEvent alloc] initWithLevel:kSentryLevelError];
-    SentryException *sentryException =
-        [[SentryException alloc] initWithValue:exception.reason ?: @"" type:exception.name ?: @""];
+    SentryException *sentryException = [[SentryException alloc] initWithValue:exception.reason
+                                                                         type:exception.name];
 
     event.exceptions = @[ sentryException ];
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Remove unused `SentryFrame.instruction` property (#6504)
 - Remove `uuid` and `name` of `SentryDebugMeta` (#6512) Use `debugID` instead of `uuid` and `codeFile` instead of `name`.
 - Enable enablePreWarmedAppStartTracing by default (#6508). With this option enabled, the SDK collects [prewarmed app starts](https://docs.sentry.io/platforms/apple/tracing/instrumentation/automatic-instrumentation/#prewarmed-app-start-tracing).
+- Change `value` and `type` of `SentryException` to be nullable (#6563)
 
 ### Features
 
--- a/Sources/Sentry/Public/SentryException.h
+++ b/Sources/Sentry/Public/SentryException.h
@@ -21,12 +21,12 @@ SENTRY_NO_INIT
 /**
  * The name of the exception
  */
-@property (nonatomic, copy) NSString *value;
+@property (nonatomic, copy) NSString *_Nullable value;
 
 /**
  * Type of the exception
  */
-@property (nonatomic, copy) NSString *type;
+@property (nonatomic, copy) NSString *_Nullable type;
 
 /**
  * Additional information about the exception
@@ -49,12 +49,13 @@ SENTRY_NO_INIT
 @property (nonatomic, strong) SentryStacktrace *_Nullable stacktrace;
 
 /**
- * Initialize an SentryException with value and type
- * @param value String
- * @param type String
+ * Initialize an SentryException with value and type.
+ * @param value Nullable string describing the exception
+ * @param type Nullable string with the type of the exception
  * @return SentryException
+ * @note At least one of value or type must be non-nil. This is asserted in debug builds.
  */
-- (instancetype)initWithValue:(NSString *)value type:(NSString *)type;
+- (instancetype)initWithValue:(NSString *_Nullable)value type:(NSString *_Nullable)type;
 
 @end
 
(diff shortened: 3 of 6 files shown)


Sources/Sentry/SentryClient.m
```diff
@@ -202,8 +203,8 @@ - (SentryId *)captureException:(NSException *)exception
 - (SentryEvent *)buildExceptionEvent:(NSException *)exception
 {
     SentryEvent *event = [[SentryEvent alloc] initWithLevel:kSentryLevelError];
-    SentryException *sentryException = [[SentryException alloc] initWithValue:exception.reason
-                                                                         type:exception.name];
+    SentryException *sentryException = [[SentryException alloc] initWithValue:exception.reason ?: @""
```
philprime: Not sure what we should use here for fallback to keep grouping, open for suggestions

Sources/Sentry/SentryDsn.m
```diff
@@ -74,7 +74,7 @@ - (NSURL *)getBaseEndpoint
     components.host = url.host;
     components.port = url.port;
     components.path = [NSString stringWithFormat:@"%@/api/%@/", path, projectId];
-    return components.URL;
+    return SENTRY_UNWRAP_NULLABLE(NSURL, components.URL);
```
philprime, line 77: To be honest, this is unsafe, but I think it's overkill to change this right now. Has been this way for a while

Sources/Sentry/SentryClient.m
```diff
@@ -202,8 +202,8 @@ - (SentryId *)captureException:(NSException *)exception
 - (SentryEvent *)buildExceptionEvent:(NSException *)exception
 {
     SentryEvent *event = [[SentryEvent alloc] initWithLevel:kSentryLevelError];
-    SentryException *sentryException = [[SentryException alloc] initWithValue:exception.reason
-                                                                         type:exception.name];
+    SentryException *sentryException =
+        [[SentryException alloc] initWithValue:exception.reason ?: @"" type:exception.name ?: @""];
```
itaybre: Does an exception like this makes sense?
No name nor reason?

Sources/Sentry/SentryDsn.m
itaybre: This will crash, right?
I think it is better for the SDK to crash in this case since the SDK won't work either way and users will notice the problem

philprime: Yes I believe it will crash with a force-null-unwrapping, but this only happens if the user injects a DSN which is invalid, that's why I believe it's fine to do it that way.

Sources/Sentry/SentryClient.m
philprime: Not really, but I also don't know if this ever happens. I see three options:

1. Use a default like I added in the proposed changes
2. Add proper null-handling, returning nil in the method if the values are not available
3. Silence the warnings

philippsuess: We read from the NSException here. If it really has `nil` for the reason, we must still report it, even if it doesn't really make sense. Nobody stops users from calling:
```swift
client.capture(exception: NSException(name: NSExceptionName("Hy"), reason: nil, userInfo: nil))
```

According to [Relay](https://github.com/getsentry/relay/blob/cf80d55e6f9b6b5cccfa4c4f06c966ffe4a22bfe/relay-event-schema/src/protocol/exception.rs#L32-L43) at least one type or value is required. So Sentry can accept exceptions with only a type or only a value.

Java has both types nullable https://github.com/getsentry/sentry-java/blob/2124a46eebc472f6610b4d99a7c906d92eb47c02/sentry/src/main/java/io/sentry/protocol/SentryException.java#L32-L44

So I think we should fix that in V9 and make these nullable as well, and don't provide default values of `""`.

philippsuess: I  noticed that the [develop docs](https://develop.sentry.dev/sdk/data-model/event-payloads/exception/) don't clearly state this so I opened a PR https://github.com/getsentry/sentry-docs/pull/15344

philprime: Thanks for the feedback, in that case I am going to change the method signature to only require at least one of the two values to be non-nullable

philprime: I updated the files and added a changelog entry

philprime: Let me know if you want it to be changed in a different PR

philprime:
I just noticed that changing the SentryException is actually todo in #4012, so this PR will also close that

Sources/Sentry/SentryClient.m
philippsuess: Nope this PR is fine for me.

PR #6573, opened 2025-10-28
fix: Disable SessionSentryReplayIntegration if the environment is unsafe

itaybre:
This PR disable the SessionReplayIntegration instead of disabling SentrySessionReplay to avoid running any SessionReplay logic if not needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Disable SessionSentryReplayIntegration if the environment is unsafe [#6573]
+- Fix crash when last replay info is missing some keys [#6577]
 
 ## 8.57.0
 
--- a/Sources/Sentry/SentrySessionReplayIntegration.m
+++ b/Sources/Sentry/SentrySessionReplayIntegration.m
@@ -216,10 +216,25 @@ static SentryTouchTracker *_touchTracker;
         return;
     }
 
-    SentryId *replayId = jsonObject[@"replayId"]
-        ? [[SentryId alloc] initWithUUIDString:jsonObject[@"replayId"]]
-        : [[SentryId alloc] init];
-    NSURL *lastReplayURL = [dir URLByAppendingPathComponent:jsonObject[@"path"]];
+    SentryId *replayId;
+    if (jsonObject[@"replayId"] && [jsonObject[@"replayId"] isKindOfClass:NSString.class]) {
+        replayId = [[SentryId alloc]
+            initWithUUIDString:SENTRY_UNWRAP_NULLABLE(NSString, jsonObject[@"replayId"])];
+    } else {
+        replayId = [[SentryId alloc] init];
+    }
+    if (!jsonObject[@"path"] || ![jsonObject[@"path"] isKindOfClass:NSString.class]) {
+        SENTRY_LOG_ERROR(@"[Session Replay] Failed to read path from last replay");
+        return;
+    }
+    NSURL *_Nullable nullableUrl =
+        [dir URLByAppendingPathComponent:SENTRY_UNWRAP_NULLABLE(NSString, jsonObject[@"path"])];
+    if (!nullableUrl) {
+        SENTRY_LOG_ERROR(
+            @"[Session Replay] Failed to create URL with path: %@", jsonObject[@"path"]);
+        return;
+    }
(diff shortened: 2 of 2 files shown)


Sources/Sentry/SentryReplayApi.m
```diff
@@ -56,8 +56,12 @@ - (void)start SENTRY_DISABLE_THREAD_SANITIZER("double-checked lock produce false
             replayIntegration = (SentrySessionReplayIntegration *)[SentrySDKInternal.currentHub
                 getInstalledIntegration:SentrySessionReplayIntegration.class];
             if (replayIntegration == nil) {
-                SENTRY_LOG_DEBUG(@"[Session Replay] Initializing replay integration");
                 SentryOptions *currentOptions = SentrySDKInternal.currentHub.client.options;
+                if (![SentrySessionReplayIntegration shouldEnableForOptions:currentOptions]) {
+                    return;
```
philprime, line 65: `l`: We should a SENTRY_LOG_ERROR warning as this is called from an SDK user

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -69,29 +64,32 @@ import UIKit
...
             guard experimentalOptions.enableSessionReplayInUnreliableEnvironment else {
                 SentrySDKLog.fatal("[Session Replay] Detected environment potentially causing PII leaks, disabling Session Replay. To override this mechanism, set `options.experimental.enableSessionReplayInUnreliableEnvironment` to `true`")
-                return
+                return false
             }
             SentrySDKLog.warning("[Session Replay] Detected environment potentially causing PII leaks, but `options.experimental.enableSessionReplayInUnreliableEnvironment` is set to `true`, ignoring and enabling Session Replay.")
         }
         
+        return true
+    }
```
philprime, line 85: `l`: Could probably be refactored as an early-exit method like this:

```
    static public func shouldEnableSessionReplay(environmentChecker: SentrySessionReplayEnvironmentCheckerProvider, experimentalOptions: SentryExperimentalOptions) -> Bool {
        // Detect if we are running on iOS 26.0 with Liquid Glass and disable session replay.
        // This needs to be done until masking for session replay is properly supported, as it can lead
        // to PII leaks otherwise.
        if environmentChecker.isReliable() {
            return true
        }
        guard experimentalOptions.enableSessionReplayInUnreliableEnvironment else {
            SentrySDKLog.fatal("[Session Replay] Detected environment potentially causing PII leaks, disabling Session Replay. To override this mechanism, set `options.experimental.enableSessionReplayInUnreliableEnvironment` to `true`")
            return false
        }
        SentrySDKLog.warning("[Session Replay] Detected environment potentially causing PII leaks, but `options.experimental.enableSessionReplayInUnreliableEnvironment` is set to `true`, ignoring and enabling Session Replay.")

        return true
    }
```

Tests/SentryTests/Integrations/SessionReplay/SentryReplayApiTests.swift
```diff
@@ -0,0 +1,106 @@
+import Foundation
+@_spi(Private) @testable import Sentry
+@_spi(Private) import SentryTestUtils
+import XCTest
+
+#if os(iOS) || os(tvOS)
+
+class SentryReplayApiTests: XCTestCase {
```
philprime, line 8: Thanks for adding missing tests 🙏

Tests/SentryTests/Integrations/SessionReplay/SentryReplayApiTests.swift
```diff
@@ -0,0 +1,106 @@
...
+        SentryDependencyContainer.sharedInstance().sessionReplayEnvironmentChecker = TestSessionReplayEnvironmentChecker(mockedIsReliableReturnValue: false)
+
+        // Act
+        sut.start()
+
+        // Assert
+        XCTAssertTrue(mockHub.installedIntegrations().isEmpty)
+    }
+
+    func testStart_whenReplayIntegrationNilWithUnreliableEnvironmentAndExperimental_shouldCreateAndInstallIntegration() throws {
```
philprime:
```suggestion
    func testStart_whenReplayIntegrationNilWithUnreliableEnvironmentAndOverrideOptionEnabled_shouldCreateAndInstallIntegration() throws {
```

philprime (approved):
LGTM, add minor comments

PR #6584, opened 2025-10-29
fix: Fix UITouch background thread access in SentryTouchTracker

### LINKED ISSUE #6231: Remove access to UITouch instances from background thread
In certain circumstances the SentryTouchTracker accesses touch events from a background thread. That should never happen with UIKit objects. If touch events are required on background threads, they should be converted to thread-safe model objects, e.g. as data structs.

Analyze the stack trace and fix the issue.

philprime:
## :scroll: Description

This PR fixes a crash caused by accessing UITouch instances from a background thread in . The fix extracts all necessary data from UITouch objects on the main thread before dispatching to the background queue, then processes only thread-safe data structures on the background thread.

### Changes:
- Added `ExtractedTouchData` struct to hold thread-safe touch information
- Changed `trackedTouches` dictionary to use `ObjectIdentifier` as key instead of `UITouch`
- Extract `touch.location`, `touch.phase`, and `ObjectIdentifier` on calling thread (main thread)
- Process only thread-safe data structures (structs, enums, value types) on background queue

Fixes #6231

The issue was that `trackTouchFrom(event:)` dispatched to a background queue and then accessed UITouch properties (`touch.location(in: nil)` and `touch.phase`) from that background thread. This violates UIKit's thread-safety requirements and caused `EXC_BAD_ACCESS` crashes.

The stack trace from the issue showed:
```
EXC_BAD_ACCESS: Attempted to dereference garbage pointer 0x400000000000bad0.
...
2  REDACTED +0xcbebc0                closure #1 in SentryTouchTracker.trackTouchFrom(event:)
...
5  libdispatch.dylib +0x1ad8       __dispatch_call_block_and_release
```

- All 12 existing tests in `SentryTouchTrackerTests` pass successfully
- Verified no UIKit objects cross the thread boundary by code review
- All UITouch accesses now happen on the calling thread before dispatching

## :pencil: Checklist

- [x] I added tests to verify the changes. (Existing tests verify behavior remains unchanged)
- [x] I updated the docs if needed. (No doc changes needed - internal fix)
- [x] I updated the wizard if needed. (No wizard changes needed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Fix crash from accessing UITouch instances from background thread and ObjectIdentifier collision when UITouch memory is reused in SentryTouchTracker (#6584)
+- Fix crash from accessing UITouch instances from background thread in SentryTouchTracker (#6584)
 - Disable SessionSentryReplayIntegration if the environment is unsafe [#6573]
 - Fix crash when last replay info is missing some keys [#6577]
 


itaybre (approved):
lgtm

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -57,14 +66,39 @@ import UIKit
...
+            for extractedTouch in extractedTouches {
+                let info: TouchInfo
+                
+                if extractedTouch.phase == .began {
+                    // Always create new TouchInfo for .began - handles ObjectIdentifier collisions
+                    info = TouchInfo(id: touchId++, identifier: extractedTouch.identifier)
+                    allTouchInfos.append(info)
+                } else {
+                    // Find existing TouchInfo with matching identifier (search backwards for most recent)
+                    if let existingInfo = allTouchInfos.last(where: { $0.identifier == extractedTouch.identifier && $0.endEvent == nil }) {
```
noahsmartin: This is like a N^2 loop right? Because of the outer `for extractedTouch in extractedTouches`. Can we not use a dictionary lookup like before?

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
```diff
@@ -57,14 +66,39 @@ import UIKit
         guard let touches = event.allTouches else { return }
         let timestamp = event.timestamp
         
+        // Extract touch data on the main thread before dispatching to background queue
+        // to avoid accessing UIKit objects from a background thread
+        let extractedTouches: [ExtractedTouchData] = touches.compactMap { touch in
+            guard touch.phase == .began || touch.phase == .ended || touch.phase == .moved || touch.phase == .cancelled else { return nil }
+            let position = touch.location(in: nil).applying(scale)
+            // There is no way to uniquely identify a touch, so we use the ObjectIdentifier of the UITouch instance.
+            // This is not a perfect solution, but it is the best we can do without holding references to UIKit objects.
```
noahsmartin, line 80: `l`: IMO this is going to be very risky, could lead to a lot of hard to diagnose bugs. Maybe I'm missing something obvious here, but can't we just remove the dispatch queue and handle touch events on the main thread? It doesn't look like anything intensive is going on in the background thread and it's normal to to gesture handling/touch tracking on the main thread

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philprime: We can but the whole thing got so messy that I went for this way. I don't think this class is optimized anyways, I just need to get the crashes fixed ASAP.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philprime: Unfortunately I don't know why it was dispatched to the background, so I need to ask the same question as you.

philprime: @philipphofmann can you help us out why this was moved onto a dispatch queue back in https://github.com/getsentry/sentry-cocoa/pull/4548? I don't understand the context from the PR discussions

philippsuess: The PR description states

> Touch tracker adds events on the main thread, but reads it from a background thread, this could lead to a race condition.

Then we had a discussion around using a lock potentially slowing it down here https://github.com/getsentry/sentry-cocoa/pull/4548#discussion_r1846756921.

I'm unsure what else you need to understand. What we missed in that PR back then was accessing UIEvent on a BG thread. I have no clue how I could miss that.

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
itaybre: Won't this be limited in amount of touches?
I think Apple limits the number of touches to 4 or 5

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philprime: But that means we can't do it on the main thread because of using a lock but we also can't use it on a background thread because UIKit elements are not thread-safe?

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philprime: Yes I think so, but for the sake of a best-practice SDK as a goal, I'll revert it to dictionary based approach. Our main issue is the multi-threading anyways

Sources/Swift/Integrations/SessionReplay/SentryTouchTracker.swift
philippsuess: Exactly. That's why the mapping makes sense.

philippsuess: Ideally, we would have a data structure that you can write on the main thread without any locks, but you can still read from a BG thread without a guarantee that all data is up to date. So basically, you would accept a slight race condition over blocking the calling thread.

philprime: Thanks for confirming, in that case we can't just move the processing of events to the main thread and need to stick with the mapping

itaybre:
CI is being very slow, I am backporting some fixes on https://github.com/getsentry/sentry-cocoa/pull/6604 to help this PR and the later release

PR #6605, opened 2025-10-30
refactor: Change SentryDsn.getHash to internal method

### LINKED ISSUE #6059: Make SentryDsn.getHash internal

philprime:
## :scroll: Description

Removes `SentryDsn.getHash` from the public header.
Adds private headers where needed.

Closes #6059

Used unit tests to confirm that it compiles and passes.

## :pencil: Checklist


Samples/iOS-Swift/iOS-Swift/Tools/iOS-Swift-Bridging-Header.h
```diff
@@ -2,4 +2,5 @@
 #import "SentryBenchmarking.h"
 #import "SentryExposure.h"
 #import <Sentry/PrivateSentrySDKOnly.h>
+#import <Sentry/SentryDsn+Private.h>
```
philippsuess, line 5: I think the Xcode proj is missing a reference.

philprime: Should've been a draft PR, but thanks for pointing this out

PR #6607, opened 2025-10-30
ref: Rename SentryMechanismMeta to SentryMechanismMetaInformation

### LINKED ISSUE #2888: Rename `SentryMechanismMeta` to something else different than `Meta`.
### Description

Running into this issue right now: https://github.com/getsentry/sentry-kotlin-multiplatform/blob/d5ecbdf84b0a7881ed9664715436f34c8d8072c8/sentry-kotlin-multiplatform/build.gradle.kts#L178-L185

philprime:
## Description

This PR renames `SentryMechanismMeta` to `SentryMechanismContext` to resolve naming conflicts with Kotlin Multiplatform's metaprogramming features.

Closes #2888

### Renamed Classes
- `SentryMechanismMeta` → `SentryMechanismContext`
- Swift name: `MechanismMeta` → `MechanismContext`

### Files Changed
- **Created:**
  - `Sources/Sentry/Public/SentryMechanismContext.h`
  - `Sources/Sentry/SentryMechanismContext.m`
  - `Sources/Swift/Protocol/Codable/SentryMechanismContextCodable.swift`
  - `Tests/SentryTests/Protocol/SentryMechanismContextTests.swift`

- **Deleted:**
  - `Sources/Sentry/Public/SentryMechanismMeta.h`
  - `Sources/Sentry/SentryMechanismMeta.m`
  - `Sources/Swift/Protocol/Codable/SentryMechanismMetaCodable.swift`
  - `Tests/SentryTests/Protocol/SentryMechanismMetaTests.swift`

- **Updated:**
  - All import statements across the codebase
  - Public headers (`Sentry.h`, `SentryWithoutUIKit.h`, `SentryMechanism.h`)
  - Xcode project file
  - SDK API file
  - Test files and utilities

## Testing

- ✅ iOS build successful
- ✅ macOS build successful  
- ✅ All `SentryMechanismContextTests` pass on iOS
- ✅ All `SentryMechanismContextTests` pass on macOS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@
   - Add layer class filtering for views used in multiple contexts (e.g., SwiftUI._UIGraphicsView)
   - Improve transform calculations for views with custom anchor points
   - Fix axis-aligned transform detection for optimized opaque view clipping
-- Rename `SentryMechanismMetaInformation` to `SentryMechanismContext` to resolve Kotlin Multi-Platform build errors (#6607)
+- Rename `SentryMechanismMeta` to `SentryMechanismContext` to resolve Kotlin Multi-Platform build errors (#6607)
 - Fix conversion of frame rate to time interval for session replay (#6623)
 
 ### Improvements


Sources/Sentry/Public/SentryMechanism.h
```diff
@@ -58,7 +58,7 @@ SENTRY_NO_INIT
  * Information from the operating system or runtime on the exception
  * mechanism.
  */
-@property (nullable, nonatomic, strong) SentryMechanismMeta *meta;
+@property (nullable, nonatomic, strong) SentryMechanismContext *meta;
```
philippsuess, line 61: `m`: This gets serialized as `meta` key. I think we should keep meta in the name. What about `SentryMechanismMetaInformation`?

philippsuess:
FYI @philprime, I pinged Gino if this is still required https://github.com/getsentry/sentry-cocoa/issues/2888#issuecomment-3467774615

Sources/Sentry/Public/SentryMechanism.h
philprime: I have no strong preference. I can change it `SentryMechanismMetaInformation`

buenaflor:
@philipphofmann I think we general have issues if Meta is in the name of the class but don't let this block it you, it's not critical at all (https://youtrack.jetbrains.com/issue/KT-41709)

Change looks fine for me

philippsuess:
> https://youtrack.jetbrains.com/issue/KT-41709

@buenaflor the class name would still contain `Meta` and we're only renaming it because of KMP. So if it doesn't improve the situation we don't do the renaming. If you're saying it's not important, we don't have to do the refactoring at all.

buenaflor:
In that case I agree this refactoring can be avoided unless you think something like `SentryMechanismContext` is alright then I'd opt for that but if you feel the change to `SentryMechanismContext` is too much work or has friction with other stuff we can just drop this

philprime:
I have already refactored `SentryMechanismContext` to `SentryMechanismMetaInformation`, so if we can not use `Meta` in the type name, let's go back to `SentryMechanismContext` or similar.

@philipphofmann I do not have a strong opinion on the naming, so even though the field is called `meta` I would go for `SentryMechanismContext` or similar. We could also consider renaming `meta` to `context` if that helps.

philippsuess:
> @philipphofmann I do not have a strong opinion on the naming, so even though the field is called `meta` I would go for `SentryMechanismContext` or similar. We could also consider renaming `meta` to `context` if that helps.

OK, then let's do `SentryMechanismContext` and add a comment on why we have to use that name. Not many people interact with that API, so it's not that important.

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@
   - Add layer class filtering for views used in multiple contexts (e.g., SwiftUI._UIGraphicsView)
   - Improve transform calculations for views with custom anchor points
   - Fix axis-aligned transform detection for optimized opaque view clipping
+- Rename `SentryMechanismMetaInformation` to `SentryMechanismContext` to resolve Kotlin Multi-Platform build errors (#6607)
```
```suggestion
- Rename `SentryMechanismMeta` to `SentryMechanismContext` to resolve Kotlin Multi-Platform build errors (#6607)
```

PR #6608, opened 2025-10-30
refactor: Change SentryFrame.function default from <redacted> to nil

### LINKED ISSUE #4686: Set SentryFrame.function default to nil instead of <redacted>
### Description

https://github.com/getsentry/sentry-cocoa/blob/f69e4da33bc05f4e20d3260e148974cf0efa3192/Sources/Sentry/SentryFrame.m#L11

The default of `SentryFrame.function` is `<redacted>` which makes no sense. We don't do this in other SDKs, such as [Java](https://github.com/getsentry/sentry-java/blob/eb7afff8ea69d85a1644b48d50a30496efa280c7/sentry/src/main/java/io/sentry/protocol/SentryStackFrame.java) for example. The Cocoa SDK doesn't symbolcate function names locally mostly because it's slow, and it usually doesn't work for app store builds. When the SDK can't detect the function name, instead it should keep it set to `nil` and let the backend set a potential `<redacted>`. This is a breaking change, though.

philprime:
## :scroll: Description

Changed `SentryFrame.function` default value from `@"<redacted>"` to `nil`. This aligns the Cocoa SDK with other Sentry SDKs (like Java) and lets the backend handle redaction when function names can't be detected.

Fixes #4686

The Cocoa SDK doesn't symbolicate function names locally (it's slow and doesn't work for App Store builds). When the function name is unknown, the SDK should keep it as `nil` and let the backend apply `<redacted>` if needed, rather than doing it client-side.

- Built SDK for iOS and macOS - both builds succeeded
- Ran full test suite on iOS Simulator (iPhone 16 Pro, iOS 18.4) - all tests passed
- Ran full test suite on macOS - all tests passed
- Updated all affected tests to expect `nil` instead of `<redacted>`:
  - `SentryFrameTests`
  - `SentryInterfacesTests`
  - `SentryCrashReportConverterTests`
  - `SentryCrashStackEntryMapperTests`
  - `SentryDefaultThreadInspectorTests`
  - `SentryMetricKitIntegrationTests`

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentryFrame.m
+++ b/Sources/Sentry/SentryFrame.m
@@ -8,7 +8,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)init
 {
-     self = [super init];
+    self = [super init];
     return self;
 }
 


Sources/Sentry/SentryFrame.m
```diff
@@ -8,10 +8,7 @@ @implementation SentryFrame
 
 - (instancetype)init
 {
-    if (self = [super init]) {
-        self.function = @"<redacted>";
-    }
-    return self;
+    return [super init];
```
philippsuess: `m`: Assigning self is good practice. There are some weird potential subclassing issues that could bite you if you don't do that. 
```suggestion
     self = [super init];
    return self;
```

philipphofmann (approved):
Thanks, LGTM.

PR #6616, opened 2025-10-30
chore: Remove some deprecated properties for v9

### LINKED ISSUE #6617: chore: Remove some deprecated properties for v9
> [!NOTE]
> The pull request "[chore: Remove some deprecated properties for v9](https://github.com/getsentry/sentry-cocoa/pull/6616)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Removing some of these that we don't need in V9

#skip-changelog

noahsmartin:
Removing some of these that we don't need in V9

#skip-changelog

Closes #6617

### DIFF
--- a/Tests/SentryTests/SentryHubTests.swift
+++ b/Tests/SentryTests/SentryHubTests.swift
@@ -1565,7 +1565,6 @@ class SentryHubTests: XCTestCase {
         init() {
             super.init(
                 replayOptions: SentryReplayOptions(sessionSampleRate: 0, onErrorSampleRate: 0),
-                experimentalOptions: SentryExperimentalOptions(),
                 replayFolderPath: FileManager.default.temporaryDirectory,
                 screenshotProvider: MockScreenshotProvider(),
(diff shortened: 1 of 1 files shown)


Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -162,22 +162,6 @@ public final class SentryReplayOptions: NSObject, SentryRedactOptions {
...
     public var unmaskedViewClasses: [AnyClass]
 
-    /**
-     * Alias for ``enableViewRendererV2``.
-     *
-     * This flag is deprecated and will be removed in a future version.
-     * Please use ``enableViewRendererV2`` instead.
-     */
-    @available(*, deprecated, renamed: "enableViewRendererV2")
-    public var enableExperimentalViewRenderer: Bool {
```
philprime, line 172: `m`: We might also want to remove it from the `init(dict:)` initializer

Issue #6612, opened 2025-10-30
Logging a breadcrumb over 1mb in size prevents future messages and errors from being uploaded after capture


dfed:
### Platform

iOS

### Environment

Production

### Installed

Swift Package Manager

### Version

8.56.0 and later (possibly earlier as well)

### Xcode Version

26.0.0

### Did it work on previous versions?

Unknown

### Steps to Reproduce

1. Log a gigantic breadcrumb (anything over 1mb)
2. Run `SentrySDK.capture(error: error)` or `SentrySDK.capture(message: message)`

### Expected Result

Breadcrumb fails to log. Error and message capture + upload

### Actual Result

Breadcrumb fails to log (400 error on upload), and then error and message captures never end up uploading. The app will fail to upload errors and messages until deleting + reinstalling the app (or possibly waiting for some long expiry period)

### Are you willing to submit a PR?

If you point me at how you'd like to approach this, I could help, though I doubt I'm the best person for this particular issue.

dfed:
Note that there also isn't a great way for me to determine a-priori if a breadcrumb will exceed the threshold and drop it in `beforeBreadcrumb`. I'm currently truncating breadcrumbs by unicode character length in a directional attempt to prevent hitting this, but ideally a breadcrumb that 400s on upload would be dropped on the floor (or maybe be replaced by a breadcrumb along the lines of "Attempted to send breadcrumb that was too large") and then future breadcrumbs and errors/messages would not get stuck in a queue that will never succeed.

sberrevoets:
A (default) setting to automatically truncate breadcrumbs to the allowed size instead of a 400 would be nice as well

philippsuess:
> A (default) setting to automatically truncate breadcrumbs to the allowed size instead of a 400 would be nice as well

@sberrevoets, we try to let the server truncate as much as possible and remove truncation on the client side as much as we can with https://github.com/getsentry/sentry-cocoa/issues/5913, but for scenarios like this we might have to revisit our approach.

philippsuess:
OK, SDKs should actually drop such envelopes and record client outcomes for it. We're going to update our [develop docs](https://develop.sentry.dev/sdk/expected-features/#dealing-with-network-failures) to more clearly specify this behavior, and after that, we can adapt the logic in the Cocoa SDK. We found out that we have the same problem on the sentry-java SDK. I'm surprised that this didn't surface sooner.

PR #6618, opened 2025-10-30
fix: Transport correctly handling 4xx and 5xx

dfed:
## :scroll: Description

If an upload receives a 4xx error, Sentry should assume that a retry will not succeed and drop it on the floor. Currently Sentry will retry the upload (seemingly forever), which means that a 4xx error on upload will prevent future uploads from succeeding.

This PR represents the shortest-path resolution to #6612. I'd prefer a solution where a 4xx error didn't cause a silent failure, or a solution where breadcrumb were automatically truncated. But this change is simple and unsticks existing clients, which seems like a good thing.

I didn't.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-## Fixes
+### Fixes
 
 - The transport now correctly discard envelopes on 4xx and 5xx responses and records client reports `send_error` (#6618) This also fixes edge cases in which the SDK kept retrying sending a faulty envelope until the offline cache overflowed.
 - Change default attributes of Logs to only include user attributes when `options.sendDefaultPii = true` (#7055)


CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
   - Add layer class filtering for views used in multiple contexts (e.g., SwiftUI._UIGraphicsView)
   - Improve transform calculations for views with custom anchor points
   - Fix axis-aligned transform detection for optimized opaque view clipping
+- Fix issue where a too-large breadcrumb would prevent future errors and messages from being uploaded (#6617)
```
dfed:
```suggestion
- Fix issue where a too-large breadcrumb would prevent future errors and messages from being uploaded (#6618)
```

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -414,6 +414,13 @@ - (void)sendEnvelope:(SentryEnvelope *)envelope
                 return;
             }
 
+            if (response.statusCode >= 400 && response.statusCode < 500) {
+                SENTRY_LOG_DEBUG(@"Received 4xx response code: %li", (long)response.statusCode);
+                // 4xx means the payload is bad and will not succeed on retry. Drop it on the floor and enable sending the next one.
+                [weakSelf deleteEnvelopeAndSendNext:envelopePath];
+                return;
+            }
```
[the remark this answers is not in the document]
dfed: fair. pushed a fix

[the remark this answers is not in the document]
dfed: fair. adding `[weakSelf recordLostEventFor:envelope.items];`

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -414,6 +414,14 @@ - (void)sendEnvelope:(SentryEnvelope *)envelope
                 return;
             }
 
+            if (error && response.statusCode >= 400 && response.statusCode < 500 && response.statusCode != 429) {
```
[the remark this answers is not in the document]
dfed: fair.
```suggestion
            if (response.statusCode >= 400 && response.statusCode < 500 && response.statusCode != 429) {
```

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -414,6 +414,14 @@ - (void)sendEnvelope:(SentryEnvelope *)envelope
...
+                SENTRY_LOG_DEBUG(@"Received 4xx response code: %li", (long)response.statusCode);
+                // 4xx means the payload is bad and will not succeed on retry. Drop it on the floor and enable sending the next one.
+                [weakSelf recordLostEventFor:envelope.items];
+                [weakSelf deleteEnvelopeAndSendNext:envelopePath];
+                return;
+            }
```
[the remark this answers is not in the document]
dfed: lol good point. Bugbot incorrectly stated we weren't logging these failures earlier.

Sources/Sentry/SentryHttpTransport.m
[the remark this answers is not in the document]
dfed: turns out this was wrong. we were logging this above!

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -414,6 +414,13 @@ - (void)sendEnvelope:(SentryEnvelope *)envelope
                 return;
             }
 
+            if (response.statusCode >= 400 && response.statusCode < 500 && response.statusCode != 429) {
+                SENTRY_LOG_DEBUG(@"Received 4xx response code: %li", (long)response.statusCode);
```
dfed: happy to make this specific to `400` status codes if we don't want to go broader, but generally speaking `4xx` is "client screwed up and shouldn't try again"

philippsuess: Yes, I agree that 400 usually means the client shouldn't retry again, but I still have to double-check if they might not have a good reason to keep the envelope. 

The somewhat good news is that the SDK will delete the faulty envelope once the cache is full. The cache has a default max size of 100 envelopes. So when you log more than 100 errors, the SDK will send out all the other errors. Of course this isn't ideal, but at least the other envelopes end up in Sentry eventually.

dfed: Sweet. Happy to update this PR. Let me know what you find!

philippsuess: Hey @dfed, sorry that it took so long. So yes, we must drop envelopes when getting 400 and also 500 status codes and we must also record a client report with the reason `send_error`. When we get a 429, we must also drop the envelope, but we must not record a client report, as the backend already does this, and otherwise we would be double-counting.

We record client reports here https://github.com/getsentry/sentry-cocoa/blob/f0d57aea65cc61752dffa3c68d13acae1baaedd6/Sources/Sentry/SentryHttpTransport.m#L429-L442

We don't have the `send_error` category yet, so we need to add it here
https://github.com/getsentry/sentry-cocoa/blob/f0d57aea65cc61752dffa3c68d13acae1baaedd6/Sources/Sentry/include/SentryDiscardReason.h#L3-L21

More info on client reports is here: https://develop.sentry.dev/sdk/telemetry/client-reports/ and this PR updates the develop docs on this topic https://github.com/getsentry/sentry-docs/pull/15571

@dfed, I leave it up to you. Feel free to implement the changes and also add tests for it, or we can take it from here and push the PR over the finish line.

dfed: Happy to let y'all take it over the line. Feel free to close this PR or push to it: whatever makes your lives easier.

philprime:
@philipphofmann are you going to pick this up?

philippsuess:
@philprime, yes, I plan to next week. I mean, you or anybody else can also pick it up if you have bandwidth.

Sources/Sentry/include/SentryDiscardReason.h
```diff
@@ -12,10 +12,6 @@ typedef NS_ENUM(NSUInteger, SentryDiscardReason) {
...
     kSentryDiscardReasonRateLimitBackoff = 6,
-    kSentryDiscardReasonInsufficientData = 7
+    kSentryDiscardReasonInsufficientData = 7,
+    kSentryDiscardReasonSendError = 8
 };
-
-static DEPRECATED_MSG_ATTRIBUTE(
-    "Use nameForSentryDiscardReason() instead.") NSString *_Nonnull const SentryDiscardReasonNames[]
-    = { @"before_send", @"event_processor", @"sample_rate", @"network_error", @"queue_overflow",
-          @"cache_overflow", @"ratelimit_backoff", @"insufficient_data" };
```
philippsuess, line 21: This isn't used anywhere. Instead of adding send_error here, I removed it.

philippsuess:
I manually tested capturing a message with the iOS-Swift sample app
```swift
let message = Data(repeating: 0, count: 1024*1024*2) // 2 MB message
let string = message.base64EncodedString()
let eventId = SentrySDK.capture(message: string)
```

And the transport successfully deleted the envelope when getting a 400 from Relay.

Doing the same thing on the main branch, led to filling up the chache until the faulty envelope with the too large message finally gets deleted due to cache overflow.

Tests/SentryTests/Integrations/ANR/SentryANRTrackerTests.swift
```diff
@@ -7,7 +7,7 @@ final class SentryANRTrackerTests: XCTestCase {
         let helper = MockSentryANRTrackerHelper()
         let tracker = SentryANRTracker(helper: helper)
         var delegate: MockANRTrackerDelegate? = MockANRTrackerDelegate()
-        weak var delegateWeak: MockANRTrackerDelegate? = delegate
+        weak let delegateWeak: MockANRTrackerDelegate? = delegate
```
philippsuess: I have no clue why, but the tests failed to compile for me locally, so I had to add this.

philippsuess:
Back to draft because some tests are failing.

Tests/SentryTests/SentrySDKTests.swift
```diff
@@ -371,34 +371,23 @@ class SentrySDKTests: XCTestCase {
...
         SentrySDK.capture(message: "Test message")
-        // Ensures that the capture envelope is written to disk before we read it.
-        SentrySDK.flush(timeout: 0.1)
 
         // Assert
-        let eventEnvelopeItems = try fileManager.getAllEnvelopes().map { fileContent in
-            return try XCTUnwrap(SentrySerializationSwift.envelope(with: fileContent.contents))
-        }.flatMap { envelope in
-            return envelope.items.filter { $0.header.type == SentryEnvelopeItemTypes.event }
-        }
```
philippsuess, line 396: This only worked because the transport didn't correctly discard the envelope on a 400 due to the test DSN above. This PR changes this behaviour. We can also check the correctness of the attached debug meta via the beforeSend hook intead.

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -437,8 +443,8 @@ - (void)recordLostEventFor:(NSArray<SentryEnvelopeItem *> *)items
             continue;
         }
         SentryDataCategory category = sentryDataCategoryForEnvelopItemType(itemType);
-        [self recordLostEvent:category reason:kSentryDiscardReasonNetworkError];
-        [self recordLostSpans:item reason:kSentryDiscardReasonNetworkError];
+        [self recordLostEvent:category reason:kSentryDiscardReasonSendError];
```
itaybre, line 461: `l`: All error while sending will be marked as `send_error`?

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -408,14 +403,25 @@ - (void)sendEnvelope:(SentryEnvelope *)envelope
...
+            SENTRY_LOG_DEBUG(@"Received response status code: %li", (long)response.statusCode);
+
+            BOOL is2xx = (response.statusCode >= 200 && response.statusCode < 300);
+            BOOL is4xxOr5xx = (response.statusCode >= 400 && response.statusCode < 600);
+
+            // Relay already records a client report for a 429, so we must not record it again
+            // to avoid double-counting.
+            BOOL isNotRateLimitStatusCode = response.statusCode != 429;
+
+            if ((error != nil || is4xxOr5xx) && isNotRateLimitStatusCode) {
```
noahsmartin: Can't `error != nil` never be true here? Above you have a check for `response == nil` that returns early. If `error != nil` evaluates to true then `response == nil` will also be true. More reason for this code to be in Swift!

Sources/Sentry/SentryHttpTransport.m
philippsuess: Yes exactly as pointed out here https://develop.sentry.dev/sdk/expected-features/#dealing-with-network-failures

Sources/Sentry/SentryHttpTransport.m
philippsuess: Yes, I agree that this should be in Swift at some point. I think there are edge cases in which the error can be nonnil and we get a response. But in that case, we should only care about the response and actually ignore the error. I removed the check here and added a comment. Thanks for pointing it out 👍

Sources/Sentry/SentryHttpTransport.m
```diff
@@ -395,27 +395,43 @@ - (void)sendEnvelope:(SentryEnvelope *)envelope
                 return;
             }
 
-            if (error && response.statusCode != 429) {
```
noahsmartin, line 398: Just pointing out, the `error` param is now not used in this completion handler. Probably OK but we might want to log it below in the `response == nil` check

dfed:
Thank you all for getting this over the line! Excited for this 🚀

PR #6623, opened 2025-10-31
fix(session-replay): Fix conversion of frame rate to time interval

### LINKED ISSUE #6624: fix(session-replay): Fix conversion of frame rate to time interval
> [!NOTE]
> The pull request "[fix(session-replay): Fix conversion of frame rate to time interval](https://github.com/getsentry/sentry-cocoa/pull/6623)" was created by @philprime but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

When using `@_spi(Private) import Sentry` it is possible to set a higher frame rate using `options.sessionReplay.frameRate = 10`.

Due to a `UInt / UInt` this will result in a time interval since last frame of `0 ms` instead of e.g. `100ms`. 

This PR fixes the division.

No changelog needed because `frameRate` is internal.

#skip-changelog

philprime:
When using `@_spi(Private) import Sentry` it is possible to set a higher frame rate using `options.sessionReplay.frameRate = 10`.

Due to a `UInt / UInt` this will result in a time interval since last frame of `0 ms` instead of e.g. `100ms`. 

This PR fixes the division.

No changelog needed because `frameRate` is internal.

#skip-changelog

Closes #6624

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
   - Add layer class filtering for views used in multiple contexts (e.g., SwiftUI._UIGraphicsView)
   - Improve transform calculations for views with custom anchor points
   - Fix axis-aligned transform detection for optimized opaque view clipping
+- Fix conversion of frame rate to time interval for session replay (#6623)
 
 ### Improvements
 
--- a/Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayTests.swift
+++ b/Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayTests.swift
@@ -577,6 +577,120 @@ class SentrySessionReplayTests: XCTestCase {
         XCTAssertTrue(SentrySessionReplay.shouldEnableSessionReplay(environmentChecker: environmentChecker, experimentalOptions: experimentalOptions))
     }
 
+    // MARK: - Frame Rate Tests
+
+    func testFrameRate_1FPS_takesScreenshotsAtCorrectInterval() {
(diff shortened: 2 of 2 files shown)


Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -233,7 +233,7 @@ import UIKit
             return
         }
 
-        if now.timeIntervalSince(lastScreenShot) >= Double(1 / replayOptions.frameRate) {
+        if now.timeIntervalSince(lastScreenShot) >= 1.0 / Double(replayOptions.frameRate) {
```
philippsuess, line 236: `m`: The impact of this bug was critical, if I'm not mistaken. IMO, we should add a test to prevent the rounding error from happening again.

philipphofmann (approved):
> No changelog needed because frameRate is internal.

If I'm not mistaken, this fixes a critical bug. I think this PR is worth a changelog entry.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
philprime: Even if it never applied to any user? I can add one in the "internal" section

philippsuess: Sorry, I thought this fixed a critical bug. If not, then don't add a changelog entry.

PR #6627, opened 2025-10-31
ref: SentryClient and SentryHub in Swift

### LINKED ISSUE #6630: ref: SentryClient and SentryHub in Swift
> [!NOTE]
> The pull request "[ref: SentryClient and SentryHub in Swift](https://github.com/getsentry/sentry-cocoa/pull/6627)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

This PR looks big in the line count, but only because it renamed some things. The real changes are just SentryId.h/.m, SentryClient.swift, and SentryHub.swift. All of these are pretty small.

Refactors the public interface of SentryClient and SentryHub to be in Swift. The implementation doesn't change at all I keep the old old classes around and just renamed them to SentryClientInternal and SentryHubInternal. This unblocks using SPM without any risk of a big Swift migration, and lets us migrate them when we want without affecting the public API.

Also converts SentryId to ObjC which is required for SPM, and lets us delete SentryIdWrapper. We can convert it to Swift again once all of the classes depending on it in the public API are also in Swift. Basically, this should be the last part of the public API to be in Swift.

#skip-changelog

noahsmartin:
This PR looks big in the line count, but only because it renamed some things. The real changes are just SentryId.h/.m, SentryClient.swift, and SentryHub.swift. All of these are pretty small.

Refactors the public interface of SentryClient and SentryHub to be in Swift. The implementation doesn't change at all I keep the old old classes around and just renamed them to SentryClientInternal and SentryHubInternal. This unblocks using SPM without any risk of a big Swift migration, and lets us migrate them when we want without affecting the public API.

Also converts SentryId to ObjC which is required for SPM, and lets us delete SentryIdWrapper. We can convert it to Swift again once all of the classes depending on it in the public API are also in Swift. Basically, this should be the last part of the public API to be in Swift.

#skip-changelog

Closes #6630


philprime:
@sentry review
@cursor review

PR #6628, opened 2025-11-02
ref: SentryOptions in Swift

### LINKED ISSUE #6690: ref: SentryOptions in Swift
> [!NOTE]
> The pull request "[ref: SentryOptions in Swift](https://github.com/getsentry/sentry-cocoa/pull/6628)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

> This change may look big, but I've done everything that can be done to make it as small as possible. It only looks big because the generated API json file changed, the actual lines of code added are ~800. So many files were touched because I had to remove a bunch of imports of the objc class

This converts SentryOptions.h/.m to SentryOptions.swift and in doing so removes the last blocker for using SPM. It also cleans up a lot of gross bridging we had to do, like with SentryExperimentalOptions and the beforeSendLog callback. This is now replaced with one piece of cross bridging, which I'll get to later...

Almost all the relevant implementation here is in `Options.swift`. This matches the old ObjC API almost exactly, except that the class is now `final`. The API json file changed a lot structurally just because it represents Swift and ObjC differently, not because the actual API changed.

The tricky part of this is that we can't call an objc header that uses a swift type from Swift code. (This works in xcodebuild but not cocoapods/SPM) To work around this the value that gets passed to`SentrySDKInternal` needs to be type-erased. This is the same as what we had to do with types defined in Swift on SentryOptions.h prior to this PR. To make it easy to follow this change I've kept the Swift type almost everywhere, but in a few header files I had to use `SentryOptionsObjc` which is just an alias for `NSObject`. There are extensions at the end of `Options.swift` that explains what is done in more detail.

#skip-changelog

noahsmartin:
> This change may look big, but I've done everything that can be done to make it as small as possible. It only looks big because the generated API json file changed, the actual lines of code added are ~800. So many files were touched because I had to remove a bunch of imports of the objc class

This converts SentryOptions.h/.m to SentryOptions.swift and in doing so removes the last blocker for using SPM. It also cleans up a lot of gross bridging we had to do, like with SentryExperimentalOptions and the beforeSendLog callback. This is now replaced with one piece of cross bridging, which I'll get to later...

Almost all the relevant implementation here is in `Options.swift`. This matches the old ObjC API almost exactly, except that the class is now `final`. The API json file changed a lot structurally just because it represents Swift and ObjC differently, not because the actual API changed.

The tricky part of this is that we can't call an objc header that uses a swift type from Swift code. (This works in xcodebuild but not cocoapods/SPM) To work around this the value that gets passed to`SentrySDKInternal` needs to be type-erased. This is the same as what we had to do with types defined in Swift on SentryOptions.h prior to this PR. To make it easy to follow this change I've kept the Swift type almost everywhere, but in a few header files I had to use `SentryOptionsObjc` which is just an alias for `NSObject`. There are extensions at the end of `Options.swift` that explains what is done in more detail.

#skip-changelog

Closes #6690

### DIFF
--- a/Sources/Sentry/Profiling/SentryProfilingSwiftHelpers.m
+++ b/Sources/Sentry/Profiling/SentryProfilingSwiftHelpers.m
@@ -167,12 +167,20 @@ sentry_scheduledTimerWithTarget(
 BOOL
 sentry_appHangsDisabled(void)
 {
-    return [[[[SentrySDKInternal currentHub] getClient] options] isAppHangTrackingDisabled];
+    SentryOptions *options = [[[SentrySDKInternal currentHub] getClient] options];
+    if (options == nil) {
+        return NO;
+    }
+    return [options isAppHangTrackingDisabled];
 }
 BOOL
 sentry_autoPerformanceTracingDisabled(void)
 {
-    return ![[[[SentrySDKInternal currentHub] getClient] options] enableAutoPerformanceTracing];
+    SentryOptions *options = [[[SentrySDKInternal currentHub] getClient] options];
+    if (options == nil) {
+        return YES;
+    }
+    return ![options enableAutoPerformanceTracing];
 }
 void
 sentry_startFramesTracker(void)
--- a/Sources/Sentry/include/SentryOptionsObjC.h
+++ b/Sources/Sentry/include/SentryOptionsObjC.h
@@ -1,1 +1,5 @@
+@class SentryOptionsHolder;
+
+// Swift code cannot see an objc header using a Swift type. So instead we use NSObject, but for
+// easier development we alias it to SentryOptionsObjC so the purpose of it is understood.
 @compatibility_alias SentryOptionsObjC NSObject;
--- a/Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
+++ b/Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
@@ -252,7 +252,7 @@ private extension SentryContinuousProfilerTests {
         }
     }
     
-    func performContinuousProfilingTest(expectedEnvironment: String = "production") throws {
+    func performContinuousProfilingTest(expectedEnvironment: String = Options.defaultEnvironment) throws {
         XCTAssertFalse(SentryContinuousProfiler.isCurrentlyProfiling())
         SentryContinuousProfiler.start()
         XCTAssert(SentryContinuousProfiler.isCurrentlyProfiling())
--- a/SentryTestUtils/Sources/TestClient.swift
+++ b/SentryTestUtils/Sources/TestClient.swift
@@ -1,6 +1,6 @@
 import _SentryPrivate
 import Foundation
-@_spi(Private) import Sentry
+@_spi(Private) @testable import Sentry
 
 public class TestClient: SentryClientInternal {
     public override init?(options: NSObject) {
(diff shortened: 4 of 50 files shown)


Sources/Sentry/SentryProfiler.mm
```diff
@@ -10,13 +10,13 @@
...
 #    import "SentryLogC.h"
 #    import "SentryMetricProfiler.h"
-#    import "SentryOptions+Private.h"
 #    import "SentryProfileConfiguration.h"
 #    import "SentryProfilerState+ObjCpp.h"
 #    import "SentryProfilerTestHelpers.h"
 #    import "SentryProfilingSwiftHelpers.h"
 #    import "SentrySDK+Private.h"
 #    import "SentrySamplingProfiler.hpp"
+#    import "SentrySwift.h"
```
itaybre: This is an ObjectiveC++ file, this swift import wont work :(

noahsmartin: It was already unused, and now it's deleted :)

Sources/Sentry/Profiling/SentryProfilingSwiftHelpers.m
```diff
@@ -159,6 +164,16 @@
...
+BOOL
+sentry_appHangsDisabled(void)
+{
+    return [[[[SentrySDKInternal currentHub] getClient] options] isAppHangTrackingDisabled];
+}
+BOOL
+sentry_autoPerformanceTracingDisabled(void)
+{
+    return ![[[[SentrySDKInternal currentHub] getClient] options] enableAutoPerformanceTracing];
+}
```
philprime, line 184: `m`: I believe the options could be nil, therefore this statement would return `nil` instead of a boolean value. I assume this is safe because it's Objective-C, but maybe we need proper null-handling here with debug logs

Sources/Sentry/include/SentryOptionsObjC.h
```diff
@@ -0,0 +1 @@
+@compatibility_alias SentryOptionsObjC NSObject;
```
philprime, line 5: `m`: Please add a documentation comment block to this type so future maintainers understand the purpose of this.

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
```diff
@@ -252,7 +252,7 @@ private extension SentryContinuousProfilerTests {
         }
     }
     
-    func performContinuousProfilingTest(expectedEnvironment: String = kSentryDefaultEnvironment) throws {
+    func performContinuousProfilingTest(expectedEnvironment: String = "production") throws {
```
philprime: `l`: Could this be defined as an enum somewhere? Would there be additional value?

Sources/Sentry/Profiling/SentryProfilingSwiftHelpers.m
noahsmartin: Main thing is to make this exactly the same logic as before, I will double check that behavior. Can make the code cleaner as long as the logic is the same

Tests/SentryProfilerTests/SentryContinuousProfilerTests.swift
noahsmartin: It could be defined as a swift `let` I'll try that

Sources/Sentry/Profiling/SentryProfilingSwiftHelpers.m
```diff
@@ -159,6 +164,24 @@
...
+BOOL
+sentry_autoPerformanceTracingDisabled(void)
+{
+    SentryOptions *options = [[[SentrySDKInternal currentHub] getClient] options];
+    if (options == nil) {
+        return YES;
+    }
+    return ![options enableAutoPerformanceTracing];
+}
```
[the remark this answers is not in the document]
noahsmartin: No, cursor has it reversed

Issue #3742, opened 2024-03-14
HTTP Client Errors don't impact session


philipphofmann:
### Description

The HTTP Client errors captured here https://github.com/getsentry/sentry-cocoa/blob/2de284cc01403c035708cf502d49bd0db0f0b847/Sources/Sentry/SentryNetworkTracker.m#L445

don't impact the session because captureEvent doesn't increment session errors in the Hub https://github.com/getsentry/sentry-cocoa/blob/2de284cc01403c035708cf502d49bd0db0f0b847/Sources/Sentry/SentryHub.m#L279-L300

PR #6633, opened 2025-11-03
fix: HTTP Client errors mark session as errored

philipphofmann:
## :scroll: Description

Automatically captured HTTP client errors now mark sessions as errored.

Fixes GH-3742

Unit tests and manually capturing a HTTP client error via the sample app: https://sentry-sdks.sentry.io/explore/releases/philipp-test-client-errors%2B1.0.0/?issuesType=all&project=5428557

Screenshot 2025-11-03 at 11 53 57

## :pencil: Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -270,14 +270,12 @@ jobs:
             device: "iPhone 16 Pro"
             scheme: "Sentry"
 
-          # iOS 26 - Download iOS 26.1 beta runtime on macOS-26
+          # iOS 26 - Use pre-installed iOS 26.1 runtime on macOS-26
           - name: iOS 26 Sentry
             runs-on: macos-26
             xcode: "26.1"
             test-destination-os: "26.1"
-            install_platforms: true
             platform: "iOS"
-            create_device: true
             device: "iPhone 17 Pro"
             scheme: "Sentry"
             # This runner seems to be slower than the others, so we give it more time because otherwise it frequently times out. The build step often alone takes 20 minutes and also booting the simulator takes a while.


Sources/Sentry/SentryClient.m
```diff
@@ -355,6 +341,22 @@ - (SentryId *)captureEvent:(SentryEvent *)event
         additionalEnvelopeItems:additionalEnvelopeItems];
 }
 
+- (SentryId *)captureErrorEvent:(SentryEvent *)event
```
philippsuess: I didn't add tests for that method because it's covered by the existing tests of `captureException` and `captureError`.

philippsuess:
Failing UI tests will be fixed with https://github.com/getsentry/sentry-cocoa/pull/6650.

Sources/Sentry/SentryHub.m
```diff
@@ -511,16 +534,38 @@ - (SentryId *)captureException:(NSException *)exception
 
 - (SentryId *)captureException:(NSException *)exception withScope:(SentryScope *)scope
 {
+    SentryId * (^captureClientBlock)(SentryClientInternal *) = ^SentryId *(
```
philprime: `m`: In previous PR feedback you mentioned that we should refrain from using NSBlock due to it's instability and memory unsafety. We should consider using private methods instead.

philprime (commented):
Almost LGTM

Sources/Sentry/SentryHub.m
philippsuess: > we should refrain from using NSBlock due to it's instability and memory unsafety.

When you're using async code and you're referencing other objects in it, yes. In this case, it should be good, but I also hate the syntax a lot, so let's do private methods instead. Thanks for pointing it out.

philippsuess: https://fuckingblocksyntax.com/ 😅

philippsuess:
Back to draft, because I'm trying to get rid of many NSBlocks.

PR #6644, opened 2025-11-03
ref: Cleanup SentryOptions implementation

### LINKED ISSUE #6645: ref: Cleanup SentryOptions implementation
> [!NOTE]
> The pull request "[ref: Cleanup SentryOptions implementation](https://github.com/getsentry/sentry-cocoa/pull/6644)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

#skip-changelog

noahsmartin:
I'm refactoring SentryOptions to be in Swift and figured we could take the opportunity to cleanup this implementation.

Closes #6645

### DIFF
--- a/Sources/Sentry/SentryOptions.m
+++ b/Sources/Sentry/SentryOptions.m
@@ -263,11 +263,6 @@ sentry_isValidSampleRate(NSNumber *sampleRate)
     }
 }
 
-- (void)setTracesSampler:(SentryTracesSamplerCallback)tracesSampler
-{
-    _tracesSampler = tracesSampler;
-}
-
 - (BOOL)isTracingEnabled
 {
     return (_tracesSampleRate != nil && [_tracesSampleRate doubleValue] > 0)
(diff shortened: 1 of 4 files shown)


Sources/Sentry/SentryOptions.m
```diff
@@ -289,39 +284,6 @@ - (BOOL)isProfilingCorrelatedToTraces
...
-        @"right configuration of Sentry that links UIKit.");
-#    endif // SENTRY_HAS_UIKIT
-}
-
-- (void)setAttachViewHierarchy:(BOOL)attachViewHierarchy
-{
-#    if SENTRY_HAS_UIKIT
-    _attachViewHierarchy = attachViewHierarchy;
-#    else
-    SENTRY_GRACEFUL_FATAL(
```
philprime, line 319: `m`: This removes the logging of this fatal warning. Is there a reason it's not worth keeping around?

noahsmartin: @philprime it's impossible to hit this case because this function is unused with the fix in the header file

noahsmartin: To explain a bit more: before this PR these functions would do nothing but log an error. With this PR the function is just deleted (breaking public API change) but there isn't any reason to have a function that does nothing but log an error, this makes it so the compiler enforces the error instead of the runtime

philprime: Thanks for the insights 👍

PR #6646, opened 2025-11-03
fix: Remove property that had no effect

### LINKED ISSUE #6647: fix: Remove property that had no effect
> [!NOTE]
> The pull request "[fix: Remove property that had no effect](https://github.com/getsentry/sentry-cocoa/pull/6646)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

#skip-changelog

noahsmartin:
This param did nothing because the default is to consider a library excluded. Since it does nothing, we can remove the property

#skip-changelog

Closes #6647


philprime:
Instead of removing the field we should fix the usage of it internally.

This field also exists for Sentry Java:

https://docs.sentry.io/platforms/java/configuration/options/#inAppExcludes

https://github.com/getsentry/sentry-java/blob/ab8a72db41b2e5c66e60cef3102294dddba90b20/sentry/src/main/java/io/sentry/SentryOptions.java#L239-L244

https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryStackTraceFactory.java#L81-L101

noahsmartin:
@philprime how would you fix it exactly? The default is for a frame that doesn't appear in inAppInclude or inAppExclude to be considered excluded, so since inAppInclude is checked first, inAppExclude does nothing

philippsuess:
Yes, currently `inAppExclude` does nothing. It's the same on Java, if I'm not mistaken.

https://github.com/getsentry/sentry-cocoa/blob/2b0eac07729f0d7f129f5423c801edd4c3a998f5/Sources/Swift/Core/Helper/SentryInAppLogic.swift#L75-L81

This is just extra logic and confuses users. I'm OK with removing it, but we must also update the docs here https://docs.sentry.io/platforms/apple/guides/ios/configuration/options/#inAppExclude

We should also udpate the code comment here https://github.com/getsentry/sentry-cocoa/blob/2b0eac07729f0d7f129f5423c801edd4c3a998f5/Sources/Swift/Core/Helper/SentryInAppLogic.swift#L51-L61

And also this branch still has a couple of occurrences of `inAppExcludes` which we should remove: 

Screenshot 2025-11-04 at 15 17 33

noahsmartin:
@philipphofmann I just made a PR for the docs: https://github.com/getsentry/sentry-docs/pull/15399

philippsuess:
I also created an issue https://github.com/getsentry/sentry-cocoa/issues/6677 and added it to the docs PR v9 milestone in Linear https://linear.app/getsentry/project/cocoa-sdk-900-19a03b661aba/overview#milestone-1389bf73-963e-490a-ac6e-3de6df5515fb, so we don't forget to merge it once V9 is out.

Issue #3901, opened 2024-04-25
Disable App Hangs for Widgets and Live Activities


philipphofmann:
### Description

The app hang detection might report false positives for Widgets and Live Activities, see https://github.com/getsentry/sentry-cocoa/issues/3616#issuecomment-2076599284. Therefore, it makes sense to disable that by default for Widgets and Live Activities, which we could do with something as pointed out [here](https://stackoverflow.com/a/64073922/1702958). As Widgets run in their own process and Live Activities in their own sandbox, app hangs could be useful for some special setups. Therefore, it's safer to do that in the next major.

PR #6670, opened 2025-11-04
feat(anr): Disable ANR tracking for widgets, live activities, action, intent and share extensions

philprime:
# :scroll: Description

This PR addresses issue #3901 by automatically disabling App Hang (ANR) tracking for Widgets, Live Activities, (Siri) Intent Extensions, Action Extensions and Share Extensions. These components run in separate processes or sandboxes with different execution characteristics, which can cause false positive app hang reports.

- Added `SentryExtensionDetector` to detect the current extension context by reading `Info.plist`
- Added `SentryExtensionType` enum for extension point identifiers (WidgetKit, Intents, Action, Share extensions)
- Modified `SentryANRTrackingIntegration` to check for extensions before installing ANR tracking
- Extended `SentryInfoPlistWrapper` to support reading dictionary values from `Info.plist`

**Testing:**
- Added comprehensive unit tests for `SentryExtensionDetector` covering all extension types and error cases
- Added integration tests for `SentryANRTrackingIntegration` verifying correct behavior in extension contexts
- Added tests for `SentryExtensionType` enum

**Sample Applications:**

To properly test the edge cases I created additional sample apps and extensions. Some of them required newer iOS versions than iOS 15 which is still used by iOS-Swift.

### Testing - General

I implemented samples with UI to visualize internal state using available UI, e.g. using a Siri Message Preview or a widget. As the App Hang tracker is disabled by default when a debugger is attached, you have two options to verify behaviour:

- Perform the test without a debugger attached
- Comment out the following lines of code, so that the App Hang Tracker is also installed with a debugger attached:

https://github.com/getsentry/sentry-cocoa/blob/3a3d93bfe5fcead190cf7804a400032cb1e09aa8/Sources/Sentry/SentryBaseIntegration.m#L136-L140

The easiest way to verify that the changes in this PR actually change anything is removing/commenting disabled extension types in the `SentryExtensionDetector`:

https://github.com/getsentry/sentry-cocoa/blob/00bb563ff3618fcf3246501f52f38f706d93ece0/Sources/Swift/Helper/SentryExtensionDetector.swift#L5-L12

### Testing - Share Extension

**Before:**

1. Remove the `.share` from the disabled app hang types in the `SentryExtensionDetector`
2. Run the Xcode scheme `iOS-Swift` on an iOS 18.6 simulator or iOS device
3. Stop the app
4. Open Safari on the device and navigate to any page, e.g. `https://docs.sentry.io`
5. Tap the share symbol in the toolbar
6. Tap on `iOS-Swift` in the list of applications
7. The share dialog will appear with `Sentry Enabled? ✅` but `ANR Disabled? ❌`, indicating that we are in the falsey state and still have ANR tracking enabled.

https://github.com/user-attachments/assets/7c201eb1-186a-42f0-ac6a-37b083e44af2

**After:**

1. Ensure the `.share` is in the list of disabled app hang types in the `SentryExtensionDetector` 
2. Perform steps 2 to 6 similar explained above
7. The share dialog will appear with `Sentry Enabled? ✅` but `ANR Disabled? ✅ `, indicating that we are in the truthy state and have ANR tracking disabled.

https://github.com/user-attachments/assets/179cb82a-8cf8-472a-8ba3-3f6beb809919

### Testing - Action Extension

1. Remove the `.action` from the disabled app hang types in the `SentryExtensionDetector`
2. Run the Xcode scheme `iOS-Swift` on an iOS 18.6 simulator or iOS device
3. Stop the app
4. Open Safari on the device and navigate to any page, e.g. `https://docs.sentry.io`
5. Tap the share symbol in the toolbar
6. Scroll to the bottom and tap on the action `iOS-Swift-ActionExtension`
7. The share dialog will appear with `✅ Sentry is enabled` but `❌ ANR Tracking installed`, indicating that we are in the falsey state and still have ANR tracking enabled.

https://github.com/user-attachments/assets/ca952290-034f-4733-9992-9b9778239015

**After:**

1. Ensure the `.action` is in the list of disabled app hang types in the `SentryExtensionDetector` 
2. Perform steps 2 to 6 similar explained above
7. The action view will appear with `✅ Sentry is enabled` and `✅ ANR Tracking not installed `, indicating that we are in the truthy state and have ANR tracking disabled.

https://github.com/user-attachments/assets/4812f986-4875-4cdb-8290-cb58c8081e92

### Testing - Intent Extension

> [!IMPORTANT]
> I was unsuccessful in using the extension in the simulator, but it worked on device. Therefore I recommend using a real device, but it might also work in the simulator

**Before:**

1. Remove the `.intent` from the disabled app hang types in the `SentryExtensionDetector`
2. Run the Xcode scheme `iOS-Swift` on an iOS device
3. Stop the app
4. Trigger Siri and ask it to "Send a message via iOS-Swift".
5. The sample will pick a fake recipient "John Snow" and show a preview message with `Sentry Enabled? ✅` but `ANR Disabled? ❌`
7. Tap on "Send" to view the sent message UI
5. The sample will show a sent message UI with `Sentry Enabled? ✅` but `ANR Disabled? ❌` too

NOTE: iOS sometimes seems to cache the preview but the message after "sending it" should be correct.

https://github.com/user-attachments/assets/e2b3d526-50e3-4e0a-a167-9bb14b336708

**After:**

1. Ensure the `.intent` is in the list of disabled app hang types in the `SentryExtensionDetector`
2. Perform steps 2 to 4 as explained above
...
5. The sample will pick a fake recipient "John Snow" and show a preview message with `Sentry Enabled? ✅` but `ANR Disabled? ✅`
7. Tap on "Send" to view the sent message UI
5. The sample will show a sent message UI with `Sentry Enabled? ✅` but `ANR Disabled? ✅` too

https://github.com/user-attachments/assets/bd86ba39-35a5-4296-9f3d-15f4be922153

### Testing - Widget

For this test we are disabling debugger detection and keep the debugger attached, because it causes a widget refresh on every run.

**Before:**

1. Remove the `.widget` from the disabled app hang types in the `SentryExtensionDetector`
2. Disable the debugger detection by commenting out the following code:

https://github.com/getsentry/sentry-cocoa/blob/3a3d93bfe5fcead190cf7804a400032cb1e09aa8/Sources/Sentry/SentryBaseIntegration.m#L136-L140

3. Run the Xcode scheme `iOS-SwiftUI-Widgets-WidgetsExtension` on an iOS simulator
4. It will install the app and create a widget on the home screen.
5. The widget will display `SDK Enabled? ✅` but `ANR Disabled? ❌` with the current timestamp (which is important because widgets can be outdated)

https://github.com/user-attachments/assets/c336a502-8c5f-40db-b6fd-c2ddd78b8a15

**After:**

1. Ensure the `.widget` is in the list of disabled app hang types in the `SentryExtensionDetector`
2. Perform steps 2 to 4 from the before test
3. The widget will display `SDK Enabled? ✅` and `ANR Disabled? ✅` with the current timestamp (which is important because widgets can be outdated)

https://github.com/user-attachments/assets/1ea2b494-ee2d-4015-830b-6f148073ff1f

### Testing - Widget Control

**Before:**

1. Remove the `.widget` from the disabled app hang types in the `SentryExtensionDetector`
2. Disable the debugger detection by commenting out the following code:

https://github.com/getsentry/sentry-cocoa/blob/3a3d93bfe5fcead190cf7804a400032cb1e09aa8/Sources/Sentry/SentryBaseIntegration.m#L136-L140

3. Run the Xcode scheme `iOS-SwiftUI-Widgets-WidgetControl` on an iOS simulator
4. It will install the app and create a control in the widget control center.
5. If the symbol of the control is a `X`, it indicates ANR tracking is enabled.

https://github.com/user-attachments/assets/e3ebae90-600e-47cb-a05e-5187e1dfda04

**After:**

1. Ensure the `.widget` is in the list of disabled app hang types in the `SentryExtensionDetector`
2. Disable the debugger detection by commenting out the following code:

https://github.com/getsentry/sentry-cocoa/blob/3a3d93bfe5fcead190cf7804a400032cb1e09aa8/Sources/Sentry/SentryBaseIntegration.m#L136-L140

3. Run the Xcode scheme `iOS-SwiftUI-Widgets-WidgetControl` on an iOS simulator
4. It will install the app and create a control in the widget control center.
5. If the symbol of the control is a `✔️`, it indicates ANR tracking is enabled.

https://github.com/user-attachments/assets/8962a356-c67b-4efb-bca8-931457d73498

### Testing - Live Activities

Live Activities are displayed differently on different devices, so I recommend you use an iPhone 16 Pro with an dynamic island to see it there too.

IMPORTANT:

During testing I experience multiple times that Live Activities were not started properly, with the push notification token never showing up. Further investigation showed that the live activity is actually a separate process which is called `WidgetRenderer_Activities` in combination with `chronod`. During compile-time some of the extensions seemed to perform only partial rebuilds, causing theses processes to not being able to unarchive the live activities.

Furthermore, live activities can be considered special kinds of widgets, therefore they are in the same kind of app extension.

**Before:**

1. Remove the `.widget` from the disabled app hang types in the `SentryExtensionDetector`
2. Disable the debugger detection by commenting out the following code:

https://github.com/getsentry/sentry-cocoa/blob/3a3d93bfe5fcead190cf7804a400032cb1e09aa8/Sources/Sentry/SentryBaseIntegration.m#L136-L140

3. Run the Xcode scheme `iOS-SwiftUI-Widgets` on an iOS simulator
4. Tap on the button "Start Live Activity"
5. It will start an indefinite local live activity and display the push notification token locally with "✅ Live Activity is running"
6. Lock the screen (CMD + L)
7. In the lockscreen it should display a live activity with "SDK Enabled? ✅" and "ANR Disabled? ❌) indicating that the ANR tracker is still installed
8. Unlock the simulator and go to the home screen (CMD + SHIFT + H, repeat 3x)
9. It will display a ✅  left to the dynamic island indicating the SDK is enabled, but ❌ on the right side indicating that the ANR tracker is still installed

https://github.com/user-attachments/assets/23485e23-183f-4a56-af50-8f33fda9b430

**After:**

1. Ensure the `.widget` is in the list of disabled app hang types in the `SentryExtensionDetector`
2. Perform steps 2 to 6 from the before test
3. In the lockscreen it should display a live activity with "SDK Enabled? ✅" and "ANR Disabled? ✅ ) indicating that the ANR tracker is not installed anymore
4. Unlock the simulator and go to the home screen (CMD + SHIFT + H, repeat 3x)
5. It will display a ✅  left to the dynamic island indicating the SDK is enabled, but ✅  on the right side indicating that the ANR tracker is not installed

https://github.com/user-attachments/assets/84f38a29-361a-4164-8a88-336ef7720b4c


Samples/iOS-Swift/iOS-Swift-ActionExtension.xcconfig
```diff
@@ -0,0 +1,20 @@
...
+PROVISIONING_PROFILE_SPECIFIER[sdk=macosx*] =
+
+CODE_SIGN_STYLE = Manual
+
+SUPPORTED_PLATFORMS = iphoneos iphonesimulator
+
+SWIFT_OBJC_BRIDGING_HEADER = iOS-Swift/Tools/iOS-Swift-Bridging-Header.h
+
+// Runtime search paths for app extensions to find frameworks in parent app
+LD_RUNPATH_SEARCH_PATHS = $(inherited) @executable_path/Frameworks @executable_path/../../Frameworks
```
itaybre, line 20: I might be overengineering this, but all xcconfig look the same except for the last (reverse) subdomain, maybe we could remove a couple lines on each target

Sources/Swift/Helper/Dependencies.swift
```diff
@@ -8,6 +8,9 @@
     @objc public static let sessionReplayEnvironmentChecker: SentrySessionReplayEnvironmentChecker = {
         SentrySessionReplayEnvironmentChecker(infoPlistWrapper: Dependencies.infoPlistWrapper)
     }()
+    @objc public static let extensionDetector: SentryExtensionDetector = {
```
itaybre: Since this is a let, it will be hard to modify this.
I wonder if you can inject `SentryExtensionDetector` as a parameter. It will make it easier to test it

itaybre (approved):
LGTM, left some comments for discussion

Samples/iOS-Swift/iOS-Swift-ActionExtension.xcconfig
philprime: We can definitely optimize this, but let's not do it in this PR, because it might also apply to other samples.

Sources/Swift/Helper/Dependencies.swift
philprime: This aligns with all the other properties in this class. Why would you want to modify it? We shouldn't use the `Dependencies` in the unit tests anyways and inject it into classes using it instead.

EDIT: I just noticed that it needs to be a a `var` so it can be overwritten when testing the `SentryANRTrackingIntegration.m`, which is the only place accessing the global dependency container. But then I need to double check why it's not necessary for the `sessionReplayEnvironmentChecker`

philprime: I'll leave a comment on the PR.

philprime:
While working on this PR I noticed that the `SentryANRTrackingIntegrationTests` are kind of patching a state of initialized dependencies, e.g. `install` only works when `SentryFileManager` is set in the `SentryDependencyContainer.sharedInstance()`, which can cause further issues in other tests due to global state. While we do call `clearTestState()` at the end, discarding the shared container, I believe this is bad practice and should be cleaned up.
To keep the scope of the PR where it is, we will do this in a follow-up PR.

PR #6686, opened 2025-11-05
chore: Require `ready-to-merge` label to run all checks

itaybre:
In order to speed up our development workflow, we plan to run a subset of test regularily on PRs (#6682).
And before merge, run the full suite of tests.

#skip-changelog

### DIFF
--- a/.github/workflows/ready-to-merge-workflow.yml
+++ b/.github/workflows/ready-to-merge-workflow.yml
@@ -37,5 +37,5 @@ jobs:
         if: ${{ inputs.is-pr && steps.check-label.outputs.label_found == 'true' }}
         run: echo "Label present. Proceeding with CI."
       - name: Gate passed
-        if: ${{ inputs.is-pr == 'false' }}
+        if: ${{ !inputs.is-pr }}
         run: echo "Not a PR. Proceeding with CI."


itaybre:
Example of what happens without the label
image

.github/workflows/ui-tests.yml
```diff
@@ -19,9 +20,18 @@ concurrency:
...
+    # Only validate the ready-to-merge gate for PR events.
+    if: github.event_name == 'pull_request'
+    uses: ./.github/workflows/ready-to-merge-workflow.yml
+    with:
+      labels: ${{ toJson(github.event.pull_request.labels) }}
+
   files-changed:
     name: Detect File Changes
     runs-on: ubuntu-latest
+    needs: ready
```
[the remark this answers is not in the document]
philippsuess: I think that's valid feedback.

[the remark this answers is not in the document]
itaybre: Yeah, I just updated the logic to have the ready job always run

.github/workflows/ready-to-merge-workflow.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  ready-to-merge-gate:
+    name: 'Missing "ready-to-merge" label'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail until the 'ready-to-merge' label is present
+        if: ${{ inputs.is-pr == 'true' && !contains(inputs.labels, 'ready-to-merge') }}
+        run: |
+          echo "Add the 'ready-to-merge' label to run CI."
+          exit 1
```
philippsuess, line 35: `m`: Why not skip but fail? The required checks should stop us from merging without all CI jobs. Skipping the not required jobs until the lable is present, seems a bit cleaner to me.

itaybre: Github's required workflows allow either `succeed` or `skipped`. If I skip everything, we will be able to accidentally merge without running the full suite of tests.
With this fail, we need to add the label before merging.

philippsuess: OK, then add this info as a comment please.

PR #6688, opened 2025-11-06
ci: Skip SPM resolution for compiling samples

philipphofmann:
Similar to #6673 but for compiling the samples directly with xcodebuild.

#skip-changelog


philippsuess:
Not working. Replaced with https://github.com/getsentry/sentry-cocoa/pull/6689.

Issue #6696, opened 2025-11-06
Unwanted dependency on `Sentry` when using `Sentry-WithoutUIKitOrAppKit`


letko-dmitry:
https://github.com/getsentry/sentry-cocoa/blob/f2460938031a47af2a45e38672d5e46271fa8a93/Package.swift#L76

It seems, when using the product `Sentry-WithoutUIKitOrAppKit` it still brings in the full `Sentry` package as a dependency. Probably, the root cause is that `Sentry-WithoutUIKitOrAppKit` depends on `SentryCppHelper` and `SentryCppHelper` depends on `Sentry`.

philippsuess:
@letko-dmitry, yes the `Sentry-WithoutUIKitOrAppKit` depends on `Sentry` but we have several macros and a compiler flag in our code base ensuring that product comes without UIKitOrAppKit see https://github.com/getsentry/sentry-cocoa/blob/a7a6331c84665a1893e58e65dcef77b69e73eee2/develop-docs/BUILD.md?plain=1#L9-L20

Do you have any problems with our current setup, @letko-dmitry?

letko-dmitry:
Thanks for the clarification. Right now I receive the following error when trying to build the product without UIKit managed by SPM:

```
SentryWithoutUIKit.h:10:13: fatal error: cyclic dependency in module 'SentryWithoutUIKit': SentryWithoutUIKit -> Sentry -> SentryWithoutUIKit
While building module 'SentryWithoutUIKit' imported from SentryWithoutUIKit-00e02de4.input:1:
```

Version: 8.57.2

letko-dmitry:
To clarify, I add the product `Sentry-WithoutUIKitOrAppKit` directly to the main app target in Xcode and also to one of the other dependencies through SPM the following way:

`.product(name: "Sentry-WithoutUIKitOrAppKit", package: "sentry-cocoa"),`

philippsuess:
Thanks, I can easily reproduce this by creating a macOS command line app and importing `Sentry-WithoutUIKitOrAppKit` [SentryCommandLine.zip](https://github.com/user-attachments/files/23414338/SentryCommandLine.zip).

That's the error I get:
```
Clang dependency scanner failure: While building module 'Sentry' imported from Sentry-91554be0.input:1:
[...]

PR #6754, opened 2025-11-11
fix: Remove full sentry being included when using SentryWithoutUIKit

### LINKED ISSUE #6755: fix: Remove full sentry being included when using SentryWithoutUIKit
> [!NOTE]
> The pull request "[fix: Remove full sentry being included when using SentryWithoutUIKit](https://github.com/getsentry/sentry-cocoa/pull/6754)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Looks like SentryCppHelper is depending on `Sentry` which is actually not necesary.
This fixes the issue

Fixes: https://github.com/getsentry/sentry-cocoa/issues/6696

itaybre:
Looks like `SentryCppHelper` is depending on `Sentry` which is actually not necesary.
This fixes the issue.

`SentryCppHelper` objective is lo add c++ to the linker, nothing else.
Adding Sentry as a dependency was a mistake.

Fixes: https://github.com/getsentry/sentry-cocoa/issues/6696

Closes #6755

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,6 @@
 
 ## Unreleased
 
-### Fixes
-
-- Remove unnecesary dependency on `SentryCppHelper` to Sentry (#6754)
-## Features
-
-- Structured Logs: Add log APIs to `Hub` and `Client` (#6737)
-
 ### Breaking Changes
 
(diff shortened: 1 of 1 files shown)


philprime (approved):
LGTM

philippsuess:
Is this something we should also merge to v8.x.x, @itaybre?

itaybre:
> Is this something we should also merge to v8.x.x, @itaybre?

We can do it there, it is an easy fix

PR #6721, opened 2025-11-07
chore: Update workflow to fetch labels using github API

itaybre:
Looks like GH actions events may be missing the PR labels.

This PR fetch them manually instead of depending on the obtained data

Closes #6722

PR #6756, opened 2025-11-12
fix(security): Resolve Command Injection vulnerabilities

Jeffreyhung:
Resolves the following Linear tickets:
- https://linear.app/getsentry/issue/ENG-5874/command-injection-vulnerabilities-in-getsentrysentry-cocoa-in-8
- https://linear.app/getsentry/issue/ENG-5875/command-injection-vulnerabilities-in-getsentrysentry-cocoa-in-8
- https://linear.app/getsentry/issue/ENG-5876/command-injection-vulnerabilities-in-getsentrysentry-cocoa-in-8
- https://linear.app/getsentry/issue/ENG-5879/command-injection-vulnerabilities-in-getsentrysentry-cocoa-in-8

--- 
[Copy-pasted from Semgrep]
Using variable interpolation $...`` with github context data in a run: step could allow an attacker to inject their own code into the runner. This would allow them to steal secrets and code. github context data can have arbitrary user input and should be treated as untrusted. Instead, use an intermediate environment variable with env: to store the data and use the environment variable in the run: script. Be sure to use double-quotes the environment variable, like this: "$ENVVAR".

Severity: High

References:

[semgrep rule](https://semgrep.dev/r/yaml.github-actions.security.run-shell-injection.run-shell-injection)

https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#understanding-the-risk-of-script-injections

https://securitylab.github.com/research/github-actions-untrusted-input/

#skip-changelog


philprime:
Thank you for tackling this issue. It looks to me like some of the issues might actually be resolved by merging https://github.com/getsentry/sentry-cocoa/pull/6721 so I would like to wait for that one to be merged first.

Jeffreyhung:
@philprime sounds good! That PR resolved the issues in `ready-to-merge` workflow, but not the rest, so we will still need this.

PR #6814, opened 2025-11-18
chore: Replace macOS runners with xcode image

itaybre:
Test using Cirrus Runners images with xcode version set


itaybre:
Confirmed by Cirrus Runners that these images are just syntactic sugar, so they actually contain all other images.

Because of this, there is no benefit on using them instead of the regular image.
Closing the PR

PR #6827, opened 2025-11-19
docs: Add strategy for project structure refactoring

philprime:
#skip-changelog

### DIFF
--- a/docs/project-structure.md
+++ b/docs/project-structure.md
@@ -1,0 +1,488 @@
+# Project Structure
+
+## Xcode Projects
+
+Xcode projects are the traditional way to build Swift code. They provide a comprehensive build system for iOS, macOS, and other Apple platforms.
+
+### Xcode Projects - Targets
+
+Xcode projects are built using targets. Targets are the basic building blocks of an Xcode project and define how code is compiled and linked.
+
+Targets configure the compilation process, specifying:
+
+- Source files to compile
+- Build settings (compiler flags, frameworks, libraries)
+- Dependencies on other targets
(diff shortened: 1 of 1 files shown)


philprime:
### From Project Structure Discussion (Nov 20, 2025):

Attendees: @philprime @philipphofmann @itaybre @noahsmartin 

Instead of adding targets to Xcode (even if in a SPM structure) we should migrate to SPM first and use targets there

Order of migration:
1. We should start off with removing `SentryDependencyContainer.sharedInstance()`
2. Replace Cocoapods with SPM to build Xcode framework
3. Remove Xcode project to test
4. Remove Xcode project to build
5. Rewrite integration types in Swift (e.g. `SentryCrashIntegration.m`) ← highest priority

Step 1, 2, 3 and 5 can be parallelized.

Important: Framework built with Cocoapods is not the same as a framework built with SPM.

Caveat: Internal types need to be public to be exposed when using in hierarchy:

```swift
// Package.swift

.target(name: "TargetA", dependencies: ["TargetB")

// Target A
import TargetB

public class TargetA {
  public func getUser() -> User? {
    nil
  }
}

// Target B

public class User {}
```

We would have to wrap internal types:

```
// Target A
import TargetB

public class TargetA {
  public func getUser() -> UserWrapper? {
    nil
  }
}

public class UserWrapper {
  private let user: User
}
```

Modules are not the only way to enforce architecture, e.g. using namespaces via `enum` types.

- [x] Create issues for todo list (@philprime)

philprime:
Closing this in favor of follow-up issues

PR #6828, opened 2025-11-19
Auto start app launch

noahsmartin:
The SDK drops app start times if the app doesn't create a ui.load transaction within 5 seconds of app start. This causes a lot of user confusion and makes it difficult to get app launch measurements. This PR instead makes the SDK automatically send app start spans, so no extra code from the user is required to get it working.

One thing missing from the implementation is to suppress sending the auto created transaction if the span was already included in another transaction

#skip-changelog


philprime:
@noahsmartin do you have any plans to continue this PR to get it merged?
If not, I would like to propose we close the PR for now until we have the resources available to work on it.

christophaigner:
Closing for now, we can reopen if necessary 👍🏻

PR #6830, opened 2025-11-20
feat: Add attributes data to `SentryScope`

### LINKED ISSUE #6835: feat: Add attributes data to `SentryScope`
> [!NOTE]
> The pull request "[feat: Add attributes data to `SentryScope`](https://github.com/getsentry/sentry-cocoa/pull/6830)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Adds `setAttribute` and `removeAttribute` to `SentryScope` as per https://develop.sentry.dev/sdk/telemetry/scopes/#scope-methods

These attributes will be used for logs in a following PR.
Currently units are not supported since Logs don't support them (yet).

itaybre:
Adds `setAttribute` and `removeAttribute` to `SentryScope` as per https://develop.sentry.dev/sdk/telemetry/scopes/#scope-methods

These attributes will be used for logs in a following PR.
Currently units are not supported since Logs don't support them (yet).

Closes #6835

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add attributes data to `SentryScope` (#6830)
+
 ## 9.0.0
 
 This changelog lists every breaking change. For a high-level overview and upgrade guidance, see the [migration guide](https://docs.sentry.io/platforms/apple/migration/).
@@ -53,7 +59,6 @@ This changelog lists every breaking change. For a high-level overview and upgrad
 
 ### Features
 
-- Add attributes data to `SentryScope` (#6830)
 - Add SentryDistribution as Swift Package Manager target (#6149)
 - Moves SentryClient and SentryHub to be written in Swift (#6627)
 - Add option `enablePropagateTraceparent` to support OTel/W3C trace propagation (#6356)


Samples/iOS-Swift/iOS-Swift/ViewControllers/ScopeViewController.swift
```diff
@@ -0,0 +1,61 @@
+import Sentry
+import UIKit
+
+class ScopeViewController: UIViewController {
```
philippsuess, line 4: `m`: What drove you to add a view controller for this? Do we need UI to test this?

Tests/SentryTests/SentryScopeTests.m
```diff
@@ -137,4 +137,44 @@ - (void)testReplaySerializes
...
+
+- (void)testRemoveAttribute
+{
+    SentryScope *scope = [[SentryScope alloc] init];
+    [scope setAttributeValue:@"test-string" forKey:@"a-key"];
+    XCTAssertEqualObjects([scope attributes][@"a-key"], @"test-string");
+
+    [scope removeAttributeForKey:@"a-key"];
+    XCTAssertEqualObjects([scope attributes][@"a-key"], nil);
+}
```
philippsuess: `m`: Please migrate these to `SentryScopeSwiftTests` instead.

Sources/Sentry/Public/SentryScope.h
```diff
@@ -137,6 +141,17 @@ NS_SWIFT_NAME(Scope)
...
+/**
+ * Set global attributes. Attributes are searchable key/value string pairs attached to every log
+ * message.
+ */
+- (void)setAttributeValue:(id)value forKey:(NSString *)key NS_SWIFT_NAME(setAttribute(value:key:));
+
+/**
+ * Remove the attribute for the specified key.
+ */
+- (void)removeAttributeForKey:(NSString *)key NS_SWIFT_NAME(removeAttribute(key:));
```
philippsuess, line 161: `h`: We have to be very clear that the attributes are only applied to logs and in the future metrics and not for the rest, such as events, transactions, spans, profiles, SR.  The same applies to the property above.

Sources/Sentry/SentryScope.m
```diff
@@ -49,6 +51,7 @@ - (instancetype)initWithMaxBreadcrumbs:(NSInteger)maxBreadcrumbs
         self.contextDictionary = [NSMutableDictionary new];
         self.attachmentArray = [NSMutableArray new];
         self.fingerprintArray = [NSMutableArray new];
+        self.attributesDictionary = [NSMutableDictionary new];
```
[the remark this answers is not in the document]
philippsuess: `h`: This is accurate feedback. Please also add to `initWithScope`.

Sources/Sentry/SentryScope.m
```diff
@@ -673,6 +675,40 @@ - (NSString *)propagationContextTraceIdString
...
+- (NSDictionary<NSString *, id> *)attributes
+{
+    @synchronized(_attributesDictionary) {
+        return _attributesDictionary.copy;
+    }
+}
+
+- (void)setAttributeValue:(id)value forKey:(NSString *)key
+{
+    if (key == nil || key.length == 0) {
```
philippsuess: `m`: You can set an empty key to a dict. Strictly speaking, it works. 
```suggestion
    if (key == nil) {
```

Sources/Sentry/SentryScope.m
```diff
@@ -673,6 +675,40 @@ - (NSString *)propagationContextTraceIdString
     return [self.propagationContext.traceId sentryIdString];
 }
 
+- (NSDictionary<NSString *, id> *)attributes
```
philippsuess: `l`: I would move all these methods a bit up in the class and put it below `- (NSArray<SentryAttachment *> *)attachments`. These are public methods and I think they should be close to the other public methods.

Sources/Sentry/SentryScope.m
```diff
@@ -673,6 +675,40 @@ - (NSString *)propagationContextTraceIdString
...
+        SENTRY_LOG_ERROR(@"Attribute's key cannot be nil nor empty");
+        return;
+    }
+
+    @synchronized(_attributesDictionary) {
+        _attributesDictionary[key] = value;
+
+        // ScopeObservers are not called since at this moment attributes are only used for Logs,
+        // which the LogBatcher obtains manually. At this moment not even Spans use this attributes.
+        // Should this change, we will need to call the observers.
```
philippsuess: `m`: The key difference is that the scope observes are used for watchdog terminations and crashes. As the SDK creates these events when a crash occurs or the next time the app launches for watchdog terminations, the scope data in memory would be lost. Logs instead are created on the fly. For logs we must ensure we don't loose the logs in memory for crashes and watchdog terminations, which we don't solve with scope observers, but in the batch processor instead by implementing a double rotating buffer: https://develop.sentry.dev/sdk/telemetry/telemetry-buffer/mobile-telemetry-buffer/#3-doublerotatingbuffer. So even for spans we won't change this. We could already put this info in this comment that we don't need to sync these attributes to the scope observers.

Sources/Sentry/SentryScope.m
```diff
@@ -672,6 +675,40 @@ - (NSString *)propagationContextTraceIdString
...
+    }
+}
+
 @end
 
 NS_ASSUME_NONNULL_END
```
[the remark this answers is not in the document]
philippsuess: `h`: Yes, I think that we must remove the attributes also when calling scope.clear. I opened a develop docs PR to clarify that https://github.com/getsentry/sentry-docs/pull/15578, but I'm not 100% it's expected behaviour. Let's wait for Michi to confirm.

Sources/Sentry/SentryScope.m
```diff
@@ -50,6 +51,7 @@ - (instancetype)initWithMaxBreadcrumbs:(NSInteger)maxBreadcrumbs
         self.contextDictionary = [NSMutableDictionary new];
         self.attachmentArray = [NSMutableArray new];
         self.fingerprintArray = [NSMutableArray new];
+        self.attributesDictionary = [NSMutableDictionary new];
```
philippsuess, line 55: `h`: The serialize method doesn't add the attributes. I just realized it doesn't make any sense that the scope implements the `SentrySerializable` protocol, so we could actually still ditch it in V9. If not, we should add the attributes to the serialize method, because otherwise it's quite confusing.

philipphofmann (commented):
Please also update the `SentryScope+Equality.m` test file.

> Currently units are not supported since Logs don't support them (yet).

Yes, but please double-check that we can add these without a breaking change. I think we should be good because we will simply add method overloads, but it is still worth checking if that works.

We should merge this PR including https://github.com/getsentry/sentry-cocoa/pull/6834 to main cause otherwise the attributes are useless. Thanks for splitting this up into two PRs.

Samples/iOS-Swift/iOS-Swift/ViewControllers/ScopeViewController.swift
itaybre: This is for manual testing, no UI test were created.

Sources/Sentry/SentryScope.m
itaybre: Yes, but do they make sense as attributes?
@cleptric

itaybre: Talked with him, they cannot be empty, will add to the docs

Tests/SentryTests/SentryScopeTests.m
itaybre: Moved

Samples/iOS-Swift/iOS-Swift/ViewControllers/ScopeViewController.swift
philippsuess: I think we can easily test the functionality with unit tests. Do you think we're going to need this in the future for manual testing?

Sources/Sentry/SentryScope.m
```diff
@@ -28,6 +28,8 @@ @interface SentryScope ()
 
 @property (atomic, strong) NSMutableArray<SentryBreadcrumb *> *breadcrumbArray;
 
+@property (atomic, strong) NSMutableDictionary<NSString *, id> *attributesDictionary;
```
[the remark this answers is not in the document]
itaybre: This is intentional, attributes are only used for logs.
The serialization is only used in crashes

Samples/iOS-Swift/iOS-Swift/ViewControllers/ScopeViewController.swift
itaybre: I can remove it, but I don't see it hurting us to keep it.
Once we add metrics or other integrations it will be easier to use this

Sources/Sentry/SentryScope.m
itaybre: We can't ditch it, it is used for crashes (unless we move to codable)

Sources/Sentry/SentryScope.m
```diff
@@ -467,6 +474,42 @@ - (void)clearAttachments
...
+
+- (void)setAttributeValue:(id)value forKey:(NSString *)key
+{
+    if (key == nil || key.length == 0) {
+        SENTRY_LOG_ERROR(@"Attribute's key cannot be nil nor empty");
+        return;
+    }
+
+    @synchronized(_attributesDictionary) {
+        _attributesDictionary[key] = value;
```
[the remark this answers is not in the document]
itaybre: We are already assuming `NONNULL`

itaybre:
> Please also update the `SentryScope+Equality.m` test file.
> 
> > Currently units are not supported since Logs don't support them (yet).
> 
> Yes, but please double-check that we can add these without a breaking change. I think we should be good because we will simply add method overloads, but it is still worth checking if that works.
> 
> We should merge this PR including #6834 to main cause otherwise the attributes are useless. Thanks for splitting this up into two PRs.

We can just add a new method with the unit parameter, I initially had it like that until I realized we cannot use the unit at this moment.

I am waiting to merge this one and then the other PR.

Sources/Sentry/SentryScope.m
```diff
@@ -509,6 +552,9 @@ - (void)clearAttachments
     if (crumbs.count > 0) {
         [serializedData setValue:crumbs forKey:@"breadcrumbs"];
     }
+    if (self.attributes.count > 0) {
+        [serializedData setValue:[self attributes] forKey:@"attributes"];
+    }
```
philippsuess, line 555: `m`: https://github.com/getsentry/sentry-cocoa/pull/6830#discussion_r2557599224 you commented we don't serialize attributes, but here we do. I like that approach actually.

The `SentryCrashIntegration.configureScope` will now use the `attributes`. So we must remove them there https://github.com/getsentry/sentry-cocoa/blob/c5d37520bf6a94f44e792c99d140468598f0c300/Sources/Sentry/SentryCrashIntegration.m#L231-L244

Sources/Sentry/SentryScope.m
```diff
@@ -467,6 +474,42 @@ - (void)clearAttachments
...
+
+- (void)removeAttributeForKey:(NSString *)key
+{
+    @synchronized(_attributesDictionary) {
+        [_attributesDictionary removeObjectForKey:key];
+
+        // ScopeObservers are not called since at this moment attributes are only used for Logs,
+        // which are stored in memory, thus do not need to be stored in processor like
+        // WatchdogTermination events do. Attributes might be added to Crashes in a later time,
+        // which will need to revisit this.
```
philippsuess: `m`: I think it would actually be cleaner if we call the scope observer here. The scope shouldn't know that the WatchdogScopeObserver and the CrashScopeObserver don't need this data. They should decide this themselves and simply make these a NoOp with a comment explaining that they arent interested.

Samples/iOS-Swift/iOS-Swift/ViewControllers/ScopeViewController.swift
philippsuess: > see it hurting us to keep it.

It's code. I only want to keep it if it makes sense. If you do, keep it. If not, ditch it. It seems like you're in favor of keeping it. No strong opinion from my side.

Sources/Sentry/Public/SentryScope.h
```diff
@@ -137,6 +142,24 @@ NS_SWIFT_NAME(Scope)
  */
 - (void)addAttachment:(SentryAttachment *)attachment NS_SWIFT_NAME(addAttachment(_:));
 
+/**
+ * Set global attributes. Attributes are searchable key/value string pairs attached to every log
+ * message.
+ * @note Attributes are only supported for Logs and Metrics in the future. Other events such as
+ * Events, Transactions, Spans, Profiles, Session Replay will ignore these attributes.
```
philippsuess: `l`: A bit cleaner, IMO
```suggestion
 * @note The SDK only applies attributes to Logs. The SDK doesn't apply the attributes to
 * Events, Transactions, Spans, Profiles, Session Replay.
```

Sources/Sentry/SentryScope.m
```diff
@@ -191,6 +195,9 @@ - (void)clear
     @synchronized(_spanLock) {
         _span = nil;
     }
+    @synchronized(_attributesDictionary) {
+        [_attributesDictionary removeAllObjects];
+    }
```
philippsuess, line 200: Yes, thanks

Sources/Sentry/SentryScope.m
philippsuess: We could make the serialize internal and remove it from the public API.

Tests/SentryTests/SentryScopeSwiftTests.swift
```diff
@@ -358,6 +362,7 @@ class SentryScopeSwiftTests: XCTestCase {
         let expected = Scope(maxBreadcrumbs: fixture.maxBreadcrumbs)
         XCTAssertEqual(expected, scope)
         XCTAssertEqual(0, scope.attachments.count)
+        XCTAssertEqual(0, scope.attributes.count)
```
philippsuess, line 365: thanks

Sources/Sentry/SentryScope.m
itaybre: I don't have a strong opinion here, I can add the calls to the observers.
I was just thinking that the less code, the better.

Sources/Sentry/SentryCrashIntegration.m
```diff
@@ -239,6 +239,10 @@ - (void)configureScope
         userInfo[@"release"] = self.options.releaseName;
         userInfo[@"dist"] = self.options.dist;
 
+        // Crashes don't use the attributes field, we remove them to avoid uploading them
+        // unnecessarily.
+        [userInfo removeObjectForKey:@"attributes"];
```
philippsuess, line 244: `m`: Please add a test for this.

Sources/Sentry/SentryScope.m
```diff
@@ -467,6 +474,40 @@ - (void)clearAttachments
...
+        SENTRY_LOG_ERROR(@"Attribute's key cannot be nil nor empty");
+        return;
+    }
+
+    @synchronized(_attributesDictionary) {
+        _attributesDictionary[key] = value;
+
+        for (id<SentryScopeObserver> observer in self.observers) {
+            [observer setAttributes:_attributesDictionary];
+        }
```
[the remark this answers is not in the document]
philippsuess: I think that's valid feedback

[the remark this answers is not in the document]
itaybre: Interesting I didn't get any error, will add it

Sources/Sentry/SentryCrashIntegration.m
itaybre: Added

Tests/SentryTests/Integrations/SentryCrash/SentryCrashIntegrationTests.swift
```diff
@@ -624,6 +624,32 @@ class SentryCrashIntegrationTests: NotificationCenterTestCase {
...
+        let sentryCrash = fixture.sentryCrash
+        let sut = SentryCrashIntegration(crashAdapter: sentryCrash, andDispatchQueueWrapper: fixture.dispatchQueueWrapper)
+        
+        sut.install(with: Options())
+
+        let userInfo = try XCTUnwrap(SentryDependencyContainer.sharedInstance().crashReporter.userInfo)
+        // Double check the environment just set is in the user info
+        assertUserInfoField(userInfo: userInfo, key: "environment", expected: "test-attributes")
+        // Validate there is no attributes in the user info
+        XCTAssertNil(userInfo["attributes"])
```
philippsuess, line 650: Thanks 🥳

philipphofmann (approved):
LGTM, but maybe wait with merging after we release 9.0.0, so the release doesn't blow up even further.

itaybre:
> LGTM, but maybe wait with merging after we release 9.0.0, so the release doesn't blow up even further.

That was my plan 👍

PR #6660, opened 2025-11-04
ref: Remove SentryIntegrationProtocol from public API

noahsmartin:
Remove SentryIntegrationProtocol from public API

#skip-changelog

Closes #6661

PR #6862, opened 2025-11-20
ref: Use Swift integrations

### LINKED ISSUE #6863: ref: Use Swift integrations
> [!NOTE]
> The pull request "[ref: Use Swift integrations](https://github.com/getsentry/sentry-cocoa/pull/6862)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Adds a new protocol for integrations that takes dependencies as an input. For the easiest migration conformers of the protocol can satisfy the `Dependencies` associated type with just `typealias Dependencies = SentryDependencyContainer`. This PR shows how it can be used to make the dependencies just define the minimum necessary dependencies so it can be easily mocked in tests.

#skip-changelog

noahsmartin:
Adds a new protocol for integrations that takes dependencies as an input. For the easiest migration conformers of the protocol can satisfy the `Dependencies` associated type with just `typealias Dependencies = SentryDependencyContainer`. This PR shows how it can be used to make the dependencies just define the minimum necessary dependencies so it can be easily mocked in tests. There are a few benefits to this over the current objc protocol:

- The protocol defines a failable initializer rather than a function returning a boolean. This means properties in the integration can be `let` constants rather than vars with default nil value. Making these immutable with help with a lot of swift properties like thread safety, and remove the need from handling null cases that should be unreachable.
- The associatedtype for dependencies decouples the SentryDependencyContainer from the integration, so each integration only defines the dependencies it needs and doesn't have a reliance on a particular type providing those dependencies.
- No need for @objc in the integration classes
- Integration names are explicitly provided rather than inferred from class names

#skip-changelog

Closes #6863

### DIFF
--- a/Sources/Swift/Integrations/UserFeedback/UserFeedbackIntegration.swift
+++ b/Sources/Swift/Integrations/UserFeedback/UserFeedbackIntegration.swift
@@ -26,7 +26,7 @@ final class UserFeedbackIntegration<Dependencies: ScreenshotSourceProvider>: NSO
         }
     }
 
-    func uninstall() { }
+    func uninstall() { /* Empty on purpose. Nothing to uninstall. */ }
     
     static var name: String {
         "SentryUserFeedbackIntegration"
--- a/Tests/SentryTests/Integrations/Feedback/UserFeedbackIntegrationTests.swift
+++ b/Tests/SentryTests/Integrations/Feedback/UserFeedbackIntegrationTests.swift
@@ -11,39 +11,32 @@ final class UserFeedbackIntegrationTests: XCTestCase {
         return options
     }
     
-    static private var optionsWithoutFeedback: Options {
-        return Options()
+    private struct TestDependencies: ScreenshotSourceProvider {
+        let screenshotSource: SentryScreenshotSource?
     }
     
-    static private var screenshotSource: SentryScreenshotSource {
+    func testUsesCorrectName() {
+        XCTAssertEqual(UserFeedbackIntegration<TestDependencies>.name, "SentryUserFeedbackIntegration")
+    }
+    
+    func testInitializerFailsWhenNoScreenshotSource() {
+        let integration = UserFeedbackIntegration(with: Self.optionsWithFeedback, dependencies: TestDependencies(screenshotSource: nil))
+        XCTAssertNil(integration)
+    }
+    
+    func testInitializerSucceedsWhenScreenshotSourceIsPresent() {
         let viewRenderer = SentryDefaultViewRenderer()
         let photographer = SentryViewPhotographer(
            renderer: viewRenderer,
            redactOptions: Options().screenshot,
             enableMaskRendererV2: false)
-        return SentryScreenshotSource(photographer: photographer)
-    }
-    
-    private struct MockDependencies: ScreenshotSourceProvider {
-        let screenshotSource: SentryScreenshotSource?
-    }
-    
-    func testUsesCorrectName() {
-        XCTAssertEqual(UserFeedbackIntegration<MockDependencies>.name, "SentryUserFeedbackIntegration")
-    }
-    
-    func testInitializerFailsWhenNoScreenshotSource() {
-        let integration = UserFeedbackIntegration(with: Self.optionsWithFeedback, dependencies: MockDependencies(screenshotSource: nil))
-        XCTAssertNil(integration)
-    }
-    
-    func testInitializerSucceedsWhenScreenshotSourceIsPresent() {
-        let integration = UserFeedbackIntegration(with: Self.optionsWithFeedback, dependencies: MockDependencies(screenshotSource: Self.screenshotSource))
+        let screenshotSource = SentryScreenshotSource(photographer: photographer)
+        let integration = UserFeedbackIntegration(with: Self.optionsWithFeedback, dependencies: TestDependencies(screenshotSource: screenshotSource))
         XCTAssertNotNil(integration)
     }
     
     func testInitializerFailsWhenFeedbackNotConfigured() {
-        let integration = UserFeedbackIntegration(with: Self.optionsWithoutFeedback, dependencies: MockDependencies(screenshotSource: nil))
+        let integration = UserFeedbackIntegration(with: Options(), dependencies: TestDependencies(screenshotSource: nil))
         XCTAssertNil(integration)
     }
 }


Sources/Sentry/include/SentryHub+Private.h
```diff
@@ -27,7 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nullable, nonatomic, strong) SentrySession *session;
 
-@property (nonatomic, strong) NSMutableArray<id<SentryIntegrationProtocol>> *installedIntegrations;
+@property (nonatomic, strong) NSMutableArray<id> *installedIntegrations;
```
noahsmartin: The new version of this protocol is not accessible to ObjC, but as far as I can tell the fact this was previously `SentryIntegrationProtocol` wasn't important so it can just be replaced with `id`

Sources/Swift/Core/Integrations/Integrations.swift
```diff
@@ -0,0 +1,41 @@
...
+        name = NSStringFromClass(I.self)
+        self.integration = integration
+        start = {
+            integration.start(with: $0, dependencies: $1)
+        }
+    }
+    func start(with options: Options, dependencies: SentryDependencyContainer) -> Bool {
+        start(options, dependencies)
+    }
+}
```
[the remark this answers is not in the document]
noahsmartin: This is wrong

Sources/Swift/Core/Integrations/Integrations.swift
```diff
@@ -0,0 +1,42 @@
...
+
+protocol Integration: AnyObject {
+    // The dependencies required for the integration. The easiest way to satisfy this requirement when migrating from ObjC
+    // is to define it as `SentryDependencyContainer` with a typealias. However, a generic Swift class that has a protocol
+    // constraint on the `Dependencies` type can make it easier to test and to use without a direct dependency on all
+    // of `SentryDependencyContainer`.
+    associatedtype Dependencies
+    
+    // The initializer is failable, return nil if the integration was not installed, for example when the options that enable the integration is not enabled.
+    @MainActor init?(with options: Options, dependencies: Dependencies)
```
philippsuess: `h`: The MainActor requires `macOS 10.15+` https://developer.apple.com/documentation/Swift/MainActor
I wonder if we can use this after merging https://github.com/getsentry/sentry-cocoa/pull/6873.

Sources/Sentry/include/SentryHub+Private.h
philippsuess: `m`: That really feels wrong to me. Integrations still must adhere to `SentryIntegrationProtocol`. 
```suggestion
@property (nonatomic, strong) NSMutableArray<id<SentryIntegrationProtocol>> *installedIntegrations;
```

Sources/Swift/Core/Integrations/Integrations.swift
noahsmartin: I'm not seeing any compiler errors with it, but we can definitely remove it and add it later, just through I'd get ahead of it

Sources/Sentry/include/SentryHub+Private.h
noahsmartin: With this change they wouldn't have to anymore, because now they could adhere to `Sentry.Integration` which doesn't have an ObjC representation. We could still make some other base class or protocol that is visible to Swift, but I don't think it would provide any value. I'll add another commit demonstrating how we could still do this so we can talk about it

Sources/Swift/Core/Integrations/Integrations.swift
noahsmartin: Updated to remove use of MainActor, we can always try adding it later

Sources/Sentry/include/SentryHub+Private.h
noahsmartin: Good catch @philipphofmann there was one place where ObjC code still tried to dynamically call one of the methods on SentryIntegrationProtocol. So we can't quite get rid of the objc compatibility yet. I updated it to have a base protocol defined in ObjC that has the one function ObjC needs to know about (close) and now the Swift protocol also requires conformance to this objc protocol.

Sources/Sentry/include/HybridPublic/SentryIntegrationProtocol.h
```diff
@@ -4,17 +4,21 @@ NS_ASSUME_NONNULL_BEGIN
 
 @class SentryOptions;
 
-@protocol SentryIntegrationProtocol <NSObject>
+@protocol SentryBaseIntegrationProtocol <NSObject>
```
philippsuess: `h`: I find the naming of the  `SentryBaseIntegrationProtocol` and `SentryIntegrationProtocol` quite confusing. This protocol isn't public anymore since https://github.com/getsentry/sentry-cocoa/pull/6660. I don't think that RN or Flutter use this protocol actually. Maybe we can even remove it from HybridPublic. I couldn't find any usage of the protocol in RN and Dart
* RN no usage of `SentryIntegrationProtocol`: https://github.com/search?q=repo%3Agetsentry%2Fsentry-react-native%20SentryIntegrationProtocol&type=code
* RN no usage of `installwith`: https://github.com/search?q=repo%3Agetsentry%2Fsentry-react-native+installwith&type=code
* Dart no usage of https://github.com/search?q=repo%3Agetsentry%2Fsentry-dart+SentryIntegrationProtocol&type=code
* Dart no usage of `installwith`:https://github.com/search?q=repo%3Agetsentry%2Fsentry-dart+installwith&type=code
* Unity only references the header file, but doesn't use it: https://github.com/search?q=repo%3Agetsentry%2Funity%20entryIntegrationProtocol&type=code
* Unreal only references the header: https://github.com/search?q=repo%3Agetsentry%2Fsentry-unreal%20SentryIntegrationProtocol&type=code
* KMP: only references it but doens't use it: https://github.com/search?q=repo%3Agetsentry%2Fsentry-kotlin-multiplatform%20SentryIntegrationProtocol&type=code

So if I'm not mistaken, we should be able to change this protocol and make it internal. We can also completely change how we want to install all the integrations here 
https://github.com/getsentry/sentry-cocoa/blob/c5d37520bf6a94f44e792c99d140468598f0c300/Sources/Sentry/SentrySDKInternal.m#L552-L579

As all the logic is internal, I would rather change how we install and manage all our integrations. As we have to put all integration classes here already, we know all the classes. So we could add proper init methods for all these classes. All init methods would ask for the required dependencies. 

https://github.com/getsentry/sentry-cocoa/blob/c5d37520bf6a94f44e792c99d140468598f0c300/Sources/Sentry/SentrySDKInternal.m#L514-L547

So this PR slightly improves the situation, but it also makes it more complicated IMO. I would rather solve the underlying issue and invest some time to simplify how we manage our integrations. Maybe it also makes sense to discuss this on a call.

noahsmartin: Discussed this offline, we are going to move `SentryIntegrationProtocol` outside of HybridPublic but it's a bit more complicated since SentrySessionReplayIntegration is currently used in hybrid public. We will make a new class that is visible to hybrid public rather than the entire integration, but letting that be its own PR

Sources/Swift/Core/Integrations/Integrations.swift
```diff
@@ -0,0 +1,48 @@
...
+    init<I: Integration>(_ integration: I.Type) where I.Dependencies == SentryDependencyContainer {
+        name = I.name
+        start = {
+            integration.init(with: $0, dependencies: $1)
+        }
+    }
+}
+
+// Bridges to ObjC code to trigger installing the integrations
+@_spi(Private) @objc public final class SentryIntegrationInstaller: NSObject {
```
philippsuess: `l`: Maybe including Swift in the name makes it a bit cleaner.
```suggestion
@_spi(Private) @objc public final class SentrySwiftIntegrationInstaller: NSObject {
```

Sources/Swift/Core/Integrations/Integrations.swift
```diff
@@ -0,0 +1,48 @@
+@_implementationOnly import _SentryPrivate
+
+// The Swift counterpart to `SentryObjcIntegrationProtocol`. This protocol allows
+// injecting the dependencies in a way that does not require the Integration to
+// depend on SentryDependencyContainer.
+protocol Integration: SentryIntegrationProtocol {
```
philippsuess: `l`: Maybe a bit cleaner to clearly communicate that this is about Swift integrations.
```suggestion
protocol SwiftIntegration: SentryIntegrationProtocol {
```

Sources/Swift/Core/Integrations/Integrations.swift
```diff
@@ -0,0 +1,48 @@
...
+    // The initializer is failable, return nil if the integration was not installed, for example when the options that enable the integration is not enabled.
+    init?(with options: Options, dependencies: Dependencies)
+    
+    // Name of the integration that is used by `SentrySdkInfo`
+    static var name: String { get }
+}
+
+// Type erases the `Integration` so that it can be stored in an array and used for `addInstalledIntegration`
+private struct AnyIntegration {
+    let start: (Options, SentryDependencyContainer) -> SentryIntegrationProtocol?
```
philippsuess: `l`: I would stick to the naming pattern of install if possible. We install integrations in the SDK. Calling it start here is inconsistent.
```suggestion
    let install: (Options, SentryDependencyContainer) -> SentryIntegrationProtocol?
```

Sources/Swift/Core/Integrations/Integrations.swift
```diff
@@ -0,0 +1,48 @@
...
+// depend on SentryDependencyContainer.
+protocol Integration: SentryIntegrationProtocol {
+    // The dependencies required for the integration. The easiest way to satisfy this requirement when migrating from ObjC
+    // is to define it as `SentryDependencyContainer` with a typealias. However, a generic Swift class that has a protocol
+    // constraint on the `Dependencies` type can make it easier to test and to use without a direct dependency on all
+    // of `SentryDependencyContainer`.
+    associatedtype Dependencies
+    
+    // The initializer is failable, return nil if the integration was not installed, for example when the options that enable the integration is not enabled.
+    init?(with options: Options, dependencies: Dependencies)
```
philippsuess, line 14: `m`: I kind of find it a bit confusing that the init method returns `nil` when an integration shouldn't be installed. The init method than does two things: checking if it should be installed and initializing. It would be great if we could limit the init method only to initializing and have something extra for checking if an integration should be enabled or not. We can also follow up on this. That doesn't need to be in this PR.

Sources/Swift/Integrations/UserFeedback/SentryFeedbackAPI.swift
```diff
@@ -10,15 +10,17 @@ public final class SentryFeedbackAPI: NSObject {
     /// - seealso: See `SentryOptions.configureUserFeedback` to configure the widget.
     @available(iOSApplicationExtension, unavailable)
     @objc public func showWidget() {
-        SentryFeedbackAPIHelper.showWidget()
+        let feedback = SentrySDKInternal.currentHub().getInstalledIntegration(UserFeedbackIntegration<SentryDependencyContainer>.self) as? UserFeedbackIntegration<SentryDependencyContainer>
```
philippsuess: `l`: We could define a method for `SentrySDKInternal.currentHub().getInstalledIntegration(UserFeedbackIntegration<SentryDependencyContainer>.self) as? UserFeedbackIntegration<SentryDependencyContainer>` as it's pretty long code and used in two places.

Sources/Swift/Integrations/UserFeedback/UserFeedbackIntegration.swift
```diff
@@ -0,0 +1,36 @@
+@_implementationOnly import _SentryPrivate
+
+#if os(iOS) && !SENTRY_NO_UIKIT
+
+protocol ScreenshotSourceProvider {
+    var screenshotSource: SentryScreenshotSource? { get }
+}
+
+final class UserFeedbackIntegration<Dependencies: ScreenshotSourceProvider>: NSObject, Integration {
```
philippsuess: `m`: I think we should add some tests for that class to see how we can test that integration with the `ScreenshotSourceProvider` protocol.

philipphofmann (commented):
Thanks, I still found a view things, but we're getting close to a LGTM

Sources/Swift/Core/Integrations/Integrations.swift
noahsmartin: I think that would be difficult to do while maintaining useful optionality in Swift. Like if we had to return an instance even when we don't want to init some other types (because it's not installed) then we would have to make a bunch of properties Optional that really shouldn't be. The other option is to make it a throwing init, but I think that's less like what we would want than the failable init.

We could have another static method like `shouldInstall(options: Options) -> Bool`, but I don't think that would remove the failable init. We'd still have cases where we would want the init to return nil even when `shouldInstall(options: Options)` returns `true`. Maybe once we migrate some of the more complicated integrations we can revisit something like that?

philippsuess: Yes, let's revisit it later. Let's not block the PR by this discussion 🙏

Sources/Swift/Integrations/UserFeedback/UserFeedbackIntegration.swift
noahsmartin: Just added some, the tests are pretty straightforward because this class didn't have much logic, but should be a good example of how it works

Sources/Swift/Core/Integrations/Integrations.swift
```diff
@@ -0,0 +1,48 @@
...
+        let dependencies = SentryDependencyContainer.sharedInstance()
+        #if os(iOS) && !SENTRY_NO_UIKIT
+        let integrations: [AnyIntegration] = [.init(UserFeedbackIntegration<SentryDependencyContainer>.self)]
+        #else
+        let integrations: [AnyIntegration] = []
+        #endif
+        integrations.forEach { anyIntegration in
+            guard let integration = anyIntegration.install(options, dependencies) else { return }
+
+            SentrySDKInternal.currentHub().addInstalledIntegration(integration, name: anyIntegration.name)
```
philippsuess, line 45: `l`: Not having a refernce to the hub in here would be great, but we can't simply pass the `SentryInternalHub` because `_SentryPrivate` is marked as implementation only. Having a delegate for adding the installed integration might be overkill. Up to you, if you think it's better to remove the reference. No strong opinion,

Sources/Swift/Integrations/UserFeedback/SentryFeedbackAPI.swift
```diff
@@ -10,15 +10,19 @@ public final class SentryFeedbackAPI: NSObject {
...
     /// Hide the feedback widget button.
     /// - warning: This is an experimental feature and may still have bugs.
     /// - seealso: See `SentryOptions.configureUserFeedback` to configure the widget.
     @available(iOSApplicationExtension, unavailable)
     @objc public func hideWidget() {
-        SentryFeedbackAPIHelper.hideWidget()
+        getIntegration()?.driver.hideWidget()
+    }
+    
+    private func getIntegration() -> UserFeedbackIntegration<SentryDependencyContainer>? {
```
philippsuess, line 24: Thanks 👍

Sources/Swift/Integrations/UserFeedback/UserFeedbackIntegration.swift
```diff
@@ -0,0 +1,36 @@
...
+        }
+        
+        // The screenshot source is coupled to the options, but due to the dependency container being
+        // tightly to the options anyways, it was decided to not pass it to the container.
+        guard let screenshotSource = dependencies.screenshotSource else {
+            return nil
+        }
+
+        driver = SentryUserFeedbackIntegrationDriver(configuration: configuration, screenshotSource: screenshotSource) { feedback in
+            SentrySDK.capture(feedback: feedback)
```
philippsuess, line 25: Not for this PR, but we'd like to decouple the integrations from the static API as much as we can. Ideally this would interact with the hub that the integration is installed on instead, but not for this PR. FYI.

Sources/Swift/Integrations/UserFeedback/UserFeedbackIntegration.swift
```diff
@@ -0,0 +1,36 @@
...
+        }
+    }
+
+    func uninstall() { }
```
philippsuess: `l`: To clearly state this is empty on purpose.
```suggestion
    func uninstall() { /* Empty on purpose. Nothing to uninstall. */ }
```

Tests/SentryTests/Integrations/Feedback/UserFeedbackIntegrationTests.swift
```diff
@@ -0,0 +1,51 @@
...
+    }
+    
+    static private var screenshotSource: SentryScreenshotSource {
+        let viewRenderer = SentryDefaultViewRenderer()
+        let photographer = SentryViewPhotographer(
+           renderer: viewRenderer,
+           redactOptions: Options().screenshot,
+            enableMaskRendererV2: false)
+        return SentryScreenshotSource(photographer: photographer)
+    }
```
philippsuess: `l`: These two properties are only used in one test each. I would inline them so the tests are easier to read.

Tests/SentryTests/Integrations/Feedback/UserFeedbackIntegrationTests.swift
```diff
@@ -0,0 +1,51 @@
...
+    
+    private struct MockDependencies: ScreenshotSourceProvider {
+        let screenshotSource: SentryScreenshotSource?
+    }
```
philippsuess: `l`: It's not really a mock dependency as we're using an actual dependency. Calling it a mock suggests behavior verification to me, when it’s really just a straightforward test fixture. I would call it `TestDependencies` 
```suggestion
    private struct TestDependencies: ScreenshotSourceProvider {
        let screenshotSource: SentryScreenshotSource?
    }
```

Sources/Swift/Core/Integrations/Integrations.swift
noahsmartin: I tried that but it also doesn't work because SentryIntegrationProtocol is implemented in ObjC, I like the idea but think we should wait until the rest of the integrations are in Swift

PR #6897, opened 2025-11-25
ref: Remove unused session.user

### LINKED ISSUE #6898: ref: Remove unused session.user
> [!NOTE]
> The pull request "[ref: Remove unused session.user](https://github.com/getsentry/sentry-cocoa/pull/6897)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

The session.user isn't really used anywhere in the code. While the scope applied the user to the session, the session doesn't do anything with the user property. Therefore, we can remove it. Also, the [develop docs](https://develop.sentry.dev/sdk/telemetry/sessions/) don't mention a user property. Only the distinctID that should be the userID.

#skip-changelog

philipphofmann:
The session.user isn't really used anywhere in the code. While the scope applied the user to the session, the session doesn't do anything with the user property. Therefore, we can remove it. Also, the [develop docs](https://develop.sentry.dev/sdk/telemetry/sessions/) don't mention a user property. Only the distinctID that should be the userID.

#skip-changelog 

Closes #6898

### DIFF
--- a/Sources/Sentry/SentryScope.m
+++ b/Sources/Sentry/SentryScope.m
@@ -526,11 +526,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)applyToSession:(SentrySession *)session
 {
-    SentryUser *userObject = self.userObject;
-    if (userObject != nil) {
-        session.user = userObject.copy;
-    }
-
     NSString *environment = self.environmentString;
     if (environment != nil) {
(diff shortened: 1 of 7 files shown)


itaybre (approved):
LGTM, we should get in v9 since it is a breaking change

philippsuess:
Nope, I don't think it's a breaking change. The session isn't public.

PR #6902, opened 2025-11-25
docs: Add integrations documentation to developer-docs section

### LINKED ISSUE #6903: chore: Add integrations documentation to developer-docs section
> [!NOTE]
> The pull request "[chore: Add integrations documentation to developer-docs section](https://github.com/getsentry/sentry-cocoa/pull/6902)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Moving these docs from Notion to Github.

### Opinions on Notion
@itaybre:
I believe the option 3 (automated releases from sentry-cocoa) is the way forward because at this moment our biggest constraint are the man hours, adding more workload for us is not going to help us as a team, and once the set up is ready, we shouldn’t need to maintain the release workflow, all with having the best solution of users which is only importing the integration they need.

If going with a repo per integration, repo name scheme:

`sentry-cocoa-*<yourfavoriteloggingframework>*`

If going with a repo per integration type, repo name scheme:

`sentry-cocoa-<integration-type>`

@philprime 
I believe the option 3 (package-mirror) is the best way, using GitHub workflows to automatically update the downstream packages and tags accordingly. As the integrations are still tightly coupled with the main SDK code, we should keep the code and documentation close to each other. Asking users to add an additional dependency to their project is a good trade-off compared to having to download unnecessary ones (also impacts their CI workflows, e.g. longer download times cause more CI build minutes, cause more costs)

@philipphofmann 
I’m also for option 3 because all the code lives in one place, and we only need to use multiple repositories because of Apple's ecosystem package manager limitations. That limitation shouldn’t force us to split up the code across multiple repositories. We should abstract that away as much as we can from the codebase if possible, and that is what option 3 is about.

## Steps forward:
- [ ] Align repository on name structure
- [ ] Talk with GitHub admins to find the best way to securely deploy to other repositories
- [ ] Create Repositories for SwiftLog
  - Initial releases will probably be manual
- [ ] Create integrations project in sentry-cocoa
  - [ ] Create a CI template for integrations to run in an isolated context

itaybre:
Moving these docs from Notion to Github.

### Opinions on Notion
@itaybre:
I believe the option 3 (automated releases from sentry-cocoa) is the way forward because at this moment our biggest constraint are the man hours, adding more workload for us is not going to help us as a team, and once the set up is ready, we shouldn’t need to maintain the release workflow, all with having the best solution of users which is only importing the integration they need.

If going with a repo per integration, repo name scheme:

`sentry-cocoa-*<yourfavoriteloggingframework>*`

If going with a repo per integration type, repo name scheme:

`sentry-cocoa-<integration-type>`

@philprime 
I believe the option 3 (package-mirror) is the best way, using GitHub workflows to automatically update the downstream packages and tags accordingly. As the integrations are still tightly coupled with the main SDK code, we should keep the code and documentation close to each other. Asking users to add an additional dependency to their project is a good trade-off compared to having to download unnecessary ones (also impacts their CI workflows, e.g. longer download times cause more CI build minutes, cause more costs)

@philipphofmann 
I’m also for option 3 because all the code lives in one place, and we only need to use multiple repositories because of Apple's ecosystem package manager limitations. That limitation shouldn’t force us to split up the code across multiple repositories. We should abstract that away as much as we can from the codebase if possible, and that is what option 3 is about.

## Steps forward:
- [ ] Align repository on name structure
- [ ] Talk with GitHub admins to find the best way to securely deploy to other repositories
- [ ] Create Repositories for SwiftLog
  - Initial releases will probably be manual
- [ ] Create integrations project in sentry-cocoa
  - [ ] Create a CI template for integrations to run in an isolated context

Closes #6903


denrase:
@itaybre  We also got [this suggestions](https://github.com/getsentry/sentry-cocoa/pull/6738#issuecomment-3520840848) to use package traits. There dependencies can also be defined as optional. Glanced over it and it uses conditional compilation in the codebase, but I'm not sure this will play out with our approach to offer pre-compiled artefacts in SPM?

https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/packagetraits/

Maybe worth considering it though, if not for the main `Package.swift`, then maybe for a downstream `sentry-cocoa-integrations` `Package.swift`, as we'd also fetch multiple dependencies, even thou a user is only interested in integrating one sentry log integration.

For example, [swift-configuration](https://github.com/apple/swift-configuration/blob/1becb8e2059b530ce9d1188cb0dcfb7d95a3f07f/Package.swift#L15) has a trait for a `swift-log` integration.

```swift
.trait(
    name: "LoggingSupport",
    description: "Adds support for swift-log integration."
),

...

.product(
    name: "Logging",
    package: "swift-log",
    condition: .when(traits: ["LoggingSupport"])
),
```

itaybre:
> @itaybre We also got [this suggestions](https://github.com/getsentry/sentry-cocoa/pull/6738#issuecomment-3520840848) to use package traits. There dependencies can also be defined as optional. Glanced over it and it uses conditional compilation in the codebase, but I'm not sure this will play out with our approach to offer pre-compiled artefacts in SPM?
> 
> https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/packagetraits/
> 
> Maybe worth considering it though, if not for the main `Package.swift`, then maybe for a downstream `sentry-cocoa-integrations` `Package.swift`, as we'd also fetch multiple dependencies, even thou a user is only interested in integrating one sentry log integration.
> 
> For example, [swift-configuration](https://github.com/apple/swift-configuration/blob/1becb8e2059b530ce9d1188cb0dcfb7d95a3f07f/Package.swift#L15) has a trait for a `swift-log` integration.
> 
> ```swift
> .trait(
>     name: "LoggingSupport",
>     description: "Adds support for swift-log integration."
> ),
> 
> ...
> 
> .product(
>     name: "Logging",
>     package: "swift-log",
>     condition: .when(traits: ["LoggingSupport"])
> ),
> ```

Looks like traits are only supported in Swift 6.1+

develop-docs/INTEGRATIONS.md
```diff
@@ -0,0 +1,93 @@
+# Integrations in other repositories
```
philippsuess: `m`: I would clearly state it's about 3rd party integrations in the title.

develop-docs/INTEGRATIONS.md
```diff
@@ -0,0 +1,93 @@
...
+
+We have identified that **SPM** downloads _all_ declared dependencies in a package, even if none the added actually added modules use them.
+
+This means that if `sentry-cocoa` declares dependencies like **CocoaLumberjack** or **SwiftLog**, _all_ downstream consumers download these libraries, even if they don’t use the corresponding integrations.
+
+To avoid forcing unnecessary dependencies on users, we already agreed to **remove integrations from the main repository**.
```
philippsuess: `m`: I mean it's not removing from the repository, but from the Package.swift or any other package maner file actually.

develop-docs/INTEGRATIONS.md
```diff
@@ -0,0 +1,93 @@
...
+
+- Maintain flexibility in release schedules.
+
+### Agreed solution
+
+- **3: Keep all code in `sentry-cocoa`, but mirror releases into individual repositories**
+
+  SPM users import the integration repos, but implementation lives in `sentry-cocoa`.
+
+  Automated workflows push integration-specific code into dedicated repos during release.
```
philippsuess, line 29: `h`: I think we should explain the mirrored repository naming pattern in here and we should also mention the folder structure where the 3rd party integrations source code is going to live.

develop-docs/INTEGRATIONS.md
philprime: But is it a 3rd party integration? It's still maintained by us and treated as a first-party-citizen in the SDK

philippsuess: Sorry, I should have been more clear. It's about our integration for a 3rd party lib. We also have internal integrations. It's all a bit confusing.

develop-docs/DECISIONS.md
```diff
@@ -479,3 +479,84 @@ Related links:
...
+
+- Requires building a **custom mirroring release workflow**.
+- Potential risk of divergence if mirror fails or is misconfigured.
+- Release cadence may still be tied to `sentry-cocoa` unless new workflows are built.
+- Requires tooling to ensure code in the main repo remains cleanly partitioned.
+
+### Discarded options
+
+<details>
+  <summary>See options</summary>
```
philippsuess, line 526: Lovely thanks 🙏

develop-docs/INTEGRATIONS.md
```diff
@@ -0,0 +1,90 @@
...
+  - Mirrors how some SDKs manage platform-specific or optional components
+
+  Cons:
+
+  - Requires building a **custom mirroring release workflow**.
+  - Potential risk of divergence if mirror fails or is misconfigured.
+  - Release cadence may still be tied to `sentry-cocoa` unless new workflows are built.
+  - Requires tooling to ensure code in the main repo remains cleanly partitioned.
+
+> [!NOTE]
```
philippsuess, line 35: `l`: I think you can shorten this. We don't need pros and cons. They're explained in the decisions.

develop-docs/INTEGRATIONS.md
```diff
@@ -0,0 +1,90 @@
...
+- `Sources/` - Integration source code
+- `README.md` - Integration-specific documentation
+- `Package.swift` - SPM package definition
+- `*.podspec` - CocoaPods specification
+
+**Example:**
+
+```
+3rd-party-integration/
+  ├── CocoaLumberjack/
```
denrase: One thing i encountered while adding #6941 , `swift package resolve` will fail if one of the dependencies has the exact same name as the parent folder. I was using `swift-log` as the parent folder name. There is no issue with `SwiftLog`, but `CocoaLumberjack` will probably fail, as it matches the dependency name exactly. Just something to keep in mind, so we might want to add `Sentry` prefix in those cases if there is no other solution to this.

develop-docs/INTEGRATIONS.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Contributing moving forward
+
+All integration development will continue in the main `sentry-cocoa` repository, organized in dedicated subdirectories for clean CI isolation.
+
+#### Directory Structure
+
+Each integration will be self-contained in `3rd-party-integration/INTEGRATION-NAME/` with:
+
+- `Sources/` - Integration source code
```
denrase, line 46: We should also add the `Tests` folder.

develop-docs/INTEGRATIONS.md
denrase: Image

denrase: One solution would be to use our name as the folder name, if this was the intent with `3rd-party-integration/INTEGRATION-NAME/`. So it would be named `SentrySwiftLog` instead of `SwiftLog`, `SentryCocoaLumberjack` etc. This will also be consistent if we ever have an integration wich uses multiple 3rd party dependencies. Also it will somewhat create a connection to the repo name where this code will be pushed to.

itaybre: Added this, thanks for noticing this

Issue #5424, opened 2025-06-17
Dispatch access of UIWindow screen to main thread in SentryCrashWrapper


philprime:
### Description

As discussed in #3559, #5393 and #4015, there is a edge case where accessing the `[SentryHub scope]` before starting the SDK or after closing the SDK (when `_scope` is nil) will create a new scope and call `[SentryCrashWrapper enrichScope:scope]`. 

In `enrichScope` we are accessing the current app windows via `SentryDependencyContainer.sharedInstance.application.windows`, which correctly dispatches the access to the current application windows to the main thread as expected, because UIKit object must only be accessed from main thread.

https://github.com/getsentry/sentry-cocoa/blob/2e790c65299f50859e726bdbaa93dafe178bf2fb/Sources/Sentry/SentryCrashWrapper.m#L187-L194
https://github.com/getsentry/sentry-cocoa/blob/2e790c65299f50859e726bdbaa93dafe178bf2fb/Sources/Sentry/SentryUIApplication.m#L74-L79

But the references to the windows are then transferred to the calling thread where `enrichScope` is executed, therefore the references are accessed from a background thread when calling `appWindows.firstObject.screen` and `appScreen.bounds.size.height`.

We need to move the access to the screen size to the main thread in `SentryUIApplication`, e.g. `[SentryDependencyContainer.sharedInstance.application getActiveWindowScreenSize]` and return a non-reference return value, i.e. a `CGSize`, instead.

philprime:
From Cocoa Sync 08/16:<br><br>We need to investigate if the `windows` are not used anywhere else and instead just return the data that is requested from the `windows` so all access to UIKit objects is done on the `main` thread

PR #6905, opened 2025-11-25
fix: UIWindow being possibly accessed from a background thread in SentryCrashWrapper

itaybre:
Under some scenarios, calling `addBreadcrumb` in the background, may en up in SentryCrashWrapper using UIWindows in the background thread.
Since all we need is the window's size, this is enough.

Closes: #5424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 
 - Fix UIWindow being possibly accessed from a background thread in SentryCrashWrapper (#6905)
 - Ensure SentrySDK.close resets everything on the main thread (#6907)
-- Ensure SentrySDK.close resets everything on the main thread (#6907)
 - Allow transaction tags to be accessed and modified in `beforeSend` (#6910)
 
 ## 9.0.0-rc.1


Sources/Swift/Helper/SentryApplicationExtensions.swift
```diff
@@ -55,6 +59,19 @@ extension SentryApplication {
         return Array(windows)
     }
     
+    public func internal_getActiveWindowSize() -> CGSize {
```
noahsmartin, line 65: Can we make this Optional instead of defaulting to zero?

Sources/Swift/Helper/SentryApplicationExtensions.swift
```diff
@@ -55,6 +59,19 @@ extension SentryApplication {
...
+        var size = CGSize.zero
+        Dependencies.dispatchQueueWrapper.dispatchSyncOnMainQueue({ [weak self] in
```
noahsmartin, line 67: Since this is sync, just need to make sure that no public APIs call it directly. Like we wouldn't want addBreadcrumb to internally block until the main thread runs, we'd need it to return before the breadcrumb is actually added, right?

Sources/Swift/Helper/SentryApplicationExtensions.swift
```diff
@@ -55,6 +59,19 @@ extension SentryApplication {
...
+            guard let self,
+                  let window = self.internal_getWindows()?.first else {
+                return
+            }
+            
+            size = window.bounds.size
```
philprime: `h`: As mentioned by Cursor but, you changed from the `window.screen.size` to `window.bounds.size`. I don't think this is correct

itaybre: ohh right, it was `screen.bounds`, not just `bounds`

Sources/Swift/Helper/SentryApplicationExtensions.swift
itaybre: It doesn't play nice with ObjC, this is why I am using the default `.zero`

Sources/Swift/Helper/SentryApplicationExtensions.swift
itaybre: `internal_getWindows` already ran synchronously in the main thread, so nothing actually changed.

PR #6907, opened 2025-11-25
fix: Ensure SentrySDK.close resets everything on the main thread

itaybre:
When working on #6905, I found that closing the SDK in the background executes everything in the same thread.

This is different from the initialization which executes in the main thread.

Linked to this, I also found an unsafe access to application state in that background thread:
Screenshot 2025-11-25 at 5 34 28 PM

### DIFF
--- a/Sources/Sentry/SentrySDKInternal.m
+++ b/Sources/Sentry/SentrySDKInternal.m
@@ -613,40 +613,39 @@ static NSDate *_Nullable startTimestamp = nil;
 {
     SENTRY_LOG_DEBUG(@"Starting to close SDK.");
 
-    startTimestamp = nil;
-
-    [SentryDependencyContainer.sharedInstance.dispatchQueueWrapper
-        dispatchAsyncOnMainQueueIfNotMainThread:^{
+    [SentryDependencyContainer.sharedInstance.dispatchQueueWrapper dispatchSyncOnMainQueue:^{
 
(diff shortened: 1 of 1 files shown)


Sources/Sentry/SentrySDKInternal.m
```diff
@@ -613,36 +613,40 @@ + (void)close
 {
     SENTRY_LOG_DEBUG(@"Starting to close SDK.");
 
+    startTimestamp = nil;
```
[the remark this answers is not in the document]
noahsmartin: Was about to comment this same thing, shouldn't it be in the main queue check too?

Sources/Sentry/SentrySDKInternal.m
```diff
@@ -613,36 +613,40 @@ + (void)close
...
+
+    [SentryDependencyContainer.sharedInstance.dispatchQueueWrapper
+        dispatchAsyncOnMainQueueIfNotMainThread:^{
```
philippsuess: `h`: Previously this was a blocking operation. The `client.close` calls flush https://github.com/getsentry/sentry-cocoa/blob/c5d37520bf6a94f44e792c99d140468598f0c300/Sources/Sentry/SentryClient.m#L640-L645

This makes it an async operation when not called on the main thread.

itaybre: You are right, I should make it sync, updating it

Issue #4040, opened 2024-06-06
SDK wrongly parses MetricKit MXHangDiagnostic stack trace


philipphofmann:
### Description

The SDK reports MXCPUException and MXDiskWriteException correctly, but the stacktraces for MXHangDiagnostic don't make any sense.

To find sample events, use this [Looker dashboard](https://sentryio.cloud.looker.com/dashboards/1400) to find customers using the MetricKitIntegration.

philipphofmann:
We finally got a sample event with a raw [MXDiagnosticPayload.json](https://github.com/user-attachments/files/17280627/MXDiagnosticPayload.json). Big thanks to @brustolin for sharing the data with us from his [personal project](https://brustolin.sentry.io/issues/5656146470/?project=5579037&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D%20mx&referrer=issue-stream&statsPeriod=30d&stream_index=0).

PR #6908, opened 2025-11-25
fix: Use correct parsing for stackframes

noahsmartin:
Fixes https://github.com/getsentry/sentry-cocoa/issues/4040

The problem was that for some events such as app hangs and CPUDiagnostics the metric kit payload contains an entire flamegraph. This change parses the flame graph into samples, and then picks the longest duration sample as the one we use for a stack trace. It also fixes over-counting the CPUDiagnostics which was emitting more than one event per exception. This is also fixed by identifying the longest duration sample.

Eventually we should support showing the full flame graph on the frontend, but with this PR the bug is fixed and we can consider that a future improvement.

I tested this on test flight by triggering crashes and verifying they are still correct, as well as manually replaying hang reports received in production

### DIFF
--- a/Sources/Swift/Core/MetricKit/MXSample.swift
+++ b/Sources/Swift/Core/MetricKit/MXSample.swift
@@ -11,13 +11,5 @@ struct MXSample {
         let offsetIntoBinaryTextSegment: Int
         let binaryName: String?
         let address: UInt64
-        
-        func toSentryFrame() -> Frame {
-            let frame = Frame()
-            frame.package = binaryName
-            frame.instructionAddress = sentry_formatHexAddressUInt64Swift(address)
-            frame.imageAddress = sentry_formatHexAddressUInt64Swift(address - UInt64(offsetIntoBinaryTextSegment))
-            return frame
-        }
     }
 }
--- a/Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
+++ b/Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
@@ -70,7 +70,7 @@ extension SentryMXFrame {
         result.type = "macho"
         result.debugID = binaryUUID.uuidString
         result.codeFile = binaryName
-        if offsetIntoBinaryTextSegment > 0 && offsetIntoBinaryTextSegment < address {
+        if offsetIntoBinaryTextSegment >= 0 && offsetIntoBinaryTextSegment < address {
             result.imageAddress = sentry_formatHexAddressUInt64Swift(address - UInt64(offsetIntoBinaryTextSegment))
         }
         return [result] + (subFrames?.flatMap { $0.toDebugMeta() } ?? [])
@@ -90,23 +90,26 @@ extension SentryMXFrame {
     }
 }
 
-extension Sequence {
-    func unique<T: Hashable>(by key: (Element) -> T) -> [Element] {
-        var seen = Set<T>()
-        var result: [Element] = []
-        for element in self {
-            let k = key(element)
-            if !seen.contains(k) {
-                seen.insert(k)
-                result.append(element)
-            }
+private extension MXSample.MXFrame {
+    func toSentryFrame() -> Frame {
+        let frame = Frame()
+        frame.package = binaryName
+        frame.instructionAddress = sentry_formatHexAddressUInt64Swift(address)
+        if offsetIntoBinaryTextSegment >= 0 && offsetIntoBinaryTextSegment < address {
+            frame.imageAddress = sentry_formatHexAddressUInt64Swift(address - UInt64(offsetIntoBinaryTextSegment))
         }
-
-        return result
+        return frame
     }
 }
 
-extension Dictionary where Value == Int {
+private extension Sequence {
+    func unique<T: Hashable>(by key: (Element) -> T) -> [Element] {
+        var seen = Set<T>()
+        return filter { seen.insert(key($0)).inserted }
+    }
+}
+
+private extension Dictionary where Value == Int {
     func mostSampled() -> Key? {
         var mostSamples = -1
         var mostSampledKey: Key?


Tests/SentryTests/Integrations/MetricKit/SentryMetricKitIntegrationTests.swift
```diff
@@ -260,19 +259,19 @@ final class SentryMetricKitIntegrationTests: SentrySDKIntegrationTestsBase {
...
-        XCTAssertEqual(4, client.captureEventWithScopeInvocations.count, "Client expected to capture 2 events.")
+        XCTAssertEqual(1, invocations.count, "Client expected to capture 1 event.")
         
-        let firstEvent = try XCTUnwrap(invocations.first).event
-        let secondEvent = try XCTUnwrap(invocations.element(at: 1)).event
-        let thirdEvent = try XCTUnwrap(invocations.element(at: 2)).event
-        let fourthEvent = try XCTUnwrap(invocations.element(at: 3)).event
+//        let firstEvent = try XCTUnwrap(invocations.first).event
+//        let secondEvent = try XCTUnwrap(invocations.element(at: 1)).event
+//        let thirdEvent = try XCTUnwrap(invocations.element(at: 2)).event
```
philippsuess: `m`: There is still some test code commented out. Please either fixe the test or remove them if they aren't relevant anymore.

Sources/Swift/Core/MetricKit/SentryMXCallStackTree.swift
```diff
@@ -7,39 +8,153 @@ import Foundation
...
+    // a thread backtrace. A flamegraph is just a collection of many thread backtraces
+    // generated by taking multiple samples. For example a hang from metric kit will
+    // be a group of samples of the main thread while it is hanging. To make MetricKit
+    // data compatible with Sentry data this function finds the most commonly sampled stack.
+    // Some metric kit events contain multiple threads (like a crash report) and others
+    // contain all threads in one flamegraph. That happens when "callStackPerThread"
+    // is false. In these cases we can't really make a "thread" because
+    // it represents data that is sampled across many threads and aggregated, we do not
+    // know which samples came from which thread. Instead we create just one fake thread
+    // that just contains the most common callstack.
```
philippsuess: `m`: Maybe worth adding info that we plan to change this by adding a flamegraph to app hang events. If we don't have GH issue, I think we could create one and link to the Android implementation and/or the frontend work that makes this possible.

Sources/Swift/Core/MetricKit/SentryMXCallStackTree.swift
```diff
@@ -7,39 +8,153 @@ import Foundation
...
+    func sentryMXBacktrace(inAppLogic: SentryInAppLogic?, handled: Bool) -> [SentryThread] {
+        callStacks.map { callStack in
+            let thread = SentryThread(threadId: 0)
```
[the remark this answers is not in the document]
philippsuess: `h`: Yes that's valid feedback. I think the backend and frontend might get confused when having multiple threads with the threadId 0. If this is intendended, we should add a comment explaining why we have 0 for all threads.

Sources/Swift/Core/MetricKit/SentryMXCallStackTree.swift
```diff
@@ -1,3 +1,4 @@
+@_implementationOnly import _SentryPrivate
```
philippsuess: `m`: I feels a bit strange to me now that we have a dependency on `_SentryPrivate` here. This is now a bit mixed. We have both MetricKitCallStack JSON parsing with the `JSONDecoder` and also `toDebugMeta` and `public func prepare(event: Event, inAppLogic: SentryInAppLogic?, handled: Bool) `.

I would prefer keeping this file focused on only parsing JSON data. I would move the conversion logic to a different file. It can be an extension, but doing this in a different file feels a bit cleaner to me.

Sources/Swift/Core/MetricKit/SentryMXCallStackTree.swift
```diff
@@ -7,39 +8,153 @@ import Foundation
...
+    }
 }
 
-@objcMembers
-@_spi(Private) public class SentryMXCallStack: NSObject, Decodable {
-    public let threadAttributed: Bool?
-    public let callStackRootFrames: [SentryMXFrame]
+// A Sample is the standard data format for a flamegraph taken from https://github.com/brendangregg/FlameGraph
+// It is less compact than Apple's MetricKit format, but contains the same data and is easier to work with
+struct Sample {
```
philippsuess: `m`: I think plenty of the structs and methods here could be fileprivate. We should hide that complexity and only make public what's required.

Tests/SentryTests/Integrations/MetricKit/SentryMetricKitIntegrationTests.swift
noahsmartin: Yep sorry I missed this

Sources/Swift/Core/MetricKit/SentryMXCallStackTree.swift
noahsmartin: Don't think any could be fileprivate, but only one is public now

noahsmartin:
Thanks @philipphofmann this is ready for another look!

Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
```diff
@@ -0,0 +1,123 @@
...
+    // data compatible with Sentry data this function finds the most commonly sampled stack.
+    // Some metric kit events contain multiple threads (like a crash report) and others
+    // contain all threads in one flamegraph. That happens when "callStackPerThread"
+    // is false. In these cases we can't really make a "thread" because
+    // it represents data that is sampled across many threads and aggregated, we do not
+    // know which samples came from which thread. Instead we create just one fake thread
+    // that just contains the most common callstack.
+    //
+    // We hope to add flamegraph support at some point and that is tracked here: https://github.com/getsentry/sentry-cocoa/issues/7062
+    func sentryMXBacktrace(inAppLogic: SentryInAppLogic?, handled: Bool) -> [SentryThread] {
```
philippsuess, line 40: Thank you 🥔

Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
```diff
@@ -0,0 +1,123 @@
...
+        let selfCount = (sampleCount ?? 0) - childCount
+        var result = subframes.flatMap { $0.toSamples() }.map { MXSample(count: $0.count, frames: [selfFrame] + $0.frames) }
+        if selfCount > 0 {
+            result.append(MXSample(count: selfCount, frames: [selfFrame]))
+        }
+        return result
+    }
+}
+
+extension Sequence {
```
philippsuess: `l`: this can be private, I think.
```suggestion
private extension Sequence {
```

Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
```diff
@@ -0,0 +1,123 @@
...
+                seen.insert(k)
+                result.append(element)
+            }
+        }
+
+        return result
+    }
+}
+
+extension Dictionary where Value == Int {
```
philippsuess: `l`:
```suggestion
private extension Dictionary where Value == Int {
```

Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
```diff
@@ -0,0 +1,123 @@
...
+        var result: [Element] = []
+        for element in self {
+            let k = key(element)
+            if !seen.contains(k) {
+                seen.insert(k)
+                result.append(element)
+            }
+        }
+
+        return result
```
philippsuess: `l`: I think this should be the same but a bit shorter
```suggestion
        var seen = Set<T>()
        return filter { seen.insert(key($0)).inserted }
```

Sources/Swift/Core/MetricKit/MXSample.swift
```diff
@@ -0,0 +1,23 @@
...
+        let binaryName: String?
+        let address: UInt64
+        
+        func toSentryFrame() -> Frame {
+            let frame = Frame()
+            frame.package = binaryName
+            frame.instructionAddress = sentry_formatHexAddressUInt64Swift(address)
+            frame.imageAddress = sentry_formatHexAddressUInt64Swift(address - UInt64(offsetIntoBinaryTextSegment))
+            return frame
+        }
```
philippsuess: `l`: All the other structs now don't have Sentry logic. Instead, we now have extensions for that Sentry sepcific logic in `SentryMXCallStackTree+Parsing` as extensions. We could also move this method to `SentryMXCallStackTree+Parsing`.

noahsmartin: Yeah this is not a decodable struct it's more of an intermediate model, I originally put the entire thing in that extension file, but we can also split this one up into two files

Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
noahsmartin: Yes much better, thanks!

Sources/Swift/Core/MetricKit/MXSample.swift
philippsuess: I'm fine with both. It's a nit

PR #6910, opened 2025-11-26
fix: allow transaction tags to be accessed and modified in beforeSend

### LINKED ISSUE #5308: Tags can not be edited in `beforeSend`
### Description

The tags of the tracer/transactions are not available in `beforeSend` but they are added to the serialized transaction. Example to reproduce:

```
SentrySDK.start { config in
    config.dsn = "https://b4cac8e75bb5fbd66350d2bd81015fac@o447951.ingest.us.sentry.io/4509348709924865"
    config.debug = true
    config.tracesSampleRate = 1.0
    config.enableUIViewControllerTracing = false
    config.enableAutoSessionTracking = false
    config.beforeSend = { transaction in
        print(transaction.tags) // prints "[:])"
        transaction.tags = ["key2": "value2"]
        return transaction
    }
    config.beforeSendSpan = { span in
        print(span.tags) // prints ["nested-key": "nested-value"]
        span.setTag(value: "nested-value2", key: "nested-key2")
        return span
    }
}

let tracer = SentrySDK.startTransaction(name: "test", operation: "test")
tracer.setTag(value: "value", key: "key")
let child = tracer.startChild(operation: "nested-1")
child.setTag(value: "nested-value", key: "nested-key")
sleep(1)
child.finish()
tracer.finish()
```

The example above sends a transaction with these tags:

```
"tags": {
  "key": "value",
  "key2": "value2"
},
```

and a span with these tags:

```
"tags": {
  "nested-key": "nested-value",
  "nested-key2": "nested-value2"
},
```

When changing the example to remove the existing tags, it removes the tag for the span, but not for the transaction:

```
config.beforeSend = { transaction in
    print(transaction.tags) // prints "[:])"
    transaction.tags = [:]
    return transaction
}
config.beforeSendSpan = { span in
    print(span.tags) // prints ["nested-key": "nested-value"]
    span.removeTag(key: "nested-key")
    return span
}
```

The tags of the trace are still:

```
"tags": {
    "key": "value"
},
```

philprime:
## Description

Fixes #5308

This PR implements getter and setter overrides for the `tags` property in `SentryTransaction` to allow full CRUD operations (Read, Add, Modify, Remove, Replace) on transaction tags within `beforeSend` callbacks.

**See [this issue comment](https://github.com/getsentry/sentry-cocoa/issues/5308#issuecomment-3582174265) for a full explanation reproducer.**

## Problem

When accessing `transaction.tags` in `beforeSend`, it only returned tags from the superclass `SentryEvent`, missing tags set on the underlying `SentryTracer`. Additionally, modifications to `transaction.tags` in `beforeSend` didn't persist because they only affected the event tags, not the tracer tags that are actually serialized.

## Solution Approach

We implemented a computed property pattern that:

1. **Getter**: Merges tags from both `SentryEvent` (superclass) and `SentryTracer` (wrapped tracer), with tracer tags taking precedence
   - Returns a mutable dictionary copy to allow modifications
   - Handles nil/empty cases gracefully

2. **Setter**: Replaces all tags by:
   - Clearing all existing tags from the tracer
   - Clearing event tags
   - Setting all new tags on the tracer (since transaction tags belong on the tracer)

### Swift vs Objective-C Behavior

**Swift**: When code does `transaction.tags?["key"] = "value"`, Swift automatically expands this to get tags, modify copy, and call the setter with the modified copy. This means modifications persist automatically.

**Objective-C**: Modifying the returned dictionary directly does NOT persist. Users must explicitly call the setter: `transaction.tags = modifiedDict`.

This difference is documented in code comments to help developers understand the behavior in both languages.

### Testing

Added comprehensive tests covering all CRUD operations:
- **Read**: Verify merged tags are returned correctly
- **Add**: Verify new tags can be added
- **Modify**: Verify existing tags can be modified (both tracer and event tags)
- **Remove**: Verify tags can be removed (both tracer and event tags)
- **Replace**: Verify entire tag dictionary can be replaced

All tests verify behavior both in `beforeSend` callbacks and in the final serialized output.

### 1. Custom NSMutableDictionary Subclass

**Approach**: Create a custom subclass that automatically calls the setter on every mutation.

**Why Not Implemented**:
- **Performance overhead**: The setter does significant work (clears all tags, then sets new ones). Calling it on every mutation would be inefficient, especially with multiple mutations.
- **Complexity**: Would need to override many mutation methods (`setObject:forKey:`, `removeObjectForKey:`, `removeAllObjects`, etc.)
- **Memory management**: Would require either weak references (risking nil) or strong references (creating retain cycles)
- **Thread safety**: Would need synchronization, adding more complexity
- **Architecture mismatch**: Tags belong on the tracer, not the event, so this approach doesn't align with the data model

### 2. Proxy Pattern with Method Swizzling

**Approach**: Use method swizzling or a proxy to intercept dictionary mutations.

**Why Not Implemented**:
- Too complex and fragile
- Method swizzling can have unintended side effects
- Doesn't solve the fundamental issue of where tags are stored

### 3. Direct Tracer Access

**Approach**: Document that users should access `transaction.trace.tags` directly.

**Why Not Implemented**:
- Breaks the abstraction - users shouldn't need to know about the internal tracer
- Inconsistent with how other properties work
- Doesn't solve the problem of reading merged tags

## Changes

- Added `tags` getter override in `SentryTransaction.m` that merges event and tracer tags
- Added `tags` setter override in `SentryTransaction.m` that replaces all tags on the tracer
- Added comprehensive tests in `SentryTransactionTests.swift` for all tag operations
- Added tests in `SentryClientTests.swift` verifying tag operations work in `beforeSend` callbacks
- Added documentation comments explaining Swift vs Objective-C behavior differences

## Testing

- All existing tests pass
- New comprehensive test suite covers all CRUD operations
- Verified behavior in both Swift and Objective-C contexts
- Tested with real `beforeSend` callbacks in sample apps

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -41,7 +41,7 @@ jobs:
     runs-on: macos-26
     steps:
       - uses: actions/checkout@v6
-      - run: ./scripts/ci-select-xcode.sh 26.1
+      - run: ./scripts/ci-select-xcode.sh 26.1.1
       - name: Setup Ruby
         uses: ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71 # v1.268.0
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -304,7 +304,7 @@ jobs:
           # macOS 26
           - name: macOS 26 Sentry
             runs-on: macos-26
-            xcode: "26.1"
+            xcode: "26.1.1"
             test-destination-os: "26.1"
             platform: "macOS"
             scheme: "Sentry"
@@ -361,7 +361,7 @@ jobs:
           # tvOS 26
           - name: tvOS 26 Sentry
             runs-on: macos-26
-            xcode: "26.1"
+            xcode: "26.1.1"
             test-destination-os: "26.1"
             install_platforms: true
             platform: "tvOS"


Sources/Sentry/SentryTransaction.m
```diff
@@ -22,6 +22,65 @@ - (instancetype)initWithTrace:(SentryTracer *)trace children:(NSArray<id<SentryS
...
+        // Fallback to superclass if trace is nil (shouldn't happen for valid transactions)
+        [super setTags:tags];
+        return;
+    }
+
+    // Remove all existing tags from the tracer
+    NSDictionary<NSString *, NSString *> *currentTracerTags = self.trace.tags;
+    for (NSString *key in currentTracerTags.allKeys) {
+        [self.trace removeTagForKey:key];
+    }
```
itaybre, line 68: `m:` Wouldn't it be better to add a `removeAllObjects` to `SentrySpanProtocol` and call that one?

philprime: IIUC this would be a public facing change as `SentrySpanProtocol.h` is a public header, therefore I wouldn't do that.

itaybre: Fair, I don't have a strong opinion on this.
But if we need it, there might be another used needing this too

philprime: But we only need it internally, so I would postpone this addition until it's requested by users.

PR #6917, opened 2025-11-27
chore: Add platform for iOS in Swift6 and other UI test

### LINKED ISSUE #6919: chore: Add platform for iOS in Swift6 and other UI test
> [!NOTE]
> The pull request "[chore: Add platform for iOS in Swift6 and other UI test](https://github.com/getsentry/sentry-cocoa/pull/6917)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Adds iOS platform field to enable step process which reloads runtimes

#skip-changelog

itaybre:
Adds iOS platform field to enable step process which reloads runtimes

#skip-changelog

Closes #6919

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -58,11 +58,13 @@ jobs:
             target: ios_objc
             test_destination_os: "18.6"
             device: iPhone 16 Pro
+            platform: iOS
           - job_id: tvos_swift
             name: tvOS Swift
             target: tvos_swift
             test_destination_os: "18.5"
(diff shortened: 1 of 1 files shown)


itaybre:
> I assume this is related to CirrusLabs runners being flaky, due to Xcode being flaky.
> 
> LGTM

Exactly, we use the platform field to identify when we need to run the step which verifies runtimes being correctly loaded.
This is because we only run that one for iOS platform.

We could use a new field to trigger that, but I prefer the simplification of using the common workflow and reduce the number of options

PR #6921, opened 2025-11-27
chore: Improve Carthage support with a json file for versions

itaybre:
When using Carthage to download xcframework files, the downloaded zip [is random](https://github.com/Carthage/Carthage/issues/3207).
This makes using Sentry with prebuilt binaries unreliable.

We can provide a json file per package to allow users to manually define which version they want to download, for example if they intend to use `Sentry-Dynamic` only, they just need to add this to their Cartfile
```
binary "https://raw.githubusercontent.com/getsentry/sentry-cocoa/refs/heads/main/Sentry-Dynamic.json"
```

i also updated `VersionBump` util to create the new entries on each release


itaybre:
Closing this since we are fully deprecating Carthage

PR #6922, opened 2025-11-28
fix: Convert bools to Swift in Options code docs

### LINKED ISSUE #6923: fix: Convert bools to Swift in Options code docs
> [!NOTE]
> The pull request "[fix: Convert bools to Swift in Options code docs](https://github.com/getsentry/sentry-cocoa/pull/6922)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

## :scroll: Description

The options frequently mentioned YES and NO, but this is Swift code now, and should be true and false.

philipphofmann:
## :scroll: Description

The options frequently mentioned YES and NO, but this is Swift code now, and should be true and false.

Closes #6923

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- Change the `http.response.status_code` for spans to int (#6896)
+
 ### Feature
 
 - Add options to customize UserFeedback error messages (#6790)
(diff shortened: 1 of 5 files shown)


philippsuess:
> Is there a benefit between `@c false` and `false` with backticks?

Good catch, no that's wrong actually. I'm going to fix this.

PR #6926, opened 2025-11-28
chore: Remove Carthage supporte from ReadMe

itaybre:
Since we are no longer supporting Carthage from v9, I am removing it from the Readme.md

Closes #6930

PR #6929, opened 2025-11-28
chore: Rename `CARTHAGE` macro to `XCODE`

itaybre:
In sync with #6926, since we no longer plan to use Carthage, I am renaming the precompiler flags to `XCODE`, to avoid confussions.

### DIFF
--- a/Tests/SentryTests/Protocol/SentrySdkInfoTests.swift
+++ b/Tests/SentryTests/Protocol/SentrySdkInfoTests.swift
@@ -108,18 +108,7 @@ class SentrySdkInfoTests: XCTestCase {
         XCTAssertEqual(packages[0]["version"] as? String, SentryMeta.versionString)
     }
 
-    func testCarthage_packageInfo() throws {
-        SentrySdkPackage.setPackageManager(2)
-        let actual = SentrySdkInfo.global()
-        let serialization = actual.serialize()
-
-        let packages = try XCTUnwrap(serialization["packages"] as? [[String: Any]])
-        XCTAssertEqual(1, packages.count)
-        XCTAssertEqual(packages[0]["name"] as? String, "carthage:getsentry/\(SentryMeta.sdkName)")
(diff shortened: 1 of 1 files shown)


scripts/build-xcframework-slice.sh
```diff
@@ -45,7 +45,6 @@ if [ "$sdk" = "maccatalyst" ]; then
         -derivedDataPath ./Carthage/DerivedData \
         CODE_SIGNING_REQUIRED=NO \
         CODE_SIGN_IDENTITY= \
-        CARTHAGE=YES \
```
itaybre, line 48: These are not being used anywhere, so I just removed them

Sources/Swift/Helper/SentrySdkPackage.swift
```diff
@@ -5,18 +5,13 @@ import Foundation
 private enum SentryPackageManagerOption: UInt {
     case swiftPackageManager = 0
     case cocoaPods = 1
-    case carthage = 2
-    case unknown = 3
+    case unknown = 2
```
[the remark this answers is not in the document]
itaybre: Fixed

Sources/SentrySwiftUI/Preview/SentryReplayMaskPreview.swift
```diff
@@ -3,7 +3,7 @@ import Sentry
 import SwiftUI
 import UIKit
 
-#if CARTHAGE || SWIFT_PACKAGE
+#if XCODE || SWIFT_PACKAGE
```
philippsuess, line 6: `m`: But wait a second. We now always have the `DXCODE` swift flag set. This will always be true. So does this flag even make sense? We can also follow up with a PR.

itaybre: I believe this is false for cocoapods

PR #6931, opened 2025-11-28
chore: Rename Carthage validation project to XCFramework validation project

itaybre:
Renames the Carthage validation project which uses xcframeworks to just XCFramework Validation project

Cleanups some script that wont be used anymore too.

Follows #6926 in removing Carthage references.


Brewfile-ci-deploy
```diff
@@ -1 +0,0 @@
-brew 'carthage'
```
itaybre, line 1: We should be able to delete this file now

Samples/XCFramework-Validation/XCFramework.xcodeproj/project.pbxproj
```diff
@@ -36,7 +36,7 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
-		1573CAE22BBCA4820018A2B7 /* SentrySwiftUI.xcframework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.xcframework; name = SentrySwiftUI.xcframework; path = Carthage/Build/SentrySwiftUI.xcframework; sourceTree = "<group>"; };
+		1573CAE22BBCA4820018A2B7 /* SentrySwiftUI.xcframework */ = {isa = PBXFileReference; expectedSignature = "AppleDeveloperProgram:97JCY7859U:GetSentry LLC"; lastKnownFileType = wrapper.xcframework; name = SentrySwiftUI.xcframework; path = ../../Carthage/SentrySwiftUI.xcframework; sourceTree = "<group>"; };
```
itaybre, line 39: Mmmm, I wonder if having the expected signature may bite us in CI

philippsuess: Could be. I would trigger an alpha release from this branch to see if the relase workflow breaks.

.github/workflows/release.yml
```diff
@@ -127,6 +127,9 @@ jobs:
           pattern: xcframework-${{github.sha}}-sentry-swiftui
           path: Carthage/
       - run: ./scripts/ci-select-xcode.sh 16.4
+      - name: Unzip all XCFrameworks
+        run: |
+          find Carthage -name "*.xcframework.zip" -print0 | xargs -t0I @ unzip @ -d Carthage
```
philippsuess, line 132: `m`: It feels a bit weird that we still have `Carthage` here. We don't have to fix it in this PR, but maybe later.

itaybre: Yes, that's the build path I mentioned earlier.
I want to change that one, but in a different PR, since that path is used in many places.

Samples/XCFramework-Validation/XCFramework.xcodeproj/project.pbxproj
itaybre: Confirmed no issues here: https://github.com/getsentry/sentry-cocoa/actions/runs/19833508140/job/56825670966

PR #6939, opened 2025-12-01
feat: Add isiOSAppOnVisionOS, isiOSAppOnMac, isMacCatalystApp to device context

### LINKED ISSUE #3824: Identify when iOS app runs on VisionOS
### Description

It's useful for users to know if their iOS app runs on VisionOS. As of April 4th, 2024, there is no official API for this yet; see [Apple forum](https://developer.apple.com/forums/thread/733029) and [Stackoverflow](https://stackoverflow.com/questions/76760180/how-to-check-if-device-is-vision-pro-xcode-15-beta-4). Once Apple adds an official API similar to `NSProcessInfo.iOSAppOnMac` we can send this information on the device context.

### LINKED ISSUE #3825: Identify when iOS app runs on a Mac
### Description

Use [NSProcessInfo.iOSApoOnMac](https://developer.apple.com/documentation/foundation/nsprocessinfo/3608556-iosapponmac?language=objc) and [NSProcessInfo.macCatalystApp](https://developer.apple.com/documentation/foundation/nsprocessinfo/3362531-maccatalystapp?language=objc) when an iOS app runs on a Mac and send this information along with the device context. Maybe users can already determine if an iOS app runs on a Mac so we have to double check before implementing this.

This issue is related to https://github.com/getsentry/sentry-cocoa/issues/3824.

philprime:
## :scroll: Description

- Adds check for `isiOSAppOnVisionOS` supporting all visionOS versions
- Adds isiOSAppOnVisionOS, isiOSAppOnMac, isMacCatalystApp to device context

Closes #3824
Closes #3825 

### visionOS

- Added "Apple Vision (Designed for iPad)" to the list of supported destinations for the sample app iOS-Swift
- Ran the scheme `iOS-Swift`
- Tapped on "Capture Error" to send an error event
- Checked the event context in Sentry if `ios_app_on_visionos`, `ios_app_on_macos` and `mac_catalyst_app` are set

| OS Version    | Event | `ios_app_on_visionos` | `ios_app_on_macos` | `mac_catalyst_app` |
|---------------|-------|-----------------------|--------------------|--------------------|
| visionOS 26.1 | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/4bbfe10fd6184586a5825cd9b360359d/) | ✅ | ❌ | ❌ |
| visionOS 26.0 | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/7ac5ae7f5fb74c9c8e2107c10c260eb8/) | ✅ | ❌ | ❌ |
| visionOS 2.5  | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/867e56f7559f44b4bd2c870999c653f8/) | ✅ | ❌ | ❌ |
| visionOS 2.4  | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/c50de56c12f640e0b888f6df9729133d/) | ✅ | ❌ | ❌ |
| visionOS 2.3  | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/0c8f82d80d934a50bda9d4e5111317c2/) | ✅ | ❌ | ❌ |
| visionOS 2.2  | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/18d6639da29344e197b4316994a6ca46/) | ✅ | ❌ | ❌ |
| visionOS 2.1  | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/07c5f1a6efb245c6b8efc45dc93bcb96/) | ✅ | ❌ | ❌ |
| visionOS 2.0  | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/413cca4065fe41a0a87b5adcb8214a3f/) | ✅ | ❌ | ❌ |
| visionOS 1.2  | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/6ec31d980ee240cda68fb5e096913a34/) | ✅ | ❌ | ❌ |
| visionOS 1.1  | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/6c89eebdd0854f65abc2d79e240843b5/) | ✅ | ❌ | ❌ |
| visionOS 1.0  | [Link](https://sentry-sdks.sentry.io/issues/6607086585/events/23e341f91cb64faea03b399a6481c132/) | ✅  | ❌ | ❌ |

**Note:** During testing I encountered an `unrecognized selector` SDK crash on visionOS 1.1, even though it was gated with an availability check. I fixed it by adding the recognized selector check.

### Unit Tests

In addition I added a unit test which asserts that the flag `isiOSAppOnVisionOS` is default `false`.

## :pencil: Checklist

### DIFF
--- a/.github/workflows/test-cross-platform.yml
+++ b/.github/workflows/test-cross-platform.yml
@@ -86,7 +86,7 @@ jobs:
         run: yarn install
 
       - name: Remove fixed version of Sentry Dependency from RNSentry.podspec
-        run: sed -i '' "s/s.dependency 'Sentry', '[0-9]*\.[0-9]*\.[0-9]*'/s.dependency 'Sentry'/" sentry-react-native/packages/core/RNSentry.podspec
+        run: sed -E -i '' "s/s.dependency 'Sentry', '[0-9]*\.[0-9]*\.[0-9]*(-[a-z]+\.[0-9]+)?'/s.dependency 'Sentry'/" sentry-react-native/packages/core/RNSentry.podspec
 
       - name: Add Sentry Dependency with path to Sentry Cocoa SDK to RNSentryCocoaTester/Podfile
         run: sed -i '' -e "s/RNSentry.podspec'/RNSentry.podspec'\\n  pod 'Sentry', :path => '..\/..\/..\/..\/sentry-cocoa'/" sentry-react-native/packages/core/RNSentryCocoaTester/Podfile


itaybre:
@philprime are you waiting on something in particular for this?

Sources/Swift/Helper/SentryProcessInfo.swift
```diff
@@ -26,4 +28,26 @@
...
+                // Use value(forKey:) to dynamically access the property at runtime
+                // This avoids compile-time errors when the API isn't available in the SDK headers of Xcode 16 or older.
+                if let value = self.value(forKey: "isiOSAppOnVision") as? Bool {
+                    return value
+                }
+            }
+        }
+        // Fallback for older versions: `UIWindowSceneGeometryPreferencesVision` is only available on visionOS
+        // https://developer.apple.com/documentation/uikit/uiwindowscene/geometrypreferences/vision?language=objc
+        return NSClassFromString("UIWindowSceneGeometryPreferencesVision") != nil
```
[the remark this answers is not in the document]
philippsuess: `m`: Yes, I think that's a legit comment. The property is called `isiOSAppOnVisionOS`, so this must only return true if it's an iOS app running on visionOS. I think it's OK to skip the fallback. It's better to only report the data when we know it's accurate and not report it for older versions for which we can't provide a reliable solution. If we decide to do that, then we could use `@available` for the property to clearly communicate that this only works for iOS and visionOS.

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -170,6 +170,12 @@ public final class SentryCrashWrapper: NSObject {
         
         deviceData["locale"] = Locale.autoupdatingCurrent.identifier
         
+        if #available(macOS 12, *) {
+            deviceData["ios_app_on_macos"] = self.processInfoWrapper.isiOSAppOnMac
+            deviceData["mac_catalyst_app"] = self.processInfoWrapper.isMacCatalystApp
+        } 
+        deviceData["ios_app_on_visionos"] = self.processInfoWrapper.isiOSAppOnVisionOS
```
philippsuess: `m`: These are going to be `false` for almost all events. What about only sending these when they are `true`. For most events, that doesn't add any value and it increases the payload size.

Sources/Swift/Helper/SentryProcessInfo.swift
```diff
@@ -26,4 +28,26 @@
...
+            // For unknown reasons when running an iOS app "Designed for iPad" on visionOS 1.1, the simulator system
+            // version is simulator 17.4, but it still enters this block.
+            //
+            // Due to that it crashes with an uncaught exception 'NSInvalidArgumentException', reason: '-[NSProcessInfo isiOSAppOnVision]: unrecognized selector sent to instance 0x600001549230'
+            if self.responds(to: NSSelectorFromString("isiOSAppOnVision")) {
+                // Use value(forKey:) to dynamically access the property at runtime
+                // This avoids compile-time errors when the API isn't available in the SDK headers of Xcode 16 or older.
+                if let value = self.value(forKey: "isiOSAppOnVision") as? Bool {
+                    return value
+                }
```
philippsuess, line 49: Thanks for the comments to clearly explain why we need this solution. It still seems a bit dirty, but I think it's acceptable. I don't know any other way to make this work when compiling the XCFramework with Xcode 16. We could create an issue to follow up on to remove this workaround once we use Xcode 26.

philprime: I think this issue will come back when people start using it via SPM and compile it locally using Xcode 16. I had a similar issue recently in when working with SFSafeSymbols: https://github.com/SFSafeSymbols/SFSafeSymbols/issues/148

So as long as we support compiling with Xcode 16 in general we need to keep this in.

Sources/Swift/Helper/SentryProcessInfo.swift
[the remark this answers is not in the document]
philprime: What do you think about this:

```
        #if os(iOS)
        // Fallback for older versions: `UIWindowSceneGeometryPreferencesVision` is only available on visionOS
        // https://developer.apple.com/documentation/uikit/uiwindowscene/geometrypreferences/vision?language=objc
        return NSClassFromString("UIWindowSceneGeometryPreferencesVision") != nil
        #else
        return false
        #endif
```

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
philprime: Do we have a common guideline for sending boolean flags which are mostly one value?

Sources/Swift/Helper/SentryProcessInfo.swift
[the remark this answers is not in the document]
philprime: I could also do an early exit if we are not on iOS:

```
#if !os(iOS)
return false
#else
...
#endif

File: Sources/Swift/Helper/SentryProcessInfo.swift (hunk 3 above)
Comment (philippsuess) [reply Sources/Swift/Helper/SentryProcessInfo.swift:49]: Then maybe update the comment that we can't remove this when we compile the SDK with Xcode 26, and we need to keep this as long as people use Xcode 16.

File: Sources/Swift/SentryCrash/SentryCrashWrapper.swift (hunk 2 above)
Comment (philippsuess) [reply Sources/Swift/SentryCrash/SentryCrashWrapper.swift]: Not real guidelines, no, but IMO we should always try to keep the payload small, if we can, for everything we do.

File: Sources/Swift/Helper/SentryProcessInfo.swift [hunk 4]
```diff
@@ -26,4 +28,35 @@
...
+        }
+        // Fallback for older visionOS versions (< 26.1) when the official API isn't available.
+        // The `#if os(iOS)` check is necessary because `UIWindowSceneGeometryPreferencesVision` exists on all
+        // visionOS builds, so without this compile-time check, native visionOS apps would incorrectly return `true`.
+        #if os(iOS)
+        // https://developer.apple.com/documentation/uikit/uiwindowscene/geometrypreferences/vision?language=objc
+        return NSClassFromString("UIWindowSceneGeometryPreferencesVision") != nil
+        #else
+        return false
+        #endif
```
philippsuess, line 60: Yes that's better thanks.

Sources/Swift/SentryCrash/SentryCrashWrapper.swift
```diff
@@ -170,6 +170,21 @@ public final class SentryCrashWrapper: NSObject {
         
         deviceData["locale"] = Locale.autoupdatingCurrent.identifier
         
+        // Only include these boolean flags when they are `true` to reduce payload size.
+        // These flags are `false` for the vast majority of events (regular iOS apps, native macOS apps, etc.),
+        // so omitting them when `false` significantly reduces the payload size without losing meaningful information.
```
philippsuess, line 175: Thanks for the context.

PR #6942, opened 2025-12-01
ref: Convert SentryDsn to swift

antonis:
## :scroll: Description

Converts SentryDsn to swift

See https://github.com/getsentry/sentry-react-native/issues/5273#issuecomment-3533215421

Ci checks, Sample apps

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 > The class `SentryDsn` has been converted from Objective-C to Swift. While the public API remains the same, you might have to remove imports of the `SentryDsn.h` from your Objective-C code.
 > We acknowledge that this is a breaking change introduced in a minor release, but it should have been part of the major release 9.0.0 already before - apologies for any inconveniences caused.
 
-## Breaking Changes
+### Breaking Changes
 
 - Refactored type `SentryDsn` from Objective-C to Swift, removing the `SentryDsn.h` from the public header files.
 


Sources/Swift/SentryDsn.swift
```diff
@@ -0,0 +1,189 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+#if canImport(CommonCrypto)
+import CommonCrypto
+#endif
+
+/// Represents a Sentry Data Source Name (DSN) which identifies a Sentry project.
+@objc(SentryDsn)
+public class SentryDsn: NSObject {
```
noahsmartin: `final`?

Sources/Swift/SentryDsn.swift
```diff
@@ -0,0 +1,189 @@
...
+    /// Internal initializer for use by the throwing convenience initializer.
+    private init(url: URL) {
+        self.url = url
+        super.init()
+    }
+    
+    /// Generates a SHA1 hash of the DSN URL.
+    /// - Returns: A hexadecimal string representation of the hash.
+    @objc
+    public func getHash() -> String {
```
noahsmartin, line 66: IIRC this function isn't used, maybe we can delete it?

antonis:
@sentry review

Sources/Swift/SentryDsn.swift
```diff
@@ -0,0 +1,189 @@
...
+#if canImport(CommonCrypto)
+import CommonCrypto
+#endif
+
+/// Represents a Sentry Data Source Name (DSN) which identifies a Sentry project.
+@objc(SentryDsn)
+public class SentryDsn: NSObject {
+    
+    // Error constants - must match SentryError.h
+    private static let kSentryErrorInvalidDsnError: Int = 100
```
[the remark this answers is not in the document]
antonis: Updated with 450436b222d2d8ec9801b606ed825d033497af3a

Sources/Swift/SentryDsn.swift
```diff
@@ -0,0 +1,189 @@
...
+    /// - Returns: A hexadecimal string representation of the hash.
+    @objc
+    public func getHash() -> String {
+        guard let data = url.absoluteString.data(using: .utf8) else {
+            return ""
+        }
+        
+        var digest = [UInt8](repeating: 0, count: Int(CC_SHA1_DIGEST_LENGTH))
+        data.withUnsafeBytes { bytes in
+            _ = CC_SHA1(bytes.baseAddress, CC_LONG(data.count), &digest)
```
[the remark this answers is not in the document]
antonis: Updated with a73aa554f477806fa3ae5fca6ac0618e0e6e644a

Sources/Swift/SentryDsn.swift
antonis: I think it is called from objective-c [here](https://github.com/getsentry/sentry-cocoa/blob/98e0ee854b026d0dedda6bd85bab32b9bdaa1d61/Sources/Sentry/SentryFileManagerHelper.m#L148)

Sources/Swift/SentryDsn.swift
antonis: Good point 👍 Updated with 035a80c873777fd55ac1edad8620323f2cf73895

sdk_api.json
antonis, line 1: I've run `make generate-public-api ` with `xcode 16.4`

Sources/Swift/SentryDsn.swift
```diff
@@ -0,0 +1,202 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+#if canImport(CommonCrypto)
+import CommonCrypto
```
philprime, line 4: `h`: What is used if CommonCrypto is absent? The logic still relies on types from this framework so we need to have a proper fallback or gate all logic using CC* types

antonis: Good catch @philprime 👍 
My understanding is that CommonCrypto should always be available on Apple platforms. I'm torn between removing the conditional import (we were just importing in the ObjC code) or add a fallback to be on the safe side (e.g. [CryptoKit](https://developer.apple.com/documentation/cryptokit/)). Wdyt?

ref: https://developer.apple.com/security/

antonis: Added a guard with 9eb91593ded0bfaedc69987c525202d82ad359e6

philprime: I would just remove the canImport, we already used it in Objective-C so I don't see a reason why it would become unavailable now.

antonis: Sounds good 👍 Updated with ac7ef0199a9ac803f549b2af6f858a95095ed373

philprime:
@philipphofmann @itaybre I had a quick discussion on this with @antonis and @noahsmartin in person and we believe this is a breaking change because everyone importing `SentryDsn.h` will not find the header anymore and manually change their code base. We just shipped the v9 major release so I think going for a v10 because it's a breaking change is overkill and we should just quickly ship v9.0.1, and consider it as a change we missed for the major release.

Do you agree with this approach?

Sources/Swift/SentryDsn.swift
```diff
@@ -0,0 +1,200 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+import CommonCrypto
```
noahsmartin, line 4: It would be nice to replace this with `CryptoKit` and use `Insecure.SHA1.hash(data:)` rather than the C functions that aren't very Swift friendly. But not a blocker for this PR, we can follow up with that later

noahsmartin: https://developer.apple.com/documentation/cryptokit/insecure/sha1

antonis: Sounds good 👍 I've actually tested it out earlier ([related discussion](https://github.com/getsentry/sentry-cocoa/pull/6942#discussion_r2585706912)).
I'll follow up with a separate PR on this.

antonis: @noahsmartin heads up that I've tried to use `CryptoKit` but I've run to a deployment target mismatch since [the project targets macOS `10.14`](https://github.com/getsentry/sentry-cocoa/blob/3b189e189e160e86d4eeb863a545fa359a2d4d44/Sources/Configuration/DeploymentTargets.xcconfig#L3) and [`Insecure.SHA1` from CryptoKit requires macOS `10.15+`](https://developer.apple.com/documentation/cryptokit/insecure/sha1). The actual code change is minimal (see patch below) but since this PR [is already a "breaking change"](https://github.com/getsentry/sentry-cocoa/pull/6942#issuecomment-3615860854) also bumping the macos deployment target might not be desired.

<details><summary>Related CryptoKit patch</summary>
<p>

```
diff --git a/Sources/Swift/SentryDsn.swift b/Sources/Swift/SentryDsn.swift
index d12ee3b3a..0722bf543 100644
--- a/Sources/Swift/SentryDsn.swift
+++ b/Sources/Swift/SentryDsn.swift
@@ -1,8 +1,7 @@
+import CryptoKit
 import Foundation
 
-import CommonCrypto
-
 /// Represents a Sentry Data Source Name (DSN) which identifies a Sentry project.
 @objc(SentryDsn)
 public final class SentryDsn: NSObject {
@@ -78,12 +77,8 @@ public final class SentryDsn: NSObject {
             data = Data()
         }
         
-        var digest = [UInt8](repeating: 0, count: Int(CC_SHA1_DIGEST_LENGTH))
-        data.withUnsafeBytes { bytes in
-            _ = CC_SHA1(bytes.baseAddress, CC_LONG(data.count), &digest)
-        }
-        
-        return digest.map { String(format: "%02x", $0) }.joined()
+        let hash = Insecure.SHA1.hash(data: data)
+        return hash.map { String(format: "%02x", $0) }.joined()
     }
     
     /// Returns the envelope endpoint URL for this DSN.
diff --git a/Tests/SentryTests/Networking/SentryDsnTests.m b/Tests/SentryTests/Networking/SentryDsnTests.m
index 00636569c..e09318541 100644
--- a/Tests/SentryTests/Networking/SentryDsnTests.m
+++ b/Tests/SentryTests/Networking/SentryDsnTests.m
@@ -81,4 +81,46 @@
     XCTAssertTrue([dsn getEnvelopeEndpoint] == [dsn getEnvelopeEndpoint]);
 }
 
+- (void)testGetHash_ReturnsSHA1Hash
+{
+    SentryDsn *dsn = [[SentryDsn alloc] initWithString:@"https://username:password@getsentry.net/1"
+                                      didFailWithError:nil];
+
+    NSString *hash = [dsn getHash];
+
+    XCTAssertNotNil(hash);
+    XCTAssertEqual(hash.length, 40, @"SHA1 hash should be 40 characters");
+
+    NSCharacterSet *hexCharacterSet =
+        [NSCharacterSet characterSetWithCharactersInString:@"0123456789abcdef"];
+    NSCharacterSet *hashCharacterSet = [NSCharacterSet characterSetWithCharactersInString:hash];
+    XCTAssertTrue([hexCharacterSet isSupersetOfSet:hashCharacterSet],
+        @"Hash should only contain hexadecimal characters");
+}
+
+- (void)testGetHash_IsConsistent
+{
+    NSString *dsnString = @"https://username:password@getsentry.net/1";
+    SentryDsn *dsn1 = [[SentryDsn alloc] initWithString:dsnString didFailWithError:nil];
+    SentryDsn *dsn2 = [[SentryDsn alloc] initWithString:dsnString didFailWithError:nil];
+
+    NSString *hash1 = [dsn1 getHash];
+    NSString *hash2 = [dsn2 getHash];
+
+    XCTAssertEqualObjects(hash1, hash2, @"Same DSN should produce the same hash");
+}
+
+- (void)testGetHash_DifferentDsnProducesDifferentHash
+{
+    SentryDsn *dsn1 = [[SentryDsn alloc] initWithString:@"https://user1:pass1@getsentry.net/1"
+                                       didFailWithError:nil];
+    SentryDsn *dsn2 = [[SentryDsn alloc] initWithString:@"https://user2:pass2@getsentry.net/2"
+                                       didFailWithError:nil];
+
+    NSString *hash1 = [dsn1 getHash];
+    NSString *hash2 = [dsn2 getHash];
+
+    XCTAssertNotEqualObjects(hash1, hash2, @"Different DSNs should produce different hashes");
+}
+
 @end
```
</p>
</details>

philprime: @antonis can we stay with the current approach using CommonCrypto to get this PR merged now, and do this in a follow-up PR? This issue is currently blocking the release 9.1.0 so I think we are good to go.

philprime:
It was decided today in the Cocoa Sync with @itaybre @philipphofmann and myself that we will release this breaking change in a minor release, as the impact is only going to Objective-C users having to remove the `#import` statements

philprime:
Actually, @antonis can you please add a note to the CHANGELOG about this change with a warning callout at the top of the unreleased section so users adopting the next version are aware we had to do a breaking change?

Sources/Swift/SentryDsn.swift
antonis: I agree @philprime 👍

antonis:
> Actually, @antonis can you please add a note to the CHANGELOG about this change with a warning callout at the top of the unreleased section so users adopting the next version are aware we had to do a breaking change?

Good idea @philprime! I've added a warning with 1764341be7dae80631445372730d7d01156bbbad
Feel free to proceed with approving/merging if you consider this sufficient

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+> [!Warning]
+> **`SentryDsn` converted from Objective-C to Swift**
+> The `SentryDsn` class has been converted from Objective-C to Swift. The public API remains the same but this may affect Objective-C code that directly interacts with `SentryDsn`.
```
```suggestion
> [!Warning]
> The class `SentryDsn` has been converted from Objective-C to Swift. While the public API remains the same, you might have to remove imports of the `SentryDsn.h` from your Objective-C code. 
We acknowledge that this is a breaking change introduced in a minor release, but it should have been part of the major release 9.0.0 already before - apologies for any inconveniences caused.
```

philprime:
Thank you @antonis for the update. I'll merge this now so we can proceed with a release.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+> [!Warning]
+> The class `SentryDsn` has been converted from Objective-C to Swift. While the public API remains the same, you might have to remove imports of the `SentryDsn.h` from your Objective-C code. 
+We acknowledge that this is a breaking change introduced in a minor release, but it should have been part of the major release 9.0.0 already before - apologies for any inconveniences caused.
+
```
philprime, line 8:
```suggestion

## Breaking Changes

- Refactored type `SentryDsn` from Objective-C to Swift, removing the `SentryDsn.h` from the public header files.

```

Sources/Swift/SentryDsn.swift
antonis: Opened a PR to wrap up this discussion https://github.com/getsentry/sentry-cocoa/pull/7037

PR #6945, opened 2025-12-01
chore: Add dedicated CI workflow for 3rd-party integrations

### LINKED ISSUE #6973: chore: Add test workflow for swift log
> \[!NOTE\]
> The pull request "[chore: Add test workflow for swift log](<https://github.com/getsentry/sentry-cocoa/pull/6945>)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Adds:

* Adds a common workflow for 3rd party integrations
* Adds a script to easy testing SPM packages
* Adds a workflow for SwiftLog

itaybre:
#skip-changelog

## Add dedicated CI workflow for 3rd-party integrations

Moves 3rd-party integration tests (SentrySwiftLog, SentrySwiftyBeaver, SentryCocoaLumberjack, SentryPulse) from `test.yml` into a dedicated `test-3rd-party-integrations.yml` workflow with its own file-filter trigger.

- Pre-builds XCFrameworks from source once and shares them across all integration test jobs (instead of using swift package edit or the latest release)
- Each integration is tested via swift test on macOS.

Closes #6973


.github/workflows/integrations-common.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Build Tests
+        working-directory: ${{ inputs.working_directory }}
+        env:
+          PLATFORM: ${{ inputs.platform }}
+          OS_VERSION: ${{ inputs.test-destination-os }}
+          REF_NAME: ${{ github.ref_name }}
+          DEVICE_NAME: ${{ inputs.device }}
+          SCHEME: ${{ inputs.package-name }}
+        run: |
+          ../../../scripts/integration-xcodebuild.sh \
```
philippsuess: `m`: Applies to all script invocations.
```suggestion
          ./scripts/integration-xcodebuild.sh \
```

scripts/integration-xcodebuild.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+set -euxo pipefail
+
+# Disable SC1091 because it won't work with pre-commit
```
philippsuess: `m`: This script is quite similar to `sentry-xcodebuild.sh`. I would really appreciate if we don't have all of that logic duplicated.

.github/workflows/integrations-common.yml
```diff
@@ -0,0 +1,105 @@
...
+        description: "The package name to test (same as Package.swift name)"
+        required: true
+        type: string
+      macos_version:
+        description: "macOS version"
+        required: true
+        default: macos-26
+        type: string
+
+jobs:
```
philippsuess: `m`: There are many similarities to `unit-test-common.yml`. Can't we reuse `unit-test-common.yml` instead of having plenty of code duplicated?

scripts/use-local-sentry-dependency.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+set -euo pipefail
+
+# Script to replace sentry-cocoa remote dependency with local path in integration packages
+# This is useful for CI testing where we want to test integrations against the current code
```
philippsuess: `m`: I think this could be part of https://github.com/getsentry/sentry-cocoa/blob/main/scripts/prepare-package.sh

philipphofmann (commented):
Thanks, I added a few comments.

.github/workflows/integrations-common.yml
itaybre: Because of the `working-directory` it doesn't

scripts/use-local-sentry-dependency.sh
itaybre: I can move if there, but the only similarity is that both scripts manipulate Package.swift 
But they are not so similar in which modifications they do

scripts/prepare-package.sh
```diff
@@ -131,6 +139,10 @@ var targets: [Target] = [\
   sed -i '' '/^let env = getenv("EXPERIMENTAL_SPM_BUILDS")/,/^}/d' "$PACKAGE_FILE"
 fi
 
+if is_enabled "$UPDATE_PATH_TO_SENTRY_COCOA"; then
+  sed -i '' 's|\.package(url: "https://github\.com/getsentry/sentry-cocoa", from: "[^"]*")|.package(path: "'../../../'")|g' "$PACKAGE_FILE"
+fi
```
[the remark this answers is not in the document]
denrase: We could use SPM functionality instead of replacing strings. 

```bash
swift package edit sentry-cocoa --path ../..
```

denrase:
@philipphofmann @itaybre Updated the workflow to run tests for all 3rd party integrations with SPM and Xcodebuild on different platforms/devices.

Bildschirmfoto 2026-01-20 um 17 31 12

.github/file-filters.yml
```diff
@@ -527,3 +527,10 @@ run_size_analysis_for_prs: &run_size_analysis_for_prs
   # Build configuration
   - "Makefile"
   - "Brewfile*"
+
+run_3rd_party_integrations_tests_for_prs: &run_3rd_party_integrations_tests_for_prs
+  - "3rd-party-integrations/**"
```
itaybre, line 533: `h`: Any change to the source code of sentry-cocoa may affect the integrations, so the tests should run for them too

.github/workflows/unit-test-common.yml
```diff
@@ -65,51 +83,94 @@ jobs:
...
+      # Install platform runtimes that are not included by default
+      - name: Install required platforms
+        working-directory: ${{ inputs.working_directory }}
+        if: ${{ inputs.install_platforms }}
+        env:
+          PLATFORMS: ${{ inputs.platform }}
+          OS_VERSION: ${{ inputs.test-destination-os }}
+          SCRIPTS_DIRECTORY: ${{ inputs.scripts_directory }}
+        run: |
+          "$SCRIPTS_DIRECTORY/ci-install-platforms.sh" --platforms "$PLATFORMS" --os-version "$OS_VERSION"
```
itaybre: `l`: Are we missing platforms

.github/workflows/test-3rd-party-integrations.yml
```diff
@@ -0,0 +1,284 @@
...
+      platform: ${{ matrix.platform }}
+      runs-on: ${{ matrix.macos_version }}
+      timeout: 10
+      test_spm_package: true
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # SwiftLog tests
+          - integration_dir: SentrySwiftLog
```
itaybre: `m`: At this moment I don't see the need to test so many platforms as I don't see any any reason for a Log API to be different between platforms or OS versions

.github/workflows/test-3rd-party-integrations.yml
```diff
@@ -0,0 +1,284 @@
+name: Test 3rd Party Integrations
```
itaybre, line 1: `h`: Something in noticed after fixing #7240, is that this tests use the latest xcarchive released in GitHub, not the source code.
We should add a step to build the xcframework and then modify sentry-cocoa's `Package.swift` to use it instead of the URL

denrase: Is this something that should be done for other tests (or the release flow) as well? I'm worried that this build step will take a long time, as these tests also run on PRs. WDYT?

.github/workflows/test-3rd-party-integrations.yml
denrase: Only running on iOS 18.5 with Xcode 16.4 now.

.github/workflows/test-3rd-party-integrations.yml
denrase: Updated it to build a macos/iphonesim slice with Sentry target only (removing the dynamic targets).

.github/workflows/unit-test-common.yml
```diff
@@ -65,51 +83,94 @@ jobs:
...
+          DEVICE_NAME: ${{inputs.device}}
+          SCRIPTS_DIRECTORY: ${{ inputs.scripts_directory }}
+        run: |
+          "$SCRIPTS_DIRECTORY/ci-create-simulator.sh" --platform "$PLATFORM" --os-version "$OS_VERSION" --device-name "$DEVICE_NAME"
 
       # Boot created simulators to ensure they're ready before tests run
       # Based on CircleCI forum comment, booting is especially important for Xcode 26: https://discuss.circleci.com/t/xcode-26-rc/54066/18
       - name: Boot simulator
-        if: ${{ inputs.platform == 'iOS' || inputs.platform == 'visionOS' }}
+        if: ${{ (inputs.create_device && inputs.platform == 'iOS') || inputs.platform == 'visionOS' }}
```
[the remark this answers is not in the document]
denrase: @itaybre This seems to have been an issue before as well, can we safely remove this?

.github/workflows/unit-test-common.yml
denrase: Is this still relevant?

.github/workflows/unit-test-common.yml
```diff
@@ -65,51 +83,94 @@ jobs:
...
+        run: |
+          "$SCRIPTS_DIRECTORY/ci-select-xcode.sh" "$XCODE_VERSION"
 
       - name: Install Slather
         run: gem install slather
 
+      # Install platform runtimes that are not included by default
+      - name: Install required platforms
+        working-directory: ${{ inputs.working_directory }}
+        if: ${{ inputs.install_platforms }}
```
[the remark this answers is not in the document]
denrase: @itaybre Same thing here, there is no input for `install_platforms`

denrase:
@itay Updated the PR:

- Updated the workflow to build an artifact (Sentry Static macos & iphonesimulator)
- Use it instead of current release version
- xcodebuild is no wonly running on iOS 18.5, but i was wondering if we even need this and could just rely on the spm macos tests, wdyt? If not, we could revert the 3rd party specific changes from the shared tests workflow

I also introduced a flag to skip swiftlint, making the builds for testing purposes much faster.

denrase:
@itay they seem to be running quite fast now.

Bildschirmfoto 2026-01-26 um 14 54 28

.github/workflows/unit-test-common.yml
itaybre: I am not sure, this step it to install missing simulators.
I believe we shouldn't need it since we use preinstalled ones.

.github/workflows/unit-test-common.yml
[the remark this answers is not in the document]
itaybre: Probably a leftover from before.
We might have deleted this input

.github/workflows/test-3rd-party-integrations.yml
```diff
@@ -0,0 +1,171 @@
...
+    runs-on: ["ghcr.io/cirruslabs/macos-runner:sequoia", "runner_group_id:10"]
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
+      - name: Select Xcode
+        run: ./scripts/ci-select-xcode.sh 16.4
+      - name: Build XCFramework (Static)
+        run: |
+          rm -rf XCFrameworkBuildPath/
+          mkdir -p XCFrameworkBuildPath
```
itaybre: `l`: This shouldn't be necessary in CI

.github/workflows/test-3rd-party-integrations.yml
```diff
@@ -0,0 +1,171 @@
...
+
+          # Build Sentry (Static) - the only variant needed by 3rd-party integrations
+          # Build both macosx and iphonesimulator slices since tests run on iOS Simulator
+          ./scripts/build-xcframework-slice.sh "macosx" "Sentry" "" "staticlib" "" "skip-swiftlint"
+          ./scripts/build-xcframework-slice.sh "iphonesimulator" "Sentry" "" "staticlib" "" "skip-swiftlint"
+          ./scripts/assemble-xcframework.sh "Sentry" "" "" "macosx,iphonesimulator" "$(pwd)/XCFrameworkBuildPath/archive/Sentry/SDK_NAME.xcarchive"
+          ./scripts/validate-xcframework-format.sh "Sentry.xcframework"
+          ./scripts/compress-xcframework.sh --not-signed Sentry
```
itaybre: `m`: If we added a new variant to `build-xcframework-variant.sh`, this could become easier to mantain:
```
elif [ "$sdks_to_build" = "simulatorAndmacOSOnly" ]; then
    sdks=( iphonesimulator macosx )
```

Then you only need to call `./scripts/build-xcframework-local.sh simulatorAndmacOSOnly StaticOnly --not-signed`

.github/workflows/test-3rd-party-integrations.yml
```diff
@@ -0,0 +1,171 @@
...
+        run: ./scripts/ci-select-xcode.sh 16.4
+      - name: Download XCFrameworks
+        uses: actions/download-artifact@v7
+        with:
+          name: xcframeworks-for-3rd-party-integration-tests
+      - name: Modify Package.swift to use local xcframeworks
+        run: ./scripts/prepare-package.sh --package-file Package.swift --is-pr true --remove-duplicate true --change-path true
+      - name: Use local sentry-cocoa
+        working-directory: 3rd-party-integrations/${{matrix.integration_dir}}
+        run: swift package edit sentry-cocoa --path ../..
```
itaybre: `h`: I was testing this, and seems like by using `swift package edit sentry-cocoa --path ../..`, swift resolves all dependencies and downloads all xcframeworks before performing the change

This would fail in release (changed the version to a non existing one):
image

.github/workflows/unit-test-common.yml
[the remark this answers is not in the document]
itaybre: I would remove `Install required platforms` and `Create simulator device`, this is probably a merge conflict after they were removed from main

Sentry.xcodeproj/project.pbxproj
```diff
@@ -4777,7 +4777,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if [[ \"$(uname -m)\" == arm64 ]]\nthen\n    export PATH=\"/opt/homebrew/bin:$PATH\"\nfi\n\nif command -v swiftlint >/dev/null 2>&1\nthen\n    swiftlint\nelse\n    echo \"warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions.\"\nfi\n";
+			shellScript = "if [ \"${SKIP_SWIFTLINT}\" = \"1\" ] || [ \"$SKIP_SWIFTLINT\" = \"1\" ]; then\n    echo \"Skipping SwiftLint (SKIP_SWIFTLINT=1)\"\n    exit 0\nfi\n\nif [[ \"$(uname -m)\" == arm64 ]]\nthen\n    export PATH=\"/opt/homebrew/bin:$PATH\"\nfi\n\nif command -v swiftlint >/dev/null 2>&1\nthen\n    swiftlint\nelse\n    echo \"warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions.\"\nfi\n";
```
itaybre: `h`: Why do we need to skip swiftlint? 
We shouldn't need to skip them, if there are issues we should fix them or disable the rule for that line if needed

denrase: Yeah since it was hanging locally for me, I assumed that it's running for a long time, which was incorrect. Will remove.

.github/workflows/test-3rd-party-integrations.yml
denrase: Great catch, updated to use sed.

denrase:
@itaybre Added your feedback.

We do both `swift test` on macOS and `xcodebuild` style testing to test on iOS 18.5 simulator. Both test against sentry-cocoa build form source. We could still think about removing the latter, i'm not sure on this. It would be much simpler then.

@philipphofmann @philprime Would be great if one of you could also take a look, as we need a approving review. 🙇

denrase:
Updated to only do swift tests on macOS, as there is no iOS specific code to test. /cc @itaybre

denrase:
@itaybre I'll approve this if it's ok with you.

itaybre:
> @itaybre I'll approve this if it's ok with you.

yes please, approve so we can merge it

I kinda forgot I couldn't approve

PR #6955, opened 2025-12-02
chore(dx): Change generate-public-api to find Xcode 16

philprime:
When managing Xcode installations using `xcodes` it installs them at the path `/Applications/Xcode-<VERSION>.app`, especially `Xcode-16.4.0.app`. This PR changes the script executed by `make generate-public-api` to find the correct version of Xcode automatically, in case the currently selected on is not already Xcode 16

#skip-changelog

### DIFF
--- a/scripts/update-api.sh
+++ b/scripts/update-api.sh
@@ -6,9 +6,30 @@ XCODE_VERSION=$(xcodebuild -version | head -n 1 | awk '{print $2}')
 XCODE_MAJOR_VERSION=$(echo "$XCODE_VERSION" | cut -d. -f1)
 
 if [[ "$XCODE_MAJOR_VERSION" != "16" ]]; then
-    echo "Error: Xcode 16 is required for running the update-api.sh script, because Xcode 26 doesn't include the ObjC public API, but Xcode $XCODE_VERSION is currently selected."
-    echo "Please select Xcode 16 using 'sudo xcode-select' or 'xcodes select 16.x'"
-    exit 1
+    # Try to find an Xcode 16 installation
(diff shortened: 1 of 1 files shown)


scripts/update-api.sh
```diff
@@ -6,9 +6,30 @@ XCODE_VERSION=$(xcodebuild -version | head -n 1 | awk '{print $2}')
 XCODE_MAJOR_VERSION=$(echo "$XCODE_VERSION" | cut -d. -f1)
 
 if [[ "$XCODE_MAJOR_VERSION" != "16" ]]; then
-    echo "Error: Xcode 16 is required for running the update-api.sh script, because Xcode 26 doesn't include the ObjC public API, but Xcode $XCODE_VERSION is currently selected."
-    echo "Please select Xcode 16 using 'sudo xcode-select' or 'xcodes select 16.x'"
-    exit 1
+    # Try to find an Xcode 16 installation
+    XCODE_16_PATH=$(find /Applications -maxdepth 1 -type d -name "Xcode-16*" 2>/dev/null | head -n 1)
```
itaybre, line 10: Wouldn't it be easier to use `xcodes` to provide the path?

philprime: I didn't want to assume that xcodes is installed

itaybre: Since you mentioned this is the path `xcodes` uses, I would say it is safe to assume it is installed.

philprime: No, because [XcodesApp](https://github.com/XcodesOrg/XcodesApp) doesn't install the CLI

PR #6956, opened 2025-12-02
feat(metrics): Add integration with installation by SDK

philprime:
**This PR is part of a merge-chain and should be merged one-by-one into `main` as soon as all of them are ready to be merged:**

1. https://github.com/getsentry/sentry-cocoa/pull/6956
2. https://github.com/getsentry/sentry-cocoa/pull/6960
3. https://github.com/getsentry/sentry-cocoa/pull/6957

## :scroll: Description

This pull request introduces a new metrics integration to the Sentry SDK for Swift, allowing users to enable metrics collection and reporting via the SDK. The changes include the addition of the `enableMetrics` option, implementation of the metrics integration, updates to the SDK configuration, and new tests to verify the integration. The project structure is also updated to include the new integration and its tests.

**Metrics Integration Feature:**

* Added a new `enableMetrics` option to `SentryOptions`, allowing users to enable or disable metrics collection. [[1]](diffhunk://#diff-26ba9567bf49152733f934daac20a4f24b9c887032ff7aa87c4385537b9fb7daR133-R137) [[2]](diffhunk://#diff-944eaed649050d2dc6fad0eb85a106ae792a88680dabca5852ea7396cc5bede5R46223-R46298)
* Implemented `MetricsIntegration`, which is installed when `enableMetrics` is set to `true`. [[1]](diffhunk://#diff-338f88658224c60dc501be89a7ff034c6966bae8e44470bd34953f24510cd038R1-R13) [[2]](diffhunk://#diff-7802b2b5432fe13cbef0398a5cd60ea96feaa86070126271f4315c45907d42fdL37-R40)
* Updated the SDK wrapper and options validation to support the new `enableMetrics` option. [[1]](diffhunk://#diff-3e2347a4fe0f477de6a9eb0fc6a2e9be48e9060580c48a999d1cc15adf77655bR159-R164) [[2]](diffhunk://#diff-9206345e30e699cbc04a94385feb33e7b19ce4168efbc6107f6a89c68c366ef7R99-R101)
* Updated the enabled features builder to include "metrics" when the feature is enabled.

## :bulb: Motivation and Context

Closes #6954

## :green_heart: How did you test it?

- Added unit tests based on the existing option `enableLogs`
- Added unit tests based on the session replay integration
- Added option to enable metrics to run arguments in scheme
  - Enabled the option, ran the macOS sample app and checked that the "Integration initialized" log message shows up.

## :pencil: Checklist

You have to check all boxes before merging:

- [X] I added tests to verify the changes.
[...]

PR #6957, opened 2025-12-02
feat(metrics): Add public API to collect count, distribution and gauge

### LINKED ISSUE #6949: [Metrics] Add public API for integration
* Add integration root at `Sentry.metrics`
* Add `Sentry.metrics.count(String, Double)` ([Reference](https://docs.sentry.io/platforms/javascript/metrics/#counter))
* Add `Sentry.metrics.distribution(String, Double)` ([Reference](https://docs.sentry.io/platforms/javascript/metrics/#gauge))
* Add `Sentry.metrics.gauge(String, Double)` ([Reference](https://docs.sentry.io/platforms/javascript/metrics/#distribution))

philprime:
**This PR is part of a merge-chain and should be merged one-by-one into `main` as soon as all of them are ready to be merged:**

1. https://github.com/getsentry/sentry-cocoa/pull/6956
2. https://github.com/getsentry/sentry-cocoa/pull/6960
3. https://github.com/getsentry/sentry-cocoa/pull/6957

This PR must also wait for #7077 to be merged

## :scroll: Description

This pull request introduces a new metrics API to the codebase, allowing clients to record custom metrics such as counts, distributions, and gauges. The API is exposed via the `SentrySDK.metrics` property and is designed for future extensibility, though the current implementation is a no-op.

This PR also adds an example UI to manually collect metrics in iOS-Swift.

Closes #6949

- Added unit tests

## :pencil: Checklist

- [X] I updated the docs if needed. → https://github.com/getsentry/sentry-docs/pull/15822

## TODO

- [X] Add `Attributable` and `AttributeValue` to support type-safe bool, string, int, float, double literal values and variables
- [x] ~Add `AttributeArrayValue` to support type-safe arrays without mixed types~

### DIFF
--- a/Samples/iOS-Swift/iOS-Benchmarking/SentrySDKPerformanceBenchmarkTests.m
+++ b/Samples/iOS-Swift/iOS-Benchmarking/SentrySDKPerformanceBenchmarkTests.m
@@ -50,6 +50,7 @@
             XCUIElement *moreButton = tabBar.buttons[@"More"];
             if (![moreButton waitForExistenceWithTimeout:1]) {
                 XCTFail("Failed to find more tab item");
+                return;
             }
             [moreButton tap];
 
@@ -58,6 +59,7 @@
                 [app.cells.staticTexts matchingIdentifier:@"Benchmarking"].firstMatch;
             if (![buttonBenchmarking waitForExistenceWithTimeout:1]) {
                 XCTFail("Failed to find button to benchmarking page");
+                return;
             }
             [buttonBenchmarking tap];
         }


philprime:
@sentry review

Sources/Swift/Protocol/Attributes/SentryAttributeValue.swift
```diff
@@ -0,0 +1,167 @@
...
+            self = .stringArray(stringArrayValue)
+        case let boolArrayValue as [Bool]:
+            self = .booleanArray(boolArrayValue)
+        case let intArrayValue as [Int]:
+            self = .integerArray(intArrayValue)
+        case let doubleArrayValue as [Double]:
+            self = .doubleArray(doubleArrayValue)
+        case let floatArrayValue as [Float]:
+            self = .doubleArray(floatArrayValue.map { Double($0) })
+        case let arrayValue as [SentryAttribute]:
```
philprime: I'll look into how this can be improved, so we are not even in the situation of having mixed arrays

Tests/SentryTests/Integrations/Metrics/SentryMetricsApiTests.swift
```diff
@@ -0,0 +1,415 @@
+import Foundation
+@_spi(Private) @testable import Sentry
+@_spi(Private) import SentryTestUtils
+import XCTest
+
+class MetricsApiTests: XCTestCase {
```
philprime: I need to clean these tests up, because they do not make much sense

philprime:
Converted back to draft until I properly handle the typing of the integrations

philprime:
@sentry review
@cursor review

Samples/tvOS-Swift/tvOS-Swift/MetricsView.swift
```diff
@@ -0,0 +1,295 @@
...
+        let interactionType = "button_press"
+        SentrySDK.metrics.count(
+            key: "tvos.app.interaction.string",
+            value: 1,
+            unit: "interaction",
+            attributes: [
+                "interaction_type": interactionType,
+                "device": "apple_tv"
+            ]
+        )
```
itaybre, line 31: `m`: All platform's test apps seem to have the same metrics for testing, wouldn't it be better to move this to SentrySampleShared and call it with the specific prefix for the platform as en argument?
In case we add a new metric, it will be easier a smaller change

Sources/Swift/Integrations/Metrics/SentryMetricsApi.swift
```diff
@@ -0,0 +1,77 @@
...
+
+    // MARK: - Private
+
+    private func recordMetric(
+        name: String,
+        value: SentryMetric.Value,
+        unit: SentryMetricsUnit?,
+        attributes: [String: SentryAttributeValue]
+    ) {
+        guard dependencies.isSDKEnabled else {
```
itaybre, line 42: `l`: Looks like a good placement for some logs

Sources/Swift/Integrations/Metrics/SentryMetricsApiProtocol.swift
```diff
@@ -0,0 +1,65 @@
...
+    ///   - unit: Optional unit of measurement (e.g., "byte", "connection")
+    ///   - attributes: Optional dictionary of attributes to attach to the metric.
+    ///                 Values can be String, Bool, Int, Double, Float, or SentryAttribute.
+    ///                 Example: `["process": "main_app", "compressed": true, "pressure_level": 2]`
+    func gauge(key: String, value: Double, unit: SentryMetricsUnit?, attributes: [String: SentryAttributeValue])
+}
+
+public extension SentryMetricsApiProtocol {
+    func count(key: String, value: UInt, unit: SentryMetricsUnit? = nil, attributes: [String: SentryAttributeValue] = [:]) {
+        self.count(key: key, value: value, unit: unit, attributes: attributes)
```
itaybre, line 70: What is this? Is it calling itself?

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
```diff
@@ -0,0 +1,218 @@
+/// Type used for known attribute units used by Metrics
+///
+/// - SeeAlso: https://develop.sentry.dev/sdk/telemetry/attributes/#units
+public enum SentryMetricsUnit {
```
itaybre: `l`: This feels the same as `SentryMeasurementUnit`, I wonder if we can avoid duplication of units

Sources/Swift/Protocol/SentryUnit.swift
```diff
@@ -0,0 +1,218 @@
...
+
+    // MARK: - Generic
+
+    case generic(String)
+}
+
+extension SentryMetricsUnit: RawRepresentable {
+    public typealias RawValue = String
+
+    public init?(rawValue: String) { // swiftlint:disable:this cyclomatic_complexity function_body_length
```
itaybre, line 57: `l` feels easier to read

```suggestion
    // swiftlint:disable:this cyclomatic_complexity function_body_length
    public init?(rawValue: String) { 
```

Tests/SentryTests/Integrations/Metrics/SentryMetricsApiTests.swift
```diff
@@ -0,0 +1,520 @@
...
+            metricsIntegration: nil
+        )
+        let sut = SentryMetricsApi(dependencies: dependencies)
+        
+        // -- Act --
+        sut.count(key: "test.metric", value: 1)
+        
+        // -- Assert --
+        // When metricsIntegration is nil, no metrics should be created
+        // This simulates the case where metrics are disabled (integration returns nil from init)
```
itaybre: `m`: I think we are missing an assert here, we are not validating anything

Tests/SentryTests/Integrations/Metrics/SentryMetricsApiTests.swift
```diff
@@ -0,0 +1,520 @@
...
+        let metric: SentryMetric = invocation.0
+        XCTAssertEqual(metric.name, "api.request.count", "Metric key should match")
+        XCTAssertEqual(metric.value, SentryMetric.Value.counter(1), "Metric value should match")
+        
+        XCTAssertEqual(metric.attributes["endpoint"]?.anyValue as? String, "api/users", "endpoint attribute value should match")
+        XCTAssertEqual(metric.attributes["success"]?.anyValue as? Bool, true, "success attribute value should match")
+        XCTAssertEqual(metric.attributes["status_code"]?.anyValue as? Int, 200, "status_code attribute value should match")
+        XCTAssertEqual(try XCTUnwrap(metric.attributes["response_time"]?.anyValue as? Double), 0.125, accuracy: 0.001, "response_time attribute value should match")
+    }
+//
```
itaybre:
```suggestion

```

Tests/SentryTests/Integrations/Metrics/SentryMetricsApiTests.swift
```diff
@@ -0,0 +1,520 @@
...
+            metricsIntegration: nil
+        )
+        let sut = SentryMetricsApi(dependencies: dependencies)
+        
+        // -- Act --
+        sut.distribution(key: "test.metric", value: 1.0)
+        
+        // -- Assert --
+        // When metricsIntegration is nil, no metrics should be created
+        // This simulates the case where metrics are disabled (integration returns nil from init)
```
itaybre: Same as above

Samples/tvOS-Swift/tvOS-Swift/MetricsView.swift
philippsuess: I actually prefer having duplicate code because it's easier to modify and experiment with the samples. The samples should also be something for the users to look at, actually. They should showcase a bit on how to use the SDK. Now we also use them for testing and have all this SentrySampleShared, which makes it harder to quickly play around with the SDK in the samples. I would prefer ditching the SentrySampleShared completely, actually, and have most code duplicated in the samples.

Sources/Swift/Helper/SentrySDK.swift
```diff
@@ -37,7 +37,10 @@ import Foundation
             return SentryLogger(dateProvider: SentryDependencyContainer.sharedInstance().dateProvider)
         }
     }
-    
+
+    /// API to access Sentry metrics
+    public static var metrics: SentryMetricsApiProtocol = SentryMetricsApi(dependencies: SentryDependencyContainer.sharedInstance())
```
philippsuess: `m`: I think we should add an explanation why this isn't available from ObjC, because all other code in here is.

Sources/Swift/Helper/SentrySDK.swift
```diff
@@ -37,7 +37,10 @@ import Foundation
             return SentryLogger(dateProvider: SentryDependencyContainer.sharedInstance().dateProvider)
         }
     }
-    
+
+    /// API to access Sentry metrics
```
philippsuess: `m`: I know the logs API above has the same code comment, but I think we could add a more detailed comment about the logs API. This is the main entry point for metrics. It could explain a bit what it is or reference other classes to see how to use it.

Sources/Swift/Integrations/Metrics/SentryMetricsApi.swift
philippsuess: Yes, indeed 👍 . Same for the integration check below.

Sources/Swift/Integrations/Metrics/SentryMetricsApi.swift
```diff
@@ -0,0 +1,77 @@
...
+            return
+        }
+        guard let integration = dependencies.metricsIntegration else {
+            return
+        }
+
+        let metric = SentryMetric(
+            timestamp: Date(),
```
philippsuess: `m`: It would be great to use the `SentryCurrentDateProvider` here, then we can also properly test the timetampe in `SentryMetricsApiTests`.

Sources/Swift/Integrations/Metrics/SentryMetricsApi.swift
```diff
@@ -0,0 +1,77 @@
...
+            traceId: SentryId.empty, // Will be set by batcher from scope
```
philippsuess: `m`: I wonder if it's not better to set the fallback of the `propagationContextTraceId` already here, and not to wait until we do it here https://github.com/getsentry/sentry-cocoa/blob/f402adbfdf5961e6d75564c4e8acdb95acd843d1/Sources/Swift/Tools/Batcher/BatcherScope.swift#L50-L52

Because we now have a metric with a somewhat wrong traceID. Even when this works now, it's so easy to mess this up in the future, when we don't set the `propagationContextTraceId` in the BatcherScope. I would rather already create a metric with the correct fallback traceID here.

Sources/Swift/Integrations/Metrics/SentryMetricsApiProtocol.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// Records a count metric for the specified key.
+    ///
+    /// Use this to increment or set a discrete occurrence count associated with a metric key,
+    /// such as the number of events, requests, or errors.
+    ///
+    /// - Parameters:
+    ///   - key: A namespaced identifier for the metric (for example, "network.request.count").
+    ///          Prefer stable, lowercase, dot-delimited names to aid aggregation and filtering.
+    ///   - value: The count value to record. Typically a non-negative integer (e.g., 1 to increment by one).
+    ///            Values less than zero may be ignored or clamped by the metrics backend.
```
philippsuess: `l`: It's not just typically a non-negative integer. The type is UInt, it's alsways a non-negative integer. As it's a UInt we can remove the note about values less than zero.
```suggestion
    ///   - value: The count value to record. A non-negative integer (e.g., 1 to increment by one).
```

Sources/Swift/Integrations/Metrics/SentryMetricsApiProtocol.swift
```diff
@@ -0,0 +1,65 @@
...
+    ///   - unit: Optional unit of measurement (e.g., "request", "error")
```
philippsuess: `l`: request and error aren't typical units of measurements. I think that should be 
```suggestion
    ///   - unit: Optional unit of measurement (e.g., "millisecond", "byte")
```

Sources/Swift/Integrations/Metrics/SentryMetricsApiProtocol.swift
```diff
@@ -0,0 +1,65 @@
...
+    ///   - attributes: Optional dictionary of attributes to attach to the metric.
+    ///                 Values can be String, Bool, Int, Double, Float, or SentryAttribute.
```
philippsuess: `l`: I think it would be good to also mention string[], boolean[], integer[], double[], and Float isn't supported actually. The following code doesn't compile

```swift
let myfloat: Float = 3.14

SentrySDK.metrics.count(
    key: "ios.app.action.boolean",
    value: myfloat
)
```

I'm unsure whether we should mention `SentryAttribute` here, because that's the ObjC compatible API. We use `SentryAttributeContent` / `SentryAttributeValue` here.

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
```diff
@@ -0,0 +1,218 @@
...
+             (.ratio, .ratio),
+             (.percent, .percent):
+            return true
+        case (.generic(let lhsValue), .generic(let rhsValue)):
+            return lhsValue == rhsValue
+        default:
+            return false
+        }
+    }
+}
```
philippsuess: `m`: I think if we do `public enum SentryMetricsUnit : Equatable` above, we can ditch this here. The compiler should synthesize this for you.

Sources/Swift/Protocol/SentryUnit.swift
```diff
@@ -0,0 +1,218 @@
...
+    case ratio
+    case percent
+
+    // MARK: - Generic
+
+    case generic(String)
+}
+
+extension SentryMetricsUnit: RawRepresentable {
+    public typealias RawValue = String
```
philippsuess, line 55: `m`: It took me a bit to figure out that we need this extension to support custom metric units. That's also the reason why we can't use `SentryMetricsUnit : String`. It would be great to briefly explain this here, because the generic(String) is at the bottom.

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
philippsuess: Yes, indeed, but I guess the main reason was ObjC to Swift compatibility. These units aren't restricted to metrics, actually. They are overall telemetry units, so I don't think we should call this enum `SentryMetricUnits`, but rather `SentryUnits` or something similar to communicate that these aren't specific to metrics. According to the develop docs, these can also be used for (Spans, Logs, Metrics): https://develop.sentry.dev/sdk/telemetry/attributes/#units

Sources/Swift/Integrations/Metrics/SentryMetricsApiProtocol.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// - Parameters:
+    ///   - key: A namespaced identifier for the metric (for example, "http.request.duration").
+    ///          Prefer stable, lowercase, dot-delimited names to aid aggregation and filtering.
+    ///   - value: The value to record in the distribution. This can be any numeric value
+    ///            representing the measurement (e.g., milliseconds for response time).
+    ///   - unit: Optional unit of measurement (e.g., "millisecond", "byte")
+    ///   - attributes: Optional dictionary of attributes to attach to the metric.
+    ///                 Values can be String, Bool, Int, Double, Float, or SentryAttribute.
+    ///                 Example: `["endpoint": "/api/data", "cached": false, "response_size": 1024.5]`
+    func distribution(key: String, value: Double, unit: SentryMetricsUnit?, attributes: [String: SentryAttributeValue])
```
philippsuess: `m`: I wonder, though, if the current approach with `SentryMetricsUnit` here isn't too loose. While we don't want to limit users and which units to use, the current API could do a better job communicating which units we support. When using the API, I just type a string.
```
SentrySDK.metrics.count(
            key: "ios.app.action.boolean",
            value: 1,
            unit: "action",
            attributes: [
                "window_active": windowActive,
                "focused": false
            ]
        )
```

The easiest fix would be to more clearly point this out in the here in the code docs.

Samples/tvOS-Swift/tvOS-Swift/MetricsView.swift
philprime: Also all of the metrics have different keys so we can easily differentiate them.
I agree with @philipphofmann on this, our samples should reflect real-life usage, and not use some shared framework no one else has.

Sources/Swift/Protocol/SentryUnit.swift
philprime: Yes, but if we add documentation above it, there would be a comment between documentation and the method, that's why is use inline comments instead

Sources/Swift/Integrations/Metrics/SentryMetricsApiProtocol.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// - Parameters:
+    ///   - key: A namespaced identifier for the metric (for example, "network.request.count").
+    ///          Prefer stable, lowercase, dot-delimited names to aid aggregation and filtering.
+    ///   - value: The count value to record. Typically a non-negative integer (e.g., 1 to increment by one).
+    ///            Values less than zero may be ignored or clamped by the metrics backend.
+    ///   - unit: Optional unit of measurement (e.g., "request", "error")
+    ///   - attributes: Optional dictionary of attributes to attach to the metric.
+    ///                 Values can be String, Bool, Int, Double, Float, or SentryAttribute.
+    ///                 Example: `["endpoint": "api/users", "success": true, "status_code": 200]`
+    func count(key: String, value: UInt, unit: SentryMetricsUnit?, attributes: [String: SentryAttributeValue])
```
philippsuess: `h`: Missing default value for counter parameter

According to the [metrics spec](https://develop.sentry.dev/sdk/telemetry/metrics/), the counter should default to 1: "The count to increment by (should default to 1)".

The JavaScript SDK implements this as:
```typescript
export function count(name: string, value: number = 1, options?: MetricOptions): void
```

**Current behavior:**
```swift
SentrySDK.metrics.count(key: "button.click", value: 1) // verbose
```

**Expected behavior:**
```swift
SentrySDK.metrics.count(key: "button.click") // should default to 1
```

**Suggested fix:**
```swift
func count(key: String, value: UInt = 1, unit: SentryMetricsUnit?, attributes: [String: SentryAttributeValue])
```

This will also require updating the default extension at line 54-56 to remove the explicit call since the protocol now has a default.

Sources/Swift/Integrations/Metrics/SentryMetricsApi.swift
philippsuess: `h`: Potential issue with empty trace IDs

Setting `traceId: SentryId.empty` with the expectation that "it will be set by batcher from scope" is fragile. If the batcher enrichment fails or is skipped for any reason, metrics will be sent with empty trace IDs, breaking trace correlation.

**JavaScript SDK approach:**
They set trace context during serialization, not at metric creation:
```typescript
const traceId = span ? span.spanContext().traceId : traceContext?.trace_id;
```

**Recommendations:**
1. **Preferred:** Fetch and set the trace ID here at creation time from the scope:
```swift
let traceId = dependencies.scope.span?.traceId ?? dependencies.scope.propagationContext.traceID
```

2. **Alternative:** If deferring to batcher is intentional (e.g., for performance), add validation in the batcher to ensure trace ID is never empty, and log a warning if it is.

Could you clarify the design decision here? Is there a reason to defer trace ID resolution to the batcher?

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
philippsuess: `m`: Document the typed unit enum as an intentional spec deviation

Your strongly-typed unit enum is a nice Swift-native improvement over the spec's free-form strings! The `.generic(String)` escape hatch provides the flexibility needed.

**Spec:** `unit?: string`  
**Cocoa SDK:** Typed enum with known units + generic fallback

**Benefits:**
- ✅ Type safety and autocomplete
- ✅ Prevents typos for common units
- ✅ Still extensible via `.generic(String)`

**Suggestion:** Add a doc comment explaining this design decision:

```swift
/// Type-safe enum for metric units following the Sentry attributes specification.
///
/// This provides compile-time safety and IDE autocomplete for common units while remaining
/// extensible via the `.generic(String)` case. This is a Swift-native improvement over the
/// spec's free-form string approach.
///
/// - SeeAlso: https://develop.sentry.dev/sdk/telemetry/attributes/#units
public enum SentryMetricsUnit {
```

This will help future maintainers understand it's intentional, not an oversight.

Sources/Swift/Integrations/Metrics/SentryMetricsApi.swift
```diff
@@ -0,0 +1,77 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+protocol SentryMetricsApiDependencies {
```
philippsuess, line 4: praise: 👏 Exemplary dependency injection pattern for testability!

The protocol-based dependencies make this infinitely more testable than the JavaScript SDK's global singleton approach:

```swift
protocol SentryMetricsApiDependencies {
    var isSDKEnabled: Bool { get }
    var scope: Scope { get }
    var metricsIntegration: Integration? { get }
}
```

**Benefits:**
- ✅ Easy to mock in unit tests (no global state)
- ✅ Clear dependencies visible in the protocol
- ✅ Integration is properly nullable (fail-safe if not installed)

The test file `SentryMetricsApiTests.swift` shows this pays off immediately with clean, isolated unit tests. This is a great example of how to structure SDK code for maintainability.

Tests/SentryTests/Integrations/Metrics/SentryMetricsApiTests.swift
```diff
@@ -0,0 +1,520 @@
+import Foundation
+@_spi(Private) @testable import Sentry
+@_spi(Private) import SentryTestUtils
+import XCTest
+
+/// This test suite tests the SentryMetricsApi on a unit test level to verify the logic without bootstrapping an entire SDK setup.
+///
+/// Having additional unit tests can help pin down if failing end-to-end tests are caused by unrelated internal programming errors.
+final class SentryMetricsApiTests: XCTestCase {
```
philippsuess, line 9: praise: ✅ Excellent test coverage and organization!

Having both unit tests (`SentryMetricsApiTests.swift`) and E2E tests (`SentryMetricsApiE2ETests.swift`) is exactly the right approach.

**Test coverage highlights:**
- ✅ SDK disabled scenarios (line 114)
- ✅ Metrics integration disabled (line 130)
- ✅ Zero values (line 147)
- ✅ Large values (line 166)
- ✅ All attribute types with variables and literals (lines 185+)
- ✅ Scope passing validation (lines 34-35)

The test naming convention (`test<Function>_when<Condition>_should<Expected>`) is also spot-on and matches your AGENTS.md guidelines. Great consistency!

**Suggestion:** Consider adding one more test case:
```swift
func testCount_whenSendDefaultPiiDisabled_shouldNotIncludeUserAttributes()
```
to verify PII handling (may be in a separate PR if the batcher handles this).

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
```diff
@@ -0,0 +1,218 @@
...
+
+extension SentryMetricsUnit: Encodable {
+    public func encode(to encoder: any Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(self.rawValue)
+    }
+}
+
+extension SentryMetricsUnit: ExpressibleByStringLiteral {
```
philippsuess: praise: 🎯 Clever use of `ExpressibleByStringLiteral`!

```swift
extension SentryMetricsUnit: ExpressibleByStringLiteral {
    public init(stringLiteral value: StringLiteralType) {
        self = .generic(value)
    }
}
```

This provides the best of both worlds:
- ✅ Type safety: `unit: .millisecond`
- ✅ Flexibility: `unit: "custom_unit"`
- ✅ No conversion ceremony required

Users can write `unit: "requests"` and it automatically becomes `.generic("requests")`. This is a great example of Swift's expressiveness making the API more pleasant to use.

Samples/iOS-Swift/iOS-Swift/MetricsViewController.swift
```diff
@@ -0,0 +1,283 @@
+// swiftlint:disable function_body_length
+import Sentry
+import UIKit
+
+class MetricsViewController: UIViewController {
```
philippsuess, line 5: praise: 📚 Comprehensive sample code showing all attribute types!

The sample app demonstrates:
- ✅ All metric types (count, distribution, gauge)
- ✅ All attribute types (String, Bool, Int, Double, arrays)
- ✅ Both variables and literals
- ✅ With and without units

This is exactly what developers need to understand the API. The comments are clear (e.g., line 6: "demonstrates all supported attribute types as variables and literals").

**Suggestion:** Consider adding one example showing the `beforeSendMetric` callback in action:
```swift
options.experimental.beforeSendMetric = { metric in
    var modified = metric
    // Example: Add environment tag
    modified.attributes["environment"] = "production"
    return modified
}
```

This would help developers understand when and how to use the callback.

Sources/Swift/Integrations/Metrics/SentryMetricsApiProtocol.swift
philprime: Protocols can't have default values, therefore without this extension the SDK users must set `unit: nil` and `attributes: [:]` when using the method every time.
This extension results in these methods to be valid to:

```
count(key: "key", value: 123)
count(key: "key", value: 123, unit: "bytes")
count(key: "key", value: 123, attributes: ["key": "value"])
```

If all parameters are set by the SDK user, it will directly call the implementation of the protocol and not this extension. The unit tests proof that it will not end up in a recursion.

I am going to add a comment to keep this behaviour documented

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
philprime: You are right, when I did the original implementation it didn't synthesize it, but now it seems to work fine.

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
philprime: I wasn't aware we already had a `SentryMeasurementUnit`, but this one is now the most Swift-style implementation possible, especially as it allows associated values for `.generic("my unit")` which Objective-C does not support.

What do you think about having feature specific types, e.g. `SentryLogUnit`, `SentryMetricsUnit` and `SentrySpanUnit` which are all typealias for `SentryMeasurementUnit` or the enum I created here (which I would rename to `SentryUnit`)?

That way we can have them alongside for now and merge them at a later stage.

Tests/SentryTests/Integrations/Metrics/SentryMetricsApiTests.swift
philprime: It can not be asserted, because it doesn't do antyhing. I added a comment explaining it.

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
itaybre: I believe it would be better to have a single unit since as @philipphofmann said, it is a telemetry concept not linked to any feature.

But I would do it only if it is easy to replace the old ObjC one, which may be a breaking change.

philprime: @itaybre how would you expose a Swift-only type into Objective-C? This is technically not possible, and I don't think we should give up Swift-like functionality if this feature as we all decided this is going to be best-practice implementation for Swift

philprime: Also the typealias is just for convenience, so they stay changeable without breaking public API

philippsuess: I would do what gets us closer to an ideal solution. At some point, we're going to deprecate `SentryMeasurementUnit` and then delete it. Once we get rid of transactions, we're going to do that.
I think it's not a good idea to use the `SentryMeasurementUnit` for logs, spansV2 and metrics. I also wouldn't do different unit types.

TLDR; I would do SentryUnit and use it in the metrics. When we can, we should deprecate `SentryMeasurementUnit` and ditch it.

Sources/Swift/Helper/SentrySDK.swift
```diff
@@ -37,7 +37,51 @@ import Foundation
...
+    ///     key: "http.request.duration",
+    ///     value: 187.5,
+    ///     unit: .millisecond,
+    ///     attributes: ["endpoint": "/api/data", "cached": false]
+    /// )
+    /// ```
+    ///
+    /// ## Requirements
+    ///
+    /// Metrics must be enabled via ``Options/enableMetrics`` before metrics are sent to Sentry.
```
philippsuess: `h`: I think we decided to enable metric by default on all SDKs, because they don't send anything by default and you have to call the APIs manually. See [Slack](https://sentry.slack.com/archives/C09H3UT3X2Q/p1767727212911739?thread_ts=1767625890.869059&cid=C09H3UT3X2Q) and also develop docs: https://develop.sentry.dev/sdk/telemetry/metrics/#initialization-options

Sources/Swift/Integrations/Metrics/SentryMetricsApi.swift
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    func gauge(key: String, value: Double, unit: SentryMetricsUnit? = nil, attributes: [String: SentryAttributeValue] = [:]
+    ) {
+        recordMetric(name: key, value: .gauge(value), unit: unit, attributes: attributes)
+    }
+
+    // MARK: - Private
+
+    private func recordMetric(
```
philippsuess: `l`: I just noticed this when reading the log messages below, we usually have the term capture for events, transactions, etc. Maybe it make sense to name this method `captureMetric` to align. Also, the integration below uses the naming `addMetric`. It would be great to stick to the same term, IMO.

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
philprime: @itaybre is renaming it to SentryUnit without an typealias ok for you?

Sources/Swift/Integrations/Metrics/SentryMetricsApi.swift
```diff
@@ -0,0 +1,97 @@
...
+        //    rather than empty ones, which would break trace correlation entirely.
+        // 2. Semantic correctness: A metric recorded during a network request should correlate
+        //    with that request's trace, matching how we capture timestamp at creation time.
+        // 3. Fail-safe redundancy: The batchers scope enrichment will overwrite this value, producing
+        //    the same result but with a safety net for edge cases. We can not remove the duplication from
+        //    the batcher scope because it is used by Logs and other integrations.
+        //
+        // When a span is active, use its traceId to ensure consistency with span_id.
+        // Otherwise, fall back to propagationContext traceId.
+        let traceId = dependencies.scope.span?.traceId ?? dependencies.scope.propagationContextTraceId
```
philippsuess, line 66: Yep better thanks 💯

Sources/Swift/Integrations/Metrics/SentryMetricsUnit.swift
```diff
@@ -0,0 +1,193 @@
...
+    case generic(String)
+}
+
+// MARK: - RawRepresentable
+//
+// We implement RawRepresentable manually instead of using `enum SentryMetricsUnit: String` because
+// the enum includes `.generic(String)` for custom unit values. Swift's automatic String raw value
+// synthesis doesn't support associated values, so we need this custom implementation to:
+// 1. Map known unit strings to their corresponding enum cases
+// 2. Fall back to `.generic(rawValue)` for any unrecognized string (custom units)
```
philippsuess: Better thanks 💯

Sources/Swift/Helper/SentrySDK.swift
philippsuess: We can also follow up on this in another PR.

philprime: changed it

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -31,8 +31,8 @@ public final class SentryExperimentalOptions: NSObject {
 
     /// When enabled, the SDK sends metrics to Sentry. Metrics can be captured using the SentrySDK.metrics
     /// API, which allows you to send, view and query counters, gauges and measurements.
-    /// @note Default value is @c false.
-    @objc public var enableMetrics: Bool = false
+    /// @note Default value is @c true.
+    @objc public var enableMetrics: Bool = true
```
philippsuess, line 35: 🥇

And same as:  https://github.com/getsentry/sentry-cocoa/pull/6957#discussion_r2695112355

Sources/Swift/Helper/SentrySDK.swift
```diff
@@ -71,7 +71,7 @@ import Foundation
     ///
     /// ## Requirements
     ///
-    /// Metrics must be enabled via ``Options/enableMetrics`` before metrics are sent to Sentry.
+    /// Metrics are enabled by default. To disable metrics, set ``Options/experimental`` ``SentryExperimentalOptions/enableMetrics`` to `false`.
```
philippsuess: `l`: Maybe worth explaining why we enable an experimental feature by default. It's because you have to manually call the API to use it, so just enabling it per default doesn't do anything.

PR #6960, opened 2025-12-02
feat(metrics): Add implementation for metrics envelope item

### LINKED ISSUE #6948: [Metrics] Add options for configuring integration
* Add `beforeSendMetric` to options ([JavaScript Reference](https://docs.sentry.io/platforms/javascript/metrics/#filtering-and-modifying-metrics))

### LINKED ISSUE #6951: [Metrics] Implement Metrics Protocol and `trace_metric` envelope type
* Define the [Metrics Protocol](https://develop.sentry.dev/sdk/telemetry/metrics/#metrics-protocol) and `trace_metric` envelope type in the SDK.

### LINKED ISSUE #6952: [Metrics] Attach default attributes and custom user attributes to metrics

philprime:
**This PR is part of a merge-chain and should be merged one-by-one into `main` as soon as all of them are ready to be merged:**

1. https://github.com/getsentry/sentry-cocoa/pull/6956
2. https://github.com/getsentry/sentry-cocoa/pull/6960
3. https://github.com/getsentry/sentry-cocoa/pull/6957

## :scroll: Description

This PR implements the metrics envelope item functionality for the Sentry Cocoa SDK. It adds the core infrastructure for capturing, batching, and sending metrics to Sentry.

### Key Changes

- **SentryMetric Protocol**: Added a new `SentryMetric` class that represents a metric entry with support for:
  - Three metric types: counter, gauge, and distribution
  - Timestamp, trace ID, and optional span ID for distributed tracing
  - Metric name, value, unit, and structured attributes
  - JSON encoding support for envelope serialization

- **SentryMetricBatcher**: Implemented a batcher that:
  - Batches metrics together before sending (configurable max count: 100, max buffer size: 1MB)
  - Flushes metrics automatically after a timeout (default: 5 seconds) or when limits are reached
  - Enriches metrics with scope attributes (environment, release, user info, scope attributes)
  - Sets trace ID from propagation context and span ID from active span
  - Supports `beforeSendMetric` callback for filtering/modifying metrics

- **MetricsIntegration Updates**: Enhanced the integration to:
  - Initialize and manage the metric batcher
  - Provide `addMetric(_:scope:)` API for adding metrics
  - Flush pending metrics on uninstall

- **Client Support**: Added `captureMetricsData(_:with:)` method to `SentryClient` that:
  - Creates envelope items with type `trace_metric`
  - Uses content type `application/vnd.sentry.items.trace-metric+json`
  - Includes item count in the envelope item header

- **Data Category Support**: Added `trace_metric` data category to:
  - `SentryDataCategory` enum
  - `SentryDataCategoryMapper` for mapping envelope item types to categories
  - `SentryEnvelopeItemType` for the envelope item type constant

- **Options**: Added `beforeSendMetric` callback option for filtering/modifying metrics before sending

### Testing

Added comprehensive test coverage:
- **SentryMetricTests**: Tests for metric creation, encoding, attribute management, and metric type handling
- **SentryMetricBatcherTests**: Tests for batching behavior, timeout handling, buffer size limits, attribute enrichment, scope integration, and beforeSendMetric callback
- **MetricsIntegrationTests**: Tests for integration initialization and metric addition
- **SentryClientTests**: Tests for `captureMetricsData` envelope creation

Closes #6948
Closes #6951
Closes #6952

- Added comprehensive unit tests for `SentryMetric` covering encoding, attribute management, and metric types
- Added extensive unit tests for `SentryMetricBatcher` covering batching, flushing, timeout behavior, attribute enrichment, and edge cases
- Added integration tests for `MetricsIntegration` and `SentryClient` metrics capture
- Verified envelope item creation with correct type and content type
- Tested data category mapping for trace_metric

## :pencil: Checklist


philprime:
@sentry review

philprime:
@sentry review

philprime:
@sentry review

Sources/Swift/Integrations/Metrics/MetricsIntegration.swift
```diff
@@ -1,13 +1,45 @@
...
+            capturedDataCallback: { data, count in
+                let hub = SentrySDKInternal.currentHub()
+                guard let client = hub.getClient() else {
+                    SentrySDKLog.debug("MetricsIntegration: No client available, dropping metrics")
+                    return
+                }
+                client.captureMetricsData(data, with: NSNumber(value: count))
+            }
+        )
+        super.init()
```
noahsmartin: No need for this line, swift adds it automatically

philprime:
```suggestion
```

philprime:
@sentry review

Sources/Swift/Tools/Batcher/BatcherItem.swift
```diff
@@ -1,5 +1,4 @@
 protocol BatcherItem: Encodable {
     var attributes: [String: SentryAttribute] { get set }
     var traceId: SentryId { get set }
-    var body: String { get }
 }
```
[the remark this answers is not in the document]
philprime: It's only used in the SentryLog.encode method, so it's safe to remove

Sources/Swift/Integrations/Metrics/SentryMetricsBatcher.swift
```diff
@@ -0,0 +1,90 @@
...
+    ///              Passing a concurrent queue will result in undefined behavior and potential data races.
+    ///
+    /// - Note: Metrics are flushed when either `maxMetricCount` or `maxBufferSizeBytes` limit is reached.
+    init(
+        options: MetricBatcherOptionsProtocol,
+        flushTimeout: TimeInterval = 5,
+        maxMetricCount: Int = 100, // Maximum 100 metrics per batch
+        maxBufferSizeBytes: Int = 1_024 * 1_024, // 1MB buffer size
+        dateProvider: SentryCurrentDateProvider,
+        dispatchQueue: SentryDispatchQueueWrapper,
```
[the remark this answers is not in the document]
philprime: Options are not changed over time

Sources/Swift/Integrations/Metrics/SentryMetricsIntegration.swift
```diff
@@ -1,13 +1,44 @@
-final class MetricsIntegration<Dependencies>: NSObject, SwiftIntegration {
+@_implementationOnly import _SentryPrivate
+
+final class MetricsIntegration<Dependencies: DateProviderProvider & DispatchQueueWrapperProvider>: NSObject, SwiftIntegration {
+    private let options: Options
```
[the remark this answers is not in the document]
philprime: Options are not changed after SDK initialization

Sources/Sentry/include/SentryDataCategory.h
```diff
@@ -20,5 +20,6 @@ typedef NS_ENUM(NSUInteger, SentryDataCategory) {
     kSentryDataCategorySpan = 11,
     kSentryDataCategoryFeedback = 12,
     kSentryDataCategoryLogItem = 13,
-    kSentryDataCategoryUnknown = 14,
+    kSentryDataCategoryTraceMetric = 14,
+    kSentryDataCategoryUnknown = 15,
 };
```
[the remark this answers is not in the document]
philprime: It's an internal number used to bridge Objective-C

Sources/Sentry/SentryClient.m
```diff
@@ -1124,6 +1124,18 @@ - (void)captureLogsData:(NSData *)data with:(NSNumber *)itemCount
...
+{
+    SentryEnvelopeItem *envelopeItem =
+        [[SentryEnvelopeItem alloc] initWithType:SentryEnvelopeItemTypes.traceMetric
+                                            data:data
+                                     contentType:@"application/vnd.sentry.items.trace-metric+json"
+                                       itemCount:itemCount];
+    SentryEnvelope *envelope = [[SentryEnvelope alloc] initWithHeader:[SentryEnvelopeHeader empty]
+                                                           singleItem:envelopeItem];
+    [self captureEnvelope:envelope];
+}
```
philippsuess, line 1141: `l`: This is basically the same as `captureLogsData` but only the `contentType` and `SentryEnvelopeItemTypes` are different. We could extract a common method that both methods use.

Sources/Swift/Integrations/Metrics/MetricBatcher.swift
```diff
@@ -0,0 +1,90 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+/// Protocol for batching metrics with scope-based attribute enrichment.
+protocol MetricBatcherProtocol {
```
philippsuess: `l`: We already have the Metric**s**Integration. I would keep the naming consistend.
```suggestion
protocol MetricsBatcherProtocol {
```

Sources/Swift/Integrations/Metrics/MetricBatcher.swift
```diff
@@ -0,0 +1,90 @@
...
+    var enableMetrics: Bool { get }
+    var beforeSendMetric: ((Metric) -> Metric?)? { get }
+    var environment: String { get }
+    var releaseName: String? { get }
+    var cacheDirectoryPath: String { get }
+}
+
+struct MetricBatcher: MetricBatcherProtocol {
+    private let isEnabled: Bool
+    private let batcher: any BatcherProtocol<Metric, Scope>
```
philippsuess: `m`: I find this a bit confusing. The `MetricBatcher` implements the `MetricBatcherProtocol` and also has an internal batcher. That doesn't seem right to me. With this approach, we now need a different Batcher implementation for each new telemetry type we need a batcher for. 

When we look at the `SentryLogBatcher` and the `MetricBatcher`, I see similarities. It would be great if we could have one "GenericBatcher" (we for sure need a better name), that you can use for both logs and metrics, but just the config needs to be different.

Sources/Swift/Protocol/Metric.swift
```diff
@@ -0,0 +1,289 @@
+/// A metric entry that captures metric data with associated attribute metadata.
+///
+/// Use the `options.beforeSendMetric` callback to modify or filter metric data.
+public struct Metric {
```
philippsuess: `m`: Should we maybe add a comment explaining that this is Swift only by intention, and that we still figuring out wheter or how this should be exposed to Objc?

Sources/Swift/Protocol/Metric.swift
```diff
@@ -0,0 +1,289 @@
...
+
+// MARK: - MetricType
+
+/// The type of metric being recorded.
+///
+/// Different metric types serve different purposes:
+/// - **Counter**: Incrementing integer values that only increase (e.g., total requests, errors)
+/// - **Gauge**: Current value at a point in time that can go up or down (e.g., active connections, queue size)
+/// - **Distribution**: Statistical distribution of values for aggregation (e.g., response times, payload sizes)
+public enum MetricType: Encodable {
```
philippsuess: `m`: This actually implements both decode and encode
```suggestion
public enum MetricType: Codeable {
```

philippsuess: `h`: Can't we define the Codable on an extension and keep it internal as we do for `Metric`? I would really appreciate it if we didn't make conformances to Codable public, because it allows for more flexibility. We can always make that public later.

```
extension MetricType: Codable
```

Sources/Swift/Protocol/Metric.swift
philippsuess: `h`: Is it now SentryMetric or Metric? If I'm not mistaken we decided on prefixing it with Sentry on Slack. This applies for all public APIs here. So 
```suggestion
public struct SentryMetric {
```

Sources/Swift/Protocol/Metric.swift
```diff
@@ -0,0 +1,289 @@
...
+
+    /// The timestamp when the metric was recorded.
+    public var timestamp: Date
+
+    /// The type of metric (counter, gauge, or distribution).
+    ///
+    /// - `.counter`: Incrementing integer values (e.g., request counts)
+    /// - `.gauge`: Current value at a point in time (e.g., active connections)
+    /// - `.distribution`: Statistical distribution of values (e.g., response times)
+    public let metricType: MetricType
```
philippsuess: `m`: Making this let doesn't allow changing the type in hooks. If that's desired, I would add a comment explaining why you can't change it.

Sources/Swift/Protocol/SentryMetric.swift
```diff
@@ -0,0 +1,289 @@
...
+            case (.gauge, .double), (.distribution, .double):
+                _value = newValue
+            case (.gauge, .integer(let intValue)):
+                // Convert integer to double
+                SentrySDKLog.warning("Attempted to set an integer value (\(intValue)) on a gauge metric. Converting to double: \(Double(intValue)).")
+                _value = .double(Double(intValue))
+            case (.distribution, .integer(let intValue)):
+                // Convert integer to double
+                SentrySDKLog.warning("Attempted to set an integer value (\(intValue)) on a distribution metric. Converting to double: \(Double(intValue)).")
+                _value = .double(Double(intValue))
```
philippsuess: `m`: That code looks very similar to the switch statement in the init method below. It would be great to extract a method to remove code duplication. I also noticed that this isn't covered by tests, or maybe I'm mistaken.

Sources/Swift/Protocol/SentryMetric.swift
```diff
@@ -0,0 +1,289 @@
...
+    /// Adds or updates an attribute in the metric entry.
+    /// - Parameters:
+    ///   - attribute: The attribute value to add
+    ///   - key: The key for the attribute
+    public mutating func setAttribute(_ attribute: SentryAttribute?, forKey key: String) {
+        if let attribute = attribute {
+            attributes[key] = attribute
+        } else {
+            attributes.removeValue(forKey: key)
+        }
```
philippsuess: `l`: A bit shorter, if I'm not mistaken.
```suggestion
        attributes[key] = attribute
```

Sources/Swift/Protocol/SentryMetric.swift
```diff
@@ -0,0 +1,289 @@
...
+            SentrySDKLog.warning("Distribution metric created with integer value \(intValue). Converting to double: \(Double(intValue)).")
+            self._value = .double(Double(intValue))
+        }
+    }
+    
+    /// Adds or updates an attribute in the metric entry.
+    /// - Parameters:
+    ///   - attribute: The attribute value to add
+    ///   - key: The key for the attribute
+    public mutating func setAttribute(_ attribute: SentryAttribute?, forKey key: String) {
```
philippsuess: `m`: Do we actually need this method? You can set attributes via the property `attributes`. I think we should be able to delete this method.

Sources/Swift/Protocol/SentryMetric.swift
```diff
@@ -0,0 +1,289 @@
...
+    var stringValue: String {
+        switch self {
+        case .counter:
+            return "counter"
+        case .gauge:
+            return "gauge"
+        case .distribution:
+            return "distribution"
+        }
+    }
```
philippsuess: `m`: If you define the enum like this `MetricType : String`, I think we don't need the stringValue. You can get the string value via the `rawType`. Any reason that we can't do this?

Sources/Swift/Protocol/Metric.swift
```diff
@@ -0,0 +1,289 @@
...
+///         // Or use literal syntax: modified.value = intValue + 1
+///     case .double(let doubleValue):
+///         // Can safely modify as double - e.g., for distributions
+///         modified.value = .double(doubleValue * 1.5)
+///         // Or use literal syntax: modified.value = doubleValue * 1.5
+///     }
+///     return modified
+/// }
+/// ```
+public enum MetricValue: Encodable, Equatable, ExpressibleByIntegerLiteral, ExpressibleByFloatLiteral, Hashable {
```
philippsuess: TIL about ExpressibleByIntegerLiteral

Sources/Swift/Protocol/SentryMetric.swift
```diff
@@ -0,0 +1,289 @@
...
+///     }
+///     return modified
+/// }
+/// ```
+public enum MetricValue: Encodable, Equatable, ExpressibleByIntegerLiteral, ExpressibleByFloatLiteral, Hashable {
+    /// A 64-bit signed integer value, typically used for counters.
+    case integer(Int64)
+
+    /// A 64-bit floating point value, typically used for distributions and gauges.
+    case double(Double)
```
philippsuess: @noahsmartin and @itaybre, I would like your input on whether this is a proper Swifty way of doing such an API. It looks OK to me, but I would love your input.

Sources/Swift/Protocol/SentryMetric.swift
philprime: You are correct, this was taken over from SentryLog and not necessary

Sources/Swift/Integrations/Metrics/MetricBatcher.swift
philprime: The MetricBatcher is typed specifically to Metric types, offering a minimal API to work with them. 
To reduce the duplicate implementation between MetricBatcher and LogBatcher, I extracted the generic Batcher (as you are requesting in this comment).
The MetricBatcher and LogBatcher are both wrapper around the generic Batcher with pre-defined configuration in the initializer (see couple of lines below).

So yes, this should already be what you are asking for, happy to discuss this better in a call if it's still unclear

Sources/Swift/Protocol/Metric.swift
philprime: I don't think there's any value in that comment, because if we add it here, Objective-C users won't see it because they don't have access to the Swift type. And if a Swift user sees it, they are not the target group of the comment. I have already added a note to our documentation draft PR: https://github.com/getsentry/sentry-docs/pull/15822

Sources/Swift/Protocol/Metric.swift
philprime: This is done on purpose, because in which scenario would a `gauge` metric suddenly be a `count` metric?

Sources/Swift/Protocol/Metric.swift
philprime: Good catch, we don't need decodable anymore and we don't need encoding public

Sources/Swift/Protocol/Metric.swift
philprime: Especially because counter metrics are integer values and gauge/distribution metrics are doubles

Sources/Swift/Protocol/SentryMetric.swift
philprime: Glad for feedback, to add more context, this enum makes the value type-safe because counter expectes integers while gauge/distribution expects floats
https://develop.sentry.dev/sdk/telemetry/metrics/#metrics-protocol

Sources/Swift/Integrations/Metrics/MetricBatcher.swift
philippsuess: Then it's a naming thing. We could name this The `MetricsBatcherFacade` or `MetricsBatcherAdapter` or we just keep it as is. No strong opinion.

Sources/Swift/Protocol/Metric.swift
philippsuess: Yes, I thought so, it would be great to add a comment.

Sources/Swift/Protocol/SentryMetric.swift
```diff
@@ -0,0 +1,240 @@
...
+            case (.gauge, .integer(let intValue)):
+                // Convert integer to double
+                SentrySDKLog.warning("Attempted to set an integer value (\(intValue)) on a gauge metric. Converting to double: \(Double(intValue)).")
+                _value = .double(Double(intValue))
+            case (.distribution, .integer(let intValue)):
+                // Convert integer to double
+                SentrySDKLog.warning("Attempted to set an integer value (\(intValue)) on a distribution metric. Converting to double: \(Double(intValue)).")
+                _value = .double(Double(intValue))
+            }
+        }
```
philippsuess: `h`: This still feels a bit odd. We're informing the user if they choose a somewhat wrong combination of value and type. It would be great if we could come up with an API that enforces this out of the box.

I would need to spend some time to come up with a proper solution, but something like this could work
 SentryCounterMetric() // only offers an init method accepting integer and double
SenrtryDistributionMetric() // only offers an init method accepting double

So basically enforcing wrong usage and type casting via the API

philprime: So the only case where the user interacts with this metric object is in `beforeSendMetric`. The `metricType` and the `value` need to be tightly coupled. I can do something like this:

```swift
options.beforeSendMetric = { metric in 
  if let counterMetric = metric as? SentryDistributionMetric {
    counterMetric.value = 1.23 // is integer
  } else if let distributionMetric = metric as? SentryDistributionMetric {
    distributionMetric.value = 1.23 // is double
  } else if let gaugeMetric = metric as? SentryGaugeMetric {
    gaugeMetric.value = 1.23 // is double
  } else {
    // should never happen
  }
}
```

The Swift-ier way would be something like this:

```swift
options.beforeSendMetric = { metric in 
  switch metric {
  case .counter(let counterMetric):
    counterMetric.value = 1.23 // is integer  
  case .gauge(let gaugeMetric):
    gaugeMetric.value = 1.23 // is double  
  case .distribution(let distributionMetric):
    distributionMetric.value = 1.23 // is double
  }
}
```

But i can also combine the type and the value:


```swift
options.beforeSendMetric = { metric in 
  // metric has no 'type', it's implied from the 'value'
  metric.value = .counterValue(/* Integer */)
  metric.value = .distributionValue(/* Double */)
  metric.value = .gaugeValue(/* Double */)
}
```

The last one is already quite close to `SentryMetricValue` but basically one more step with the metric type merged in too.

philprime: @noahsmartin @itaybre feel free to raise if you have ideas on the syntax. The goal is not allowing users to use double/float for counter metrics

itaybre: I would say that option 3 looks better to me, it doesn't allow unexpected types

noahsmartin: I'm not sure why we need this type conversion. This is a classic Swift enum w/ associated value right? The counter type would be int, and gauge/distribution would be double. The enum would probably be a part of the public API like:

```
    enum Metric {
        struct Attributes {
            let timestamp: Date
            let name: String
            let traceId: SentryId
        }
        case counter(Int, Attributes)
        case gauge(Double, Attributes)
    }
```

The same can be achieved by making `Metric` a struct as it is in this PR with a generic type for `MetricData` like:

```
structure Metric<MetricData: MetricData> {
let timestamp: Date
let name: String
let traceId: SentryId
...

let value: MetricData
}

protocol SentryMetricData {
  associatedtype MetricDataType
  var type: String
  var data: MetricDataType
}

structure CounterMetric: SentryMetricData {
  var type: String {
    "counter"
  }

  var data: Int
}
```

I think the enum approach is much cleaner. The important thing here is that the field `metricType: SentryMetricType` that is in this PR needs to be coupled to the type of `value`. It can be done by grouping them in a protocol like that or by letting the type be a case in the enum.

philprime: @noahsmartin Yes, we can refactor the `SentryMetric` from a `struct` to a `enum` with an associated value to keep the actual data payload. IIUC that would result in a `beforeSendMetric` like this:

```swift
struct MetricData {
   var timestamp: Date
   var name: String
   var traceId: SentryId
}
enum Metric {
   case counter(Int, MetricData)
   case distribution(Double, MetricData)
}

var beforeSendMetric: (Metric) -> Metric = { metric in
   var metric = metric
   switch metric {
   case .counter(let value, var data):
       // manipulating the associated data
       data.name = "new name"
   case .distribution(let value, var data):
       // Setting a different value means replacing the enum
       metric = .distribution(123, data)
   }
   return metric
}
```

noahsmartin: Yes that looks right to me. Personally I avoid mutable data as much as possible (causes race conditions) so I would instead not even make name mutable and always require replacing the value, but I think that is a more minor detail. Here's another approach:

```
    struct Metric {
        enum Data {
            case counter(Int)
            case gauge(Double)
        }
        let timestamp: Date
        let name: String
        let traceId: SentryId
        let data: Data
    }
```

I inverted the struct/enum, I think it makes a bit more sense that way but again either way is good, up to you. You also might want to consider nested types like in that snippet above, makes it easier than having everything at file scope IMO

philprime: I just did a test and mutating associated values means that the return value must always be set again, so like this:

```swift
let beforeSendMetric: (Metric) -> Metric = { metric in
    var metric = metric
    switch metric {
    case .counter(let value, var data):
        // manipulating the associated data
        data.name = "new name"
        // MUST create the enum again, because associated data is a copy
        metric = .counter(value, data)
    case .distribution(var value, let data):
        // Setting a different value means replacing the enum
        value = 123
        // MUST create the enum again, because associated value is a copy
        metric = .distribution(value, data)
    }
    return metric
}
```

This is terrible UX, because if an user forgets to return a new enum, they will not realize that the metric wasn't actually modified, but our specs say:

> beforeSendMetric/before_send_metric: A function that takes a metric object and returns a metric object. This function is called before sending the metric to Sentry. It can be used to modify the metric object or to prevent the metric from being sent to Sentry. This function is optional.

philprime: With the growing complexity in the `beforeSendMetric`, which is customer-facing, I believe my option 3 where `type` and `value` is combined seems to be the best approach, eventhough it allows changing the type of a metric in the `beforeSendMetric`

```
options.beforeSendMetric = { metric in 
  // metric has no 'type', it's implied from the 'value'
  metric.value = .counterValue(/* Integer */)
  metric.value = .distributionValue(/* Double */)
  metric.value = .gaugeValue(/* Double */)
}
```

@philipphofmann what do you think based on this conversation?

philippsuess: I'm also all for avoiding mutable types, but almost everything in your public API is mutable. People are used to chaning single properties in beforeSend hooks. When we change that, I think we should change it for all public APIs and not just metrics now. 

Users should be able to mess around with the metric as much as they want in beforeSend. This enables plenty of use cases. If we don't allow this, we limit our users.

I'm not too worried about a switch statement in beforeSend. Currently, it's worse IMO. We do the conversions on the setter and change the metric values for our users, so we don't clearly communicate this limitation. In the end, they would end up with a switch statement anyway, I think. If we combine the value and type, it's crystal clear. You don't get any surprises about rounding and such. While this might be OK to do in JS and Python, in which you have to worry about type safety yourself all the time, in languages as Swift, that's uncommon, and we should make types stricter.

So I'm also in favor of the inverted enum, but I think we have to make things var to stay consistend with all other beforeSend hooks. So basically something like this. I would call the enum value and not data, because it's about the metric value:
```swift
struct Metric {

    var timestamp: Date
    var name: String
    var traceId: SentryId
    var value: Value

    enum Value {
        case counter(UInt)
        case gauge(Double)
        case distribution(Double)
    }
}

options.beforeSendMetric = { metric in
    metric.value = .counter(10)
    metric.value = .gauge(20.0)
    metric.value = .distribution(-1.0)
}
```

We also have to keep in mind what happens when we add a new metric type, and with this solution, we simply add a new case to the enum Value.

philprime: Refactored it to combine type and value into an enum with associated data

philprime:
@sentry review
@cursor review

Sources/Swift/Protocol/SentryMetricValue.swift
```diff
@@ -0,0 +1,60 @@
...
+///         modified.value = .counter(intValue + 1)
+///     case .gauge(let doubleValue):
+///         // Can safely modify as double - e.g., for gauges
+///         modified.value = .gauge(doubleValue * 1.5)
+///     case .distribution(let doubleValue):
+///         // Can safely modify as double - e.g., for distributions
+///         modified.value = .distribution(doubleValue * 1.5)
+///     }
+///     return modified
+/// }
```
philippsuess, line 23: Thanks for the snippet

Tests/SentryTests/Integrations/Metrics/SentryMetricsBatcherTests.swift
```diff
@@ -0,0 +1,801 @@
...
+
+        // Assert no additional metrics
+        XCTAssertEqual(capturedMetrics.count, 1)
+    }
+    
+    func testAddMetric_whenDistributionType_shouldCaptureDistribution() throws {
+        // -- Arrange --
+        let metric = createTestMetric(
+            name: "distribution.metric",
+            value: .distribution(42.0),
```
philippsuess: `l`: We could add a few digits after the `.` to ensure the encoding works properly.
```suggestion
            value: .distribution(42.12345),
```

Tests/SentryTests/Integrations/Metrics/SentryMetricsBatcherTests.swift
```diff
@@ -0,0 +1,801 @@
...
+        let capturedMetric = try XCTUnwrap(capturedMetrics.first)
+        XCTAssertEqual(capturedMetric["type"] as? String, "counter")
+        XCTAssertEqual(capturedMetric["value"] as? Int64, 42)
+        XCTAssertEqual(capturedMetric["unit"] as? String, "connection")
+
+        // Assert no additional metrics
+        XCTAssertEqual(capturedMetrics.count, 1)
+    }
+    
+    func testAddMetric_whenDistributionType_shouldCaptureDistribution() throws {
```
philippsuess, line 702: `m`: We could add one test witha negative distribution to ensure encoding negative doubles works correctly. Maybe it's already covered somewhere else. If so, please ignore me.

Tests/SentryTests/Integrations/Metrics/SentryMetricsIntegrationTests.swift
```diff
@@ -21,23 +24,165 @@ class MetricsIntegrationTests: XCTestCase {
...
+
+        // We can not rely on the SentrySDK.flush(), because we are using a test client which is not actually
+        // flushing integrations as of Dec 16, 2025.
+        //
+        // Calling uninstall will flush the data, allowing us to assert the client invocations
+        integration.uninstall()
+
+        // -- Assert --
+        let capturedMetrics = try XCTUnwrap(client.captureMetricsDataInvocations.first)
+        XCTAssertEqual(1, capturedMetrics.count.intValue, "Should capture 1 metric")
```
philippsuess: `l`: We do the actual value first and then the expected. This applies to more tests in this test class. Also the assert message could be a bit more specific. 
```suggestion
        XCTAssertEqual(capturedMetrics.count.intValue, 1, "Should capture 1 metric with invValue 1, but got value \(capturedMetrics.count.intValue)")
```

Tests/SentryTests/Protocol/SentryMetricTests.swift
```diff
@@ -0,0 +1,128 @@
...
+
+    // MARK: - Encoding Tests
+    
+    func testEncode_whenCounterMetric_shouldEncodeCorrectly() throws {
+        // -- Arrange --
+        let metric = SentryMetric(
+            timestamp: testTimestamp,
+            traceId: testTraceId,
+            name: "api.requests",
+            value: .counter(1),
```
philippsuess, line 41: `l`: Adding one test with UInt.max would slightly improve the test coverage to ensure that encoding large Integers works correctly.

Tests/SentryTests/Protocol/SentryMetricTests.swift
```diff
@@ -0,0 +1,128 @@
...
+        )
+        
+        // -- Act --
+        let data = try encodeToJSONData(data: metric)
+        let json = try XCTUnwrap(JSONSerialization.jsonObject(with: data) as? [String: Any])
+        
+        // -- Assert --
+        XCTAssertEqual(json["name"] as? String, "api.response_time")
+        let value = try XCTUnwrap(json["value"] as? Double)
+        XCTAssertEqual(value, 125.5, accuracy: 0.001)
```
philippsuess, line 88: `l`: Testing with more numbers after the `.` would slightly increase the test quality.
```suggestion
        XCTAssertEqual(value, 125.523434, accuracy: 0.001)
```

Tests/SentryTests/Protocol/SentryMetricValueTests.swift
```diff
@@ -0,0 +1,261 @@
...
+        XCTAssertEqual(json["type"] as? String, "counter")
+        XCTAssertEqual(json["value"] as? Int64, 42)
+    }
+    
+    func testEncode_whenCounterWithLargeValue_shouldEncodeAsInt64() throws {
+        // -- Arrange --
+        // Use a value larger than Int64.max (9,223,372,036,854,775,807) but within UInt64 range
+        // to verify truncation behavior
+        let largeValue: UInt = 10_000_000_000_000_000_000 // 10 quintillion
+        let metricValue = SentryMetricValue.counter(largeValue)
```
philippsuess, line 26: Thanks for the test case. I added a few other comments to cover this edge case, you can ignore these then, I guess.

Tests/SentryTests/Integrations/Metrics/SentryMetricsBatcherTests.swift
philippsuess: It could be that this is already covered somewhere else. If so, ignore me.

Tests/SentryTests/Integrations/Metrics/SentryMetricsBatcherTests.swift
philippsuess: I think that's already covered in another unit test. If so, ignore that comment.

Tests/SentryTests/Protocol/SentryMetricTests.swift
philippsuess: That's already covered. Sorry about the comment.

Tests/SentryTests/Protocol/SentryMetricTests.swift
philippsuess: Already covered somewhere else.

Tests/SentryTests/Integrations/Metrics/SentryMetricsBatcherTests.swift
philprime: Yes, it's already covered in SentryMetricValueTests.testEncode_whenDistributionNegative_shouldEncodeCorrectly

Tests/SentryTests/Integrations/Metrics/SentryMetricsBatcherTests.swift
philprime: Thanks, I'll add it anyways

philprime:
Thanks @philipphofmann, I'll keep this PR open and ready-for-merge until #6957 is ready to be merged too (blocked by #7077) so I merge all three PRs directly after another.

Issue #6901, opened 2025-11-25
8.57.3 turns System Extensions (sysext) into UI processes on Mac


CodingMarkus:
### Platform

macOS

### Environment

Production

### Installed

CocoaPods

### Version

8.57.3

### Xcode Version

26.1

### Did it work on previous versions?

Yes, works fine with 8.51.1

### Steps to Reproduce

1. Integrate Sentry into a system extension

### Expected Result

It just works.

### Actual Result

As soon as Sentry is initialized, the system extension shows up in Dock like a UI application. First it bounces for quite a while, then it just sits there an cannot be quit anymore. 

Also it can no longer use system API, e.g. configuring a tunnel interface (network extension) always fails and it will no longer quit when the app is quit, it just hangs forever and can only be force killed in Activity Monitor.

Reverting back to 8.51.1 and everything works fine again. No Dock icon, system API can be used without error, and the system extension quits when the controlling app quits.

philippsuess:
@CodingMarkus, please share your options for starting the SDK and tell us which system extension you're using. That will help us to identify the problem more quickly.

itaybre:
Using Apple's [example network filter app](https://developer.apple.com/documentation/NetworkExtension/filtering-network-traffic), I was able to reproduce the jumping icon in the dock but the extension seems to be working.

<img width="1252" height="332" alt="Image" src="https://github.com/user-attachments/assets/608cbf8f-bf10-4bf2-b2cc-26ff55f955c6" />

I see this happening since [8.53.0](https://github.com/getsentry/sentry-cocoa/compare/8.52.1...8.53.0) (8.52.1 seems fine).
But I haven't been able to find the reason yet

CodingMarkus:
A network filter doesn't try to create a tunnel interface, though. I think the reason why this fails for us is somehow threading or runloop related. It seems like the extension gets stuck somehow and this prevents certain completion callbacks from ever getting executed.

UI apps (those that appear in Dock) require an active runloop on the main thread and a lot of actions are still performed via that runloop. Network extensions are executed in a context that does have a runloop but are not supposed to interact with it, they are supposed to only use GCD. Messing around with the process' runloop can certainly break things.

Also does the extension ever quit for you? Usually it does quit on its own, as soon as the app process dies that triggered it. Not immediately, but shortly after. For me it never quits and also ignores TERM signal; it can only be forcefully killed. This also speaks for some runloop issue as a running runloop is also required to quit apps.

[...]

PR #6961, opened 2025-12-02
chore: Convert `SentryAutoSessionTrackingIntegration` to Swift

itaybre:
Converts `SentryAutoSessionTrackingIntegration` to Swift in preparation for some fixes for System Extensions (see [this](https://github.com/getsentry/sentry-cocoa/issues/6901) issue)

### DIFF
--- a/Sources/Swift/Integrations/Session/SentryAutoSessionTrackingIntegration.swift
+++ b/Sources/Swift/Integrations/Session/SentryAutoSessionTrackingIntegration.swift
@@ -18,10 +18,6 @@ final class SentryAutoSessionTrackingIntegration<Dependencies: SessionTrackerPro
     }
     
     func uninstall() {
-        stop()
-    }
-    
-    func stop() {
         tracker.stop()
     }
     
--- a/Tests/SentryTests/Integrations/Session/SentrySessionGeneratorTests.swift
+++ b/Tests/SentryTests/Integrations/Session/SentrySessionGeneratorTests.swift
@@ -54,7 +54,7 @@ class SentrySessionGeneratorTests: NotificationCenterTestCase {
         fileManager.deleteCrashedSession()
         fileManager.deleteTimestampLastInForeground()
         fileManager.deleteAppState()
-        autoSessionTrackingIntegration.stop()
+        autoSessionTrackingIntegration.uninstall()
     }
     
     /**
@@ -92,7 +92,7 @@ class SentrySessionGeneratorTests: NotificationCenterTestCase {
         for _ in Array(1...amount.crashed) {
             // send crashed session
             crashIntegration.install(with: options)
-            autoSessionTrackingIntegration.stop()
+            autoSessionTrackingIntegration.uninstall()
             autoSessionTrackingIntegration = SentryAutoSessionTrackingIntegration(with: options, dependencies: SentryDependencyContainer.sharedInstance())
             goToForeground()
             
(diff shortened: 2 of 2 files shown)


Sources/Swift/Integrations/Session/SentryAutoSessionTrackingIntegration.swift
```diff
@@ -0,0 +1,31 @@
...
+    let tracker: SessionTracker
+    
+    init?(with options: Options, dependencies: Dependencies) {
+        guard options.enableAutoSessionTracking else { 
+            SentrySDKLog.debug("Not going to enable \(Self.name) because enableAutoSessionTracking is disabled.")
+            return nil 
+        }
+
+        tracker = dependencies.getSessionTracker(with: options)
+        tracker.start()
```
philprime, line 17: `l`: If we have a log message that the integration was skipped, we should consider a message that it was installed.

```suggestion
        tracker.start()
        
        SentrySDKLog.debug("\(Self.name) installed")
```

Tests/SentryProfilerTests/SentryProfilingPublicAPITests.swift
```diff
@@ -57,12 +58,6 @@ class SentryProfilingPublicAPITests: XCTestCase {
 
         givenSdkWithHubButNoClient()
 
-        if let autoSessionTracking = SentrySDKInternal.currentHub().installedIntegrations().first(where: { it in
```
philprime, line 60: `h`: Where is the tear down logic for the integration?

itaybre: I disabled it with `enableAutoSessionTracking`, no need to tear it down manually.

Sources/Swift/Integrations/Session/SentryAutoSessionTrackingIntegration.swift
itaybre: Sounds good, I added it at the `SentrySwiftIntegrationInstaller` level to have the same behavior for all integrations

itaybre: PR: https://github.com/getsentry/sentry-cocoa/pull/6976

Tests/SentryTests/Integrations/Session/SentrySessionGeneratorTests.swift
```diff
@@ -149,8 +149,7 @@ class SentrySessionGeneratorTests: NotificationCenterTestCase {
         crashIntegration = SentryCrashIntegration(crashAdapter: sentryCrash, andDispatchQueueWrapper: TestSentryDispatchQueueWrapper())
         crashIntegration.install(with: options)
         
-        autoSessionTrackingIntegration = SentryAutoSessionTrackingIntegration()
-        autoSessionTrackingIntegration.install(with: options)
+        autoSessionTrackingIntegration = SentryAutoSessionTrackingIntegration(with: options, dependencies: SentryDependencyContainer.sharedInstance())
```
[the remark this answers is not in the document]
itaybre: This was right in the end, the test didn't fail because it was disabled.

But in fact it wasn't a test, but used to generate sessions during testing

Sources/Swift/Integrations/Session/SentryAutoSessionTrackingIntegration.swift
```diff
@@ -0,0 +1,31 @@
...
+    }
+    
+    func uninstall() {
+        stop()
```
noahsmartin: Can we just inline the call to tracker.stop() ? We don't need an extra function

noahsmartin (approved):

Sources/Swift/Integrations/Session/SentryAutoSessionTrackingIntegration.swift
itaybre: Sure, we don't call stop from tests anymore.
So I can do that

PR #6963, opened 2025-12-02
chore: Rename build folder from `Carthage` to `XCFrameworkBuildPath `

### LINKED ISSUE #6964: chore: Rename build folder from `Carthage` to `SentryOutput`
> [!NOTE]
> The pull request "[chore: Rename build folder from `Carthage` to `SentryOutput`](https://github.com/getsentry/sentry-cocoa/pull/6963)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Renames the build folder from `Carthage` to `SentryOutput`.

Nothing should change for users from this change, since it only affects the folder where we extract the xcframeworks.

#skip-changelog

itaybre:
Renames the build folder from `Carthage` to `XCFrameworkBuildPath `.

Nothing should change for users from this change, since it only affects the folder where we extract the xcframeworks.

#skip-changelog

Closes #6964

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,8 @@ playground.xcworkspace
 Packages
 
 # Carthage
+# We don't use Carthage anymore, but we keep this folder to avoid uploading local files by mistake.
+# Will be removed in the future.
 Carthage/Build
 Carthage/Checkouts
 Carthage


itaybre:
`SentryOutput` is pretty much a random name, any one will do, I am open to suggestions

.gitignore
```diff
@@ -116,3 +116,6 @@ analyzer
 
 # Output of snapshot testing
 **/__Snapshots__
+
+# Output folder for Sentry local builds
```
philippsuess, line 122: `m`: We should be able to remove Carthage here as well, I believe.

https://github.com/getsentry/sentry-cocoa/blob/072fada45028a106587279d10802d6a4c771bd9a/.gitignore#L43-L46

.github/workflows/release-upload-xcframework.yml
```diff
@@ -44,4 +44,4 @@ jobs:
           if-no-files-found: error
           overwrite: true
           path: |
-            ${{github.workspace}}/Carthage/*.zip
+            ${{github.workspace}}/SentryOutput/*.zip
```
philippsuess: `m`: Maybe a slightly better name
```suggestion
            ${{github.workspace}}/XCFrameworkBuildPath/*.zip
```

.gitignore
itaybre: Since we may have that folder locally, I left it there to avoid accidentally uploading stuff.
But I can remove it if you have a strong opinion

philippsuess: No strong opinion. Maybe safer to keep it around for a while. It does no harm, but I would then add a comment why we didn't delete it yet. And maybe create a follow up issue in Linear to ditch this in x months.

.github/workflows/release-upload-xcframework.yml
itaybre: Sounds good

.gitignore
itaybre: Sounds good, created it: https://github.com/getsentry/sentry-cocoa/issues/6974

itaybre:
> LGTM, but I think we should trigger an alpha release that we don't publish to ensure the release workflow doesn't break before merging this.

Release worked fine: https://github.com/getsentry/sentry-cocoa/actions/runs/19902408367/job/57052133551

PR #6976, opened 2025-12-03
chore: Add log when a Swift Integration is loaded

### LINKED ISSUE #6977: chore: Add log when a Swift Integration is loaded
> [!NOTE]
> The pull request "[chore: Add log when a Swift Integration is loaded](https://github.com/getsentry/sentry-cocoa/pull/6976)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Added a SDKLog message when a `SwiftIntegration` is installed.

Also added some tests to verify `SentrySwiftIntegrationInstaller` works as expected

#skip-changelog

itaybre:
Added a SDKLog message when a integrations are intalled in `Sources/Sentry/SentryHub.m`

Also added some tests to verify `SentrySwiftIntegrationInstaller` works as expected

#skip-changelog

Closes #6977

### DIFF
--- a/Tests/SentryTests/Integrations/SentrySwiftIntegrationInstallerTests.swift
+++ b/Tests/SentryTests/Integrations/SentrySwiftIntegrationInstallerTests.swift
@@ -4,58 +4,47 @@ import XCTest
 
 final class SentrySwiftIntegrationInstallerTests: XCTestCase {
 
-    private var testHub: TestHub!
-    private var options: Options!
-    private var logOutput: TestLogOutput!
-
-    override func setUp() {
-        super.setUp()
-
-        // Create test hub
-        testHub = TestHub(client: nil, andScope: nil)
-        SentrySDKInternal.setCurrentHub(testHub)
-
-        // Create options
-        options = Options()
-        options.dsn = TestConstants.dsnAsString(username: "SentrySwiftIntegrationInstallerTests")
-        options.debug = true
-
-        // Setup log output
-        logOutput = TestLogOutput()
-        SentrySDKLog.setLogOutput(logOutput)
-        SentrySDKLog.configureLog(true, diagnosticLevel: .debug)
-    }
-
     override func tearDown() {
-        clearTestState()
+        SentrySDKInternal.setCurrentHub(nil)
 
         super.tearDown()
     }
     
     // We are not testing `SwiftAsyncIntegration`, but use it as an example for an installed integration
-    func testInstall_LogsInstalledIntegrations() {
+    func testInstall_AddsInstalledIntegrations() {
         // Arrange
+        let options = Options()
+        options.dsn = TestConstants.dsnAsString(username: "SentrySwiftIntegrationInstallerTests")
+        options.debug = true
         options.swiftAsyncStacktraces = true
+        
+        let testHub = TestHub(client: nil, andScope: nil)
+        SentrySDKInternal.setCurrentHub(testHub)
 
         // Act
         SentrySwiftIntegrationInstaller.install(with: options)
 
         // Assert
-        let logs = logOutput.loggedMessages.joined()
-        XCTAssertTrue(logs.contains("Integration installed: SentrySwiftAsyncIntegration"),
-                     "Should log when SwiftAsyncIntegration is installed")
+        XCTAssertEqual(testHub.installedIntegrationNames().count, 1)
+        XCTAssertEqual(try XCTUnwrap(testHub.installedIntegrationNames().first), "SentrySwiftAsyncIntegration")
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/Integrations/SentrySwiftIntegrationInstallerTests.swift
```diff
@@ -0,0 +1,61 @@
...
+        options.debug = true
+
+        // Setup log output
+        logOutput = TestLogOutput()
+        SentrySDKLog.setLogOutput(logOutput)
+        SentrySDKLog.configureLog(true, diagnosticLevel: .debug)
+    }
+
+    override func tearDown() {
+        clearTestState()
```
philippsuess: `m`: I would really try to avoid using this method if we can. It does so many things. I would rather unset only what is required.

Tests/SentryTests/Integrations/SentrySwiftIntegrationInstallerTests.swift
```diff
@@ -0,0 +1,61 @@
...
+    private var testHub: TestHub!
+    private var options: Options!
+    private var logOutput: TestLogOutput!
+
+    override func setUp() {
+        super.setUp()
+
+        // Create test hub
+        testHub = TestHub(client: nil, andScope: nil)
+        SentrySDKInternal.setCurrentHub(testHub)
```
philippsuess: `m`: The tests below don't even interact with the hub. I think you could remove it as the assert only checks for the log output. Or even better, the assert checks if the hub has the integration installed.

Tests/SentryTests/Integrations/SentrySwiftIntegrationInstallerTests.swift
```diff
@@ -0,0 +1,61 @@
...
+
+        // Create options
+        options = Options()
+        options.dsn = TestConstants.dsnAsString(username: "SentrySwiftIntegrationInstallerTests")
+        options.debug = true
```
philippsuess: `l`: I would rather define the option as local variables in the tests below. It makes the tests easier to read.

Tests/SentryTests/Integrations/SentrySwiftIntegrationInstallerTests.swift
```diff
@@ -0,0 +1,61 @@
+@_spi(Private) @testable import Sentry
+@_spi(Private) import SentryTestUtils
+import XCTest
+
+final class SentrySwiftIntegrationInstallerTests: XCTestCase {
+
+    private var testHub: TestHub!
+    private var options: Options!
+    private var logOutput: TestLogOutput!
```
philippsuess: `m`: I'm not a real fan of  using `!`, cause it can break without proper error messages. I would rather define the variables locally in the tests below.

Sources/Swift/Core/Integrations/Integrations.swift
```diff
@@ -44,6 +44,7 @@ private struct AnyIntegration {
             guard let integration = anyIntegration.install(options, dependencies) else { return }
 
             SentrySDKInternal.currentHub().addInstalledIntegration(integration, name: anyIntegration.name)
+            SentrySDKLog.debug("Integration installed: \(anyIntegration.name)")
```
philippsuess: `m`: We now have a very similar log messages in two different places. Here and in https://github.com/getsentry/sentry-cocoa/blob/b2295d4e405a79560ed2e630f71413ce12ba18bc/Sources/Sentry/SentrySDKInternal.m#L565

I would rather add one log message here to remove duplication https://github.com/getsentry/sentry-cocoa/blob/b2295d4e405a79560ed2e630f71413ce12ba18bc/Sources/Sentry/SentryHub.m#L654-L660

philipphofmann (commented):
Thanks for adding the tests and a log message. I left a few comments.

Tests/SentryTests/Integrations/SentrySwiftIntegrationInstallerTests.swift
itaybre: `SentrySDKInternal.currentHub().addInstalledIntegration` is called internally.

Improved the test to make it more clear and verify that call

Tests/SentryTests/Integrations/SentrySwiftIntegrationInstallerTests.swift
itaybre: Updated

Sources/Swift/Core/Integrations/Integrations.swift
itaybre: Done, moved the log to the common place and added tests there

PR #6985, opened 2025-12-04
ref: Remove dependency container singleton from SentryAutoSessionTrac…

### LINKED ISSUE #6986: ref: Remove dependency container singleton from SentryAutoSessionTrac…
> [!NOTE]
> The pull request "[ref: Remove dependency container singleton from SentryAutoSessionTrac…](https://github.com/getsentry/sentry-cocoa/pull/6985)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

This integration class was pretty small, and was the only thing keeping `SessionTracker` objc compatible. So I converted it to a Swift integration which removes the use of the dependency container singleton.

#skip-changelog

noahsmartin:
This integration class was pretty small, and was the only thing keeping `SessionTracker` objc compatible. So I converted it to a Swift integration which removes the use of the dependency container singleton.

#skip-changelog

Closes #6986


itaybre (commented):
I already did it [here](https://github.com/getsentry/sentry-cocoa/pull/6985/files)

PR #6987, opened 2025-12-04
ref: Adding missing final attribute

### LINKED ISSUE #6988: ref: Adding missing final attribute
> [!NOTE]
> The pull request "[ref: Adding missing final attribute](https://github.com/getsentry/sentry-cocoa/pull/6987)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Found a few places where final wasn't used and added it

#skip-changelog

noahsmartin:
Found a few places where final wasn't used and added it

#skip-changelog

Closes #6988

### DIFF
--- a/Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebOptionsEvent.swift
+++ b/Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebOptionsEvent.swift
@@ -1,7 +1,7 @@
 @_implementationOnly import _SentryPrivate
 import Foundation
 
-@objc class SentryRRWebOptionsEvent: SentryRRWebCustomEvent {
+final class SentryRRWebOptionsEvent: SentryRRWebCustomEvent {
     
     init(timestamp: Date, customOptions: [String: Any]) {
         super.init(timestamp: timestamp, tag: "options", payload: customOptions)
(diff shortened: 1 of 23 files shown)


Sources/Swift/Integrations/SessionReplay/RRWeb/SentryRRWebOptionsEvent.swift
```diff
@@ -1,7 +1,7 @@
 @_implementationOnly import _SentryPrivate
 import Foundation
 
-@objc class SentryRRWebOptionsEvent: SentryRRWebCustomEvent {
+final class SentryRRWebOptionsEvent: SentryRRWebCustomEvent {
```
philippsuess, line 4: `l`: Can we really remove the `@objc` here?
```suggestion
@objc final class SentryRRWebOptionsEvent: SentryRRWebCustomEvent {
```

noahsmartin: Yeah it wasn't used from objc anywhere, it also wasn't `public` so it never wouldn't have worked with objc anyways

noahsmartin:
@philipphofmann I was thinking about a lint rule as well, but not sure if there is a good way for the linter to detect if a class is overridden somewhere or not (to know if final can be used without a compiler error)

PR #6989, opened 2025-12-04
fix: Development builds with prebuilt binary

### LINKED ISSUE #6990: fix: Development builds with prebuilt binary
> [!NOTE]
> The pull request "[fix: Development builds with prebuilt binary](https://github.com/getsentry/sentry-cocoa/pull/6989)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

The only way this would ever report debug is if the code was built with the DEBUG preprocessor flag. But we support building from a pre-built binary which is pre-built in release mode (I hope) so this would not work. A more reliable way is to check the entitlements to see if it's a debug build.

#skip-changelog

noahsmartin:
The only way this would ever report debug is if the code was built with the DEBUG preprocessor flag. But we support building from a pre-built binary which is pre-built in release mode (I hope) so this would not work. A more reliable way is to check the entitlements to see if it's a debug build.

#skip-changelog

Closes #6990

### DIFF
--- a/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
+++ b/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
@@ -469,8 +469,11 @@ getBuildType(void)
         return "debug";
     }
     SentryMobileProvisionParser *parser = [[SentryMobileProvisionParser alloc] init];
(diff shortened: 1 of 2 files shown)


noahsmartin:
Closing this for now to clean up the open draft PRs that I'm not expecting to get back to soon

PR #6991, opened 2025-12-04
ref: Pass reachability to init for HTTPTransport

### LINKED ISSUE #6992: ref: Pass reachability to init for HTTPTransport
> [!NOTE]
> The pull request "[ref: Pass reachability to init for HTTPTransport](https://github.com/getsentry/sentry-cocoa/pull/6991)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Instead of retrieving the reachability class via the DependencyContainer, pass it in via the init method.

#skip-changelog

philipphofmann:
Instead of retrieving the reachability class via the DependencyContainer, pass it in via the init method.

#skip-changelog

Closes #6992

### DIFF
--- a/develop-docs/DECISIONS.md
+++ b/develop-docs/DECISIONS.md
@@ -491,3 +491,81 @@ We also decided to streamline our integrations by reducing the number of package
 Related:
 
 - Fix for multiple files in Releases (stale for months): https://github.com/Carthage/Carthage/pull/3398
+
+## 3rd Party Library Integrations
+
+### Background
+
+SPM downloads all declared dependencies in a package, even if none of the actually added modules use them. If `sentry-cocoa` declares dependencies like **CocoaLumberjack** or **SwiftLog**, all downstream consumers download these libraries, even if they don't use the corresponding integrations.
+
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentryHttpTransport.m
```diff
@@ -63,6 +64,7 @@ - (id)initWithDsn:(SentryDsn *)dsn
                  rateLimits:(id<SentryRateLimits>)rateLimits
           envelopeRateLimit:(SentryEnvelopeRateLimit *)envelopeRateLimit
        dispatchQueueWrapper:(SentryDispatchQueueWrapper *)dispatchQueueWrapper
+               reachability:(SentryReachability *)reachability
```
itaybre, line 67: Not for now, but if we need 11 parameters in the initializer, we might have too many responsibilities in this class

philippsuess: > we might have too many responsibilities in this class

Good point. We definitely have too many responsibilities in there 😅

PR #6996, opened 2025-12-05
chore: Update cocoapods to use prebuild xcframework

itaybre:
This PR attempts to update Cocoapods to use our prebuild xcframework.

Because we have some special headers for HybridSDK only, we cannot use the default http source option and have to manually download the built version.
This might change since this is just a draft.

For now this only applies to Sentry, it does not apply to SentrySwiftUI


itaybre:
Closing as not planned

Issue #6999, opened 2025-12-05
User ID not falling back to Installation ID


Maschina:
### Platform

macOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

9.0.0

### Xcode Version

26.1.1

### Steps to Reproduce

1. Initialize SDK:

```swift
SentrySDK.start { options in
	options.dsn = "..."
	// Enable logs to be sent to Sentry
	options.enableLogs = true

	#if DEBUG
	options.environment = "development"
	#else
	options.environment = "production"
	#endif

	// Adds IP for users.
	// For more information, visit: https://docs.sentry.io/platforms/apple/data-management/data-collected/
	options.sendDefaultPii = true
	
	options.enableUncaughtNSExceptionReporting = true

	// automatically captures HTTP client errors, like bad response codes, as error events and reports them to Sentry.
	options.enableCaptureFailedRequests = true
}

let user = User() // Don’t set user.userId – SDK should fall back to installationId
SentrySDK.setUser(user)
```

2. Make some logs

### Expected Result

In the logs dashboard, I am expecting to see the user data payload for each log, but I am not seeing any information with regards to user, unless I am explicitly specifying `userId` during initialization

### Actual Result

Here a screenshot of what I am seeing for any log in my dashboard:

<img width="2458" height="420" alt="Image" src="https://github.com/user-attachments/assets/9c7ac6b5-5cb3-4d10-bbaa-7ba9161c48a6" />

### Are you willing to submit a PR?

Maybe

itaybre:
Hey @Maschina, thanks for submitting the report.

This seems to be intended behavior, when an user is set, we do not set anything (that's also the same for Events).

But it seems like there is a related bug, we do not set `user.id` when no user is set.
I am fixing that one as I write this comment.

Maschina:
Oh interesting, I was a bit mislead by the [doc](https://docs.sentry.io/platforms/apple/guides/macos/enriching-events/identify-user/) saying, "If you don't provide a userId, the SDK falls back to installationId, which the SDK randomly generates once during an app's installation."

I wasn't aware that there is a difference between manually setting the user with a custom instance, or rely on the default values.

philippsuess:
FYI, we need some userID mostly for Release Health statistics. If all users are `nil`, we can't give you numbers on how many users have updated to a specific version, or how many users are crash-free. This is a fallback.

[...]

PR #7005, opened 2025-12-08
fix: Fixes `user.id` not set to installationId if no user is set

itaybre:
After reviewing a related user report (#6999), I noticed we are not setting the user.id on logs when no user was set.

This is the same behavior than in Java: https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/logger/LoggerApi.java#L275

### DIFF
--- a/Tests/SentryTests/SentryLogBatcherTests.swift
+++ b/Tests/SentryTests/SentryLogBatcherTests.swift
@@ -414,8 +414,27 @@ final class SentryLogBatcherTests: XCTestCase {
         XCTAssertNil(attributes["user.email"])
     }
     
+    func testAddLog_NoUserAtributesAreSetIfInstallationIdIsNotCached() throws {
+        // No user set on scope
+        // InstallationId not cached
+        
+        let log = createTestLog(body: "Test log message without user")
+        sut.addLog(log, scope: scope)
+        sut.captureLogs()
+        
+        let capturedLogs = testDelegate.getCapturedLogs()
+        let capturedLog = try XCTUnwrap(capturedLogs.first)
+        let attributes = capturedLog.attributes
+        
+        XCTAssertNil(attributes["user.id"])
+        XCTAssertNil(attributes["user.name"])
+        XCTAssertNil(attributes["user.email"])
+    }
+    
     func testAddLog_OnlySetsUserIdToInstallationIdWhenNoUserIsSet() throws {
         // No user set on scope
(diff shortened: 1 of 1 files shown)


Sources/Swift/Tools/SentryLogBatcher.swift
```diff
@@ -195,6 +196,16 @@ import Foundation
...
     }
+    
+    private func addDefaultUserIdIfNeeded(to attributes: inout [String: SentryLog.Attribute], scope: Scope, options: Options) {
+        guard attributes["user.id"] == nil && attributes["user.name"] == nil && attributes["user.email"] == nil else {
+            return
+        }
+        
+        // We only want to set the id if the customer didn't set a user so we at least set something to
+        // identify the user.
+        attributes["user.id"] = .init(value: SentryInstallation.id(withCacheDirectoryPath: options.cacheDirectoryPath))
```
noahsmartin: This is going to do file I/o, are we sure it's called off the main thread?

itaybre: This is executed in the calling thread, but there should not be any file i/o because we cache the installation id, see https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentryClient.m#L125

noahsmartin: It looks like the method class header has a separate function the only reads from the cache, can we switch to that? That will make it easier to ensure we don’t cause hangs and work with structured concurrency in the future

itaybre: Added a cached only method that will never read from the file system

Sources/Swift/Tools/SentryLogBatcher.swift
```diff
@@ -195,6 +196,18 @@ import Foundation
...
+        
+        if let installationId = SentryInstallation.cachedId(withCacheDirectoryPath: options.cacheDirectoryPath) {
+            // We only want to set the id if the customer didn't set a user so we at least set something to
+            // identify the user.
+            attributes["user.id"] = .init(value: installationId)
+        }
+    }
 
     // Only ever call this from the serial dispatch queue.
     private func encodeAndBuffer(log: SentryLog) {
```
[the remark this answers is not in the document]
itaybre: This is known, but the ID should be cached when initializing the SDK

PR #7012, opened 2025-12-08
chore: Use at least xcode 16 for all jobs

### LINKED ISSUE #7033: chore: Use at least xcode 16 for all jobs
> [!NOTE]
> The pull request "[chore: Use at least xcode 16 for all jobs](https://github.com/getsentry/sentry-cocoa/pull/7012)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Bumps the version of Xcode used to at least 16 (16.2 on macOS 14) so we can use newer Xcode 16's features.

Also cleanups some steps and parameters that aren't neeed anymore.

itaybre:
Bumps the version of Xcode used to at least 16 (16.2 on macOS 14) so we can use newer Xcode 16's features.

Also cleanups some steps and parameters that aren't neeed anymore.

Closes #7033


.github/workflows/ui-tests-common.yml
```diff
@@ -89,14 +89,6 @@ jobs:
         with:
           bundler-cache: true
 
-      - name: Install old xcodegen
-        if: ${{ inputs.xcode_version == '14.3.1' || inputs.xcode_version == '15.4' }}
```
itaybre, line 93: This is not needed anymore

.github/workflows/ui-tests-common.yml
```diff
@@ -41,21 +41,11 @@ on:
...
       device:
         description: "Device"
         required: false
         default: ""
         type: string
-      install_platforms:
-        description: "Whether to install the platforms"
-        required: false
-        default: false
-        type: boolean
```
itaybre, line 58: We don't need to install new platforms nor create simulators anymore

philprime (approved):
LGTM

PR #7017, opened 2025-12-09
refactor(logs): Add generic internal item batcher

### LINKED ISSUE #7013: [Metrics] Extract generic item batcher from SentryLogBatcher and use it for SentryMetricBatcher
Both of the implementations are very similar and can reuse the same internal batcher logic. We will implement it as an Swift-only generic type using protocols.

philprime:
## Overview

This PR refactors the `SentryLogBatcher` implementation by extracting its core batching logic into a generic, reusable `Batcher` component using **Protocol-Oriented Programming (POP)**. This abstraction enables future reuse for other item types (e.g., feedback items) while maintaining type safety through Swift generics.

### Protocol-Oriented Architecture

The refactoring uses Protocol-Oriented Programming with the following protocol hierarchy:

- **`BatcherProtocol`**: Defines the public interface for batching operations
- **`BatcherItem`**: Protocol for items that can be batched (must be `Encodable`)
- **`BatcherScope`**: Protocol for scope objects that provide context (attributes, trace IDs, user info)
- **`BatcherConfig`**: Protocol for batcher configuration (timeouts, limits, callbacks)
- **`BatchStorage`**: Protocol for storage backends (in-memory, file-based, etc.)

The generic `Batcher<Storage, Item, Scope>` class implements `BatcherProtocol` and works with any types conforming to these protocols, enabling code reuse without sacrificing type safety.

### Implementation Details

1. **Generic Batcher Implementation** (`Sources/Swift/Tools/Batcher/Batcher.swift`)
   - Thread-safe batching with configurable flush triggers (item count, buffer size, timeout)
   - Scope application happens synchronously; encoding/buffering happens asynchronously
   - Proper deadlock prevention documentation for `capture()` method

2. **Scope Attribute Enrichment** (`Sources/Swift/Tools/Batcher/BatcherScope.swift`)
   - Protocol extension provides default implementation for applying scope attributes
   - Adds SDK metadata, OS/device info, user attributes, replay IDs, and custom scope attributes
   - Non-overriding behavior: scope attributes don't override existing item attributes

3. **Storage Abstraction** (`Sources/Swift/Tools/Batcher/InMemoryBatchStorage.swift`)
   - Protocol-based storage allows swapping implementations (in-memory, file-based, etc.)
   - Current implementation uses in-memory storage with JSON encoding

4. **Type Migration**
   - Moved `SentryLog.Attribute` to `SentryAttribute` as a public type
   - Added backwards compatibility header for `SentryLogAttribute` to maintain Hybrid SDK compatibility

### Bug Fixes

- Fixed `TestCurrentDateProvider.getAbsoluteTime()` to properly update `_absoluteTime` in advance methods
- Fixed callback to pass item count instead of byte size (`batchStorage.count` vs `batchStorage.size`)
- Enhanced documentation for execution model and deadlock prevention
- Added comprehensive test coverage for scope attributes

### Testing

- Added extensive unit tests for `Batcher`, `BatcherScope`, and `InMemoryBatchStorage`
- Test coverage includes edge cases, thread safety, and protocol conformance
- All existing `SentryLogBatcher` tests continue to pass

## Backwards Compatibility

- Public API remains unchanged
- `SentryLogBatcher` maintains the same interface
- Hybrid SDKs unaffected due to compatibility header

## Related Issues

Closes #7013

#skip-changelog

### DIFF
--- a/Sources/Swift/Tools/Batcher/InMemoryBatchBuffer.swift
+++ b/Sources/Swift/Tools/Batcher/InMemoryBatchBuffer.swift
@@ -2,11 +2,15 @@ struct InMemoryBatchBuffer<Item: Encodable>: BatchBuffer {
     private var elements: [Data] = []
     var itemsDataSize: Int = 0
 
+    private let encoder: JSONEncoder = {
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = .secondsSince1970
+        return encoder
+    }()
+
     init() {}
 
     mutating func append(_ item: Item) throws {
-        let encoder = JSONEncoder()
-        encoder.dateEncodingStrategy = .secondsSince1970
         let encoded = try encoder.encode(item)
         elements.append(encoded)
         itemsDataSize += encoded.count


philprime:
Moving back to draft to explore protocol-based implementation.

Also I noticed that my implementation changed the requirements of `options` not changing after the batcher is created. Need to double-check if that is an issue.

philprime:
Further resources to consider:

- [ ] https://www.emergetools.com/blog/posts/SwiftProtocolConformance
- [x] https://nonstrict.eu/wwdcindex/wwdc2015/408/?t=2739

Sources/Swift/Tools/SentryItemBatcher.swift
```diff
@@ -0,0 +1,252 @@
...
+    ///   - flushTimeout: The timeout interval after which buffered items will be flushed
+    ///   - maxItemCount: Maximum number of items to batch before triggering an immediate flush.
+    ///   - maxBufferSizeBytes: The maximum buffer size in bytes before triggering an immediate flush
+    ///   - dispatchQueue: A **serial** dispatch queue wrapper for thread-safe access to mutable state
+    ///
+    /// - Important: The `dispatchQueue` parameter MUST be a serial queue to ensure thread safety.
+    ///              Passing a concurrent queue will result in undefined behavior and potential data races.
+    ///
+    /// - Note: Items are flushed when either `maxItemCount` or `maxBufferSizeBytes` limit is reached.
+    @_spi(Private) public init(
```
noahsmartin: We shouldn't need spi(Private) if it's not @objc

philprime:
@sentry review

AGENTS.md
```diff
@@ -37,6 +37,65 @@ This file provides comprehensive guidance for AI coding agents working with the
 - Fix any test or type errors until the whole suite is green.
 - Add or update tests for the code you change, even if nobody asked.
 
+#### Test Naming Convention
```
itaybre, line 40: Looks good, but also, somewhat unrelated to the PR change

Sources/Sentry/Public/SentryDefines.h
```diff
@@ -65,13 +65,20 @@
...
 @class SentrySamplingContext;
 @class SentryUserFeedbackConfiguration;
 @class SentryLog;
 @protocol SentrySpan;
 
+// Compatibility alias to maintain backward compatibility with existing Objective-C code.
+// SentryLogAttribute is an alias for SentryAttribute, allowing code like
+// [[SentryLogAttribute alloc] initWithString:...] to continue working, after `SentryLog.Attribute`
+// was renamed to `SentryAttribute`.
+@compatibility_alias SentryLogAttribute SentryAttribute;
```
itaybre, line 80: Nice

AGENTS.md
philprime: Yes, I can put that into another PR

philprime: https://github.com/getsentry/sentry-cocoa/pull/7036

philprime:
~BLOCKED BY https://github.com/getsentry/sentry-cocoa/pull/7036~

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,160 @@
...
+        let maxBufferSizeBytes: Int
+        let beforeSendItem: ((Item) -> Item?)?
+        let getInstallationId: () -> String?
+
+        var capturedDataCallback: (Data, Int) -> Void = { _, _ in }
+    }
+
+    private let config: Config
+
+    private var batchStorage: Storage
```
denrase: Could `batchStorage` also be `let` instead of `var`?. We could also name this `storage` instead of `batchStorage`, but thats just a nit.

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /// Encodes and buffers an item, triggering a flush if limits are reached.
+    ///
+    /// - Important: Only call this method from the serial dispatch queue to ensure thread safety.
+    /// - Parameter item: The item to encode and add to the buffer
+    private func encodeAndBuffer(item: Item) {
+        do {
+            let encodedItemsWereEmpty = batchStorage.size == 0
+            try batchStorage.append(item)
```
denrase: I was thinking, what if a user adds an item that is already larger than the buffer, or worse, larger than the 3MB backend limit. We'd add it to the transport, and would then fail to send, as the transport would just re-try.

This has nothing to do with this refactor, as the same behaviour was present before. Also, this will be resolved once the re-try logic is fixed. I was just wondering if we should do some additional measures here. WDYT? Doesn't have to be in this PR.

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,160 @@
...
+        defer {
+            batchStorage.flush()
+        }
+        
+        // Reset timer state
+        timerWorkItem?.cancel()
+        timerWorkItem = nil
+
+        // Fetch and send any available data
+        guard batchStorage.size > 0 else {
```
denrase: We are still checking size here, while checking count below. Shouldn't we also check count, as it represents the number of items that need to be send? Technically it comes out to the same, but it would be more consistent to check the same property as below IMHO.

Sources/Swift/Tools/Batcher/BatcherScope.swift
```diff
@@ -0,0 +1,109 @@
...
+    var span: Span? { get }
+    var userObject: User? { get }
+    func getContextForKey(_ key: String) -> [String: Any]?
+    var attributes: [String: Any] { get }
+
+    func applyToItem<Item: BatcherItem, Config: BatcherConfig<Item>>(_ item: inout Item, config: Config)
+}
+
+extension BatcherScope {
+    func applyToItem<Item: BatcherItem, Config: BatcherConfig<Item>>(_ item: inout Item, config: Config) {
```
denrase: Love how this is extracted, was bugging me to have all this responsibility in the batcher. ❤️

Sources/Swift/Tools/Batcher/BatchStorage.swift
```diff
@@ -0,0 +1,10 @@
+protocol BatchStorage<Element> {
+    associatedtype Element
+
+    mutating func append(_ element: Element) throws
+    mutating func flush()
+
+    var data: Data { get }
+    var count: Int { get }
```
denrase: Maybe we could rename count to `elementCount` to further reduce this risk of confusing it with `size` in the future.

Sources/Swift/Tools/Batcher/BatchStorage.swift
```diff
@@ -0,0 +1,10 @@
+protocol BatchStorage<Element> {
```
denrase: Naming wise, would `Item` instead of `Element` conflict with something? Would be more in line with the naming in the batcher, as well as with the encoded JSON.

Sources/Swift/Tools/Batcher/InMemoryBatchStorage.swift
```diff
@@ -0,0 +1,25 @@
...
+    mutating func flush() {
+        elements.removeAll()
+        size = 0
+    }
+
+    var count: Int {
+        elements.count
+    }
+
+    var data: Data {
```
denrase: It's kinda opaque that we are encoding the data here when we are using a computed property named `data`. Do you think having a `encode` or `encodeItems` method instead would better represent what is being done here?

Sources/Swift/Tools/SentryLogBatcher.swift
```diff
@@ -67,204 +57,55 @@ import Foundation
...
+                },
+                capturedDataCallback: { [weak delegate] data, count in
+                    guard let delegate else {
+                        SentrySDKLog.debug("SentryLogBatcher: Delegate not set, not capturing logs data.")
+                        return
+                    }
+                    delegate.capture(logsData: data as NSData, count: NSNumber(value: count))
+                }
+            ),
+            batchStorage: InMemoryBatchStorage(),
```
denrase: Very nice ❤️

Sources/Sentry/Public/SentryDefines.h
philippsuess: Thanks for the comment.

Sources/Swift/Core/Helper/SentryCurrentDateProvider.swift
```diff
@@ -10,6 +10,13 @@ import Foundation
     func timezoneOffset() -> Int
     func systemTime() -> UInt64
     func systemUptime() -> TimeInterval
+    func getAbsoluteTime() -> UInt64
+}
+
+extension SentryCurrentDateProvider {
+    public func getAbsoluteTime() -> UInt64 {
+        clock_gettime_nsec_np(CLOCK_UPTIME_RAW)
+    }
```
philippsuess, line 19: `m`: I know that this is the "new" way of doing this with protocols, but now we have `getAbsoluteTime` duplicated in the extension and the `SentryDefaultCurrentDateProvider`. I wonder if it actually makes sense then do have this extension.

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,160 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+
+protocol BatcherProtocol<Item, Scope>: AnyObject {
```
philippsuess, line 4: `m`: We renamed the batcher to buffer. FYI, the batcher spec is [deprecated](https://develop.sentry.dev/sdk/telemetry/telemetry-buffer/batch-processor/). You can also follow up in another PR, or change it now.

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,160 @@
...
+        let environment: String
+        let releaseName: String?
+        let flushTimeout: TimeInterval
+        let maxItemCount: Int
+        let maxBufferSizeBytes: Int
+        let beforeSendItem: ((Item) -> Item?)?
+        let getInstallationId: () -> String?
+
+        var capturedDataCallback: (Data, Int) -> Void = { _, _ in }
+    }
```
philippsuess, line 21: `l`: What about split up the config into two structs. The `environment`, `releaseName`, and `getInstallationId` aren't config-related, really. These are properties for the attributes. For me a config controls the behaviour of a class and it shouldn't really contain meta data for attributes. They are only used when applying the attributes: https://github.com/getsentry/sentry-cocoa/blob/a217014241855efbff7f4b6ea1f99f673e8266ae/Sources/Swift/Tools/Batcher/BatcherScope.swift#L15-L25

Sources/Swift/Tools/Batcher/BatchStorage.swift
philprime: Not really, I was thinking the same but I tried to align with Array using Element. But I was considering it too, so I'll go for it for consistency

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,160 @@
...
+
+final class Batcher<Storage: BatchStorage<Item>, Item: BatcherItem, Scope: BatcherScope>: BatcherProtocol {
+    struct Config: BatcherConfig {
+        let environment: String
+        let releaseName: String?
+        let flushTimeout: TimeInterval
+        let maxItemCount: Int
+        let maxBufferSizeBytes: Int
+        let beforeSendItem: ((Item) -> Item?)?
+        let getInstallationId: () -> String?
```
philippsuess: `m`: What drove you to make this a lambda method? Can't we just pass the installationID as a string?

Sources/Swift/Tools/Batcher/Batcher.swift
philprime: I'll rename it to storage. I must be a `var` because it's a data type (`struct`), therefore adding new items is a mutation conflicting with `let`

Sources/Swift/Tools/Batcher/BatchStorage.swift
philippsuess: `m`: I know on Go we call this a storage now, but it's not a real storage, IMO. We're most likely going to call this Buffer on most SDKs. The concept is still WIP, but we already agreed that Storage isn't the best name. You an also change the name later if you want https://www.notion.so/sentry/Telemetry-Processor-2b58b10e4b5d805e819be97fbfc0cd53

Sources/Swift/Tools/Batcher/InMemoryBatchStorage.swift
philprime: I'll add more documentation and rename it to `encodedBatchData` so we know it's returning not the encoded items alone but in the batch format

Sources/Swift/Tools/Batcher/BatchStorage.swift
```diff
@@ -0,0 +1,10 @@
...
+    associatedtype Element
+
+    mutating func append(_ element: Element) throws
```
philippsuess: `l`: I think we're usually adding to the buffers. Append kind of suggests we add it to the end of the buffer or something, but that's not always the case. What if we use a ring buffer internally. 
```suggestion
    mutating func add(_ element: Element) throws
```

Sources/Swift/Tools/Batcher/BatchStorage.swift
```diff
@@ -0,0 +1,10 @@
...
+    mutating func flush()
+
+    var data: Data { get }
```
philippsuess: `l`: For the spans buffer most likely we're not serializng the elements when adding it, because we need a buffer that stores the raw spans, because we need to compute the DCS as late as possible, so most likely the main buffer interface/protocolk won't expose the `Data`, but we can also fix this once we add the span buffer.

Sources/Swift/Tools/Batcher/BatchStorage.swift
```diff
@@ -0,0 +1,10 @@
+protocol BatchStorage<Element> {
+    associatedtype Element
```
philippsuess: `m`: The `BatcherProtocol` uses the term `Item`. Here we have `Element`. We should align on the same term.

Sources/Swift/Tools/Batcher/InMemoryBatchStorage.swift
```diff
@@ -0,0 +1,25 @@
...
+    mutating func append(_ element: Element) throws {
+        let encoded = try JSONEncoder().encode(element)
+        elements.append(encoded)
+        size += encoded.count
+    }
+
+    mutating func flush() {
+        elements.removeAll()
+        size = 0
+    }
```
philippsuess: `m`: The `flush` method name is a bit misleading. This method doesn't flush; it clears the storage. Either we rename it to something like `clear` or we put the flushing logic in here.

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,160 @@
...
+    ///
+    /// - Important: Only call this method from the serial dispatch queue to ensure thread safety.
+    /// - Parameter item: The item to encode and add to the buffer
+    private func encodeAndBuffer(item: Item) {
+        do {
+            let encodedItemsWereEmpty = storage.itemsDataSize == 0
+            try storage.append(item)
+
+            // Flush when we reach max item count or max buffer size
+            if storage.itemsCount >= config.maxItemCount || storage.itemsDataSize >= config.maxBufferSizeBytes {
```
philippsuess: `h`: We're going to have different types of buffers in the future. The span buffer will have buckets per traceID and have a different condition for flushing. Ideally, we should move responsibility for flushing to the storage/buffer. They should know themselves when they are full and they need to flush. Sorry about this, we're just drafting this in here https://www.notion.so/sentry/Telemetry-Processor-2b58b10e4b5d805e819be97fbfc0cd53.

So basically, this Batcher class, which should be called Buffer, knows all different types of buffers and forwards the data items to these buffers. Each buffer knows when to flush and flushes itself when it needs to flush.

As the concept is still in draft, I think it makes sense to discuss this on a call before making changes to ensure the concept actually works in our scenario here.

We could also move forward with this and adapt it later.

Sources/Swift/Tools/Batcher/Batcher.swift
philprime: At this time the batcher is not applying any constraints on added data, so it might make sense to handle it in the SentryLogBatcher instead, because the limits could be different per data point (logs vs metrics).
It might make sense to add a configurable "BatcherValidator" at some point so we can inject the logic

Sources/Swift/Tools/Batcher/BatchStorage.swift
philprime: Aren't items still in order when accessing them as serialized data?

Sources/Swift/Tools/Batcher/Batcher.swift
philprime: Because the field was set using `SentryInstallation.cacheId(...)` [here](https://github.com/getsentry/sentry-cocoa/blob/47a168453ec80c9eb5fe8511ba3ebdfc9cceb7b9/Sources/Swift/Tools/SentryLogBatcher.swift#L204-L208) and there is no documentation or contract in place to ensure that this method always responds with the same value, so theoretically even if the `options` don't change over time, the `installationId` could. 

So this is mostly defensive programming, but logically the installation id doesn't change, so I am fine to change it a fixed value

Sources/Swift/Tools/Batcher/Batcher.swift
philprime: Makes sense, I also believe having so many fields in there is a code smell

Sources/Swift/Tools/Batcher/Batcher.swift
philprime: My intention is not implementing the telemetry buffer spec, I am just refactoring the `SentryLogBatcher` into a generic `Batcher` so I can use it for metrics. It would be nice if we could postpone this to another PR, but we'll align later today in a call to discuss this anyways

Sources/Swift/Core/Helper/SentryCurrentDateProvider.swift
philprime: There are a couple of reasons why I added this here even though it's a duplicate:

- I want to fake the time returned by `getAbsoluteTime()` in tests, but I can't do that in the static method → new method for the protocol `SentryCurrentDateProvider`
- I added it as the extension because that's the default implementation.
- I did not use `SentryDefaultCurrentDateProvider.getAbsoluteTime()` in the extension, because I did not want the protocol to rely on a type implementing the protocol.
- I can not remove the `public func getAbsoluteTime()` from the `SentryDefaultCurrentDateProvider` because it fails with `Non-'@objc' method 'getAbsoluteTime()' does not satisfy requirement of '@objc' protocol 'SentryCurrentDateProvider'` because it's an Objective-C compatible type and protocol extensions are unavailable to Objective-C.

Sources/Swift/Tools/Batcher/Batcher.swift
philippsuess: Anyway, this PR brings us a huge step closer to the TelemetryBuffer. As these APIs are internal, we can follow up.

Sources/Swift/Tools/Batcher/Batcher.swift
philippsuess: Postponing is fine 👍 . The proposal isn't final yet.

Sources/Swift/Core/Helper/SentryCurrentDateProvider.swift
philippsuess: Then all good thanks

Sources/Swift/Tools/Batcher/Batcher.swift
philippsuess: If they all belong together, I don't care too much about the number. But in this case they don't really.

Sources/Swift/Tools/Batcher/BatchStorage.swift
philippsuess: They can be, but they don't have to. No strong opinion

Sources/Swift/Tools/Batcher/Batcher.swift
philippsuess: I would rather make it a fixed value. Also, it would be weird if log items linked to one trace suddenly have a different installationID.

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,160 @@
...
+        return TimeInterval(endTimeNs - startTimeNs) / 1_000_000_000.0 // Convert nanoseconds to seconds
+    }
+
+    /// Encodes and buffers an item, triggering a flush if limits are reached.
+    ///
+    /// - Important: Only call this method from the serial dispatch queue to ensure thread safety.
+    /// - Parameter item: The item to encode and add to the buffer
+    private func encodeAndBuffer(item: Item) {
+        do {
+            let encodedItemsWereEmpty = storage.itemsDataSize == 0
```
philippsuess: `l`: `encodedItemsWereEmpty` is a bit misleading, IMO. 
```suggestion
            let storageWasEmpty = storage.itemsDataSize == 0
```

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,160 @@
...
+
+            // Flush when we reach max item count or max buffer size
+            if storage.itemsCount >= config.maxItemCount || storage.itemsDataSize >= config.maxBufferSizeBytes {
+                performCaptureItems()
+            } else if encodedItemsWereEmpty && timerWorkItem == nil {
+                startTimer()
+            }
+        } catch {
+            SentrySDKLog.error("Failed to encode item: \(error)")
+        }
```
philippsuess, line 128: `l`: Only the storage.append(item) can throw here. The error message is also related to this specific `try`. This makes the code easier to read, because the try and the related error message are close to each other.
```suggestion
        let encodedItemsWereEmpty = storage.itemsDataSize == 0
        do {
            try storage.append(item)
        } catch {
            SentrySDKLog.error("Failed to encode item: \(error)")
            return
        }

        // Flush when we reach max item count or max buffer size
        if storage.itemsCount >= config.maxItemCount || storage.itemsDataSize >= config.maxBufferSizeBytes {
            performCaptureItems()
        } else if encodedItemsWereEmpty && timerWorkItem == nil {
            startTimer()
        }
```

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,167 @@
...
+    ///              **Do not call this method from within the batcher's dispatch queue or from within the
+    ///              `capturedDataCallback` closure**, as this will cause a deadlock. This method should only be
+    ///              called from external threads or queues (e.g., main thread, app lifecycle callbacks).
+    @discardableResult func capture() -> TimeInterval {
+        let startTimeNs = dateProvider.getAbsoluteTime()
+        dispatchQueue.dispatchSync { [weak self] in
+            self?.performCaptureItems()
+        }
+        let endTimeNs = dateProvider.getAbsoluteTime()
+        return TimeInterval(endTimeNs - startTimeNs) / 1_000_000_000.0 // Convert nanoseconds to seconds
```
philippsuess, line 108: `l`: I know this already existed, but maybe worth reusing methods from https://github.com/getsentry/sentry-cocoa/blob/main/SentryTestUtils/Sources/TimeInterval%2BSentry.swift 

or adding an extra method

Sources/Swift/Tools/Batcher/BatchBuffer.swift
```diff
@@ -0,0 +1,20 @@
...
+    mutating func append(_ item: Item) throws
+
+    /// Clears all items from the storage
+    mutating func clear()
+
+    /// Number of elements in the storage
+    var itemsCount: Int { get }
+
+    /// Sum of the size of encoded items in the storage
+    var itemsDataSize: Int { get }
```
philippsuess, line 16: `l`: I personally find such comments not really useful, because the methods are crystal clear. I know we already had a chat about this in the past and you find them useful. I would like to understand if you, @itaybre and @noahsmartin, find them useful? We can keep them, of course.

Sources/Swift/Tools/Batcher/Batcher.swift
```diff
@@ -0,0 +1,167 @@
...
+    func capture() -> TimeInterval
+}
+
+final class Batcher<Buffer: BatchBuffer<Item>, Item: BatcherItem, Scope: BatcherScope>: BatcherProtocol {
+    struct Config: BatcherConfig {
+        let flushTimeout: TimeInterval
+        let maxItemCount: Int
+        let maxBufferSizeBytes: Int
+        let beforeSendItem: ((Item) -> Item?)?
+        let installationId: String?
```
philippsuess: `l`: I think the `installationId` should also be part of the `Metadata`.

Sources/Swift/Tools/Batcher/InMemoryBatchBuffer.swift
```diff
@@ -0,0 +1,27 @@
+struct InMemoryBatchBuffer<Item: Encodable>: BatchBuffer {
+    private var elements: [Data] = []
+    var itemsDataSize: Int = 0
+
+    init() {}
+
+    mutating func append(_ item: Item) throws {
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = .secondsSince1970
```
philippsuess: `l`: Wouldn't it make sense to init the `JSONEncoder` once and keep it as a private property?

philipphofmann (commented):
A few more minor comments.

Sources/Swift/Tools/Batcher/BatchBuffer.swift
philprime: I added them because @denrase mentioned that the naming of `count` and `size` i used before was arbitrary. Please also consider we are now using review bots using LLMs which work better with more documentation in code

Sources/Swift/Tools/Batcher/InMemoryBatchBuffer.swift
philprime: Good catch. It's a class so it could cause a performance impact due to many allocations.

PR #7019, opened 2025-12-09
chore: Removes HybridSDK subspec

itaybre:
This PR is to explore the possibility of removing the HybridSDK subspec since either way all API are public.

This change would also make maintenance easier for the maintainers

More reasons in the decision in #7024


.github/workflows/test-cross-platform.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         with:
           repository: getsentry/sentry-react-native
           path: sentry-react-native
-          ref: cocoa-v9
+          ref: itay/use_regular_podspec
```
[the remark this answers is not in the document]
philippsuess: `m`: Do we maybe need to update this ref before merging?

.github/workflows/test-cross-platform.yml
```diff
@@ -86,10 +86,10 @@ jobs:
         run: yarn install
 
       - name: Remove Sentry/HybridSDK Dependency from RNSentry.podspec
```
philippsuess: `l`: The name needs updating. It still states HybridSDK dependency. The same applies to the job name below.

Sources/Sentry/Public/Sentry.h
```diff
@@ -43,4 +43,15 @@ FOUNDATION_EXPORT const unsigned char SentryVersionString[];
...
+// Headers here are to be used by Hybrid SDKs
+// Be careful when using them
+#    import <Sentry/PrivateSentrySDKOnly.h>
+#    import <Sentry/PrivatesHeader.h>
+#    import <Sentry/SentryAppStartMeasurement.h>
+#    import <Sentry/SentryBreadcrumb+HybridSDK.h>
+#    import <Sentry/SentryFormatter.h>
+#    import <Sentry/SentryOptionsInternal.h>
+#    import <Sentry/SentryProfilingConditionals.h>
+#    import <Sentry/SentrySessionReplayHybridSDK.h>
```
philippsuess, line 52: `h`: I'm unsure about simply making all of these public. Previously, you had to explicitly use the hybrid podspec. Now, all of these are public, but it's unclear whether these are somewhat reserved for hybrid SDKS. I think we should not make all of these simply public, but instead have a discussion with Hybrid about exactly what they need, and then offer proper APIs, or somehow expose them all via a single class or file. I find it confusing that we have PrivatesHeader, and SentryOptionsInternal now public. We should change the names for these if we make them public IMO.

.github/workflows/test-cross-platform.yml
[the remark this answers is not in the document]
itaybre: That branch changes the CocoaPods dependencies, or all RN builds will fail

Sources/Sentry/Public/Sentry.h
itaybre: I checked the headers:
- `PrivateSentrySDKOnly.h`: the PrivateSentrySDKOnly name points to be non public
- `PrivatesHeader.h`: is a helper file to import `SentryProfilingConditionals` and `SentryDefine` macros, I don't see any reason for this to be private, but changing its name would be breaking change
- `SentryProfilingConditionals.h`: Macros, needs to be public now
- `SentrySessionReplayHybridSDK.h`: The class name shows it is not public
- `SentryFormatter.h`: a couple of utils to convert strings and UInt, nothing that cannot be public.

Now public models:
- `SentryAppStartMeasurement.h`
- `SentryBreadcrumb+HybridSDK`: just exposes the initializer
Even on a new API, we would probably need to expose some models, if they are the same than the internal ones, won't make differences

I would say `SentryOptionsInternal` is the only one that is now exposed and maybe it could be improved. 
We could move it to `PrivateSentrySDKOnly`, and but it is being used in many [other SDK](https://github.com/search?q=org%3Agetsentry+SentryOptionsInternal&type=code). I wanted to avoid causing unnecessary changes

itaybre: Also, I would say `SentryOptionsInternal` is confusing, it is just a constructor for `SentryOptions` from a dictionary. I believe it stayed with this name to avoid a breaking change

itaybre: Updated this to expose less headers, only:
```
// Macros
#    import <Sentry/PrivatesHeader.h>
// HybridSDK APIS
#    import <Sentry/PrivateSentrySDKOnly.h>
#    import <Sentry/SentrySessionReplayHybridSDK.h>
// Models
#    import <Sentry/SentryAppStartMeasurement.h>
// Inlinable helper functions
#    import <Sentry/SentryFormatter.h>
```

Sources/Sentry/Public/Sentry.h
```diff
@@ -43,4 +43,12 @@ FOUNDATION_EXPORT const unsigned char SentryVersionString[];
 #    import <Sentry/SentryTransactionContext.h>
 #    import <Sentry/SentryUser.h>
 #    import <Sentry/SentryWithoutUIKit.h>
+
+// Headers here are to be used by Hybrid SDKs
+// Be careful when using them
+#    import <Sentry/PrivateSentrySDKOnly.h>
+#    import <Sentry/PrivatesHeader.h>
```
philippsuess, line 50: `m`: Wouldn't it make more sense to simply move the `SentryProfilingConditionals` file to the SentryDefines.h

https://github.com/getsentry/sentry-cocoa/blob/7e7673487a8133ac359ec3832d25681666682082/Sources/Sentry/Public/PrivatesHeader.h#L10-L14

Sources/Sentry/Public/Sentry.h
philippsuess: Yep, it's cleaner now thanks.

philipphofmann (approved):
Thank you for ditching this complexity.

Sources/Sentry/Public/Sentry.h
itaybre: I agree but that change looks a bit more complex, I should do this on a different PR

PR #7025, opened 2025-12-10
chore: Add flutter cross platform integration test

itaybre:
Adds an integration test for Flutter to verify our changes are not introducing any breaking change for them


itaybre:
There are still many build issues, so we cannot enable this test yet

PR #7031, opened 2025-12-11
chore: Cleanup file filter for required files modified

itaybre:
In addition to #7030, found more files that shouldn't be in the filters since they aren't used in those jobs


.github/file-filters.yml
```diff
@@ -70,17 +70,11 @@ run_api_stability_for_prs: &run_api_stability_for_prs
   - "Package*.swift"
 
   # Scripts
```
itaybre:
```suggestion
  # Scripts  
  - "scripts/build-xcframework-slice.sh"
  - "scripts/assemble-xcframework.sh"
```

PR #7043, opened 2025-12-12
chore: Convert `SentryScreenshotIntegration` to Swift

### LINKED ISSUE #7047: chore: Convert `SentryScreenshotIntegration` to Swift
> [!NOTE]
> The pull request "[chore: Convert `SentryScreenshotIntegration` to Swift](https://github.com/getsentry/sentry-cocoa/pull/7043)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Converts `SentryScreenshotIntegration` to Swift

While I tried not using `SentryDependencyContainer` anymore, the C block requires a global variable.

#skip-changelog

itaybre:
Converts `SentryScreenshotIntegration` to Swift

While I tried not using `SentryDependencyContainer` anymore, the C block requires a global variable.

#skip-changelog

Closes #7047

### DIFF
--- a/Sources/Swift/Integrations/Screenshot/SentryScreenshotIntegration.swift
+++ b/Sources/Swift/Integrations/Screenshot/SentryScreenshotIntegration.swift
@@ -6,10 +6,6 @@
 // nor we want to continue using the DependencyContainer
 private weak var globalScreenshotSource: SentryScreenshotSource?
 
-protocol ScreenshotIntegrationProvider {
-    var screenshotSource: SentryScreenshotSource? { get }
-}
-
 final class SentryScreenshotIntegration<Dependencies: ScreenshotIntegrationProvider>: NSObject, SwiftIntegration, SentryClientAttachmentProcessor {
     private let options: Options
     private let screenshotSource: SentryScreenshotSource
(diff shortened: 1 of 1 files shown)


Sources/Swift/Integrations/Screenshot/SentryScreenshotIntegration.swift
```diff
@@ -0,0 +1,101 @@
+@_implementationOnly import _SentryPrivate
+
+#if (os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)) && !SENTRY_NO_UIKIT
+
+// We need to use a global variable because C doesn't allow capturing var
+// nor we want to continue using the DependencyContainer
+private weak var globalScreenshotSource: SentryScreenshotSource?
+
+protocol ScreenshotIntegrationProvider {
```
philprime: `l`: I believe we should keep these protocols and the extensions of the SentryDependencyContainer in the `SentryDependencyContainer.swift`, because we will most likely reuse them in other integrations too

Sources/Swift/Integrations/Screenshot/SentryScreenshotIntegration.swift
```diff
@@ -0,0 +1,101 @@
...
+
+        let screenshots = screenshotSource.appScreenshotDatasFromMainThread()
+
+        for (index, data) in screenshots.enumerated() {
+            let name = index == 0 ? "screenshot.png" : "screenshot-\(index + 1).png"
+            let attachment = Attachment(data: data, filename: name, contentType: "image/png")
+            result.append(attachment)
+        }
+
+        return result
```
philprime: `l`: Could be simplified:

```suggestion
        let screenshots = screenshotSource.appScreenshotDatasFromMainThread()
        
        return screenshots.enumerated().map { (index, data) in 
            let name = index == 0 ? "screenshot.png" : "screenshot-\(index + 1).png"
            return Attachment(data: data, filename: name, contentType: "image/png")
        }
```

Sources/Swift/Integrations/Screenshot/SentryScreenshotIntegration.swift
itaybre: I don't have any preference here, I can move it

Sources/Swift/Integrations/Screenshot/SentryScreenshotIntegration.swift
itaybre: Using map sounds good, but I will apply a modification since we need to keep the previous attachments

Issue #7054, opened 2025-12-16
[Logs] Add guard for sendDefaultPII for user attributes


philprime:
### Description

[User attributes](https://develop.sentry.dev/sdk/telemetry/logs/#user-attributes) should be guarded by `sendDefaultPII` as they are for [Metrics](https://develop.sentry.dev/sdk/telemetry/metrics/#user-attributes). This information was missing in the develop-docs previously and added with https://github.com/getsentry/sentry-docs/pull/15833

PR #7055, opened 2025-12-16
fix(logs): Use sendDefaultPii and span_id for attributes

philprime:
## :scroll: Description

- Changes the default attribute `sentry.trace.parent_span_id` to `span_id` as defined in the [Log Protocol](https://develop.sentry.dev/sdk/telemetry/logs/)
- Adds the `sendDefaultPii` option to the scope to use it for guarding of user-related attributes (as recently changed in the [Log Protocol](https://develop.sentry.dev/sdk/telemetry/logs/#user-attributes))

Closes https://github.com/getsentry/sentry-cocoa/issues/7054

## :pencil: Checklist

### DIFF
--- a/Tests/SentryTests/SentryLogBatcherTests.swift
+++ b/Tests/SentryTests/SentryLogBatcherTests.swift
@@ -344,7 +344,7 @@ final class SentryLogBatcherTests: XCTestCase {
         let attributes = capturedLog.attributes
         
         XCTAssertNil(attributes["sentry.release"])
-        XCTAssertNil(attributes["sentry.trace.parent_span_id"])
+        XCTAssertNil(attributes["span_id"])
         XCTAssertEqual(attributes["sentry.sdk.name"]?.value as? String, SentryMeta.sdkName)
         XCTAssertEqual(attributes["sentry.sdk.version"]?.value as? String, SentryMeta.versionString)
         XCTAssertNotNil(attributes["sentry.environment"])


Sources/Sentry/Public/SentryScope.h
```diff
@@ -46,6 +46,12 @@ NS_SWIFT_NAME(Scope)
  */
 @property (nonatomic, readonly, copy) NSDictionary<NSString *, id> *attributes;
 
+/**
+ * Indicates if default PII should be sent with events.
+ * This is set from ``SentryOptions.sendDefaultPii`` during SDK initialization.
+ */
+@property (nonatomic) BOOL sendDefaultPii;
```
philprime: I tried to add this to `SentryScope+Private.h` instead so it's not a property, but it wasn't available in Swift, so this was easier.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Use `sendDefaultPii` and `span_id` for attributes in `BatcherScope` (#7055)
```
itaybre: `l`: I would say it isn't very clear what this change means

philprime: What do you think about this:

```suggestion
- Change default attributes of Logs to only include user attributes when `options.sendDefaultPii = true` (#7055)
- Rename log attribute `sentry.trace.parent_span_id` to `span_id` (#7055)
```

Sources/Sentry/Public/SentryScope.h
philippsuess: `h`: I think this could confuse users a bit. You have one setting `sendDefaultPii` on the options and now also here on the scope. Do we have a use case for this to be different for scopes? If not, I would prefer if we found another way of keeping this internal. Wherever we apply the scope we should somehow get access to `sendDefaultPii` of the options instead.

philprime:
Moving back to draft until I find a way to do this without adding public API

philprime:
I found the rather obvious way of adding it to the BatcherConfig (not BatcherMetadata even though it originates from the options, but isn't additional metadata).

Now there's on change in public API.

Tests/SentryTests/SentryLogBatcherTests.swift
```diff
@@ -368,35 +368,70 @@ final class SentryLogBatcherTests: XCTestCase {
...
+        // -- Act --
+        sut.addLog(log, scope: scope)
+        sut.captureLogs()
+
+        // -- Assert --
+        let capturedLogs = testDelegate.getCapturedLogs()
+        let capturedLog = try XCTUnwrap(capturedLogs.first)
+        let attributes = capturedLog.attributes
+
+        // The installation id is used as a fallback for the user.id
```
philippsuess, line 421: Thanks for the comment

PR #7061, opened 2025-12-17
Add `CrashSafeBatchBuffer` and save/restore/send logs on crash

denrase:
## :scroll: Description

This PR implements crash-safe batch buffering for log capture, ensuring that logs are preserved even when the app crashes. The implementation uses C-level APIs that are accessible from crash handlers, storing logs in crash-safe memory and persisting them to disk upon crash detection, then automatically recovering and sending them on the next app launch.

### Overview

Implements a crash-safe storage mechanism for batched logs using C-level APIs that are accessible from crash handlers. Logs are stored in crash-safe memory and persisted to disk when a crash is detected, then automatically recovered and sent on the next app launch.

### Key Changes

**Crash-Safe Storage Infrastructure:**
- Added `SentryBatchBufferC` (C API) for crash-safe batch buffer operations
- Added `SentryLogSyncC` to manage a global log buffer accessible during crashes
- Buffer automatically persists logs to disk when a crash is detected

**Swift Integration:**
- Implemented `CrashSafeBatchBuffer` conforming to `BatchBuffer` protocol
- Updated `SentryLogBatcher` to use crash-safe buffers with automatic fallback to in-memory storage if initialization fails
- Buffer uses double capacity (2x `maxBufferSizeBytes`) to ensure sufficient space (adding slightly larger items due to json structures, as we only cound the items sizes)

**Crash Recovery:**
- Added to `FlushLogsIntegration`:
  - Create telemetry buffer folder structure
  - Set up crash detection file paths
  - Recover and send logs from previous crashes on app launch

**Testing:**
- Added  `CrashSafeBatchBufferTests`
- Updated `InMemoryBatchBufferTests` with additional test coverage
- Added sample app crash testing functionality in `ErrorsViewController`

### Technical Details

- Batcher is not dispathcing to a serial queue anymode, but rather doing thread-safe operations using `NSLock` synchronization
- Automatic fallback to `InMemoryBatchBuffer` if crash-safe initialization fails
- Logs are written to disk in JSON format during crash detection
- Recovery happens automatically on next app launch via `FlushLogsIntegration`
- Buffer capacity: 2x `maxBufferSizeBytes` for data, `maxLogCount` for items

Relates to [COCOA-480](https://linear.app/getsentry/issue/COCOA-480/structured-logs-logs-for-crashes)

Unit tests. Ran sample app.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Add isiOSAppOnVisionOS, isiOSAppOnMac, isMacCatalystApp to device context #6939
+- Add `CrashSafeBatchBuffer` and save/restore/send logs on crash ([](#7061)
 
 ### Fixes
 


Sources/Swift/Integrations/Log/FlushLogsIntegration.swift
```diff
@@ -76,6 +83,94 @@ final class FlushLogsIntegration<Dependencies: NotificationCenterProvider>: NSOb
...
+            }
+            
+            let itemCount = items.count
+            guard itemCount > 0 else {
+                return
+            }
+            
+            SentrySDKLog.debug("FlushLogsIntegration: Recovering \(itemCount) crash logs.")
+            
+            // TEPM: Send the logs through the client (which conforms to SentryLogBatcherDelegate)
```
denrase, line 162: Not sure how to handle this. Easiest would be to either provide an internal function, or write the envelope directly to the out directory, like it is currently described in the spec for the double rotating buffer. If we do the latter, we should extract the envelope construction out of the client.

Sources/Swift/Tools/Batcher/CrashSafeBatchBuffer.swift
```diff
@@ -0,0 +1,108 @@
...
+/// A wrapper around the global crash-safe log buffer.
+///
+/// This class manages a buffer that is accessible from the crash handler during a crash.
+/// The buffer is allocated in C memory and remains stable for the lifetime of this object.
+///
+/// - Important: Only one `CrashSafeBatchBuffer` should be active at a time.
+///              Creating a new instance will destroy any previously created buffer.
+final class CrashSafeBatchBuffer<Item: Encodable>: BatchBuffer {
+    
+    let lock = NSLock()
```
denrase, line 18: We could consider removing this lock, as it is already guarded by the batcher lock and this class should not be used anywhere else anyway.

Sources/Swift/Tools/SentryLogBatcher.swift
```diff
@@ -61,33 +61,80 @@ import Foundation
         dispatchQueue: SentryDispatchQueueWrapper,
         delegate: SentryLogBatcherDelegate
     ) {
-        self.batcher = Batcher(
+        let capturedDataCallback = Self.createCapturedDataCallback(delegate: delegate)
+        
+        do {
+            let crashSafeBuffer = try CrashSafeBatchBuffer<SentryLog>(
```
denrase, line 67: We could also put this behind a compiler/feature flag, as long as we don't have the complete double rotating buffer spec implemented.

Sources/SentryCrash/Recording/SentryCrashC.c
```diff
@@ -85,6 +86,7 @@ onCrash(struct SentryCrash_MonitorContext *monitorContext)
         strlcpy(g_lastCrashReportFilePath, crashReportFilePath, sizeof(g_lastCrashReportFilePath));
         sentrycrashreport_writeStandardReport(monitorContext, crashReportFilePath);
         sentrySessionReplaySync_writeInfo();
+        sentryLogSync_writeToFile();
```
philippsuess, line 89: `h`: We definitely must call this before the `sentrySessionReplaySync_writeInfo` directly after the `sentrycrashreport_writeStandardReport`. Logs have higher prio that SessionReplay.

Sources/Swift/Tools/Batcher/CrashSafeBatchBuffer.swift
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/// A wrapper around the global crash-safe log buffer.
+///
+/// This class manages a buffer that is accessible from the crash handler during a crash.
+/// The buffer is allocated in C memory and remains stable for the lifetime of this object.
+///
+/// - Important: Only one `CrashSafeBatchBuffer` should be active at a time.
+///              Creating a new instance will destroy any previously created buffer.
+final class CrashSafeBatchBuffer<Item: Encodable>: BatchBuffer {
```
philippsuess, line 16: `m `: I know you're just sticking to the existing naming, but I think we should ditch the `Batch` from the name. These are just buffers. Mixing batch and buffer in the same name is confusing to me.
```suggestion
final class CrashSafeBuffer<Item: Encodable>: BatchBuffer {
```

philippsuess, line 16: `h`: This implementation is tied to logs, actually. But I could use it with anything that implements encoding. So this would compile

```
let buffer = try CrashSafeBatchBuffer<SentryAttribute>(
    dataCapacity: maxBufferSizeBytes * 2,
    itemsCapacity: maxLogCount
)
```

I would keep this only specific to logs at the moment. 
```suggestion
final class CrashSafeLogBuffer: BatchBuffer {
```

Sources/Swift/Tools/SentryLogBatcher.swift
philippsuess: Yes indeed, this should be a feature flag in the options, I think. What I would do is to have this SentryLogsBatcher act as a kind of facade. When the crash safe feature is enabled, we forward the logs to the CrashSafeBuffer, if not to the InMemoryBuffer.

Sources/Swift/Tools/SentryLogBatcher.swift
```diff
@@ -61,33 +61,80 @@ import Foundation
...
+                options: options,
+                flushTimeout: flushTimeout,
+                maxLogCount: maxLogCount,
+                maxBufferSizeBytes: maxBufferSizeBytes,
+                capturedDataCallback: capturedDataCallback,
+                buffer: inMemoryBuffer,
+                dateProvider: dateProvider,
+                dispatchQueue: dispatchQueue
+            )
+        }
```
philippsuess, line 94: `m`: I wouldn't do that. We should always be able to init the crash safe store. If that doesn't work, I think we should disable logs completely. Logs shouldn't have to use the in-memory buffer. They should rely on the crash safe buffer.

Sources/Swift/Integrations/Log/FlushLogsIntegration.swift
```diff
@@ -76,6 +85,94 @@ final class FlushLogsIntegration<Dependencies: NotificationCenterProvider>: NSOb
...
+        url.appendPathComponent("io.sentry")
+        let dsnHash = dsnHash ?? ""
+        url.appendPathComponent(dsnHash)
+        url.appendPathComponent("telemetry-buffer")
+        return url
+    }
+    
+    private func getDetectedTerminationFileURL(folder: URL) -> URL {
+        return folder.appendingPathComponent("detected-termination")
+    }
```
philippsuess, line 125: `m`: We usually do all that path handling in the SentryFileManager / SentryFileManagerHelper. Even though this isn't a great pattern, I would like to keep following it, because the SentryFileManagerHelper already handles a couple of edge cases when it creates the required folders and such.

Sources/Swift/Integrations/Log/FlushLogsIntegration.swift
```diff
@@ -76,6 +85,94 @@ final class FlushLogsIntegration<Dependencies: NotificationCenterProvider>: NSOb
         client.captureLogs()
     }
     
+    // MARK: - Telemetry Buffer Setup
+    
+    private func createTelemetryBufferFolderIfNeeded() {
```
philippsuess, line 90: `m`: I don't think any of the telemetry buffer code in here belongs here. The FlushLogIntegration shouldn't care which type of buffer we use. This code should live in the `CrashSafeBatchBuffer`, IMO.

Sources/Sentry/SentryLogSyncC.c
```diff
@@ -0,0 +1,126 @@
+#include "SentryLogSyncC.h"
+#include "SentryAsyncSafeLog.h"
+#include <SentryCrashFileUtils.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+static SentryBatchBufferC *logBuffer = NULL;
```
philippsuess, line 10: `h`: We defenitely need some native dev review here.

Sources/SentryCrash/Recording/Tools/SentryBatchBufferC.h
```diff
@@ -0,0 +1,50 @@
...
+//
+
+#ifndef HDR_SENTRY_BATCH_BUFFER_H
+#define HDR_SENTRY_BATCH_BUFFER_H
+
+#include <stdbool.h>
+#include <stddef.h>
+
+typedef struct {
+    char *data;
```
philippsuess, line 14: `m`: I wonder if it wouldn't be better to use a `char **` array as we do for https://github.com/getsentry/sentry-cocoa/blob/0710ba958a35640d73f1482dd0020e0ddb5317ee/Sources/Sentry/include/SentryScopeSyncC.h#L14

philipphofmann (commented):
Thanks for tackling this, this is a good first start, but I think we will need a few iteractions until we can push this over the finish line.

Sources/SentryCrash/Recording/Tools/SentryBatchBufferC.h
denrase: Yeah I will check if we can simplify this. Since this is one-dimensional buffer, we are doing quite a bit of "bookkeeping" by also keeping track of the individual items lengths and indices.

Sources/Swift/Tools/Batcher/CrashSafeBatchBuffer.swift
denrase: Naming is following the protocol that this class is implementing. If we find that it is incorrect wording, we should extend our re-naming efforts to the protocol.

philippsuess:
Let's close this for now. We can reopen this PR when we start working on this topic again.

PR #7063, opened 2025-12-17
fix(session-replay): Add options to ignore views from subtree traversal

### LINKED ISSUE #7053: [Session Replay] Add public API to configure sub-tree traversal
### Description

We need a public API to add and remove view types from the subtree traversal in the SentryUIRedactBuilder:

https://github.com/getsentry/sentry-cocoa/blob/4558de25dfe0e6989e994f5ddc49aebe40444616/Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift#L552-L590

It will allow SDK users to extend the list of views, in case they are experiencing crashes caused by the view traversal, e.g. activating internal animations for CoreAnimation.

**Example:**

```
Application Specific Information:
-[NSConcreteValue doubleValue]: unrecognized selector sent to instance 0x1409e0cc0

Thread 0 Crashed:
0   CoreFoundation                  0x3252a1994         __exceptionPreprocess
1   libobjc.A.dylib                 0x31f169810         objc_exception_throw
2   CoreFoundation                  0x32533a8fc         -[NSObject(NSObject) doesNotRecognizeSelector:]
3   CoreFoundation                  0x325221358         ___forwarding___
4   CoreFoundation                  0x3252291fc         __forwarding_prep_0___
5   QuartzCore                      0x326843c68         -[NSNumber(CAAnimatableValue) CA_interpolateValue:byFraction:]
6   QuartzCore                      0x32683f99c         -[CABasicAnimation applyForTime:presentationObject:modelObject:]
7   QuartzCore                      0x3265f98bc         CA::Layer::presentation_layer
8   QuartzCore                      0x3265f4008         CA::Layer::sublayers
9   MyApp                           0x200a98d40         SentryUIRedactBuilder.mapRedactRegion
10  MyApp.                          0x200a98ed4         [inlined] SentryUIRedactBuilder.mapRedactRegion
```

A reference for similar options is the list of masked and unmasked view classes in the session replay options:

https://github.com/getsentry/sentry-cocoa/blob/4558de25dfe0e6989e994f5ddc49aebe40444616/Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift#L145-L163

We want to offer the SDK users to add additional view types, while also removing ours pre-defined ones (for flexibility), which is the `CameraUI.ChromeSwiftUIView` at this point in time only.

https://github.com/getsentry/sentry-cocoa/blob/4558de25dfe0e6989e994f5ddc49aebe40444616/Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift#L72-L77

Therefore, define the list should be a private field on the [SentryReplayOptions.swift](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift) with two public helper methods `includeSubtreeTraversalForViewType(String)` and `excludeSubtreeTraversalForViewType(String)`. The given string value will be compared to `type(of: view).description()`.

philprime:
**This PR should be merged after #7089**

## :scroll: Description

This PR adds configurable subtree traversal ignoring functionality to prevent crashes when traversing problematic view hierarchies during session replay and screenshot capture.

**Key Changes:**

- Added `viewTypesIgnoredFromSubtreeTraversal: Set<String>` property to `SentryRedactOptions` protocol
- Implemented the property in all conforming classes:
  - `SentryReplayOptions` - with helper methods `excludeViewTypeFromSubtreeTraversal(_:)` and `includeViewTypeInSubtreeTraversal(_:)`
  - `SentryRedactDefaultOptions` - computed property with iOS 26+ default
  - `SentryViewScreenshotOptions` - mutable property with empty default
  - `PreviewRedactOptions` - immutable property with default from `SentryReplayOptions`
- Modified `SentryUIRedactBuilder` to check ignored view types before accessing `layer.sublayers`, preventing crashes
- Added dictionary initialization support for `viewTypesIgnoredFromSubtreeTraversal` in `SentryReplayOptions` and `SentryViewScreenshotOptions`

**Default Behavior:**

- On iOS 26+, `CameraUI.ChromeSwiftUIView` is automatically included in the ignore set by default for `SentryReplayOptions` and `SentryRedactDefaultOptions` to prevent crashes when accessing `CameraUI.ModeLoupeLayer`
- Other platforms/versions have an empty ignore set by default
- `SentryViewScreenshotOptions` has an empty ignore set by default

**Implementation Details:**

When a view type is in the ignore set, `SentryUIRedactBuilder` will:
- Still redact the view itself (unless explicitly marked to be ignored, like `UISwitch`)
- Skip traversing the view's subtree by returning early
- Prevent crashes from accessing problematic sublayers

Closes #7053

Some view hierarchies contain layers that crash when accessed during traversal. Specifically, `CameraUI.ChromeSwiftUIView` on iOS 26+ contains `CameraUI.ModeLoupeLayer` instances that cause fatal errors when their `sublayers` property is accessed, resulting in crashes during session replay or screenshot capture.

This change provides a configurable way to exclude problematic view types from subtree traversal while still redacting the views themselves.

- Added comprehensive unit tests for all new functionality:
  - Default value tests verifying `CameraUI.ChromeSwiftUIView` is included on iOS 26+
  - Dictionary initialization tests covering valid/invalid/mixed values
  - Helper method tests for `excludeViewTypeFromSubtreeTraversal` and `includeViewTypeInSubtreeTraversal`
  - Subtree traversal behavior tests using `ProblematicLayer`/`ProblematicView` test classes that track `sublayers` access
  - Tests verify that ignored view types skip subtree traversal (no `sublayers` access) while non-ignored types traverse normally
- Updated existing tests to use the new property name and API
- All tests pass on macOS and iOS simulators

## :pencil: Checklist

- [X] I updated the docs if needed. → https://github.com/getsentry/sentry-docs/pull/15854

### DIFF
--- a/Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
+++ b/Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
@@ -139,7 +139,7 @@ final class SentryUIRedactBuilder {
     ///
     /// - note: On iOS, views such as `WKWebView` and `UIWebView` are always redacted, and controls like
     ///   `UISlider` and `UISwitch` are ignored by default.
-    init(options: SentryRedactOptions) {
+    init(options: SentryRedactOptions) { // swiftlint:disable:this function_body_length
         var redactClasses = Set<ClassIdentifier>()
 
         if options.maskAllText {


Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -330,7 +387,8 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
             frameRate: (dictionary["frameRate"] as? NSNumber)?.uintValue,
             errorReplayDuration: (dictionary["errorReplayDuration"] as? NSNumber)?.doubleValue,
             sessionSegmentDuration: (dictionary["sessionSegmentDuration"] as? NSNumber)?.doubleValue,
-            maximumDuration: (dictionary["maximumDuration"] as? NSNumber)?.doubleValue
+            maximumDuration: (dictionary["maximumDuration"] as? NSNumber)?.doubleValue,
+            viewTypesIgnoredFromSubtreeTraversal: (dictionary["viewTypesIgnoredFromSubtreeTraversal"] as? [String]).map { Set($0) }
```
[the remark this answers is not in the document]
philprime: There is not failure handling implemented anyways

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -198,17 +204,18 @@ final class SentryUIRedactBuilder {
...
-        
+
         for type in options.maskedViewClasses {
             redactClasses.insert(ClassIdentifier(class: type))
         }
-        
+
         redactClassesIdentifiers = redactClasses
-        
+        viewTypesIgnoredFromSubtreeTraversal = options.viewTypesIgnoredFromSubtreeTraversal
```
benjaminpeters: I'm running into this bug in a react native app, I pulled your branch and wired in `viewTypesIgnoredFromSubtreeTraversal` to sentry-react-native's `MobileReplayOptions` to test everything out. Right now passing a value in my Sentry.mobileReplayIntegration ( ie `viewTypesIgnoredFromSubtreeTraversal: ['RCTView']`) overrides the default array: `[CameraUI.ChromeSwiftUIView]` with `['RCTView']`. I'm thinking it's better to keep the default and insert new viewTypes.

Right now I've added this to keep testing 
```
viewTypesIgnoredFromSubtreeTraversal = SentryReplayOptions.DefaultValues.viewTypesIgnoredFromSubtreeTraversal
        
for viewType in options.viewTypesIgnoredFromSubtreeTraversal {
       viewTypesIgnoredFromSubtreeTraversal.insert(viewType)
}
```

philprime: Thanks for exploring the solution in this PR and bringing in your feedback!
The reason why I didn't want to do it that way is that I want to allow removing types as well, so that SDK users have full control with adding and removing.

But I also notice there is a risk in accidentally removing safe-guards we have added before like the `CameraUI.ChromeSwiftUIView`, so I might reconsider it.

Do I understand correctly that you set the array of views in the `Sentry.mobileReplayIntegration` which is then deserialized by the `SentryReplayOptions.init(dictionary:)`?

benjaminpeters: ah yeah good call. What do you think about a separate property to add viewTypes? maybe something like this:
    
```
@objcMembers
@_spi(Private) public final class SentryRedactDefaultOptions: NSObject, SentryRedactOptions {
    public var maskAllText: Bool = true
    public var maskAllImages: Bool = true
    public var maskedViewClasses: [AnyClass] = []
    public var unmaskedViewClasses: [AnyClass] = []
    
    // Stored property for user additions
    public var InsertAdditionalViewTypesIgnoredFromSubtreeTraversal: Set<String> = []
    /// Default view types for which subtree traversal should be ignored.
    ///
    /// By default, includes `CameraUI.ChromeSwiftUIView` on iOS 26+ to avoid crashes
    /// when accessing `CameraUI.ModeLoupeLayer`.
    
    // Computed property that combines defaults + custom
    public var viewTypesIgnoredFromSubtreeTraversal: Set<String> {
        var defaults: Set<String> = []
#if os(iOS)
        if #available(iOS 26.0, *) {
            defaults.insert("CameraUI.ChromeSwiftUIView")
        }
#endif
        return defaults.union(additionalViewTypesIgnoredFromSubtreeTraversal)
    }
}

```

This way you can insert new viewTypes or replace the entire array 
`viewTypesIgnoredFromSubtreeTraversal: ['RCTView'] ` or  `InsertAdditionalViewTypesIgnoredFromSubtreeTraversal:  ['RCTView'] `

maybe better to keep it simple though if a user needs to add additional viewTypes they can always include CameraUI.ModeLoupeLayer

Sources/SentrySwiftUI/Preview/PreviewRedactOptions.swift
```diff
@@ -35,6 +35,21 @@ public class PreviewRedactOptions: SentryRedactOptions {
...
+     * caused by traversing problematic view hierarchies (e.g., views that activate internal CoreAnimation
+     * animations when their layers are accessed).
+     *
+     * The string values should match the result of `type(of: view).description()`.
+     *
+     * - Note: See ``SentryReplayOptions.DefaultValues.viewTypesIgnoredFromSubtreeTraversal`` for the default value.
+     * - Note: By default, this includes `CameraUI.ChromeSwiftUIView` on iOS 26+ to avoid crashes
+     *         when accessing `CameraUI.ModeLoupeLayer`.
+     */
+    public let viewTypesIgnoredFromSubtreeTraversal: Set<String>
```
philippsuess: `m`: Above we have the naming unmaskedView**Classes**. Here we have **types**. Can't we also use **classes** here for consistency?

Sources/SentrySwiftUI/Preview/PreviewRedactOptions.swift
```diff
@@ -35,6 +35,21 @@ public class PreviewRedactOptions: SentryRedactOptions {
...
     public let unmaskedViewClasses: [AnyClass]
     
+    /**
+     * A set of view type identifiers (as strings) for which subtree traversal should be ignored.
+     *
+     * Views matching these types will have their subtrees skipped during redaction to avoid crashes
+     * caused by traversing problematic view hierarchies (e.g., views that activate internal CoreAnimation
+     * animations when their layers are accessed).
+     *
+     * The string values should match the result of `type(of: view).description()`.
```
philippsuess: `h`: For swizzleClassNameExcludes the string doesn't need to match the exact type, but we check if the whole class name contains the excluded string `className.contains(exclude) `

https://github.com/getsentry/sentry-cocoa/blob/f73c5c8240685bd4daf7ed546c41e5eb2d247b7c/Sources/Sentry/Public/SentryOptions.h#L518-L531

Would it make sense to do the same here or does need to be the exact type name?

Sources/SentrySwiftUI/Preview/PreviewRedactOptions.swift
philippsuess: `h`: I'm pretty sure most users don't know what subtree traversing is, and they shouldn't have to. We also have `unmaskedViewClasses`, which are already classes excluded from masking. I wonder if we shouldn't let users only have to worry about `unmaskedViewClasses`. `viewTypesIgnoredFromSubtreeTraversal` seems quite specific. What's the benefit of having both?

philprime:
@philipphofmann  and I have discussed the pull request review, and the confusion I need to resolve is mainly around **terminology and mental models** exposed to the user.

The core issue is that two **conceptually different mechanisms** are currently perceived as related because of naming:

1. **Masking-related configuration**
   - `maskedViewClasses`
   - `unmaskedViewClasses`

   These purely control **what gets redacted** in a screenshot.  
   `unmaskedViewClasses` can be used to “cut holes” into an otherwise masked area, i.e. explicitly allow certain subviews to remain visible even if their parent would normally be masked.

2. **Traversal-related configuration**
   - `viewTypesIgnoredFromSubtreeTraversal`

   This does **not control masking** as it's primary purpose.  
   Its purpose is to **stop view hierarchy traversal** for specific view types in order to avoid crashes (e.g. when accessing `view.sublayers` on problematic views).  
   When traversal is stopped, the children of that view effectively do not exist in the algorithm. As an implementation detail, the view itself is currently masked to ensure no data leaks, but that is secondary to its primary purpose.

The confusion arises because:
- The name `viewTypesIgnoredFromSubtreeTraversal` suggests a masking-related behavior (“ignored”), while it actually affects traversal control flow.
- From a user’s perspective, it is unclear whether views in this list are *masked*, *skipped*, or *both*.
- Users who only think in terms of “masking” should ideally not need to understand subtree traversal internals at all.

The action item on my side is therefore to:
- Improve **naming** so that traversal-stopping semantics are explicit and not conflated with masking.
- Improve **documentation** to clearly separate:
  - *what is masked* vs.
  - *where traversal stops and why*.

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
philprime: I've decided to rename it to `excludedViewClasses` to add additional view types that should be ignored from the subtree traversal, and `includedViewClasses` to overrule our defaults. That way we have to assignable sets which are merged

philippsuess:
That https://github.com/getsentry/sentry-cocoa/pull/7063#issuecomment-3686515062 summarizes it pretty well @philprime  what we discussed 👏 ⬆️

Sources/SentrySwiftUI/Preview/PreviewRedactOptions.swift
philprime: I agree, will change it

philprime: I am renaming it to `excludeViewClasses` to exclude them from view hierarchy traversal, and allow overruling using `includeViewClasses`.

Sources/SentrySwiftUI/Preview/PreviewRedactOptions.swift
philprime: Changed to partial matching for the `exclude` list with exact matching for the `include` list

benjaminpeters:
Tested on react-native using [this branch](https://github.com/benjaminpeters/sentry-react-native/tree/add-viewTypesIgnoredFromSubtreeTraversal) (local `sentry-cocoa` with your branch checked out) and everything looks to be working for me

philprime:
Thank you @benjaminpeters for your feedback! I am now back from my Christmas break and will continue working on getting this done in the next few days.

@itaybre I noticed you picked up the PR as discussed, the checks are green. Anything missing to get this approved?

Sources/Swift/Integrations/Screenshot/SentryScreenshotOptions.swift
```diff
@@ -133,7 +211,9 @@ public class SentryViewScreenshotOptions: NSObject, SentryRedactOptions {
             }) ?? DefaultValues.maskedViewClasses,
             unmaskedViewClasses: (dictionary["unmaskedViewClasses"] as? NSArray)?.compactMap({ element in
                 NSClassFromString((element as? String) ?? "")
-            }) ?? DefaultValues.unmaskedViewClasses
+            }) ?? DefaultValues.unmaskedViewClasses,
+            excludedViewClasses: (dictionary["excludedViewClasses"] as? [String]).map { Set($0) } ?? (dictionary["viewClassesExcludedFromSubtreeTraversal"] as? [String]).map { Set($0) } ?? DefaultValues.excludedViewClasses,
```
philprime: `m`: Seems like an unnecessary fallback added during refactoring

```suggestion
            excludedViewClasses: (dictionary["excludedViewClasses"] as? [String]).map { Set($0) } ?? DefaultValues.excludedViewClasses,
```

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -330,7 +407,9 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
             frameRate: (dictionary["frameRate"] as? NSNumber)?.uintValue,
             errorReplayDuration: (dictionary["errorReplayDuration"] as? NSNumber)?.doubleValue,
             sessionSegmentDuration: (dictionary["sessionSegmentDuration"] as? NSNumber)?.doubleValue,
-            maximumDuration: (dictionary["maximumDuration"] as? NSNumber)?.doubleValue
+            maximumDuration: (dictionary["maximumDuration"] as? NSNumber)?.doubleValue,
+            excludedViewClasses: (dictionary["excludedViewClasses"] as? [String]).map { Set($0) } ?? (dictionary["viewClassesExcludedFromSubtreeTraversal"] as? [String]).map { Set($0) },
```
philprime:
```suggestion
            excludedViewClasses: (dictionary["excludedViewClasses"] as? [String]).map { Set($0) },
```

itaybre (approved):
LGTM

Issue #7015, opened 2025-12-09
Ensure user facing options docs stay up-to-date


philipphofmann:
### Description

This issue aims to implement an automated process that ensures that when adding a new [`Options`](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Swift/Options.swift) option, that option is also added to the user-facing options [doc](https://docs.sentry.io/platforms/apple/configuration/options/).

One idea is to have a similar bot the [`danger.yml`](https://github.com/getsentry/sentry-cocoa/blob/main/.github/workflows/danger.yml) that fails when a PR doesn't include a proper changelog entry, but we can come up with any process that works for us.

Some edge cases to cover:

* Ignore specific options with a reason.
* The Cocoa SDK PR adding new options usually gets merged before merging the docs PR, because we often only merge the docs PR after a Cocoa SDK release.

philippsuess:
We might use a prompt as mentioned in https://github.com/getsentry/sentry-cocoa/issues/7014#issuecomment-3635920987 to automate this.

PR #7075, opened 2025-12-19
test: Add Options Documentation Sync Tests

philipphofmann:
This PR adds a unit test to ensure all public properties in `Options.swift` have public-facing user documentation. This helps us remember to document every new option we add for SDK users.

The test works like this

1. **Extract properties** from `Options.swift` using Objective-C runtime + Swift Mirror
2. **Fetch documentation** from the sentry-docs GitHub repo
3. **Compare** and fail if any options are missing documentation

When you add a new option and not a user facing doc, it fails like this

Screenshot 2025-12-19 at 09 39 22

For a temp fix, you can add any option to the `SentryOptionsDocumentationSyncTests.undocumentedOptions`. We only have to make sure to remove the added option to the `undocumentedOptions` after adding the docs.

Fixes GH-7015

### DIFF
--- a/Tests/SentryTests/OptionsInSyncWithDocs/SentryOptionsDocumentationSyncTests.swift
+++ b/Tests/SentryTests/OptionsInSyncWithDocs/SentryOptionsDocumentationSyncTests.swift
@@ -15,111 +15,27 @@ final class SentryOptionsDocumentationSyncTests: XCTestCase {
     /// to match the availability in Options.swift.
     private var undocumentedOptions: Set<String> {
         var options: Set<String> = [
-            // Internal/derived properties
             "parsedDsn",
-            "isTracingEnabled",
-            
-            // SDK behavior
-            "enabled",
-            "shutdownTimeInterval",
-            
-            // Crash handling
-            "enableCrashHandler",
-            "enableWatchdogTerminationTracking",
-            
-            // Session tracking
-            "sessionTrackingIntervalMillis",
-            
-            // Attachments
-            "maxAttachmentSize",
-            
-            // Performance tracing
-            "enableAutoPerformanceTracing",
-            "enableNetworkTracking",
-            "enableFileIOTracing",
-            "enableDataSwizzling",
-            "enableFileManagerSwizzling",
-            "enableCoreDataTracing",
-            "enableTimeToFullDisplayTracing",
-            
-            // App hangs
-            "enableAppHangTracking",
-            "appHangTimeoutInterval",
-            
-            // Swizzling
-            "enableSwizzling",
-            "swizzleClassNameExcludes",
-            
-            // URLSession
-            "urlSessionDelegate",
-            "urlSession",
-            
-            // Breadcrumbs
-            "enableAutoBreadcrumbTracking",
-            
-            // Failed requests
-            "failedRequestStatusCodes",
-            "failedRequestTargets",
-            
-            // Experimental
-            "swiftAsyncStacktraces",
-            "experimental",
-            
-            // Storage
-            "cacheDirectoryPath",
-            
-            // Spotlight
-            "enableSpotlight",
-            "spotlightUrl",
-            
-            // Callbacks
-            "beforeSendSpan",
-            "beforeSendLog",
-            "beforeCaptureScreenshot",
-            "beforeCaptureViewHierarchy",
-            "onCrashedLastRun",
-            
-            // Logs
-            "enableLogs"
+            "experimental"
         ]
         
-        // macOS only (#if os(macOS))
-        #if os(macOS)
-        options.insert("enableUncaughtNSExceptionReporting")
-        #endif
-        
-        // iOS/tvOS/visionOS with UIKit (#if (os(iOS) || os(tvOS) || os(visionOS)) && !SENTRY_NO_UIKIT)
         #if (os(iOS) || os(tvOS) || os(visionOS)) && !SENTRY_NO_UIKIT
-        options.insert("enableUIViewControllerTracing")
         options.insert("screenshot")
-        options.insert("reportAccessibilityIdentifier")
-        options.insert("enableUserInteractionTracing")
-        options.insert("enablePreWarmedAppStartTracing")
-        options.insert("enableReportNonFullyBlockingAppHangs")
         #endif
-        
-        // iOS/tvOS with UIKit (#if (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT)
+
         // Session replay (documented at https://docs.sentry.io/platforms/apple/session-replay/)
         #if (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT
         options.insert("sessionReplay")
         #endif
-        
-        // iOS only with UIKit (#if os(iOS) && !SENTRY_NO_UIKIT)
+
         #if os(iOS) && !SENTRY_NO_UIKIT
         options.insert("userFeedbackConfiguration") // @_spi(Private) - internal backing for configureUserFeedback
         options.insert("configureUserFeedback")
         #endif
-        
-        // Profiling - iOS/macOS/macCatalyst only (#if !(os(watchOS) || os(tvOS) || os(visionOS)))
-        #if !(os(watchOS) || os(tvOS) || os(visionOS))
+
+        #if !(os(tvOS) || os(visionOS))
         options.insert("profiling") // @_spi(Private) - internal backing for configureProfiling
         options.insert("configureProfiling")
-        #endif
-        
-        // MetricKit - iOS/macOS only (#if canImport(MetricKit) && !os(tvOS) && !os(visionOS))
-        #if canImport(MetricKit) && !os(tvOS) && !os(visionOS)
-        options.insert("enableMetricKit")
-        options.insert("enableMetricKitRawPayload")
         #endif
         
         return options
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/OptionsInSyncWithDocs/README.md
```diff
@@ -0,0 +1,19 @@
+# Options Documentation Sync Tests
+
+Automated tests to ensure all public properties in `Options.swift` have public-facing user documentation. This helps us remember to document every new option we add for SDK users.
+
+- Docs repo: [sentry-docs](https://github.com/getsentry/sentry-docs)
+- Options page: [options.mdx](https://github.com/getsentry/sentry-docs/blob/master/docs/platforms/apple/common/configuration/options.mdx)
+
+## How It Works
+
+1. **Extract properties** from `Options.swift` using Objective-C runtime + Swift Mirror
```
philprime, line 10: `m`: Have you considered using the `sdk_api.json` as it already performed the extraction tasks for us, including all public information? I believe using scripts and `jq` we could have a structured approach

Tests/SentryTests/OptionsInSyncWithDocs/SentryOptionsDocumentationSyncTests.swift
```diff
@@ -0,0 +1,202 @@
...
+            .filter { property in
+                // Check if there's a mapping with a documented name
+                guard let mapping = optionNameMappings.first(where: { $0.codeName == property }) else {
+                    return true // No mapping exists, property is missing
+                }
+                return !documentedOptions.contains(mapping.docsName)
+            }
+            .sorted()
+        
+        XCTAssertTrue(propertiesMissingDocs.isEmpty, """
```
philprime, line 88: Is the idea here that we don't let PRs pass unless (1) the new option are explicitly in the `undocumentedOptions` list or (2) if we added it to the docs already? 

`m`: If yes for 1, this seems like we are just building a list of undocumented options which has value, but we still need to create follow-up tasks for it. Maybe we can extend it with an auto-updating issue which always contains the current list of undocumented options as a todo list, similar to how the Renovate Dependencies bot creates a [dashboard issue](https://github.com/techprimate/Flinky/issues/10)

`h`: If yes for 2, I don't think we should enforce documentation for something which is not yet merged, nor released. If we update the docs to include a new option which is not yet publicly available, I believe it will confuse users.

Tests/SentryTests/OptionsInSyncWithDocs/README.md
philippsuess: Yes, I shortly considered it, but the major downside is that you need to use Xcode 16 and you need to parse the JSON file, which I find more complicated than using `property_getName` and Swift `Mirror`. Also, the JSON format seems to change across Xcode versions, and I couldn't find official docs for the JSON format. 
I also prefer using Swift whenever I can, because everybody working on this codebase, including contributors and contractors, knows Swift, but not everybody is fluent at scripting.

Tests/SentryTests/OptionsInSyncWithDocs/SentryOptionsDocumentationSyncTests.swift
philippsuess: I'd recommend option 1: use this as a reminder to add docs for new options. Merge the PR after adding the item to undocumented options, then create a follow-up issue to remove it once the docs are added. If needed, we can automate reminders later.

Tests/SentryTests/OptionsInSyncWithDocs/README.md
philprime: Ah I just noticed that the check runs as part of the test target, not as an additional script like `swiftlint` or `swiftformat`, because if it's a CLI script, we could just use `bash` with `jq` (JSON-Query) to query data like this:

```shell
$ jq '[.ABIRoot.children[] | select(.name == "Options") | .children[] | select(.kind == "Var") | .name]' sdk_api.json
[
  "dsn",
  "parsedDsn",
  "debug",
  "diagnosticLevel",
  "releaseName",
  "dist",
  "environment",
  "enabled",
  "shutdownTimeInterval",
  "enableCrashHandler",
  "enableSigtermReporting",
  "maxBreadcrumbs",
  "enableNetworkBreadcrumbs",
  "maxCacheItems",
...
]
```

As we enforce the sdk_api.json to be up-to-date before merging a PR, we always have the latest dump available, and no Xcode is needed.

philprime:
Based on this PR this could also be implemented as a bash script which is run in the pre-commit scripts and in CI:

```shell
#!/bin/bash

set -euo pipefail

# Colors for output
RED='\033[0;31m'
GREEN='\033[0;32m'
YELLOW='\033[1;33m'
NC='\033[0m' # No Color

# URLs and paths
SDK_API_JSON="sdk_api.json"
DOCS_URL="https://raw.githubusercontent.com/getsentry/sentry-docs/master/docs/platforms/apple/common/configuration/options.mdx"

# Options that are intentionally not documented (matching the Swift test)
UNDOCUMENTED_OPTIONS=(
    "parsedDsn"
    "experimental"
    "screenshot"
    "sessionReplay"
    "userFeedbackConfiguration"
    "configureUserFeedback"
    "profiling"
    "configureProfiling"
)

# Function to check if option is in undocumented list
is_undocumented() {
    local option="$1"
    for undoc in "${UNDOCUMENTED_OPTIONS[@]}"; do
        if [ "$option" = "$undoc" ]; then
            return 0
        fi
    done
    return 1
}

# Function to get mapped docs name (returns original if no mapping)
get_docs_name() {
    local option="$1"
    case "$option" in
        "inAppIncludes") echo "inAppInclude" ;;
        "enablePropagateTraceparent") echo "enable-propagate-trace-parent" ;;
        *) echo "$option" ;;
    esac
}

echo "📊 Comparing SentryOptions properties with documentation..."
echo ""

# Step 1: Extract property names from sdk_api.json
echo "1️⃣  Extracting properties from $SDK_API_JSON..."
if [ ! -f "$SDK_API_JSON" ]; then
    echo -e "${RED}❌ Error: $SDK_API_JSON not found${NC}"
    exit 1
fi

SDK_OPTIONS=$(jq -r '.ABIRoot.children[] | select(.name == "Options") | .children[] | select(.kind == "Var") | .name' "$SDK_API_JSON" | sort)
SDK_OPTIONS_COUNT=$(echo "$SDK_OPTIONS" | wc -l | tr -d ' ')
echo -e "${GREEN}   Found $SDK_OPTIONS_COUNT properties${NC}"
echo ""

# Step 2: Fetch and parse documentation from sentry-docs
echo "2️⃣  Fetching documentation from GitHub..."
DOCS_CONTENT=$(curl -fsSL "$DOCS_URL")
if [ $? -ne 0 ]; then
    echo -e "${RED}❌ Error: Failed to fetch documentation from $DOCS_URL${NC}"
    exit 1
fi

# Extract option names from MDX (matches <SdkOption name="optionName"> or <SdkOption name='optionName'>)
DOCUMENTED_OPTIONS=$(echo "$DOCS_CONTENT" | grep -o '<SdkOption name=['"'"'"][^'"'"'"]*['"'"'"]' | sed -E 's/.*name=['"'"'"]([^'"'"'"]*)['"'"'"].*/\1/' | sort -u)
DOCUMENTED_OPTIONS_COUNT=$(echo "$DOCUMENTED_OPTIONS" | wc -l | tr -d ' ')
echo -e "${GREEN}   Found $DOCUMENTED_OPTIONS_COUNT documented options${NC}"
echo ""

# Step 3: Compare and find missing documentation
echo "3️⃣  Comparing options..."
echo ""

MISSING_DOCS=()
SHOULD_REMOVE_FROM_UNDOCUMENTED=()

# Check each SDK option
while IFS= read -r option; do
    # Skip empty lines
    [ -z "$option" ] && continue

    # Get the docs name (might be mapped)
    docs_name=$(get_docs_name "$option")

    # Check if documented
    if echo "$DOCUMENTED_OPTIONS" | grep -q "^${docs_name}$"; then
        # Option is documented
        if is_undocumented "$option"; then
            SHOULD_REMOVE_FROM_UNDOCUMENTED+=("$option")
        fi
    else
        # Option is not documented
        if ! is_undocumented "$option"; then
            MISSING_DOCS+=("$option")
        fi
    fi
done <<< "$SDK_OPTIONS"

# Step 4: Report results
echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
echo ""

if [ ${#SHOULD_REMOVE_FROM_UNDOCUMENTED[@]} -gt 0 ]; then
    echo -e "${YELLOW}⚠️  The following options are now documented and can be removed from undocumented list:${NC}"
    echo ""
    for option in "${SHOULD_REMOVE_FROM_UNDOCUMENTED[@]}"; do
        echo "   - $option"
    done
    echo ""
fi

if [ ${#MISSING_DOCS[@]} -gt 0 ]; then
    echo -e "${RED}❌ The following Options properties are not documented in sentry-docs:${NC}"
    echo ""
    for option in "${MISSING_DOCS[@]}"; do
        echo "   - $option"
    done
    echo ""
    echo "To fix this:"
    echo "1. Add documentation for these options in sentry-docs:"
    echo "   https://github.com/getsentry/sentry-docs/blob/master/docs/platforms/apple/common/configuration/options.mdx"
    echo ""
    echo "2. OR add them to the UNDOCUMENTED_OPTIONS array in this script with a reason."
    echo ""
    exit 1
else
    echo -e "${GREEN}✅ All Options properties are documented or explicitly marked as undocumented!${NC}"
    echo ""
fi
```

philprime (approved):
I am going to approve this so you can proceed as you believe best, but I still want to highlight again that I don't believe this needs to be implemented in unit tests with dynamic member lookup, as we already rely on the `sdk_api.json` for making sure we are aware of changing the public API, and a script is simpler to maintain than this much Swift parsing code.

philippsuess:
After an internal Slack discussion, we also think that Bash isn't ideal. Swift is good enough for now to validate this approach. We can always change it. We could use something like Ruby if we wanted to extend this to multiple repos.

PR #7076, opened 2025-12-19
ref: Convert SentryMetricKitIntegration to Swift

noahsmartin:
Resolves https://linear.app/getsentry/issue/COCOA-989/refactor-sentrymetrickitintegrationm-in-swift

I converted most of this in [my last PR](https://github.com/getsentry/sentry-cocoa/pull/6908) so this just follows up to convert the rest

There is one logic change here. If MetricKit ever delivered a CPU exception event with `callStackPerThread = true` we would now still handle it. It's not expected to happen, but there's no reason to have a guard case specifically for it, just like we don't have any for app hangs.

### DIFF
--- a/Sources/Swift/Core/MetricKit/SentryMXCallStackTree.swift
+++ b/Sources/Swift/Core/MetricKit/SentryMXCallStackTree.swift
@@ -7,7 +7,7 @@ import Foundation
 struct SentryMXCallStackTree: Decodable {
     
     let callStacks: [SentryMXCallStack]
-    public let callStackPerThread: Bool
+    let callStackPerThread: Bool
     
     static func from(data: Data) throws -> SentryMXCallStackTree {
         return try JSONDecoder().decode(SentryMXCallStackTree.self, from: data)
--- a/Sources/Swift/Core/MetricKit/SentryMetricKitIntegration.swift
+++ b/Sources/Swift/Core/MetricKit/SentryMetricKitIntegration.swift
@@ -73,7 +73,7 @@ private let hangDiagnosticMechanism = "mx_hang_diagnostic"
 @available(macOS 12.0, *)
 extension SentryMetricKitIntegration: SentryMXManagerDelegate {
     func didReceiveCrashDiagnostic(_ diagnostic: MXCrashDiagnostic, callStackTree: SentryMXCallStackTree, timeStampBegin: Date) {
-        let exceptionValue = "MachException Type:\(diagnostic.exceptionType ?? 0) Code:\(diagnostic.exceptionCode ?? 0) Signal:\(diagnostic.signal ?? 0)"
+        let exceptionValue = "MachException Type:\(String(describing: diagnostic.exceptionType)) Code:\(String(describing: diagnostic.exceptionCode)) Signal:\(String(describing: diagnostic.signal))"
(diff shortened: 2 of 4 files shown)


Sources/Swift/Core/Integrations/Integrations.swift
```diff
@@ -41,15 +41,21 @@ private struct AnyIntegration {
...
+        integrations.append(.init(FlushLogsIntegration.self))
         #endif
 
         #if (os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)) && !SENTRY_NO_UIKIT
-        integrations.append(.init(SentryScreenshotIntegration<SentryDependencyContainer>.self))
+        integrations.append(.init(SentryScreenshotIntegration.self))
+        #endif
+        
+        #if os(iOS) || os(macOS)
+        if #available(macOS 12.0, *) {
```
itaybre, line 56: it requires UIKit?

Sources/Swift/Core/MetricKit/SentryMXCallStackTree.swift
```diff
@@ -4,8 +4,7 @@ import Foundation
 /**
  * JSON specification of MXCallStackTree can be found here https://developer.apple.com/documentation/metrickit/mxcallstacktree/3552293-jsonrepresentation
  */
-@objcMembers
-@_spi(Private) public class SentryMXCallStackTree: NSObject, Decodable {
+struct SentryMXCallStackTree: Decodable {
```
itaybre, line 7: 👏

Sources/Swift/Core/MetricKit/SentryMetricKitIntegration.swift
```diff
@@ -0,0 +1,116 @@
...
+    }
+    
+    func didReceiveCpuExceptionDiagnostic(_ diagnostic: MXCPUExceptionDiagnostic, callStackTree: SentryMXCallStackTree, timeStampBegin: Date) {
+        let totalCPUTime = measurementFormatter.string(from: diagnostic.totalCPUTime)
+        let totalSampledTime = measurementFormatter.string(from: diagnostic.totalSampledTime)
+
+        let exceptionValue = "MXCPUException totalCPUTime:\(totalCPUTime) totalSampledTime:\(totalSampledTime)"
+        let event = Self.createEvent(handled: true, level: .warning, exceptionValue: exceptionValue, exceptionType: "MXCPUException", exceptionMechanism: cpuExceptionMechanism, timeStampBegin: timeStampBegin)
+        capture(event: event, handled: true, callStackTree: callStackTree, diagnosticJSON: diagnostic.jsonRepresentation())
+    }
```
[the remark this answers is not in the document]
itaybre: @noahsmartin check if this is relevant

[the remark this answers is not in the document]
noahsmartin: It was unnecessary, we can remove it

Sources/Swift/Core/Integrations/Integrations.swift
noahsmartin: Keeping this logic the same as before, we only support metric kit on iOS/macOS. It's a bit arbitrary, we could support on other platforms, but no-one has tested it

PR #7077, opened 2025-12-22
refactor: Add attributable protocol for typed attribute values

philprime:
## :scroll: Description

This PR introduces a protocol-based system for typed attribute values to improve type safety when working with structured logging attributes. The changes add:

- A type-safe enum `SentryAttributeValue` that represents attribute values (string, boolean, integer, double, and their array variants)
- A protocol `SentryAttributeValuable` that allows types to convert themselves to `SentryAttributeValue`
- Extensions on `String`, `Bool`, `Int`, `Double`, and `Float` to conform to `SentryAttributeValuable`

This change refactors the Objective-C compatible `SentryAttribute` to use the Swift enum `SentryAttributeValue` as its backing type. It also updates `BatcherScope` to use `SentryAttributeValue` directly instead of `SentryAttribute` objects.

This PR also introduces support for Array attribute types as defined [in the documentation](https://develop.sentry.dev/sdk/telemetry/attributes/).

**Changes in Public API:**

The currently proposed changes contain new public API endpoints for the `SentryLogger` to accept `[String: SentryAttributeValuable]` instead of `[String: Any]` for improved type safety. **I acknowledge this is a breaking change**, therefore I am open to remove it until the next major version.

If we keep it it could cause ambiguity issues because we have to method signatures both with the same parameters, only having different Dictionary types as the attributes.

The main use case are the new methods introduced by Metrics in #6957

Previously, attributes were stored as `[String: Any]` dictionaries, which provided no compile-time type safety. This made it easy to accidentally pass unsupported types or incorrect values, which are converted to `String` as a fallback, sending unpredictable data. By introducing a protocol-based system with `SentryAttributeValuable`, we can:

1. Improve the type-safety as the compiler now enforces that only supported types (String, Bool, Int, Double, Float) can be used as attribute values
2. Simplify API usage as developers can directly pass native Swift types without wrapping them in `SentryAttribute` objects (using th `ExpressibleBy...` protocol extensions)
3. Maintain backward compatibility the `SentryAttribute` class still exists and can convert from `SentryAttributeValue` for Objective-C interop.

I also need this change to implement #6957.

- Updated existing unit tests in `BatcherScopeTests`, `BatcherTests`, and `SentryLoggerTests` to use the new typed attribute system
- All tests pass with the new protocol-based approach
- Verified that the changes maintain the same runtime behavior while improving compile-time safety

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Tools/Batcher/BatcherScope.swift
+++ b/Sources/Swift/Tools/Batcher/BatcherScope.swift
@@ -32,14 +32,21 @@ extension BatcherScope {
         config: Config,
         metadata: Metadata
     ) {
-        addDefaultAttributes(to: &item.attributesDict, config: config, metadata: metadata)
-        addOSAttributes(to: &item.attributesDict, config: config)
-        addDeviceAttributes(to: &item.attributesDict, config: config)
-        addUserAttributes(to: &item.attributesDict, config: config)
-        addReplayAttributes(to: &item.attributesDict, config: config)
-        addScopeAttributes(to: &item.attributesDict, config: config)
-        addDefaultUserIdIfNeeded(to: &item.attributesDict, config: config, metadata: metadata)
+        // Extract attributesDict once to avoid multiple getter/setter calls on computed property
+        // Each inout parameter access triggers both getter and setter, which is expensive for
+        // computed properties that perform dictionary conversions.
+        var attributes = item.attributesDict
+        
+        addDefaultAttributes(to: &attributes, config: config, metadata: metadata)
+        addOSAttributes(to: &attributes, config: config)
+        addDeviceAttributes(to: &attributes, config: config)
+        addUserAttributes(to: &attributes, config: config)
+        addReplayAttributes(to: &attributes, config: config)
+        addScopeAttributes(to: &attributes, config: config)
+        addDefaultUserIdIfNeeded(to: &attributes, config: config, metadata: metadata)
 
+        // Set the modified dictionary back once
+        item.attributesDict = attributes
         item.traceId = propagationContextTraceId
     }
 


philprime:
@sentry review
@cursor review

Sources/Swift/Tools/Batcher/BatcherItem.swift
```diff
@@ -1,5 +1,5 @@
 protocol BatcherItem: Encodable {
-    var attributes: [String: SentryAttribute] { get set }
+    var attributesMap: [String: SentryAttributeValue] { get set }
```
philprime: This was renamed because it would otherwise collide with the type of `SentryAttribute.attributes`

Sources/Swift/Tools/SentryLogger.swift
```diff
@@ -173,13 +209,22 @@ public final class SentryLogger: NSObject {
     
     // MARK: - Private
     
-    private func captureLog(level: SentryLog.Level, logMessage: SentryLogMessage, attributes: [String: Any]) {
+    private func captureLog(
+        level: SentryLog.Level,
+        logMessage: SentryLogMessage,
+        attributes: [String: Any]
+    ) {
```
philprime:
```suggestion
    private func captureLog(level: SentryLog.Level, logMessage: SentryLogMessage, attributes: [String: Any]) {
```

Sources/Swift/Tools/SentryLogger.swift
```diff
@@ -51,7 +51,18 @@ public final class SentryLogger: NSObject {
     public func trace(_ message: SentryLogMessage, attributes: [String: Any] = [:]) {
         captureLog(level: .trace, logMessage: message, attributes: attributes)
     }
-    
+
+    /// Logs a trace-level message with structured string interpolation and optional attributes.
```
philprime: These newly introduced logger methods are not strictly necessary but can be beneficial for SDK users due to type hints. We still need the type-erased method to keep supporting Objective-C but eventually we should deprecate the Any-typed version for Swift and have an Objective-C wrapper

philprime:
Array handling is incomplete, moved back to draft

philprime:
@sentry review
@cursor review

Sources/Swift/Tools/Batcher/BatcherScope.swift
```diff
@@ -6,8 +6,13 @@ protocol BatcherScope {
     var span: Span? { get }
     var userObject: User? { get }
     func getContextForKey(_ key: String) -> [String: Any]?
+
+    /// List of attributes with erased value type for compatibility with public ``Scope``.
     var attributes: [String: Any] { get }
```
itaybre, line 11: Should we add `@deprecated`?

philprime: Not sure why we would want to do that. It's not deprecated, the `attributesMap` is just because for the Metrics API it won't be the type `[String: Any]`

Tests/SentryTests/Protocol/SentryAttributeTests.swift
```diff
@@ -46,6 +46,82 @@ final class SentryAttributeTests: XCTestCase {
         XCTAssertEqual(doubleValue, 3.14159, accuracy: 0.00001)
     }
     
+    func testEncodeStringArrayAttribute() throws {
+        // -- Arrange --
+        let attribute = SentryLog.Attribute(stringArray: ["hello", "world", "test"])
```
philippsuess: `l`: As we're not only going to use this for logs but also for metrics, I would rather use this naming in the tests
```suggestion
        let attribute = SentryAttribute(stringArray: ["hello", "world", "test"])
```

Sources/Swift/Protocol/SentryAttribute.swift
```diff
@@ -40,7 +40,67 @@ public final class SentryAttribute: NSObject {
         super.init()
     }
 
-    internal init(value: Any) {
+    public init(stringArray values: [String]) {
+        self.type = "string[]"
+        self.value = values
+        super.init()
+    }
```
philippsuess, line 82: `h`: I would highly appreciate proper code docs for all these new init methods, as these are public API.

Sources/Swift/Protocol/SentryAttribute.swift
```diff
@@ -40,7 +40,67 @@ public final class SentryAttribute: NSObject {
...
+
+    public init(booleanArray values: [Bool]) {
+        self.type = "boolean[]"
```
philippsuess: `m`: What about defining an internal string-based enum for all the allowed types? So something like
```Swift
private enum AttributeType: String {
    case stringType = "string",
         booleanType = "boolean",
         integerType = "integer",
         integerArrayType = "integer[]",
         doubleType = "double",
         doubleArrayType = "double[]"
}
```

SentryAttributeValue could also use it for the types.

Tests/SentryTests/Protocol/SentryAttributeValuableTests.swift
```diff
@@ -0,0 +1,421 @@
+@testable import Sentry
+import XCTest
+
+final class SentryAttributeValuableTests: XCTestCase {
```
philippsuess: Thanks for the clean test coverage.

Sources/Swift/Protocol/SentryAttributeValuable.swift
```diff
@@ -0,0 +1,145 @@
+public protocol SentryAttributeValuable {
+    var asAttributeValue: SentryAttributeValue { get }
```
philippsuess: `m`: We have sentry in the naming everywhere. I think this would make it a bit clearer that this returns a SentryAttributeValue.
```suggestion
    var asSentryAttributeValue: SentryAttributeValue { get }
```

Sources/Swift/Protocol/SentryAttributeValuable.swift
```diff
@@ -0,0 +1,145 @@
...
+        }
+        if let values = self as? [SentryAttributeValuable] {
+            return castArrayToAttributeValue(values: values)
+        }
+        return .stringArray(self.map { element in
+            String(describing: element)
+        })
+    }
+
+    private func castArrayToAttributeValue(values: [SentryAttributeValuable]) -> SentryAttributeValue {
```
philippsuess: `m`: Honestly, I find that a bit confusing. I can do this, and the compiler doesn't complain. When looking at this code, I could assume that we support arrays with different types for the attributes, but instead, we silently convert them to a string array. Instead, it would be great if the following code wouldn't even compile.
```
let attribute : [SentryAttributeValuable] = [1,2,3, 3.3, "test"]
let attributeValue = attribute.asAttributeValue
```

Sources/Swift/Tools/SentryLogger.swift
```diff
@@ -51,7 +51,18 @@ public final class SentryLogger: NSObject {
...
+    public func trace(_ message: SentryLogMessage, attributes: [String: SentryAttributeValuable] = [:]) {
+        captureLog(level: .trace, logMessage: message, attributes: attributes)
+    }
```
philippsuess: `h`: I'm all in favor of adding type safety. This is the API our users interact with most, and currently, this approach could do a better job of communicating to our users which types they can use. When I declare the attributes, I see  `SentryAttributeValuable`, which I find a bit confusing. What is `SentryAttributeValuable`? Which types can I use? 

OK, so this works:
```
let attributes: [String: SentryAttributeValuable] = ["retry_policy": "exponential"])]
```

This leads to two compiler errors, which is excellent:

```
let attributes: [String: SentryAttributeValuable] = ["string": "hello", "double": 0.0, "int": 1, "bool": true, "array": [1,2,3, 3.3], "dict": ["hello": "you"], "null": NSNull()]
```
Image

But if I do this, the code compiles. It should also fail to compile here. It doesn't because it falls back to using `[String: Any]` for the attributes.
```
SentrySDK.logger.warn(logString, attributes: ["string": "hello", "double": 0.0, "int": 1, "bool": true, "array": [1,2,3, 3.3], "dict": ["hello": "you"], "null": NSNull()])
```

philipphofmann (commented):
I really like that we're adding type safety to the attributes with this PR. I think we need some slight tweaking for the public logging API to better communicate which types our users can use, such as:
```
public func warn(_ message: SentryLogMessage, attributes: [String: SentryAttributeValuable] = [:])
```

Sources/Swift/Tools/SentryLogger.swift
philprime: That's what I am afraid off, we can't really offer this API for Logs, because it's so ambiguous with `Any`. I actually fee like we shouldn't introduce this for Logs right now and instead when we have a Objective-C wrapper framework.

Sources/Swift/Protocol/SentryAttributeValuable.swift
philprime: I tried to do that but it's not possible in Swift to have multiple conformances like this:

```swift
extension Array: SentryAttributeValuable where Element == Bool {
    public var asAttributeValue: SentryAttributeValue {
      return .booleanArray(self)
    }
}

extension Array: SentryAttributeValuable where Element == Int {
    public var asAttributeValue: SentryAttributeValue {
      return .integerArray(self)
    }
}

extension Array: SentryAttributeValuable where Element == String {
    public var asAttributeValue: SentryAttributeValue {
      return .stringArray(self)
    }
}

extension Array: SentryAttributeValuable where Element == Double {
    public var asAttributeValue: SentryAttributeValue {
      return .doubleArray(self)
    }
}
```

...because Swift does not allow multiple conditional conformances of the same generic type to the same protocol — even if the constraints are mutually exclusive. This is a hard compiler rule, not a limitation of my code.

If you, @itaybre or @noahsmartin know how we can make only these types of arrays can actually be casted to `SentryAttributeValuable`, I happily integrate it. I've already tried it before and couldn't find a way.

Sources/Swift/Tools/SentryLogger.swift
philippsuess: We could make it work by using a different name for attributes such as typeSafeAttributes, or typedAttributes and even consider using `_ typeSafeAttributes`.
```
public func warn(_ message: SentryLogMessage, _ typeSafeAttributes: [String: SentryAttributeValuable] = [:]) {
```

And I think we could rename `SentryAttributeValuable` to `SentryTypedAttribute` or `SentryTypeSafeAttribute`.

Sources/Swift/Protocol/SentryAttributeValuable.swift
philippsuess: OK, maybe we need a completely different approach if it doesn't work with the current one. I can try to come up with something, but that will take some time, and that will have to wait after the holidays.

philprime: @philipphofmann what is your rough ETA for proposing an alternative?

Sources/Swift/Tools/SentryLogger.swift
philprime: I would not use `_` because named arguments are more deterministic. So it would be `typeSafeAttributes` which I think is clunky. TBH I would just remove it for Logs right now, and use it with Metrics. We can always adopt typed attributes for Logs in a future PR

Sources/Swift/Protocol/SentryAttribute.swift
philprime: This should have been feedback when attributes where first introduced, but yes I can add them here.

Tests/SentryTests/Protocol/SentryAttributeTests.swift
philprime: Thanks for pointing this out

Sources/Swift/Tools/SentryLogger.swift
philippsuess: I agree (also @itaybre) that we can only use this new API for metrics, because it allows iteration, cause it's still experimental. We can add it to logs later.

Sources/Swift/Protocol/SentryAttributeValuable.swift
philippsuess: Thanks for the ping. I can do that tomorrow. Please keep pinging me if I don't.

philippsuess: Here's why I find this API a bit confusing:
```
public func warn(_ message: SentryLogMessage, attributes: [String: SentryAttributeValuable] = [:])
```

When I call this method with the wrong types, I get this error message in Xcode, which isn't really actionable:
Screenshot 2026-01-08 at 09 07 18

More importantly, with the current solution, code completion doesn't prompt me which types I can use.


What if we made this fully type safe? 

An attribute has to have a key: String and a value. What about making this a list of attributes instead, so the API for calling logs or metrics then looks like:
```swift
func testFullyTypeSafe() throws {
    logWarn("message", attributes: [
        SentryAttribute("hello", .string("world")),
        SentryAttribute("pi", .doubleArray([12.2, 43.3])),
        SentryAttribute("isTrue", .bool(true))
    ])
}


func logWarn(_ message: String, attributes: [SentryAttribute]) {

}
```


With this approach, you get full type safety and autocomplete code suggestions when you write the code. 
Screenshot 2026-01-08 at 09 11 17



We could also shorten the SentryAttribute to simply Attribute. You could also then write

```swift
logWarn("message", attributes: [
    .init("hello", .string("world")),
    .init("pi", .doubleArray([12.2, 43.3])),
    .init("isTrue", .bool(true)),
])
```


It's similar to what we have right now. I mostly copied this approach from https://github.com/apple/swift-configuration/blob/main/Sources/Configuration/ConfigProvider.swift
```
@frozen public enum SentryAttributeType: String, Sendable, Equatable, Hashable {

    /// A string value.
    case string

    /// An integer value.
    case int

    /// A double value.
    case double

    /// A Boolean value.
    case bool

    /// An array of bytes.
    case bytes

    /// An array of string values.
    case stringArray

    /// An array of integer values.
    case intArray

    /// An array of double values.
    case doubleArray

    /// An array of Boolean values.
    case boolArray

    /// An array of byte chunks.
    case byteChunkArray
}

@frozen public enum SentryAttributeContent: Sendable, Equatable, Hashable {

    /// A string value.
    case string(String)

    /// An integer value.
    case int(Int)

    /// A double value.
    case double(Double)

    /// A Boolean value.
    case bool(Bool)

    /// An array of bytes.
    case bytes([UInt8])

    /// An array of string values.
    case stringArray([String])

    /// An array of integer values.
    case intArray([Int])

    /// An array of double values.
    case doubleArray([Double])

    /// An array of Boolean value.
    case boolArray([Bool])

    /// An array of byte arrays.
    case byteChunkArray([[UInt8]])

    /// The configuration type of this content.
    var type: SentryAttributeType {
        switch self {
        case .string:
            return .string
        case .int:
            return .int
        case .double:
            return .double
        case .bool:
            return .bool
        case .bytes:
            return .bytes
        case .stringArray:
            return .stringArray
        case .intArray:
            return .intArray
        case .doubleArray:
            return .doubleArray
        case .boolArray:
            return .boolArray
        case .byteChunkArray:
            return .byteChunkArray
        }
    }

    /// A string description of the underlying value without type information.
    var underlyingValueDescription: String {
        switch self {
        case .string(let value):
            return value.description
        case .int(let value):
            return value.description
        case .double(let value):
            return value.description
        case .bool(let value):
            return value.description
        case .bytes(let value):
            // Show the first 32 bytes as hexadecimal.
            let hex = value.prefix(32)
                .map {
                    let hexString = String($0, radix: 16)
                    return hexString.count == 1 ? "0" + hexString : hexString
                }
                .joined()
            return "\(value.count) bytes, prefix: \(hex)"
        case .stringArray(let values):
            return values.map(\.description).joined(separator: ", ")
        case .intArray(let values):
            return values.map(\.description).joined(separator: ", ")
        case .doubleArray(let values):
            return values.map(\.description).joined(separator: ", ")
        case .boolArray(let values):
            return values.map(\.description).joined(separator: ", ")
        case .byteChunkArray(let values):
            let descriptions = values.map { value in
                // Show the first 32 bytes as hexadecimal.
                let hex = value.prefix(32)
                    .map {
                        let hexString = String($0, radix: 16)
                        return hexString.count == 1 ? "0" + hexString : hexString
                    }
                    .joined()
                let count = value.count
                return "\(count) bytes, prefix: \(hex)"
            }
            return descriptions.joined(separator: ", ")
        }
    }

    /// An error thrown when the actual type doesn't match the requested type.
    internal struct UnwrapError: Error {

        /// The actual type of the configuration value.
        var actualType: SentryAttributeType

        /// The requested type.
        var requestedType: SentryAttributeType
    }

    /// Returns the string value, or throws an error if the content is not a string.
    var asString: String {
        get throws {
            guard case .string(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .string)
            }
            return value
        }
    }

    /// Returns the integer value, or throws an error if the content is not an integer.
    var asInt: Int {
        get throws {
            guard case .int(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .int)
            }
            return value
        }
    }

    /// Returns the double value, or throws an error if the content is not a double.
    var asDouble: Double {
        get throws {
            guard case .double(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .double)
            }
            return value
        }
    }

    /// Returns the Boolean value, or throws an error if the content is not a Boolean.
    var asBool: Bool {
        get throws {
            guard case .bool(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .bool)
            }
            return value
        }
    }

    /// Returns the byte array, or throws an error if the content is not a byte array.
    var asBytes: [UInt8] {
        get throws {
            guard case .bytes(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .bytes)
            }
            return value
        }
    }

    /// Returns the array of strings, or throws an error if the content is not an array of strings.
    var asStringArray: [String] {
        get throws {
            guard case .stringArray(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .stringArray)
            }
            return value
        }
    }

    /// Returns the array of integers, or throws an error if the content is not an array of integers.
    var asIntArray: [Int] {
        get throws {
            guard case .intArray(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .intArray)
            }
            return value
        }
    }

    /// Returns the array of doubles, or throws an error if the content is not an array of doubles.
    var asDoubleArray: [Double] {
        get throws {
            guard case .doubleArray(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .doubleArray)
            }
            return value
        }
    }

    /// Returns the array of Booleans, or throws an error if the content is not an array of Booleans.
    var asBoolArray: [Bool] {
        get throws {
            guard case .boolArray(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .boolArray)
            }
            return value
        }
    }

    /// Returns the array of byte arrays, or throws an error if the content is not an array of byte arrays.
    var asByteChunkArray: [[UInt8]] {
        get throws {
            guard case .byteChunkArray(let value) = self else {
                throw UnwrapError(actualType: type, requestedType: .byteChunkArray)
            }
            return value
        }
    }
}



struct SentryAttribute: Sendable, Equatable, Hashable {
    /// The configuration content.
   public var content: SentryAttributeContent
    public let key: String

    public init(_ key: String, _ content: SentryAttributeContent) {
        self.key = key
        self.content = content
    }
}
```

Let me know if you think a fully type safe approach is better or you think it's not a great idea.

philippsuess: The outcome of an internal call from (@philprime and @philipphofmann)  is to keep protocol oriented approach, because it allows a more concise way of passing the attributes. Furthermore, the [String:SentryAttributeValuable] clearly communicates that it is a dict and you can't have duplicate keys.

To make the API easier to understand, we decided to rename `SentryAttributeValuable` to `SentryAttributeValue`, because that is what the user cares about, and 95% the users see. Furthermore, we want to rename the enum `SentryAttributeValue` to `SentryAttributeContent`, mostly to have a different name from `SentryAttributeValue`. We also have to rename the `asSentryAttributeValue` to `asSentryAttributeContent`.

One minor disadvantage of keeping the current approach is that we can't enforce compile-time safety to avoid mixed arrays, because the array extension can't be scoped to single types. Swift converts an array with multiple types to an array of `Any`. But we accept this trade-off as we can convert the values to strings and not loose any data.

Sources/Swift/Protocol/SentryAttribute.swift
```diff
@@ -4,103 +4,250 @@
 /// Supports String, Bool, Int, and Double types.
 @objcMembers
 public final class SentryAttribute: NSObject {
```
philippsuess, line 13: `l`: What about adding a comment for users but also SDK maintainers explaining why we now have both SentryAttribute inheriting from NSObject and the other SentryAttributeValue. So basically that this is for ObjC and the other for Swift and the ObjC version will be removed in the future.

Sources/Swift/Protocol/SentryAttributeValuable.swift
```diff
@@ -0,0 +1,203 @@
+/// A protocol that represents values that can be used as structured logging attributes.
+///
+/// `SentryAttributeValue` provides a protocol-oriented approach for accepting attribute values
+/// in public APIs. This allows APIs to accept a wide variety of types without requiring a concrete
+/// enum type, making the API more flexible and user-friendly.
```
philippsuess: `l`: I would somewhere in here also mention SentryAttribute inheriting from NSObject and briefly explain why we need both.

Sources/Swift/Protocol/SentryAttributeValuable.swift
```diff
@@ -0,0 +1,203 @@
+/// A protocol that represents values that can be used as structured logging attributes.
```
philippsuess: `m`: I think we should also rename this file to `SentryAttributeValue` and the `SentryAttributeValue.swift` file to `SentryAttributeContent.swift`.

Sources/Swift/Protocol/SentryAttributeValuable.swift
```diff
@@ -0,0 +1,203 @@
...
+/// - `Double`
+/// - `Float` (converted to Double)
+/// - `Array` (with various element types)
+///
+/// ## Implementation
+///
+/// Conforming types implement `asSentryAttributeContent` to convert themselves to a
+/// `SentryAttributeContent` enum value, which provides the structured representation used
+/// internally for encoding and type identification.
+public protocol SentryAttributeValue {
```
philippsuess: `m`: As we're only going to use this for the experimental logging API, I think we should also add a comment to this protocol and the SentryAttributeContent to mention that they are still experimental, so could still change it in the future. WDTY?

Sources/Swift/Protocol/SentryAttributeValuable.swift
```diff
@@ -0,0 +1,203 @@
...
+
+extension Array: SentryAttributeValue {
+    /// Converts an array to a SentryAttributeContent value.
+    ///
+    /// This extension cannot be scoped to `where Element == SentryAttributeValue` because:
+    /// - Mixed arrays (arrays containing elements of different types) are automatically converted to `[Any]` by Swift
+    /// - If we used `where Element == SentryAttributeValue`, mixed arrays would not compile, preventing users
+    ///   from passing heterogeneous arrays
+    /// - We accept this trade-off: while we can't enforce compile-time safety for mixed arrays, we can convert
+    ///   the values to strings at runtime without losing any data
```
philippsuess: Thanks 💯

Sources/Swift/Protocol/SentryAttributeValuable.swift
```diff
@@ -0,0 +1,203 @@
...
+        }
+        if let doubleArray = castValuesToDoubleArray(values) {
+            return doubleArray
+        }
+        if let integerArray = castValuesToIntegerArray(values) {
+            return integerArray
+        }
+        if let stringArray = castValuesToStringArray(values) {
+            return stringArray
+        }
```
philippsuess: `l`: Can't we avoid iterating over the whole array and cast individual items, by applying a similar logic as above in `asSentryAttributeContent`? We could check the type of the first array element and only try to cast the full array to booleans if the first array is a boolean? For an stringArray here we try to cast all items first to bools then double then int and then finally to string. We could also solve this by doing an early exit in the `castValuesTo..` methods below. Maybe it's also a micro optimization.

Sources/Swift/Protocol/SentryAttributeValue.swift
```diff
@@ -0,0 +1,206 @@
...
+    case integerArray = "integer[]"
+    case doubleArray = "double[]"
+}
+
+/// A type-safe representation of attribute values used by structured logging.
+///
+/// `SentryAttributeContent` provides a strongly-typed enum for representing various attribute types
+/// including strings, booleans, integers, doubles, and their array variants. This enum conforms to
+/// `Equatable`, `Hashable`, and `Encodable` for use in structured data contexts.
+public enum SentryAttributeContent: Equatable, Hashable {
```
philippsuess: `l`: Maybe worth adding a hint that users shouldn't have to interact with this, but rather the `SentryAttributeValue` protocol.

Sources/Swift/Protocol/SentryAttributeValue.swift
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+extension SentryAttributeContent: ExpressibleByStringLiteral {
+    /// Creates a string attribute value from a string literal.
+    ///
+    /// This allows `SentryAttributeContent` to be initialized directly from string literals:
+    /// ```swift
+    /// let value: SentryAttributeContent = "hello"
+    /// ```
```
philippsuess: `l`: Maybe worth mentioning that we use this is only really useful when passing a dict of SentryAttributeValues via the dict on the metric APIs.

Sources/Swift/Tools/Batcher/BatcherScope.swift
```diff
@@ -6,8 +6,13 @@ protocol BatcherScope {
...
 
+    /// Used for type-safe access of the attributes, uses default implementation in extension
+    var attributesMap: [String: SentryAttributeContent] { get }
```
philippsuess: `l`: In Swift we usually use dictionaries and not maps. The same applies to Sources/Swift/Tools/SentryLogBatcher.swift.
```suggestion
    var attributesDict: [String: SentryAttributeContent] { get }
```

Sources/Swift/Protocol/SentryAttributeValuable.swift
philprime: Yes, thank you, I believed I did this but it didn't accept the file name

Sources/Swift/Protocol/SentryAttributeValuable.swift
philprime: Yes, I can add that until Metrics is out of experimental state

Sources/Swift/Protocol/SentryAttributeValue.swift
philprime: I'll add the metrics API in the upcoming days, which will introduce a use-case, so I'll skip this for now.

Sources/Swift/Protocol/SentryAttribute.swift
```diff
@@ -2,105 +2,259 @@
 ///
 /// `Attribute` provides a type-safe way to store structured data alongside item messages.
 /// Supports String, Bool, Int, and Double types.
+///
+/// This class exists primarily for Objective-C compatibility, inheriting from `NSObject` and
+/// using `@objcMembers` to ensure it can be used from Objective-C code. For Swift code, prefer
+/// using the `SentryAttributeValue` protocol instead, which provides a more idiomatic Swift API
+/// and allows you to pass native Swift types (String, Bool, Int, Double, Float, and their arrays)
+/// directly without wrapping them in a class instance. This Objective-C-compatible class will be
+/// removed in a future major version release.
```
philippsuess, line 11: Yes exactly what I had in mind thanks.

Sources/Swift/Protocol/SentryAttributeValue.swift
```diff
@@ -0,0 +1,233 @@
+/// A protocol that represents values that can be used as structured logging attributes.
+///
+/// - Warning: **Experimental API** - This API is experimental and will be used by the upcoming
+///   experimental Metrics product. The API may change in future releases.
```
philippsuess, line 4: Thanks 👍

Sources/Swift/Protocol/SentryAttributeValue.swift
```diff
@@ -0,0 +1,233 @@
...
+                return String(describing: value)
+            }
+        })
+    }
+
+    func castValuesToBooleanArray(_ values: [SentryAttributeValue]) -> SentryAttributeContent? {
+        // Early exit optimization: if the first element isn't a boolean, skip the full iteration
+        guard !values.isEmpty, case .boolean = values[0].asSentryAttributeContent else {
+            return nil
+        }
```
philippsuess, line 166: Thanks

PR #7100, opened 2025-12-26
ref: HangTracker integration in Swift

### LINKED ISSUE #7101: ref: HangTracker integration in Swift
> [!NOTE]
> The pull request "[ref: HangTracker integration in Swift](https://github.com/getsentry/sentry-cocoa/pull/7100)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Converts SentryANRTrackingIntegration to Swift and renames it HangTracker since that's a more appropriate name for iOS

#skip-changelog

noahsmartin:
Converts SentryANRTrackingIntegration to Swift and renames it HangTracker since that's a more appropriate name for iOS

#skip-changelog

Closes #7101


Tests/SentryTests/Integrations/ANR/SentryHangTrackingIntegrationTests.swift
```diff
@@ -273,8 +272,8 @@ class SentryANRTrackingIntegrationTests: SentrySDKIntegrationTestsBase {
     func testANRDetected_DetectingPausedResumed_EventCaptured() throws {
         givenInitializedTracker()
         setUpThreadInspector()
-        sut.pauseAppHangTracking()
-        sut.resumeAppHangTracking()
+        sut?.pauseAppHangTracking()
```
itaybre: XCTUnwrap?

noahsmartin: It's not really what's being tested, but it could be used here, don't think it really matters

itaybre (approved):
LGTM

Sources/Swift/Integrations/SentryHangTrackingIntegration.swift
```diff
@@ -0,0 +1,236 @@
+@_implementationOnly import _SentryPrivate
+import Foundation
+#if (os(iOS) || os(tvOS) || os(visionOS)) && !SENTRY_NO_UIKIT
+import UIKit
+#endif
+
+typealias HangTrackingIntegrationScope = DispatchQueueWrapperProvider & CrashWrapperProvider & ExtensionDetectorProvider & DebugImageProvider & ThreadInspectorProvider & FileManagerProvider & ANRTrackerBuilder
```
philprime, line 7: @noahsmartin shouldn't we use a protocol here with only the methods we need for this class/integration to compile properly, so we can easily implement a type in tests with minimal effort?

noahsmartin: Looks like from the slack discussion this is already using the minimum protocols the type needs

PR #7102, opened 2025-12-29
chore: Convert SentrySessionReplayIntegration to Swift

### LINKED ISSUE #7105: chore: Convert SentrySessionReplayIntegration to Swift
> [!NOTE]
> The pull request "[chore: Convert SentrySessionReplayIntegration to Swift](https://github.com/getsentry/sentry-cocoa/pull/7102)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

## :scroll: Description

Converts SentrySessionReplayIntegration to Swift.
Took the chance to remove some logic from SentrySessionReplayIntegration to other classes 

Work to deprecate SentryDependencyContainer and lower the amount of code in ObjC

Ran the demo APP

## :pencil: Checklist

itaybre:
## :scroll: Description

Converts SentrySessionReplayIntegration to Swift.
Took the chance to remove some logic from SentrySessionReplayIntegration to other classes 

Work to deprecate SentryDependencyContainer and lower the amount of code in ObjC

Ran the demo APP

## :pencil: Checklist

Closes #7105

### DIFF
--- a/Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
+++ b/Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
@@ -355,7 +355,10 @@ public class SentrySessionReplayIntegration: NSObject, SwiftIntegration, SentryS
             stop()
             return
         }
-        guard let timestamp = replayEvent.timestamp else { return }
+        guard let timestamp = replayEvent.timestamp else { 
+            SentrySDKLog.warning("[Session Replay] Could not capture replay event, reason: timestamp is nil")
+            return 
+        }
         SentrySDKInternal.currentHub().captureReplayEvent(replayEvent, replayRecording: replayRecording, video: videoUrl)
         sentrySessionReplaySync_updateInfo(UInt32(replayEvent.segmentId), timestamp.timeIntervalSinceReferenceDate)
     }
@@ -372,7 +375,7 @@ public class SentrySessionReplayIntegration: NSObject, SwiftIntegration, SentryS
     }
 
     public func currentScreenNameForSessionReplay() -> String? {
-        SentrySDKInternal.currentHub().scope.currentScreen ?? application?.relevantViewControllersNames()?.first ?? ""
+        SentrySDKInternal.currentHub().scope.currentScreen ?? application?.relevantViewControllersNames()?.first
     }
 
     // MARK: - SentryReachabilityObserver


Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,893 @@
...
+
+typealias SessionReplayIntegrationScope = SessionReplayEnvironmentCheckerProvider & NotificationCenterProvider & RateLimitsProvider & CurrentDateProvider & RandomProvider & FileManagerProvider & CrashWrapperProvider & ReachabilityProvider & GlobalEventProcessorProvider & DispatchQueueWrapperProvider & ApplicationProvider & DispatchFactoryProvider
+
+private typealias CrossPlatformApplication = UIApplication
+
+// This is static because it will be used for swizzling and would cause retain cycles
+fileprivate var touchTracker: SentryTouchTracker?
+
+#if SENTRY_TEST || SENTRY_TEST_CI
+// This non final class is used for testing
```
noahsmartin: If the only difference here is `final` I think it would be cleaner just to have the non-final version and a comment saying we should update the tests not to subclass it

Sources/Sentry/SentryHub.m
```diff
@@ -128,7 +127,8 @@ - (void)startSession
         endSessionExitedWithTimestamp:[SentryDependencyContainer.sharedInstance.dateProvider date]];
     [self captureSession:lastSession];
 
-    [_sessionListener sentrySessionStarted:SENTRY_UNWRAP_NULLABLE(SentrySession, _session)];
+    [_sessionListener
+        sentrySessionStartedWithSession:SENTRY_UNWRAP_NULLABLE(SentrySession, _session)];
```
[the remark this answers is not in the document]
philprime: @itaybre please verify this

Sources/Sentry/SentryReplayApi.m
```diff
@@ -59,19 +58,26 @@ - (void)start SENTRY_DISABLE_THREAD_SANITIZER("double-checked lock produce false
             if (replayIntegration == nil && SentrySDKInternal.currentHub.client.options) {
                 SentryOptions *currentOptions = SENTRY_UNWRAP_NULLABLE(
                     SentryOptions, SentrySDKInternal.currentHub.client.options);
-                if (![SentrySessionReplayIntegration shouldEnableForOptions:currentOptions]) {
+                SentryDependencyContainer *sharedContainer =
+                    [SentryDependencyContainer sharedInstance];
+                if (![SentrySessionReplay
+                        shouldEnableSessionReplayWithEnvironmentChecker:
+                            [sharedContainer sessionReplayEnvironmentChecker]
```
philprime, line 65: `m`: Can't we inject the environment checker into this class instead of accessing it via the shared container as we want to get rid of that?

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,396 @@
...
+    
+    // MARK: - Properties
+    var replayProcessingQueue: SentryDispatchQueueWrapper
+    var replayAssetWorkerQueue: SentryDispatchQueueWrapper
+    @objc public var sessionReplay: SentrySessionReplay?
+    @objc public let viewPhotographer: SentryViewPhotographer
+    
+    private let replayOptions: SentryReplayOptions
+    private let rateLimits: RateLimits
+    private var rateLimited = false
```
philprime: `m`: We should keep comments giving additonal context:


```suggestion
    /// Flag to indicate if rate limited was activated
    ///
    /// We need to use this variable to identify whether rate limiting was ever activated for session
    /// replay in this session, instead of always looking for the rate status in `SentryRateLimits`
    /// This is the easiest way to ensure segment 0 will always reach the server, because session
    /// replay absolutely needs segment 0 to make replay work.
    private var rateLimited = false
```

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,396 @@
...
+        // The dispatch queue is used to asynchronously wait for the asset worker queue to finish its
+        // work. To avoid a deadlock, the priority of the processing queue must be lower than the asset
+        // worker queue. Use a relative priority of -2 to make it lower than the asset worker queue.
+        let processingQueue = dependencies.dispatchFactory.createUtilityQueue("io.sentry.session-replay.processing", relativePriority: -2)
+        
+        return (processingQueue, assetWorkerQueue)
+    }
+    
+    private func setupTouchTrackerIfNeeded(options: Options) {
+        guard options.enableSwizzling else { return }
```
philprime, line 146: `h`: Checking the previous implementation in `SentrySessionReplayIntegration.m` I notice that `options.enableSwizzling` was mapped to `enableTouchTracker` which wasn't used at all. Does that mean SR was swizzling events even if `enableSwizzling` was set to `false`? If yes, while this seems correct now, it's a behavioural change.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,396 @@
...
+        if !startedAsFullSession && replayOptions.onErrorSampleRate == 0 {
+            SentrySDKLog.debug("[Session Replay] Not full session and onErrorSampleRate is 0, not starting session")
+            return
+        }
+        
+        runReplayForAvailableWindow()
+    }
+
+    private func runReplayForAvailableWindow() {
+        if application?.getWindows()?.count ?? 0 > 0 {
```
philprime: `l`: Could be cleaner and easier to read without a `0 > 0`

```suggestion
        if let windowCount = application?.getWindows()?.count, windowCount > 0 {
```

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,396 @@
...
+                        screenshotProvider: currentScreenshotProvider ?? viewPhotographer,
+                        breadcrumbConverter: currentBreadcrumbConverter ?? SentrySRDefaultBreadcrumbConverter(),
+                        fullSession: fullSession)
+    }
+
+    private func startWithOptions(_ replayOptions: SentryReplayOptions, experimentalOptions: SentryExperimentalOptions,
+                                  screenshotProvider: SentryViewScreenshotProvider, breadcrumbConverter: SentryReplayBreadcrumbConverter, fullSession: Bool) {
+        SentrySDKLog.debug("[Session Replay] Starting session")
+        guard let sessionDocs = replayFileManager.createSessionDirectory() else { return }
+
```
philprime, line 229: `m`: We should have clear logging statements so we don't see a "Starting session" log statement, but then it didn't start due to the early return


```suggestion
        SentrySDKLog.debug("[Session Replay] Starting session")
        guard let sessionDocs = replayFileManager.createSessionDirectory() else { 
            SentrySDKLog.warning("[Session Replay] Failed to create session direcotry, cancelling starting session")
            return
        }

```

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,396 @@
...
+            replayOptions: replayOptions, replayFolderPath: sessionDocs, screenshotProvider: screenshotProvider,
+            replayMaker: replayMaker, breadcrumbConverter: breadcrumbConverter, touchTracker: touchTracker,
+            dateProvider: dateProvider, delegate: self, displayLinkWrapper: SentryDisplayLinkWrapper())
+
+        self.sessionReplay = newSessionReplay
+        newSessionReplay.start(rootView: application?.getWindows()?.first, fullSession: fullSession)
+        addBackgroundForegroundObservers()
+
+        guard let replayId = newSessionReplay.sessionReplayId else { return }
+        replayFileManager.saveCurrentSessionInfo(replayId, path: sessionDocs.path, options: replayOptions)
```
philprime: `l`: While I am generally an advocate for the `guard` statement, I believe here it could cause more problems if we just extend the `startWithOptions` setup instructions, with this early-exit causing partial setup.


```suggestion
        if let replayId = newSessionReplay.sessionReplayId {
            replayFileManager.saveCurrentSessionInfo(replayId, path: sessionDocs.path, options: replayOptions)
        }
```

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,396 @@
...
+        sessionReplay?.pause() 
+    }
+    @objc public func resume() {
+        SentrySDKLog.debug("[Session Replay] Resuming session")
+        sessionReplay?.resume() 
+    }
+
+    // MARK: - Public API
+
+    @objc public func start() {
```
philprime, line 281: `m`: If it's public API, please add documentation

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,396 @@
...
+    }
+    
+    private func removeBackgroundForegroundObservers() {
+        notificationCenter.removeObserver(self, name: UIApplication.didEnterBackgroundNotification, object: nil)
+        notificationCenter.removeObserver(self, name: UIApplication.willEnterForegroundNotification, object: nil)
+    }
+    
+    // MARK: - Public API for Testing
+    func replayDirectory() -> URL? { replayFileManager.replayDirectory() }
+    func moveCurrentReplay() { replayFileManager.moveCurrentReplay() }
```
philprime: `m`: If this is only used for testing, we should move it to the bottom of the file in the `#if TEST` segment

Sources/Swift/Integrations/SessionReplay/SessionReplayFileManager.swift
```diff
@@ -0,0 +1,162 @@
...
+    // MARK: - Directory Access
+    
+    func replayDirectory() -> URL? {
+        guard let sentryPath = fileManager?.sentryPath else { return nil }
+        return URL(fileURLWithPath: sentryPath).appendingPathComponent("replay")
+    }
+    
+    // MARK: - Session Info
+    
+    func saveCurrentSessionInfo(_ sessionId: SentryId, path: String, options: SentryReplayOptions) {
```
philprime, line 33: `h`: In the original implementation the `sessionId` was nullable` and would only be set if it's non-null, but would still create and write information. This implemenation only writes information if a session id is given. This is behavioural differences, so please explain why this is still the same. Assume that tests are incomplete as SR was not fully tested when it was implemented.

Sources/Swift/Integrations/SessionReplay/SessionReplayFileManager.swift
```diff
@@ -0,0 +1,162 @@
...
+            SentrySDKLog.error("[Session Replay] Could not get replay directory")
+            return nil
+        }
+
+        let currentSession = UUID().uuidString
+        let sessionDocs = docs.appendingPathComponent(currentSession)
+
+        if !FileManager.default.fileExists(atPath: sessionDocs.path) {
+            SentrySDKLog.debug("[Session Replay] Creating directory at path: \(sessionDocs.path)")
+            try? FileManager.default.createDirectory(
```
philprime: `h`: Do not use `try?` and silently fail. This is an easy pitfall for unexplainable behaviour. Always use do-catch and either rethrow or at least debug log the error

Sources/Swift/Integrations/SessionReplay/SessionReplayFileManager.swift
```diff
@@ -0,0 +1,162 @@
...
+        try? data.write(to: URL(fileURLWithPath: infoPath), options: .atomic)
+
+        SentrySDKLog.debug("[Session Replay] Saved current session info at path: \(infoPath)")
+        let crashInfoPath = (path as NSString).appendingPathComponent("crashInfo")
+        sentrySessionReplaySync_start(crashInfoPath)
+    }
+    
+    func lastReplayInfo() -> [String: Any]? {
+        guard let dir = replayDirectory() else { return nil }
+        let lastReplayUrl = dir.appendingPathComponent("replay.last")
```
philprime: `h`: You are reusing the same file name in multiple locations. Use an `enum` or at least constants to make sure they stay the same even if the name is changed

Sources/Swift/Integrations/SessionReplay/SessionReplayFileManager.swift
```diff
@@ -0,0 +1,162 @@
...
+            "path": (path as NSString).lastPathComponent,
+            "errorSampleRate": options.onErrorSampleRate
+        ]
+
+        guard let data = SentrySerializationSwift.data(withJSONObject: info) else {
+            SentrySDKLog.error("[Session Replay] Failed to serialize session info")
+            return
+        }
+
+        let infoPath = ((path as NSString).deletingLastPathComponent as NSString).appendingPathComponent("replay.current")
```
philprime: `h`: You are using the same file name in multiple locations. Use an `enum` or at least constants to make sure the file name stays consistent. This was already done in the original implementation, so this is implementation might be correct but less ergonomic to work with.

Sources/Swift/Integrations/SessionReplay/SessionReplayFileManager.swift
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+    }
+    
+    // MARK: - File Utilities
+    
+    func removeFileIfExists(atPath path: String) {
+        guard FileManager.default.fileExists(atPath: path) else { return }
+        SentrySDKLog.debug("[Session Replay] Removing file at path: \(path)")
+        try? FileManager.default.removeItem(atPath: path)
```
philprime: `h`: Do not use `try?` and silently fail. This is an easy pitfall for unexplainable behaviour. Always use do-catch and either rethrow or at least debug log the error

Sources/Swift/Integrations/SessionReplay/SessionReplayFileManager.swift
```diff
@@ -0,0 +1,162 @@
...
+    }
+    
+    func removeFileIfExists(at url: URL) {
+        guard FileManager.default.fileExists(atPath: url.path) else {
+            SentrySDKLog.debug("[Session Replay] No file to remove at path: \(url)")
+            return
+        }
+        SentrySDKLog.debug("[Session Replay] Removing file at path: \(url)")
+        try? FileManager.default.removeItem(at: url)
```
philprime: `h`: Do not use `try?` and silently fail. This is an easy pitfall for unexplainable behaviour. Always use do-catch and either rethrow or at least debug log the error

Sources/Swift/Integrations/SessionReplay/SessionReplayFileManager.swift
```diff
@@ -0,0 +1,162 @@
...
+    }
+    
+    private func moveFileIfExists(from source: URL, to destination: URL) {
+        guard FileManager.default.fileExists(atPath: source.path) else {
+            SentrySDKLog.debug("[Session Replay] No file to move at path: \(source)")
+            return
+        }
+        SentrySDKLog.debug("[Session Replay] Moving file from: \(source) to: \(destination)")
+        try? FileManager.default.moveItem(at: source, to: destination)
```
philprime: `h`: Do not use `try?` and silently fail. This is an easy pitfall for unexplainable behaviour. Always use do-catch and either rethrow or at least debug log the error

Sources/Swift/Integrations/SessionReplay/SessionReplayRecovery.swift
```diff
@@ -0,0 +1,195 @@
+@_implementationOnly import _SentryPrivate
+
+#if (os(iOS) || os(tvOS)) && !SENTRY_NO_UIKIT
+import Foundation
+
+/// Configuration for recovering a previous session replay after a crash or app restart.
+struct PreviousReplayConfig {
```
philprime: `l`: If this type is only used in this file, let's make it private for minimal availability scope


```suggestion
private struct PreviousReplayConfig {
```

Sources/Swift/Integrations/SessionReplay/SessionReplayRecovery.swift
```diff
@@ -0,0 +1,195 @@
...
+        SentrySDKInternal.currentHub().captureReplayEvent(
+            replayEvent,
+            replayRecording: recording,
+            video: video.path
+        )
+
+        do {
+            try FileManager.default.removeItem(at: video.path)
+        } catch {
+            SentrySDKLog.debug("[Session Replay] Could not delete replay segment from disk: \(error.localizedDescription)")
```
philprime: `m`: Thanks for useing do-catch here with proper error logging. I think this could potentially be useful to know for the user, so we should use a `warning` here

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -24,6 +24,20 @@ + (void)swizzle:(void (^)(SEL action, _Nullable id target, _Nullable id sender,
 #    pragma clang diagnostic pop
 }
 
++ (void)swizzleSendEvent:(void (^)(UIEvent *_Nullable event))callback;
```
philprime, line 31: `m`: We should add a test for this

Sources/Sentry/SentryHub.m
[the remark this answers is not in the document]
itaybre: False alarm, these names were created by Swift automatically.
I can make them more verbose to avoid alerting AI

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
itaybre: There is no behavior change, I had to break the functions a little bit to avoid swift lint errors.
It used to be:
```
if (touchTracker) {
        SENTRY_LOG_DEBUG(
            @"[Session Replay] Setting up touch tracker, scale: %f", replayOptions.sizeScale);
        _touchTracker = [[SentryTouchTracker alloc] initWithDateProvider:_dateProvider
                                                                   scale:replayOptions.sizeScale];
        [self swizzleApplicationTouch];
    }
```

This is now a new function:
```
private func setupTouchTrackerIfNeeded(options: Options) {
        guard options.enableSwizzling else { return }
        SentrySDKLog.debug("[Session Replay] Setting up touch tracker, scale: \(replayOptions.sizeScale)")
        touchTracker = SentryTouchTracker(dateProvider: dateProvider, scale: options.sessionReplay.sizeScale)
        SentrySwizzleWrapperHelper.swizzleSendEvent { event in
            guard let event = event else { return }
            touchTracker?.trackTouchFrom(event: event)
        }
    }
```

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
itaybre: I have no preference here

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
itaybre: Fair comment, it is not public, just exposed to ObjC

Sources/Swift/Integrations/SessionReplay/SessionReplayFileManager.swift
itaybre: We used `SENTRY_UNWRAP_NULLABLE` before, so we didn't actually support nullable values.
So no behavior change here

Sources/Sentry/SentrySwizzleWrapperHelper.m
itaybre: I could wrap the swizzling, but UIApplication doesn't work on unit tests.
So it isn't really possible to test it

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,400 @@
...
+    // MARK: - Notification Observers
+    
+    private func addBackgroundForegroundObservers() {
+        notificationCenter.addObserver(self, selector: #selector(pause), name: UIApplication.didEnterBackgroundNotification, object: nil)
+        notificationCenter.addObserver(self, selector: #selector(resume), name: UIApplication.didBecomeActiveNotification, object: nil)
+    }
+    
+    private func removeBackgroundForegroundObservers() {
+        notificationCenter.removeObserver(self, name: UIApplication.didEnterBackgroundNotification, object: nil)
+        notificationCenter.removeObserver(self, name: UIApplication.willEnterForegroundNotification, object: nil)
```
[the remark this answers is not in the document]
itaybre: Good catch, looks like this was a bug from before

Sources/Sentry/SentryReplayApi.m
itaybre: We can, but I believe this PR is big already, a new PR would be better for this.
Changing this means changing many more tests that already rely on this being in the DependencyContainer

Sources/Sentry/SentrySwizzleWrapperHelper.m
philprime: I believe we test other swizzling methods, can't we use that as a reference?

Sources/Sentry/SentrySDKInternal.m
```diff
@@ -506,11 +505,8 @@ + (void)endSession
     // The order of integrations here is important.
     // SentryCrashIntegration needs to be initialized before SentryAutoSessionTrackingIntegration.
     // And SentrySessionReplayIntegration before SentryCrashIntegration.
-    NSMutableArray<Class> *defaultIntegrations = [NSMutableArray<Class> arrayWithObjects:
-#if SENTRY_TARGET_REPLAY_SUPPORTED
-            [SentrySessionReplayIntegration class],
-#endif // SENTRY_TARGET_REPLAY_SUPPORTED
-        [SentryCrashIntegration class],
+    NSMutableArray<Class> *defaultIntegrations = [NSMutableArray<Class>
+        arrayWithObjects:[SentryCrashIntegration class],
```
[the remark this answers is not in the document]
itaybre: This is valid, looks like we need to convert `SentryCrashIntegration` first
Moving this back to draft

Sources/Sentry/SentrySwizzleWrapperHelper.m
itaybre: There is no UIApplication during unit tests, so this one in particular is hard to test.
Others should be testable

Sources/Swift/Integrations/SessionReplay/SessionReplayRecovery.swift
```diff
@@ -0,0 +1,205 @@
...
+        for video in videos {
+            captureVideo(video, replayId: replayId, segmentId: currentSegmentId, type: currentType)
+            currentSegmentId += 1
+            // type buffer is only for the first segment
+            currentType = .session
+        }
+
+        var eventContext = event.context ?? [:]
+        eventContext["replay"] = ["replay_id": replayId.sentryIdString]
+        event.context = eventContext
```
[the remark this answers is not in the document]
itaybre: Fixed above

philprime:
@itaybre please update the PR, it shows a lot of unrelated changes in the review UI.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,402 @@
...
+
+    public func sessionReplayNewSegment(replayEvent: SentryReplayEvent, replayRecording: SentryReplayRecording, videoUrl: URL) {
+        SentrySDKLog.debug("[Session Replay] New segment with replay event, eventId: \(replayEvent.eventId), segmentId: \(replayEvent.segmentId)")
+
+        if rateLimits.isRateLimitActive(SentryDataCategory.replay.rawValue) || rateLimits.isRateLimitActive(SentryDataCategory.all.rawValue) {
+            rateLimited = true
+            stop()
+            return
+        }
+        guard let timestamp = replayEvent.timestamp else { return }
```
philprime: `m`: As cursor already mentioned, could be worth adding a debug log statement

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    public func breadcrumbsForSessionReplay() -> [Breadcrumb] {
+        var result: [Breadcrumb] = []
+        SentrySDKInternal.currentHub().configureScope { scope in result = scope.breadcrumbs() }
+        return result
+    }
+
+    public func currentScreenNameForSessionReplay() -> String? {
+        SentrySDKInternal.currentHub().scope.currentScreen ?? application?.relevantViewControllersNames()?.first ?? ""
```
philprime: `m`: Either change this method to return non-optional `String` or remove the `?? ""`. I believe we shouldn't do a empty string fallback here

Sources/Sentry/SentryReplayApi.m
```diff
@@ -59,19 +58,26 @@ - (void)start SENTRY_DISABLE_THREAD_SANITIZER("double-checked lock produce false
...
+                                                    experimentalOptions:currentOptions
+                                                                            .experimental]) {
```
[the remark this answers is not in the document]
itaybre: False alarm

Issue #7112, opened 2026-01-04
Feedback dark mode doesn't adjust to theme changes


noahsmartin:
### Description

If I use `configureDarkTheme` to set different colors for dark mode of the user feedback button, they only work if the app is launched in dark mode and doesn't adjust if I then change it back to light mode.

To reproduce:
```
      config.configureDarkTheme = { theme in
        theme.background = .blue
      }
      config.configureTheme = { theme in
        theme.background = .black
      }
```

Then change between light and dark mode and see the button background color stays whatever color it was when the app launched

PR #7114, opened 2026-01-05
fix: Mark dark theme deprecated

noahsmartin:
Mark dark theme deprecated because UIColor's built in dark mode support ([dynamic provider](https://developer.apple.com/documentation/uikit/uicolor/init(dynamicprovider:))) should be used instead

Resolves https://github.com/getsentry/sentry-cocoa/issues/7112

### DIFF
--- a/Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
+++ b/Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
@@ -88,8 +88,6 @@ public final class SentryUserFeedbackConfiguration: NSObject {
     
     /**
      * Builder for default/light theme overrides.
-     * - note: On iOS versions predating dark mode (≤12) this is the only theme override used. Apps
-     * running on later versions that include dark mode should also consider `configureDarkTheme`.
      * - note: Default: `nil`
      */
     public var configureTheme: ((SentryUserFeedbackThemeConfiguration) -> Void)?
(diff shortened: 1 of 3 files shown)


Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
```diff
@@ -101,9 +99,17 @@ public final class SentryUserFeedbackConfiguration: NSObject {
...
+    public var configureDarkTheme: ((SentryUserFeedbackThemeConfiguration) -> Void)? {
+        @available(*, deprecated, message: "Use dynamic UIColor instead of the dark theme.")
+        set {
+            _configureDarkTheme = newValue
+        }
+        get {
+            _configureDarkTheme
+        }
+    }
+    var _configureDarkTheme: ((SentryUserFeedbackThemeConfiguration) -> Void)?
```
philippsuess, line 114: `m`: Why do we only mark the setter as deprecated? Can't we mark the whole property as deprecated? If there's a good reason, please add the info as a comment.

noahsmartin: It's because we access the getter in the SDK code during SDK init, and we have warnings as errors on so the warning about accessing a deprecated property causes the SDK not to compile. The setter is what users access though, so this gives users of the SDK a deprecated warning without causing a new warning when building the SDK

philippsuess: That's an acceptable workaround. I think it makes sense to add the info as a comment, please.

noahsmartin: yep done

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
```diff
@@ -101,9 +99,19 @@ public final class SentryUserFeedbackConfiguration: NSObject {
      * mode, but you still want to override some theme settings, assign the same builder to this
      * property as you do for `configureTheme`.
      * - note: Default: `nil`
-     * - note: Only applies to iOS ≤12.
      */
-    public var configureDarkTheme: ((SentryUserFeedbackThemeConfiguration) -> Void)?
+    public var configureDarkTheme: ((SentryUserFeedbackThemeConfiguration) -> Void)? {
+        // Only the setter is deprecated to avoid warnings when compiling the SDK. The getter is
+        // used from SentrySDK.init
```
philippsuess, line 105: Thanks 👍

Sources/Swift/Integrations/UserFeedback/Configuration/SentryUserFeedbackConfiguration.swift
itaybre: Should we remove this API on v10?
If yes, we should create a ticket to remember this

PR #7153, opened 2026-01-10
ci(release): Switch from action-prepare-release to Craft (minimal)

BYK:
## Summary

This is a **minimal migration** that preserves all existing workflow structure.

## Changes

- Updated `getsentry/action-prepare-release@v1` to `getsentry/craft@v2` (SHA pinned)
- Added `changelog-preview.yml` workflow with proper permissions block

## What's NOT changed

- All XCFramework build jobs preserved
- All validation jobs preserved (SPM, duplication tests, app metrics)
- All triggers preserved (push, pull_request, workflow_dispatch)
- Concurrency settings preserved

## Why this PR?

The previous PR #7150 was too aggressive and removed essential build infrastructure. This minimal approach only updates the action reference while preserving all existing functionality.

Replaces #7150

### DIFF
--- a/.github/workflows/api-stability.yml
+++ b/.github/workflows/api-stability.yml
@@ -25,7 +25,7 @@ jobs:
     outputs:
       run_api_stability_for_prs: ${{ steps.changes.outputs.run_api_stability_for_prs }}
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
       - name: Get changed files
         id: changes
         uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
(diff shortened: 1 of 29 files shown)


.github/workflows/changelog-preview.yml
```diff
@@ -0,0 +1,13 @@
+name: Changelog Preview
```
philprime, line 1: `m`: @BYK can you share some insights with us on the benefit of having the full changelog on every single PR? It feels to me like this is mostly noise.

BYK: @philprime it's a preference and an educational thing. It is especially useful for the following scenarios:

1. See what kind of a version bump your PR will cause
2. How will your PR will be shown amond other changes with section and scope grouping _and_ custom changelog entries when using https://getsentry.github.io/craft/configuration/#custom-changelog-entries-from-pr-descriptions

If these are not useful, we can definitely remove this workflow. (FYI it only leaves a single comment and keeps updating that when things change)

philprime: Then let's go for it an we can iterate on it.

philprime (approved):
LGTM

PR #7155, opened 2026-01-12
fix:  Crash in `SentryFramesTracker.removeListener`

denrase:
## :scroll: Description

Mutate listeners in `SentryFramesTracker` only on main queue.

Access to listeners was moved to main thread dispatch instead of locks, but one lock was probably forgotten. The reason why this was changed is added as a comment in code, so because of this we do not use the locking anymore.

> The DisplayLink callback always runs on the main thread. We dispatch this to the main thread instead to avoid using locks in the DisplayLink callback.

If this issue should remain even after this fix, we can still additionally add locks/synchronisation around listeners access.

Closes https://linear.app/getsentry/issue/COCOA-869/sentryframestracker-removelistener-crash

Unfortunately, I could not reproduce the crash, but the sync was only in one place, suggesting it was forgotten to move this to main thread access only.

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
+++ b/Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
@@ -20,9 +20,23 @@ public protocol SentryFramesTrackerListener: NSObjectProtocol {
 public class SentryFramesTracker: NSObject {
     
     private var isStarted: Bool = false
-    @objc public private(set) var isRunning: Bool = false
     
     // MARK: Private properties
+    private var isRunningLock = NSLock()
+    private var _isRunning: Bool = false
+    
+    @objc public private(set) var isRunning: Bool {
+        get {
+            return isRunningLock.synchronized {
+                return _isRunning
+            }
+        }
+        set {
+            isRunningLock.synchronized {
+                _isRunning = newValue
+            }
+        }
+    }
     private var previousFrameTimestamp: CFTimeInterval = SentryFramesTracker.previousFrameInitialValue
     private var previousFrameSystemTimestamp: UInt64 = 0
     private var currentFrameRate: UInt64 = 60


denrase:
@philprime Due to the comment in the code, moved the last remaining lock to main thread dispatch instead.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Fix `raw_description` in `runtime` context on Mac Catalyst (#7082)
 - Deprecates `configureDarkTheme` for user feedback (#7114)
 - Fix incorrect variable assignment for 'sampled' key (#7120)
+- [SentryFramesTracker removeListener:] crash ([#7155](https://github.com/getsentry/sentry-cocoa/pull/7155))
```
```suggestion
- Resolve crash in caused by calling `SentryFramesTracker.removeListener(_:)` (#7155)
```

denrase:
@philprime Not because of locking the main thread, but the code base was moving away from locks, with the reasoning that `DisplayLink` is always calling on the main thread anyway, so this PR completes this transition.

But there's the catch that I was not able to reproduce the crash from the issue, but basically it was possible to clear the listeners from a non-main thread, while adding/removing individual listeners was always being done on the main-thread.

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -170,6 +167,12 @@ public class SentryFramesTracker: NSObject {
             self.listeners.remove(listener)
         }
     }
+
+    func removeAllListeners() {
+        dispatchQueueWrapper.dispatchAsyncOnMainQueueIfNotMainThread {
+            self.listeners.removeAllObjects()
+        }
+    }
```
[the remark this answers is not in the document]
denrase: Fixed by calling stop sync on main thread.

denrase:
@philprime Would be great if you could take another look. I tried reproducing again after the cursor bot comment and could get the frames tracker to crash using this code in the sample app. I added additional precautions to fix this, which did work. While I still could not reproduce the initial issue, this should offer more protection against potential races.

```swift
    @IBAction func useAfterFree(_ sender: UIButton) {
        highlightButton(sender)
        
        let framesTracker = SentryDependencyContainer.sharedInstance().framesTracker
        
        class TestListener: NSObject, SentryFramesTrackerListener {
            func framesTrackerHasNewFrame(_ newFrameDate: Date) {
                // Access during callback to create more contention
            }
        }
        
        var listeners: [TestListener] = []
        for _ in 0..<50 {
            let listener = TestListener()
            listeners.append(listener)
            framesTracker.addListener(listener)
        }
        for _ in 0..<20 {
                for listener in listeners {
                    for i in 0..<100 {
                        framesTracker.removeListener(listener)
                        if i % 2 == 0 {
                            framesTracker.stop()
                        } else {
                            framesTracker.start()
                        }
                    }
                }
            }
        }
    }
```

Bildschirmfoto 2026-01-13 um 11 53 31

philprime (approved):
LGTM, this also seems to be more consistent. Maybe it makes sense if @itaybre or @philipphofmann also gives this a second review

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -18,12 +18,11 @@ public protocol SentryFramesTrackerListener: NSObjectProtocol {
 
 @_spi(Private) @objc
 public class SentryFramesTracker: NSObject {
-
-    var isStarted: Bool = false
+    
+    private var isStarted: Bool = false
```
philippsuess, line 22: Makes sense thanks 👍

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -68,11 +67,24 @@ public class SentryFramesTracker: NSObject {
         resetFrames()
         SentrySDKLog.debug("Initialized frame tracker")
     }
+    
+    deinit {
+        // Avoid async dispatch with self capture on deinit.
+        dispatchQueueWrapper.dispatchSyncOnMainQueue {
```
philippsuess: `h`: I don't think it'a good idea to call dispatch_sync on a deinit method. Even though most of the time we're going to call this from the main thread and we won't run `dispatch_sync` from a BG thread, this can cause weird behaviour and hard-to-track bugs. To me, it seems this is a duct tape around a deeper problem of how we handle synchronization around the listeners. I think it would be better to come up with a different strategy. For example, I don't think we need to remove all listeners in deinit. The listeners array anyways is going to be deallocated. Maybe we should rather call a stop method that only removes the observers and calls pause, but not removing the observers.

denrase: Yeah that makes sense, what we def need to do is invalidate the display link, as it retains the object is is reporting to. Will update, thx for the insights. 👍

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -68,11 +67,22 @@ public class SentryFramesTracker: NSObject {
...
     // MARK: - Public Methods
 
     @objc
     public func start() {
+        dispatchQueueWrapper.dispatchAsyncOnMainQueueIfNotMainThread {
+            self.startInternal()
+        }
+    }
+    
+    private func startInternal() {
```
itaybre, line 100: Can we double check that no one calling `start` expects data immediately?

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -68,11 +67,22 @@ public class SentryFramesTracker: NSObject {
         resetFrames()
         SentrySDKLog.debug("Initialized frame tracker")
     }
+    
+    deinit {
+        // Need to invalidate so DisplayLink is releasing this object. Calling this is thread-safe.
+        displayLinkWrapper.invalidate()
```
philippsuess, line 88: `m`: @denrase I think we should also remove the observers as we do in `stopInternal`. I think we could add a new method that calls `displayLinkWrapper.invalidate()` and also removes the observers. Then we can call it from here and `stopInternal`

denrase: We are already invalidating the display link wrapper on pause. AFAIK removing the observers is per-se not necessary, as the notification center should not retain objects, but the docs i referenced were not super clear about this. thx!

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
denrase: Looked over where this is used and it should not be an issue. `SentryFramesTrackingIntegration ` just calls start without reading more values. `SentryAppStartTracker` registers as listener, which is a callback that gets called.

Issue #5423, opened 2025-06-17
Let UI tests run for releases


philipphofmann:
### Description

When creating a release, the UI tests don't run for the release branch. We tried to enable this with https://github.com/getsentry/sentry-cocoa/pull/5362, but then we ran into permission issues during the release https://github.com/getsentry/sentry-cocoa/pull/5370 and https://github.com/getsentry/sentry-cocoa/actions/runs/15468574493/job/43548210375.

philprime:
@itaybre @philipphofmann now that we have more stable UI tests due to the switch to CirrusLabs, should we consider enabling it for the release workflows?

philippsuess:
Yes, they're even more stable than our unit tests. Let's do it.

philippsuess:
Blocked by the SPM chicken egg problem during the release process:  [https://github.com/getsentry/sentry-cocoa/pull/7156#issuecomment-3799572016](<https://github.com/getsentry/sentry-cocoa/pull/7156#issuecomment-3799572016>)

PR #7156, opened 2026-01-12
ci: Run UI tests on release branches

philipphofmann:
Our UI tests are now stable enough to also run for release branches.

Fixes GH-5423

#skip-changelog

We should do an alpha release before merging this branch, to ensure we don't run into permission issues as mentioned in https://github.com/getsentry/sentry-cocoa/issues/5423#issue-3152680547.


philippsuess:
This PR is blocked by the broken release workflow caused by migrating from action-prepare-release to Craft https://github.com/getsentry/sentry-cocoa/pull/7153.

itaybre:
> This PR is blocked by the broken release workflow caused by migrating from action-prepare-release to Craft #7153.

This should be ready to be enabled now, right?

philippsuess:
Yes, indeed. I didn't get to that yet. Thanks for the ping @itaybre.

philippsuess:
We can't enable that because the UI tests also surface the SPM chicken egg problem: see   

```
target 'Sentry-Dynamic-WithARM64e': badResponseStatusCode(404)failed downloading 'https://github.com/getsentry/sentry-cocoa/releases/download/9.2.0-alpha.0/Sentry-WithoutUIKitOrAppKit-WithARM64e.xcframework.zip' which is required by binary target 'Sentry-WithoutUIKitOrAppKit-WithARM64e': badResponseStatusCode(404)fatalError
```

https://github.com/getsentry/sentry-cocoa/actions/runs/21358000971/job/61469949383

I'm going to close this PR, until we have an proper way to fix the chicken egg SPM resolution problem. cc @itaybre

PR #7157, opened 2026-01-12
ref: convert SentryViewHierarchyIntegration to Swift

### LINKED ISSUE #6853: Refactor `SentryViewHierarchyIntegration.m` in Swift

philprime:
## :scroll: Description

Convert SentryViewHierarchyIntegration from Objective-C to Swift, following the same pattern as SentryScreenshotIntegration. The Swift implementation uses dependency injection and the SwiftIntegration protocol.

Changes:
- Convert SentryViewHierarchyIntegration.m/.h to Swift
- Add ViewHierarchyIntegrationProvider protocol for DI
- Convert test helper from Objective-C to Swift using @_cdecl
- Remove unused TestSentryViewHierarchyProvider.h header
- Update tests to use Swift integration initialization pattern

Closes #6853 

- Unit tests

#skip-changelog

### DIFF
--- a/Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
+++ b/Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
@@ -7,7 +7,7 @@ typealias SentryViewHierarchyIntegrationProvider = ViewHierarchyProviderProvider
 final class SentryViewHierarchyIntegration<Dependencies: SentryViewHierarchyIntegrationProvider>: NSObject, SwiftIntegration, SentryClientAttachmentProcessor {
     private let options: Options
     private let viewHierarchyProvider: SentryViewHierarchyProvider
-    private let client: SentryClientInternal
+    private weak var client: SentryClientInternal?
 
     init?(with options: Options, dependencies: Dependencies) {
         guard options.attachViewHierarchy else {
@@ -44,7 +44,7 @@ final class SentryViewHierarchyIntegration<Dependencies: SentryViewHierarchyInte
 
     func uninstall() {
         sentrycrash_setSaveViewHierarchy(nil)
-        client.removeAttachmentProcessor(self)
+        client?.removeAttachmentProcessor(self)
     }
 
     static var name: String {


philprime:
@sentry review

Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
```diff
@@ -0,0 +1,91 @@
...
+    init?(with options: Options, dependencies: Dependencies) {
+        guard options.attachViewHierarchy else {
+            SentrySDKLog.debug("Not going to enable \(Self.name) because attachViewHierarchy is disabled.")
+            return nil
+        }
+
+        guard let viewHierarchyProvider = dependencies.viewHierarchyProvider else {
+            SentrySDKLog.debug("Not going to enable \(Self.name) because viewHierarchyProvider is not available.")
+            return nil
+        }
```
philippsuess, line 21: That reads way nicer than the old way of enabling and disabling the integrations 💯

Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        self.options = options
+        self.viewHierarchyProvider = viewHierarchyProvider
+
+        super.init()
+
+        if let client = SentrySDKInternal.currentHub().getClient() {
+            self.client = client
+            client.addAttachmentProcessor(self)
+        }
```
philippsuess: `m`: I know that this was already the same before, but maybe it makes even sense to return nil here if there is no client, because if the client is `nil`, the integration won't work.

itaybre: Yeah, not having any client doesn't make much sense right now

Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
```diff
@@ -0,0 +1,91 @@
...
+            guard let path = path else { return }
+            let reportPath = String(cString: path)
+            let filePath = (reportPath as NSString).appendingPathComponent("view-hierarchy.json")
+            SentryDependencyContainer.sharedInstance().viewHierarchyProvider?.saveViewHierarchy(filePath)
+        }
+    }
+
+    func uninstall() {
+        sentrycrash_setSaveViewHierarchy(nil)
+        client?.removeAttachmentProcessor(self)
```
itaybre, line 47: `l`: As far as I know we only create a new client on startSDK, so I wouldn't expect this client reference to change to a different one if we use `SentrySDKInternal.currentHub().getClient()?.removeAttachmentProcessor(self)`

This would just avoid saving the client reference

philippsuess: Yes, you're absolutely right. That's a main problem in our integration architecture right now. When the client or the hub changes, this breaks. We don't need to fix this in this PR. Thanks for pointing it out.

philprime: I am not sure if I understand correctly. If the client never changes, I don't see a reason to dynamically fetch it via `SentrySDKInternal.currentHub().getClient()` and instead only have the weak reference?

philippsuess: The client only changes if you restart the SDK. In that case, also the integration gets deallocated and recreated. So right now it's not a real problem. It was previously, when you could still manually bind a different hub or client to the static SentrySDK. So my bad, I think it's not a problem now, but it could easily become a problem in the future. We should pass in the client or hub as a dependency and not retrieve it via the Static API.

philippsuess: We could dynamically fetch it, but then it wouldn't work if you have multiple hubs, and each hub has its own integrations. That's not a valid use case right now, but it could be in the future, for Swift Vapor, for example. The integrations shouldn't depend on the static API, but rather on a specific hub or a client. They shouldn't even know there is a static API, or you can have different hubs. Does it make sense now?

philprime: So, I could inject a factory "getter" method via the dependencies, so the integration itself doesn't have to know where the client comes from, it just knows that it needs to be fetched every time without keeping a reference. Would that resolve your concerns?

Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
philprime: As long as the getClient returns a nullable value, I need to do null handling. So the only two options are:

1. Do it as is, the integration exists but it's not installed as an attachment processor
2. Cancel the setup of the integration, returning nil from the initializer

Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
philippsuess: Yes, and I think we could then use this factory getter in all integrations that need to access the client, but we don't have to do that in this PR. We can also follow up in another PR.

Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
itaybre: 2 is more correct IMO

philprime: I added a guard statement to return nil on absence of the client

Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
philprime: I introduced a `ClientProvider` protocol which just wraps the static accessor of `SentrySDKInternal.currentHub().getClient()` for now, and accesses it once during integration init.

Sources/Swift/Integrations/ViewHierarchy/SentryViewHierarchyIntegration.swift
```diff
@@ -0,0 +1,95 @@
+@_implementationOnly import _SentryPrivate
+
+#if (os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)) && !SENTRY_NO_UIKIT
+
+typealias SentryViewHierarchyIntegrationProvider = ViewHierarchyProviderProvider & ClientProvider
+
+final class SentryViewHierarchyIntegration<Dependencies: SentryViewHierarchyIntegrationProvider>: NSObject, SwiftIntegration, SentryClientAttachmentProcessor {
+    private let options: Options
+    private let viewHierarchyProvider: SentryViewHierarchyProvider
+    private let client: SentryClientInternal
```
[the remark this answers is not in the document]
philprime: This is valid feedback, but the attachment processors are stored as a strong map in the client, so I can only change the client back to be weak.

PR #7172, opened 2026-01-14
ref: Renames `SentrySpan` class to `SentrySpanInternal`

### LINKED ISSUE #7176: ref: Renames `SentrySpan` class to `SentrySpanInternal`
> [!NOTE]
> The pull request "[ref: Renames `SentrySpan` class to `SentrySpanInternal`](https://github.com/getsentry/sentry-cocoa/pull/7172)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

## :scroll: Description

Renames the class `SentrySpan` to `SentrySpanInternal`, keeping `SentrySpan` only to the protocol.
Only the protocol is a public API, so no breaking changes were made

We have 2 types with identical names, one a protocol and another a class.
While converting the SentryCrashIntegration to Swift, I ran into issues with them having the same name in ObjC.

## :pencil: Checklist

itaybre:
## :scroll: Description

Renames the class `SentrySpan` to `SentrySpanInternal`, keeping `SentrySpan` only to the protocol.
Only the protocol is a public API, so no breaking changes were made

We have 2 types with identical names, one a protocol and another a class.
While converting the SentryCrashIntegration to Swift, I ran into issues with them having the same name in ObjC.

## :pencil: Checklist

Closes #7176

### DIFF
--- a/Sources/Sentry/SentryCoreDataTracker.m
+++ b/Sources/Sentry/SentryCoreDataTracker.m
@@ -37,7 +37,8 @@
                             error:(NSError **)error
                       originalImp:(NSArray *(NS_NOESCAPE ^)(NSFetchRequest *, NSError **))original
 {
-    SentrySpanInternal *_Nullable currentSpan = [SentrySDKInternal.currentHub.scope span];
+    SentrySpanInternal *_Nullable currentSpan
+        = (SentrySpanInternal *_Nullable)[SentrySDKInternal.currentHub.scope span];
     SentrySpanInternal *_Nullable fetchSpan;
     if (currentSpan) {
         NSString *spanDescription = [self descriptionFromRequest:request];
--- a/Sources/SentrySwiftUI/SentryInternal/SentryInternal.h
+++ b/Sources/SentrySwiftUI/SentryInternal/SentryInternal.h
@@ -18,7 +18,7 @@
 #endif
 
 #if SENTRY_TEST
-#    import "SentrySpan.h"
+#    import "SentrySpanInternal.h"
 #    import "SentryTracer.h"
 #else
 @protocol SentrySpan;


itaybre:
@sentry review

itaybre:
> LGTM, but might we worth to ask one of our coding agents to verify we caught all use cases of `SentrySpan` which need to be renamed

Should be fine since if I missed any case, it shouldn't compile
But I triggered the review anyway

Sources/Sentry/include/SentrySpanInternal.h
```diff
@@ -15,7 +15,7 @@ NS_ASSUME_NONNULL_BEGIN
 @class SentryFramesTracker;
 #endif // SENTRY_HAS_UIKIT
 
-@interface SentrySpan : NSObject <SentrySpan, SentrySerializable>
+@interface SentrySpanInternal : NSObject <SentrySpan, SentrySerializable>
```
philippsuess, line 20: `m`: Maybe it's worth renaming the file as well.

philippsuess, line 20: `m`: Maybe worth adding a comment explaining that this was called `SentrySpan` before, and there's not a good reason for that, and why we renamed it.

Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -37,8 +37,8 @@ - (NSArray *)managedObjectContext:(NSManagedObjectContext *)context
                             error:(NSError **)error
                       originalImp:(NSArray *(NS_NOESCAPE ^)(NSFetchRequest *, NSError **))original
 {
-    id<SentrySpan> _Nullable currentSpan = [SentrySDKInternal.currentHub.scope span];
-    id<SentrySpan> _Nullable fetchSpan;
+    SentrySpanInternal *_Nullable currentSpan = [SentrySDKInternal.currentHub.scope span];
+    SentrySpanInternal *_Nullable fetchSpan;
```
philippsuess: `m`: I don't think we should have to change this here. We should still be able to use the protocol and not the class. What drove you to change this?

Sources/Sentry/Public/SentryScope.h
```diff
@@ -13,6 +13,7 @@
 @class SentryBreadcrumb;
 @class SentryOptions;
 @class SentryUser;
+@class SentrySpanInternal;
```
philippsuess: `h`: SentryScope is public. That shouldn't have to be here.
```suggestion
```

Sources/Sentry/SentryScope.m
```diff
@@ -147,23 +147,23 @@ - (void)setPropagationContext:(SentryPropagationContext *)propagationContext
     }
 }
 
-- (nullable id<SentrySpan>)span
+- (nullable SentrySpanInternal *)span
```
philippsuess: `m`: That shouldn't have to change. This should stay the same. What drove you to change this?

philipphofmann (commented):
There are several replacements of `id<SentrySpan>` with `SentrySpanInternal`. I don't think these should be required.

Tests/SentryTests/SentryScopeSwiftTests.swift
```diff
@@ -1057,5 +1057,5 @@ private final class NotOfTypeSpan: NSObject, Span {
     func serialize() -> [String: Any] { return [:] }
 }
 
-private final class SubClassOfSentrySpan: SentrySpan {}
+private final class SubClassOfSentrySpan: SentrySpanInternal {}
```
philippsuess: `l`: 
```suggestion
private final class SubClassOfSentrySpanInternal: SentrySpanInternal {}
```

Sources/Sentry/Public/SentryScope.h
itaybre: Good catch, it should be the protocol in here then

Sources/Sentry/SentryScope.m
itaybre: Probably CrashIntegration, but doesn't seems needed here.
Will undo this change

Sources/Sentry/SentryCoreDataTracker.m
itaybre: `SentrySpanInternal` has an extra function `setFrames` which is used below.
SentrySpan and SentrySpanInternal are almost interchangeable, the internal class just has some additional functions

itaybre:
> There are several replacements of `id<SentrySpan>` with `SentrySpanInternal`. I don't think these should be required.

There is no reason to use `id<SentrySpan>` instead of `SentrySpanInternal` internally, the protocol is just the public interface

Sources/Sentry/include/SentrySpanInternal.h
```diff
@@ -15,7 +15,9 @@ NS_ASSUME_NONNULL_BEGIN
 @class SentryFramesTracker;
 #endif // SENTRY_HAS_UIKIT
 
-@interface SentrySpan : NSObject <SentrySpan, SentrySerializable>
+// This class used to be named SentrySpan, but it was renamed to SentrySpanInternal to avoid
+// confusion with the public SentrySpan protocol.
```
philippsuess, line 19: Thank you 🥇

PR #7177, opened 2026-01-14
fix(tests): Remove fatal error logging in SentryFileManager

### LINKED ISSUE #5317: Tests: Spammy fatal logs from SentryFileManager
### Description

When checking the test logs in CI we can see that the SentryFileManager spams the logs with fatal messages like the following ones

```
[Sentry] [fatal] [timeIntervalSince1970:1748454849.6005049] [SentryFileManager:63] Failed to create directory, path is too long: (null)
[Sentry] [fatal] [timeIntervalSince1970:1748454849.6005502] [SentryFileManager:140] Failed to create Sentry SDK working directory: (null)
[Sentry] [fatal] [timeIntervalSince1970:1748454849.6008] [SentryFileManager:63] Failed to create directory, path is too long: (null)
[Sentry] [fatal] [timeIntervalSince1970:1748454849.600837] [SentryFileManager:140] Failed to create Sentry SDK working directory: (null)
[Sentry] [fatal] [timeIntervalSince1970:1748454849.605884] [SentryFileManager:63] Failed to create directory, path is too long: (null)
[Sentry] [fatal] [timeIntervalSince1970:1748454849.605921] [SentryFileManager:140] Failed to create Sentry SDK working directory: (null)
[Sentry] [fatal] [timeIntervalSince1970:1748454849.606196] [SentryFileManager:63] Failed to create directory, path is too long: (null)
[Sentry] [fatal] [timeIntervalSince1970:1748454849.60624] [SentryFileManager:140] Failed to create Sentry SDK working directory: (null)
```

You can reproduce this by running the unit tests `SentrySerializationTests` on iOS. The `clearTestState` calls SentrySDK.close, which then calls AppStateManager, which tries to init a FileManager, but the options are nil.

Image

The SentryDependencyContainer.fileManger property is nonnullable, but it actually return `nil` when there are no options bound to the SDK. 

A PR that could be related to fixing this problem is https://github.com/getsentry/sentry-cocoa/pull/5315, which will require passing in the options when retrieving a file manager.

An easy fix is that the SentryFileManager doesn't print a fatal error message when passing in nil for the options.

philprime:
## :scroll: Description

This change prevents fatal error logs from cluttering test outputs when the Sentry SDK working directory or envelopes directory cannot be created during tests. The logging is conditionally compiled to exclude test environments.

**Matches of `Failed to create Sentry SDK working directory`:**

- Before: 5636 matches
- After: 0 matches 

Closes #5317 

The created Xcode Results bundle contains the test logs, where I counted the occurences of the spam text.

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryFileManagerHelper.m
+++ b/Sources/Sentry/SentryFileManagerHelper.m
@@ -120,12 +120,18 @@ _non_thread_safe_removeFileAtPath(NSString *path)
         [self removeFileAtPath:self.eventsPath];
 
         if (!createDirectoryIfNotExists(self.sentryPath, error)) {
+// Disabled fatal error logging in tests to reduce noise in test logs
+#if !SENTRY_TEST && !SENTRY_TEST_CI
             SENTRY_LOG_FATAL(@"Failed to create Sentry SDK working directory: %@", self.sentryPath);
+#endif
             return nil;
         }
         if (!createDirectoryIfNotExists(self.envelopesPath, error)) {
+// Disabled fatal error logging in tests to reduce noise in test logs
+#if !SENTRY_TEST && !SENTRY_TEST_CI
             SENTRY_LOG_FATAL(
                 @"Failed to create Sentry SDK envelopes directory: %@", self.envelopesPath);
+#endif
             return nil;
         }
 


Sources/Sentry/SentryFileManagerHelper.m
```diff
@@ -120,12 +120,18 @@ - (nullable instancetype)initWithOptions:(SentryOptions *_Nullable)options error
         [self removeFileAtPath:self.eventsPath];
 
         if (!createDirectoryIfNotExists(self.sentryPath, error)) {
+// Disabled fatal error logging in tests to reduce noise in test logs
```
philippsuess, line 123: `m`: I'm not sure if this is the right approach, honestly. I would rather avoid printing these fatal log messages in the specific scenario that leads to these fatal log messages. So basically not trying to create the directories at all and do an early exit instead. One idea was to not print these when passing `nil` for the options (https://github.com/getsentry/sentry-cocoa/issues/5317). 

> An easy fix is that the SentryFileManager doesn't print a fatal error message when passing in nil for the options.

We could do an early return when the options are `nil`. That early return still prints an error message, because it's an important error, but we could use `#if !SENTRY_TEST && !SENTRY_TEST_CI` for that specific message.

philippsuess: Ideally, we would fix the root cause of passing `nil` as the options to the file manager, but that could be some work.

philprime: I believe we need these fatal logs for edge cases happening for the customer. That's why I introduced them in the first place. I do not want to refactor this now, that's part of [this Linear project](https://linear.app/getsentry/project/typed-file-storage-6babfee308f1/overview) which is going to be a larger task

philippsuess: Fair enough. Maybe add this info as a comment, so it doesn't get lost here.

philipphofmann (approved):
LGTM.

PR #7178, opened 2026-01-14
chore: Add temporary macro for Swift migration tracking of types erased

### LINKED ISSUE #7179: chore: Add temporary macro for Swift migration tracking of types erased
> [!NOTE]
> The pull request "[chore: Add temporary macro for Swift migration tracking of types erased](https://github.com/getsentry/sentry-cocoa/pull/7178)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

## :scroll: Description

Adds a simple macro to help us track the types erased due to the Swift migration, so we can update them later.

Due to ObjC and Swift incompatibilities and doing a migration based on needs instead of bottom up, we run into issues like ObjC functions not visible to Swift.

## :pencil: Checklist

itaybre:
## :scroll: Description

Adds a simple macro to help us track the types erased due to the Swift migration, so we can update them later.

Due to ObjC and Swift incompatibilities and doing a migration based on needs instead of bottom up, we run into issues like ObjC functions not visible to Swift.

## :pencil: Checklist

Closes #7179

### DIFF
--- a/Sources/Sentry/SentryScope.m
+++ b/Sources/Sentry/SentryScope.m
@@ -678,7 +678,7 @@ NS_ASSUME_NONNULL_BEGIN
     return event;
 }
 
-- (void)addScopeObserver:(id)observer
+- (void)addScopeObserver:(SENTRY_SWIFT_MIGRATION_ID(id<SentryScopeObserver>))observer
 {
     // Validate if the observer conforms because the API doesn't require
     // observers to conform anymore due to the protocol being written in Swift
--- a/Sources/Sentry/include/SentryScope+PrivateSwift.h
+++ b/Sources/Sentry/include/SentryScope+PrivateSwift.h
@@ -23,7 +23,8 @@ static NSString *const SENTRY_CONTEXT_APP_KEY = @"app";
  * The observer should conform to SentryScopeObserver protocol, but the type is id
  * because Swift-defined protocols don't work correctly with Objective-C's id<Protocol> syntax.
  */
-- (void)addScopeObserver:(id)observer NS_SWIFT_NAME(add(_:));
+- (void)addScopeObserver:(SENTRY_SWIFT_MIGRATION_ID(id<SentryScopeObserver>))observer
+    NS_SWIFT_NAME(add(_:));
 
 @end
 


scripts/check-objc-id-usage.py
```diff
@@ -0,0 +1,327 @@
+#!/usr/bin/env python3
```
philprime: `m`: Python3 is not pre-installed on macOS by default, therefore we should include either a `.python-version` which allows installing it via version managers such as `pyenv`, or add it to the `Brewfile`

scripts/check-objc-id-usage.py
```diff
@@ -0,0 +1,327 @@
...
+- Is not in a comment
+- Does not have an inline comment explaining why bare 'id' is needed
+
+Allowed patterns:
+1. SENTRY_SWIFT_MIGRATION_ID(ClassName)
+2. id someVar; // OK: explanation why bare id is needed
+3. id someVar; /* OK: explanation */
+"""
+
+from __future__ import print_function, unicode_literals
```
philprime: `m`: Instead of introducing yet another scripting language in addition to bash, ruby, swift, javascript (in CI) could we instead implement this as a custom fastlane action instead?

scripts/check-objc-id-usage.py
itaybre: I am almost sure, it does now

scripts/check-objc-id-usage.py
itaybre: I am fine with either scripting language, I let claude use python since we already use that for [clang](https://github.com/getsentry/sentry-cocoa/blob/main/scripts/check-clang-format.py)

Migrating this to ruby should be easy

philprime: Ah my bad then, I didn't realize that script is part of our repo. Then feel free to leave it as-is

itaybre: Rewrote it in ruby

Makefile
```diff
@@ -339,21 +339,23 @@ STAGED_SWIFT_FILES := $(shell git diff --cached --diff-filter=d --name-only | gr
...
 
 ## Run linting checks on staged files only
 #
-# Runs SwiftLint, Clang-Format checks, and dprint checks on staged files only.
+# Runs SwiftLint, Clang-Format checks, Objective-C id usage checks, and dprint checks on staged files only.
 .PHONY: lint-staged
 lint-staged:
 	@echo "--> Running Swiftlint and Clang-Format on staged files"
 	./scripts/check-clang-format.py -r Sources Tests
+	ruby ./scripts/check-objc-id-usage.rb -r Sources/Sentry
```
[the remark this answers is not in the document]
itaybre: This is fine

philprime:
## Review: Manual Testing Results ✅

I manually tested the `check-objc-id-usage.rb` script and verified it works as expected for the core use cases.

### Test Results

| Pattern | Expected | Result |
|---------|----------|--------|
| `@property (...) id prop` | Flag | ✅ Flagged |
| `- (id)methodName` | Flag | ✅ Flagged |
| `:(id)param` | Flag | ✅ Flagged |
| `id _ivar;` | Flag | ✅ Flagged |
| `id<Protocol>` | Skip | ✅ Skipped |
| `- (id)init` / `- (id)initWith...` | Skip | ✅ Skipped |
| `// OK: comment` | Skip | ✅ Skipped |
| `/* OK: comment */` | Skip | ✅ Skipped |
| Multi-line declaration with `// OK:` at end | Skip | ✅ Skipped |
| `SENTRY_SWIFT_MIGRATION_ID(X)` | Skip | ✅ Skipped |
| Code inside comments | Skip | ✅ Skipped |
| Both bare `id` and `id<Protocol>` on same line | Flag bare only | ✅ Works |

### Potential Gaps (for future consideration)

The script currently does not detect these patterns:

1. **Nullability specifiers before `id`**: `(nullable id)`, `(nonnull id)`, `(_Nullable id)`
   - Example: `SentrySpanProtocol.h:119` - `setDataValue:(nullable id)value` is not flagged

2. **C-style function parameters** (without parentheses):
   - Example: `SentryProfilingSwiftHelpers.h` - `id observer`, `id target` are not flagged

These may be intentional exclusions since the script focuses on Objective-C method/property declarations. If these should be covered, it could be addressed in a follow-up PR.

### Verification

```bash
# Script runs clean on codebase
ruby ./scripts/check-objc-id-usage.rb -r Sources/Sentry
# Exit code: 0
```

The existing `SentryScope.h` change correctly demonstrates the `// OK:` comment pattern for multi-line declarations.

LGTM! 🚀

PR #7182, opened 2026-01-15
feat: add metrics flushing support to SentrySDK.flush()

### LINKED ISSUE #6950: [Metrics] Verify metrics can be flushed manually
Add or update the root-level method `SentrySDK.flush()` to manually flush pending metrics and events to Sentry

philprime:
## Summary

This PR adds manual metrics flushing support to `SentrySDK.flush()` by implementing a protocol-based approach that allows flushable integrations to be flushed during manual flush operations and app lifecycle events.

## Changes

- **Added FlushableIntegration protocol** - Type-safe protocol for integrations that support manual flushing
- **Made SentryMetricsIntegration conform to FlushableIntegration** - Enables metrics to be flushed via the protocol
- **Moved integration flushing from Client to Hub** - Fixes circular dependency violation where Client was accessing Hub
- **Added lifecycle-based flushing** - Metrics are automatically flushed on `willResignActive` and `willTerminate` notifications
- **Updated Hub.flush()** - Now flushes integrations before delegating to Client for logs and transport

## Architecture

**Before (circular dependency):**
```
SentrySDK.flush() → SentryHub.flush() → SentryClient.flush() → SentrySDKInternal.currentHub ❌
```

**After (clean dependency tree):**
```
SentrySDK.flush() → SentryHub.flush() → Hub.flushIntegrations() → Client.flush() ✅
```

## Testing

- Added tests for Hub.flush() behavior with metrics integration
- Added lifecycle event tests for willResignActive and willTerminate
- Added protocol conformance tests

## Related

- Follows the same pattern as logs flushing but uses protocol-based approach for better extensibility
- Uses existing `captureMetricsData` method correctly (receives batched data from callback)

Closes #6950

### DIFF
--- a/Sources/Swift/Integrations/Metrics/SentryMetricsIntegration.swift
+++ b/Sources/Swift/Integrations/Metrics/SentryMetricsIntegration.swift
@@ -20,7 +20,7 @@ typealias SentryMetricsIntegrationDependencies = DateProviderProvider & Dispatch
 
 final class SentryMetricsIntegration<Dependencies: SentryMetricsIntegrationDependencies>: NSObject, SwiftIntegration, SentryMetricsIntegrationProtocol, FlushableIntegration {
     private let metricBatcher: SentryMetricsBatcherProtocol
-#if ((os(iOS) || os(tvOS) || os(visionOS)) && !SENTRY_NO_UIKIT) || os(macOS)
+    #if ((os(iOS) || os(tvOS) || os(visionOS)) && !SENTRY_NO_UIKIT) || os(macOS)
     private let notificationCenter: SentryNSNotificationCenterWrapper
     #endif
 
@@ -41,7 +41,7 @@ final class SentryMetricsIntegration<Dependencies: SentryMetricsIntegrationDepen
             }
         )
 
-#if ((os(iOS) || os(tvOS) || os(visionOS)) && !SENTRY_NO_UIKIT) || os(macOS)
+        #if ((os(iOS) || os(tvOS) || os(visionOS)) && !SENTRY_NO_UIKIT) || os(macOS)
         self.notificationCenter = dependencies.notificationCenterWrapper
         #endif
 


philprime:
@sentry review

Sentry.xcodeproj/project.pbxproj
```diff
@@ -4974,6 +4974,7 @@
 				9264E1EA2E2E385B00B077CF /* SentryLogMessage.swift */,
 				D46B04472EDF25E100AF4A0A /* SentryMetric.swift */,
 				D48953FD2EF5756A0086F240 /* SentryMetricValue.swift */,
+				D48954D92EF94FD20086F240 /* SentryUnit.swift */,
```
philprime:
```suggestion
```

philprime:
@sentry review
@cursor review

Sources/Sentry/SentryHub.m
```diff
@@ -818,12 +818,52 @@ - (NSString *)createSessionDebugString:(SentrySession *)session
 
 - (void)flush:(NSTimeInterval)timeout
 {
+    // Flush flushable integrations first (e.g., metrics)
+    NSTimeInterval flushIntegrationsDuration = [self flushIntegrations];
+
+    // Calculate remaining timeout for client flush (logs and transport)
+    NSTimeInterval remainingTimeout = fmax(timeout / 2, timeout - flushIntegrationsDuration);
```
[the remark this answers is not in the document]
philippsuess: `m`: I think that's a valid comment. If flushing of the integrations already consumed all the time, we should just trigger a flush with 0.0 for the client. This will then trigger the flush, but we don't wait for it to succeed.

Sources/Sentry/SentryHub.m
```diff
@@ -818,12 +818,52 @@ - (NSString *)createSessionDebugString:(SentrySession *)session
...
+        for (id<SentryIntegrationProtocol> integration in _installedIntegrations) {
+            // We use respondsToSelector: instead of conformsToProtocol: for FlushableIntegration
+            // because Swift @objc protocols defined in files with @_implementationOnly import
+            // are not reliably accessible from Objective-C code at compile time. The protocol
+            // definition exists for Swift type safety and documentation, but Objective-C code
+            // needs to use respondsToSelector: to check for the flush method dynamically.
+            // This Swift-first approach allows us to maintain protocol-oriented design in Swift
+            // while ensuring Objective-C interop works correctly.
+            SEL flushSelector = NSSelectorFromString(@"flush");
+            if ([integration respondsToSelector:flushSelector]) {
```
philippsuess, line 854: `m`: We're currently converting all integrations to Swift, and right now only one Swift integration implements `FlushableIntegration`. What about migrating this code to Swift so we don't need to use selectors here?

Sources/Sentry/SentryClient.m
```diff
@@ -625,10 +625,12 @@ - (SentryEvent *_Nullable)prepareEvent:(SentryEvent *)event
...
     NSTimeInterval captureLogsDuration = [self.logBatcher captureLogs];
-    // Capturing batched logs should never take long, but we need to fall back to a sane value.
-    // This is a workaround for in-memory logs, until we'll write batched logs to disk,
-    // to avoid data loss due to crashes. This is a trade-off until then.
-    [self.transportAdapter flush:fmax(timeout / 2, timeout - captureLogsDuration)];
+    // Calculate remaining timeout for transport flush.
+    // We subtract the time already spent capturing logs to respect the overall timeout.
+    // If log capture took longer than the timeout, we still give the transport a small
+    // chance (0.1s minimum) to flush critical data.
+    NSTimeInterval remainingTimeout = fmax(0.1, timeout - captureLogsDuration);
```
philippsuess: `m`: I think calling this with 0.0 is legit. 0.0 will still trigger sending events, but it won't block.

philipphofmann (commented):
I'm a bit unsure about putting the flushing logic and lifecycle hooks into the integration regarding the [telemetry processor architecture](https://develop.sentry.dev/sdk/telemetry/telemetry-processor/). There, the telemetry buffer is responsible for flushing all data on [mobile](https://develop.sentry.dev/sdk/telemetry/telemetry-processor/mobile-telemetry-processor/#telemetry-buffer) for the lifecycle events.

So either we adapt the architecture of the telemetry processor, or we fix it here. I don't think that the metrics integration should be responsible for flushing, actually. IMO, it should only forward the metrics and let the telemetry processor take care of the rest. Please have a look at the [telemetry processor](https://develop.sentry.dev/sdk/telemetry/telemetry-processor) and let me know if we could adapt to that architecture or if we should revisit the architecture. I'm also happy to discuss this on a call.

Sources/Sentry/SentryHub.m
philprime: What do you mean with "this code"? The entire hub?

philprime:
@philipphofmann the key reason why I introduced flushable integrations is this because the Logs feature was designed with the log batcher living on the SentryClient, but for Metrics it's part of the integration. 

Logs doesn't even have a proper integration, @denrase only created a minimal `FlushLogsIntegration` to get flushing working (which is fine!), while I created a full integration for Metrics - they are a bit different, because the public `SentryLogger` sends events directly to the hub via a delegate pattern, but the `SentryMetricsApi` sends it to the `SentryMetricsIntegration` which then performs batching and sends the batched data to the client (fetched via the hub). So Metrics bypasses the hub and goes straight for the client. Ideally the client would also be injected into the integration, to further get rid of static accessors (e.g. `SentrySDKInternal.currentHub()`)

My proposed flow is this:

```
SentrySDK.flush(timeout:)                                     (App)
└── SentrySDKInternal.flush(timeout:)                         (SentrySDK.swift)
    └── [SentrySDKInternal.currentHub flush:]                 (SentrySDKInternal.m)
        └── [flushIntegrations flush]                         (SentryHubInternal.m)
            └── SentryMetricsIntegration.flush()            
                └── metricBatcher.captureMetrics()            (SentryMetricsIntegration.swift
                    └── batcher.captureMetrics()              (SentryMetricsBatcher.swift)
                        └── performCaptureItems               (SentryBatcher.swift)
                            └── client.captureMetricsData(_:) (SentryMetricsIntegration.swift)   
                                └── self.captureData(_:)      (SentryClient.m)
```

This was designed and built before you started working on the TelemetryProcessor/-Buffer and merged in multiple refactorings. I created a generic batcher in https://github.com/getsentry/sentry-cocoa/pull/7017 so we can abstract the shared logic, so until you have finished working out the telemetry processor/buffer, that's how we have it in the code base.

Now the SentryClient is used by the SentryHub, so the hub knows about the client, but the client should not know about the hub. If flushing of metrics needs to happen in the client, 
the SentryClient does not have access to the MetricsBatcher right now. The workaround right now would be in the client to get the global SentryHub, to then grab the SentryMetricsIntegration, to then access the MetricsBatcher to then flush it. That's a circular dependency from client to hub.

Alternatively, we would have to move the MetricsBatcher to live the SentryClient too, which is pretty much the first step torwards the implementation of the Telemetry Processor, so feel free to refactor the LogsBatcher, MetricsBatcher and Batcher into the Telemetry Processor/Buffer - but that's not the scope of this PR.

What's the ETA for the Telemetry Processor? If it's done by end of next week, I am fine to wait for it before finishing implementing flushing of Metrics. But if not, I do not think we should hold up such a pure internal change like this PR for that. We must refactor the whole thing anyways, but at least we could get done the intermediate step and conclude the project Metrics (this PR and the feature docs are the final two tasks).

philippsuess:
> What's the ETA for the Telemetry Processor? If it's done by end of next week, I am fine to wait for it before finishing implementing flushing of Metrics.

We already agreed on the general approach. What you see in the develop docs, is meant to be ready to implement. We still might adapt it due to feedback, but we would only change it if we get feedback that it doesn't work. 

The scope of this PR shouldn't be refactoring everything to already match the telemetry processor architecture. While I understand the reasons for how you came up with the current approach, I would try to get us a step closer to telemetry processor architecture with this PR. 

The proposed flush flow in this PR shows some drawbacks, IMO:
1. Subtracting flushing durations in the hub seems a bit wrong to me. The flushing logic should be in a single place, not multiple places. Ensuring correct flushing behaviour is already more challenging now.
2. The integrations know too much. Previously, they mostly collected data and forwarded it. Now we're mixing that with flushing logic. The flushing logic should be close to where the data is collected and forwarded. This is the telemetry processor.

We can keep this approach and merge the PR, but we should already mark the FlushableIntegration with a comment that it's about to be ditched soon, to avoid using it in other integrations.

philippsuess:
@philprime and I just decided in a DM on Slack, that we can keep the current approach of the FlushableIntegration. I plan on refactoring the logs and metrics towards the telemetry processor and will address that problem then.

philprime:
I added more timeout logic in the for-loop and added more tests.

PR #7185, opened 2026-01-15
feat: Use full flamegraph for metrickit app hangs

### LINKED ISSUE #7636: feat: Use full flamegraph for metrickit app hangs
> [!NOTE]
> The pull request "[feat: Use full flamegraph for metrickit app hangs](https://github.com/getsentry/sentry-cocoa/pull/7185)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Before my PR https://github.com/getsentry/sentry-cocoa/pull/6908 we would send the full flamegraph as a linear stacktrace, so when users looked at the metric kit hang stacktrace it wouldn't make sense. The previous PR fixed the stacktrace UI, but only one sample from the hang was captured. This could also negatively affect grouping because all the data was not being recorded.

Now, we again capture the entire hang but add the new parent/count parameters which I implemented on the frontend here: https://github.com/getsentry/sentry/pull/106412 This enables seeing the entire hang and all the functions that contributed it to it.

Screenshot 2026-03-05 at 5 53 44 PM

noahsmartin:
Before my PR https://github.com/getsentry/sentry-cocoa/pull/6908 we would send the full flamegraph as a linear stacktrace, so when users looked at the metric kit hang stacktrace it wouldn't make sense. The previous PR fixed the stacktrace UI, but only one sample from the hang was captured. This could also negatively affect grouping because all the data was not being recorded.

Now, we again capture the entire hang but add the new parent/count parameters which I implemented on the frontend here: https://github.com/getsentry/sentry/pull/106412 This enables seeing the entire hang and all the functions that contributed it to it.

Screenshot 2026-03-05 at 5 53 44 PM

Closes #7636

### DIFF
--- a/Sources/Sentry/Public/SentryFrame.h
+++ b/Sources/Sentry/Public/SentryFrame.h
@@ -75,6 +75,9 @@ NS_SWIFT_NAME(Frame)
  */
 @property (nonatomic, copy) NSNumber *_Nullable parentIndex;
 
+/**
+ * Number of samples recorded that contained this frame, used for flamegraphs.
+ */
 @property (nonatomic, copy) NSNumber *_Nullable sampleCount;
 
 /**
--- a/Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
+++ b/Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
@@ -20,8 +20,6 @@ extension SentryMXCallStackTree {
     // it represents data that is sampled across many threads and aggregated, we do not
     // know which samples came from which thread. Instead we create just one fake thread
     // that just contains the most common callstack.
-    //
-    // We hope to add flamegraph support at some point and that is tracked here: https://github.com/getsentry/sentry-cocoa/issues/7062
     func sentryMXBacktrace(inAppLogic: SentryInAppLogic?, handled: Bool) -> [SentryThread] {
         callStacks.enumerated().map { index, callStack in
             let thread = SentryThread(threadId: NSNumber(value: index))
@@ -53,7 +51,7 @@ extension SentryMXCallStackTree {
     }
 
     /// Flattens the call stack tree into a single thread with all frames.
-    /// Each frame includes metadata in its `vars` field to allow reconstructing the original tree:
+    /// Each frame includes metadata to allow reconstructing the original tree:
     /// - `parent_frame_index`: The index of the parent frame in the flat list (-1 for root frames)
     /// - `sample_count`: The number of samples at this frame
     ///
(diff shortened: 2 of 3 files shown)


Sources/Swift/Core/MetricKit/SentryMXManager.swift
```diff
@@ -95,12 +95,29 @@ final class SentryMXManager: NSObject, MXMetricManagerSubscriber {
...
+        // First look for the crashing thread, but for events that were not a crash (like a hang) take the first thread
+        // since those events only report one thread
+        let exceptionThread = threads.first { $0.crashed?.boolValue == true } ?? threads.first
+        event.debugMeta = debugMeta
+        event.threads = threads
+
+        if let exceptionThread, let exception = event.exceptions?[0] {
+            exception.stacktrace = exceptionThread.stacktrace
+            exception.threadId = exceptionThread.threadId
+        }
```
philippsuess, line 120: `h`: I know this is a draft, but this question is crucial to ask. Do you have any clue how grouping is going to work for these types of exceptions?

noahsmartin: It will be back to how it was before my PR here: https://github.com/getsentry/sentry-cocoa/pull/6908 which means grouping can again use all frames from the hang, rather than just the ones from the single sample.

Sources/Sentry/Public/SentryFrame.h
```diff
@@ -70,6 +70,13 @@ NS_SWIFT_NAME(Frame)
  */
 @property (nonatomic, copy) NSString *_Nullable contextLine;
 
+/**
+ * Index of the parent frame used for flamegraphs.
+ */
+@property (nonatomic, copy) NSNumber *_Nullable parentIndex;
+
+@property (nonatomic, copy) NSNumber *_Nullable sampleCount;
```
itaybre, line 81: `m`: Can we have a simple explanation about this item?

Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
```diff
@@ -66,6 +51,40 @@ extension SentryMXCallStackTree {
...
+            thread.stacktrace = SentryStacktrace(frames: frames, registers: [:])
+            thread.crashed = NSNumber(value: (callStack.threadAttributed ?? false) && !handled)
+            return thread
+        }
+    }
+
+    private func flattenFrame(_ mxFrame: SentryMXFrame, parentIndex: Int, frames: inout [Frame], inAppLogic: SentryInAppLogic?) {
+        let currentIndex = frames.count
+        let frame = mxFrame.toSentryFrameWithTreeData(frameIndex: currentIndex, parentFrameIndex: parentIndex)
+        frame.inApp = NSNumber(value: inAppLogic?.is(inApp: frame.package) ?? false)
```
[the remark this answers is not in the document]
itaybre: @noahsmartin this is right, isn't it?
When the package is missing, you mentioned it was always inApp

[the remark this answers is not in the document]
noahsmartin: Yes I think so, will fix. I also had a test for this so will need to see what happened

Issue #6178, opened 2025-09-17
App context in_foreground is true without foreground crumb


philipphofmann:
### Description

When an app crashes when coming to the foreground in between `UIApplicationWillEnterForegroundNotification` and `UIApplicationDidBecomeActiveNotification` the app context has `in_foreground = true`, but the inForeground breadcrumb is missing. This confuses users.

This is because the `SentryBreadcrumbTracker` only adds a breadcrumb for the `UIApplicationDidBecomeActiveNotification`.

https://github.com/getsentry/sentry-cocoa/blob/fccc4e5b5f10915b8a43d4ed756c02daef5e88dd/Sources/Sentry/SentryBreadcrumbTracker.m#L92

SentryCrash already sets `in_foreground = true` on the `UIApplicationWillEnterForegroundNotification` here https://github.com/getsentry/sentry-cocoa/blob/fccc4e5b5f10915b8a43d4ed756c02daef5e88dd/Sources/SentryCrash/Recording/SentryCrash.m#L539-L542

Strictly speaking, an app isn't in the foreground when receiving the `applicationWillEnterForeground`, as the OS posts this when `the app is about to enter the foreground`. So we should only update the app context on the `applicationDidBecomeActive`. Ideally, we could also add a breadcrumb for `applicationWillEnterForeground`.

philprime:
Additional note: it looks like KSCrash is still doing it the same way, so this has been wrong even in the upstream implementation

https://github.com/kstenerud/KSCrash/blob/257e3754f676b52707afaeb3e9ef7dbbc188d03a/Sources/KSCrashRecording/KSCrash.m#L448-L451

naftaly:
> Additional note: it looks like KSCrash is still doing it the same way, so this has been wrong even in the upstream implementation
> 
> https://github.com/kstenerud/KSCrash/blob/257e3754f676b52707afaeb3e9ef7dbbc188d03a/Sources/KSCrashRecording/KSCrash.m#L448-L451

When being foregrounded, the user intent is usually to interact with the app and the "foregrounded" flag is usually an important one that apps use to filter in important crashes. That's the reason why it's set up as it is in KSCrash. 

Some new work from a few months ago tries to make this a lot more explicit with a much more robust app state tracker. https://github.com/kstenerud/KSCrash/blob/257e3754f676b52707afaeb3e9ef7dbbc188d03a/Sources/KSCrashRecording/include/KSCrashAppTransitionState.h#L48

[...]

PR #7188, opened 2026-01-16
fix: in_foreground and is_active app context

philipphofmann:
## :scroll: Description

This PR aligns the `in_foreground` app context behaviour for both handled and unhandled events and also adds a `is_active` property to the app context.

https://github.com/getsentry/sentry/pull/107047 will correctly format is_active in the issue details view.

Next steps:
* I'm going to add the breadcrumbs for when the app becomes active in an upcoming PR to fix the final part of GH-6178. 

First step for GH-6178

Screenshot 2026-01-27 at 13 50 45

Unit tests and simulator

## :pencil: Checklist

### DIFF
--- a/Tests/Perf/metrics-test.yml
+++ b/Tests/Perf/metrics-test.yml
@@ -11,4 +11,4 @@ startupTimeTest:
 
 binarySizeTest:
   diffMin: 200 KiB
-  diffMax: 1100 KiB
+  diffMax: 1200 KiB


Sources/Sentry/SentryBreadcrumbTracker.m
```diff
@@ -77,7 +77,10 @@ - (void)stop
 - (void)trackApplicationNotifications
 {
 #if SENTRY_HAS_UIKIT
-    NSNotificationName foregroundNotificationName = UIApplicationDidBecomeActiveNotification;
```
philippsuess: Here's a key change. The foreground crumb now gets added when `UIApplicationWillEnterForegroundNotification` and not when `UIApplicationDidBecomeActiveNotification`. I think that's more accurate.

Sources/Sentry/SentryClient.m
```diff
@@ -716,15 +716,22 @@ - (SentryEvent *_Nullable)prepareEvent:(SentryEvent *_Nullable)event
...
+                && (context[@"app"][@"in_foreground"] == nil
+                    || context[@"app"][@"is_active"] == nil))) {
             NSMutableDictionary *app =
                 [(NSDictionary *)context[@"app"] mutableCopy] ?: [NSMutableDictionary dictionary];
             context[@"app"] = app;
 
             UIApplicationState appState =
                 [SentryDependencyContainer sharedInstance].threadsafeApplication.applicationState;
-            BOOL inForeground = appState == UIApplicationStateActive;
-            app[@"in_foreground"] = @(inForeground);
```
philippsuess, line 727: Here was the mismatch between fatal and non-fatal events. Here we used to set in_foreground to true if the app was active, but the app could also be in the foreground but not active and then this was false before. Now it's aligned with SentryCrash https://github.com/getsentry/sentry-cocoa/blob/7e93d856f3af1f778ddc993c42b46497833c9b0a/Sources/SentryCrash/Recording/SentryCrash.m#L521-L539

Sources/Sentry/SentryBreadcrumbTracker.m
```diff
@@ -128,6 +131,28 @@ - (void)trackApplicationNotifications
                                                                     withDataKey:@"state"
                                                                   withDataValue:@"foreground"];
                                                 }];
+
+    [NSNotificationCenter.defaultCenter addObserverForName:didBecomeActiveNotificationName
```
itaybre: Looks like a swift name

Sources/Swift/Helper/SentryEventContextEnricher.swift
```diff
@@ -0,0 +1,73 @@
...
+    /// Enriches the event context dictionary with app state fields (in_foreground and is_active).
+    /// - Parameter context: The current event context dictionary.
+    /// - Returns: The enriched context dictionary with app state information added.
+    func enrichWithAppState(_ context: [String: Any]) -> [String: Any]
+}
+
+/// Default implementation of event context enrichment.
+/// On UIKit platforms, adds in_foreground and is_active fields to the app context based on UIApplicationState.
+/// On other platforms, returns the context unchanged.
+class SentryDefaultEventContextEnricher: SentryEventContextEnricher {
```
philprime, line 18: `m`: Is there a reason you decided against POP here and decided to go for the old approach of creating a subclass `TestEventContextEnricher` instead?

philprime (approved):
LGTM with one feedback comment

Sources/Swift/Helper/SentryEventContextEnricher.swift
philippsuess: Old habits. Thanks for pointing it out.

philippsuess: Ah sorry, I forgot: The `SentryEventContextEnricher` protocol needs to be available in ObjC. I think the POP approach doesn't work then, or am I mistaken?

philprime: I think it could still work, but also fine to leave it as-is and refactor it in another PR

philippsuess: I then just keep it as is. We can change it later if required.

PR #7190, opened 2026-01-16
chore: Use pull_request_target for changelog preview

### LINKED ISSUE #7191: chore: Use pull_request_target for changelog preview
> [!NOTE]
> The pull request "[chore: Use pull_request_target for changelog preview](https://github.com/getsentry/sentry-cocoa/pull/7190)" was created by @BYK but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

## Summary

Updates the changelog-preview workflow to use `pull_request_target` instead of `pull_request`.

## Why?

This change is required for the changelog preview to work correctly with PRs from forks. With `pull_request`, the workflow runs with a read-only `GITHUB_TOKEN` which cannot post comments or create statuses on fork PRs.

This aligns with the recommended usage in the [craft changelog-preview workflow](https://github.com/getsentry/craft/blob/master/.github/workflows/changelog-preview.yml).

## Security Note

This change is safe because the changelog-preview workflow:
- Downloads the Craft binary from releases (not from the PR)
- Only reads git metadata and configuration  
- Does not execute any code from the PR

BYK:
## Summary

Updates the changelog-preview workflow to use `pull_request_target` instead of `pull_request`.

## Why?

This change is required for the changelog preview to work correctly with PRs from forks. With `pull_request`, the workflow runs with a read-only `GITHUB_TOKEN` which cannot post comments or create statuses on fork PRs.

This aligns with the recommended usage in the [craft changelog-preview workflow](https://github.com/getsentry/craft/blob/master/.github/workflows/changelog-preview.yml).

## Security Note

This change is safe because the changelog-preview workflow:
- Downloads the Craft binary from releases (not from the PR)
- Only reads git metadata and configuration  
- Does not execute any code from the PR

Closes #7191

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -1,6 +1,6 @@
 name: Changelog Preview
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened, edited, labeled]
 
 permissions:


.github/workflows/changelog-preview.yml
```diff
@@ -1,6 +1,6 @@
 name: Changelog Preview
 on:
-  pull_request:
+  pull_request_target:
```
itaybre, line 3: TIL, definitely worth reviewing if we need this on some other workflows that post to GH

philprime: I believe we do not for now as this can be a security risk and we should enable it on a need-to-have basis.

PR #7211, opened 2026-01-20
chore: add file renaming guidelines to AGENTS.md

### LINKED ISSUE #7212: chore: add file renaming guidelines to AGENTS.md
> [!NOTE]
> The pull request "[chore: add file renaming guidelines to AGENTS.md](https://github.com/getsentry/sentry-cocoa/pull/7211)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Add guidelines to AGENTS.md for preserving git history when renaming files using `git mv`.

Also added `git mv` to Claude Code settings permissions.

#skip-changelog

philipphofmann:
Add guidelines to AGENTS.md for preserving git history when renaming files using `git mv`.

Also added `git mv` to Claude Code settings permissions.

#skip-changelog

Closes #7212

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -294,6 +294,10 @@ git commit -m "ref: rename File.swift to NewName.swift"
 - Code archaeology and debugging are easier
 - Refactoring history is preserved
 
+**Note on Squash Merges:**
+
+These benefits apply even when PRs are squash-merged. Git's rename detection works on the final diff of the squashed commit, so using `git mv` in feature branches ensures the squashed commit is recognized as a rename (shown as `R` in git log) rather than a delete + add. This preserves `git log --follow` and other history-tracking features after merging.
+
 **Verification:**
 
 After renaming, verify that git recognizes the rename:


AGENTS.md
```diff
@@ -245,6 +245,64 @@ When an error path cannot be reliably tested:
...
+```bash
+# WRONG - breaks history tracking
+mv old-name.swift new-name.swift
+git add new-name.swift
+git commit -m "ref: rename old-name to new-name"
+```
+
+**Multi-step Renames:**
+
+If a rename requires multiple steps (e.g., moving and renaming, or updating imports before renaming), perform each step in a separate commit:
```
philprime, line 273: `l`: I mean while I agree, there isn't much of a point in this as we are using squash commits anyways, so all multi-commit refactoring is lost in the end anyways

philippsuess: Thanks for the feedback! I've addressed this concern by adding a \"Note on Squash Merges\" section to AGENTS.md explaining why \`git mv\` still matters even with squash commits.

I also tested this to verify: when Git creates a squashed commit, it applies rename detection on the final diff. Using \`git mv\` in the feature branch ensures the squashed commit shows the change as a rename (R050 in git log) rather than delete + add, which preserves \`git log --follow\` and other history-tracking features.

See the detailed testing results in my comment below.

philippsuess:
## TL;DR

Using `git mv` **does** preserve history even with squash merges. Git detects renames in the squashed commit based on content similarity, and `git log --follow` works correctly after squash merging.

## Detailed Testing

I tested this claim with a real Git scenario to verify whether `git mv` matters when using squash commits. Here's what I found:

### Test Setup

I created a test repository and simulated the exact workflow we use:

```bash
# Create initial file
git init test-repo && cd test-repo
echo "line 1" > old.txt
git add old.txt
git commit -m "initial"

# Create feature branch with rename using git mv
git checkout -b feature
git mv old.txt new.txt
git commit -m "rename"

# Make additional changes
echo "line 2" >> new.txt
git commit -am "modify"

# Squash merge to main
git checkout main
git merge --squash feature
git commit -m "squashed: rename and modify"
```

### Results

After the squash merge, checking the git history shows:

```bash
$ git log --oneline --name-status
87adb4a squashed: rename and modify
R050    old.txt new.txt    ← Git correctly detects this as a RENAME
d58ba47 initial
A       old.txt
```

The key indicator is **`R050`** which means Git recognized this as a **rename with 50% similarity**, not a delete + add.

More importantly, `git log --follow` works correctly:

```bash
$ git log --follow --oneline new.txt
87adb4a squashed: rename and modify
d58ba47 initial    ← History traces back through the rename!
```

### Why This Works

When Git creates a squashed commit, it:
1. Compares the base branch state with the final state after all changes
2. Applies **rename detection** on that final diff
3. Recognizes files as renamed if content similarity exceeds the threshold

Using `git mv` in the feature branch helps Git detect the rename because:
- It explicitly marks the operation as a rename (100% similarity initially)
- Even with subsequent content changes, Git's similarity detection still works
- The squashed commit preserves this rename detection in the final diff

### Without `git mv`

If we had used filesystem operations (`mv` + `git add`), Git might show:
```
D       old.txt
A       new.txt
```

This breaks:
- `git log --follow new.txt` (won't show history before rename)
- `git blame` (shows all lines as "new")
- History bisection across the rename

### Conclusion

The multi-step process with `git mv` **absolutely matters** for squash merges. It ensures that even after squashing, Git's rename detection works correctly, and all history-tracking features (`git log --follow`, `git blame`, etc.) continue to function properly.

philprime:
I performed the same validation approach using `mv` + `git add` and it yields the exact same result:

**Test Setup:**

```shell
# Create initial file
git init test-repo && cd test-repo
echo "line 1" > old.txt
git add old.txt
git commit -m "initial"

# Create feature branch with rename using mv + git add
git checkout -b feature
mv old.txt new.txt
git add old.txt new.txt
git commit -m "rename"

# Make additional changes
echo "line 2" >> new.txt
git commit -am "modify"

# Squash merge to main
git checkout main
git merge --squash feature
git commit -m "squashed: rename and modify"
```

**Result:** 

```
$ git log --oneline --name-status
7a6875b (HEAD -> main) squashed: rename and modify
R050    old.txt new.txt
a6b2b66 initial
A       old.txt
```

Therefore it's important to note that `git mv` does **not** have an impact in squash commits as it's purely based on a similarity diff.

PR #7245, opened 2026-01-21
ref: Migrate SentrySession from Objective-C to Swift

### LINKED ISSUE #7247: ref: Migrate SentrySession from Objective-C to Swift
> [!NOTE]
> The pull request "[ref: Migrate SentrySession from Objective-C to Swift](https://github.com/getsentry/sentry-cocoa/pull/7245)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

- Move all implementation from SentrySessionInternal (ObjC) into SentrySession (Swift)
- Convert SentrySessionStatus to a String-based enum
- Add equality methods (isEqual, hash) directly to SentrySession
- Delete SentrySessionInternal.h, SentrySessionInternal.m
- Delete SentrySession+Equality.h, SentrySession+Equality.m
- Update SentryPrivate.h to remove SentrySessionInternal import
- Update tests to use Swift implementation

#skip-changelog

noahsmartin:
- Move all implementation from SentrySessionInternal (ObjC) into SentrySession (Swift)
- Convert SentrySessionStatus to a String-based enum
- Add equality methods (isEqual, hash) directly to SentrySession
- Delete SentrySessionInternal.h, SentrySessionInternal.m
- Delete SentrySession+Equality.h, SentrySession+Equality.m
- Update SentryPrivate.h to remove SentrySessionInternal import
- Update tests to use Swift implementation

#skip-changelog

Closes #7247

### DIFF
--- a/Sources/Swift/SentrySession.swift
+++ b/Sources/Swift/SentrySession.swift
@@ -15,7 +15,7 @@ enum SentrySessionStatus: String {
 
     // MARK: - Private properties
 
-    private let lock = NSLock()
+    private let lock = NSRecursiveLock()
     private var _sessionId: UUID
     private var _started: Date
     private var _status: SentrySessionStatus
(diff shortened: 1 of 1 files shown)


Sources/Swift/SentrySession.swift
```diff
@@ -2,136 +2,378 @@
...
     }
+    // swiftlint:enable cyclomatic_complexity
+
+    // MARK: - Public methods
 
     @objc(endSessionExitedWithTimestamp:)
     public func endExited(withTimestamp timestamp: Date) {
-        session.endSessionExited(withTimestamp: timestamp)
+        lock.lock()
+        defer { lock.unlock() }
```
philippsuess: You could also use the NSLock extension

https://github.com/getsentry/sentry-cocoa/blob/d8b9a958ef9dad14594e5b57185291feae894c61/Sources/Swift/Core/Extensions/Locks.swift#L12-L16

Sources/Swift/SentrySession.swift
```diff
@@ -2,136 +2,378 @@
...
+
+    @objc public func serialize() -> [String: Any] {
+        lock.lock()
+        defer { lock.unlock() }
+
+        var serializedData: [String: Any] = [
+            "sid": _sessionId.uuidString,
+            "errors": _errors,
+            "started": sentry_toIso8601String(_started)
+        ]
```
philippsuess: `h`: As far as I know, NSLock isn't reentrant, but it seems like we're acquiring the lock here two times. First at the start of the method and then in the getters. I wonder why this doesn't end up in a deadlock actually. If I'm not mistaken, `@synchronized` in ObjC uses a reentrant lock. It's worth double-checking if we shouldn't rather use `NSRecursiveLock`

Sources/Swift/SentrySession.swift
noahsmartin: Yes Claude didn't know about it apparently, I will teach it

Sources/Swift/SentrySession.swift
noahsmartin: We only acquire the lock here once at the start of the function, the getters aren't called, the values are accessed directly once the lock is held. It should be more performant than acquiring the lock a bunch of times. That being said, can't really hurt to just use NSRecursiveLock here, the perf isn't too critical. Next re-write we should make this a struct and copy the value anytime we need to change it so we don't have to worry about locks, but that will have to wait until more is in Swift

Sources/Swift/SentrySession.swift
philippsuess: @noahsmartin, you could put the info into the Agent.md file. Then, all AI agents in our codebase should be aware.

Sources/Swift/SentrySession.swift
philippsuess: Thanks, for double-checking 🙏

PR #7261, opened 2026-01-23
feat: Transport logs error message for HTTP 413 responses

### LINKED ISSUE #7248: SDK Handling HTTP 413 [Apple]
Relay previously returned HTTP 400 Bad Request for oversized envelopes. This was changed in Relay [#5474](<https://github.com/getsentry/relay/pull/5474>) to HTTP 413 (Content Too Large) to allow SDKs to distinguish size-related rejections from other errors. For more context on this decision, see the [DACI: Relay Returns 413 for too-large envelopes.](<https://www.notion.so/sentry/DACI-Relay-returns-413-for-too-large-envelopes-2c68b10e4b5d8034be83e524c855e73d>)

The main benefit is for our users. With the HTTP 413 response, SDKs can now log specific, actionable error messages when envelopes are rejected due to size limits. Previously, users had no clear indication of why their data was being dropped when envelopes exceeded size limits. Now, they'll understand the cause and can take steps to resolve it.

### Requirements

See the [413 section in Dealing With Network Failures in the SDK develop docs](<https://develop.sentry.dev/sdk/expected-features/#dealing-with-network-failures>) for the full specification.

philipphofmann:
## :scroll: Description

Implements error logging for HTTP 413 (Content Too Large) responses to help users understand why their envelopes are being rejected. When the SDK receives an HTTP 413 response, it now logs an error message.

Fixes #7248

Users need visibility into why their envelopes are being rejected by the server. HTTP 413 responses indicate the envelope is too large, but without detailed logging, it's difficult to diagnose which events or attachments are causing the issue.

This follows the Sentry SDK specification: https://develop.sentry.dev/sdk/expected-features/#dealing-with-network-failures

Unit tests and tested with a sample app and capturing a message exceeding 1 MiB to double-check if Relay returns 413.

## :pencil: Checklist

---

### DIFF
--- a/Sources/Swift/Networking/SentryHttpStatusCode.swift
+++ b/Sources/Swift/Networking/SentryHttpStatusCode.swift
@@ -30,15 +30,13 @@ public enum SentryHttpStatusCode: Int {
 /// Enables direct comparison between Int and SentryHttpStatusCode.
 /// Allows writing `statusCode == SentryHttpStatusCode.contentTooLarge` instead of `statusCode == SentryHttpStatusCode.contentTooLarge.rawValue`.
 /// Note: This operator is only needed for Swift. In Objective-C, you can already compare the enum directly to an int without .rawValue.
-@_spi(Private)
-public func == (lhs: Int, rhs: SentryHttpStatusCode) -> Bool {
+func == (lhs: Int, rhs: SentryHttpStatusCode) -> Bool {
     return lhs == rhs.rawValue
 }
 
 /// Enables direct comparison between SentryHttpStatusCode and Int.
 /// Allows writing `SentryHttpStatusCode.contentTooLarge == statusCode` instead of `SentryHttpStatusCode.contentTooLarge.rawValue == statusCode`.
 /// Note: This operator is only needed for Swift. In Objective-C, you can already compare the enum directly to an int without .rawValue.
-@_spi(Private)
-public func == (lhs: SentryHttpStatusCode, rhs: Int) -> Bool {
+func == (lhs: SentryHttpStatusCode, rhs: Int) -> Bool {
     return lhs.rawValue == rhs
 }


Sources/Sentry/SentryHttpTransport.m
```diff
@@ -420,6 +420,14 @@ - (void)sendEnvelope:(SentryEnvelope *)envelope
 
             SENTRY_LOG_DEBUG(@"Received response status code: %li", (long)response.statusCode);
 
+            // Log specific error messages for certain HTTP status codes.
+            // See https://develop.sentry.dev/sdk/expected-features/#dealing-with-network-failures
+            // We didn't want to inline the logic in here, so we can have more Swift code.
+            [SentryHttpTransportHttpStatusCodeLogger
```
philprime, line 426: `l`: Instead of using a static method, could we create an instance of a logger which is injected/access via the SentryDependencyContainer? Using a protocol-based approach could unlock easier assertions in tests

Sources/Swift/Networking/SentryHttpTransportHttpStatusCodeLogger.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    /// Logs an error if the HTTP status code requires specific error messaging.
+    /// Currently handles HTTP 413 (Content Too Large) by logging envelope size and item types.
+    @objc
+    public static func logHttpResponseError(
+        statusCode: Int,
+        envelope: SentryEnvelope,
+        request: URLRequest
+    ) {
+        guard statusCode == 413 else {
```
philprime: `m`: We should use/introduce an enum for HTTP codes

Sources/Sentry/SentryHttpTransport.m
philippsuess: Oh, the whole SentrySDKLog is optimized for being used as a static class. I agree that we should change this, but not in this PR.

philippsuess:
@philprime, I added the HTTP status code enum. Please have another look 😃

Tests/SentryTests/Networking/SentryHttpStatusCodeTests.swift
```diff
@@ -0,0 +1,61 @@
+@_spi(Private) @testable import Sentry
+import XCTest
+
+final class SentryHttpStatusCodeTests: XCTestCase {
+
+    // MARK: - Int == SentryHttpStatusCode Tests
+
+    func testIntEqualsEnum_whenValuesMatch_shouldReturnTrue() {
+        XCTAssertTrue(413 == SentryHttpStatusCode.contentTooLarge)
```
philprime, line 9: `l`: Any strong reason you are not using `XCTAssertEqual`?

philippsuess: Yes, to test the ==. XCTAssertEqual needs the same type. XCTAssertEqual doesn't call ==.

PR #7276, opened 2026-01-23
ref: Removes dead code from DebugImageProvider

### LINKED ISSUE #7277: ref: Removes dead code from DeubgImageProvider
> [!NOTE]
> The pull request "[ref: Removes dead code from DeubgImageProvider](https://github.com/getsentry/sentry-cocoa/pull/7276)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Removes an unused function

#skip-changelog

noahsmartin:
Removes an unused function

#skip-changelog

Closes #7277


noahsmartin:
This is still used only by react native :(

philprime:
@noahsmartin this is one of the reasons I advocated back when we had the discussion on a modular project structure to have a SDK wrapper for the HybridSDK, which does not just re-export internal methods and types to the public, but instead defines a clean and clear public API which wraps internal types of our SDK.

That would allow us to refactor internally as much as we want, and as long as the HybridSDK API is not changing, we are fine.

PR #7294, opened 2026-01-27
fix(profiling): Remove launch profiling logs in sandboxed environments breaking CLI tools

### LINKED ISSUE #7281: Remove app launch profiling error for sandboxed execution
### Description

# Description

When executing a binary linked to the Sentry SDK in a sandbox environment, which is common for Homebrew code completion evaluation, our app launch profiling logs an error message when trying to create the `sentry.io` directory in the default cache directory.

```
[Sentry] [error] [1769418490.135415] [SentryFileManager:848] Can't create base path to store launch profile config file: Error Domain=SentryErrorDomain Code=108 "Failed to create the directory at path /Users/philip/Library/Caches/tart/io.sentry." UserInfo={NSLocalizedDescription=Failed to create the directory at path /Users/philip/Library/Caches/tart/io.sentry., NSUnderlyingError=0x104e8b970 {Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “io.sentry” in the folder “tart”." UserInfo={NSFilePath=/Users/philip/Library/Caches/tart/io.sentry, NSURL=file:///Users/philip/Library/Caches/tart/io.sentry, NSUnderlyingError=0x104e8b940 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}}}}
[Sentry] [error] [1769418490.135479] [SentryFileManager:895] Failed to construct the URL to check for launch profile configs.
Hello, World!

```

To reproduce the issue create a new SPM project with this `Package.swift`:

```swift
// swift-tools-version:5.10

import PackageDescription
let package = Package(
  name: "Tart",
  platforms: [
    .macOS(.v13)
  ],
  products: [
    .executable(name: "tart", targets: ["tart"])
  ],
  dependencies: [
    .package(url: "https://github.com/grpc/grpc-swift.git", .upToNextMajor(from: "1.24.2")),
    .package(url: "https://github.com/getsentry/sentry-cocoa", from: "8.51.1"),
  ],
  targets: [
    .executableTarget(name: "tart", dependencies: [
      .product(name: "GRPC", package: "grpc-swift"),
      .product(name: "Sentry", package: "sentry-cocoa"),
    ])
  ]
)
```

Create a single file at `Sources/tart/Root.swift`:

```swift
print("Hello, World!")
```

To create a sandboxed environment create a `sandbox.sb`:

```
(version 1)
(allow default)
(deny file* (subpath "/Users"))
```

And to execute it using `make clean` and `make`, create this `Makefile`:

```make
all:
	swift build
	sandbox-exec -f sandbox.sb -- .build/arm64-apple-macosx/debug/tart

clean:
	rm -rf .build
```

During initial investigation I was only able to reproduce the error if the dependency `https://github.com/grpc/grpc-swift.git` was present too. When having `sentry-cocoa` as the only dependency the error message **was not shown**, maybe due to the absence of the Objective-C runtime triggering the `+ (void) load;` method, starting app launch profiling.

Furthermore, this reproduce **does not** even contain a `SentrySDK.configure` block, meaning that setting the error log level will not work.

# Other

- Relevant Issue: https://github.com/cirruslabs/tart/issues/1175
- Reproducer Project: https://github.com/edigaryev/tart-sentry-reproduce

philprime:
## Problem

When using the Sentry SDK in a sandboxed macOS environment (e.g., CLI tools with code completion), launch profiling code runs in `+load` before any SDK configuration and attempts to create directories and read config files. In sandboxed environments where `/Users` is restricted, this produces `SENTRY_LOG_ERROR` messages that cannot be suppressed by the user — since `+load` runs before `SentrySDK.start`, there is no way to configure log levels.

Additionally, the `DEBUG` preprocessor guard in `SentryProfiler +load` caused inconsistent behavior: SPM consumers saw launch profiling in release builds but not debug builds, while binary (XCFramework) consumers always had it enabled.

Closes #7281

### 1. Remove DEBUG guard from launch profiling
Remove `defined(DEBUG)` from the `+load` compile-time guard in `SentryProfiler.mm`, keeping only `SENTRY_TEST || SENTRY_TEST_CI`. This ensures launch profiling behaves consistently for all consumer builds (debug and release), matching the XCFramework behavior.

### 2. Separate read and write paths in SentryFileManagerHelper
- Extract `ensureLaunchProfileConfigDirectoryExists()` from `launchProfileConfigFileURL()` — directory creation now only happens in the write path
- `launchProfileConfigFileURL()` becomes a pure URL builder with no side effects (no directory creation, no error logs)
- Move directory creation to `writeAppLaunchProfilingConfigFile()` where it belongs

### 3. Downgrade read-path log levels
Change `SENTRY_LOG_ERROR` to `SENTRY_LOG_DEBUG` in `sentry_persistedLaunchProfileConfigurationOptions()` and `appLaunchProfileConfigFileExists()` when URL/path construction fails. Missing config is a normal state (first launch, sandboxed environment, profiling not configured).

### 4. Fix pre-existing macOS test failures
`SentryFileIOTrackingIntegrationTests` and `SentryFileIOTrackingIntegrationObjCTests` used `NSDocumentDirectory` (`~/Documents`), which is TCC-protected on macOS. Switched to temporary directory.

## How did you test it?

- `make test-macos` — all tests pass with zero failures
- Added 4 new tests for `SentryFileManagerHelper` profiling functions:
  - `testEnsureLaunchProfileConfigDirectoryExists_whenBasePathAvailable_shouldReturnTrue`
  - `testPersistedLaunchProfileConfigOptions_whenURLIsNil_shouldReturnNil`
  - `testPersistedLaunchProfileConfigOptions_whenNoConfigFileExists_shouldReturnNil`
  - `testWriteAppLaunchProfilingConfigFile_whenDirectoryDoesNotExist_shouldCreateItAndWrite`
- Reproduced sandbox issue using `Samples/macOS-CLI` with `sandbox-exec`
- Verified fix eliminates error logs in sandboxed execution

## Reproduction sample

The `Samples/macOS-CLI` directory contains two executables for reproducing the issue:
- `cli-with-binary`: Uses pre-built XCFramework
- `cli-with-spm`: Uses source-built SPM target

Run with: `make exec-sandboxed-cli-with-binary` or `make exec-sandboxed-cli-with-spm`

### DIFF
--- a/Tests/SentryTests/Helper/SentryFileManagerTests.swift
+++ b/Tests/SentryTests/Helper/SentryFileManagerTests.swift
@@ -1451,8 +1451,9 @@ extension SentryFileManagerTests {
 private extension SentryFileManagerTests {
     func ensureAppLaunchProfileConfig(exists: Bool = true, tracesSampleRate: Double = 1, tracesSampleRand: Double = 1.0, profilesSampleRate: Double = 1, profilesSampleRand: Double = 1.0) throws {
         let url = try XCTUnwrap(launchProfileConfigFileURL())
-        
+
         if exists {
+            XCTAssertTrue(ensureLaunchProfileConfigDirectoryExists(), "Directory for launch profile config must exist before writing")
             let dict = [
                 kSentryLaunchProfileConfigKeyTracesSampleRate: tracesSampleRate,
                 kSentryLaunchProfileConfigKeyTracesSampleRand: tracesSampleRand,


philprime:
Write-up summary made with Cursor from local analysis.

---

# Issue #7281: Launch Profiling Error in Sandboxed Environments

Raw findings from investigating [getsentry/sentry-cocoa#7281](https://github.com/getsentry/sentry-cocoa/issues/7281).
Related upstream issue: [cirruslabs/tart#1175](https://github.com/cirruslabs/tart/issues/1175).

## Problem

When a macOS CLI binary linked against the Sentry SDK runs inside a `sandbox-exec` sandbox that denies file access under `/Users`, the SDK prints error-level log messages to stderr during `+load` — before `main()` and before any SDK configuration.

```
[Sentry] [error] [...] [SentryFileManagerHelper:799] Can't create base path to store launch profile config file: ...
[Sentry] [error] [...] [SentryFileManagerHelper:846] Failed to construct the URL to check for launch profile configs.
```

This is disruptive because:

- It pollutes shell output (e.g. during Homebrew completion script evaluation).
- The user has no way to suppress it — `SentryOptions.diagnosticLevel` is not yet configured.
- The app does not even call `SentrySDK.start`.

## Reproduction

Sample app: `Samples/macOS-CLI/`

```bash
cd Samples/macOS-CLI
make exec-sandboxed-cli-with-binary   # Shows the error
make exec-sandboxed-cli-with-spm      # No error (debug build, see below)
```

Building SPM in release also triggers it:

```bash
swift build --product cli-with-spm -c release
sandbox-exec -f sandbox.sb -- .build/arm64-apple-macosx/release/cli-with-spm  # Shows the error
```

## Root Cause Chain

### 1. `+[SentryProfiler load]` runs unconditionally in release builds

`SentryProfiler.mm` line 69-88:

```objc
+ (void)load
{
#if defined(SENTRY_TEST) || defined(SENTRY_TEST_CI) || defined(DEBUG)
    if (NSProcessInfo.processInfo.environment[@"--io.sentry.ui-test.test-name"] == nil) {
        return;
    }
#endif
    sentry_startLaunchProfile();
}
```

The pre-built XCFramework (`Sentry` product) is compiled without `DEBUG`, so `sentry_startLaunchProfile()` always executes. The SPM source build (`SentrySPM`) only skips it in debug mode because of the early return guard.

### 2. `sentry_startLaunchProfile()` checks for config file existence

`SentryLaunchProfiling.m` line 237-242:

```objc
void _sentry_nondeduplicated_startLaunchProfile(void) {
    if (!appLaunchProfileConfigFileExists()) {
        SENTRY_LOG_DEBUG(@"No launch profile config exists, will not profile launch.");
        return;
    }
    // ...
}
```

### 3. `appLaunchProfileConfigFileExists()` calls `launchProfileConfigFileURL()`

`SentryFileManagerHelper.m` line 841-856:

```objc
BOOL appLaunchProfileConfigFileExists(void) {
    NSURL *url = launchProfileConfigFileURL();
    if (url == nil) {
        SENTRY_LOG_ERROR(@"Failed to construct the URL to check for launch profile configs.");
        return NO;
    }
    // ...
}
```

### 4. `launchProfileConfigFileURL()` tries to create the base directory

`SentryFileManagerHelper.m` line 787-807:

```objc
NSURL *launchProfileConfigFileURL(void) {
    dispatch_once(&onceToken, ^{
        NSString *basePath = sentryStaticBasePath();
        // ...
        if (!createDirectoryIfNotExists(basePath, &error)) {
            SENTRY_LOG_ERROR(@"Can't create base path to store launch profile config file: %@", error);
            return;
        }
        // ...
    });
    // ...
}
```

### 5. The base path resolves to a `/Users/...` path

`sentryStaticBasePath()` -> `sentryStaticCachesPath()` -> `sentryGetScopedCachesDirectory()`

`SentryFileManagerHelper.m` line 678-710:

```objc
NSString *sentryGetScopedCachesDirectory(NSString *cachesDirectory) {
#if !TARGET_OS_OSX
    return cachesDirectory;  // iOS always sandboxed
#else
    BOOL isSandboxed = processInfo.environment[@"APP_SANDBOX_CONTAINER_ID"] != nil;
    // ...
    return sentryBuildScopedCachesDirectoryPath(cachesDirectory, isSandboxed, bundleIdentifier, lastPathComponent);
#endif
}
```

The `sandbox-exec` sandbox is NOT macOS App Sandbox, so `APP_SANDBOX_CONTAINER_ID` is not set. The code treats the app as unsandboxed and builds a path like `/Users/philip/Library/Caches/cli-with-binary/io.sentry`, which the sandbox denies.

### 6. Error messages are logged at ERROR level

Both `SENTRY_LOG_ERROR` calls in steps 4 and 5 write to stderr. Since these run in `+load` (before `main()`), there is no opportunity for the user to configure logging.

## Why SPM debug builds didn't show the error

The `#if defined(DEBUG)` guard in `SentryProfiler +load` causes an early return when the `--io.sentry.ui-test.test-name` environment variable is absent. This is the normal case for all non-UI-test runs. So the entire `sentry_startLaunchProfile()` code path is never reached in debug SPM builds.

This was confirmed by:

1. Disassembling the `+load` in both binaries (release = single `b _sentry_startLaunchProfile` instruction; debug = full env var check with early return)
2. Building `cli-with-spm` in release mode (`swift build -c release`) and confirming the error appears

## Secondary Finding: DEBUG guard disables launch profiling for all SPM debug builds

While investigating the sandbox issue, we discovered that the `#if defined(DEBUG)` guard in `SentryProfiler +load` has a broader effect than intended. The `SENTRY_TEST` and `SENTRY_TEST_CI` macros are Sentry-internal and correctly scoped, but `DEBUG` is defined by SPM for **all** targets (including dependencies) when building in debug configuration. This means launch profiling is silently disabled for every consumer app that uses `SentrySPM` in a debug scheme — not just the SDK's own test suite.

This is a separate issue from the sandbox error and should be evaluated independently. See `develop-docs/OBJC-LOAD-AND-LINKING.md` for details.

## Findings Summary

The core issue has three layers:

**Layer 1: `+load` runs launch profiling unconditionally in production builds.**
There is no check for whether the SDK has been configured or even intended to be used. The `+load` runs file system operations just because the binary links the Sentry framework.

**Layer 2: File system operations in `+load` fail and log errors that users can't suppress.**
The `launchProfileConfigFileURL()` function eagerly creates directories via `createDirectoryIfNotExists()`. When this fails (due to sandbox, permissions, or any other reason), it logs at ERROR level. Since this happens pre-`main()`, there is no way to set `SentryOptions.diagnosticLevel`.

**Layer 3: The DEBUG guard masks the issue in SPM debug builds.**
The `defined(DEBUG)` condition in the `+load` guard disables launch profiling for all SPM source-built debug builds, not just the SDK's test suite. This masks the sandbox error during development (where most testing happens) and defers it to release builds. It also means SPM users never get launch profiling during development, which is likely unintended.

## Possible Fix Directions

These are rough ideas collected during investigation, not yet evaluated for feasibility or side effects.

### Option A: Don't create directories in `launchProfileConfigFileURL()`

Currently `launchProfileConfigFileURL()` calls `createDirectoryIfNotExists(basePath, &error)`. This eagerly creates the `io.sentry` directory even when just checking if a config file exists. The config file won't exist on a fresh install anyway, so checking for file existence should not require creating directories.

Split the function into "read" (check existence) and "write" (create directory + write file) paths. The read path would just build the URL and check `access()` without creating anything. The write path would create directories only when actually persisting config.

### Option B: Downgrade the log level

Change `SENTRY_LOG_ERROR` to `SENTRY_LOG_WARN` or `SENTRY_LOG_DEBUG` for the directory creation failure in `launchProfileConfigFileURL()`. This would suppress the messages at default log levels but doesn't fix the underlying issue of running file system operations unnecessarily.

### Option C: Guard `sentry_startLaunchProfile()` to bail out early if config can't exist

If we can determine early that the environment won't support file operations (e.g. no writable caches directory), bail out before attempting directory creation.

### Option D: Check file existence without directory creation

Instead of building the full URL path (which involves `createDirectoryIfNotExists`), use a simpler check that just tests whether the config file path could exist. If the parent directory doesn't exist, the config file can't exist either, so return NO immediately.

## Files Involved

| File                                                  | Role                                                                                                                                                           |
| ----------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `Sources/Sentry/SentryProfiler.mm`                    | `+load` entry point calling `sentry_startLaunchProfile()`                                                                                                      |
| `Sources/Sentry/Profiling/SentryLaunchProfiling.m`    | `sentry_startLaunchProfile()` and `_sentry_nondeduplicated_startLaunchProfile()`                                                                               |
| `Sources/Sentry/SentryFileManagerHelper.m`            | `sentryGetScopedCachesDirectory()`, `sentryStaticCachesPath()`, `sentryStaticBasePath()`, `launchProfileConfigFileURL()`, `appLaunchProfileConfigFileExists()` |
| `Sources/Sentry/Public/SentryProfilingConditionals.h` | `SENTRY_TARGET_PROFILING_SUPPORTED` definition                                                                                                                 |
| `Samples/macOS-CLI/`                                  | Reproduction sample with sandbox setup                                                                                                                         |

philprime:
I removed the `DEBUG` compiler flag check from the launch profiler so it's not skipped for debug builds

philprime:
## Update: Fix implemented and tests passing

### Changes in latest commits:

**1. Remove DEBUG guard from launch profiling (`SentryProfiler.mm`)**
The `#if defined(DEBUG)` was removed from the `+load` guard, leaving only `SENTRY_TEST || SENTRY_TEST_CI`. This ensures launch profiling behaves consistently for SPM consumers across debug and release builds. See `develop-docs/OBJC-LOAD-AND-LINKING.md` for full analysis.

**2. Fix sandbox logging issue (`SentryFileManagerHelper.m`)**
- Extracted `ensureLaunchProfileConfigDirectoryExists()` from `launchProfileConfigFileURL()` — directory creation now only happens in the write path (`writeAppLaunchProfilingConfigFile`)
- `launchProfileConfigFileURL()` is now a pure URL builder with no side effects
- Downgraded log levels from `SENTRY_LOG_ERROR` to `SENTRY_LOG_DEBUG` in read-path functions (`sentry_persistedLaunchProfileConfigurationOptions`, `appLaunchProfileConfigFileExists`) since missing config is a normal state

**3. Fix pre-existing macOS test failures**
- `SentryFileIOTrackingIntegrationTests` and `SentryFileIOTrackingIntegrationObjCTests` were using `NSDocumentDirectory` (`~/Documents`), which is TCC-protected on macOS
- Switched to `NSTemporaryDirectory()` / `FileManager.default.temporaryDirectory`

**4. Added test coverage**
- `testEnsureLaunchProfileConfigDirectoryExists_whenBasePathAvailable_shouldReturnTrue`
- `testPersistedLaunchProfileConfigOptions_whenURLIsNil_shouldReturnNil`
- `testPersistedLaunchProfileConfigOptions_whenNoConfigFileExists_shouldReturnNil`
- `testWriteAppLaunchProfilingConfigFile_whenDirectoryDoesNotExist_shouldCreateItAndWrite`

### Test results:
- `make test-macos` ✅ (all tests pass, zero failures)

philprime:
@sentry review

Sources/Sentry/SentryFileManagerHelper.m
```diff
@@ -858,6 +868,9 @@ - (BOOL)isDirectory:(NSString *)path
 void
 writeAppLaunchProfilingConfigFile(NSMutableDictionary<NSString *, NSNumber *> *config)
 {
+    if (!ensureLaunchProfileConfigDirectoryExists()) {
+        return;
+    }
```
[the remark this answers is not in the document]
philprime: False positive — the `dispatch_once` caching pattern is pre-existing and unchanged by this PR. `sentryStaticBasePath()` depends on `NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES)`, which is deterministic for a given process. It won't be temporarily nil and then recover later; if it's nil (e.g. sandboxed), it's consistently nil for the process lifetime, so caching it is correct.

Tests/SentryTests/Integrations/Performance/IO/SentryFileIOTrackingIntegrationTests.swift
```diff
@@ -24,8 +24,7 @@ class SentryFileIOTrackingIntegrationTests: XCTestCase {
         }
         
         init() throws {
-            let paths = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
-            fileDirectory = try XCTUnwrap(paths.first)
```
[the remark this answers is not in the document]
philprime: Fixed — each test class now uses a unique temp directory name (`SentryFileIOTrackingIntegrationTests` vs `SentryFileIOTrackingIntegrationObjCTests`) to eliminate any potential interference.

Sources/Sentry/SentryFileManagerHelper.m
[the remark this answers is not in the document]
philprime: This is a false positive since the `dispatch_once` caching pattern is pre-existing and `sentryStaticBasePath()` returns a deterministic value per process.

.swiftlint.yml
```diff
@@ -1,6 +1,7 @@
 excluded:
   - Pods
   - Utils/**/.build
+  - Samples/**/.build
```
philippsuess, line 4: That makes sense, thanks.

Samples/macOS-CLI/.gitignore
```diff
@@ -0,0 +1,8 @@
+.DS_Store
+/.build
+/Packages
+xcuserdata/
+DerivedData/
+.swiftpm/configuration/registries.json
+.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+.netrc
```
philippsuess: `l`: I wonder if we can define these rules in the top-level git ignore, because we could use file patterns to basically ignore all build folders, derived data folders, even if they're subfolders, for example.

Samples/macOS-CLI/Package.swift
```diff
@@ -0,0 +1,28 @@
...
+    name: "macOS-CLI",
+    platforms: [
+        .macOS(.v13)
+    ],
+    products: [
+        .executable(name: "cli-with-binary", targets: ["cli-with-binary"]),
+        .executable(name: "cli-with-spm", targets: ["cli-with-spm"])
+    ],
+    dependencies: [
+        .package(url: "https://github.com/grpc/grpc-swift.git", .upToNextMajor(from: "1.24.2")),
```
philippsuess: `m`: I’m wondering if it would be better to pin this to a specific version and only update it intentionally. That way, we don’t accidentally pick up changes on their side that could break us.

Sources/Sentry/SentryFileManagerHelper.m
```diff
@@ -858,6 +868,9 @@ - (BOOL)isDirectory:(NSString *)path
 void
 writeAppLaunchProfilingConfigFile(NSMutableDictionary<NSString *, NSNumber *> *config)
 {
+    if (!ensureLaunchProfileConfigDirectoryExists()) {
+        return;
```
philippsuess, line 873: `m`: Maybe worth printing a debug log here.

develop-docs/OBJC-LOAD-AND-LINKING.md
```diff
@@ -0,0 +1,135 @@
...
+The SDK uses `+load` in several classes for early initialization:
+
+| Class                                     | Purpose                                                       |
+| ----------------------------------------- | ------------------------------------------------------------- |
+| `SentryProfiler`                          | Starts app launch profiling via `sentry_startLaunchProfile()` |
+| `SentryAppStartTrackerHelper`             | Captures runtime init timestamp and checks prewarm status     |
+| `SentrySysctlObjC`                        | Captures runtime init timestamp for launch profiles           |
+| `SentryCrashDefaultMachineContextWrapper` | Captures main thread ID                                       |
+
+Because `+load` runs before any SDK configuration, there is no way for the user to control or suppress it via `SentryOptions` or log level settings.
```
philippsuess, line 22: Just as a side note: this is another strong argument for splitting the SDK into multiple modules. I’m generally not a fan of always initializing the Sentry profiler code, even when profiling isn’t being used, since it still adds some overhead.

Definitely not something to address in this PR — just pointing it out. And thanks a lot for adding this document.

Samples/macOS-CLI/Package.swift
philprime: Good point, I will pin it so it stays reproducible.

PR #7324, opened 2026-01-30
refactor: Update SentryAttribute to use a generic array type

### LINKED ISSUE #7341: refactor: Update SentryAttribute to use a generic array type
> [!NOTE]
> The pull request "[refactor: Update SentryAttribute to use a generic array type](https://github.com/getsentry/sentry-cocoa/pull/7324)" was created by @itaybre but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

## :scroll: Description

Replaces the array attributes type with a generic `array`

After reviewing the [developer docs](https://develop.sentry.dev/sdk/telemetry/attributes/), I found out that array attributes now expect `array` as the type.
Related:
- Docs https://github.com/getsentry/sentry-docs/pull/15994
- Change in Sentry Native https://github.com/getsentry/sentry-native/pull/1479
- Relay https://github.com/getsentry/relay/pull/5394

For now only unit tests, array attributes are not visible in Sentry UI so only unit tests

## :pencil: Checklist

itaybre:
## :scroll: Description

Replaces the array attributes type with a generic `array`

After reviewing the [developer docs](https://develop.sentry.dev/sdk/telemetry/attributes/), I found out that array attributes now expect `array` as the type.
Related:
- Docs https://github.com/getsentry/sentry-docs/pull/15994
- Change in Sentry Native https://github.com/getsentry/sentry-native/pull/1479
- Relay https://github.com/getsentry/relay/pull/5394

For now only unit tests, array attributes are not visible in Sentry UI so only unit tests

## :pencil: Checklist

Closes #7341

### DIFF
--- a/Sources/Swift/Protocol/SentryAttributeContent.swift
+++ b/Sources/Swift/Protocol/SentryAttributeContent.swift
@@ -8,7 +8,6 @@ internal enum SentryAttributeType: String {
     case booleanArray = "boolean[]"
     case integerArray = "integer[]"
     case doubleArray = "double[]"
-    case array = "array"
 }
 
 /// A type-safe representation of attribute values used by structured logging.
@@ -49,6 +48,10 @@ public enum SentryAttributeContent: Equatable, Hashable {
             return SentryAttributeType.doubleArray.rawValue
         }
     }
+    
+    // This type is used to encode array types as "array" regardless of element type.
+    // This is what Relay expects for array types: https://develop.sentry.dev/sdk/telemetry/attributes/
+    static let genericArrayType: String = "array"
 }
 
 extension SentryAttributeContent: Encodable {
@@ -67,7 +70,7 @@ extension SentryAttributeContent: Encodable {
     public func encode(to encoder: any Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
         
-        try container.encode(typeForContent(), forKey: .type)
+        try container.encode(typeForEncoding, forKey: .type)
 
         switch self {
         case .string(let value):
@@ -89,12 +92,12 @@ extension SentryAttributeContent: Encodable {
         }
     }
     
-    private func typeForContent() -> String {
+    private var typeForEncoding: String {
         switch self {
         case .string(_), .boolean(_), .integer(_), .double(_):
             return type
         case .stringArray(_), .booleanArray(_), .integerArray(_), .doubleArray(_):
-            return SentryAttributeType.array.rawValue
+            return SentryAttributeContent.genericArrayType
         }
     }
 }


philprime (commented):
What's still unclear to me now is what element types we should support for the arrays. The develop docs and internal comms mention that mixed arrays might become available in the future at some point. For now they are homogenous, but do we have to indicate which element type it is? Do we just convert everything to string? Depending on that we might also want to adapt the entire public API of Metrics too

philprime:
Relevant docs PR to clarify specs: https://github.com/getsentry/sentry-docs/pull/16242

itaybre:
> What's still unclear to me now is what element types we should support for the arrays. The develop docs and internal comms mention that mixed arrays might become available in the future at some point. For now they are homogenous, but do we have to indicate which element type it is? Do we just convert everything to string? Depending on that we might also want to adapt the entire public API of Metrics too

They might become available doesn't seems like a good enough reason to undo all Metrics API, `eventually` might be in a couple months.
When the time comes, users would be already comfortable using the current API and the new mixed arrays will just give them more uses for the API.

Tests/SentryTests/SentryLoggerTests.swift
```diff
@@ -547,22 +547,20 @@ final class SentryLoggerTests: XCTestCase {
...
+                } else if let expectedValue = expectedAttribute.value as? [Bool] {
+                    let actualValue = actualAttribute.value as! [Bool]
+                    XCTAssertEqual(actualValue, expectedValue, "Boolean array attribute value mismatch for key: \(key)", file: file, line: line)
+                } else if let expectedValue = expectedAttribute.value as? [Int] {
+                    let actualValue = actualAttribute.value as! [Int]
+                    XCTAssertEqual(actualValue, expectedValue, "Integer array attribute value mismatch for key: \(key)", file: file, line: line)
+                } else if let expectedValue = expectedAttribute.value as? [Double] {
+                    let actualValue = actualAttribute.value as! [Double]
+                    XCTAssertEqual(actualValue, expectedValue, "Double array attribute value mismatch for key: \(key)", file: file, line: line)
+                }
```
philprime:
```suggestion
                if let expectedValue = expectedAttribute.value as? [String] {
                    let actualValue = try XCTUnwrap(actualAttribute.value as? [String])
                    XCTAssertEqual(actualValue, expectedValue, "String array attribute value mismatch for key: \(key)", file: file, line: line)
                } else if let expectedValue = expectedAttribute.value as? [Bool] {
                    let actualValue = try XCTUnwrap(actualAttribute.value as? [Bool])
                    XCTAssertEqual(actualValue, expectedValue, "Boolean array attribute value mismatch for key: \(key)", file: file, line: line)
                } else if let expectedValue = expectedAttribute.value as? [Int] {
                    let actualValue = try XCTUnwrap(actualAttribute.value as? [Int])
                    XCTAssertEqual(actualValue, expectedValue, "Integer array attribute value mismatch for key: \(key)", file: file, line: line)
                } else if let expectedValue = expectedAttribute.value as? [Double] {
                    let actualValue = try XCTUnwrap(actualAttribute.value as? [Double])
                    XCTAssertEqual(actualValue, expectedValue, "Double array attribute value mismatch for key: \(key)", file: file, line: line)
                }
```

Sources/Swift/Protocol/SentryAttributeContent.swift
```diff
@@ -39,13 +36,13 @@ public enum SentryAttributeContent: Equatable, Hashable {
         case .double:
             return SentryAttributeType.double.rawValue
         case .stringArray:
-            return SentryAttributeType.stringArray.rawValue
+            return SentryAttributeType.array.rawValue
```
philprime: `h`: This `type` is used to convert from and to the enum. I don't think. we should change them all to be `array` here, but instead we should keep using our typed arrays as much as possible. Instead would like to propose that we change it on a lower encoding level, that only when encoding to JSON the type is set to `array.

The problem I see is that if all enum type use the same value for `type`, it's an too easy mistake of using this property in cases unrelated to encoding, which brings up the question why we should even keep the typed array enum cases, i.e. `stringArray` around and not just type-erase to `Any`

philippsuess:
> I still think it's beneficial to offer our SDK users type-safety for homogenous arrays, e.g. [Int], and only on the lowest level where we do the encoding set it to array to be compatible with the protocol

I agree. It's quite uncommon to do this in Swift or any other strictly typed language: 

```Swift
// Which Swift dev does this?
let array: [Any] = [1, "hi", 3.14]
```

Doing mixed typed arrays is more common in JS and Python, but not in Swift.

Even if the protocol allows it, our API here should be optimized for Swift. I would keep the type safe approach. We can allow mixed typed arrays, IMO, but that shouldn't be the default, and I would be OK with using some extra API that is not so nice to use.

Sources/Swift/Protocol/SentryAttributeContent.swift
itaybre: Updated the code so we only use array when encoding

Sources/Swift/Protocol/SentryAttributeContent.swift
```diff
@@ -85,6 +88,15 @@ extension SentryAttributeContent: Encodable {
...
     }
+    
+    private func typeForContent() -> String {
+        switch self {
+        case .string(_), .boolean(_), .integer(_), .double(_):
+            return type
+        case .stringArray(_), .booleanArray(_), .integerArray(_), .doubleArray(_):
+            return SentryAttributeType.array.rawValue
+        }
+    }
```
philippsuess: `m`: Shouldn't we rather remove the SentryAttributeType for typed arrays and only keep array? We only use these internally, so having the typed arrays still internally doesn't really make sense to me. WDYT @philprime?

itaybre: I might be answering for @philprime, but this internal type is used for casting between `SentryAttributeContent` and ``SentryAttribute` without needed to check the array's content

Maybe I should rename this function to `typeForEncoding`

philprime: Yes, `typeForEncoding` makes more sense, and you could also define it as a getter-variable instead of a method.

And yes, @itaybre is correct, we need `SentryAttributeType` in other parts of the SDK

philippsuess: Do we also need `SentryAttributeType.array` in other parts of the SDK?

If not, we could remove it, so we would hide the serialization implementation detail from the rest of the SDK. I'm asking because it would be great if we could keep the detail of not being able to send typed arrays a serialization detail. Ideally, the rest of the SDK has no clue that we can only send `array` even though we use types arrays everywhere else. But if we also need `SentryAttributeType.array`  in other parts of the SDK, then lets keep it the `SentryAttributeType.array` .

itaybre: Renamed it and removed the array enum case.

philprime:
@itaybre what is missing for getting this merged?

itaybre:
> @itaybre what is missing for getting this merged?

I never saw it was approved, merging it

PR #7337, opened 2026-02-02
test: add tests for custom type array conversion

### LINKED ISSUE #7338: test: add tests for custom type array conversion
> [!NOTE]
> The pull request "[test: add tests for custom type array conversion](https://github.com/getsentry/sentry-cocoa/pull/7337)" was created by @philprime but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Add tests to verify that arrays containing different custom types adopting SentryAttributeValue are correctly converted to typed arrays (integer[], double[], string[]) when all elements return the same SentryAttributeContent case, rather than falling back to string[].

#skip-changelog

philprime:
Add tests to verify that arrays containing different custom types adopting SentryAttributeValue are correctly converted to typed arrays (integer[], double[], string[]) when all elements return the same SentryAttributeContent case, rather than falling back to string[].

#skip-changelog

Closes #7338

### DIFF
--- a/Tests/SentryTests/Protocol/SentryAttributeValuableTests.swift
+++ b/Tests/SentryTests/Protocol/SentryAttributeValuableTests.swift
@@ -368,6 +368,125 @@ final class SentryAttributeValueTests: XCTestCase {
         XCTAssertEqual(result, .booleanArray([true, true]))
     }
     
+    func testAsSentryAttributeContent_whenHomogenousIntegerArrayWithSingleCustomType_shouldReturnIntegerArrayCase() {
+        // -- Arrange --
+        struct CustomType: SentryAttributeValue {
+            var asSentryAttributeContent: SentryAttributeContent {
+                return .integer(42)
+            }
+        }
+        
+        let array: [CustomType] = [CustomType(), CustomType()]
+        
+        // -- Act --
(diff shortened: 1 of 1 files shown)


Tests/SentryTests/Protocol/SentryAttributeValuableTests.swift
```diff
@@ -368,6 +368,125 @@ final class SentryAttributeValueTests: XCTestCase {
...
+        
+        let array: [CustomType] = [CustomType(), CustomType()]
+        
+        // -- Act --
+        let result = array.asSentryAttributeContent
+        
+        // -- Assert --
+        guard case .integerArray(let value) = result else {
+            return XCTFail("Expected .integerArray case")
+        }
```
philippsuess, line 387: `l`: I prefer using casting and using XCTUnwrap over this because it usually fits on one line.

philipphofmann (approved):
LGTM

Tests/SentryTests/Protocol/SentryAttributeValuableTests.swift
philprime: Not applicable for guard-case statements using associated values

PR #7339, opened 2026-02-02
feat: Expose metrics API to Objective-C   

tustanivsky:
## :scroll: Description

This PR introduces `SentryMetrics` wrapper class that bridges the Swift-only metrics API (count, distribution, gauge) to Objective-C.

This pattern was used because there is no way to make the underlying types `@objc`-compatible: 
- `SentryUnit` is a Swift enum with associated values
- `SentryAttributeValue` is a Swift protocol
- `SentryMetricsApi` is a generic struct which also cannot be `@objc

Even though the static type of `SentrySDK.metrics` changed from `SentryMetricsApiProtocol` to `SentryMetrics`, read access remains fully backward-compatible because `SentryMetrics` conforms to `SentryMetricsApiProtocol`, so existing Swift call sites and variables typed as the protocol continue to work.

The actual breaks would only be:                                                                                                                                                                                                         
- Assigning a custom conformer to the property: `SentrySDK.metrics = MyCustomImpl()`
- Explicitly using the old concrete type: `let m: SentryMetricsApi<...> = SentrySDK.metrics`

However, neither of these look like supported use-cases.

The Unreal Engine plugin uses `sentry-cocoa` internally via Objective-C which previously had no access to the metrics API. This change enables metrics calls from any Objective-C context by accepting `String` for units and `[String: Any]` for attributes converting them to the Swift-native types internally. 

Related items:
- https://github.com/getsentry/sentry-unreal/issues/1213
- https://github.com/getsentry/sentry-unreal/pull/1214

Manually emitted counter, gauge, and distribution metrics in Unreal Engine using a patched version of the Cocoa SDK in the plugin.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Expose metrics API to Objective-C ([#7339](https://github.com/getsentry/sentry-cocoa/pull/7339))
+
 ### Fixes
 
 - Fix mismatch of `in_foreground` app context (#7188) The app context `in_foreground` for handled and unhandled events was sometimes different. This is fixed now by aligning the implementation and adding a new `is_active` app context field.
--- a/Sources/Swift/Helper/SentrySDK.swift
+++ b/Sources/Swift/Helper/SentrySDK.swift
@@ -80,12 +80,8 @@ import Foundation
     ///
     /// - Note: This feature is currently in open beta.
     ///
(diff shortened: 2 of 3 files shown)


philprime (commented):
I am going to raise an internal discussion on this, because there might have been a miscommunication, so please hold back from further adapting this PR for now

tustanivsky:
Closing this for now as the Metrics API is intended to remain a Swift-only interface.

PR #7340, opened 2026-02-02
fix: Write reports on concurrent crashes

### LINKED ISSUE #3296: Sentry crash in multiple simultaneous fatal crashes in structured concurrency
### Platform

macOS

### Installed

Swift Package Manager

### Version

8.10.0

### Steps to Reproduce

1.  Create a code that triggers multiple simultaneous Tasks, each one calling `fatalError()`. Ensure that they are not restricted to run in the MainActor, so they can run concurrently. For example:

```
        AsyncStream { continuation in
            Task {
                await withTaskGroup(of: Void.self) { group in
                    for key in [“key1”, “key2”, “key3”, “key4”, “key5”, “key6”] {
                        group.addTask {
                            fatalError("Testing")
                            continuation.yield(key)
                        }
                    }
                }
                continuation.finish()
            }
        }
```

2. When the code above is executed, the app crashes, but Sentry does NOT generate any report. When Sentry debug is enabled, this is what is printed:

```
ImageCache/ImageCache.swift:56: Fatal error: Testing
ImageCache/ImageCache.swift:56: Fatal error: Testing
ImageCache/ImageCache.swift:56: Fatal error: Testing
ImageCache/ImageCache.swift:56: Fatal error: Testing
ImageCache/ImageCache.swift:56: Fatal error: Testing
ImageCache/ImageCache.swift:56: Fatal error: Testing
ERROR: SentryCrashReport.c (1487): void sentrycrashreport_writeRecrashReport(const SentryCrash_MonitorContext *const, const char *const): Could not rename  to : No such file or directory
ERROR: SentryCrashFileUtils.c (414): _Bool sentrycrashfu_openBufferedWriter(SentryCrashBufferedWriter *, const char *const, char *, int): Could not open crash report file : No such file or directory
Trace/BPT trap: 5
```

If the code above is restricted to run on a single actor, for example by moving the addTask closure to a function marked as @MainActor, they will run serially, the app will crash once and stop, and Sentry will have a chance to save the crash report to disk successfully.

My naive understanding is that Sentry’s file operations should only execute in the main thread or at least be better shielded for concurrent executions.

### Expected Result

Sentry code crashes and no crash report is stored in disk:

```
ERROR: SentryCrashReport.c (1487): void sentrycrashreport_writeRecrashReport(const SentryCrash_MonitorContext *const, const char *const): Could not rename  to : No such file or directory
ERROR: SentryCrashFileUtils.c (414): _Bool sentrycrashfu_openBufferedWriter(SentryCrashBufferedWriter *, const char *const, char *, int): Could not open crash report file : No such file or directory
Trace/BPT trap: 5
```

### Actual Result

App crashes in the first `fatalError` and Sentry saves the crash report to disk. 

If given some guidance on how to fix it, yes.

denrase:
## :scroll: Description

- Track which thread is handling the crash via g_crashingThread
- Same thread re-entering = recrash (allowed)
- Different thread = block for 2s, then continue
- Use C11 atomics (<stdatomic.h>)
- Rename `sentrycrashcm_notifyFatalExceptionCaptured` to `sentrycrashcm_notifyFatalException` to move it closer to KSCrash naming.
- Merged notify + suspend into `sentrycrashcm_notifyFatalException` to prevent deadlock (concurrency check must run before any threads are suspended)

This eliminates the ENOENT race by ensuring only one thread ever enters crash handling at a time.

Aligned with KSCrash upstream ([KSCrashMonitor.c](https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashRecordingCore/KSCrashMonitor.c)):

When multiple threads crash simultaneously, the second crash was incorrectly treated as a "recrash" (crash during crash handling) because g_handlingFatalException had no thread awareness.

Fixes #3296

## :pencil: Checklist

### DIFF
--- a/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
+++ b/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
@@ -220,6 +220,13 @@ sentrycrashcm_notifyFatalException(
     // - Recrash handling means build a minimal recrash report and be very cautious.
     // - Block means block this thread for a few seconds so it doesn't return
     //   before the other handler does.
+    //
+    // NOTE: This logic is tuned to POSIX signal handling semantics, where
+    // the handler runs on the crashing thread. For monitors where that is
+    // not the case (e.g. Mach exceptions run on a port listener thread),
+    // g_crashingThread tracks the handler thread, not the faulting thread,
+    // and same-thread recrash detection does not apply. See the inline
+    // comments at each monitor's call site for details.
 
     g_requiresAsyncSafety |= isAsyncSafeEnvironment;
 


denrase:
Please look closely at the monitor changes (`SentryCrashMonitor_MachException.c`, `SentryCrashMonitor_Signal.c`, `SentryCrashMonitor_CPPException.cpp`, `SentryCrashMonitor_NSException.m`). This is all sensitive async-signal-safe code which i'm not familiar with, as we also don't seem to have extensive test coverage here. So please give me feedback if bailing out in this concurrent crash case is the correct thing to do here.

denrase:
Reproduced:

```swift
@IBAction func asyncCrash(_ sender: Any) {
    for i in 1...6 {
            fatalError("Concurrent crash #\(i)")
        }
    }
}
```

Capture:
https://denrase.sentry.io/issues/7236198917/?project=4508007036551168&query=is%3Aunresolved&referrer=issue-stream

philippsuess:
Sorry that it took so long. It just took me 2 minutes to give input 🤦 

So first, I would double-check KSCrash. They did plenty of fixes over the past, and they might have a similar fix. And second, we need a review from a native dev. We could ping Mischan, but I think he might be sick this week. 

Thanks for looking into this @denrase 😃

denrase:
@philipphofmann Ok, I aligned this with the KSCrash solution, where they suspend for 2 seconds when a concurrent thread also crashed. Also updated naming. This fixes our issue with incorrectly detecting a re-crash and also aligns us closer with KSCrash. Of course we still need a native team review here. 🙇

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_MachException.c
```diff
@@ -325,9 +325,8 @@ handleExceptions(void *const userData)
     if (g_isEnabled) {
         thread_act_array_t threads = NULL;
         mach_msg_type_number_t numThreads = 0;
-        sentrycrashmc_suspendEnvironment(&threads, &numThreads);
+        sentrycrashcm_notifyFatalException(true, &threads, &numThreads);
```
itaybre: `m`: Shouldn't `g_isHandlingCrash` be set to true before notifying the exception?

itaybre: `m`: Shouldn't `g_isHandlingCrash` be set to true before notifying the exception?

itaybre: `m`: Shouldn't `g_isHandlingCrash` be set to true before notifying the exception?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## UNreleased
```
```suggestion
## Unreleased
```
`l`: small typo

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_MachException.c
itaybre: `m`: Shouldn't `g_isHandlingCrash` be set to true before notifying the exception?

itaybre:
(sorry for the duplicated comments, github was having issues)

Sources/Sentry/include/SentryCrashMonitor.h
```diff
@@ -79,20 +80,34 @@ typedef struct {
...
 
-/** Notify that a fatal exception has been captured.
- *  This allows the system to take appropriate steps in preparation.
+/** Notify that a fatal exception has been captured, then suspend the
+ * runtime environment. Combining both into one call ensures the concurrency
+ * check always runs before any threads are suspended, preventing a deadlock
+ * where a second crashing thread freezes the first handler.
  *
- * @oaram isAsyncSafeEnvironment If true, only async-safe functions are allowed
+ * Simplified version of KSCrash's notifyException() decision logic.
```
philprime, line 88: `m`: How is it simpler? Please add a comment explaining the differences in case we need to revisit it.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
```diff
@@ -189,20 +194,63 @@ sentrycrashcm_getActiveMonitors(void)
...
         SENTRY_ASYNC_SAFE_LOG_INFO(
             "Detected crash in the crash reporter. Uninstalling SentryCrash.");
         sentrycrashcm_setActiveMonitors(SentryCrashMonitorTypeNone);
+    } else if (wasHandlingFatalException) {
+        // This is an incidental exception that happened while we were handling a
+        // fatal exception. Pause this handler to allow the other handler to finish.
+        // 2 seconds should be ample time for it to finish and terminate the app.
+        SENTRY_ASYNC_SAFE_LOG_DEBUG(
+            "Concurrent crash from different thread. Blocking to let first handler finish.");
+        sleep(2);
```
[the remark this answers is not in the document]
philprime: `h`: Instead of using a 2 second timeout, could we do this with a locking mechanism instead and combine it with a longer timeout, e.g. wait for a lock/semaphore to release or continue after 5 seconds?

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
```diff
@@ -189,20 +194,63 @@ sentrycrashcm_getActiveMonitors(void)
...
+    } else {
+        atomic_store(&g_crashingThread, self);
+    }
```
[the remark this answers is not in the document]
philprime: `h`: This seems to be valid feedback

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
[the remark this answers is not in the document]
denrase: The goal is to keep the code-base closely aligned with the KSCrash, so we can also add future upstream improvements/fixes more easily.

We are already in crashed state here, so we can't do standard locking, as we can only do async safe operations.

The assumption is that the process will, most likely be terminated anyway. In the case where it is still alive, I'm not sure that the first thread taking > 2 seconds is a valid scenario.

Sources/Sentry/include/SentryCrashMonitor.h
denrase: Updated the comment to be more precise. We "only" do fatal exception handling, but the comment structure + is closely aligned with KSCrash comments, we can follow the link if we want to compare more deeply.
 https://github.com/kstenerud/KSCrash/blob/4f5034dda74184900faeffb1717c510b6749e556/Sources/KSCrashRecordingCore/KSCrashMonitor.c#L208

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_MachException.c
denrase: Yup, good catch. Messed this up when moving suspend into notify. 🙇

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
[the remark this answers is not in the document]
denrase: @philprime Unresolved this (the bot resolved it) so you can see the answer/comment.

denrase:
@supervacuus I am once again asking for your review. 🙇

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
```diff
@@ -189,20 +194,67 @@ sentrycrashcm_getActiveMonitors(void)
...
+    //
+    // Where:
+    // - Normal handling means build a standard crash report.
+    // - Recrash handling means build a minimal recrash report and be very cautious.
+    // - Block means block this thread for a few seconds so it doesn't return
+    //   before the other handler does.
+
+    g_requiresAsyncSafety |= isAsyncSafeEnvironment;
+
+    const pthread_t self = pthread_self();
```
supervacuus, line 235: Sadly, `phtread_self()` is not async-signal-safe (the Linux and POSIX manpages list it as added for `POSIX.1-2008 TC1`, but `xOS` `sigaction` docs still leave it out). This is likely not a problem on `xOS`, but I would remark with a comment that this is an accepted trade-off. 

The only alternative is to have TLS IDs for threads that you assign in the signal handler to incoming threads from an atomic counter. However, that approach also has problems (for instance, when the SDK was `dlopen`ed after the incoming thread was created, since that would trigger the dynamic TLS machinery inside the signal handler). Fixing these problems can get hairy pretty quickly, so `pthread_self()` is a sensible trade-off.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
```diff
@@ -82,8 +86,9 @@ static int g_monitorsCount = sizeof(g_monitors) / sizeof(*g_monitors);
 
 static SentryCrashMonitorType g_activeMonitors = SentryCrashMonitorTypeNone;
 
-static bool g_handlingFatalException = false;
-static bool g_crashedDuringExceptionHandling = false;
+static _Atomic bool g_isHandlingFatalException = false;
```
supervacuus, line 92: Maybe also add a quick remark here: C11 atomics are not "officially" async-signal-safe from the POSIX POV, but on `xOS` they should all boil down to the expected atomic intrinsics/instructions and nothing more.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
```diff
@@ -189,20 +194,67 @@ sentrycrashcm_getActiveMonitors(void)
...
         SENTRY_ASYNC_SAFE_LOG_INFO(
             "Detected crash in the crash reporter. Uninstalling SentryCrash.");
         sentrycrashcm_setActiveMonitors(SentryCrashMonitorTypeNone);
+    } else if (wasHandlingFatalException) {
+        // This is an incidental exception that happened while we were handling a
+        // fatal exception. Pause this handler to allow the other handler to finish.
+        // 2 seconds should be ample time for it to finish and terminate the app.
+        SENTRY_ASYNC_SAFE_LOG_DEBUG(
+            "Concurrent crash from different thread. Blocking to let first handler finish.");
+        sleep(2);
```
supervacuus, line 262: For me, the essential design question is: do you want additional threads to report anything (and is the reporting pipeline fully reentrant to absorb them), and if not, how much did you put in the way of those threads to prevent that? Time-based is always hopium.

I fully get that you want to stay close to upstream KSCrash, and this is much better than before. In particular, this occurs at the root of 4 semantically very different OS/library-level handler mechanisms, so this seems like a great trade-off.

In the Native SDK, I recently attempted to order/block incoming threads using 3 atomic states (`IDLE`, `HANDLING`, `DONE`); this was handled orthogonally to tracking handler threads entering the signal handler (and counting preemptive recursion). But that approach only needs to map to signal handlers (on all POSIXy platforms we support) and to UnhandledExceptionFilters for Windows. I don't want to bore you, but I will report how that worked out.

denrase: The main intent was to give the first one enough time to finish and not have a second thread be in-correctly detected as same-thread re-crash. For that measure I merged in what KSCrash did to prevent this, and did not fully think through or flesh out what concurrent crash report handling would look like.

The assumption was that the sleep is long enough to give the first thread a chance to write its report and then assume the app will get terminated before the sleep is even up. Does that make sense?

denrase:
@supervacuus I have added additional comments derived from your feedback, do you think this is good enough?

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
denrase: Added a comment.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor.c
denrase: Added a comment.

supervacuus:
> @supervacuus I have added additional comments derived from your feedback, do you think this is good enough?

I think so. It is now clear that this is primarily a first preventative measure, using the worst-case monitor (the signal handler) as a baseline in the face of concurrent crashes, and is much better than how it was handled before. I know inline docs can get old quickly, but it certainly helps to orient whoever works on this next.

PR #7334, opened 2026-02-02
ref: Extract beforeSend from telemetry buffers

philipphofmann:
## :scroll: Description

This PR moves the beforeSend invocations out of the buffers, so buffers focus on buffering. This is required for aligning the telemetry buffers with the telemetry processor architecture.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/7314

#skip-changelog

## :bulb: Motivation and Context

Fixes GH-7332

## :green_heart: How did you test it?

Unit tests

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #7350, opened 2026-02-03
ref: Extract enabled check from LogBuffer

### LINKED ISSUE #7351: ref: Extract enabled check from LogBuffer
> [!NOTE]
> The pull request "[ref: Extract enabled check from LogBuffer](https://github.com/getsentry/sentry-cocoa/pull/7350)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

The log buffer should only focus on buffering logs. The client now takes care of the enabledLogs check.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/7334.

#skip-changelog

philipphofmann:
The log buffer should only focus on buffering logs. The client now takes care of the enabledLogs check.

This PR is based on https://github.com/getsentry/sentry-cocoa/pull/7334.

#skip-changelog 

Closes #7351

### DIFF
--- a/.github/workflows/test-cross-platform.yml
+++ b/.github/workflows/test-cross-platform.yml
@@ -57,7 +57,7 @@ jobs:
         with:
           repository: getsentry/sentry-react-native
           path: sentry-react-native
-          ref: antonis/cocoa-9.1.1
+          ref: v8
 
       - name: Enable Corepack
         working-directory: sentry-react-native


Sources/Swift/Tools/SentryLogBuffer.swift
```diff
@@ -10,27 +10,24 @@ import Foundation
 @objc
 @objcMembers
```
itaybre, line 11: No related to this change, but I believe having both `@objcMembers` and `@objc` is redundant

philippsuess: Thanks, I'm going to ditch this in an upcoming PR 😄, so it won't be visible to objc anymore anyways.

PR #7373, opened 2026-02-04
ref: Migrate app start measurement to swift

### LINKED ISSUE #7374: ref: Migrate app start measurement to swift
> [!NOTE]
> The pull request "[ref: Migrate app start measurement to swift](https://github.com/getsentry/sentry-cocoa/pull/7373)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Pretty straightforward migration to Swift, used claude to generate it

noahsmartin:
Pretty straightforward migration to Swift, used claude to generate it

Closes #7374


itaybre (commented):
@noahsmartin Looks good, but seems like it missed some spots

noahsmartin:
Going to close for now because SentrySDKInternal.h still uses the class and we can't use the swift class from an objc header

PR #7375, opened 2026-02-04
ref: Make integrations pure swift

### LINKED ISSUE #7376: ref: Make integrations pure swift
> [!NOTE]
> The pull request "[ref: Make integrations pure swift](https://github.com/getsentry/sentry-cocoa/pull/7375)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Now that all the integrations are out of objc, they don't need to be `NSObject` subclasses anymore

noahsmartin:
Now that all the integrations are out of objc, they don't need to be `NSObject` subclasses anymore

Closes #7376


noahsmartin:
Going to close this for now, I had just given it to claude to do in the background but it didn't quite get it, left a lot of "id" in header files

PR #3291, opened 2023-09-19
fix: Start SDK only in the main thread

brustolin:
## :scroll: Description

We always assume the SDK would be initialized in the main thread during app initialization and we design the SDK around this idea (although not documented). This change ensures the initialization of the SDK in the main thread which will avoid other issues like #3280 

## :bulb: Motivation and Context

closes #3280 

## :green_heart: How did you test it?

Unit test

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I reviewed the submitted code.
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

Issue #3277, opened 2023-09-12
Sentry 8.11.0 hangs Main thread on internal sync lock


username0x0a:
### Platform

iOS

### Installed

Swift Package Manager

### Version

8.11.0

### Steps to Reproduce

1. Integrate Sentry
2. Launch the app
3. Wait until the hang is triggered

### Expected Result

App not hanging

### Actual Result

App hanging. We've only discovered after…finding an increasing number of `WatchdogTermination` issues in Sentry soon after a release. :crazy-laugh: 🤦‍♂️

This is caused by a re-entry into the `@sychrnoized(...)` lock in `-[SentryDependencyContainer uiDeviceWrapper]`:

```objc
#if TARGET_OS_IOS
- (SentryUIDeviceWrapper *)uiDeviceWrapper
{
    if (_uiDeviceWrapper == nil) {
        @synchronized(sentryDependencyContainerLock) {
            if (_uiDeviceWrapper == nil) {
                _uiDeviceWrapper = [[SentryUIDeviceWrapper alloc] init];
            }
        }
    }
    return _uiDeviceWrapper;
}
#endif // TARGET_OS_IOS
```

which has been introduced in `8.11.0` update & has a potential to deadlock if re-entries around the `@synchronized(sentryDependencyContainerLock)` aren't sanitised properly.

“Luckily”, this also involves Main thread at some point so that it causes the issue to become visible sooner or later (app freezes dead); I can't even imagine the catastrophe if this'd silently happen purely on non-Main threads.

Here's a stack trace of colliding threads locking simultaneously:

```
Thread 1 Queue : com.apple.main-thread (serial)
#0	0x00000001d9ce77dc in __ulock_wait ()
#1	0x00000001faacefb0 in _os_unfair_lock_lock_slow ()
#2	0x0000000193dbf84c in objc_sync_enter ()
#3	0x0000000106fd80c4 in -[SentryDependencyContainer uiDeviceWrapper] ()
#4	0x0000000106f9c2a8 in -[SentryExtraContextProvider getExtraDeviceContext] ()
#5	0x0000000106f9c0bc in -[SentryExtraContextProvider getExtraContext] ()
#6	0x0000000106f7e5d8 in -[SentryClient applyExtraDeviceContextToEvent:] ()
#7	0x0000000106f7d640 in -[SentryClient prepareEvent:withScope:alwaysAttachStacktrace:isCrashEvent:] ()
[...]

PR #7377, opened 2026-02-04
feat: Add pending task queue for SDK initialization

### LINKED ISSUE #6872: setUser not work after upgade to 8.54.0
### Platform

iOS

### Environment

Production

### Installed

Swift Package Manager

### Version

8.54.0

### Xcode Version

16.4.0

_No response_

### Steps to Reproduce

be sure to set `SentrySDK.setUser(user)` after  `SentrySDK.start`

### Expected Result

`SentrySDK.setUser(user)` work

### Actual Result

no date set successfully

_No response_

itaybre:
## :scroll: Description

Fixes setUser not working when the SDK is started on a background thread.

When calling SentrySDK.start() on a background thread, the SDK dispatches initialization work to the main thread. During this window, isEnabled returns false, causing setUser calls to be silently ignored. This PR introduces a pending task queue that stores these operations and executes them once the SDK is fully initialized.

When users start the SDK on a background thread and immediately call setUser, there's a race condition where the main thread initialization hasn't completed yet. Previously, this would just log a fatal message and the user would never be set. Now, the operation is queued and executed once initialization completes.

Fixes #6872

## :pencil: Checklist

- [] I updated the docs if needed.

<sup><a href="https://cursor.com/dashboard?tab=bugbot">Cursor Bugbot</a> found 1 potential issue for commit <u>e438753</u></sup>


philprime:
While I see this is still a draft, I would like to raise two concerns/topics:

1. While we advocate to initialize the SDK on the main thread, we do not prohibit doing it from a background thread. Now when users are moving SDK initialization to the background thread because the main thread is busy on launch, the SDK initialization waiting to dispatch initialization work to the main thread could further delay SDK initialization. Not sure yet if this applies but maybe worth investigating.
2. We should consider adding a deduplication logic, so that identical tasks with side-effects are not run multiple times, e.g. if `setUser` is called once with a user and once without directly afterwards, we could skip the first call.

itaybre:
> 2. We should consider adding a deduplication logic, so that identical tasks with side-effects are not run multiple times, e.g. if `setUser` is called once with a user and once without directly afterwards, we could skip the first call.

Yeah, that's why I moved it back to draft.

> 1. While we advocate to initialize the SDK on the main thread, we do not prohibit doing it from a background thread. Now when users are moving SDK initialization to the background thread because the main thread is busy on launch, the SDK initialization waiting to dispatch initialization work to the main thread could further delay SDK initialization. Not sure yet if this applies but maybe worth investigating.

We can go down a rabbithole with enabling initializing the SDK from a background thread has 2 problems:
- Stuff might not be tracked (like events or breadcrumbs)
- Settings wont be set correctly

This fixed setting the user, but for example configuring the scope is harder to fix here.
Using the same strategy might make it asynchronous, while this is a callback I am not sure if users expect this.

I would recommend moving slowly and as needed

philippsuess:
I’m not 100% sure this solution actually addresses the root cause. The core issue seems to be a race condition in `SentrySDK.start` when it’s called from a background thread: the method returns before the hub is fully set up.

I’m mostly brainstorming here, but I see a few possible options — none of them are perfect:

### Option A: Install the hub immediately in `SentrySDK.start` (on the calling thread)
Set up a working hub right away, and delay enriching the scope (which currently happens on the main thread).
- **Pros:** We at least have a usable hub immediately.
- **Cons:** If `captureMessage` is called right after `start`, some scope data might be missing.

### Option B: Block captures until scope enrichment completes
If someone calls `captureMessage` before the main-thread work finishes, we wait until it’s done.
- **Pros:** Scope data is always correct.
- **Cons:** Calling this from a background thread can suddenly block and slow things down; this mostly just shifts the cost and only helps a subset of users.

### Option C: Block all SDK interactions until initialization fully completes
Allow `start` from a background thread, but make all subsequent SDK calls wait until the main-thread initialization finishes.
- **Pros:** Clear and consistent behavior.
- **Cons:** Adds implicit blocking and can negatively impact performance; not great ergonomics.

### Option D: Disallow background-thread initialization
Explicitly require `SentrySDK.start` to be called on the main thread.
- **Pros:** Simple and avoids the race entirely.
- **Cons:** Not very flexible and likely frustrating for some users.

### Option E: Provide a “SDK fully initialized” callback
Expose a callback that fires once all initialization (including main-thread work) is done.
- **Pros:** Explicit and gives users control.
- **Cons:** Requires users to know about and correctly use the callback.

So overall, I’m not fully convinced yet. The proposed solution seems to address only part of the problem and still introduces blocking in some scenarios. I think we need to be careful here, because all of these options come with tradeoffs, and none of them feel ideal.

itaybre:
> I’m not 100% sure this solution actually addresses the root cause. The core issue seems to be a race condition in `SentrySDK.start` when it’s called from a background thread: the method returns before the hub is fully set up.
> 
> I’m mostly brainstorming here, but I see a few possible options — none of them are perfect:
> 
> ### Option A: Install the hub immediately in `SentrySDK.start` (on the calling thread)
> Set up a working hub right away, and delay enriching the scope (which currently happens on the main thread).
> 
> * **Pros:** We at least have a usable hub immediately.
> * **Cons:** If `captureMessage` is called right after `start`, some scope data might be missing.
> 
> ### Option B: Block captures until scope enrichment completes
> If someone calls `captureMessage` before the main-thread work finishes, we wait until it’s done.
> 
> * **Pros:** Scope data is always correct.
> * **Cons:** Calling this from a background thread can suddenly block and slow things down; this mostly just shifts the cost and only helps a subset of users.
> 
> ### Option C: Block all SDK interactions until initialization fully completes
> Allow `start` from a background thread, but make all subsequent SDK calls wait until the main-thread initialization finishes.
> 
> * **Pros:** Clear and consistent behavior.
> * **Cons:** Adds implicit blocking and can negatively impact performance; not great ergonomics.
> 
> ### Option D: Disallow background-thread initialization
> Explicitly require `SentrySDK.start` to be called on the main thread.
> 
> * **Pros:** Simple and avoids the race entirely.
> * **Cons:** Not very flexible and likely frustrating for some users.
> 
> ### Option E: Provide a “SDK fully initialized” callback
> Expose a callback that fires once all initialization (including main-thread work) is done.
> 
> * **Pros:** Explicit and gives users control.
> * **Cons:** Requires users to know about and correctly use the callback.
> 
> So overall, I’m not fully convinced yet. The proposed solution seems to address only part of the problem and still introduces blocking in some scenarios. I think we need to be careful here, because all of these options come with tradeoffs, and none of them feel ideal.

I don't believe there is going to be any perfect solution here.

Doing an async initialization will always introduce problems, this is why I went for a solution that was the most friendly with the user:
- Allow setting the configuration while the SDK is not fully initialized
  - Apply it when the SDK starts
- Events, breadcrumbs, messages, etc are time sensitive and require the _full_ SDK initialized to make sense, this is why I left them as they are currently.

philippsuess:
OK, I did some digging. We moved starting parts of the SDK on the main thread in https://github.com/getsentry/sentry-cocoa/pull/3291, but we didn't want to make this synchronized because it can lead to deadlocks such as https://github.com/getsentry/sentry-cocoa/issues/3277. That's why we intentionally documented this tradeoff https://github.com/getsentry/sentry-cocoa/blob/b9e1a29cfe3cb86b8d330944b45ff2736c5c5ec5/Sources/Sentry/SentrySDKInternal.m#L186-L189

So that is the root cause, IMO. Ideally we should ensure that the SDK start is atomic, meaning after calling SentrySDK.start the SDK is completely functional, which means we would need to make the call to the main thread blocking when being called from a bg thread without introducing any deadlocks. I would prefer doing that, instead of adding a task queue or any other workaround, which adds more complexity, IMO. My comment here stating that this is a race condition we must fix, might have been a bit misleading sorry https://github.com/getsentry/sentry-cocoa/issues/6872#issuecomment-3588546914.

Or we state you shouldn't start the SDK from a bg thread. If you do that, you have to live with this tradeoff.

itaybre:
Closing this PR since is has been open for some time and no plans to immediately address this

Issue #7370, opened 2026-02-04
Migrate types from SentrySwiftUI to base SDK


philprime:
We want to get rid of the need of SentrySwiftUI to reduce the amount of distributed artifacts and build processes. SwiftUI is a first-class citizen such as UIKit and AppKit, therefore there is no reason to keep it separate.

Due to backwards compatibility, we should keep the framework itself around but move all public types. Instead it should export the same API as the base SDK does.

We need to add warnings to our changelog to make sure users are removing SentrySwiftUI as much as possible without breaking them

PR #7419, opened 2026-02-09
feat(ci): Add SentrySwiftUI to public API scripts

philprime:
In preparation for #7370 we want to add a JSON describing the public API of the SentrySwiftUI framework ABI to ensure we keep backwards compatibility even after migrating types from SentrySwiftUI to Sentry and only keep typealias.

#skip-changelog

### DIFF
--- a/.github/workflows/api-stability.yml
+++ b/.github/workflows/api-stability.yml
@@ -51,6 +51,7 @@ jobs:
           ./scripts/update-api.sh
 
       - name: Diagnose breaking changes for Sentry
+        if: always()
         run: |
           if diff -q "sdk_api_base.json" "sdk_api.json" > /dev/null; then
             echo "No Sentry API changes detected."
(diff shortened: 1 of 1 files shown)


.github/workflows/api-stability.yml
```diff
@@ -66,6 +67,24 @@ jobs:
             cat result.json
             exit 1
           fi
+
+      - name: Diagnose breaking changes for SentrySwiftUI
```
itaybre, line 72: m: if would be great if this step runs even if the sentry public api fails?

itaybre (approved):
LGTM, with a suggestion

.github/workflows/api-stability.yml
philprime: Yes, but this would make it more complicated with `if: always()` and status processing afterwards, so I feel it adds unnecessary complexity. WDYT?

Issue #5756, opened 2025-07-30
Cocoa SDK does not discard replayId after session duration limit, causing extremely long replays


philprime:
### Description

When a replay session exceeds the maximumDuration limit (60 minutes), the Cocoa SDK pauses capturing but does not reset or discard the replayId. As a result, any subsequent calls to `captureReplay()` or `getReplayId()` return the same `replayId`, leading to session replays that span multiple hours or even days.

This behavior causes events far beyond the original session to be stitched together under the same replay on the backend, violating the expected session boundaries.

### Expected Behavior:
After reaching the 60-minute limit, the SDK should:
- Stop the replay session.
- Discard the current replayId.

See https://github.com/getsentry/sentry-docs/pull/14493

### Actual Behavior:
- SDK pauses but retains the original replayId.
- `getReplayId()` continues to return the same value.
- Replays can reach excessive lengths (e.g., 336 hours), incorrectly associating unrelated events under a single session.

### Suggested Fix:
Ensure that once maximumDuration is reached:
- The replayId is removed from scope.
- A new session and replayId are generated on next start/capture.

PR #7421, opened 2026-02-09
fix: Cleanup SessionReplay when maximum duration reached

itaybre:
## :scroll: Description

Cleanup session replay Id once the replay reached the maximum duration

Fixes: #5756

Session replay were paused after reaching the maximum duration but the replay Id was not cleared, thus events long after the recording were linked to it.

## :pencil: Checklist


Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayTests.swift
```diff
@@ -88,6 +88,7 @@ class SentrySessionReplayTests: XCTestCase {
         var lastVideoUrl: URL?
         var lastReplayId: SentryId?
         var currentScreen: String?
+        var sessionReplayEndedCalled = false
```
philprime: `l`: We should use `Invocations` for call tracking for concsistency, even if it's only called once

PR #7430, opened 2026-02-11
feat: Add SentrySPM compile-from-source product and multi-platform sample apps

philprime:
## Summary

- Adds a new `SentrySPM` SPM product that compiles Sentry directly from source via a thin Swift wrapper module (`Sources/SentrySPM/`) using `@_exported import`, so consumers only need `import SentrySPM`
- Restructures the compile-from-source targets: renames `SentryObjc` to `SentryObjCInternal` (reserving `SentryObjC` for a future public wrapper), adds extensive inline documentation in `Package.swift`
- Adds SwiftUI sample apps for all supported platforms (iOS, watchOS, tvOS, macOS, visionOS) using `SentrySPM` with XcodeGen specs, Makefile build targets, and CI workflow integration

## Architecture

```
SentryHeaders      → public ObjC headers (Sources/Sentry/Public/)
_SentryPrivate     → internal ObjC headers (Sources/Sentry/include/)
SentrySwift        → all Swift source (Sources/Swift/)
SentryObjCInternal → all ObjC/C source (Sources/Sentry/, Sources/SentryCrash/)
SentrySPM          → thin re-export wrapper (@_exported import of ObjC + Swift)
```

## Test plan

- [x] CI builds all five platform sample apps (iOS, watchOS, tvOS, macOS, visionOS) via `build-sample-spm` matrix job
- [x] Verify `make build-sample-iOS-SwiftUI-SPM` succeeds locally
- [x] Verify existing SPM/xcframework products are unaffected
- [x] Verify `import SentrySPM` provides access to both ObjC types (SentrySDK, Breadcrumb) and Swift types (.sentryTrace())

#skip-changelog

Closes #7431

Issue #6342, opened 2025-10-03
Create an ObjC SDK


noahsmartin:
### Problem Statement

The current SDK aims to support both Swift and ObjC projects. However, we want to use Swift only APIs, such as enums with associated values, to improve the public interface for Swift users. This however would break objc users. To still support ObjC we should create a new SDK that wraps the Swift one. It would also be written in Swift so that it can access the Swift only APIs, but it would wrap them in an objc interface using @objc.

For example if the Swift SDK had:
```
public struct Options {
  enum AppHangTrackingType {
    case .disabled
    case .enabled(hangDuration: Int)
    case .v2(hangDuration: Int)
  }

  public let appHangTracking: AppHangTrackingType
}
```

The ObjC SDK could have:

```
import Sentry

@objc public final class SentryOptions: NSObject {
  @objc public var enableAppHangTracking: Bool = false
  @objc public var enableAppHangTrackingV2: Bool = false
  @objc public var hangDuration: Int = 0

  func makeOptions() -> Sentry.Options {
    var appHangTracking = Sentry.Options.AppHangTrackingType.disabled
    if enableAppHangTracking {
        appHangTracking = .enabled(hangDuration)
    } else if enableAppHangTrackingV2 {
        appHangTracking = .v2(hangDuration)
    }
    return .init(appHangTracking: appHangTracking)
  }
}
```

If we further want to support ObjC that doesn't use modules, such as ObjC++ in https://github.com/getsentry/sentry-cocoa/issues/4543, then we could make another SDK that wraps the Swift+@objc one that is written entirely in ObjC. This would then be able to be imported without a module import

philprime:
cc @antonis @alwx @bitsandfoxes @Flash0ver for awareness, as this new SDK might then also be consumed by the Hybrid Sentry SDKs (unless we the ObjC SDK and ObjC++ SDKs are different wrappers than an HybridSDK)

itaybre:
We will investigate the scope of this, to make sure we have a proper implementation of all the wrapper features and testing required. So we know the actual scope of what needs to be done.

philprime:
### XCFramework artifacts validated

The `SentryObjC` XCFrameworks produced by PR #7598 (run [24890645262](https://github.com/getsentry/sentry-cocoa/actions/runs/24890645262)) are structurally valid and consumable.

**Artifacts checked**
- `SentryObjC-Static.xcframework` (45 MB, static archives)
[...]

PR #7433, opened 2026-02-11
feat: Add iOS-ObjectiveCpp-NoModules sample reproducing #4543

philprime:
**This PR needs to be merged after #7430**

## Summary

- Add a sample project (`Samples/iOS-ObjectiveCpp-NoModules`) that reproduces the ObjC++ without-modules issue from #4543
- The sample uses `.mm` files with `CLANG_ENABLE_MODULES=NO` and `#import <Sentry/Sentry.h>`, which fails with `error: use of undeclared identifier 'SentrySDK'` since SDK 8.54 moved `SentrySDK` to Swift
- This serves as a test case for the pure ObjC SDK wrapper proposed in #6342 — once implemented, this sample should build successfully

## Background

Since SDK 8.54, `SentrySDK` is a Swift class only visible via `Sentry-Swift.h` or `@import Sentry`. Projects that cannot enable Clang modules (React Native ≤0.76, Haxe, custom build systems) can only use `#import <Sentry/Sentry.h>`, which no longer includes `SentrySDK`. The sample is intentionally **not** in the CI build matrix since it does not build with the current SDK.

## Test plan

- [x] `xcodegen --spec Samples/iOS-ObjectiveCpp-NoModules/iOS-ObjectiveCpp-NoModules.yml` generates the project
- [x] Building the scheme `iOS-ObjectiveCpp-NoModules` fails with `error: use of undeclared identifier 'SentrySDK'` — confirming the reproducer works
- [x] Existing CI is unaffected (sample not in build matrix)

#skip-changelog

Refs #4543
Refs #6342

### DIFF
--- a/Samples/iOS-ObjectiveCpp-NoModules/README.md
+++ b/Samples/iOS-ObjectiveCpp-NoModules/README.md
@@ -22,10 +22,10 @@ With modules disabled:
 
 - Uses **Objective-C++** (`.mm` files) for AppDelegate and ViewController
 - Sets **`CLANG_ENABLE_MODULES = NO`** in the build configuration
-- Uses `#import <Sentry/Sentry.h>` and `#import <Sentry/Sentry-Swift.h>`
-- Attempts to call `SentrySDK.startWithConfigureOptions:` and `options.sessionReplay`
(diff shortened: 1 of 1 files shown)


philipphofmann (approved):
Thanks.

Issue #7115, opened 2026-01-05
Logs not connected to spans


noahsmartin:
### Description

When I send a log with this code:

```swift
let transaction = SentrySDK.startTransaction(name: "camera-test", operation: "configure-test")
DispatchQueue.main.async {
    SentrySDK.logger.warn("Going to error")
    SentrySDK.capture(error: NSError(domain: "camera", code: 404))
    DispatchQueue.main.async {
    SentrySDK.logger.warn("After error")
    transaction.finish()
    }
}
```

I can find the log from my error in Sentry, like this: (ignore the "info pressed" that's a different log in my app)
<img width="613" height="274" alt="Image" src="https://github.com/user-attachments/assets/9ad88310-c606-47d3-9f79-c52e302091f3" />

However, when I view the trace I can see the logs and error but not the "configure-test" span. The logs and error just get put in a 0 duration transaction.

<img width="863" height="471" alt="Image" src="https://github.com/user-attachments/assets/48e05418-b4e1-43fa-86da-591287f0df99" />
<img width="808" height="596" alt="Image" src="https://github.com/user-attachments/assets/2e81362c-0a2a-4749-8e18-369b1345fe77" />

I can separately find my span through explore, and see it as expected but the logs are not connected to the same trace:

<img width="868" height="402" alt="Image" src="https://github.com/user-attachments/assets/1d08b396-010d-44e4-b9d5-36c67865bd6e" />

Isn't the expected behavior here is that logs would be visible on the trace waterfall along with other spans in the app?

philippsuess:
> Isn't the expected behavior here is that logs would be visible on the trace waterfall along with other spans in the app?

Yes, that's the expected behavior. This is a bug and we need to fix it.

philprime:
cc @denrase as you might know if this is a known issue

denrase:
Not sure what's the issue here, we could check if the correct span id is set to the logs in \`addDefaultAttributes\`.

philprime:
Now that I think about it, this might actually be fixed here:

https://github.com/getsentry/sentry-cocoa/pull/6960/files#diff-1be80454bfe69ca57aa4b50ae1c46d833019a8da059f58f07146910cfc393f43R50-R52

based on this PR comment:

https://github.com/getsentry/sentry-cocoa/pull/6960#discussion_r2676117220

philprime:
I will investigate this after #6960 is merged

philprime:
This seems to be related to #7415

[...]

PR #7439, opened 2026-02-11
docs: clarify trace_id and span_id source in TelemetryScopeApplier

### LINKED ISSUE #7415: Revisit TelemetryScopeApplier span_id assignment for logs
TelemetryScopeApplier currently sets log item traceId from the active span traceId or the propagation context traceId. Per telemetry logs docs and guidance, logs should only set `span_id` when there is an active span; a propagated span id is not considered an active span. Revisit the TelemetryScopeApplier logic to ensure `span_id` is only set when an active span exists, and confirm the correct fields (`trace_id` vs `span_id`) for log envelope items.

philprime:
## Description

Add documentation comments to `TelemetryScopeApplier` clarifying how `trace_id` and `span_id` are set for logs (and other telemetry items), with links to the official SDK telemetry docs. No behavior change; this PR is for discussion and to capture what we discovered while investigating #7415

**This PR serves as common grounds as we need to discuss if either our develop docs are wrong.**

## Motivation and Context

Closes #7415

The issue asks to ensure `span_id` is only set when there is an active span, and that a propagated `span_id` is not considered an active span.

We cross-checked the implementation against develop.sentry.dev and the codebase and concluded the following.

**Docs (develop.sentry.dev):**

- Log envelope item payload: https://develop.sentry.dev/sdk/telemetry/logs/#log-envelope-item-payload
  - `trace_id`:
> The trace id should be grabbed from the current propagation context in the SDK.

  - `span_id`: "The span id of the span that was active when the log was collected."
> The span id should be grabbed from the current active span in the SDK.

- Tracing section: https://develop.sentry.dev/sdk/telemetry/logs/#tracing

> If a log is recorded during an active span, the SDK should set the `span_id` property of the log to the span id of the span that was active when the log was collected.

- Tracing without Performance: https://develop.sentry.dev/sdk/telemetry/traces/tracing-without-performance/
  - Propagation context holds `trace_id` and `span_id` (from incoming headers or random). 

> A non-existing spanId will be propagated along with the trace" and that propagated `span_id` must not be used for logs.

**Current implementation (`TelemetryScopeApplier.swift`):**

- `trace_id`: `item.traceId = span?.traceId ?? propagationContextTraceId`. When a span is active we use its `traceId`; otherwise propagation context. When span is active, scope has been set with that span so the two sources match.
- `span_id`: Only set inside `addDefaultAttributes` when `self.span` is non-nil: `if let span = self.span { attributes["span_id"] = .string(span.spanId.sentrySpanIdString) }`. So `span_id` is only set when `scope.span` exists.

**Why no code change is required:**

- `scope.span` in Sentry Cocoa is only set when a transaction/tracer is started (`SentryHub.m` `setSpan` when starting a tracer). It is never set from propagation context. So `scope.span` always represents an active span we created.
- `TelemetryScopeApplier` only receives `propagationContextTraceId` from the scope, not the propagation context's `span_id`. So we never have access to a "propagated span_id" in this path and cannot accidentally use it. The only span we see is `scope.span`, which is the active span.
- Therefore we already only set `span_id` when an active span exists. The comments we added document this and link to the docs so future readers and other SDKs can align.

**Possible follow-ups (not in this PR):**

- The docs describe `span_id` as a top-level field in the log envelope payload. We currently put `span_id` in the `attributes` dict. Moving it to a top-level field on the log payload might require `SentryLog`/encoding changes and could be a separate change.

## How did you test it?

No behavior change. `make build-macos` and `make test-ios ONLY_TESTING=TelemetryScopeApplierTests` were run; existing tests pass.

## Checklist

- [ ] I added tests to verify the changes. (N/A: documentation only; tests could be added in follow-up)
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPii` is enabled.
- [ ] I updated the docs if needed. (Only inline comments with doc links.)

#skip-changelog

### DIFF
--- a/Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
+++ b/Samples/iOS-Swift/iOS-Swift/AppDelegate.swift
@@ -15,18 +15,38 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
-        print("[iOS-Swift] [debug] launch arguments: \(args)")
-        print("[iOS-Swift] [debug] launch environment: \(env)")
+//        print("[iOS-Swift] [debug] launch arguments: \(args)")
+//        print("[iOS-Swift] [debug] launch environment: \(env)")
+//
+//        if args.contains(SentrySDKOverrides.Special.wipeDataOnLaunch.rawValue) {
+//            removeAppData()
+//        }
+//
+//        SentrySDKWrapper.shared.startSentry()
+//        SampleAppDebugMenu.shared.display()
+//        
+//        metricKit.receiveReports()
 
-        if args.contains(SentrySDKOverrides.Special.wipeDataOnLaunch.rawValue) {
-            removeAppData()
+        SentrySDK.start { options in
+            options.dsn = SentrySDKWrapper.defaultDSN
+            options.debug = true
+            options.enableLogs = true
+            options.tracesSampleRate = 1
         }
 
-        SentrySDKWrapper.shared.startSentry()
-        SampleAppDebugMenu.shared.display()
-        
-        metricKit.receiveReports()
-        
+        let transaction = SentrySDK.startTransaction(name: "camera-test", operation: "configure-test")
+        DispatchQueue.main.async {
+            SentrySDK.logger.warn("Going to error")
+            SentrySDK.capture(error: NSError(domain: "camera", code: 404))
+            DispatchQueue.main.async {
+                SentrySDK.logger.warn("After error")
+                transaction.finish()
+                // Can't run this because it will block the main thread
+                // SentrySDK.flush(timeout: 5) 
+                print("done")
+            }
+        }
+
         return true
     }
     
--- a/Sources/Swift/Protocol/SentryLog.swift
+++ b/Sources/Swift/Protocol/SentryLog.swift
@@ -10,6 +10,9 @@ public final class SentryLog: NSObject {
     public var timestamp: Date
     /// The trace ID to associate this log with distributed tracing. This will be set to a valid non-empty value during processing.
     public var traceId: SentryId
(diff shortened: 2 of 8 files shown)


itaybre (approved):
LGTM

philprime:
## Update: `span_id` moved from attributes to top-level field

We identified that the SDK was incorrectly placing `span_id` inside the log item's `attributes` dictionary, but according to the [develop docs](https://develop.sentry.dev/sdk/telemetry/logs/#log-envelope-item-payload) and the [Python SDK reference implementation](https://github.com/getsentry/sentry-python), `span_id` should be a **top-level field** on the log item payload (alongside `trace_id`, `timestamp`, `level`, `body`, etc.).

### Changes made in this branch:
- **`TelemetryItem` protocol**: Added `var spanId: SpanId? { get set }` as a top-level property
- **`SentryLog`**: Added `spanId` as a top-level property with `Encodable` support (`"span_id"` key)
- **`SentryMetric`**: Added `spanId` stub for `TelemetryItem` conformance (metrics don't use span_id)
- **`TelemetryScopeApplier`**: Sets `item.spanId = span?.spanId` directly instead of adding `"span_id"` to `attributesDict`
- **Tests**: Updated all assertions to check `item.spanId` instead of `item.attributesDict["span_id"]`

### Still investigating: logs not appearing in trace explorer

Logs are being sent successfully (HTTP 200) and contain the correct `trace_id`, but they are **not showing up** in the Sentry trace explorer. For example: https://sentry-sdks.sentry.io/explore/traces/trace/b295c61cc5e74068b5d7e6aedce2e816

We also confirmed that the transaction was being dropped due to `sample_rate` (visible in the `client_report` in the envelope), which has been fixed by setting `tracesSampleRate = 1` in the sample app. However, even with logs sent and associated with a valid `trace_id`, they don't appear in the trace view. This may be a project configuration issue (e.g., the `ourlogs` feature flag) or a Relay/ingestion issue on the `sentry-sdks` org.

philprime:
Closing this issue as #7115 is resolved, will create a follow-up issue for moving `span_id` to the right spot

PR #7451, opened 2026-02-13
fix: Set installationId as fallback userId when user has no userId

### LINKED ISSUE #4036: ID missing after setting username field
### Platform

iOS

### Environment

Production, Develop

### Installed

Swift Package Manager

### Version

8.10.0

### Xcode Version

15.1

_No response_

### Steps to Reproduce

iOS SDK and android SDK have difference in terms of setting username. ID dissapears from webinterface after setting username on iOS, but staying on android. Example:

Android:
Снимок экрана 2024-06-04 в 23 40 40

iOS:
Снимок экрана 2024-06-04 в 23 41 12

There is ID field missing on iOS, and naming of fields is different after setting username. As i suppose if ID not set, we have fallback to installationID (which not work for some reason on iOS?..)

### Expected Result

ID field still present, after setting username or other User fields.

### Actual Result

ID field missing.

_No response_

asynclet:
## Summary
- Fixes #4036
- When a `SentryUser` object exists with fields like `email`/`username` but no `userId`, the SDK now falls back to `installationId` as the userId — aligning with Android SDK behavior
- Previously, the fallback only applied when no user object was set at all

## Test plan
- [x] Updated existing test `testInstallationIdSetWhenUserIsSetWithoutId` to verify installationId is now set as fallback
- [x] Verified `testInstallationIdNotSetWhenUserIsSetWithId` still passes — explicit userId is not overwritten
- [x] Verified `testInstallationIdSetWhenNoUserId` still passes — no-user case unchanged

### DIFF
--- a/Sources/Sentry/SentryClient.m
+++ b/Sources/Sentry/SentryClient.m
@@ -925,12 +925,15 @@ NSString *const DropSessionLogMessage = @"Session has no release name. Won't sen
 
 - (void)setUserIdIfNoUserSet:(SentryEvent *)event
 {
-    // We only want to set the id if the customer didn't set a user so we at least set something to
-    // identify the user.
+    // We want to set the installationId as the userId if no userId is set so we at least set
+    // something to identify the user. This aligns with the Android SDK behavior.
     if (event.user == nil) {
         SentryUser *user = [[SentryUser alloc] init];
         user.userId = [SentryInstallation idWithCacheDirectoryPath:self.options.cacheDirectoryPath];
         event.user = user;
+    } else if (event.user.userId == nil || event.user.userId.length == 0) {
+        event.user.userId =
+            [SentryInstallation idWithCacheDirectoryPath:self.options.cacheDirectoryPath];
     }
 }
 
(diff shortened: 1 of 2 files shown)


itaybre:
Hey @asynclet,
Thanks for the PR, looks like we forgot to add it on v9 release.
Thanks for implementing it, but this might need to be released on v10, since it is a breaking change due to the behavior change.

What do you think @philipphofmann @philprime ? We are in v9.4.1, should we add the change now?

philprime:
@itaybre I also believe it should be in a major release as discussed in https://github.com/getsentry/sentry-cocoa/issues/4036#issuecomment-2149978443

We could consider adding a V10 compilation flag now so we can merge changes already.

philippsuess:
> We could consider adding a V10 compilation flag now so we can merge changes already.

I think that's a bit too early.

We didn't include this change in V9 because we time-boxed, including breaking changes. IMO, we should wait until V10 for this.

asynclet:
Got it, makes sense to defer this to v10. Let me know if I should keep the PR open or close it until then.

philippsuess:
@asynclet I would close it and then reopen it again for v10. I added a link to this PR in the original GitHub issue. See https://github.com/getsentry/sentry-cocoa/issues/4036#issuecomment-3914324635

PR #7464, opened 2026-02-15
feat: Hang tracker using runloop observer

### LINKED ISSUE #7470: feat: Hang tracker using runloop observer
> [!NOTE]
> The pull request "[feat: Hang tracker using runloop observer](https://github.com/getsentry/sentry-cocoa/pull/7464)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

WIP just opening for unit tests to run

noahsmartin:
Converts the watchdog tracker to use a runloop observer. This avoids a CADisplayLink constantly spinning the main runloop. The hang tracker class is generic enough to be used for other use cases that are currently using display links.

Closes #7470

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Add watchdog termination tracker using RunLoop observer. Can be enabled by setting `options.experimental.enableWatchdogTerminationsV2 = true` (#7464)
+- Add watchdog termination hang tracker using RunLoop observer. Can be enabled by setting `options.experimental.enableWatchdogTerminationsV2 = true` (#7464)
 
 ### Fixes
 


philippsuess:
@noahsmartin, as the PR description states WIP, I converted this to a draft PR so it clearly signals it's not ready for review. Feel free to make it ready for review once it's ready 😃

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
```diff
@@ -33,13 +38,26 @@ final class SentryWatchdogTerminationTrackingIntegration<Dependencies: WatchdogT
...
 
         tracker.start()
-        anrTracker.add(listener: self)
+        let late = hangTracker.addLateRunLoopObserver { [weak self] id, interval in
+            guard let self, id != currentHangId, interval > timeoutInterval else {
+                return
+            }
+
+            currentHangId = id
+            hangStarted()
```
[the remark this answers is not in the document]
noahsmartin: This is wrong, it is only ever used from the background queue

philippsuess:
@cursoragent review

philippsuess:
@claude review

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
```diff
@@ -33,13 +38,26 @@ final class SentryWatchdogTerminationTrackingIntegration<Dependencies: WatchdogT
...
+        }
+        
+        let finished = hangTracker.addFinishedRunLoopObserver { [weak self] _ in
+            self?.hangStopped()
+        }
```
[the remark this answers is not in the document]
noahsmartin: Wrong, init is always called on the main thread

Sources/Swift/SentryDependencyContainer.swift
```diff
@@ -130,6 +130,7 @@ extension SentryFileManager: SentryFileManagerProtocol { }
         SentryExtensionDetector(infoPlistWrapper: Dependencies.infoPlistWrapper)
     }()
     var coreDataSwizzling = SentryCoreDataSwizzling()
+    let hangTracker: HangTracker = DefaultHangTracker(dateProvider: Dependencies.dateProvider)
```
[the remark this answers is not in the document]
noahsmartin: This isn't right, the dependency container is initialized on the main thread

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,214 @@
...
+                if self?.backgroundQueueState.lateRunLoop.isEmpty ?? false {
+                    DispatchQueue.main.async {
+                        if self?.mainQueueState.finishedRunLoop.isEmpty ?? false {
+                            self?.stop()
+                        }
+                    }
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
noahsmartin: Good catch, fixed

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,214 @@
...
+                mainQueueState.semaphore = localSemaphore
+                queue.async { [weak self] in
+                    self?.waitForHang(semaphore: localSemaphore, started: started, isStarting: true)
+                }
+            default:
+                fatalError()
+            }
+        }
+        mainQueueState.observer = observer
+        addObserver(CFRunLoopGetMain(), observer, .commonModes)
```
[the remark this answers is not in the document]
noahsmartin: These are bridged from CoreFoundation to Swift so all have `!` this is the right way to use them

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
```diff
@@ -68,18 +86,21 @@ final class SentryWatchdogTerminationTrackingIntegration<Dependencies: WatchdogT
 
     func uninstall() {
         tracker.stop()
-        anrTracker.remove(listener: self)
+        
+        guard let callbackId else {
+            return
+        }
+        hangTracker.removeLateRunLoopObserver(id: callbackId.late)
+        hangTracker.removeFinishedRunLoopObserver(id: callbackId.finishedRunLoop)
```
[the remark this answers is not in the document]
noahsmartin: Uninstall is always called on the main queue

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,214 @@
...
+        queue: DispatchQueue = DispatchQueue(label: "io.sentry.runloop-observer-checker")
+    ) {
+        self.dateProvider = dateProvider
+        self.createObserver = createObserver
+        self.addObserver = addObserver
+        self.removeObserver = removeObserver
+        self.queue = queue
+#if canImport(UIKit) && !SENTRY_NO_UI_FRAMEWORK && !os(visionOS) && !os(watchOS)
+        let window = UIApplication.shared.connectedScenes.flatMap { ($0 as? UIWindowScene)?.windows ?? [] }.first { $0.isKeyWindow }
+        let maxFPS = Double(window?.screen.maximumFramesPerSecond ?? 60)
```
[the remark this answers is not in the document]
noahsmartin: This is wrong, I tested it and worked fine in a live activity

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
```diff
@@ -3,13 +3,18 @@ import Foundation
...
+final class SentryWatchdogTerminationTrackingIntegration<Dependencies: WatchdogTerminationTrackingProvider>: NSObject, SwiftIntegration {
 
     private let tracker: SentryWatchdogTerminationTracker
-    private let anrTracker: SentryANRTracker
+    private let timeoutInterval: TimeInterval
+    private let hangTracker: HangTracker
     private let appStateManager: SentryAppStateManager
+    
+    private var callbackId: (late: UUID, finishedRunLoop: UUID)?
+    private var finishedRunLoopId: UUID?
```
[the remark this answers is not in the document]
noahsmartin: fixed

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,209 @@
...
+                    if case let .late(callback) = $0 {
+                        callback(backgroundQueueState.hangId, duration)
+                    }
+                }
+            }
+            waitForHang(semaphore: semaphore, started: started, isStarting: false)
+        case .success:
+            break
+        }
+    }
```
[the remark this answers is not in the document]
noahsmartin: Yep, will convert it

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,190 @@
...
+        return id
+    }
+    
+    // Must be called on main queue
+    func removeObserver(id: UUID) {
+        // Modifying observers requires holding the lock
+        observersLock.synchronized {
+            _ = observers.removeValue(forKey: id)
+        }
+        if observers.isEmpty {
```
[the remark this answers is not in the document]
noahsmartin: No this is fine, observers is only changed on the main thread

Sources/Swift/SentryDependencyContainer.swift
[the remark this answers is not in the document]
noahsmartin: No cursor made the same comment, the hang tracker needs to be initialized on the main thread

noahsmartin:
@cursoragent review

noahsmartin:
Cursor is getting pretty confused and commenting about things that aren't in the diff anymore when I manually ping it, but it's automated comments seem to work better and didn't notice anything, so this is ready for review!

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,226 @@
...
```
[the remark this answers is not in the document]
noahsmartin: This is a hallucination

Tests/SentryTests/Integrations/ANR/HangTrackerTests.swift
```diff
@@ -0,0 +1,210 @@
...
+    func testHangTrackerDeallocates() {
+        let dateProvider = TestCurrentDateProvider()
+        dateProvider.setSystemUptime(0)
+        var sut: DefaultHangTracker? = DefaultHangTracker(
+            dateProvider: dateProvider,
+            createObserver: createObserver,
+            addObserver: addObserver,
+            removeObserver: removeObserver,
+            queue: queue)
+        weak let weakSut = sut
```
itaybre, line 200:
```suggestion
        weak var weakSut = sut
```

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,226 @@
...
+// "beforeWaiting" we signal the semaphore. This way, if the background queue waiting times out, we know
+// it took too long for the runloop to go back to waiting and a hang has occured.
+//
+// Design requirements:
+// 1: We don't want the hang tracker to cause any thing more than a constant number of extra runloop iterations.
+// A fixed number of extra runloops per hang is acceptable, for example if it needed to run a dispatch_async
+// on the main queue, but spinning the runloop indefinitely is not acceptable.
+// 2: We don't want to aquire any locks every iteration of the runloop. It's ok to aquire locks in general
+// (the code does not need to be async signal safe) but it's not ok to aquire them on every runloop iteration.
+// 3: As simple as possible. We think it can be straightforward to detect a hang, it shouldn't require 400+ LOC.
```
itaybre:
```suggestion
// 3: As simple as possible.
```
`l`: I don't think opinions should be documented + 400+ LOC mention is irrelevant.

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,226 @@
...
+        let expectedFrameDuration = 1.0 / maxFPS
+        hangNotifyThreshold = expectedFrameDuration * 1.5
+        mainQueueState = .init()
+    }
+    
+    deinit {
+        guard let observer = mainQueueState.observer else {
+            return
+        }
+        removeObserver(CFRunLoopGetMain(), observer, .commonModes)
```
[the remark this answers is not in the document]
itaybre: I don't believe there is anything we can do here, unless we use `actor`

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,226 @@
...
+    
+    // Must be called on main queue
+    private func stop() {
+        guard let observer = mainQueueState.observer else {
+            return
+        }
+        removeObserver(CFRunLoopGetMain(), observer, .commonModes)
+        mainQueueState.observer = nil
+        mainQueueState.loopStartTime = nil
+        // If we are between beforeWaiting and afterWaiting the background queue is waiting for a signal, so let it proceed.
```
itaybre, line 180: Without DI, it don't think we can test this, right?

Sources/Swift/SentryDependencyContainer.swift
philprime: `h`: DefaultHangTracker is initialized eagerly in SentryDependencyContainer as a let property

The init comment says "Must be initialized on the main queue" (it reads UIApplication.shared.connectedScenes on UIKit platforms). If SentryDependencyContainer.sharedInstance() is first accessed from a background thread, this could crash or produce undefined behavior. Consider making it lazy or adding a main-queue assertion.

Sources/Swift/HangTracker.swift
philprime: Yes, this is unnecessary

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,226 @@
...
+#else
+        let maxFPS: Double = 60.0
+        #endif
+        let expectedFrameDuration = 1.0 / maxFPS
+        hangNotifyThreshold = expectedFrameDuration * 1.5
+        mainQueueState = .init()
+    }
+    
+    deinit {
+        guard let observer = mainQueueState.observer else {
```
philprime, line 80: `m`: `deinit` could theoretically run from a non-main thread. While `CFRunLoopRemoveObserver` for the main runloop is generally safe from any thread, the access to `mainQueueState.observer` could race with the runloop observer callback (which holds a temporary strong self via `guard let self`). This is unlikely in practice but worth noting.

Sources/Swift/HangTracker.swift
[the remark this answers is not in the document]
philprime: This is exactly the reason why I chose to use the SentryDispatchWrapper in my pull request

Sources/Swift/HangTracker.swift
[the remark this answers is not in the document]
philprime: If it is, it's very consistent. Saw this also in my PR

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,226 @@
...
+    }
+    
+    // MARK: Background queue
+    
+    // Must be called on background queue
+    private func waitForHang(semaphore: DispatchSemaphore, started: TimeInterval) {
+        var hasTimedOut = false
+        var semaphoreSuccess = false
+        while !semaphoreSuccess {
+            let timeout = DispatchTime.now() + DispatchTimeInterval.milliseconds(Int(hangNotifyThreshold * 1_000))
```
philprime, line 192: `m`: Even with a mocked dateProvider, the semaphore wait uses real wall-clock time. On heavily loaded CI machines, the background queue might not get scheduled promptly, potentially causing `testHangTrackerCallsLateRunLoop` to be slow or flaky. Consider injecting the timeout interval or making `hangNotifyThreshold` configurable for tests.

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
```diff
@@ -68,18 +83,24 @@ final class SentryWatchdogTerminationTrackingIntegration<Dependencies: WatchdogT
...
-    func anrDetected(type: SentryANRType) {
+        hasStartedHang = true
         appStateManager.updateAppState { appState in
             appState.isANROngoing = true
         }
     }
 
-    func anrStopped(result: SentryANRStoppedResult?) {
+    func hangStopped() {
+        hasStartedHang = false
```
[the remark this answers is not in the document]
philprime: I believe this is valid feedback. While in practice the callbacks arrive sequentially from waitForHang's loop, there's nothing enforcing this structurally. A lock or atomic would make the contract explicit which breaks with our core rules.

philprime, line 112: `l`: This causes unnecessary app state synchronization to disk, we could add a guard to avoid unnecessary disk I/O

```suggestion
        guard hasStartedHang else { return }
        hasStartedHang = false
```

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,226 @@
...
+    private var mainQueueState: MainQueueState
+    
+    // Must be called on main queue
+    private func startIfNecessary() {
+        guard mainQueueState.observer == nil else {
+            // Already running
+            return
+        }
+
+        let observer = createObserver(nil, CFRunLoopActivity.beforeWaiting.rawValue | CFRunLoopActivity.afterWaiting.rawValue, true, CFIndex(INT_MAX)) { [weak self] _, activity in
```
philprime, line 146: `m`: The `createObserver` could return `nil`, which is unhandled

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
```diff
@@ -33,13 +37,24 @@ final class SentryWatchdogTerminationTrackingIntegration<Dependencies: WatchdogT
...
+        timeoutInterval = options.appHangTimeoutInterval
+        hangTracker = dependencies.hangTracker
         appStateManager = dependencies.appStateManager
 
         super.init()
 
         tracker.start()
-        anrTracker.add(listener: self)
+        callbackId = hangTracker.addOngoingHangObserver { [weak self] interval, ongoing in
+            guard let self, interval > timeoutInterval else {
```
philprime, line 55: `m`: I think we need a test to make sure this filter is there.

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,226 @@
...
+    private let dateProvider: SentryCurrentDateProvider
+    private let createObserver: CreateObserverFunc<T>
+    private let addObserver: AddObserverFunc<T>
+    private let removeObserver: RemoveObserverFunc<T>
+    
+    // Observers is the only state that uses a lock. It should only be modified
+    // on the main queue, while the lock is held. Reading it on the main queue
+    // does not require a lock. Reading it on a background queue does require the lock.
+    private let observersLock = NSRecursiveLock()
+    private var observers = [UUID: (TimeInterval, Bool) -> Void]()
```
philprime, line 126: `m`: The tests are all set up with single observers, but we will have multiple ones with the ANR tracker and the watchdog termination tracker using it

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
```diff
@@ -33,13 +37,24 @@ final class SentryWatchdogTerminationTrackingIntegration<Dependencies: WatchdogT
...
+                return
+            }
+
+            if ongoing {
```
philprime, line 59: `m`: No test verifies that after one hang completes (ongoing=true then ongoing=false), a second hang is properly detected. This would catch state-reset bugs with consecutive hangs.

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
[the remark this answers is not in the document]
noahsmartin: This is AI hallucination, it's guaranteed to be called sequentially because it uses the same queue. Will update the comment

Sources/Swift/HangTracker.swift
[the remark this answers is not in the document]
noahsmartin: Part of the interface for hand tracker is that it gets used by the main thread, which is the case already, so this comment isn't really telling us anything useful, we already have the right interface to prevent bugs AFAIK

Tests/SentryTests/Integrations/ANR/HangTrackerTests.swift
noahsmartin: This would create a warning "Weak variable 'weakSut' was never mutated; consider changing to 'let' constant"

Sources/Swift/HangTracker.swift
noahsmartin: There is already a test for this in HangTrackerTests: `testRemovesObserverDuringRunloop`

Sources/Swift/SentryDependencyContainer.swift
noahsmartin: I thought we already required SentryDependencyContainer.sharedInstance() to be inited on the main queue. We definitely should... but that seems out of scope for this PR. Will make it a `var` even though it's immutable so that it's lazy

Sources/Swift/HangTracker.swift
noahsmartin: It could be, but this would be an API violation. Will add a comment

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
noahsmartin: I don't think this is the case, hangStopped() is only called when a hang actually ends, so it should always write to disk. We essentially add a precondition like `assert(hasStartedHang, "Hangs should have been started if they are stopped")` but IMO that would be an unnecessary precondition

Sources/Swift/HangTracker.swift
noahsmartin: I believe it's still handled because `addObserver` accepts a `nil` observer. So it doesn't need any special handling here, already handled by that addObserver function

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
```diff
@@ -33,13 +37,24 @@ final class SentryWatchdogTerminationTrackingIntegration<Dependencies: WatchdogT
...
+                hangStarted()
+            } else {
+                hangStopped()
+            }
+        }
```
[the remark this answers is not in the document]
noahsmartin: AI hallucination. `hasStartedHang` is always used from the same background thread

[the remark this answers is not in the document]
noahsmartin: AI hallucination again

Sources/Swift/HangTracker.swift
noahsmartin: I don't think there are any race conditions related to this? The tests inject the background queue which gives sufficient control to avoid these by starting/stopping the background queue and waiting for it to not be blocked

Sources/Swift/HangTracker.swift
noahsmartin: Good idea, added!

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
noahsmartin: Good idea, added!

Sources/Swift/HangTracker.swift
```diff
@@ -0,0 +1,230 @@
...
+        mainQueueState = .init()
+    }
+    
+    deinit {
+        guard let observer = mainQueueState.observer else {
+            return
+        }
+        removeObserver(CFRunLoopGetMain(), observer, .commonModes)
+    }
+
```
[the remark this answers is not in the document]
noahsmartin: This is a hallucination, deist won't be called until waitForHang is done running since it is an instance method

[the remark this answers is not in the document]
noahsmartin: I added an extra test to prove the AI wrong

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
noahsmartin: Yep, added!

Sources/Swift/Integrations/WatchdogTerminations/SentryWatchdogTerminationTrackingIntegration.swift
```diff
@@ -3,13 +3,17 @@ import Foundation
...
     private let tracker: SentryWatchdogTerminationTracker
-    private let anrTracker: SentryANRTracker
+    private let timeoutInterval: TimeInterval
+    private let hangTracker: HangTracker
     private let appStateManager: SentryAppStateManager
+    
+    private var hasStartedHang: Bool = false
+    private var callbackId: UUID?
 
     init?(with options: Options, dependencies: Dependencies) {
```
philprime, line 19: `h`: This should be behind an experimental flag, or at least an option to disable it.

philprime (commented):
@noahsmartin please review and resolve all open comments so we can merge it. I am uncertain which ones you already addressed

philprime (approved):
LGTM, it's experimental now so we can iterate on it 👍 Thank you for tackling this. I'll add the CHANGELOG.md and update the public API json so we can get this done

PR #7469, opened 2026-02-16
feat: Add `SentrySDK.lastRunStatus` to distinguish unknown, no-crash and crash

### LINKED ISSUE #4958: crashedLastRun returns unreliable and unexpected values and shows undocumented behavior
### Platform

iOS

### Environment

Production

### Installed

Swift Package Manager

### Version

8.46.0

### Xcode Version

16.2

_No response_

### Steps to Reproduce

The way in which it is currently possible to determine whether an application has crashed in the last session has a design flaw that makes it impossible to reliably determine whether an application has **not** crashed.

Consider the following simple example to reproduce the issue:
```swift
NSLog("### [1] Crashed last run: \(SentrySDK.crashedLastRun)")

SentrySDK.start { options in
    NSLog("### [2] Crashed last run: \(SentrySDK.crashedLastRun)")
}

NSLog("### [3] Crashed last run: \(SentrySDK.crashedLastRun)")
```

The log outputs for [1], [2] and [3] will display `false` even if the app crashed in the last session.

### Expected Result

`crashedLastRun` should not return `false` if it's currently unknown if the app did crash in the last session or not.
Returning `false` although the status is actually still unknown is at least unexpected and leads to the problem that you can never be sure in an app whether the app has actually not crashed in the last session.

An expected behavior would be that when attempting to retrieve `crashedLastRun`, although the status is currently still unknown, a value is returned that clearly signals this status, e.g. via a `nil` value or by returning an error.

In addition, the behaviour in this regard should also be documented, which it currently is not.

Now there is the `onCrashedLastRun` handler, which is called asynchronously as soon as the SDK knows that the app has crashed in the last session, but this method is not called if the app has not crashed.
In other words, there is no mechanism that signals the status as soon as the SDK knows that the app has not crashed in the last session.

A possible solution would be if `onCrashedLastRun` is also called at the moment when the SDK knows the actual status and passes this status via a parameter. However, you should probably then rename the handler method to avoid confusion.

### Actual Result

1. `SentrySDK.crashedLastRun` **always** returns `false` before the `SentrySDK.start` method has been called, no matter if the app did crash in the last session or not.
2. `SentrySDK.crashedLastRun` **always** returns `false` in the options closure of the `SentrySDK.start` method, no matter if the app did crash in the last session or not.
3. `SentrySDK.crashedLastRun` **may** return `false` after the `SentrySDK.start` call even if the app did crash in the last session since the `start` method is an asynchronous process.

_No response_

denrase:
## :scroll: Description

SentrySDK.crashedLastRun returns false both when the app didn't crash and when the status is unknown (before SDK init). This adds SentrySDK.lastRunStatus returning a tri-state enum (.unknown, .didNotCrash, .didCrash) and an onLastRunStatus callback that fires regardless of crash outcome. The old crashedLastRun and onCrashedLastRun are deprecated.

 Closes #4958

Unit tests + added logging in sample app.

Bildschirmfoto 2026-02-16 um 17 07 47

## :pencil: Checklist

### DIFF
--- a/Sources/Sentry/SentrySDKInternal.m
+++ b/Sources/Sentry/SentrySDKInternal.m
@@ -140,7 +140,7 @@ static NSDate *_Nullable startTimestamp = nil;
     return fatalDetected;
 }
 
-+ (void)setFatalEventDetected:(BOOL)value
++ (void)setFatalDetected:(BOOL)value
 {
     fatalDetected = value;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add `SentrySDK.lastRunStatus` to distinguish unknown, no-crash and crash (#7469)
+
 ## 9.7.0
 
 ### Features
@@ -10,8 +16,6 @@
 - Add package traits for UI framework opt-out (#7578).
   When building from source with Swift 6.1+ (using `Package@swift-6.1.swift`), you can enable the `NoUIFramework` trait to avoid linking UIKit or AppKit. Use this for command-line tools, headless server contexts, or other environments where UI frameworks are unavailable.
   In Xcode 26.4 and later, add the Sentry package as a dependency and the `SentrySPM` product, then enable the `NoUIFramework` trait on the package reference (Package Dependencies → select Sentry → Traits).
-  - Metric kit app hangs now report a full flamegraph rather than just one stacktrace during the hang. (#7185)
-- Add `SentrySDK.lastRunStatus` to distinguish unknown, no-crash and crash (#7469)
 - Metric kit app hangs now report a full flamegraph rather than just one stacktrace during the hang. (#7185)
 
 ### Fixes


Sources/Swift/Options.swift
```diff
@@ -156,8 +156,26 @@
     /// if you prefer a callback for every event.
     /// @warning It is not guaranteed that this is called on the main thread.
     /// @note Crash reporting is automatically disabled if a debugger is attached.
+    @available(*, deprecated, message: "Use onLastRunStatus instead, which is called regardless of whether the app crashed.")
     @objc public var onCrashedLastRun: SentryOnCrashedLastRunCallback?
 
+    /// A block called shortly after the initialization of the SDK when the crash status of the
```
denrase, line 162: Not sure if there's enough value in also calling it for normal program execution (not crashed), but added if for completnes sake. If we don't need it, we keep calling `onCrashedLastRun` like before and not deprecate it.

Sources/Sentry/SentrySDKInternal.m
```diff
@@ -469,6 +480,20 @@ + (BOOL)crashedLastRun
...
+        // SentryLastRunStatusUnknown
+        return 0;
+    }
+    if (SentryDependencyContainer.sharedInstance.crashReporter.crashedLastLaunch) {
+        // SentryLastRunStatusDidCrash
+        return 2;
+    }
+    // SentryLastRunStatusDidNotCrash
+    return 1;
+}
```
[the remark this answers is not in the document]
itaybre: Worth checking this

Sources/Swift/Options.swift
```diff
@@ -156,8 +156,26 @@
...
+    /// - If the last run did **not** end with a crash, `status` is
+    ///   ``SentryLastRunStatus/didNotCrash`` and `crashEvent` is `nil`.
+    ///
+    /// This callback is only executed once during the entire run of the program.
+    ///
+    /// - warning: It is not guaranteed that this is called on the main thread.
+    /// - note: Crashes that occur while a debugger is attached are not recorded.
+    ///   In that case, the callback reports ``SentryLastRunStatus/didNotCrash``
+    ///   even though the app did crash.
+    @objc public var onLastRunStatus: ((SentryLastRunStatus, Event?) -> Void)?
```
itaybre: `l`: The name feels odd, like missing a verb `onLastRunStatusUpdate`

denrase: Changed to `onLastRunStatusDetermined`

Sources/Sentry/SentrySDKInternal.m
```diff
@@ -39,7 +39,8 @@ @interface SentrySDKInternal ()
 @implementation SentrySDKInternal
 static SentryHubInternal *_Nullable currentHub;
 static NSObject *currentHubLock;
-static BOOL crashedLastRunCalled;
+static BOOL lastRunStatusCalled;
+static BOOL crashReporterInstalled;
```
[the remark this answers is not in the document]
denrase: Previous flag `crashedLastRunCalled ` also worked without synchronization.

Sources/Sentry/SentrySDKInternal.m
```diff
@@ -562,6 +584,9 @@ + (void)close
 
         [SentrySDKInternal setCurrentHub:nil];
 
+        crashReporterInstalled = NO;
+        lastRunStatusCalled = NO;
```
itaybre, line 593: `l`: This means that closing and starting the SDK will call `onLastRunStatus` again right?
But in theory it will always return false (if the SDK closed and started again, it would be old to have crashed)

denrase: Good catch, i will check & address this.

Sources/Sentry/SentrySDKInternal.m
```diff
@@ -452,6 +484,17 @@ + (BOOL)crashedLastRun
...
++ (NSInteger)lastRunStatus
+{
+    if (!crashReporterInstalled) {
+        return SentryLastRunStatusUnknown;
+    }
+    if (SentryDependencyContainer.sharedInstance.crashWrapper.crashedLastLaunch) {
+        return SentryLastRunStatusDidCrash;
+    }
+    return SentryLastRunStatusDidNotCrash;
+}
```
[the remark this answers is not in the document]
denrase: Updated to also handle this case.

Sources/Sentry/SentrySDKInternal.m
```diff
@@ -235,6 +257,16 @@ + (void)startWithOptions:(SentryOptions *)options
...
 
+            // The .didCrash case is handled by SentryClient.prepareEvent when
+            // a fatal event arrives. Here we report .didNotCrash if no
+            // integration set fatalDetected during init.
+            if (crashReporterInstalled && !fatalDetected) {
+                lastRunStatusCalled = YES;
+                if (nil != options.onLastRunStatusDetermined) {
+                    options.onLastRunStatusDetermined(SentryLastRunStatusDidNotCrash, nil);
+                }
+            }
```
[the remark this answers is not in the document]
denrase: Calling SDK start multiple times without end counterpart is incorrect usage.

PR #7471, opened 2026-02-17
fix: Don't finish network spans for suspended URL session tasks

### LINKED ISSUE #6836: fix: NetworkTracker wrongly finishes suspended task as aborted
### Description

The NetworkTracker treats [NSURLSessionTaskStateSuspended](https://developer.apple.com/documentation/foundation/urlsessiontask/state-swift.enum/suspended?language=objc) as a terminal state and finishes the corresponding span with a status like “aborted”. This is misleading for NSURLSessionTask–based auto-instrumentation.

In [NSURLSession](https://developer.apple.com/documentation/foundation/urlsession), a suspended task:
* is not cancelled or failed
* can be resumed later and continue the same logical HTTP operation
* may be suspended implicitly by the system (e.g. app backgrounding), which is not an error from the app’s perspective

By finishing the span on suspension, we effectively report an in-flight request as aborted, even though it may later resume and complete successfully.

Current behavior
* When a task transitions to NSURLSessionTaskState.suspended, the SDK finishes the span and sets its status to something equivalent to “aborted”.
* When a task is explicitly cancelled, the SDK finishes the span with a “cancelled” status (which is correct).

https://github.com/getsentry/sentry-cocoa/blob/f0d57aea65cc61752dffa3c68d13acae1baaedd6/Sources/Sentry/SentryNetworkTracker.m#L480-L495

Why this is a problem
* Suspended tasks are not logically finished operations.
* Spans for requests that are temporarily paused (or auto-suspended by the OS) are reported as aborted, which gives a false impression of network failures.
* This also makes it harder to reason about real cancellation vs. temporary suspension in performance data.

Expected / Open question

Suspension should not be treated as an aborted request. Instead, we should treat it as a non-terminal state and come up with a more accurate span handling strategy for NSURLSessionTaskState.suspended (and its transitions back to running, cancellation, or completion).

denrase:
## :scroll: Description

NSURLSessionTaskStateSuspended is a non-terminal state — the task can be resumed later. Treating it as terminal was causing false "aborted" spans for requests that were merely paused (e.g., by the OS during backgrounding).

Closes #6836

Unit tests

## :pencil: Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -278,8 +278,8 @@ jobs:
           # iOS 26
           - name: iOS 26 Sentry
             runs-on: tahoe
-            xcode: "26.1.1"
-            test-destination-os: "26.1"
+            xcode: "26.2"
+            test-destination-os: "26.2"
             platform: "iOS"
             device: "iPhone 17 Pro"
 
@@ -304,8 +304,8 @@ jobs:
           # macOS 26
           - name: macOS 26 Sentry
             runs-on: tahoe
-            xcode: "26.1.1"
-            test-destination-os: "26.1"
+            xcode: "26.2"
+            test-destination-os: "26.2"
             platform: "macOS"
 
           # Catalyst. We test the latest version, as the risk something breaking on Catalyst and not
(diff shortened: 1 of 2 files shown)


Sources/Sentry/SentryNetworkTracker.m
```diff
@@ -485,6 +483,8 @@ - (SentrySpanStatus)statusForSessionTask:(NSURLSessionTask *)task state:(NSURLSe
             : [self spanStatusForHttpResponseStatusCode:[self urlResponseStatusCode:task.response]];
     case NSURLSessionTaskStateRunning:
         break;
+    case NSURLSessionTaskStateSuspended:
```
philippsuess, line 501: `l`: I think it would be great to also briefly explain why suspended is not a terminal state. Not just having it in the tests, but also here in the code. The same applies to checking the new state above.

PR #7473, opened 2026-02-17
fix: Use different fallback for when MetricKit does not have file path

noahsmartin:
Sometimes metric kit only provides the binaryUUID and not the image name. This will break the in app logic and requires a full fix on the backend https://github.com/getsentry/sentry/issues/108352

In the meantime though, I noticed this only ever occurring with frames that are in the app, so we can change this default to `true` for a slight improvement. It still won't work with overrides from the default in-app behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Write reports on concurrent crashes (#7340)
 - Resolve data race crash in monitorCachedData (#7423)
 - Don't finish network spans for suspended URL session tasks (#7471)
+- Use different fallback for when MetricKit does not have file path ([#7473](https://github.com/getsentry/sentry-cocoa/pull/7473))
 
 ## 9.5.0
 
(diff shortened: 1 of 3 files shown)


Sources/Swift/Core/MetricKit/SentryMXCallStackTree+Parsing.swift
```diff
@@ -48,7 +48,19 @@ extension SentryMXCallStackTree {
                 samplesToCount[sample.frames] = sample.count + count
             }
             let frames = samplesToCount.mostSampled()?.map { $0.toSentryFrame() } ?? []
-            frames.forEach { $0.inApp = NSNumber(value: inAppLogic?.is(inApp: $0.package) ?? false) }
+            frames.forEach {
```
philippsuess, line 51: `m`: Thanks for all the context and the fix. It would be great if you could add a unit test for this.

Issue #7333, opened 2026-02-02
TelemetryProcessor: Migrate metrics


philipphofmann:
### Description

Same as https://github.com/getsentry/sentry-cocoa/issues/7280 but for metrics.

Issue #7335, opened 2026-02-02
Metrics: Record client report when dropping in beforeSendMetric


philipphofmann:
### Description

The SDK must record a client report with `kSentryDiscardReasonBeforeSend` when it drops a metric in beforeSendMetric. 

We should tackle this issue after merging https://github.com/getsentry/sentry-cocoa/pull/7334, as this PR moves beforeSendMetric to a different location.

philipphofmann:
This was implemented in #8490 (commit 0303c4fe8), which built on top of #7334's beforeSend relocation.

`SentryMetricsIntegration.addMetric` now records a client report when `beforeSendMetric` drops a metric:

```swift
guard let processedItem = beforeSendMetric(mutableMetric) else {
    SentrySDKLog.debug("Metric dropped by beforeSendMetric callback.")
    client.recordDroppedTraceMetric(inClientReport: SentryMetricObjC(metric: metricToSend))
    return
}
```
(`Sources/Swift/Integrations/Metrics/SentryMetricsIntegration.swift:64-75`)

This forwards to `SentryClient.m:1204` (`recordDroppedTraceMetricInClientReport:`), which records both a `trace_metric` count outcome and a `trace_metric_byte` size outcome, both with reason `SentryDiscardReasonBeforeSend`.

Covered by `testAddMetric_beforeSendMetricReturnsNil_recordsClientReport` in `Tests/SentryTests/Integrations/Metrics/SentryMetricsIntegrationTests.swift:203-242`.

Closing as already fixed.

Issue #10, opened 2016-05-19
Add support for file, function and line logging


milancermak:
The previous client, raven-swift, has a way how to capture messages with function name, file name and line number. This is super helpful information to have for logged events. Can you please add this functionality to this client too?

joshdholtz:
This is currently being worked on by @ahti in #12 🙌

joshdholtz:
I kind of lied on previous comment 😇  We currently don't have a good format for single frames on the server side and this also runs into some issues with symbolication of the code. This is being discussed with the team in the future on how to handle properly so keeping open for now.

edenman:
Any progress on this?  Should I just construct an `Event` and populate a single-frame Stacktrace myself?

joshdholtz:
@edenman Support for this is being added into #59 but there are some server side things that need to get taken care of first for this to work (since it has to go through the desymbolification process) in order to work properly 😇

edenman:
@joshdholtz 🙌

HazAT:
Hey guys ... checkout the latest release `1.3.0` ... You can send a stacktrace with your event.
https://docs.sentry.io/clients/cocoa/advanced/#change-event-before-sending-it

PR #7487, opened 2026-02-19
ref: Route metrics through TelemetryProcessor

philipphofmann:
Heads Up: Some CI jobs are failing due to problems that will be fixed in this PR: https://github.com/getsentry/sentry-cocoa/pull/7502

## Summary

This is the major step for achieving https://github.com/getsentry/sentry-cocoa/issues/7333 — aligning the metrics telemetry pipeline with the logs pipeline so both flow through **TelemetryProcessor → TelemetryScheduler → transport**.

- **Add metrics buffer to TelemetryProcessor and TelemetryScheduler**: Metrics now flow through the same pipeline as logs, with `SentryDefaultTelemetryProcessor` owning both a log buffer and a metrics buffer, and `SentryTelemetryScheduler` flushing both.
- **Route metrics through the client**: `SentryMetricsIntegration.addMetric()` now forwards metrics to the client's telemetry processor instead of owning its own buffer. The integration no longer conforms to `FlushableIntegration`.
- **Introduce dual-protocol hierarchy**: `SentryObjCTelemetryProcessor` (@objc, for ObjC code) and `SentryTelemetryProcessor` (Swift-only, adds `add(metric:)`) — necessary because `SentryMetric` is a Swift struct that cannot cross the ObjC boundary.
- **Remove dead code**: `captureMetricsData`/`captureData` from `SentryClient`, `FlushableIntegration` protocol, `flushIntegrations` from `SentryHub`.
- **Rewrite E2E tests**: Metrics API end-to-end tests now assert on the telemetry processor's metrics buffer directly.

Fixes GH-7333

### Manual test run

1. Running the iOS Swift sample app on the feature and the main branch
2. Checking the metric manually in the product

Screenshot 2026-02-23 at 10 43 29

#skip-changelog

### DIFF
--- a/.github/workflows/analyze-language-trends.yml
+++ b/.github/workflows/analyze-language-trends.yml
@@ -1,0 +1,77 @@
+name: Analyze Language Trends
+
+on:
+  push:
+    branches:
+      - main
+
+  workflow_dispatch:
+
+# Only the latest result matters, so cancel any in-progress run.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  analyze-languages:
+    name: Analyze Language Trends
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          # Full clone required because the script walks the entire git history
+          # to analyze language breakdown at monthly intervals since 2019.
+          fetch-depth: 0
+
+      - name: Install linguist build dependencies
+        # These are OS-level C libraries and build tools needed to compile
+        # native extensions in charlock_holmes and rugged gems. Pinning
+        # specific apt versions is impractical as runner images update
+        # frequently; the Gemfile.lock pins the actual gem versions.
+        run: sudo apt-get update && sudo apt-get install -y cmake pkg-config libicu-dev
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@09a7688d3b55cf0e976497ff046b70949eeaccfd # v1.288.0
+        with:
+          bundler-cache: true
+
+      - name: Run language analysis
+        # Start from 2019; earlier data fluctuates significantly between months.
+        run: ./scripts/analyze-languages/analyze-languages.sh --since 2019-01-01
+
+      - name: Prepare Pages site
+        run: |
+          mkdir _site
+          cp develop-docs/github-pages/index.html _site/index.html
+          cp language-trends.html _site/language-trends.html
+
+      - name: Upload Pages artifact
+        uses: actions/upload-pages-artifact@7b1f4a764d45c48632c6b24a0339c27f5614fb0b # v4.0.0
+
+      - name: Run CI Diagnostics
+        if: failure()
+        run: ./scripts/ci-diagnostics.sh
+
+  deploy:
+    name: Deploy to GitHub Pages
+    needs: analyze-languages
+    runs-on: ubuntu-latest
+
+    permissions:
+      pages: write
+      id-token: write
+
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e # v4.0.5
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,6 @@ XCFrameworkBuildPath/
 
 # Local Claude Code settings that should not be committed
 .claude/settings.local.json
+
+# Output of `make analyze-languages`
+language-trends.html
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -564,6 +564,11 @@ The repository includes a Makefile that contains common commands for building, t
 - To build the SDK for macOS use `make build-macos`, for iOS use `make build-ios`
 - To run tests use `make test-macos` or `make test-ios` for the respective platforms.
 
+## Shell Script Conventions
+
+- **Use named parameters** (`--since`, `--output`) over positional parameters (`$1`, `$2`) to prevent wrong parameters being passed as scripts evolve
+- **Extract complex logic** into separate scripts (e.g., Python for data processing) rather than inlining via heredocs, to enable IDE support and testing
+
 ## Helpful Commands
 
 - format code: `make format`
--- a/Makefile
+++ b/Makefile
@@ -541,6 +541,16 @@ format-yaml:
 # ANALYSIS
 # ============================================================================
 
+## Analyze repository language trends
+#
+# Uses github-linguist to compute the language breakdown at monthly intervals.
+# Produces an interactive HTML chart (language-trends.html) and opens it in the browser.
+# The linguist gem is managed via Bundler in scripts/analyze-languages/.
+# Optionally pass SINCE=YYYY-MM-DD to set the start date (default: 2019-01-01).
+.PHONY: analyze-languages
+analyze-languages:
+	./scripts/analyze-languages/analyze-languages.sh $(if $(SINCE),--since $(SINCE))
+
 ## Run static analysis
 #
 # Runs Xcode's static analyzer and reports any issues found.
--- a/develop-docs/README.md
+++ b/develop-docs/README.md
@@ -7,3 +7,8 @@ This is a collection of documents that can help you develop for the SentrySDK.
 - TEST.md: unit testing, UI testing and static/runtime analysis
 - BUILD.md: how we configure and build our SDK deliverables
 - RELEASE.md: our release processes and best practices
+- [Language Trends Report](https://getsentry.github.io/sentry-cocoa/language-trends.html): interactive chart of the repository's language breakdown over time, updated on every merge to main via GitHub Pages
+
+## GitHub Pages
+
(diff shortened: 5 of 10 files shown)


Sources/Swift/Tools/TelemetryProcessor/TelemetryProcessor.swift
```diff
@@ -3,31 +3,46 @@
 import Foundation
 
 /// The Telemetry processor is sitting between the client and transport to efficiently deliver telemetry to Sentry (as of 2026-02-04).
-/// Currently used for logs only; planned to cover all telemetry (e.g. metrics) with buffering, rate limiting, client reports, and priority-based sending.
+/// Currently used for logs and metrics only; planned to cover all telemetry (e.g. metrics) with buffering, rate limiting, client reports, and priority-based sending.
```
philprime: `l`: outdated comment

```suggestion
/// Currently used for logs and metrics only; planned to cover all telemetry with buffering, rate limiting, client reports, and priority-based sending.
```

Sources/Swift/Tools/TelemetryProcessor/TelemetryScheduler.swift
```diff
@@ -41,6 +42,7 @@ final class DefaultTelemetryScheduler: TelemetryScheduler {
     private func getEnvelopeInfo(telemetryType: TelemetrySchedulerItemType) -> EnvelopeInfo {
         switch telemetryType {
             case .log: return EnvelopeInfo(itemType: SentryEnvelopeItemTypes.log, contentType: "application/vnd.sentry.items.log+json")
+            case .metric: return EnvelopeInfo(itemType: SentryEnvelopeItemTypes.traceMetric, contentType: "application/vnd.sentry.items.trace-metric+json")
```
philprime: `l`: The content-type could live in an enum so it can be used in unit tests

Sources/Swift/Integrations/Metrics/SentryMetricsIntegration.swift
```diff
@@ -85,25 +50,28 @@ final class SentryMetricsIntegration<Dependencies: SentryMetricsIntegrationDepen
...
-    ///
-    /// This method is called by SentryHub.flush() via respondsToSelector: check.
-    /// We implement it directly in the class body (not in an extension) because
-    /// extensions of generic classes cannot contain @objc members.
-    /// The @objc attribute is required so Objective-C code can find this method
-    /// via respondsToSelector: at runtime.
-    @objc func flush() -> TimeInterval {
-        return captureMetrics()
-    }
+extension SentryClientInternal {
```
itaybre, line 63: `m`: Can we verify this works on static frameworks? I remember we had an issue with an extension for ObjC classes before. I am unsure if this is the case, I think @noahsmartin worked on this

Sources/Swift/Tools/TelemetryProcessor/TelemetryScheduler.swift
philippsuess: I really like the idea of using a dedicated enum-based content type (i.e. a string-backed enum) instead. That keeps things more type-safe and avoids duplication. Thanks for pointing that out 👍

But, I’d actually prefer not to reuse the same string literals in the enum tests, since that means we’d have to update the same value in multiple places if it changes.

Sources/Swift/Integrations/Metrics/SentryMetricsIntegration.swift
```diff
@@ -4,51 +4,14 @@ protocol SentryMetricsIntegrationProtocol {
...
-                }
-            ),
-            buffer: InMemoryInternalTelemetryBuffer(),
-            dateProvider: dependencies.dateProvider,
-            dispatchQueue: dependencies.dispatchQueueWrapper,
-            itemForwardingTriggers: itemForwardingTriggers
-        )
-
-        self.init(with: options, dependencies: dependencies, metricsBuffer: metricsBuffer)
-    }
```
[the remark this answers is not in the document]
philippsuess: Excellent catch. We will address this in this issue. https://github.com/getsentry/sentry-cocoa/issues/7335

Sources/Swift/Integrations/Metrics/SentryMetricsIntegration.swift
```diff
@@ -4,51 +4,14 @@ protocol SentryMetricsIntegrationProtocol {
...
-        )
-        #else
-        let itemForwardingTriggers = DefaultTelemetryBufferDataForwardingTriggers()
-        #endif
-
-        let metricsBuffer = DefaultTelemetryBuffer<InMemoryInternalTelemetryBuffer<SentryMetric>, SentryMetric>(
-            config: .init(
-                flushTimeout: 5,
-                maxItemCount: 100,
-                maxBufferSizeBytes: 1_024 * 1_024,
```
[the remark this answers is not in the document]
philippsuess: That's okay because the because the client close calls flush, which then calls the telemetry buffer, calling forward telemetry data and then calling the transport adapter, calling flush, this is an invalid concern.

Sources/Sentry/SentryHub.m
```diff
@@ -817,71 +817,12 @@ - (NSString *)createSessionDebugString:(SentrySession *)session
 
 - (void)flush:(NSTimeInterval)timeout
 {
-    // Flush flushable integrations first (e.g., metrics)
-    NSTimeInterval flushIntegrationsDuration = [self flushIntegrations:timeout];
-
-    // Calculate remaining timeout for client flush (logs and transport)
```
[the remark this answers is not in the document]
philippsuess: The client now takes care of all the flushing and still respects the time-added logic. This is a false positive.

Sources/Swift/Integrations/Metrics/SentryMetricsIntegration.swift
philippsuess: Thanks a lot for the feedback! This is not a concern because `captureMetric` is a **pure Swift extension** (no `@objc`) on `SentryClientInternal`, only called from Swift code. The known static framework stripping issue (#6125) only affects `@objc` extensions. We manually tested this with a temporary SwiftUI iOS app using SPM (statically linked) and confirmed metrics are captured and sent to Sentry without any issues.

<details>
<summary>Detailed validation and reproduction steps</summary>

### Why this is safe

- `captureMetric` has no `@objc` annotation — it's a pure Swift method
- It takes `SentryMetric` (a Swift struct) which can't cross the ObjC boundary
- Only called from Swift code (`SentryMetricsIntegration.addMetric`)
- The linker stripping issue from #6125 only affects files containing *only* `@objc` extensions

### What we tested

1. Created a temporary SwiftUI iOS app using `xcodegen`, adding `sentry-cocoa` as an SPM dependency pinned to commit `1f51b1e` (this branch)
2. Confirmed static linking with `otool -L` — no Sentry dynamic library in the binary
3. App calls `SentrySDK.start` with `enableMetrics = true`, then captures 10 counter metrics via `SentrySDK.metrics.count(key:value:attributes:)` on view appear
4. Built and launched on iPhone 16 Pro Max simulator (iOS 18.6)

### Results from debug logs

- `Integration installed: SentryMetricsIntegration` — integration loaded
- `STATIC_METRIC_TEST: Auto-captured metric #1` through `#10` — all 10 metrics captured without crash
- `TelemetryBuffer: Timer fired, calling performFlush()` — buffer flushed correctly
- `Request status: 200` — metric envelope sent to Sentry successfully

### How to reproduce (paste into Claude Code)

> Create a temporary SwiftUI iOS app at /tmp/StaticMetricTest that validates the SentryClientInternal.captureMetric Swift extension works with static linking. Use xcodegen with SPM dependency on sentry-cocoa pinned to the current branch HEAD commit. The app should call SentrySDK.start with enableMetrics=true and capture 10 counter metrics on view appear via SentrySDK.metrics.count. Confirm static linking with otool -L, build and run on the iOS simulator, capture logs, and verify: (1) no crash, (2) SentryMetricsIntegration installed, (3) metrics flushed and sent (HTTP 200). Clean up the temp project when done.

### Reproduction script

```bash
# 1. Setup
mkdir -p /tmp/StaticMetricTest/StaticMetricTest && cd /tmp/StaticMetricTest

# 2. Create xcodegen project spec
cat > project.yml << 'EOF'
name: StaticMetricTest
options:
  bundleIdPrefix: com.sentry.test
  deploymentTarget:
    iOS: "16.0"
packages:
  Sentry:
    url: https://github.com/getsentry/sentry-cocoa.git
    revision: <COMMIT_SHA>  # replace with branch HEAD
targets:
  StaticMetricTest:
    type: application
    platform: iOS
    sources:
      - StaticMetricTest
    dependencies:
      - package: Sentry
        product: Sentry
    settings:
      base:
        GENERATE_INFOPLIST_FILE: true
EOF

# 3. Create app source
cat > StaticMetricTest/StaticMetricTestApp.swift << 'SWIFT'
import Sentry; import SwiftUI
@main struct StaticMetricTestApp: App {
    init() {
        SentrySDK.start { options in
            options.dsn = "https://your-dsn@sentry.io/project"
            options.debug = true
            options.experimental.enableMetrics = true
        }
    }
    var body: some Scene { WindowGroup { ContentView() } }
}
SWIFT

cat > StaticMetricTest/ContentView.swift << 'SWIFT'
import Sentry; import SwiftUI
struct ContentView: View {
    @State private var counter = 0
    var body: some View {
        VStack(spacing: 20) {
            Text("Static Framework Metric Test").font(.headline)
            Text("Metrics captured: \(counter)").font(.title2)
        }
        .onAppear {
            for i in 1...10 {
                counter += 1
                SentrySDK.metrics.count(key: "static.framework.test.counter", value: 1, attributes: ["count": "\(i)"])
                print("STATIC_METRIC_TEST: Auto-captured metric #\(i)")
            }
        }
    }
}
SWIFT

# 4. Generate, build, verify static linking, run
xcodegen generate
xcodebuild -project StaticMetricTest.xcodeproj -scheme StaticMetricTest \
  -destination 'platform=iOS Simulator,name=iPhone 16 Pro Max' build
otool -L $(find ~/Library/Developer/Xcode/DerivedData/StaticMetricTest-*/Build -name StaticMetricTest -type f) | grep -i sentry
# Expected: no output (= statically linked)

# 5. Install and launch, check logs for "STATIC_METRIC_TEST" and no crashes

# 6. Cleanup
rm -rf /tmp/StaticMetricTest
```

</details>

PR #7503, opened 2026-02-23
chore(dx): Adopt Xcode 16 synchronized folders in XcodeGen specs

philprime:
Adopt Xcode 16's synchronized folders across all 22 XcodeGen YAML files by setting `defaultSourceDirectoryType: syncedFolder` as the project-wide default.

Synchronized folders keep the Xcode project automatically in sync with the file system, eliminating the need to regenerate projects when files are added or removed. This also speeds up XcodeGen generation since synced folders skip per-file tracking.

The iOS-ObjectiveC spec explicitly overrides to `type: group` for its source with `excludes`, since `syncedFolder` does not support that option.

Also updates the XcodeGen Homebrew formula from 2.43.0 to 2.44.1, which is the minimum version supporting synchronized folders, and includes regenerated pbxproj/workspace/test plan changes.

#skip-changelog


philprime:
Superseded by https://github.com/getsentry/sentry-cocoa/pull/7530 and https://github.com/getsentry/sentry-cocoa/pull/7531

PR #7158, opened 2026-01-12
ref: convert AutoBreadcrumbTrackingIntegration to Swift

philprime:
## :scroll: Description

This PR converts the auto breadcrumb tracking integration and related types from Objective-C to Swift. The main conversion includes:

- `SentryAutoBreadcrumbTrackingIntegration` - Main integration class
- `SentrySystemEventBreadcrumbs` - System event breadcrumb tracking

Additionally, `SentryBreadcrumbDelegate` and `SentryBreadcrumbTracker` were also converted to Swift due to Swift-Objective-C interoperability compilation issues encountered during the migration. Converting these types to Swift was more reasonable than attempting to fix the complex interop issues, especially given the project's direction toward Swift.

The conversion also includes:

- Refactored `UICurrentDeviceProvider` to use `uiDeviceWrapper` from `SentryDependencyContainer` via a new `SentryUIDeviceWrapperProvider` protocol
- Added `swizzleViewDidAppear` helper method to `SentrySwizzleWrapperHelper` to support Swift-based swizzling
- Fixed pre-existing header path issue for `SentryAsyncSafeLog.h` that was exposed by stricter SPM build requirements

## :bulb: Motivation and Context

This change is part of the ongoing migration from Objective-C to Swift in the Sentry Cocoa SDK. Converting the breadcrumb tracking integration to Swift improves code maintainability, type safety, and consistency with the rest of the Swift codebase.

The additional conversion of `SentryBreadcrumbDelegate` and `SentryBreadcrumbTracker` was necessary due to Swift-Objective-C interoperability compilation issues. Rather than attempting to resolve complex interop problems, converting these types to Swift was the more pragmatic approach and aligns with the project's Swift migration goals.

## :green_heart: How did you test it?

- Verified SDK builds successfully for iOS, macOS, tvOS, watchOS, and visionOS
- Ran unit tests: `make test-ios`
- Verified breadcrumb tracking functionality works correctly
- Confirmed no linter errors
- Fixed pre-existing `SentryAsyncSafeLog.h` header path issue for SPM builds

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

[...]

PR #7508, opened 2026-02-23
Fix optional chaining for window property access

benjaminpeters:
## :scroll: Description

if `view` is an implicitly unwrapped optional (UIView!), accessing it when it’s nil throws `“Unexpectedly found nil while implicitly unwrapping an Optional value.”` This PR makes the Swift version safe like previous Obj-C `SentryBreadcrumbTracker.m`‎ by not forcing unwrap view. Use optional chaining so the whole expression can be nil without crashing

Similar to how it worked before: https://github.com/getsentry/sentry-cocoa/pull/7158/changes#diff-da54f2acc5bad1b3a40d7541411b3965ff7c51aaba660842e1f899ba3b526ce2

I'm seeing a crash while trying to extract data from a google IMA: `IMAWebUIViewController`
The crash first appeared when I bumped the react-native-sentry sdk from v7 -> v8. I tracked the crash to `SentryBreadcrumbTracker`

Made the change locally and confirmed everything worked.
I've written unit tests to confirm the breadcrumb will not try to add missing window data if it is not available

## :pencil: Checklist


Sources/Swift/Integrations/Breadcrumbs/SentryBreadcrumbTracker.swift
```diff
@@ -275,7 +275,7 @@ import Cocoa
             info["parentViewController"] = SwiftDescriptor.getViewControllerClassName(parentViewController)
         }
         
-        if let window = controller.view.window {
+        if let window = controller.view?.window {
```
philprime, line 278: `m`: It would be great if you could add a test to the `SentryBreadcrumbTrackerTests` which covers this case.

Sources/Swift/Integrations/Breadcrumbs/SentryBreadcrumbTracker.swift
```diff
@@ -275,7 +275,9 @@ import Cocoa
             info["parentViewController"] = SwiftDescriptor.getViewControllerClassName(parentViewController)
         }
         
-        if let window = controller.view.window {
+        // Access view to ensure it's loaded (triggers loadView() if view is nil)
```
benjaminpeters: For [uiviewcontroller's](https://developer.apple.com/documentation/uikit/uiviewcontroller/view) we need to access the view to trigger `loadView()` when the value is `nil`

philprime: How is there a difference between
```
let view = controller.view
if let window = view?.window {
```
and 
```
if let window = controller.view?.window {
```

According to the documentation both should work equally:
> If you access this property when its value is nil, the view controller automatically calls the [loadView()](https://developer.apple.com/documentation/uikit/uiviewcontroller/loadview()) method and returns the resulting view.

Tests/SentryTests/Integrations/Breadcrumbs/SentryBreadcrumbTrackerTests.swift
```diff
@@ -319,6 +319,80 @@ class SentryBreadcrumbTrackerTests: XCTestCase {
...
+        let lifeCycleCrumb = try XCTUnwrap(crumbs.last)
+        XCTAssertEqual("navigation", lifeCycleCrumb.type)
+        XCTAssertEqual("ui.lifecycle", lifeCycleCrumb.category)
+
+        let data = try XCTUnwrap(lifeCycleCrumb.data)
+        XCTAssertNotNil(data["window"] as? String, "Breadcrumb should include window description when view controller is in a window")
+        XCTAssertNotNil(data["window_isKeyWindow"] as? String)
+        XCTAssertNotNil(data["window_windowLevel"] as? String)
+        XCTAssertEqual(data["is_window_rootViewController"] as? String, "true", "ViewController is root of the window")
+    }
```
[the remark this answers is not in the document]
philprime: Tests are always running on the main thread

Sources/Swift/Integrations/Breadcrumbs/SentryBreadcrumbTracker.swift
benjaminpeters: yeah they're identical I thought somehow assigning it was causing it to trigger that loadView() func but I was wrong. I'm not entirely sure why it's not working. This is what I'm seeing when inspecting the problematic view:
Screenshot 2026-02-25 at 9 11 55 AM
it's possible I guess that the view controller has been deallocated or some how in a bad state before getting to this part of the code

philprime: It seems this is a bug in the view controller, maybe overwriting something it shouldn't (wouldn't be the first time [this happens](https://sentry.engineering/blog/reverse-engineering-ios-to-fix-sdk-crashes))

Let's just add the `?` to the if-let statement and we can merge this PR

benjaminpeters: sounds good!

philprime:
Almost LGTM, please add an entry to the `CHANGELOG.md` like this at the top:

```
# Changelog

## Unreleased

### Fixes

- Fix unexpected null value when unwrapping view controller window  (#7508)
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- Fix unexpected null value when unwrapping view controller window  (#7508)
```
philprime: Unfortunately it was added to the already released section but I don't think I can write to your PR.


```suggestion
## Unreleased

### Fixes

- Fix unexpected null value when unwrapping view controller window  (#7508)

## 9.5.1

### Fixes

```

philprime: Looks like this made it worse 😄 Good job, GitHub.
Sorry for bothering you with this.

benjaminpeters: Ah whoops! just pushed in a fix

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Fix unexpected null value when unwrapping view controller window  (#7508)
```
```suggestion
- Fix unexpected null value when unwrapping view controller window (#7508)
```

PR #7515, opened 2026-02-25
fix: Preserve NSException reason over unrelated crash_info_message

### LINKED ISSUE #7516: fix: Preserve NSException reason over unrelated crash_info_message
> [!NOTE]
> The pull request "[fix: Preserve NSException reason over unrelated crash_info_message](https://github.com/getsentry/sentry-cocoa/pull/7515)" was created by @philprime but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

Preserve the original NSException and C++ exception reason instead of
overwriting it with an unrelated `crash_info_message` from
libswiftCore.dylib's `__crash_info` section.

The `__crash_info` section is a shared buffer — any Swift runtime warning
written there (e.g. `SWIFT TASK CONTINUATION MISUSE`) persists even if
an unrelated crash occurs afterwards. The `enhanceValueFromCrashInfoMessage:`
method unconditionally replaced `exception.value` with whatever was in
that buffer, causing Sentry issues to show misleading descriptions that
sent users debugging the wrong thing.

For `nsexception` and `cpp_exception` types, the SDK already has the
authoritative crash reason from the exception context. The
`crash_info_message` override is now skipped for these types, while
preserved for `mach`/`signal` exceptions where it provides the actual
crash cause (e.g. `fatalError`, `assertionFailure`).

philprime:
Preserve the original NSException, C++ exception, and user-reported exception reason instead of overwriting it with an unrelated `crash_info_message` from libswiftCore.dylib's `__crash_info` section.

The `__crash_info` section is a shared buffer and any Swift runtime warning written there (e.g. `SWIFT TASK CONTINUATION MISUSE`) persists even if an unrelated crash occurs afterwards. The `enhanceValueFromCrashInfoMessage:` method unconditionally replaced `exception.value` with whatever was in that buffer, causing Sentry issues to show misleading descriptions that sent users debugging the wrong thing.

For `nsexception`, `cpp_exception`, and `user` types, the SDK already has the authoritative crash reason from the exception context. The `crash_info_message` override is now skipped for these types, while preserved for `mach`/`signal` exceptions where it provides the actual crash cause (e.g. `fatalError`, `assertionFailure`).

When we don't override, we attach `crash_info_messages` to `mechanism.data` per the [Exception Interface](https://develop.sentry.dev/sdk/foundations/transport/event-payloads/exception/#data) so the info remains available for debugging. Both `crash_info_message` and `crash_info_message2` are preserved when present.

Closes #7516


philprime:
@sentry review

Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -461,7 +461,23 @@ - (SentryDebugMeta *)debugMetaFromBinaryImageDictionary:(NSDictionary *)sourceIm
...
+    // Swift runtime warnings (e.g. SWIFT TASK CONTINUATION MISUSE) that happened before an
+    // unrelated crash. Only override for mach/signal exceptions where crash_info_message IS the
+    // crash cause (fatalError, assertionFailure, etc.).
+    //
+    // The "user" type covers programmatically reported exceptions (SentrySDK.capture). While
+    // unlikely to co-occur with a stale __crash_info buffer, excluding it is correct because the
+    // user-supplied reason is always authoritative.
+    if (![exceptionType isEqualToString:@"nsexception"]
+        && ![exceptionType isEqualToString:@"cpp_exception"]
+        && ![exceptionType isEqualToString:@"user"]) {
```
philippsuess: `m`: Maybe a better approach would actually be to have an allow list and not a disallow list, so that we only enhance this for specific types of exceptions. Is that too complicated to narrow down?

Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -461,7 +461,23 @@ - (SentryDebugMeta *)debugMetaFromBinaryImageDictionary:(NSDictionary *)sourceIm
...
+        [self enhanceValueFromCrashInfoMessage:exception];
```
philippsuess: `m`: There might be valuable information in some issues. The current problem is that, even when we stop overriding it in the wrong scenarios, we should still check if there's information present. We need to find a place within the event to attach this information. I'm not entirely sure where exactly we could add it, but perhaps examining the event payload could help us identify a spot where this information could still be stored if available.

Sources/Sentry/SentryCrashReportConverter.m
philprime: I believe the key challenge there would be figuring out all possible cases where we want to enhance it, and if we miss one, it could cause a regression with data not being reported to Sentry

Sources/Sentry/SentryCrashReportConverter.m
philprime: Can you please elaborate on what you mean with "help us identify a spot where this information"

philippsuess: Basically looking at https://develop.sentry.dev/sdk/foundations/transport/event-payloads/exception/ and find a place to send the information. Sorry, I could have posted that link earlier.

Sources/Sentry/SentryCrashReportConverter.m
itaybre: Even if we don't know it, wouldn't it be better to use a `NSSet` here?
```
NSSet *allowedExceptionTypes = [NSSet setWithObjects:
    @"nsexception",
    @"cpp_exception",
    @"user",
    nil];

if (![allowedExceptionTypes containsObject:exceptionType]) {
...
```

Mainly because I expect this list to change over time

philprime: Yes, I can change it to a set, definitely cleaner, it just matched the existing code pattern.

Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -461,9 +461,34 @@ - (SentryDebugMeta *)debugMetaFromBinaryImageDictionary:(NSDictionary *)sourceIm
...
+
     exception.mechanism = [self extractMechanismOfType:exceptionType];
 
+    // Attach crash_info to mechanism.data so it remains available for debugging even when we
+    // don't use it as the primary value. See
+    // https://develop.sentry.dev/sdk/foundations/transport/event-payloads/exception/#data
+    NSSet<NSString *> *authoritativeExceptionTypes =
+        [NSSet setWithObjects:@"nsexception", @"cpp_exception", @"user", nil];
+    if ([authoritativeExceptionTypes containsObject:exceptionType] && exception.mechanism != nil) {
+        NSArray<NSString *> *crashInfoMessages = [self crashInfoMessagesFromBinaryImages];
```
philippsuess: `l`: We now get the crash info from the binary images twice. We could only get it once and use it here and also above for `enhanceValueFromCrashInfoMessage`. So basically, call `crashInfoMessagesFromBinaryImages`, store it as a variable, pass it to `enhanceValueFromCrashInfoMessage`, and use it here as well.

Tests/SentryTests/SentryCrashReportConverterTests.m
```diff
@@ -839,4 +839,317 @@ - (void)testNotableAddressesWithNilValue_ShouldBeSkipped
...
+
+    // -- Assert --
+    SentryException *exception = event.exceptions.firstObject;
+    XCTAssertEqualObjects(exception.type, @"SIGABRT");
+    XCTAssertTrue([exception.value containsString:crashInfoMessage],
+        @"Signal exception value should be overridden by crash_info_message. Got: %@",
+        exception.value);
+}
+
+- (void)testUserException_whenCrashInfoMessagePresent_shouldPreserveOriginalReason
```
philippsuess, line 1109: Thanks for the extensive test coverage.

Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -461,9 +461,35 @@ - (SentryDebugMeta *)debugMetaFromBinaryImageDictionary:(NSDictionary *)sourceIm
     }
 
     [self enhanceValueFromNotableAddresses:exception];
-    [self enhanceValueFromCrashInfoMessage:exception];
+
+    NSArray<NSString *> *crashInfoMessages = [self crashInfoMessagesFromBinaryImages];
```
philippsuess, line 465: Better thanks 💯

philipphofmann (approved):
LGTM

PR #7524, opened 2026-02-25
chore: Don't swiftlint generated sample code 

43jay:
## :scroll: Description

1) Exclude intermediate build artifacts from `iOS-Swift` and `iOS-SwiftClip` build targets (.swiftlint.yml).
2) Don't run swiftlint if there are no staged .swift files (`make lint-staged`)

Saw these while extending the iOS-Swift sample app to test [MOBILE-935](https://linear.app/getsentry/issue/MOBILE-935/mobile-replay-capture-network-requestresponse-bodies) and claude proposed the fixes, YMMV. 

### 1) pre-commit failing from lint errors in build artifacts. 
This seems unnecessary; swiftlint should lint source .swift files (which it does), and not worry about generated .swift files. 
The errors surfaced were unrelated to my changes => unclear fix => resorted to `--no-verify`.

### 2) pre-commit failing from lint errors unrelated to my commit (edge-case)
steps to repro:
 - Try to create a commit with no staged .swift files and some unstaged .swift changes.
 - `make lint-staged` was generating an empty `$(STAGED_SWIFT_FILES)`
 - `swiftlint` with no `<paths>` argument would run and fail on unrelated (unstaged) .swift lint errors

swiftlint was erroring in 2 scenarios before; not erroring after

## :pencil: Checklist


.swiftlint.yml
```diff
@@ -14,6 +14,9 @@ excluded:
   - XCFrameworkBuildPath/**
   - 3rd-party-integrations/**/Packages/sentry-cocoa
   - 3rd-party-integrations/**/.build
+  # Exclude auto-generated files from sample app builds
+  - "**/Build/Intermediates.noindex/iOS-Swift.build/**/DerivedSources/**"
+  - "**/Build/Intermediates.noindex/iOS-SwiftClip.build/**/DerivedSources/**"
```
philippsuess, line 19: `l`: Maybe we could even make this more generic and say that if the path contains `DerivedSources` we shouldn't run the linter, then it would work for other sample apps out of the box as well.

Makefile
```diff
@@ -497,7 +497,9 @@ lint-staged:
 	@echo "--> Running Swiftlint and Clang-Format on staged files"
 	./scripts/check-clang-format.py -r Sources Tests
 	ruby ./scripts/check-objc-id-usage.rb -r Sources/Sentry
-	swiftlint --strict --quiet $(STAGED_SWIFT_FILES)
+	@if [ -n "$(STAGED_SWIFT_FILES)" ]; then \
```
philippsuess, line 500: Nice thanks.

philipphofmann (approved):
Thanks.

.swiftlint.yml
43jay: hmm. I'm hesitant to make it less constrained w/out spending time to understand what it would affect. E.g. how that could interact w/ sdk builds.

PR #7533, opened 2026-02-26
style: Apply formatting with clang-format v22.1.0

philprime:
Updates clang-format to v22.1.0 (superseding #7532) and fixes all issues caught by `make format`.

#skip-changelog

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ init: init-local init-ci-build init-ci-format
 .PHONY: init-local
 init-local:
 	which brew || /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
-	brew bundle
+	brew update && brew bundle
 	pre-commit install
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-clang-formatting.yml
```diff
@@ -45,7 +45,9 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
       - name: Install tooling
-        run: make init-ci-format
+        run: |
+          make init-ci-format
+          clang-format --version
```
philippsuess, line 50: `m`: The step is named `Install tooling`, but it now also calls `clang-format --version` and we still have `make format-clang` below. Is this intentional?

philprime: It is intentional because for some reason it keeps installing 21.1.8 and not 22.1.0 so I added this to see what was installed

PR #7549, opened 2026-02-27
ref: Address CWE-676 use of potentially dangerous functions

### LINKED ISSUE #2785: Use of potentially dangerous functions
### Description

A customer reported that their security vulnerability tool reported our repository has the following security issue: [CWE-676](https://cwe.mitre.org/data/definitions/676.html), which stands for the use of potentially dangerous functions.

Dangerous functions: 
- [ ] https://github.com/getsentry/sentry-cocoa/issues/2783
- [ ] https://github.com/getsentry/sentry-cocoa/issues/2784

The following usage functions need clarification:
- [ ] [memcpy](https://github.com/search?q=repo%3Agetsentry%2Fsentry-cocoa+memcpy&type=code) - what are the downsides? Should we replace their usage?
- [ ] [sscanf](https://github.com/search?q=repo%3Agetsentry%2Fsentry-cocoa+sscanf&type=code) - Maybe replace it with safer scanf_s?
- [ ] [strlen](https://github.com/search?q=repo%3Agetsentry%2Fsentry-cocoa+strlen&type=code) - you do not know the size of the original source buffer when using it. Is there a safer API?
- [ ] [calloc](https://github.com/search?q=repo%3Agetsentry%2Fsentry-cocoa+calloc&type=code). Should we replace our usages with `malloc`?

philprime:
Address CWE-676 (use of potentially dangerous functions) reported by customer security tools. Replaces sscanf with strtoull/strtod in the report store and JSON codec so we avoid flagged APIs; scanf_s is not available on Apple platforms. Adds CWE-676 safety comments at all memcpy, strlen, and calloc sites documenting bounds or null-termination guarantees. Keeps calloc for zero-initialization and adds overflow guards where count*size could overflow (dirContents, image renderer). Adds unit tests for report ID parsing and floating-point decode.

Fixes #2785


philprime:
@sentry review

Sources/SentryCrash/Recording/Tools/SentryCrashJSONCodec.c
```diff
@@ -1265,9 +1271,7 @@ decodeElement(const char *const name, SentryCrashJSONDecodeContext *context)
         strncpy(context->stringBuffer, start, len);
         context->stringBuffer[len] = '\0';
 
-        // Parses a floating point number from the string buffer into value using %lg format
-        // %lg uses shortest decimal representation and removes trailing zeros
-        sscanf(context->stringBuffer, "%lg", &value);
+        value = strtod(context->stringBuffer, NULL);
```
philprime, line 1308: Need to double-check if this is truly the same

philprime:
@sentry review

Sources/Swift/Core/Tools/ViewCapture/SentryGraphicsImageRenderer.swift
```diff
@@ -48,9 +48,23 @@ final class SentryGraphicsImageRenderer {
...
+        // back the bitmap before the drawing actions run (e.g. when the image is queried mid-
+        // capture), and uninitialized bytes would surface as garbled pixels.
+        let byteCount = pixelsPerColumn.multipliedReportingOverflow(by: bytesPerRow)
+        guard !byteCount.overflow, byteCount.partialValue > 0,
+              let rawData = calloc(byteCount.partialValue, MemoryLayout<UInt8>.size) else {
+            if byteCount.overflow {
+                SentrySDKLog.error("Image dimensions cause allocation overflow")
+            } else {
+                SentrySDKLog.error("Unable to allocate memory for image data")
+            }
```
supervacuus: This seems to mix two things (at least when reading the comment).
 
- If the concern is only that `calloc()` might be called with an overflowed size, then this is arguably more than we need: Swift evaluates the `Int` arithmetic before the call, so if `pixelsPerColumn * bytesPerRow`  overflows, Swift traps before `calloc()` is even entered. In that sense, we are already safe from passing an overflowed size to `calloc()`. A simple `byteCount > 0` sanity check would still make sense.
- If the concern is instead that these calculations could trap, and we want to fail gracefully, then the current change is incomplete: `bytesPerRow = bytesPerPixel * pixelsPerRow` is still unchecked and can trap before the later `multipliedReportingOverflow` ever runs. In that case, I would check the whole arithmetic chain, and probably also validate the `CGFloat` -> `Int` conversions at the top of `image()`.

Sources/Sentry/SentryAsyncSafeLog.c
```diff
@@ -82,6 +82,9 @@ static void
 writeToLog(const char *const str)
 {
     if (g_fd >= 0) {
```
supervacuus, line 86: Maybe we should still check for `str != NULL` given that we unconditionally call `strlen()` next. Even if the current code is never called with a null pointer. Or we add a comment that this is the contract of `writeToLog()`.

Sources/SentryCrash/Recording/SentryCrashReportStore.c
```diff
@@ -75,26 +75,43 @@ getCrashReportPathByID(int64_t id, char *pathBuffer)
...
+    if (strncmp(filename, g_appName, appNameLen) != 0) {
+        return 0;
+    }
+    const char *rest = filename + appNameLen;
+    if (strncmp(rest, "-report-", 8) != 0) {
+        return 0;
+    }
+    const char *hexStart = rest + 8;
+    char *endPtr = NULL;
+    const uint64_t id = strtoull(hexStart, &endPtr, 16);
```
supervacuus, line 96: While it is true that this is safer wrt buffer bounds, `strtoull()`, like the previous format specifier `PRIx64`, accepts more than you might want. This is not necessarily something that needs to be fixed in this PR, just wanted to raise that this code also accepts:

- `myapp-report- 1.json`
- `myapp-report-+1.json`
- `myapp-report-0x1.json`
- `myapp-report--1.json`

which can be fine, I am not sure how tight the format bounds are.

supervacuus (commented):
This looks largely sensible. Only a couple of remarks.

Sources/Swift/Core/Tools/ViewCapture/SentryGraphicsImageRenderer.swift
philprime: Thank you for your feedback, I also considered this as being over-the-top because the graphics image renderer is an internal tool, so using it with a way to larger buffer would also cause other problems in the SDK.

I'll revert the changes and add a comment instead.

Sources/Sentry/SentryAsyncSafeLog.c
philprime: It's unlikely to happen because writeToLog is only used at call sites where it's already null-safe, but I'll add it anyways for safety 👍

Sources/SentryCrash/Recording/SentryCrashReportStore.c
philprime: Thanks for raising this! Tracing back the call sites I can see that `getReportIDFromFilename` is called from `getReportIDs` and `getReportCount` for all files in the `g_reportsPath`. 

While it is technically possible for someone to tamper with the names of the files in the `g_reportsPath`, the SDK is only writing to paths using zero-padded `%016llx`.

Unless you believe there is a risk I would leave it as-is.

supervacuus (approved):

Sources/SentryCrash/Recording/SentryCrashReportStore.c
supervacuus: If the `id` isn't used as a literal index to do an unsafe memory access, everything is fine. The buffer safety has improved, and you also fixed the type of `id` (previously `int64_t`, which would cause signed overflow for a considerable range of valid parse values, which, in C, sadly, is undefined behavior). So the worst thing tampering could achieve would be having the wrong report ID. Can't say how "risky" this is downstream, but locally it is not.

Sources/Swift/Core/Tools/ViewCapture/SentryGraphicsImageRenderer.swift
supervacuus: bounding by `byteCount > 0` can still make sense, but thanks to the sanity of Swift, overflows trap.

Issue #4944, opened 2025-03-05
[Mobile Replay] Capture Network request/response bodies


philprime:
### Description

See https://github.com/getsentry/sentry/issues/84596

43jay:
Been working through this with input from different sources. Follows a write-up of the discussion (@philprime) to swizzle additional methods instead of using NSURLProtocol - for posterity and in case of missing important input 💪.

cc @itaybre @philprime @philipphofmann 

## Approach for Capturing Network Request/Response Bodies

### Context

We need to capture HTTP request/response bodies transparently on iOS. Two approaches were evaluated:

1. **NSURLProtocol** — intercept requests at the URL loading system level

2. **Additional swizzling** — extend `SentryNetworkTracker` (which already swizzles `NSURLSessionTask.resume` and `setState`) to also swizzle methods like `dataTaskWithRequest:completionHandler:`

### Why Not NSURLProtocol

- Easily bypassed — any configuration that overwrites `protocolClasses` drops Sentry's protocol silently

- Breaks certificate pinning (the protocol's internal session handles TLS, bypassing the app's pinning delegate)

- Risks silently dropping, e.g. delegate callbacks

### Recommendation: Swizzling

Swizzling is preferred because there's already a stable implementation (`SentrySwizzleWrapperHelper`), and we will ship the new implementation with:

- Gate behind an **explicit opt-in** with description of potential issues so that responsibility to verify it works is delegated to the app developer.
- Thorough testing of the new swizzled methods for different iOS versions.

### TODOs

- [x] What swizzling is needed to get sufficient coverage of the different types of network requests?  [see below](https://github.com/getsentry/sentry-cocoa/issues/4944#issuecomment-3856073897)

- [x] Do the crash scenarios [referenced in `SentryNSURLSessionTaskSearch`](https://github.com/getsentry/sentry-cocoa/blob/ee272e8ed2975860f399337706be68f511d327a8/Sources/Sentry/SentryNSURLSessionTaskSearch.m#L22) apply to these new methods? [see below](https://github.com/getsentry/sentry-cocoa/issues/4944#issuecomment-3856482983)

- [ ] Any additional safeguards missing?

---

43jay:
> What swizzling is needed to get sufficient coverage of the different types of network requests?
# Breakdown - new swizzling for response body

[...]

PR #7580, opened 2026-03-03
feat(network-details): Extend SentryReplayOptions API with Session Replay Network Details configuration

43jay:
Implement SDK fields, setters and validation for network details capture: https://docs.sentry.io/platforms/javascript/session-replay/configuration/#network-details

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `networkDetailAllowUrls` | `(string\|RegExp)[]` | `[]` | Capture request and response details for XHR and fetch requests that match the given URLs. |
| `networkDetailDenyUrls` | `(string\|RegExp)[]` | `[]` | Do not capture request and response details for these URLs. Takes precedence over `networkDetailAllowUrls`. |
| `networkCaptureBodies` | `boolean` | `true` | Decide whether to capture request and response bodies for URLs defined in `networkDetailAllowUrls`. |
| `networkRequestHeaders` | `string[]` | `[]` | Additional request headers to capture for URLs defined in `networkDetailAllowUrls`. |
| `networkResponseHeaders` | `string[]` | `[]` | Additional response headers to capture for URLs defined in `networkDetailAllowUrls`. |

### `networkDetail[Allow|Deny]Urls`
- When client specifies a string => used for substring matching 
> E.g., "example.com" matches "api.example.com"
- When client specifies a regex => used for NSRegularExpression#firstMatch
- networkDetailDenyUrls takes precedence over networkDetailAllowUrls
- Empty string filtering to handle invalid input gracefully

### `network[Request|Response]Headers`
- Default headers are always extracted and can't be disabled - _"Content-Type", "Content-Length",  "Accept"_
- The header literal seen on the request is extracted (not the header name provided via SentryReplayOptions)
- header comparison is case-insensitive (see PR 5/8)

### `networkCaptureBodies`
- nm, just a boolean.

## :scroll: Description

**[PR 1/N](https://github.com/getsentry/sentry-cocoa/pull/7580). Extend SentryReplayOptions API with Session Replay Network Details configuration**
_[PR 2/N](https://github.com/getsentry/sentry-cocoa/pull/7581). Adds test app UI to configure and test network details collection.
[PR 3/N](https://github.com/getsentry/sentry-cocoa/pull/7582). Adds data holder classes to define structure of data being extracted
[PR 4/N](https://github.com/getsentry/sentry-cocoa/pull/7584). Adds new swizzling introduced to capture response bodies.
[PR 5/N](https://github.com/getsentry/sentry-cocoa/pull/7585). Implements extraction logic for headers & bodies.
[PR 6/N](https://github.com/getsentry/sentry-cocoa/pull/7588). Hook into existing SentryNetworkTracker.m|h
[PR 7/N](https://github.com/getsentry/sentry-cocoa/pull/7590). Implement conversion from breadcrumb data -> Session Replay compatible RRWebEvent_

Parent issue (android, cocoa, RN) - https://github.com/getsentry/sentry/issues/84596
Cocoa sub-issue - https://github.com/getsentry/sentry-cocoa/issues/4944

This PR adds the SDK fields required for developers to enable network details extraction for network requests made during a session replay capture, by following the impl in [sentry-javascript](https://github.com/getsentry/sentry-javascript/blob/d975bcd26955f906f0ebd67281bc88a67251a1c1/packages/replay-internal/src/types/replay.ts#L51) and [sentry-java (android)](https://github.com/getsentry/sentry-java/blob/b8bd88061259ff86c9472d203d3f17d1c106336e/sentry/src/main/java/io/sentry/SentryReplayOptions.java#L134-L150).

### Unit tests

**SentryReplayOptionsTests**
> xcodebuild test -workspace Sentry.xcworkspace -scheme Sentry -destination 'platform=iOS Simulator,name=iPhone 16 Pro' -only-testing:SentryTests/SentryReplayOptionsTests
```
(77 tests - showing last 30)

  Test Case '-[SentryTests.SentryReplayOptionsTests testInitFromDict_unmaskedViewClasses_whenInvalidArrayValues_shouldUseDefaultValue]' 
  passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testInitFromDict_unmaskedViewClasses_whenKeyOmitted_shouldUseDefaultValue]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests
  testInitFromDict_unmaskedViewClasses_whenMixedValidAndInvalidValues_shouldKeepOnlyValidValues]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testInitFromDict_unmaskedViewClasses_whenMultipleValidValue_shouldKeepAll]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testInitFromDict_unmaskedViewClasses_whenNotValidValue_shouldUseDefaultValue]' 
  passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testInitFromDict_unmaskedViewClasses_whenValidValue_shouldSetValue]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testInitFromDict_withMultipleOptions_shouldSetAllValues]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testQuality_fromName_shouldParseKnownValues]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testQuality_fromName_shouldReturnMediumForUnknownValues]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testQualityFromName]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testQualityFromName_invalidValue_shouldReturnDefaultQuality]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testQualityHigh]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testQualityLow]' passed
  Test Case '-[SentryTests.SentryReplayOptionsTests testQualityMedium]' passed
  Executed 77 tests, with 0 failures (0 unexpected)
```

**SentryReplayOptionsNetworkTests**
Additional unit tests for network details (I didn't put them in SentryReplayOptionsTests to avoid over-crowding that file / I assumed SentryReplayOptionsTests should be for high level testing of SentryReplayOptions).

> xcodebuild test -workspace Sentry.xcworkspace -scheme Sentry -destination 'platform=iOS Simulator,name=iPhone 16 Pro' -only-testing:SentryTests/SentryReplayOptionsNetworkTests 2>&1 | xcbeautify
```
 Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testIsNetworkDetailCaptureEnabled_withAllowList_shouldMatchCorrectly]' passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testIsNetworkDetailCaptureEnabled_withDenyList_shouldDenyMatchingUrls]'
  passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testIsNetworkDetailCaptureEnabled_withRegexPatterns_shouldMatchCorrectly]'
  passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testNetworkDetailUrls_withEmptyStrings_shouldFilterOutEmptyEntries]' passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testNetworkDetailUrls_withInvalidTypes_shouldFilterOutInvalidEntries]' passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testNetworkDetailUrls_withLeadingAndTrailingWhitespace_shouldTrimAndMatch]'
  passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testNetworkDetailUrls_withMixedStringAndRegexTypes_shouldMatchCorrectly]'
  passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests
  testNetworkHeaders_withCaseInsensitiveDuplicates_shouldPreventDuplicateHeaders]' passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testNetworkHeaders_withCustomHeaders_shouldAlwaysIncludeDefaultHeaders]'
  passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testNetworkHeaders_withVariousCases_shouldDeduplicateCaseInsensitively]'
  passed
  Test Case '-[SentryTests.SentryReplayOptionsNetworkTests testNetworkHeaders_withVariousConfigurations_shouldHandleCorrectly]' passed
  Executed 11 tests, with 0 failures (0 unexpected)
```

**SentryReplayOptionsObjcTests**
ObjC <> Swift compatibility tests. Mostly for Regex matching
> xcodebuild test -workspace Sentry.xcworkspace -scheme Sentry -destination 'platform=iOS 
  Simulator,name=iPhone 16 Pro' -only-testing:SentryTests/SentryReplayOptionsObjcTests
```
  Test Case '-[SentryReplayOptionsObjcTests testInit_withAllArguments_shouldSetAllValues]' passed
  Test Case '-[SentryReplayOptionsObjcTests testInit_withoutArguments_shouldUseDefaults]' passed
  Test Case '-[SentryReplayOptionsObjcTests testIsNetworkDetailCaptureEnabled_withDenyPatterns_shouldRespectDenyOverAllow]' passed
  Test Case '-[SentryReplayOptionsObjcTests testIsNetworkDetailCaptureEnabled_withMixedPatterns_shouldSupportBoth]' passed
  Test Case '-[SentryReplayOptionsObjcTests testIsNetworkDetailCaptureEnabled_withNSRegularExpression_shouldUseProvidedRegexMatching]'
  passed
  Test Case '-[SentryReplayOptionsObjcTests testIsNetworkDetailCaptureEnabled_withStringPatterns_shouldUseSubstringMatching]' passed
  Executed 6 tests, with 0 failures (0 unexpected)
```
## :pencil: Checklist

- [X] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled. **Nothing extracted in this PR - SDKOptions available but no backing implementation.** 
- [X] I updated the docs if needed. **Docs updated when CHANGELOG goes in**
- [X] I updated the wizard if needed. **No wizard updates**
- [x] Review from the native team if needed. **N/A**
- [x] No breaking change or entry added to the changelog. **future PR** #skip-changelog
- [X] No breaking change for hybrid SDKs or communicated to hybrid SDKs. **Extends API -> nothing breaking expected.**


Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -292,6 +299,131 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
 
+    /**
+     * A list of URL patterns to capture request and response details for during session replay.
+     * 
+     * When non-empty, network requests with URLs matching any of these patterns will have their
+     * headers and bodies captured for session replay.
+     * 
+     * Supports both String and NSRegularExpression patterns (See [JavaScript SDK](https://github.com/getsentry/sentry-javascript/blob/6fb1ee139a92a6055b52b0bbf5136fa0e5a9353f/packages/core/src/utils/string.ts#L114-L119)):
+     * - String: Uses substring contains
+     * - NSRegularExpression: Uses full regex matching
```
philprime: `m`: Let's not reference another SDK the user might not know nor ever use in our public  documentation.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -292,6 +299,131 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+     * Example:
+     * ```swift
+     * // String patterns (substring matching)
+     * options.sessionReplay.networkDetailAllowUrls = [
+     *     "api.example.com",              // Matches any URL containing this string
+     *     "/api/v1/",                      // Matches any URL containing this path
+     *     "https://analytics.myapp.com"   // Matches any URL containing this prefix
+     * ]
+     * 
+     * // NSRegularExpression patterns (full regex matching)
```
philprime, line 323: `m`: Any chance we can also support Swift's Regex?

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -292,6 +299,131 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+     *     "api.example.com",               // String: substring match
+     *     apiRegex!,                       // Regex: versioned API endpoints
+     *     imageRegex!                      // Regex: image files
+     * ]
+     * ```
+     *
+     * - Note: Request and response bodies are truncated to 150KB maximum.
+     * - Note: See ``SentryReplayOptions.DefaultValues.networkDetailAllowUrls`` for the default value.
+     */
+    public var networkDetailAllowUrls: [Any]
```
philprime: `m`: This might be a great opportunity to use protocol-based such as we did in [SentryMetricsApiProtocol](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Swift/Integrations/Metrics/SentryMetricsApiProtocol.swift) where the attributes are actually just the protocol [SentryAttributeValue]https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Swift/Protocol/SentryAttributeValue.swift#L46)

If not we could also consider using enums with associated values instead of type-erasing it. See these two blog posts for context:

- https://sentry.engineering/blog/building-type-safe-metrics-api-in-swift-part-i
- https://sentry.engineering/blog/building-type-safe-metrics-api-in-swift-part-ii

Might not work with Objective-C though

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -292,6 +299,131 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+     * - NSRegularExpression: Uses full regex matching
+     * 
+     * Default: empty array (no URLs explicitly denied)
+     *
+     * Examples:
+     * - String patterns: "/auth/", "/payment/", "password", ".internal."
+     * - NSRegularExpression patterns: Use try NSRegularExpression(pattern:) to create regex objects
+     * - Mixed arrays are supported with both types
+     */
+    public var networkDetailDenyUrls: [Any]
```
philprime: `m`: Same as above, should use typing if possible

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -351,6 +483,60 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+                // Filter out empty strings and whitespace-only strings
+                let trimmed = stringPattern.trimmingCharacters(in: .whitespacesAndNewlines)
+                guard !trimmed.isEmpty else { continue }
+                
+                if urlString.contains(stringPattern) {
+                    return true
+                }
+            } else if let regexPattern = pattern as? NSRegularExpression {
+                // NSRegularExpression: use regex matching
+                let range = NSRange(location: 0, length: urlString.utf16.count)
```
philprime: `l`: Specific reason to use the utf16 version?

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -409,7 +600,12 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
             sessionSegmentDuration: (dictionary["sessionSegmentDuration"] as? NSNumber)?.doubleValue,
             maximumDuration: (dictionary["maximumDuration"] as? NSNumber)?.doubleValue,
             excludedViewClasses: (dictionary["excludedViewClasses"] as? [String]).map { Set($0) },
-            includedViewClasses: (dictionary["includedViewClasses"] as? [String]).map { Set($0) }
+            includedViewClasses: (dictionary["includedViewClasses"] as? [String]).map { Set($0) },
+            networkDetailAllowUrls: Self.validateNetworkDetailUrlPatterns(from: dictionary["networkDetailAllowUrls"]),
+            networkDetailDenyUrls: Self.validateNetworkDetailUrlPatterns(from: dictionary["networkDetailDenyUrls"]),
```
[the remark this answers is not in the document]
philprime: This seems to be a hallucinations because it's checking two different dictionaries. please double-check

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -456,10 +652,81 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+    /**
+     * Helper method to parse and filter string arrays from dictionary configuration.
+     * 
+     * Filters out non-string entries from mixed arrays while preserving valid strings.
+     * Returns nil when the input is not an array type, allowing callers to fall back to defaults.
+     * 
+     * - Parameter value: The value from the dictionary to parse
+     * - Returns: Filtered array of strings, or nil if input is not an array
+     */
+    private static func parseStringArray(from value: Any?) -> [String]? {
```
philprime: `m`: Could live in a helper utility class rather than in this here.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -456,10 +652,81 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+        var validPatterns: [Any] = []
+        var invalidCount = 0
+        
+        for (index, element) in array.enumerated() {
+            if let stringElement = element as? String {
+                // Filter out empty strings and whitespace-only strings
+                let trimmed = stringElement.trimmingCharacters(in: .whitespacesAndNewlines)
+                if trimmed.isEmpty {
+                    SentrySDKLog.log(message: "Invalid networkDetail URL pattern at index \(index): empty or whitespace-only string discarded", 
+                                   andLevel: .warning)
```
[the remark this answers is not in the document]
philprime: `m`: This seems to be valid feedback

philprime (commented):
Good progress, almost LGTM

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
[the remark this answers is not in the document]
43jay: fixed

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: removed

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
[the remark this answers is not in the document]
43jay: was legit, should be resolved now (changed a bunch of this to introduce the protocol type thingy)

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: removed, and elsewhre

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: short answer - yes. Longer answer - let's add it later? 

Swift's `Regex` seems 1000x better than NSRegularExpression... and once the API changes (SentryUrlMatchable protocol) get merged it should be easy to add support for Swift Regex.

I could do it now if you think it's less likely to happen later ... but thinking there's still 7 more PRs to go 😂

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: Great suggestion. i took a crack at it. [PTAL](https://github.com/getsentry/sentry-cocoa/pull/7580/changes/cfc450ac0e89953f32a85b025bf0a5ac3fef1766)

> Might not work with Objective-C though

The only oddity i needed was introducing an objC getter for networkDetailAllowUrls and networkDetailDenyUrls ... an oddity b/c the getters are currently only called from `SentryReplayOptionsObjcTests` so it's odd to have them hanging around on the SentryReplayOptions.swift API (and i couldn't see a way to hide them).

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: done

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: nice catch - no, I think this was a bug / weird code. Replaced by simpler impl [here](https://github.com/getsentry/sentry-cocoa/blob/cfc450ac0e89953f32a85b025bf0a5ac3fef1766/Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift#L532)

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: Yes, we can do this in a follow-up PR

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -292,6 +299,145 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+     * options.sessionReplay.networkRequestHeaders = [
+     *     "Authorization",
+     *     "User-Agent"
+     * ]
+     * ```
+     *
+     * - Note: This setting only applies when ``networkDetailAllowUrls`` is non-empty.
+     * - Note: Header names preserve the case seen on the request, not the case specified here.
+     */
+    public var networkRequestHeaders: [String] {
```
itaybre, line 395: `m`" This API won't be available for ObjC?

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -292,6 +299,145 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+     * options.sessionReplay.networkResponseHeaders = [
+     *     "Cache-Control",    // Custom header
+     *     "Set-Cookie"        // Custom header
+     * ]
+     * ```
+     *
+     * - Note: This setting only applies when ``networkDetailAllowUrls`` is non-empty.
+     * - Note: Header names preserve the case seen on the response, not the case specified here.
+     */
+    public var networkResponseHeaders: [String] {
```
itaybre, line 421: `m`: Same question as above, not available for ObjC?

itaybre (commented):
Almost LGTM, just some small comments

And you will need to run `make generate-public-api` due to the new public APIs

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: I did not ( / did not mean to) exclude them from objC. why do you bring it up?

IIUC objC users should be able to do 

```objc
@import Sentry;

- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {

    [SentrySDK startWithConfigureOptions:^(SentryOptions *options) {
        ...
       options.sessionReplay.networkRequestHeaders = @[@"custom-req-header"];
    }];

    return YES;
}
```

---

Unless you're referring to the `[SentryUrlMatchable]` types 
 (i.e. not the `[String] networkRequestHeaders /networkResponseHeaders)` ?

```swift
public static let networkDetailAllowUrls: [SentryUrlMatchable] = []
        public static let networkDetailDenyUrls: [SentryUrlMatchable] = []
```
My understanding is this type would also be settable in objC via code like

```objc
[SentrySDK startWithConfigureOptions:^(SentryOptions *options) {
      ...
      NSError *error = nil;
      NSRegularExpression *regex = [NSRegularExpression
  regularExpressionWithPattern:@"^https://api\\.example\\.com/.*" options:0 error:&error];

      options.sessionReplay.networkDetailAllowUrls = @[
          @"https://myapi.com",
          regex
      ];
  }];
}
```

I then added two custom getters (`networkDetailDenyUrlsForObjC:``networkDetailAllowUrlsForObjC:`) to querying the value of these fields. I commented they are "not part of the public API" cuz they only exist so `SentryReplayOptionsObjcTests.m` can assert they are being set correctly.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: [convo above](https://github.com/getsentry/sentry-cocoa/pull/7580#discussion_r2907217134)

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: Ah, you _are_ referring to `[networkDetailAllow|DenyUrls]: [SentryUrlMatchable]`

... and in fact you _are_ right that there is no objC setter. I misunderstood the bridging.
So this code (from above) does _not_ work:

```
options.sessionReplay.networkDetailAllowUrls = @[
          @"https://myapi.com",
          regex
      ];
```

There's no setter exposed because [SentryUrlMatchable] can't be represented in ObjC.

**Q=> What do you suggest we do in this case?** I added [SentryUrlMatchable] b/c it _is_ the nicer API for swift users. And it makes it trivial to add Swift's `Regex` syntax in a later PR.

option 1) revert `[networkDetailAllow|DenyUrls]: [SentryUrlMatchable]` to `[networkDetailAllow|DenyUrls]: [String]` 

option 2) Add explicit setter to match the existing getters to `networkDetailDenyUrlsForObjC`

In absence of your input / until you chime in, **i'm gonna go option 2) and add the objC setters.** =>

Based on my time reading the [sentry-ios docs](https://docs.sentry.io/platforms/apple/guides/ios/session-replay/) there's not a lot of objC sample code => we want to be nice to swift users and "provide a path" for objC users.
the `@objc(networkDetailAllowUrls)` annotation actually does the right thing and shows objC users a property called `networkDetailAllowUrls` (not `networkDetailAllowUrlsForObjC`

itaybre: I would sync with @philprime since he is adding an ObjC SDK that may be able to cover this: https://linear.app/getsentry/issue/COCOA-762/create-an-objc-sdk

43jay: ok will get his input thx.

FWIW i have amended option 2 into [this commit](https://github.com/getsentry/sentry-cocoa/pull/7580/changes/efa97aae56c02030358893637233f403c9ca07d3). 


It doesn't seem at odds w/ COCOA-762. The [@objc bridge code added](https://github.com/getsentry/sentry-cocoa/blob/efa97aae56c02030358893637233f403c9ca07d3/Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift#L358-L360) should be easy to refactor into this objC SDK (or have the objC SDK rely on the setters) when it's stood up.

43jay:
> Almost LGTM, just some small comments
> 
> And you will need to run `make generate-public-api` due to the new public APIs

@itaybre ran make generate-public api [here](https://github.com/getsentry/sentry-cocoa/pull/7580/changes/037f0e4f86d4d21e13464451c436d8ed398b4a95)

... seems like it picked up a bunch of arbitrary stuff (not sure, didn't look too hard). P sure i'll need to rebase and run it again

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
philprime: The point of the ObjC SDK is that it wraps and exposes Swift-only API endpoints as ObjC code. You can see it in https://github.com/getsentry/sentry-cocoa/pull/7598/changes

We can keep it as-is so it works also for the base Sentry SDK, but we will remove it in a future major release

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -495,8 +730,64 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+        return result
+    }
+
+    // MARK: - ObjC Bridge Helpers
+
+    /// Converts an NSArray from ObjC into [SentryUrlMatchable].
+    /// Accepts NSString and NSRegularExpression elements; unsupported types are silently dropped.
+    private static func convertToUrlMatchables(_ array: NSArray) -> [SentryUrlMatchable] {
+        return array.compactMap { element -> SentryUrlMatchable? in
+            if let string = element as? String {
```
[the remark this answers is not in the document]
philprime: @43jay this is valid feedback, unless empty strings have a special meaning.

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -495,8 +730,64 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+                return string
+            } else if let regex = element as? NSRegularExpression {
+                return regex
+            }
+            return nil
+        }
+    }
```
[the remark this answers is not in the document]
philprime: Same here

philprime (commented):
Almost LGTM

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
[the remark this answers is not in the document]
43jay: dup of [comment above](https://github.com/getsentry/sentry-cocoa/pull/7580#discussion_r2907598517)

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
43jay: resolution: leave the objC getters in this (Swift) SDK for now; after objC SDK is stood up they will be moved out

Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift
```diff
@@ -292,6 +299,170 @@ public class SentryReplayOptions: NSObject, SentryRedactOptions {
...
+     * - Note: Request and response bodies are truncated to 150KB maximum.
+     * - Note: See ``SentryReplayOptions.DefaultValues.networkDetailAllowUrls`` for the default value.
+     */
+    public var networkDetailAllowUrls: [SentryUrlMatchable]
+    
+    /// Objective-C bridge for networkDetailAllowUrls.
+    /// - Warning: This property exists for Objective-C compatibility only. Swift code should use
+    ///            `networkDetailAllowUrls` directly.
+    @objc(networkDetailAllowUrls) public var networkDetailAllowUrlsForObjC: NSArray {
+        get { return networkDetailAllowUrls as NSArray }
```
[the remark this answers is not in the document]
philprime: @43jay I would like to propose that we do not support this addition in Objective-C for now, and defer the support to https://github.com/getsentry/sentry-cocoa/pull/7598 as this seems to be brittle and we do not want SDK crashes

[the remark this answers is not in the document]
43jay: hmm. If that's your inclination I will remove the obj-C support. 

Assuming I can accomplish simply by removing this bridge function => any obj-C clients will simply not see a `networkDetailAllow|DenyUrls` field.
> @objc(networkDetailAllowUrls) public var networkDetailAllowUrlsForObjC: NSArray

and not do anything to hide the other fields: these will still show up for obj-C  clients - (`networkRequest|ResponseHeaders`, `networknetworkCaptureBodies`)


---

> I will remove the obj-C support

THO... I suspect this bug sentry-bot flagged is not a real issue or easily remedied - [this test](https://github.com/getsentry/sentry-cocoa/blob/cad52c2683400286020c3bd8e88564deaca0c5e8/Tests/SentryTests/Integrations/SessionReplay/SentryReplayOptionsObjcTests.m#L262-L281) in SentryReplayOptionsObjcTests is successfully writing then reading the field using the bridge fn in question.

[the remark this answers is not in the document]
43jay: removed [here](https://github.com/getsentry/sentry-cocoa/pull/7580/changes/4a60ed48b6dff342d418cf6d1b973ceae7ea51f3)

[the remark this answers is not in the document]
philprime: Yes they will not see it, as they also do not see the Metrics API because it's Swift-only. This is the entire reason why we are building an Objective-C wrapper SDK to begin with.

philprime (approved):
LGTM

43jay:
rebase to land ✅

43jay:
rebase onto main

PR #7582, opened 2026-03-03
feat(network-details): Introduce data classes for extracting network details to session replay

43jay:
## :scroll: Description

**Class diagram**
```
SentryReplayNetworkDetails.swift     (@objc, @_spi(Private) public)
├── method: String?
├── statusCode: NSNumber?
├── requestBodySize: NSNumber?       (computed from request.size)
├── responseBodySize: NSNumber?      (computed from response.size)
├── request: Detail?
├── response: Detail?
│
├── Detail                           (Swift-only struct)
│   ├── size: NSNumber?
│   ├── body: Body?
│   └── headers: [String: String]
│
├── Body                             (Swift-only struct)
│   ├── content: BodyContent
│   └── warnings: [NetworkBodyWarning]
│
├── BodyContent                      (Swift-only enum)
│   ├── .json(Any)                   — parsed dict/array
│   └── .text(String)                — plain text or placeholder
│
└── NetworkBodyWarning               (Swift-only enum)
    ├── .jsonTruncated
    ├── .textTruncated
    ├── .invalidJson
    └── .bodyParseError
```

**Expected usage**  
```
SentryNetworkTracker.m (ObjC — producer)
  │
  │  let details = SentryReplayNetworkDetails(method: "POST")
  │  [details setRequestWithSize:… body:… headers:…]
  │  [details setResponseWithStatusCode:… size:… body:… headers:…]
  │  breadcrumbData["_networkDetails"] = details
  │
  ▼
SentrySRDefaultBreadcrumbConverter.swift (Swift — consumer)
  │
  │  let details = breadcrumb.data["_networkDetails"] as? SentryReplayNetworkDetails
  │  let serialized = details.serialize()
  │  // → { method, statusCode, requestBodySize, responseBodySize,
  │  //     request: { size, headers, body: { body, warnings } },
  │  //     response: { size, headers, body: { body, warnings } } }
  │
  ▼
  RRWebSpanEvent data payload
```

## :bulb: Motivation and Context

See corresponding [class definitions in sentry-java](https://github.com/getsentry/sentry-java/tree/b8bd88061259ff86c9472d203d3f17d1c106336e/sentry/src/main/java/io/sentry/util/network)

This PR only defines the data contract used by PRs higher in the stack.

See [first PR](https://github.com/getsentry/sentry-cocoa/pull/7580) for more motivation/context.

## :green_heart: How did you test it?
Unit tests are in the next PR in the stack, which introduces the body parsing and header extraction logic that operates on these types.

[...]

PR #7584, opened 2026-03-03
feat(network-details): New swizzling to capture response bodies for session replay

43jay:
## :scroll: Description

- Newly swizzled methods: 
`[NSURLSession dataTaskWithURL:completionHandler:]`
`[NSURLSession dataTaskWithRequest:completionHandler:]`

- This swizzling is [skipped](https://github.com/getsentry/sentry-cocoa/pull/7584/changes/22e2b20fdaf426597538da8c1bafc9200d28791f) when network detail capture is not enabled (SentryReplayOptions#networkDetailAllowUrls is empty)

- Added testing to validate for future iOS SDKs that the `IMP`s we are swizzling are implemented directly on NSURLSession: _SentryNSURLSessionTaskSearchTests_
    - `test_URLSessionDataTaskWithRequest_ByIosVersion`
    - `test_URLSessionDataTaskWithURL_ByIosVersion`

With this change sentry-cocoa can inspect response bodies of NSURLSession `dataTask`'s that use completionHandlers.

Swizzling is added to inspect the NSURLResponse before delegating to the original completionHandler. 
The request body will be captured via existing swizzling into SentryNetworkTracker (`setState`, `resume`).

See [swizzling discussion on #7582](https://github.com/getsentry/sentry-cocoa/issues/4944#issuecomment-3856073897) for more context.

```
test_URLSessionTask_ByIosVersion ✅
test_URLSessionDataTaskWithRequest_ByIosVersion ✅
test_URLSessionDataTaskWithURL_ByIosVersion ✅
```

>  SKIP_SWIFTLINT=1 xcodebuild test -workspace Sentry.xcworkspace -scheme Sentry -destination 'platform=iOS Simulator,name=iPhone 16 Pro'  -only-testing:SentryTests/SentrySwizzleWrapperHelperTests

> make test-ios ONLY_TESTING=SentryNetworkDetailSwizzlingTests 2>&1 | xcbeautify
> ...
> Test Suite 'SentryNetworkDetailSwizzlingTests' started at 2026-04-13 13:55:14.423.
> Executed 2 tests, with 0 failures (0 unexpected) in 2.100 (2.100) seconds

## :pencil: Checklist

- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled. **gated by networkDetailAllowUrls**
- [x] I updated the docs if needed. **future pr**
- [x] I updated the wizard if needed. **N/A**
- [x] No breaking change or entry added to the changelog. #skip-changelog **future PR**
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.  🤔

### DIFF
--- a/Sources/Sentry/SentryNetworkTracker.m
+++ b/Sources/Sentry/SentryNetworkTracker.m
@@ -562,4 +562,15 @@ static NSString *const SentryNetworkTrackerThreadSanitizerMessage
     return breadcrumbLevel;
 }
 
+- (void)captureResponseDetails:(NSData *)data
+                      response:(NSURLResponse *)response
+                    requestURL:(NSURL *)requestURL
+                          task:(NSURLSessionTask *)task
+{
+    // TODO: Implementation
+    // 2. Parse response body data
+    // 3. Store in appropriate location for session replay
+    // 4. Handle size limits and truncation if needed
+}
+
 @end
--- a/Sources/Sentry/SentrySwizzleWrapperHelper.m
+++ b/Sources/Sentry/SentrySwizzleWrapperHelper.m
@@ -131,9 +131,9 @@ NS_ASSUME_NONNULL_BEGIN
                 wrappedHandler = ^(NSData *data, NSURLResponse *response, NSError *error) {
                     if (!error && data && task) {
                         [networkTracker captureResponseDetails:data
-                                                     response:response
-                                                   requestURL:request.URL
-                                                         task:task];
+                                                      response:response
+                                                    requestURL:request.URL
+                                                          task:task];
                     }
                     completionHandler(data, response, error);
                 };
@@ -164,9 +164,9 @@ NS_ASSUME_NONNULL_BEGIN
                 wrappedHandler = ^(NSData *data, NSURLResponse *response, NSError *error) {
                     if (!error && data && task) {
                         [networkTracker captureResponseDetails:data
-                                                     response:response
-                                                   requestURL:url
-                                                         task:task];
+                                                      response:response
+                                                    requestURL:url
+                                                          task:task];
                     }
                     completionHandler(data, response, error);
                 };
--- a/Sources/Sentry/SentrySwizzleWrapperHelper.m
+++ b/Sources/Sentry/SentrySwizzleWrapperHelper.m
@@ -97,6 +97,87 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma clang diagnostic pop
 }
 
+/**
+ * Swizzles NSURLSession data task creation methods that use completion handlers
+ * to enable response body capture for session replay.
+ *
+ * Both dataTaskWithRequest: and dataTaskWithURL: are independent implementations
+ * (neither calls through to the other), so both need swizzling.
+ *
+ * See SentryNSURLSessionTaskSearchTests that verifies these assumptions still hold.
+ */
++ (void)swizzleURLSessionDataTasksForResponseCapture:(SentryNetworkTracker *)networkTracker
+{
+    [self swizzleDataTaskWithRequestCompletionHandler:networkTracker];
+    [self swizzleDataTaskWithURLCompletionHandler:networkTracker];
+}
+
+/**
+ * Swizzles -[NSURLSession dataTaskWithRequest:completionHandler:] to intercept response data.
+ */
++ (void)swizzleDataTaskWithRequestCompletionHandler:(SentryNetworkTracker *)networkTracker
+{
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshadow"
+    SEL selector = @selector(dataTaskWithRequest:completionHandler:);
+    SentrySwizzleInstanceMethod([NSURLSession class], selector,
+        SentrySWReturnType(NSURLSessionDataTask *),
+        SentrySWArguments(NSURLRequest * request,
+            void (^completionHandler)(NSData *, NSURLResponse *, NSError *)),
+        SentrySWReplacement({
+            __block NSURLSessionDataTask *task = nil;
+            void (^wrappedHandler)(NSData *, NSURLResponse *, NSError *) = nil;
+            if (completionHandler) {
+                wrappedHandler = ^(NSData *data, NSURLResponse *response, NSError *error) {
+                    if (!error && data && task) {
+                        [networkTracker captureResponseDetails:data
+                                                     response:response
+                                                   requestURL:request.URL
+                                                         task:task];
+                    }
+                    completionHandler(data, response, error);
+                };
+            }
+            task = SentrySWCallOriginal(request, wrappedHandler ?: completionHandler);
+            return task;
+        }),
+        SentrySwizzleModeOncePerClassAndSuperclasses, (void *)selector);
+#pragma clang diagnostic pop
+}
+
+/**
+ * Swizzles -[NSURLSession dataTaskWithURL:completionHandler:] to intercept response data.
+ */
++ (void)swizzleDataTaskWithURLCompletionHandler:(SentryNetworkTracker *)networkTracker
+{
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshadow"
+    SEL selector = @selector(dataTaskWithURL:completionHandler:);
+    SentrySwizzleInstanceMethod([NSURLSession class], selector,
+        SentrySWReturnType(NSURLSessionDataTask *),
+        SentrySWArguments(
+            NSURL * url, void (^completionHandler)(NSData *, NSURLResponse *, NSError *)),
+        SentrySWReplacement({
+            __block NSURLSessionDataTask *task = nil;
+            void (^wrappedHandler)(NSData *, NSURLResponse *, NSError *) = nil;
+            if (completionHandler) {
+                wrappedHandler = ^(NSData *data, NSURLResponse *response, NSError *error) {
+                    if (!error && data && task) {
+                        [networkTracker captureResponseDetails:data
+                                                     response:response
+                                                   requestURL:url
+                                                         task:task];
+                    }
+                    completionHandler(data, response, error);
+                };
+            }
+            task = SentrySWCallOriginal(url, wrappedHandler ?: completionHandler);
+            return task;
+        }),
+        SentrySwizzleModeOncePerClassAndSuperclasses, (void *)selector);
+#pragma clang diagnostic pop
+}
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Sources/Swift/Integrations/Performance/Network/SentryNetworkTrackingIntegration.swift
+++ b/Sources/Swift/Integrations/Performance/Network/SentryNetworkTrackingIntegration.swift
@@ -42,7 +42,7 @@ final class SentryNetworkTrackingIntegration<Dependencies: NetworkTrackerProvide
 
         SentrySwizzleWrapperHelper.swizzleURLSessionTask(networkTracker)
 
-        #if os(iOS) || os(tvOS)
+        #if (os(iOS) || os(tvOS)) && !SENTRY_NO_UI_FRAMEWORK
          if options.sessionReplay.networkDetailHasUrls {
              SentrySwizzleWrapperHelper.swizzleURLSessionDataTasks(forResponseCapture: networkTracker)
          }
--- a/Sources/Swift/Integrations/Performance/Network/SentryNetworkTrackingIntegration.swift
+++ b/Sources/Swift/Integrations/Performance/Network/SentryNetworkTrackingIntegration.swift
@@ -41,6 +41,12 @@ final class SentryNetworkTrackingIntegration<Dependencies: NetworkTrackerProvide
         super.init()
 
         SentrySwizzleWrapperHelper.swizzleURLSessionTask(networkTracker)
+
+        #if os(iOS) || os(tvOS)
+         if options.sessionReplay.networkDetailHasUrls {
+             SentrySwizzleWrapperHelper.swizzleURLSessionDataTasks(forResponseCapture: networkTracker)
+         }
+        #endif
     }
 
     func uninstall() {
--- a/Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
+++ b/Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
@@ -1,13 +1,156 @@
 import XCTest
 
+// We need to know whether Apple changes the NSURLSessionTask implementation.
 class SentryNSURLSessionTaskSearchTests: XCTestCase {
 
-    // We need to know whether Apple changes the NSURLSessionTask implementation.
-     func test_URLSessionTask_ByIosVersion() {
+    func test_URLSessionTask_ByIosVersion() {
         let classes = SentryNSURLSessionTaskSearch.urlSessionTaskClassesToTrack()
-        
+
         XCTAssertEqual(classes.count, 1)
         XCTAssertTrue(classes.first === URLSessionTask.self)
     }
 
+    // MARK: - NSURLSession class hierarchy validation tests
+    //
+    // Based on testing, NSURLSession implements dataTaskWithRequest:completionHandler:
+    // and dataTaskWithURL:completionHandler: directly on the base class.
+    //
+    // The swizzling code relies on this by swizzling [NSURLSession class] directly
+    // rather than doing runtime discovery. These tests verify that assumption
+    // still holds — if Apple ever moves these methods to a subclass, these tests
+    // will fail and we'll know to update the swizzling approach.
+
+    func test_URLSession_isNotClassCluster_dataTaskWithRequest() {
+        let selector = #selector(URLSession.dataTask(with:completionHandler:)
+            as (URLSession) -> (URLRequest, @escaping @Sendable (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask)
+        assertNSURLSessionImplementsDirectly(selector: selector, selectorName: "dataTaskWithRequest:completionHandler:")
+    }
+
+    func test_URLSession_isNotClassCluster_dataTaskWithURL() {
+        let selector = #selector(URLSession.dataTask(with:completionHandler:)
+            as (URLSession) -> (URL, @escaping @Sendable (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask)
+        assertNSURLSessionImplementsDirectly(selector: selector, selectorName: "dataTaskWithURL:completionHandler:")
+    }
+
+    // MARK: - dataTaskWithURL: / dataTaskWithRequest: independence
+    //
+    // We swizzle both dataTaskWithRequest:completionHandler: and
+    // dataTaskWithURL:completionHandler: because they are independent
+    // implementations — dataTaskWithURL: does NOT dispatch to
+    // dataTaskWithRequest: via objc_msgSend.
+    //
+    // If this test ever fails, Apple has changed the internal dispatch so
+    // one calls through to the other. In that case, remove the redundant
+    // swizzle and add a deduplication guard to avoid double-capture.
+
+    func test_dataTaskWithURL_doesNotCallThrough_dataTaskWithRequest() {
+        assertNoCallThrough(
+            from: NSSelectorFromString("dataTaskWithURL:completionHandler:"),
+            to: NSSelectorFromString("dataTaskWithRequest:completionHandler:"),
+            call: { session in
+                let url = URL(string: "https://example.com")!
+                let task = session.dataTask(with: url) { _, _, _ in }
+                task.cancel()
+            }
+        )
+    }
+
+    func test_dataTaskWithRequest_doesNotCallThrough_dataTaskWithURL() {
+        assertNoCallThrough(
+            from: NSSelectorFromString("dataTaskWithRequest:completionHandler:"),
+            to: NSSelectorFromString("dataTaskWithURL:completionHandler:"),
+            call: { session in
+                let request = URLRequest(url: URL(string: "https://example.com")!)
+                let task = session.dataTask(with: request) { _, _, _ in }
+                task.cancel()
+            }
+        )
+    }
+
+    /// Temporarily replaces the IMP of `targetSelector` with one that increments
+    /// a counter, then invokes `call` (which should trigger `sourceSelector`).
+    /// Asserts the counter stays at 0 — meaning `sourceSelector` does not
+    /// internally dispatch to `targetSelector` via objc_msgSend.
+    private func assertNoCallThrough(
+        from sourceSelector: Selector,
+        to targetSelector: Selector,
+        call: (URLSession) -> Void
+    ) {
+        guard let method = class_getInstanceMethod(URLSession.self, targetSelector) else {
+            XCTFail("URLSession should implement \(targetSelector)")
+            return
+        }
+
+        let originalIMP = method_getImplementation(method)
+        defer { method_setImplementation(method, originalIMP) }
+
+        var hitCount = 0
+
+        let replacementBlock: @convention(block) (NSObject, AnyObject, Any?) -> AnyObject = { obj, arg, handler in
+            hitCount += 1
+            typealias Fn = @convention(c) (NSObject, Selector, AnyObject, Any?) -> AnyObject
+            let original = unsafeBitCast(originalIMP, to: Fn.self)
+            return original(obj, targetSelector, arg, handler)
+        }
+
+        method_setImplementation(method, imp_implementationWithBlock(replacementBlock))
+
+        let session = URLSession(configuration: .ephemeral)
+        defer { session.invalidateAndCancel() }
+
+        call(session)
+
+        XCTAssertEqual(
+            hitCount, 0,
+            "\(sourceSelector) called through to \(targetSelector). "
+            + "These methods are no longer independent — remove the redundant swizzle "
+            + "in SentrySwizzleWrapperHelper and add a deduplication guard."
+        )
+    }
+
+    // MARK: - Helper
+
+    /// Walks the class hierarchy for sessions created with default and ephemeral
+    /// configurations and asserts that no subclass overrides `selector`.
+    private func assertNSURLSessionImplementsDirectly(selector: Selector, selectorName: String) {
+        let baseClass: AnyClass = URLSession.self
+
+        // The base class must implement the method.
+        XCTAssertNotNil(
+            class_getInstanceMethod(baseClass, selector),
+            "URLSession should implement \(selectorName)"
+        )
+
+        // Check sessions created with each relevant configuration.
+        let configs: [URLSessionConfiguration] = [
+            .default,
+            .ephemeral
+        ]
+
+        for config in configs {
+            let session = URLSession(configuration: config)
+            let sessionClass: AnyClass = type(of: session)
+
+            defer { session.invalidateAndCancel() }
+
+            if sessionClass === baseClass {
+                continue
+            }
+
+            // If Apple returns a subclass, it must NOT provide its own
+            // implementation — it should inherit from URLSession.
+            let subMethod = class_getInstanceMethod(sessionClass, selector)
+            let baseMethod = class_getInstanceMethod(baseClass, selector)
+
+            if let subMethod, let baseMethod {
+                let subIMP = method_getImplementation(subMethod)
+                let baseIMP = method_getImplementation(baseMethod)
+                XCTAssertEqual(
+                    subIMP, baseIMP,
+                    "\(NSStringFromClass(sessionClass)) overrides \(selectorName) with an unexpected IMP — "
+                    + "Verify swizzling in SentrySwizzleWrapperHelper is correct for dataTasks."
+                )
+            }
+        }
+    }
 }
--- a/Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
+++ b/Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
@@ -17,95 +17,19 @@ class SentryNSURLSessionTaskSearchTests: XCTestCase {
     //
     // The swizzling code relies on this by swizzling [NSURLSession class] directly
     // rather than doing runtime discovery. These tests verify that assumption
-    // still holds — if Apple ever moves these methods to a subclass, these tests
+    // still holds — if Apple ever moves these methods, these tests
     // will fail and we'll know to update the swizzling approach.
 
-    func test_URLSession_isNotClassCluster_dataTaskWithRequest() {
+    func test_URLSessionDataTaskWithRequest_ByIosVersion() {
         let selector = #selector(URLSession.dataTask(with:completionHandler:)
             as (URLSession) -> (URLRequest, @escaping @Sendable (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask)
         assertNSURLSessionImplementsDirectly(selector: selector, selectorName: "dataTaskWithRequest:completionHandler:")
     }
 
-    func test_URLSession_isNotClassCluster_dataTaskWithURL() {
+    func test_URLSessionDataTaskWithURL_ByIosVersion() {
         let selector = #selector(URLSession.dataTask(with:completionHandler:)
             as (URLSession) -> (URL, @escaping @Sendable (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask)
         assertNSURLSessionImplementsDirectly(selector: selector, selectorName: "dataTaskWithURL:completionHandler:")
-    }
-
-    // MARK: - dataTaskWithURL: / dataTaskWithRequest: independence
-    //
-    // We swizzle both dataTaskWithRequest:completionHandler: and
-    // dataTaskWithURL:completionHandler: because they are independent
-    // implementations — dataTaskWithURL: does NOT dispatch to
-    // dataTaskWithRequest: via objc_msgSend.
-    //
-    // If this test ever fails, Apple has changed the internal dispatch so
-    // one calls through to the other. In that case, remove the redundant
-    // swizzle and add a deduplication guard to avoid double-capture.
-
-    func test_dataTaskWithURL_doesNotCallThrough_dataTaskWithRequest() {
-        assertNoCallThrough(
-            from: NSSelectorFromString("dataTaskWithURL:completionHandler:"),
-            to: NSSelectorFromString("dataTaskWithRequest:completionHandler:"),
-            call: { session in
-                let url = URL(string: "https://example.com")!
-                let task = session.dataTask(with: url) { _, _, _ in }
-                task.cancel()
-            }
-        )
-    }
-
-    func test_dataTaskWithRequest_doesNotCallThrough_dataTaskWithURL() {
-        assertNoCallThrough(
-            from: NSSelectorFromString("dataTaskWithRequest:completionHandler:"),
-            to: NSSelectorFromString("dataTaskWithURL:completionHandler:"),
-            call: { session in
-                let request = URLRequest(url: URL(string: "https://example.com")!)
-                let task = session.dataTask(with: request) { _, _, _ in }
-                task.cancel()
-            }
-        )
-    }
-
-    /// Temporarily replaces the IMP of `targetSelector` with one that increments
-    /// a counter, then invokes `call` (which should trigger `sourceSelector`).
-    /// Asserts the counter stays at 0 — meaning `sourceSelector` does not
-    /// internally dispatch to `targetSelector` via objc_msgSend.
-    private func assertNoCallThrough(
-        from sourceSelector: Selector,
-        to targetSelector: Selector,
-        call: (URLSession) -> Void
-    ) {
-        guard let method = class_getInstanceMethod(URLSession.self, targetSelector) else {
-            XCTFail("URLSession should implement \(targetSelector)")
-            return
-        }
-
-        let originalIMP = method_getImplementation(method)
-        defer { method_setImplementation(method, originalIMP) }
-
-        var hitCount = 0
-
-        let replacementBlock: @convention(block) (NSObject, AnyObject, Any?) -> AnyObject = { obj, arg, handler in
-            hitCount += 1
-            typealias Fn = @convention(c) (NSObject, Selector, AnyObject, Any?) -> AnyObject
-            let original = unsafeBitCast(originalIMP, to: Fn.self)
-            return original(obj, targetSelector, arg, handler)
-        }
-
-        method_setImplementation(method, imp_implementationWithBlock(replacementBlock))
-
-        let session = URLSession(configuration: .ephemeral)
-        defer { session.invalidateAndCancel() }
-
-        call(session)
-
-        XCTAssertEqual(
-            hitCount, 0,
-            "\(sourceSelector) called through to \(targetSelector). "
-            + "These methods are no longer independent — remove the redundant swizzle "
-            + "in SentrySwizzleWrapperHelper and add a deduplication guard."
-        )
     }
 
     // MARK: - Helper
--- a/Sources/Sentry/include/SentryNetworkTracker.h
+++ b/Sources/Sentry/include/SentryNetworkTracker.h
@@ -26,6 +26,11 @@ static NSString *const SENTRY_NETWORK_REQUEST_TRACKER_BREADCRUMB
 @property (nonatomic, readonly) BOOL isCaptureFailedRequestsEnabled;
 @property (nonatomic, readonly) BOOL isGraphQLOperationTrackingEnabled;
 
+- (void)captureResponseDetails:(NSData *)data
+                      response:(NSURLResponse *)response
+                    requestURL:(nullable NSURL *)requestURL
+                          task:(NSURLSessionTask *)task;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Sources/Sentry/include/SentrySwizzleWrapperHelper.h
+++ b/Sources/Sentry/include/SentrySwizzleWrapperHelper.h
@@ -26,6 +26,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker;
 
+// Swizzle [NSURLSession dataTaskWithURL:completionHandler:]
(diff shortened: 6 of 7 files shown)


Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +99,75 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
+                            [networkTracker captureResponseDetails:data
+                                                          response:response
+                                                           request:request
+                                                              task:task];
+                        }
+                        completionHandler(data, response, error);
+                    };
+                }
+
+                return SentrySWCallOriginal(request, wrappedHandler ?: completionHandler);
```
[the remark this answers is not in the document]
43jay: nice. lost this in a merge resolution

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +99,75 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
+
+                void (^wrappedHandler)(NSData *, NSURLResponse *, NSError *) = nil;
+                if (completionHandler) {
+                    wrappedHandler = ^(NSData *data, NSURLResponse *response, NSError *error) {
+                        if (!error && data && task) {
+                            [networkTracker captureResponseDetails:data
+                                                          response:response
+                                                           request:request
+                                                              task:task];
+                        }
```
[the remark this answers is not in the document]
43jay: fixed

Sources/Sentry/SentrySwizzleWrapperHelper.m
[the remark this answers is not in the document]
43jay: fixed

Sources/Sentry/SentryNSURLSessionSwizzleTargetDiscovery.m
```diff
@@ -70,4 +70,65 @@ @implementation SentryNSURLSessionTaskSearch
...
+
+    Method superMethod = class_getInstanceMethod(superClass, selector);
+    if (!superMethod)
+        return YES;
+
+    IMP classIMP = method_getImplementation(method);
+    IMP superIMP = method_getImplementation(superMethod);
+
+    return classIMP != superIMP;
+}
```
philippsuess: Post edit. This comment might get obsolete with addressing https://github.com/getsentry/sentry-cocoa/pull/7584#discussion_r2883283138.

`h`: Isn't this very similar to what we already have in `urlSessionTaskClassesToTrack`. If yes, we already have smilar bullet proof code in `urlSessionTaskClassesToTrack`. I would really love to use the same approach that we already have for `urlSessionTaskClassesToTrack`. I still remember we had some really weird issues and extremely bad, hard-to-track-down bugs with this kind of thing before. Duplicating makes sense to minimize the potential damage, but we should use a similar approach as much as we can.

https://github.com/getsentry/sentry-cocoa/blob/164f67b9be5ca9d90c639125e7513f0c9e4a1ae8/Sources/Sentry/SentryNSURLSessionTaskSearch.m#L35-L69

Sources/Sentry/SentryNSURLSessionSwizzleTargetDiscovery.m
```diff
@@ -70,4 +70,65 @@ @implementation SentryNSURLSessionTaskSearch
     return [result copy];
 }
 
+/**
+ * Discovers which NSURLSession classes need swizzling for completion handler methods.
+ * Uses the same proven discovery pattern as urlSessionTaskClassesToTrack but for session methods.
+ */
++ (NSArray<Class> *)urlSessionCompletionHandlerClassesToSwizzle:(SEL)selector
```
philippsuess: `h`: If I'm not mistaken, `NSURLSession` might be different from `NSURLSessionTask` here. `NSURLSessionTask` needs class discovery because it's a [class cluster](https://github.com/AFNetworking/AFNetworking/blob/4eaec5b586ddd897ebeda896e332a62a9fdab818/AFNetworking/AFURLSessionManager.m#L355-L404) where `setState:` isn't implemented on the base class — only on private subclasses like `__NSCFLocalSessionTask`. For `NSURLSession`, `dataTaskWithRequest:completionHandler:` should be implemented on the base class itself.

Did you try swizzling `[NSURLSession class]` directly? If that works, we could drop the class lookup and simplify this. We could also start conservatively by limiting it to iOS 26+ / iPadOS 26+, verify it works there, and expand to older versions once we're confident.

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +99,77 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
+                        completionHandler(data, response, error);
+                    };
+                }
+
+                task = SentrySWCallOriginal(request, wrappedHandler ?: completionHandler);
+                return task;
```
philippsuess: `m`: I would really love to keep the logic here to an absolute minimum, because it's better for testing. We should follow similar approaches as all the other swizzling methods that just call a class or a method in here. We should follow similar approaches as all the other swizzling methods that just call a class or a method in here. 

Also, the code below and this code here share some similarities, and it would be great to reduce some duplication.

philipphofmann (commented):
Thanks for tackling this. I have one major high-level question before reviewing this more closely.

Sources/Sentry/SentryNSURLSessionSwizzleTargetDiscovery.m
43jay: > For NSURLSession, dataTaskWithRequest:completionHandler: should be implemented on the base class itself

That's correct / what I observed in my testing. 

> Did you try swizzling [NSURLSession class] directly

Yes I did and was able to swizzle directly. I left in the run-time discovery to make the code more bullet-proof:

 - Don't assume `[NSURLSession dataTaskWithURL:completionHandler:]` is implemented with the same inheritance as `[NSURLSession dataTaskWithRequest:completionHandler:]` => derive their swizzle targets separately
 - Don't assume `NSURLSessionConfiguration ephemeralSessionConfiguration` => check `NSURLSessionConfiguration defaultSessionConfiguration` as well

I'm glad to remove it (will be much simpler) if you say it's not warranted (I was taking an extra conservative approach). I could likely move this run-time discovery logic into a unit test that would fail if the swizzle targets discovered at test-time were not exactly the NSURLSession implementations.🤔

Sources/Sentry/SentrySwizzleWrapperHelper.m
43jay: tx. I had the same reaction 👍

> other swizzling methods that just call a class or a method in here

The code here is necessarily more bloated than existing cuz it's swizzling in order to wrap a callback (the completionHandler). i.e.

**Existing swizzling** => just swizzle the implementation with ours that invokes the original.
**This swizzling** => swizzle the implementation (factory creator) to grab the developer's `completionHandler`, wrap it in ours, then pass the `wrapped completionHandler` when invoking the original.

> great to reduce some duplication

Re reducing duplication, totally agree. The issue is the [SentrySwizzle macros](https://github.com/getsentry/sentry-cocoa/blob/20ff7162cdf4402d0ab86cb99add901bdf32105b/Sources/Sentry/include/SentrySwizzle.h#L80) didn't play well with refactoring code. Definitely didn't want to modify the macros but can take another crack at how to achieve w/out modifying

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +99,77 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
+                    };
+                }
+
+                task = SentrySWCallOriginal(url, wrappedHandler ?: completionHandler);
+                return task;
+            }),
+            SentrySwizzleModeOncePerClassAndSuperclasses, (void *)dataTaskWithURLSelector);
+    }
+#pragma clang diagnostic pop
+}
```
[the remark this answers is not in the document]
43jay: this isn't accurate - added a unit tests to confirm `dataTaskWithURL:completionHandler:` does not call `dataTaskWithRequest:completionHandler:` or vice versa (now and in future)

Sources/Sentry/SentryNSURLSessionSwizzleTargetDiscovery.m
philippsuess: > I left in the run-time discovery to make the code more bullet-proof.

Well, with swizzling, making it more bulletproof can actually turn into bullets flying towards you  (believe me I saw those bullets flying multiple times 😅). I would try with the simple approach first, make it experimental, and then try bulletproof if required.

43jay: more details on the "make it experimental" bit? is that just verbiage in CHANGELOG, or something in code?

philippsuess: Ups I'm sorry. This new feature most likely will require a feature flag / option. This option should be in the experimental options (SentryExperimentalOptions) and/or clearly be marked as experimental, as for example `enableMetrics`.

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +98,66 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
+{
+    if (!original)
+        return nil;
+    return ^(NSData *data, NSURLResponse *response, NSError *error) {
+        if (!error && data && *taskRef) {
+            [tracker captureResponseDetails:data response:response request:request task:*taskRef];
+        }
+        original(data, response, error);
+    };
+}
```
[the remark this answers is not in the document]
43jay: fixed

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +97,71 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
+{
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshadow"
+    SEL dataTaskWithRequestSelector = @selector(dataTaskWithRequest:completionHandler:);
+    SentrySwizzleInstanceMethod([NSURLSession class], dataTaskWithRequestSelector,
+        SentrySWReturnType(NSURLSessionDataTask *),
+        SentrySWArguments(NSURLRequest * request,
+            void (^completionHandler)(NSData *, NSURLResponse *, NSError *)),
+        SentrySWReplacement({
+            NSMutableArray<NSURLSessionDataTask *> *taskHolder =
```
43jay: @philipphofmann this NSMutableArray replaced the __block variable that was being used previously. 

The issue is pulling out the code (de-duplication) into `SentryWrapCompletionHandler` makes the usage of the `__block` invalid - it moves out of scope & creates dangling pointer once execution passes to `SentryWrapCompletionHandler` (quickly crashes).

(claude's) solution -> use `NSMutableArray` instead of __block, b/c NSMutableArray is heap-allocated unlike __block.

Sources/Sentry/SentryNSURLSessionSwizzleTargetDiscovery.m
43jay: resolved

Sources/Sentry/SentryNSURLSessionSwizzleTargetDiscovery.m
43jay: updated code, this should be addressed - ptal

Sources/Sentry/SentrySwizzleWrapperHelper.m
43jay: updated code; ptal

Sentry.xcodeproj/project.pbxproj
```diff
@@ -2445,14 +2445,15 @@
 				D82859442C3E753C009A28AA /* SentrySessionReplaySyncC.h in Headers */,
 				8459FCBE2BD73E820038E9C9 /* SentryProfilerSerialization.h in Headers */,
 				7BC852332458802C005A70F0 /* SentryDataCategoryMapper.h in Headers */,
+				D8370B6C273DF20F00F66E2D /* SentryNSURLSessionSwizzleTargetDiscovery.h in Headers */,
```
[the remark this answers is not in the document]
43jay: resolved (reverted rename)

Sources/Sentry/SentrySwizzleWrapperHelper.m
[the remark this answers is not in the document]
43jay: this is gone (at least should be!)

Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
```diff
@@ -1,13 +1,156 @@
...
+        var hitCount = 0
+
+        let replacementBlock: @convention(block) (NSObject, AnyObject, Any?) -> AnyObject = { obj, arg, handler in
+            hitCount += 1
+            typealias Fn = @convention(c) (NSObject, Selector, AnyObject, Any?) -> AnyObject
+            let original = unsafeBitCast(originalIMP, to: Fn.self)
+            return original(obj, targetSelector, arg, handler)
+        }
+
+        method_setImplementation(method, imp_implementationWithBlock(replacementBlock))
```
philprime: `h`: I am afraid that this swizzling changes the entire test suite and can cause flakiness. @itaybre do you agree?

itaybre: Yeah, if we can replace the swizzling it would be great.

`more info:` I see that @43jay is restoring the original implementation, but if we have concurrent tests it may not be restored in time.

Sources/Sentry/SentrySwizzleWrapperHelper.m
philippsuess: What's stopping you from using the same approach as we use in `swizzleURLSessionTask`? We basically just call the network tracker there and let the network tracker take care of everything. We know that this approach works, so I would stick to it.

Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
philippsuess: We use a similar approach here https://github.com/getsentry/sentry-cocoa/blob/316a9d6a2f6ab671829e12f86f03d77215d0f861/Tests/SentryTests/PrivateSentrySDKOnlyTests.swift#L375-L381

But yes, I agree. If we can avoid calling `method_setImplementation`, it would be great.

Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
```diff
@@ -1,13 +1,156 @@
...
+    // The swizzling code relies on this by swizzling [NSURLSession class] directly
+    // rather than doing runtime discovery. These tests verify that assumption
+    // still holds — if Apple ever moves these methods to a subclass, these tests
+    // will fail and we'll know to update the swizzling approach.
+
+    func test_URLSession_isNotClassCluster_dataTaskWithRequest() {
+        let selector = #selector(URLSession.dataTask(with:completionHandler:)
+            as (URLSession) -> (URLRequest, @escaping @Sendable (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask)
+        assertNSURLSessionImplementsDirectly(selector: selector, selectorName: "dataTaskWithRequest:completionHandler:")
+    }
```
philippsuess, line 27: `m`:I’m not sure we actually need tests at this level. Instead of testing this internal behavior, we should probably test the system at a slightly higher level. For example, we could verify that when we use the concrete implementation and interact with the API (i.e., performing web requests), we receive the expected results such as the correct response bodies.

The current test feels very close to implementation details. While we could keep it, it may not provide much real value. If our higher-level tests already verify that we correctly retrieve and process data from NSURLSession, then changes in the internal implementation shouldn’t matter.

In other words, this test could fail even though the functionality still works correctly, which would make it brittle. Conversely, it could pass while the actual behavior is broken. I would prefer tests that focus on observable behavior and validate the real functionality, without depending too heavily on implementation details.

Sources/Sentry/SentrySwizzleWrapperHelper.m
43jay: unless I'm missing something `swizzleURLSessionTask` is necessarily different; it's swizzling out the function for another function we want to call as well (1:1):
```objc
SentrySwizzleInstanceMethod(classToSwizzle, resumeSelector, SentrySWReturnType(void),
            SentrySWArguments(), 
            SentrySWReplacement({ 
                // Replacement function  - just passthrough to our impl on networkTracker
                [networkTracker urlSessionTaskResume:self];
                // and invoke SentrySWCallOriginal 👍
                SentrySWCallOriginal();
            }),
            SentrySwizzleModeOncePerClassAndSuperclasses, (void *)resumeSelector);
```

whereas the new swizzling is swizzling out the function
> `[NSURLSession dataTaskWith[URL|Request]:completionHandler:]`

to intercept  `dataTask` creation in order to wrap the `completionHandler` 
_// really we just want to swizzle the completionHandler, but can only access it at dataTask creation time._
The `completionHandler` callback has the data (NSURLResponse) we need to pass to the network tracker:

```objc
SentrySwizzleInstanceMethod([NSURLSession class], dataTaskWithRequestSelector, SentrySWReturnType(NSURLSessionDataTask *),
            SentrySWArguments(NSURLRequest * request,
                void (^completionHandler)(NSData *, NSURLResponse *, NSError *)),
            SentrySWReplacement({
                // Replacement function  -

                // Chicken-and-egg: wrappedHandler needs the task, which exists at callback time 
                // but here does not exist until `SentrySWCallOriginal` runs
                // Soln: create a holder for the to-be-initialised `NSURLSessionDataTask`
                __block NSURLSessionDataTask *task = nil;

                // Provide the wrapped completionHandler implementation.
                void (^wrappedHandler)(NSData *, NSURLResponse *, NSError *) = nil;
                if (completionHandler) {
                    wrappedHandler = ^(NSData *data, NSURLResponse *response, NSError *error) {
                        if (!error && data && task) {
                            [networkTracker captureResponseDetails:data
                                                          response:response
                                                           request:request
                                                           task:task];
                        }
                        completionHandler(data, response, error);
                    };
                }
                // Pass the wrappedHandler to `SentrySWCallOriginal` (apple's impl of `[NSURLSession dataTaskWithRequest:completionHandler:]`
                // Assign to `__block task` b/c networkTracker needs the task to know whose response this is.
                task = SentrySWCallOriginal(request, wrappedHandler ?: completionHandler);
                // Give the caller a reference to the newly initialised task 👍
                return task;
            }),
            SentrySwizzleModeOncePerClassAndSuperclasses, (void *)dataTaskWithRequestSelector);
```

Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
43jay: these tests are _trying_ to keep parity with the existing [`test_URLSessionTask_ByIosVersion()`](https://github.com/getsentry/sentry-cocoa/blob/main/Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift#L6-L11) for the new swizzling.

I.e.:
> // To know whether Apple changes NSURLSessionTask

so [`test_URLSessionTask_ByIosVersion`](https://github.com/getsentry/sentry-cocoa/blob/main/Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift#L6-L11) does
  -> initialise a localDataTask (`NSURLSessionDataTask`); walk its superclasses to discover alternative impls ("class cluster")
  -> asserts that this list only ever has the 1 expected impl. 
  -> asserts that the expected impl is on the expected class

Then,
> // To know whether Apple changes **NSURLSession dataTask factory initializers.**

`test_URLSession_isNotClassCluster_dataTaskWithURL`
`test_URLSession_isNotClassCluster_dataTaskWithRequest` will
  -> initialise a `URLSession` (.ephemeral and .default configs)
  -> asserts that regardless the returned `URLSession`, it has the expected impl from [NSURLSession].

43jay: There's 2 additional tests that i removed (Dont think they are that valuable):

`test_dataTaskWithURL_doesNotCallThrough_dataTaskWithRequest`
`test_dataTaskWithRequest_doesNotCallThrough_dataTaskWithURL`

these are testing whether the two functions we added swizzling for interact with each other. This came up in [this bot comment](https://github.com/getsentry/sentry-cocoa/pull/7584#discussion_r2886125141) which ended up being hallucination - but i had written the tests to verify so included them. 

Based on our [other convo](https://github.com/getsentry/sentry-cocoa/pull/7584#discussion_r2911502634) they can introduce flakiness in tests.

philippsuess:
## Review notes — treat with care

> **Disclaimer:** I'm on parental leave starting Monday, so I only gave this a quick glance today. Most of the investigation below came from prompting Claude Code — I walked it through the code step by step, asked it to trace the task usage across the PR stack, and challenged its initial assessment when things felt off. The findings are plausible and worth considering, but I haven't verified everything myself. **Treat this as "leads worth investigating" rather than definitive conclusions.** If needed, I can take another look tomorrow or the day after, but it might be better if @itaybre or @philprime  takes this over fully.

---

### TL;DR — suggested restructure

The `taskHolder` pattern (mutable array to break a chicken-and-egg dependency between the completion handler and the task) feels fishy. The completion handler doesn't actually need the task if you restructure the flow:

1. **Don't use the task as a bridge** between request-time and response-time data
2. **Do all replay network detail processing in the completion handler itself** — it already has `request` (captured in swizzle scope), `response`, and `data`
3. **Don't piggyback on the `setState:`/breadcrumb path** — the completion handler fires *after* `setState:`, so the timing doesn't work anyway

The completion handler swizzle would become a simple, self-contained block with no tricky patterns:

```objc
wrapped = ^(NSData *data, NSURLResponse *response, NSError *error) {
    if (!error && data) {
        [networkTracker captureNetworkDetailsForReplayWithRequest:request
                                                        response:response
                                                    responseData:data];
    }
    completionHandler(data, response, error);
};
```

No `taskHolder`, no `objc_getAssociatedObject` lookup in the completion handler, no timing dependencies.

---

### Detailed findings from Claude Code investigation

#### 1. The `taskHolder` chicken-and-egg pattern

The wrapped completion handler needs a reference to the `NSURLSessionDataTask`, but the task doesn't exist yet when the handler is constructed. The current solution:

```
taskHolder = NSMutableArray (empty)
task = SentrySWCallOriginal(request, wrappedHandler)  // wrappedHandler captures taskHolder
[taskHolder addObject:task]
return task
```

The assumption is: the completion handler will never fire before `[taskHolder addObject:task]` because the task starts in `NSURLSessionTaskStateSuspended` and must be `resume`d first.

**Why this is concerning:** This relies on an undocumented implementation detail. Apple doesn't guarantee that:
- A cached response won't trigger the completion handler synchronously during creation
- An invalid/malformed request won't deliver an error inline before returning
- An invalidated session won't call back immediately
- This behavior won't change in a future OS version

If any of these edge cases fire the completion handler during `SentrySWCallOriginal`, `taskHolder.firstObject` is `nil`, and capture silently fails. No crash, just silent data loss — which is worse because you'd never notice.

#### 2. The `setState:` vs completion handler timing

I asked Claude to trace the execution order. The `setState:` swizzle fires *before* the completion handler:

```
1. Our setState: swizzle fires (task.state is still .running)
2. Original setState: executes (state becomes .completed)
3. Completion handler fires
```

This means the approach in #7590 (reading `SentryReplayNetworkDetails` from the task in `addBreadcrumbForSessionTask:`) would read the associated object *before* the completion handler has populated the response data. This is a second potential race condition in the stack.

#### 3. Class cluster concern

The swizzle targets `[NSURLSession class]`, but `NSURLSession` is a class cluster — actual instances are private subclasses (e.g., `__NSURLSessionLocal`). If those subclasses override `dataTaskWithRequest:completionHandler:`, the swizzle on the base class won't intercept the calls. Compare with `swizzleURLSessionTask:` in the same file, which uses `SentryNSURLSessionTaskSearch` to discover the actual runtime classes. The response capture swizzle skips this step.

#### 4. The existing `SentryNetworkTracker` already captures almost everything

The existing `resume`/`setState:` flow in `SentryNetworkTracker.m` is rock solid. It already captures URL, method, status code, request/response headers, body sizes — all directly from task properties. The **only** thing it can't get is the response body bytes (`NSData`), which is only available in the completion handler.

So the completion handler swizzle is necessary, but it should be minimal: just capture the data and process it right there, rather than building an elaborate bridge back to the `setState:` path via associated objects.

#### 5. How I got here (prompting approach)

I asked Claude to:
1. Read `swizzleURLSessionDataTasksForResponseCapture` and explain it
2. Give its honest opinion on the approach — it initially said the `taskHolder` pattern was "clever and correct"
3. I pushed back on the race condition concern — Claude reconsidered and acknowledged the undocumented timing assumptions
4. Read all 4 stacked PRs (7582, 7585, 7588, 7590) to trace how the task flows through the stack
5. Read the full `SentryNetworkTracker.m` to understand the existing infrastructure
6. Propose how to merge the new functionality with the existing bulletproof flow

Again: these are Claude-assisted findings. The reasoning is sound to me on a quick read, but please verify independently before making architectural changes based on this.

Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
43jay: removed this

43jay:
> ## Review notes — treat with care
> > **Disclaimer:** I'm on parental leave starting Monday, so I only gave this a quick glance today. Most of the investigation below came from prompting Claude Code — I walked it through the code step by step, asked it to trace the task usage across the PR stack, and challenged its initial assessment when things felt off. The findings are plausible and worth considering, but I haven't verified everything myself. **Treat this as "leads worth investigating" rather than definitive conclusions.** If needed, I can take another look tomorrow or the day after, but it might be better if @itaybre or @philprime  takes this over fully.
> 
> ### TL;DR — suggested restructure
> The `taskHolder` pattern (mutable array to break a chicken-and-egg dependency between the completion handler and the task) feels fishy. The completion handler doesn't actually need the task if you restructure the flow:
> 
> 1. **Don't use the task as a bridge** between request-time and response-time data
> 2. **Do all replay network detail processing in the completion handler itself** — it already has `request` (captured in swizzle scope), `response`, and `data`
> 3. **Don't piggyback on the `setState:`/breadcrumb path** — the completion handler fires _after_ `setState:`, so the timing doesn't work anyway
> 
> The completion handler swizzle would become a simple, self-contained block with no tricky patterns:
> 
> ```objc
> wrapped = ^(NSData *data, NSURLResponse *response, NSError *error) {
>     if (!error && data) {
>         [networkTracker captureNetworkDetailsForReplayWithRequest:request
>                                                         response:response
>                                                     responseData:data];
>     }
>     completionHandler(data, response, error);
> };
> ```
> 
> No `taskHolder`, no `objc_getAssociatedObject` lookup in the completion handler, no timing dependencies.
> 
> ### Detailed findings from Claude Code investigation
> #### 1. The `taskHolder` chicken-and-egg pattern
> The wrapped completion handler needs a reference to the `NSURLSessionDataTask`, but the task doesn't exist yet when the handler is constructed. The current solution:
> 
> ```
> taskHolder = NSMutableArray (empty)
> task = SentrySWCallOriginal(request, wrappedHandler)  // wrappedHandler captures taskHolder
> [taskHolder addObject:task]
> return task
> ```
> 
> The assumption is: the completion handler will never fire before `[taskHolder addObject:task]` because the task starts in `NSURLSessionTaskStateSuspended` and must be `resume`d first.
> 
> **Why this is concerning:** This relies on an undocumented implementation detail. Apple doesn't guarantee that:
> 
> * A cached response won't trigger the completion handler synchronously during creation
> * An invalid/malformed request won't deliver an error inline before returning
> * An invalidated session won't call back immediately
> * This behavior won't change in a future OS version
> 
> If any of these edge cases fire the completion handler during `SentrySWCallOriginal`, `taskHolder.firstObject` is `nil`, and capture silently fails. No crash, just silent data loss — which is worse because you'd never notice.
> 
> #### 2. The `setState:` vs completion handler timing
> I asked Claude to trace the execution order. The `setState:` swizzle fires _before_ the completion handler:
> 
> ```
> 1. Our setState: swizzle fires (task.state is still .running)
> 2. Original setState: executes (state becomes .completed)
> 3. Completion handler fires
> ```
> 
> This means the approach in #7590 (reading `SentryReplayNetworkDetails` from the task in `addBreadcrumbForSessionTask:`) would read the associated object _before_ the completion handler has populated the response data. This is a second potential race condition in the stack.
> 
> #### 3. Class cluster concern
> The swizzle targets `[NSURLSession class]`, but `NSURLSession` is a class cluster — actual instances are private subclasses (e.g., `__NSURLSessionLocal`). If those subclasses override `dataTaskWithRequest:completionHandler:`, the swizzle on the base class won't intercept the calls. Compare with `swizzleURLSessionTask:` in the same file, which uses `SentryNSURLSessionTaskSearch` to discover the actual runtime classes. The response capture swizzle skips this step.
> 
> #### 4. The existing `SentryNetworkTracker` already captures almost everything
> The existing `resume`/`setState:` flow in `SentryNetworkTracker.m` is rock solid. It already captures URL, method, status code, request/response headers, body sizes — all directly from task properties. The **only** thing it can't get is the response body bytes (`NSData`), which is only available in the completion handler.
> 
> So the completion handler swizzle is necessary, but it should be minimal: just capture the data and process it right there, rather than building an elaborate bridge back to the `setState:` path via associated objects.
> 
> #### 5. How I got here (prompting approach)
> I asked Claude to:
> 
> 1. Read `swizzleURLSessionDataTasksForResponseCapture` and explain it
> 2. Give its honest opinion on the approach — it initially said the `taskHolder` pattern was "clever and correct"
> 3. I pushed back on the race condition concern — Claude reconsidered and acknowledged the undocumented timing assumptions
> 4. Read all 4 stacked PRs (7582, 7585, 7588, 7590) to trace how the task flows through the stack
> 5. Read the full `SentryNetworkTracker.m` to understand the existing infrastructure
> 6. Propose how to merge the new functionality with the existing bulletproof flow
> 
> Again: these are Claude-assisted findings. The reasoning is sound to me on a quick read, but please verify independently before making architectural changes based on this.

appreciate the input, will work through it fully. At a glance some is valid and some I already worked through.
**edit: worked through it. Responses below**

The main issue :
> 1. The taskHolder chicken-and-egg pattern

> Don't use the task as a bridge between request-time and response-time data
> Do all replay network detail processing in the completion handler itself — it already has request (captured in swizzle scope), response, and data

The issue I hit was that `NSURLRequest` is not enough to differentiate, for example, multiple simultaneous requests with the same parameters - afaict NSURLRequest is basically a URL and an HTTP verb.

The others don't seem relevant or seem inaccurate:

> #### 2. The `setState:` vs completion handler timing
> Don't piggyback on the setState:/breadcrumb path — the completion handler fires after setState:, so the timing doesn't work anyway

> 1. Our setState: swizzle fires (task.state is still .running)
> 2. Original setState: executes (state becomes .completed)
> 3. Completion handler fires

This isn't correct based on my local testing. The completionHandler always fired before `setState(.completed)` runs. This logically makes sense as the app needs to 'handle completion' before the task can be considered 'completed' (**edit**: Found a [counterexample](SentryReplayNetworkDetails) that invalidates this)

Regardless the code is written so that if completionHandler fires before [ANY](https://github.com/getsentry/sentry-cocoa/blob/fea89107fbee4f14589ce2196e88d95ce638032b/Sources/Sentry/SentryNetworkTracker.m#L232-L245) call to setState, then `captureResponseDetails` call  won't find a `SentryReplayNetworkDetails` and bails
https://github.com/getsentry/sentry-cocoa/blob/fea89107fbee4f14589ce2196e88d95ce638032b/Sources/Sentry/SentryNetworkTracker.m#L605-L610


> #### 3. Class cluster concern
This is our [other convo](https://github.com/getsentry/sentry-cocoa/pull/7584#discussion_r2883283138) - basically NSURLSession dataTask initializers are NOT currently class clusters (claude misspoke/inaccurate above). 

Regardless, the impl might change. [The tests](https://github.com/getsentry/sentry-cocoa/pull/7584#discussion_r2914362344) added to `SentryNSURLSessionTaskSearchTests.swift` are meant to uncover this by initializing `NSURLSession` with different configs to assert that the class returned at run-time has exactly the imp from the base [NSURLSession]

> #### 4. The existing `SentryNetworkTracker` already captures almost everything
This is point 1. The taskHolder chicken-and-egg pattern
Or rather the issue is the same - SentryNetworkTracker (aiui) doesn't have a way to associate a specific NSURLRequest with the completionHandler NSURLResponse

Sources/Sentry/SentryNSURLSessionSwizzleTargetDiscovery.m
43jay: resolution: removed the run-time swizzle target discovery for direct swizzling (much [simpler implementation](https://github.com/getsentry/sentry-cocoa/pull/7584/changes/cbac0b919ca961c1ee30eb5284de9e8325c1332c))

Tests/SentryTests/Integrations/Performance/Network/SentryNSURLSessionTaskSearchTests.swift
43jay: removed

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +97,71 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
+                [NSMutableArray arrayWithCapacity:1];
+            NSURLSessionDataTask *task = SentrySWCallOriginal(request,
+                SentryWrapCompletionHandler(networkTracker, request, taskHolder, completionHandler));
+            [taskHolder addObject:task];
```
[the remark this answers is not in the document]
43jay: reverted the `SentryWrapCompletionHandler` => resolved

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +97,88 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
+{
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshadow"
+    SEL selector = @selector(dataTaskWithURL:completionHandler:);
+    SentrySwizzleInstanceMethod([NSURLSession class], selector,
+        SentrySWReturnType(NSURLSessionDataTask *),
+        SentrySWArguments(
+            NSURL * url, void (^completionHandler)(NSData *, NSURLResponse *, NSError *)),
+        SentrySWReplacement({
+            NSURLRequest *request = [NSURLRequest requestWithURL:url];
```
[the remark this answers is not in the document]
43jay: addressed (ammened [this commit](https://github.com/getsentry/sentry-cocoa/pull/7584/changes/b833db184b0227b77c896ee9e660351ed6b9683a))

minor issue - the `request` passed in to captureResponseDetails is _only_ used for the url. but valid regardless.

Sources/Sentry/SentryNetworkTracker.m
```diff
@@ -562,4 +562,15 @@ - (SentryLevel)getBreadcrumbLevel:(NSURLSessionTask *)sessionTask
...
+- (void)captureResponseDetails:(NSData *)data
+                      response:(NSURLResponse *)response
+                    requestURL:(NSURL *)requestURL
+                          task:(NSURLSessionTask *)task
+{
+    // TODO: Implementation
+    // 2. Parse response body data
+    // 3. Store in appropriate location for session replay
+    // 4. Handle size limits and truncation if needed
+}
```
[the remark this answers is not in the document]
43jay: handled [here](https://github.com/getsentry/sentry-cocoa/pull/7588/changes/99deb05a6c8ccfa0bb79facb333535af1cb1b8f0)

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +97,87 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
+                        [networkTracker captureResponseDetails:data
+                                                     response:response
+                                                   requestURL:request.URL
+                                                         task:task];
+                    }
+                    completionHandler(data, response, error);
+                };
+            }
+            task = SentrySWCallOriginal(request, wrappedHandler ?: completionHandler);
+            return task;
```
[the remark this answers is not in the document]
43jay: hmm. The pattern (in #7588) is to use the originalRequest. Rationale is that if there's multiple requests (e.g. the request changes), we can still only show one in the replay dashboard -> use the first one seen. i think this applies here as well => **will not address**

Sources/Sentry/SentrySwizzleWrapperHelper.m
```diff
@@ -97,6 +97,87 @@ + (void)swizzleURLSessionTask:(SentryNetworkTracker *)networkTracker
...
++ (void)swizzleURLSessionDataTasksForResponseCapture:(SentryNetworkTracker *)networkTracker
+{
+    [self swizzleDataTaskWithRequestCompletionHandler:networkTracker];
+    [self swizzleDataTaskWithURLCompletionHandler:networkTracker];
+}
+
+/**
+ * Swizzles -[NSURLSession dataTaskWithRequest:completionHandler:] to intercept response data.
+ */
++ (void)swizzleDataTaskWithRequestCompletionHandler:(SentryNetworkTracker *)networkTracker
```
itaybre, line 118: `h`: I would like to see some test verifying the swizzling actually work.
While this looks good, it may break at any point in later versions.

While it is not the same, a strategy similar to [SentryNSDataSwizzlingHelperTests.swift](https://github.com/getsentry/sentry-cocoa/blob/main/Tests/SentryTests/Integrations/Performance/IO/SentryNSDataSwizzlingHelperTests.swift#L66) could be used to verify the methods are swizzled when executing a datatask

The current tests are verifying the method are changed, but not that they are actually called (what we actually care for)

Once that is added, this would look good to me to merge

43jay: something like the following ? (kind of like an API test / integration test - i.e. use the API and validate that the swizzling callbacks a) occur and ideally b) occur as expected )

**test scenario**
1) Initialize a NSUrlSessionDataTask
2) `resume` it (or whatever the API is to get it to run)
3) have it run to completion
4) verify our completionHandler called, verify the original completion handler called

**Couple different versions:**
For both newly swizzled initializers:
1) With networkDetailAllowUrls populated (feature turned on; swizzling active)
2) With networkDetailAllowUrls populated (feature turned off; swizzling non-existent)

Will do. 

**if you see this in time** => help me by weighing in on whether this sounds like a test that should use an actual server (there is some test server set-up that i haven't needed to use it but will set it up if you say it's the right direction).
Or if the current test harness (à la unit test with mocking) is good enough,

itaybre: That sounds good to me, you can use https://postman-echo.com
If it is flaky we can boot a local server, but I would prefer not to do that if possible

43jay: Done - [PTAL](https://github.com/getsentry/sentry-cocoa/pull/7584/changes/c24784c217c2f53233bce2769e98a352d743506f)

Sources/Swift/Integrations/Performance/Network/SentryNetworkTrackingIntegration.swift
```diff
@@ -41,6 +41,12 @@ final class SentryNetworkTrackingIntegration<Dependencies: NetworkTrackerProvide
         super.init()
 
         SentrySwizzleWrapperHelper.swizzleURLSessionTask(networkTracker)
+
```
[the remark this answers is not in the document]
43jay: will not fix. 
> enableNetworkBreadcrumbs

the current network detail capture relies on network breadcrumb for passing data 

> shouldEnableNetworkTracking

seems reasonable to gate network detail feature behind 'shouldEnableNetworkTracking'

> enableCaptureFailedRequests

This seems unrelated, but will not be true if `shouldEnableNetworkTracking` is false.

itaybre (approved):
LGTM

Sources/Sentry/SentryNetworkTracker.m
```diff
@@ -562,4 +562,15 @@ - (SentryLevel)getBreadcrumbLevel:(NSURLSessionTask *)sessionTask
...
 }
 
+- (void)captureResponseDetails:(NSData *)data
+                      response:(NSURLResponse *)response
+                    requestURL:(NSURL *)requestURL
+                          task:(NSURLSessionTask *)task
+{
+    // TODO: Implementation
+    // 2. Parse response body data
+    // 3. Store in appropriate location for session replay
```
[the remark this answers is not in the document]
43jay: landing in stacked PR

Tests/SentryTests/Integrations/Performance/Network/SentryNetworkDetailSwizzlingTests.swift
```diff
@@ -0,0 +1,148 @@
...
+
+        // Network details were captured via the swizzled completion handler
+        let breadcrumb = try lastHTTPBreadcrumb(for: echoURL)
+        let details = try XCTUnwrap(
+            breadcrumb.data?[SentryReplayNetworkDetails.replayNetworkDetailsKey] as? SentryReplayNetworkDetails,
+            "Swizzled completion handler should have populated network details on the breadcrumb"
+        )
+        let serialized = details.serialize()
+        XCTAssertEqual(serialized["statusCode"] as? Int, 200)
+        XCTAssertNotNil(serialized["response"], "Response details should be captured")
```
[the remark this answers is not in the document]
43jay: moved that commit into https://github.com/getsentry/sentry-cocoa/pull/7590

43jay:
moved failing test to PR with implementation (https://github.com/getsentry/sentry-cocoa/pull/7590)

43jay:
added missing &&  ! SENTRY_NO_UI_FRAMEWORK in SentryNetworkTrackingIntegration.swift

43jay:
missed make format on a file

43jay:
missed a `make generate api` - that job runs off the HEAD branch `mobile-935/ios-network details`

PR #7598, opened 2026-03-04
feat(objc): Add SentryObjC wrapper SDK

### LINKED ISSUE #6342: Create an ObjC SDK

philprime:
## 📜 Description

Adds a pure Objective-C SDK wrapper (`SentryObjC`) that enables Sentry usage in Objective-C++ projects with `CLANG_ENABLE_MODULES=NO`. This solves the long-standing issue where Swift SDK APIs are inaccessible without Clang modules, affecting React Native, Haxe, and other custom build systems.

**Read the develop-docs/SENTRY-OBJC.md architecture document first**.

## 💡 Motivation and Context

Closes getsentry/sentry-cocoa#6342

### The Problem

Since SDK 8.54.0, the Swift SDK's public API requires Clang modules to be imported. When `CLANG_ENABLE_MODULES=NO`:

* `#import <Sentry/Sentry.h>` only exposes Objective-C APIs
* Swift types like `SentrySDK`, `SentryOptions.sessionReplay`, and the Metrics API are unavailable
* Build errors: `use of undeclared identifier 'SentrySDK'`

This breaks projects that cannot enable modules, particularly:

* React Native (uses ObjC++ extensively)
* Haxe-based applications
* Custom build systems with strict compiler settings

### Initial Considerations

**Option 1: Make Swift SDK importable without modules**

* ❌ Extremely difficult - would require rewriting core SDK architecture
* ❌ Breaks existing Swift API design patterns
* ❌ Not backwards compatible

**Option 2: Pure ObjC wrapper (chosen approach)**

* ✅ Preserves existing SDK architecture
* ✅ Allows Swift and ObjC to coexist
* ✅ Minimal maintenance burden
* ✅ Clean separation of concerns

### Architecture Decision

The wrapper uses a **three-layer architecture**:

```
┌─────────────────────────────────────┐
│   SentryObjC (Pure ObjC Headers)    │  ← Public API for no-modules context
│  Sources/SentryObjC/Public/*.h      │
└──────────────┬──────────────────────┘
               │
               ▼
┌─────────────────────────────────────┐
│   SentryObjCBridge (Swift)          │  ← Forwards to Swift SDK
│  Sources/SentryObjCBridge/*.swift   │
└──────────────┬──────────────────────┘
               │
               ▼
┌─────────────────────────────────────┐
│     SentrySwift (Main SDK)          │  ← Actual implementation
└─────────────────────────────────────┘
```

**Why this design?**

See `develop-docs/SENTRY-OBJC.md`

## :pencil: Checklist

- [X] I added tests to verify the changes
- [X] No new PII added
- [X] Docs updated (sample README, [AGENTS.md](<http://AGENTS.md>))
- [ ] Wizard update (not applicable - advanced use case)
- [X] Review from native team
- [X] No breaking changes
- [X] No breaking changes for hybrid SDKs

## 🔗 Related Issues & Context

* Closes getsentry/sentry-cocoa#6342
* Related to getsentry/sentry-cocoa#4543 (original ObjC++ without modules issue)
* Thread safety fix for exception handling extracted to getsentry/sentry-cocoa#7672
* Architectural discussion in internal Slack thread


philprime:
@sentry review

philprime:
## Recent Fixes (Latest Commit)

Just pushed fixes for several critical issues discovered during implementation review:

### 🔒 Thread Safety (Critical)

**Problem**: `_insideReportException` flag used non-atomic operations, creating race conditions when multiple threads throw exceptions simultaneously.

**Impact**: Could cause duplicate exception reports or missed captures in multi-threaded apps.

**Fix**: Switched to C11 atomics (`_Atomic BOOL` + `atomic_store`/`atomic_load`). This guarantees thread-safe access without mutex overhead.

```objc
// Before (unsafe)
static BOOL _insideReportException = NO;
_insideReportException = YES;

// After (thread-safe)
static _Atomic BOOL _insideReportException = NO;
atomic_store(&_insideReportException, YES);
```

### 🔄 Session Replay Implementation

**Problem**: Replay API returned `nil` with a TODO comment instead of forwarding to the actual implementation.

**Impact**: Session replay would be completely broken on iOS/tvOS despite being advertised as supported.

**Solution Journey**:
1. **Attempt 1**: Forward-declare `SentrySDKInternal.replay` class property
   - ❌ Failed: ObjC compiler doesn't recognize class properties in forward declarations
   
2. **Attempt 2**: Declare as class method `+ (id)replay`
   - ❌ Failed: "no known class method for selector 'replay'"
   
3. **Final approach**: Expose via bridge (consistent with metrics/logger pattern)
   ```swift
   // SentryObjCBridge.swift
   @objc public static var replay: AnyObject {
       return SentrySDK.replay
   }
   ```
   - ✅ Works! Compiler happy, linker resolves correctly

**Lesson learned**: When bridging ObjC → Swift → ObjC, keep the pattern consistent. Don't mix forward declarations and bridge calls.

### 📦 Swift 6.1 Package Manifest

**Problem**: `Package@swift-6.1.swift` was missing `SentryObjCBridge` in product targets.

**Impact**: Linker errors when building with Swift 6.1 - bridge symbols wouldn't be included in the final product.

```swift
// Before (broken)
.library(name: "SentryObjC", targets: ["SentryObjCInternal", "SentryObjC", "SentryCppHelper"])

// After (fixed)
.library(name: "SentryObjC", targets: ["SentryObjCInternal", "SentryObjCBridge", "SentryObjC", "SentryCppHelper"])
```

**Why missed initially**: The main `Package.swift` had it correct, but Swift 6.1 uses a separate manifest that wasn't updated in sync.

### ⚡ Metrics API Initialization

**Problem**: Using `+initialize` for early initialization is fragile.

**Risk**: If bridge calls Swift SDK before `startWithOptions:`, could trigger undefined behavior or nil access.

**Fix**: Lazy initialization with `dispatch_once`:
```objc
+ (id<SentryObjCMetricsApi>)metrics {
    static SentryObjCMetricsApiImpl *_metricsApi = nil;
    dispatch_once(&onceToken, ^{
        _metricsApi = [[SentryObjCMetricsApiImpl alloc] init];
    });
    return _metricsApi;
}
```

Guarantees:
- Thread-safe singleton
- Initialized only when first accessed
- No premature SDK calls

### 🧹 Minor Fixes

1. **Header const placement**: `NSString const *` → `NSString *const` (correct pointer constness)
2. **Attribute validation**: Simplified from verbose logging to silent `compactMapValues` (more robust, less overhead)
3. **Test coverage**: Added tests for replay API, invalid attributes, and singleton behavior

### ✅ Verification

All fixes verified through:
```bash
make format   # ✅ Auto-formatting passed
make analyze  # ✅ 0 static analysis violations  
make build-ios  # ✅ Clean build
make build-sample-iOS-ObjectiveCpp-NoModules  # ✅ Sample compiles
```

These fixes move the implementation from "feature-complete but fragile" to "production-ready and robust."

philprime:
@sentry review

philprime:
## CI: SentryObjC XCFramework Pipeline

Added CI support for building the standalone `SentryObjC.xcframework` as part of the release workflow.

### What changed

- **2 new slice entries** in `generate_release_matrix.sh`: `SentryObjCBridge` and `SentryObjC` (both as `staticlib`). The existing `sentry-static` slices are reused for the Sentry component — no redundant builds.
- **New assembly workflow** (`assemble-xcframework-sentryobjc.yml`): downloads all 3 sets of static slices, arranges them into the expected directory layout, then runs `build-xcframework-sentryobjc-standalone.sh` which does `libtool` merge + `swiftc` dynamic link per SDK.
- **Byproduct cleanup** in slice builds: `SentryObjCBridge` and `SentryObjC` slice builds now remove `Sentry.framework` byproducts (same pattern as `SentrySwiftUI`).
- **Removed `-ForEmbedding` suffix** — investigation confirmed the `-ForEmbedding` static builds were identical to the normal static builds (same scheme, config, Mach-O type). The suffix was only a directory naming convention from development, not a different build configuration.
- **Dynamic framework detection** in the standalone script: instead of hardcoded per-platform framework lists (which broke on watchOS/tvOS due to missing WebKit etc.), the script now checks which frameworks actually exist in each SDK's `System/Library/Frameworks/` at link time.

### CI flow

```
generate_release_matrix.sh
  → build-xcframework-variant-slices.yml (matrix: 6 slices × N SDKs)
      includes: sentry-static, sentryobjc-bridge, sentryobjc-objc
  → assemble-xcframework-sentryobjc.yml (custom assembly)
      downloads sentry-static + sentryobjc-bridge + sentryobjc-objc slices
      libtool + swiftc per SDK → SentryObjC.xcframework
```

philprime:
Fixed pure-ObjC link failure for Swift-defined public classes.

The Swift compiler emits `@objcMembers` classes under mangled ObjC runtime names (e.g. `_TtC6Sentry14SentryFeedback`), so pure-ObjC consumers of SentryObjC resolved nothing for `_OBJC_CLASS_$_Sentry*` at link time. macOS had hidden this via `NSClassFromString` fallbacks; iOS surfaced it in sentry-unreal's iOS BuildCookRun.

Added `@compatibility_alias` to each affected public header, dispatched on `SWIFT_PACKAGE` (SPM compiles under `SentrySwift`, Xcode under `Sentry`). Consumer TUs now rewrite the plain name to the mangled class that `Sentry.framework` already exports.

Covered: `SentryAttribute`, `SentryEnvelope`, `SentryEnvelopeHeader`, `SentryEnvelopeItem`, `SentryExperimentalOptions`, `SentryFeedback`, `SentryLog`, `SentryLogger`, `SentryReplayOptions`.

Out of scope:
- `SentryMessage` — pure ObjC, already exports `_OBJC_CLASS_$_SentryMessage`.
- `SentryOptions` — Swift class `Options` with `@objc(SentryOptions)`, already exports the plain symbol.
- Category B (`SentryEvent`, `SentryBreadcrumb`, `SentryScope`, `PrivateSentrySDKOnly`, …) — follow-up issue.

Verified:
- `make build-sentryobjc` green.
- `make build-sample-iOS-ObjectiveCpp-NoModules` green (real no-modules consumer).
- Synthetic no-modules consumer `.o` now references `_OBJC_CLASS_$__TtC6Sentry…<Name>` (mangled) for all 9 aliased classes and links cleanly against the built `SentryObjC.framework`.

philprime:
@sentry review
@cursor review

philprime:
@sentry review
@cursor review

philprime:
@sentry review

.github/workflows/assemble-xcframework-sentryobjc.yml
```diff
@@ -0,0 +1,166 @@
...
+        required: false
+        type: string
+
+jobs:
+  assemble-xcframework-sentryobjc:
+    name: SentryObjC
+    runs-on: macos-15
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - run: ./scripts/ci-select-xcode.sh 16.4
```
itaybre: `m`: We now have the composite action https://github.com/getsentry/sentry-cocoa/blob/main/.github/actions/setup-xcode/action.yml

.github/workflows/assemble-xcframework-sentryobjc.yml
```diff
@@ -0,0 +1,166 @@
...
+      #   XCFrameworkBuildPath/archive/Sentry/{sdk}.xcarchive/...
+      #   XCFrameworkBuildPath/archive/SentryObjCTypes/{sdk}.xcarchive/...
+      #   XCFrameworkBuildPath/archive/SentryObjCBridge/{sdk}.xcarchive/...
+      #   XCFrameworkBuildPath/archive/SentryObjC/{sdk}.xcarchive/...
+      - name: Arrange slice archives
+        shell: bash
+        env:
+          SDKS: ${{ inputs.sdks }}
+        run: |
+          set -euo pipefail
```
itaybre: `m`: Can we move this to a bash script so we can run it locally (for testing)?

Samples/iOS-ObjectiveC/App/Sources/ViewController.m
```diff
@@ -65,9 +65,9 @@ - (IBAction)captureMessage:(id)sender
 - (IBAction)captureUserFeedbackV2:(id)sender
 {
     NSData *data = [NSData dataWithContentsOfURL:BundleResourceProvider.screenshotURL];
-    NSArray<NSData *> *attachments = nil;
+    NSArray<SentryAttachment *> *attachments = nil;
     if (data != nil) {
-        attachments = @[ data ];
+        attachments = @[ [[SentryAttachment alloc] initWithData:data filename:@"screenshot.png"] ];
```
itaybre, line 70: 🙈

Samples/iOS-ObjectiveCpp-NoModules/App/Sources/AppDelegate.mm
```diff
@@ -1,32 +1,21 @@
...
-// "use of undeclared identifier 'SentrySDK'" - reproducing the issue.
-//
-// The sample will NOT build until the pure ObjC SDK wrapper (#6342) is implemented.
+// Uses Sentry from Objective-C++ with CLANG_ENABLE_MODULES=NO via the pure
+// ObjC wrapper (SentryObjC). Import the umbrella header to access SentrySDK,
+// SentryOptions, sessionReplay, etc. without requiring Swift modules.
 
 #import "AppDelegate.h"
-#import <Sentry/Sentry.h>
+#import <SentryObjC.h>
```
itaybre, line 6: `l`: Just `<SentryObjC.h>` works?
I would expect it to be `SentryObjC/SentryObjC.h` but that may be the module...

Makefile
```diff
@@ -332,9 +372,8 @@ build-sample-visionOS-SwiftUI-SPM:
 
 ## Build the iOS-ObjectiveCpp-NoModules sample app
 #
-# Builds the ObjC++ without-modules sample that reproduces #4543.
-# This target is expected to FAIL until the pure ObjC SDK wrapper (#6342)
-# is implemented. Use it to verify the fix.
+# Builds the ObjC++ without-modules sample that uses SentryObjC (#6342).
+# Uses #import <SentryObjC/SentryObjC.h> for ObjC++ without -fmodules.
```
itaybre, line 375: `l`: So `#import <SentryObjC/SentryObjC.h>` or `#import <SentryObjC.h>`?

Sources/Swift/Core/Helper/Log/SentryLevel.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 
 private let levelNames = ["none", "debug", "info", "warning", "error", "fatal"]
 
-extension SentryLevel: CustomStringConvertible { 
+extension SentryLevel: CustomStringConvertible {
```
itaybre, line 6: `m`: @philprime there is a warning here:
```
extension declares a conformance of imported type 'SentryLevel' to imported protocol 'CustomStringConvertible'; this will not behave correctly if the owners of 'SentryHeaders' introduce this conformance in the future
```

Sources/SentryObjC/SentryOptionsObjCBridge.m
```diff
@@ -0,0 +1,26 @@
...
+#    import "SentryObjCMetric.h"
+#endif
+
+static const void *kBeforeSendMetricKey = &kBeforeSendMetricKey;
+
+@implementation SentryOptions (ObjCBridge)
+
+- (void)setBeforeSendMetric:(SentryBeforeSendMetricCallback)block
+{
+    objc_setAssociatedObject(self, kBeforeSendMetricKey, block, OBJC_ASSOCIATION_COPY_NONATOMIC);
```
itaybre: `l`: Will this break automatically if we rename SentryOptions?

scripts/update-package-sha.sh
```diff
@@ -13,30 +13,41 @@ if [ -z "$PACKAGE_FILES" ]; then
...
-      sed -i "s/checksum: \".*\" \/\/Sentry-WithoutUIKitOrAppKit-WithARM64e/checksum: \"$NEW_CHECKSUM_WITHOUT_UIKIT_OR_APPKIT_WITH_ARM64E\" \/\/Sentry-WithoutUIKitOrAppKit-WithARM64e/" "$package_file"
-  else
-      sed -i "" "s/checksum: \".*\" \/\/Sentry-Static/checksum: \"$NEW_CHECKSUM_STATIC\" \/\/Sentry-Static/" "$package_file"
-      sed -i "" "s/checksum: \".*\" \/\/Sentry-Dynamic/checksum: \"$NEW_CHECKSUM_DYNAMIC\" \/\/Sentry-Dynamic/" "$package_file"
-      sed -i "" "s/checksum: \".*\" \/\/Sentry-Dynamic-WithARM64e/checksum: \"$NEW_CHECKSUM_DYNAMIC_WITH_ARM64E\" \/\/Sentry-Dynamic-WithARM64e/" "$package_file"
-      sed -i "" "s/checksum: \".*\" \/\/Sentry-WithoutUIKitOrAppKit/checksum: \"$NEW_CHECKSUM_WITHOUT_UIKIT_OR_APPKIT\" \/\/Sentry-WithoutUIKitOrAppKit/" "$package_file"
-      sed -i "" "s/checksum: \".*\" \/\/Sentry-WithoutUIKitOrAppKit-WithARM64e/checksum: \"$NEW_CHECKSUM_WITHOUT_UIKIT_OR_APPKIT_WITH_ARM64E\" \/\/Sentry-WithoutUIKitOrAppKit-WithARM64e/" "$package_file"
-  fi
+    for entry in "${ZIPS_AND_MARKERS[@]}"; do
```
itaybre, line 44: Thanks for improving this 🙏

scripts/build-xcframework-sentryobjc-standalone.sh
```diff
@@ -0,0 +1,268 @@
...
+        iphonesimulator)   arch_targets=( "arm64-apple-ios15.0-simulator" "x86_64-apple-ios15.0-simulator" ) ;;
+        macosx)            arch_targets=( "arm64-apple-macos10.14" "x86_64-apple-macos10.14" ) ;;
+        maccatalyst)       arch_targets=( "arm64-apple-ios15.0-macabi" "x86_64-apple-ios15.0-macabi" )
+                           sysroot="$(xcrun --sdk macosx --show-sdk-path)" ;;
+        appletvos)         arch_targets=( "arm64-apple-tvos15.0" ) ;;
+        appletvsimulator)  arch_targets=( "arm64-apple-tvos15.0-simulator" "x86_64-apple-tvos15.0-simulator" ) ;;
+        watchos)           arch_targets=( "arm64-apple-watchos8.0" "arm64_32-apple-watchos8.0" ) ;;
+        watchsimulator)    arch_targets=( "arm64-apple-watchos8.0-simulator" "x86_64-apple-watchos8.0-simulator" ) ;;
+        xros)              arch_targets=( "arm64-apple-xros1.0" ) ;;
+        xrsimulator)       arch_targets=( "arm64-apple-xros1.0-simulator" ) ;;
```
itaybre: `m`: What's the reason for these specific archs?
Will these need to be updated if we bump the minimum versions?

itaybre (commented):
Almost LGTM, just some comments

Sources/SentryObjC/SentrySDK.m
```diff
@@ -0,0 +1,302 @@
...
+
+#import "SentryLastRunStatus.h"
+#import "SentryMetricsApiImpl.h"
+#import "SentrySDK.h"
+
+extern void SentryBridgeCallbacksForOptions(SentryOptions *_Nonnull options);
+
+NS_ASSUME_NONNULL_BEGIN
+
+@implementation SentrySDK
```
itaybre: `m`: I was wondering, wont this cause duplicate classes?

.github/workflows/assemble-xcframework-sentryobjc.yml
philprime: Yes I can do that

.github/workflows/assemble-xcframework-sentryobjc.yml
philprime: Thanks, I'll update the PR

philprime: Done — switched to the `setup-xcode` composite action in f4a634990.

.github/workflows/assemble-xcframework-sentryobjc.yml
philprime: Done — extracted to `scripts/ci-arrange-xcframework-slices.sh` in f4a634990. Uses named `--sdks` parameter per script conventions.

Sources/Swift/Core/Helper/Log/SentryLevel.swift
philprime: This is a SourceKit false positive. `SentryLevel` is defined in `Sources/Sentry/Public/SentryLevel.h` — same module as this Swift extension. The compiler confirms it: adding `@retroactive` fails with "'retroactive' attribute does not apply; 'SentryLevel' is declared in this module." The warning doesn't appear in actual builds (only in the IDE via SourceKit).

Samples/iOS-ObjectiveCpp-NoModules/App/Sources/AppDelegate.mm
philprime: Both forms work, but for different reasons.

`SentryObjC` is a compile-from-source SPM target (like `SentrySPM`, not a pre-compiled framework like `Sentry`), so SPM adds its `publicHeadersPath` directory directly to the consumer's header search paths. That's why the flat `#import <SentryObjC.h>` resolves.

The framework-qualified `#import <SentryObjC/SentryObjC.h>` would also work and is more portable (e.g. if someone later consumes the pre-built XCFramework instead). 

Happy to switch if you prefer.

Makefile
philprime: Ah an inconsistency. See my other PR feedback comment:
https://github.com/getsentry/sentry-cocoa/pull/7598#discussion_r3194134841

Happy to update afterwards

Sources/SentryObjC/SentryOptionsObjCBridge.m
philprime: The `SentryObjC/Public/SentryOptions.h` is a hand-written ObjC header, and the actual class comes from `Options.swift` via `@objc(SentryOptions)`. 

If someone renamed just the `@objc(...)` name, this header would still compile but the runtime class wouldn't match — so it would be a runtime failure, not a compile error.

This is one of the major downsides of this split declaration style (manually maintained ObjC header backed by a Swift implementation), but the only other approach would be creating a `SentryObjcOptions.h` interface + implementation, which would duplicate quite some logic.

Sources/SentryObjC/SentrySDK.m
philprime: Good catch! I assumed that the embedded `Sentry.framework` would be internal, but both export a class with the same ObjC runtime name:

```
$ nm SentryObjC.framework/SentryObjC | grep 'OBJC_CLASS.*SentrySDK' | grep -v 'Internal\|Log\|Only'
00000000000087a0 S _OBJC_CLASS_$_SentrySDK
00000000000082d0 s __OBJC_CLASS_RO_$_SentrySDK

$ nm SentryObjC.framework/Frameworks/Sentry.framework/Sentry | grep 'OBJC_CLASS.*SentrySDK' | grep -v 'Internal\|Log\|Only'
000000000022bfb8 S _OBJC_CLASS_$__TtC6Sentry9SentrySDK
```

- `SentryObjC.framework/SentryObjC`: `_OBJC_CLASS_$_SentrySDK` is the ObjC class from `SentrySDK.m`
- `SentryObjC.framework/Frameworks/Sentry.framework/Sentry`: `_OBJC_CLASS_$__TtC6Sentry9SentrySDK` is the Swift class, mangled symbol, but registers as `SentrySDK` in the ObjC runtime via `@objc`

Both resolve to the same ObjC runtime name `SentrySDK`, causing a duplicate class conflict at runtime. Interestingly it did not surface during testing, but it makes sense to fix this now.

Since `SentryObjC` is the new SDK, I'll rename its class to `SentryObjcSDK` to avoid the collision.

scripts/build-xcframework-sentryobjc-standalone.sh
philprime: I changed it so that it dynamically finds the relevant versions from what's configured in the Xcode build settings

Sources/SentryObjC/SentryOptionsObjCBridge.m
```diff
@@ -0,0 +1,26 @@
...
+
+#if __has_include(<SentryObjCTypes/SentryObjCMetric.h>)
+#    import <SentryObjCTypes/SentryObjCMetric.h>
+#else
+#    import "SentryObjCMetric.h"
+#endif
+
+static const void *kBeforeSendMetricKey = &kBeforeSendMetricKey;
+
+@implementation SentryOptions (ObjCBridge)
```
NinjaLikesCheez, line 14: `l`: To make it obvious this file is a category extension, maybe it's worth renaming this file to `SentryOptions+ObjCBridge.m`? (if you agree, SentryReplayObjCBridge.m could also be renamed)

NinjaLikesCheez (approved):

noahsmartin (commented):
Sorry to jump in with little context on this history of this PR, but as the author of the issue this is aimed at fixing I felt like I had to provide some context... 

As written the PR doesn't really implement what was proposed in the issue. The referenced issue (https://github.com/getsentry/sentry-cocoa/issues/6342) shows the objc version of the SDK starts with `import Sentry`. That's because it's proposing a new SDK that wraps the existing Sentry SDK. What this PR does though is something quite different, it introduces a new target that compiles the same source files as the current SDK but with different settings. IMO this is going to be much more difficult to maintain than what I proposed, and in particular make it harder to migrate more of the SDK to Swift.

An ObjC wrapper SDK would involve 0 changes to existing code files (objc and swift files) and just add one new target to the Package.swift file (SentryObjC - with a dependency on SentrySPM). I would really encourage trying this option since it will make long term maintenance much easier. This PR entangles the two distributions (swift public API and ObjC API) and my suggestion separates these two concerns into distinct layers where the ObjC API builds on the Swift API. A layered approach like that with a clear boundary between the two is usually easier to work with than try to support two modes like this PR is doing.

philprime:
Thanks for the feedback @noahsmartin, I hope I can give more context like this.

> As written the PR doesn't really implement what was proposed in the issue. 

I believe you and me have to different takes on the whole topic of building an Objective-C SDK. 

The goal of this PR is introducing an Objective-C SDK which wraps our existing Swift+ObjC SDK, to make our Swift-only features available to ObjC too and so we eventually can do breaking changes when turning it into a mostly-Swift SDK.

By asking SDK users to switch from `Sentry` to `SentryObjC` they can continue using the SDK from Objective-C, while we can do breaking-change migration, e.g. drop `@objc` in our Swift+ObjC SDK further down the line (in major releases).

As Objective-C is the past and Swift is the future, my approach sets focus on refactoring the main SDK to Swift, but still offering an (backwards-compatibility) adapter for Objective-C++ projects.

> The referenced issue (https://github.com/getsentry/sentry-cocoa/issues/6342) shows the objc version of the SDK starts with import Sentry.

I believe this was already discovered and discussed [in this PR comment](https://github.com/getsentry/sentry-cocoa/pull/7598#discussion_r3194268558).

I originally built this wrapper SDK to use the same naming too, so it's a direct drop-in replacement. I assumed that the embedded `Sentry.framework` would be internal only, but both export a class with the same ObjC runtime name which _could_ cause duplicate class conflicts at runtime:

- `SentryObjC.framework/SentryObjC`: `_OBJC_CLASS_$_SentrySDK` is the ObjC class from `SentrySDK.m`
- `SentryObjC.framework/Frameworks/Sentry.framework/Sentry`: `_OBJC_CLASS_$__TtC6Sentry9SentrySDK` is the Swift class, mangled symbol, but registers as `SentrySDK` in the ObjC runtime via `@objc`

Since `SentryObjC` is the new SDK, I renamed its class to `SentryObjcSDK` to avoid the collision.

> What this PR does though is something quite different, it introduces a new target that compiles the same source files as the current SDK but with different settings. 

This claim is only partially true, because it introduces three new targets (`SentryObjC`, `SentryObjCBridge`, `SentryObjCTypes`) but each with their own files, and it only reuses some of the public headers of the base SDK to deduplicate code.

What the PR actually does is create a pure Objective-C wrapper layer around the existing Swift SDK. The new code acts as a facade/bridge so that ObjC/ObjC++ consumers who can't use Clang modules can still access the full SDK API. The new targets depend on the existing SDK but don't recompile it, they call into it via @objc bridge methods and re-export its existing pure-ObjC headers.

> IMO this is going to be much more difficult to maintain than what I proposed, and in particular make it harder to migrate more of the SDK to Swift.

The maintenance overhead is real, I agree with you there, especially with the hand-written ObjC headers that need to stay in sync with the Swift SDK. But the alternative proposal would face the same bridging complexity for any Swift-only type (structs, enums with associated values, generics). The Metrics API, Logger API, and SentryAttributeContent all require manual ObjC-compatible wrappers regardless of architecture.

Regarding making it harder to migrate, I must disagree, as my approach decouples ObjC consumers from the internal SDK implementation as described above. If we migrate an internal ObjC type to a Swift struct or enum, only the bridge mapping code needs updating, the consumer-facing ObjC headers don't change.

> An ObjC wrapper SDK would involve 0 changes to existing code files (objc and swift files) and just add one new target to the Package.swift file (SentryObjC - with a dependency on SentrySPM).

I have to disagree. The Metrics API is written in pure Swift using structs and generics. It is not compatible with Objective-C and will never be, hence the bridging logic which introduces an Objective-C compatible API. **This is the main reason why we are building an Objective-C SDK right now, because [sentry-unreal](https://github.com/getsentry/sentry-unreal/pull/1352) needs it.**

> I would really encourage trying this option since it will make long term maintenance much easier. This PR entangles the two distributions (swift public API and ObjC API) and my suggestion separates these two concerns into distinct layers where the ObjC API builds on the Swift API. 

I don't believe this is is correct with regards to this pull request.

We are only reusing the public ObjC headers to reduce code duplication. We use `@objc(SentryX)` annotations to pin the ObjC runtime name so that the hand-written headers in Sentry resolve correctly at link time. They don't constraint future Swift refactoring as we can rename them in Swift and migrate to the ObjC SDK if necessary.

> A layered approach like that with a clear boundary between the two is usually easier to work with than try to support two modes like this PR is doing.

My architecture is also creating two distinct layers (with a helper layer in-between): an outer Objective-C SDK wrapping a bridging layer wrapping the Swift(+ObjC) SDK. The boundary is enforced at the module level when `SentryObjCBridge` imports the SDK as an external dependency, and `SentryObjC`'s `.m` files only see the bridge through a typed forward declaration. Changes to SDK internals that don't affect the `@objc`-exposed surface do not produce changes in the wrapper layer. It is **not two modes** of the same SDK.

I hope this clear things up and that this is not blocking a soonish merge and release.

CHANGELOG.md
```diff
@@ -16,6 +19,8 @@
 ### Features
 
 - Add Set conformance to SentryAttributeValue (#7876)
+- Add SentryObjC wrapper SDK to provide pure Objective-C compatibility layer that wraps the main Sentry SDK with stable, documented Objective-C interfaces. (#7598)
+  This SDK is designed for projects that require strict Objective-C compatibility without Swift dependencies. It ships as two xcframework variants — `SentryObjC-Static.xcframework.zip` and `SentryObjC-Dynamic.xcframework.zip` — and as a compile-from-source SPM library under the `SentryObjC` product.
```
```suggestion
> [!IMPORTANT]
> The new SentryObjC SDK introduced in this release should be considered experimental and may be subject to breaking changes.

### Fixes

- Fix race conditions in scope observer iteration and propagation context locking (#7897)
- Add SentryObjC wrapper SDK to provide pure Objective-C compatibility layer that wraps the main Sentry SDK with stable, documented Objective-C interfaces. (#7598)
  This SDK is designed for projects that require strict Objective-C compatibility without Swift dependencies. It ships as two xcframework variants — `SentryObjC-Static.xcframework.zip` and `SentryObjC-Dynamic.xcframework.zip` — and as a compile-from-source SPM library under the `SentryObjC` product.

## 9.13.0

### Fixes

- Fix crash with AVAssetDownloadTask in urlSessionTask:setState: (#7891)
- Prevent crash-loop from malformed recrash reports (#7892)

### Features

- Add Set conformance to SentryAttributeValue (#7876)
```

CHANGELOG.md
```diff
@@ -2,9 +2,21 @@
...
+> [!IMPORTANT]
+> The new SentryObjC SDK introduced in this release should be considered experimental and may be subject to breaking changes.
+
+### Fixes
+
+- Fix race conditions in scope observer iteration and propagation context locking (#7897)
+
+> [!IMPORTANT]
+> The new SentryObjC SDK introduced in this release should be considered experimental and may be subject to breaking changes.
+
```
philprime, line 12:
```suggestion
> [!IMPORTANT]
> The new SentryObjC SDK introduced in this release should be considered experimental and may be subject to breaking changes.

```

noahsmartin:
Thanks @philprime I believe we are in alignment with all the technical points it's just the public API and some subjective decisions around architecture that we don't have the same opinions on. Concretely here is how I see the differences in our approach:

My proposal is to have a wrapper SDK that only exposes an objc API. The architecture would be like this:
```
┌─────────────────────────────────────────────┐
│ SentryObjC                                  │
│                                             │
│ Public interface: Objective-C headers only  │
└──────────────────────┬──────────────────────┘
                       │
                       ▼
┌─────────────────────────────────────────────┐
│ SentryObjCCompat                            │
│                                             │
│ Public interface: Swift module              │
│                 + generated ObjC header     │
└──────────────────────┬──────────────────────┘
                       │
                       ▼
┌─────────────────────────────────────────────┐
│ Sentry                                     │
│                                             │
│ Public interface: Swift module only         │
└─────────────────────────────────────────────┘
```
In reality the `Sentry` target currently contains a mix of ObjC and Swift but the other two would be added and we would slowly move towards `Sentry` being unencumbered by ObjC code (metrics is an example of this).
Even more concretely, my proposal is to add a new library to the Package.swift like this:
```
.library(name: "SentryObjC", targets: ["SentryObjC"])
```
And targets like this:
```
    .target(
        name: "SentryObjCCompat",
        path: "Sources/SentryObjCCompat",
        publicHeadersPath: "Public",
    .target(
        name: "SentryObjC",
        dependencies: ["SentryObjCCompat"],
        path: "Sources/SentryObjC",
        publicHeadersPath: "Public",
    )
```

Your proposal does not include a "wrapper" SDK in the traditional sense. It does not wrap the existing SDK, it creates a new library that includes the existing targets (seen in your changes to Package.swift). It also would create a new "mode" of compiling the SDK where the module name would be changed to `SentryObjC` and this `SentryObjC` would include both public Swift and ObjC code. You can see this by your changes to the files in `Sentry/Public`, which now need different imports because the same file is being compiled with different names.This is a materially different public API than the ObjC only wrapper I am proposing.

These are different approaches to solve the same problem. However, I would suggest that the method I'm proposing would be easier to understand (in fact it sounds a lot like what you proposed in the PR description), easier to maintain, and easier for users. The key technical differences would be that my proposal is entirely a wrapper - no changes would be made to existing source files. You could validate this by making my proposal in an entirely new GitHub repo, the resulting framework would be identical to the one you'd make if you implemented my proposal in this repo. I am **not** proposing a new repo (too much maintenance burden). Rather just a thought experiment to demonstrate some clear separation of concerns. This is why I believe (subjectively) that my proposal would be easier, it provides clear boundaries between systems which generally makes it easier to work with the codebase.

I would defer to @itaybre and @philprime to evaluate the subjective pros/cons of these approaches. All I ask that if we go with the library that provides both objc and swift targets in one, we keep https://github.com/getsentry/sentry-cocoa/issues/6342 open as a feature request to provide a target that only has publicly visible objc code, with no swift module and no auto-generated objc code from swift `@objc` annotations.

philprime:
Superseded by #7918

PR #7601, opened 2026-03-05
ref: Use SentrySpan protocol in more places

philipphofmann:
## Summary
Follow-up to #7599. Use `id<SentrySpan>` protocol instead of `SentrySpanInternal *` in more places where only protocol methods are used.

#skip-changelog


Sources/Sentry/SentryCoreDataTracker.m
```diff
@@ -137,7 +136,7 @@ - (void)addExtraInfoToSpan:(SentrySpanInternal *)span withContext:(NSManagedObje
     }
 
     SentryStacktrace *stackTrace = [_threadInspector stacktraceForCurrentThreadAsyncUnsafe];
-    [span setFrames:stackTrace.frames];
+    [((SentrySpanInternal *)span) setFrames:stackTrace.frames];
```
philprime: `h`: Is this cast safe? What if it's not a SentrySpanInternal?

philippsuess: Ah, no, good catch. How did I miss this. Removed

Issue #6883, opened 2025-11-24
Capture app start by default


noahsmartin:
Currently app startup time is only capture by the SDK when a ui.load transaction is created within 5 seconds of app start. Instead we could always capture app startup time and report it to Sentry, even if the user hasn't added any other spans/transactions

philippsuess:
One idea from @noahsmartin is to structure sending app start transactions the same way an external user would interact with the SDK: keep all AppStartData collection outside of the SDK and use only the public APIs to create the AppStack transaction. The main goal is to clearly decouple the data collection logic from the SDK internals.

philippsuess:
## Mobile Vitals: Support standalone app start transactions

While migrating app start transactions from UIViewController-based transactions (that contain app start data as child spans) to standalone transactions with `transaction.op:app.start.cold` / `transaction.op:app.start.warm`, I found that the Mobile Vitals page in the [sentry repo](https://github.com/getsentry/sentry) doesn't pick up these new transactions. The root cause is that all Mobile Vitals queries hard-filter on `transaction.op:[ui.load,navigation]`, which excludes the new `app.start.*` ops. We need changes in the sentry repo to expand this filter.

We don't want to reuse `ui.load` for standalone app start transactions (confusing — `ui.load` semantically represents a screen load) or structure them as a root span wrapping child app start spans (confusing in the trace view — duplicates the problem we're moving away from).

[Example standalone app start transaction that doesn't show up in Mobile Vitals](https://sentry-sdks.sentry.io/insights/mobile/summary/trace/84d7b219c3374590bfc0f5fe46883c4a/?node=span-b27323e5d08649ba&project=5428557&query=transaction.op%3Aapp.start.warm&referrer=insights-mobile-overview&source=performance_transaction_summary&statsPeriod=1h&timestamp=1773048692&transaction=App%20Start%20Warm)

<details>
<summary>Detailed context and required changes</summary>

### Why it doesn't work

The Mobile Vitals page hard-filters all queries on:

```
transaction.op:[ui.load,navigation]
```

Standalone app start transactions use `transaction.op:app.start.warm` or `transaction.op:app.start.cold`, which don't match this filter, so they're completely excluded from all vitals calculations (Avg. Cold App Start, Avg. Warm App Start, and the App Starts table).

### What needs to change in the sentry repo

The `transaction.op` filter needs to be expanded to include `app.start.cold` and `app.start.warm` in the following locations:

**Frontend - Insights Mobile Views:**
[...]

PR #7602, opened 2026-03-05
fix: Capture transactions during launch profiling window

philipphofmann:
## :scroll: Description

When launch profiling with trace lifecycle is active, `sentry_stopProfilerDueToFinishedTransaction` discarded **all** transactions finishing during the launch window — not just the launch tracer itself. This prevented any concurrent transaction (e.g., manual transactions or app start transactions) from being captured and sent to Sentry.

The fix checks whether the finishing transaction is the launch tracer. Only the launch tracer is discarded (intentional, via `sentry_stopAndDiscardLaunchProfileTracer`); all other transactions are forwarded to the hub for capture.

Discovered while working on standalone app start transactions (#6883).

The launch tracer exists solely to drive profiling and is intentionally discarded — its transaction should never be sent. However, the early return in `sentry_stopProfilerDueToFinishedTransaction` was too broad: it caught every transaction that finished while `sentry_profileConfiguration` was still set during launch, silently dropping them.

Unit tests

## :pencil: Checklist

Agent transcript: https://claudescope.sentry.dev/share/Vwu89XdjtVlDOljOcvjOlrZJHIJOBNOtlqQhpy6oli4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Capture transactions that finish during the launch profiling window instead of silently discarding them (#7602)
+
 ## 9.6.0
 
 ### Features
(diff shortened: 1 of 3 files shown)


philprime (approved):
Great catch, LGTM!

PR #7637, opened 2026-03-07
test: Add crash when SentryInitializeForGettingSubclassesNotCalled is…

### LINKED ISSUE #7638: test: Add crash when SentryInitializeForGettingSubclassesNotCalled is…
> [!NOTE]
> The pull request "[test: Add crash when SentryInitializeForGettingSubclassesNotCalled is…](https://github.com/getsentry/sentry-cocoa/pull/7637)" was created by @noahsmartin but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

I ran into a flaky test that failed the assertion with `SentryInitializeForGettingSubclassesCalled`. So something was calling this "initialize" even though we expected it not to be called. I'm guessing some other test codepath is calling it and causing a race condition. With this assert we should be able to get a crash and a stacktrace if it happens again to debug what caused it

noahsmartin:
I ran into a flaky test that failed the assertion with `SentryInitializeForGettingSubclassesCalled`. So something was calling this "initialize" even though we expected it not to be called. I'm guessing some other test codepath is calling it and causing a race condition. With this assert we should be able to get a crash and a stacktrace if it happens again to debug what caused it

Closes #7638

### DIFF
--- a/Tests/SentryTests/Integrations/Performance/SentryInitializeForGettingSubclassesNotCalled.m
+++ b/Tests/SentryTests/Integrations/Performance/SentryInitializeForGettingSubclassesNotCalled.m
@@ -5,6 +5,7 @@
 
 + (void)initialize
 {
+    NSAssert(false, @"This class should never be initialized");
     if (self == [SentryInitializeForGettingSubclassesNotCalled self]) {
         _SentryInitializeForGettingSubclassesCalled = YES;
     }


Tests/SentryTests/Integrations/Performance/SentryInitializeForGettingSubclassesNotCalled.m
```diff
@@ -5,6 +5,7 @@ @implementation SentryInitializeForGettingSubclassesNotCalled
 
 + (void)initialize
 {
+    assert(false);
```
philippsuess: `m`: Could we also use NSAssert and adding a assertion message? This would add more context when this occurs? Or maybe even better, using an XCTAssert or XCFail?. The major downside of NSAssert and assert is that the entire process gets terminated.

noahsmartin: I was thinking the entire process getting terminated is what we wanted here, so that we get a backtrace and can identify the root cause

philippsuess: OK, but how does that surface in CI? If you know immidiately what went wrong, then good. Otherwise, it could be confusing.

noahsmartin: I don't know for sure because haven't been able to trigger it (otherwise I'd just open the PR to fix the root cause!) But I believe it will be visible as a backtrace in the xcresult and an error in CI

PR #7655, opened 2026-03-09
ref: Extract getAppStartMeasurement into SentryAppStartMeasurementProvider

### LINKED ISSUE #7661: ref: Extract getAppStartMeasurement into SentryAppStartMeasurementProvider
> [!NOTE]
> The pull request "[ref: Extract getAppStartMeasurement into SentryAppStartMeasurementProvider](https://github.com/getsentry/sentry-cocoa/pull/7655)" was created by @philipphofmann but did not reference an issue. Therefore this issue was created for better visibility in external tools like Linear.

## Description

Extract `getAppStartMeasurement` from `SentryTracer` into a standalone `SentryAppStartMeasurementProvider` class. Pure refactoring — no logic changes.

Small additional cleanup: removed the `profilerReferenceID` parameter from the provider API since it was only used in a debug log message and not needed for the logic.

#skip-changelog

philipphofmann:
## Description

Extract `getAppStartMeasurement` from `SentryTracer` into a standalone `SentryAppStartMeasurementProvider` class. Pure refactoring — no logic changes.

Small additional cleanup: removed the `profilerReferenceID` parameter from the provider API since it was only used in a debug log message and not needed for the logic.

The next step is to convert this logic, actually, to Swift. First did the extraction in Objective-C, so it's easy to review then. 

#skip-changelog

Closes #7661

### DIFF
--- a/Tests/SentryTests/Integrations/Performance/AppStartTracking/SentryAppStartMeasurementProviderTests.swift
+++ b/Tests/SentryTests/Integrations/Performance/AppStartTracking/SentryAppStartMeasurementProviderTests.swift
@@ -1,7 +1,7 @@
 @_spi(Private) @testable import Sentry
 import XCTest
 
-#if os(iOS) || os(tvOS)
+#if os(iOS) || os(tvOS) || os(visionOS)
 
 class SentryAppStartMeasurementProviderTests: XCTestCase {
 
@@ -304,4 +304,4 @@ class SentryAppStartMeasurementProviderTests: XCTestCase {
     }
 }
 
-#endif // os(iOS) || os(tvOS)
+#endif // os(iOS) || os(tvOS) || os(visionOS)


Sources/Sentry/SentryAppStartMeasurementProvider.m
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * The maximum amount of seconds the app start measurement end time and the start time of the
+ * transaction are allowed to be apart.
+ */
+static const NSTimeInterval SENTRY_APP_START_MEASUREMENT_DIFFERENCE = 5.0;
+
+static NSObject *appStartMeasurementLock;
+static BOOL appStartMeasurementRead;
+
+@implementation SentryAppStartMeasurementProvider
```
itaybre, line 25: `l`: Any reason not to write it in Swift?

philippsuess: Oops, sorry, I forgot to mention that first I wanted to extract it in ObjC, and the next step is to convert this to Swift.

Issue #2601, opened 2023-01-13
`SentrySDK.addBreadcrumb` crashes


sindresorhus:
### Platform

macOS

### Installed

Swift Package Manager

### Version

7.31.2

### Steps to Reproduce

I cannot reproduce. It's a crash report from a user.

### Expected Result

I expected the Sentry SDK to never crash my app.

### Actual Result

Crash log:

Also viewable at: https://sentry.io/organizations/sindresorhus/issues/3870221461/events/6fa32daaf73d409db9ee5ebe02d3f29c/

```
OS Version: macOS 13.1.0 (22C65)
Report Version: 104

Exception Type: EXC_BAD_ACCESS (SIGBUS)
Exception Codes: BUS_NOOP at 0x00000000deadbeef
Crashed Thread: 7

Application Specific Information:
isEqual: >
Attempted to dereference garbage pointer 0xdeadbeef.

Thread 7 Crashed:
0   libobjc.A.dylib                 0x30b479808         objc_msgSend
1   CoreFoundation                  0x30bc52f70         -[__NSDictionaryM setObject:forKey:]
2   Velja                           0x204537d70         -[SentryBreadcrumb serialize] (SentryBreadcrumb.m:33)
3   Velja                           0x2045382e0         -[SentryBreadcrumb description] (SentryBreadcrumb.m:85)
4   Foundation                      0x30da7a288         _NSDescriptionWithLocaleFunc
5   CoreFoundation                  0x30bc46d60         __CFStringAppendFormatCore
6   CoreFoundation                  0x30bd84d1c         _CFStringCreateWithFormatAndArgumentsReturningMetadata
7   CoreFoundation                  0x30bc4420c         _CFStringCreateWithFormatAndArgumentsAux2
8   Foundation                      0x30da8857c         +[NSString stringWithFormat:]
9   Velja                           0x20456d994         -[SentryScope addBreadcrumb:] (SentryScope.m:129)
10  Velja                           0x2045551f0         -[SentryHub addBreadcrumb:] (SentryHub.m:499)
11  Velja                           0x20456bbc0         +[SentrySDK addBreadcrumb:] (SentrySDK.m:337)
12  Velja                           0x204490860         SSApp.addBreadcrumb (SSKit.swift:36392)
13  Velja                           0x20450a330         OpenController.openURLs
14  libswift_Concurrency.dylib      0x41e94a4c4         swift::runJobInEstablishedExecutorContext

Thread 0
[...]

Issue #7643, opened 2026-03-09
SIGABRT Crash when constructing the HTTP request.


qgewfg:
### Platform

macOS

### Environment

Production

### Installed

CocoaPods

### Version

8.53.1

### Xcode Version

26.1.1

### Steps to Reproduce

SIGABRT 
4&ChannelID=8000&timestamp=1771664961155&ost=1&btea=c36e9cd28922e99fe579e9bf00b5< > SHPlayer(703,0x70000d935000) malloc: Incorrect checksum for freed object 0x7ffc59c12a88: probably modified after being freed. Corrupt value: 0x0 > Rogue memory write has corrupted memory.

### Expected Result

Normally Run.

### Actual Result

Crashed.

``` crash
OS Version: macOS 10.15.8 (19H2036)
Report Version: 104

Exception Type: EXC_CRASH (SIGABRT)
Crashed Thread: 22

Application Specific Information:
4&ChannelID=8000&timestamp=1771664961155&ost=1&btea=c36e9cd28922e99fe579e9bf00b5< > SHPlayer(703,0x70000d935000) malloc: Incorrect checksum for freed object 0x7ffc59c12a88: probably modified after being freed.
Corrupt value: 0x0
 >
Rogue memory write has corrupted memory.

Thread 22 Crashed:
0   libsystem_kernel.dylib          0xfffed5a1032a      __pthread_kill
1   libsystem_c.dylib               0xfffed58a6808      abort
2   libsystem_malloc.dylib          0xfffed5b0a50b      malloc_vreport
3   libsystem_malloc.dylib          0xfffed5b1bd27      malloc_zone_error
4   libsystem_malloc.dylib          0xfffed5afef1b      tiny_malloc_from_free_list
5   libsystem_malloc.dylib          0xfffed5afe297      tiny_malloc_should_clear
6   libsystem_malloc.dylib          0xfffed5afd0c6      szone_malloc_should_clear
7   libsystem_malloc.dylib          0xfffed5aff18b      nanov2_realloc
8   libsystem_malloc.dylib          0xfffed5aff09f      malloc_zone_realloc
9   libsystem_malloc.dylib          0xfffed5afde64      realloc
10  CoreFoundation                  0xfffe61c68950      __CFSafelyReallocate
11  Foundation                      0xfffe66a232df      _NSMutableDataGrowBytes
12  Foundation                      0xfffe66a03fe1      -[NSConcreteMutableData appendBytes:length:]
[...]

PR #7665, opened 2026-03-10
fix: Make SentryBreadcrumb thread-safe to prevent crashes in addBreadcrumb

denrase:
## :scroll: Description

SentryBreadcrumb properties were nonatomic with no synchronization, and the data property used strong instead of copy. This caused EXC_BAD_ACCESS crashes when serialize read properties concurrently with writes, or when mutable objects nested in data were mutated while sentry_sanitize iterated them.

- Add @synchronized(self) to all property accessors and serialize
- Deep-copy data in the setter so the SDK never holds references to caller-owned mutable collections
- Add defensive [copy] in sentry_sanitize and sanitizeArray before iterating
- Replace SENTRY_LOG_DEBUG(@"Add breadcrumb: %@", crumb) with category/message to avoid triggering full serialization on every addBreadcrumb

Fixes https://github.com/getsentry/sentry-cocoa/issues/2601

Reproduced issue with tests. After changes, tests are passing.

## :pencil: Checklist

### DIFF
--- a/.github/workflows/analyze-language-trends.yml
+++ b/.github/workflows/analyze-language-trends.yml
@@ -37,7 +37,7 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y cmake pkg-config libicu-dev
 
       - name: Setup Ruby
-        uses: ruby/setup-ruby@6ca151fd1bfcfd6fe0c4eb6837eb0584d0134a0c # v1.290.0
+        uses: ruby/setup-ruby@4eb9f110bac952a8b68ecf92e3b5c7a987594ba6 # v1.292.0
         with:
           bundler-cache: true
 
--- a/.github/workflows/assemble-xcframework-variant.yml
+++ b/.github/workflows/assemble-xcframework-variant.yml
@@ -71,7 +71,7 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
       - name: Setup Ruby
-        uses: ruby/setup-ruby@6ca151fd1bfcfd6fe0c4eb6837eb0584d0134a0c # v1.290.0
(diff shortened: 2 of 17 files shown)


philippsuess:
This could also fix https://github.com/getsentry/sentry-cocoa/issues/7643

Sources/Sentry/SentryBreadcrumb.m
```diff
@@ -7,8 +7,154 @@
...
+/**
+ * Recursively copies a value, producing immutable collections at every level.
+ * For @c NSDictionary / @c NSArray (including mutable subclasses) a new immutable copy is
+ * returned with all nested containers copied as well. Other values are returned as-is.
+ * This prevents the SDK from holding references to caller-owned mutable objects whose
+ * concurrent mutation would crash serialization (see
+ * https://github.com/getsentry/sentry-cocoa/issues/2601).
+ */
+static id
+sentry_deepCopyValue(id value)
```
philprime, line 19: `l`: Could this be standardized using the `NSCopying` protocol? I am not sure if it's considered legacy yet.

denrase: Checked the docs, as far as i can tell NSCopying methods only give one level of deep copies.

> The [copy(with:)](https://developer.apple.com/documentation/foundation/nscopying/copy(with:)) method performs a shallow copy. If you have a collection of arbitrary depth, passing [true](https://developer.apple.com/documentation/Swift/true) for the flag parameter will perform an immutable copy of the first level below the surface. If you pass [false](https://developer.apple.com/documentation/Swift/false) the mutability of the first level is unaffected. In either case, the mutability of all deeper levels is unaffected.

There is a core foundation method, but it only works with plist-safe values, so it's not useful for us, as we support arbitrary objects in the dictionary. If `CFPropertyListCreateDeepCopy` encounters any non-plist-safe object, the whole operation returns null.

> https://developer.apple.com/documentation/corefoundation/cfpropertylistcreatedeepcopy(_:_:_:))

PR #7510, opened 2026-02-24
fix: Capture instance and class method `[NSApplication _crashOnException]` exceptions

denrase:
## :scroll: Description

Fix macOS NSException crash reports showing _crashOnException: in the stacktrace instead of the actual exception callstack.

- Replace async captureCrashOnException: with synchronous uncaughtExceptionHandler so the crash report is written before abort() terminates the process
- Swizzle both class and instance _crashOnException: to cover different macOS versions
- Deduplicate capture: reportException: internally calls _crashOnException: (when NSApplicationCrashOnExceptions is YES, which Sentry sets), so both paths now go through SentryNSExceptionCaptureHelper which ensures a single capture
- Remove no unused code: captureCrashOnException:, SentryUseNSExceptionCallstackWrapper

Fixes both the subclass approach (SentryCrashExceptionApplication) and the swizzle approach (enableUncaughtNSExceptionReporting).

Closes #6620

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #7672, opened 2026-03-11
fix(macos): Fix thread safety in NSException capture

philprime:
## Description

Adds thread-safe atomic operations to the `SentryNSExceptionCaptureHelper` class introduced in #7510, fixing a race condition in macOS exception handling.

## Problem

PR #7510 recently refactored macOS exception handling and introduced `SentryNSExceptionCaptureHelper` with a deduplication flag to prevent duplicate captures when `_crashOnException:` is called from within `reportException:`.

However, the flag uses a simple `static BOOL`, which is not thread-safe. When multiple threads throw exceptions simultaneously, race conditions can occur:

1. Thread A sets flag = YES, starts capturing exception
2. Thread B also sets flag = YES, starts capturing same exception  
3. Both threads can bypass the deduplication check
4. Duplicate reports are sent

## Solution

Replace `static BOOL` with C11 atomic operations:
- `static BOOL _insideReportException` → `static _Atomic BOOL _insideReportException`
- Direct assignment → `atomic_store(&_insideReportException, YES/NO)`
- Direct read → `atomic_load(&_insideReportException)`

This ensures thread-safe read-modify-write operations on the deduplication flag.

## Changes

- **Modified**: `SentryNSExceptionCaptureHelper.m`
  - Add `#import <stdatomic.h>`
  - Change flag to `_Atomic BOOL`
  - Use `atomic_store()` and `atomic_load()` for all accesses
  - Add comment explaining thread safety

## Testing

- ✅ All existing exception handling tests passing
- ✅ Tests verify atomic flag prevents duplicates during nested calls
- ✅ macOS build and test suite passes

## Related

Follow-up to #7510 which introduced the helper class.


denrase (commented):
The initial issue here was that `_crashOnException` can now also be called individually, and not internally by `reportException`. So we additionally swizzle it.

But then in the case of reportException calling _crashOnException, we need to guard against it, that's why we have the flag.

As for concurrency, I can't find any documentation pointing to if this needs to be thread-safe, so I can't say if these NSApplication methods are only being called form the main thread.

If we'd want to do this, we'd need to make sure the individual threads have the interleaving protection, which i don't think we have with the atomic? Quick search turned up `_Thread_local `, but I am not familiar with it and we have not used it in the codebase.

```Objective-c
static _Thread_local BOOL _insideReportException = NO;
```

philipphofmann (commented):
`atomic_store`/`atomic_load` are called in the `_crashOnException:` path, which runs right before `abort()`. C11 atomics are **not** on POSIX's [async-signal-safe functions list](https://man7.org/linux/man-pages/man7/signal-safety.7.html).

Some findings from investigating with Claude Code (still needs validation):
- `ATOMIC_BOOL_LOCK_FREE` should be `2` (always lock-free) on Apple arm64/x86_64, since `_Bool` is 8 bits and Clang's `MaxAtomicInlineWidth` is 128 bits on both architectures ([Clang source](https://github.com/llvm/llvm-project/blob/main/clang/lib/Frontend/InitPreprocessor.cpp), [LLVM Atomics Guide](https://llvm.org/docs/Atomics.html))
- If lock-free, these should compile to single CPU instructions (`MOV`/`XCHG` on x86, `LDAR`/`STLR` on ARM) — no locks, no library calls

Can you double-check that it's safe to call C11 atomics in this context? If you already verified this, great — just want to make sure. cc @nicklama as they might know more about the crash handler specifics.

Sources/Sentry/SentryNSExceptionCaptureHelper.m
```diff
@@ -5,28 +5,31 @@
...
+    atomic_store(&_insideReportException, NO);
 }
 
 + (void)crashOnException:(NSException *)exception
 {
     // When called from within reportException: (i.e., [super reportException:] internally
     // dispatches to _crashOnException: when NSApplicationCrashOnExceptions is YES),
     // the exception was already captured, so skip to avoid duplicate reports.
-    if (!_insideReportException) {
+    if (!atomic_load(&_insideReportException)) {
```
philippsuess, line 32: `h`: `atomic_store`/`atomic_load` are called in the `_crashOnException:` path, which runs right before `abort()`. C11 atomics are **not** on POSIX's [async-signal-safe functions list](https://man7.org/linux/man-pages/man7/signal-safety.7.html).

Klaus believes it's OK to call these methods here, but I don't trust him:
- `ATOMIC_BOOL_LOCK_FREE` should be `2` (always lock-free) on Apple arm64/x86_64, since `_Bool` is 8 bits and Clang's `MaxAtomicInlineWidth` is 128 bits on both architectures ([Clang source](https://github.com/llvm/llvm-project/blob/main/clang/lib/Frontend/InitPreprocessor.cpp), [LLVM Atomics Guide](https://llvm.org/docs/Atomics.html))
- If lock-free, these should compile to single CPU instructions (`MOV`/`XCHG` on x86, `LDAR`/`STLR` on ARM) — no locks, no library calls


Did double-check that it's safe to call C11 atomics in this context? If you already verified this, great — just want to make sure. We could also double-check with @supervacuus.

philipphofmann (commented):
It would be great if we could add tests to validate this functionality. But yeah, I also get it if this is too hard to test.

philprime:
Closing this as it might not be relevant anymore and I would have to roll up entirely again.

PR #7681, opened 2026-03-12
docs: improve Updater.swift documentation

Sleywill:
## Summary

Three documentation improvements in `Sources/SentryDistribution/Updater.swift`:

1. **Remove duplicate blank line** in `checkForUpdate` — a stray `///` between the description and `- Parameters:` rendered as two blank lines in generated docs.

2. **Expand `install(url:)`** from a one-liner into a proper doc comment. The existing comment omitted the most important detail: **the method terminates the app process** after opening the URL, which callers need to know.

3. **Document `buildUrlForInstall` nil return** — the existing `- Returns:` didn't mention that the method returns `nil` when URL construction fails, which is important for call sites handling the optional.

No behaviour changes — documentation only.

### DIFF
--- a/Sources/SentryDistribution/Updater.swift
+++ b/Sources/SentryDistribution/Updater.swift
@@ -10,7 +10,6 @@ public final class Updater: Sendable {
 
   /// Checks if there is an update available for the app, based on the provided `params`.
   ///
-  ///
   /// - Parameters:
   ///   - params: A `CheckForUpdateParams` object.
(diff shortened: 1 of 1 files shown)


philprime:
Hi @Sleywill, thank you for the contribution, we highly appreciate it. As this documentation is intended for public audiences (our SDK users), they should be high-level and not go too deep into the actual implementation of the methods. In case you strongly believe these documentation changes solve a problem we have missed, please feel free to comment on this PR and we can discuss reopening it.

PR #7702, opened 2026-03-16
fix: Detect development builds via provisioning profile entitlement

### LINKED ISSUE #7507: isDebugBuild does not work for prebuilt binary distributions
### Platform

iOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

9.x

### Xcode Version

All

No

### Steps to Reproduce

Use SPM to install the sentry SDK and capture any data locally (using a test device not simulator)

### Expected Result

It should be sent to the "debug" environment

### Actual Result

It doesn't use "debug" environment. IIRC it would use "unknown". This happens because of:

```
static bool
isDebugBuild(void)
{
#ifdef DEBUG
    return YES;
#else
    return NO;
#endif
}
```

This will only evaluate to `NO` when a prebuilt binary is distributed

_No response_

denrase:
## :scroll: Description

Parses the get-task-allow entitlement from the embedded provisioning profile to correctly identify development-signed builds.

Previously, a Release-configuration build on a real device signed with a development profile was misclassified as "adhoc" — it now correctly returns "debug".

Closes #7507

Unit tests. Confirmed with apple docs. Tested on real device both with and without debugger attached.

IMG_7463

## :pencil: Checklist


CHANGELOG.md
```diff
@@ -8,7 +8,8 @@
 
 ### Fixes
 
-- Make SentryBreadcrumb thread-safe to prevent crashes in addBreadcrumb ([#7665](https://github.com/getsentry/sentry-cocoa/pull/7665))
+- Make SentryBreadcrumb thread-safe to prevent crashes in addBreadcrumb (#7665)
+- Detect development builds via provisioning profile entitlement (#7702])
```
philprime: `h`: I am not sure if this is a valid approach. IIRC it's possible to disable binary optimizations and use release code signing and still debug an app from Xcode. Am I missing something here?

itaybre (commented):
I don't think this completely fixes https://github.com/getsentry/sentry-cocoa/issues/7507.

This is definitely needed, but that issue, is related to this piece of code: https://github.com/getsentry/sentry-cocoa/blob/b98414228953f3fc2deea76300fc2e0d3f096a30/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m#L341

What we need is to complement the case where `DEBUG` is false with some heuristics, one of them can be the provisioning profile

denrase:
@itaybre @philprime Thank you both, I will re-evaluate this. 🙇

CHANGELOG.md
denrase: I have added an additional check if a debugger is currently attached.

denrase:
@itaybre Also added a check if a debugger is attached at check time. Do you think this is enough, or do we need additional fallbacks?

itaybre:
> @itaybre Also added a check if a debugger is attached at check time. Do you think this is enough, or do we need additional fallbacks?

This is better, I think the only additional case I would mention is: checking if the SDK is being added as a prebuilt framework or not.
Since that's where the issues come.

If an user is debugging a release build, technically it is still a release build

denrase:
@itaybre I want to make sure I get this right. You said the debugger-attached check is better, but also that debugging a release build is still a release build — those seem to conflict, not sure If we should keep this in then.

Also, any feedback on how to additionally detext if the SDK is added as a prebuilt framework apart from "get-task-allow"? Maybe "aps-environment" like in [launchpad](https://github.com/getsentry/launchpad/blob/56cf356da6c8eb3690baf77040849a1ec87ab275/src/launchpad/size/analyzers/apple.py#L388)?

itaybre:
> @itaybre I want to make sure I get this right. You said the debugger-attached check is better, but also that debugging a release build is still a release build — those seem to conflict, not sure If we should keep this in then.
> 
> Also, any feedback on how to additionally detext if the SDK is added as a prebuilt framework apart from "get-task-allow"? Maybe "aps-environment" like in [launchpad](https://github.com/getsentry/launchpad/blob/56cf356da6c8eb3690baf77040849a1ec87ab275/src/launchpad/size/analyzers/apple.py#L388)?

I think the confusion comes from an incomplete comment from my side. 
We have to two slightly different cases, which might need different handling:

1. Built from source
  - We can reliably check the `DEBUG` flag
  - In this case, I think we should only rely on `DEBUG` and ignore runtime signals like debugger attachment

2. XCFramework
  - We don’t have access to build-time flags
  - So here we have to rely on heuristics (get-task-allow, debugger attached, etc.)

Roughly what I had in mind:
- Source build → DEBUG only
- XCFramework → heuristics (debugger, entitlements, …)

This would align better with other platforms, where “debug vs release” is determined purely by compile-time macros, and avoid mixing runtime signals when we already have a reliable signal.

For XCFrameworks, we don’t have that option, so we effectively need to define “debug” ourselves.

Also, I don’t think there’s a perfect solution here—we probably just need to clearly document what we consider debug vs release when using the XCFramework.

denrase:
@itaybre I've implemented the two-path model:

 - Source builds: DEBUG macro only — no runtime heuristics
 - Prebuilt XCFrameworks: runtime heuristics (debugger attachment + get-task-allow)

To distinguish the two, the XCFramework build script now defines SENTRY_PREBUILT_FRAMEWORK=1 via GCC_PREPROCESSOR_DEFINITIONS. This gates the runtime checks so they only apply to prebuilt binaries. This is a sanity check to make this explicit.

Also documented in develop-docs/DECISIONS.md and develop-docs/BUILD.md.

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
```diff
@@ -459,17 +478,34 @@
...
+ *
+ * Source builds rely on the DEBUG macro only. Prebuilt XCFrameworks use runtime
+ * heuristics (debugger attachment, get-task-allow entitlement) since DEBUG is
+ * never set. See develop-docs/DECISIONS.md.
+ *
+ * @return "simulator", "debug", "enterprise", "adhoc", "test", "app store",
+ *         or "unknown".
+ */
 static const char *
 getBuildType(void)
```
itaybre, line 471: `m`: Can we add some tests fro this?

itaybre (approved):
Almost LGTM, all I think it missing is some tests (if possible) for `getBuildType`

I am pre approving this

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
```diff
@@ -459,17 +478,34 @@
...
 {
     if (isSimulatorBuild()) {
         return "simulator";
     }
+    // Source builds: compile-time DEBUG flag is the only signal.
     if (isDebugBuild()) {
```
noahsmartin, line 476: Can we delete this? The intention is to simplify the logic so we don't have as many corner cases. Ideally the only way to consider something debug would be through the provisioning profile and a prebuilt binary wouldn't behave differently

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
```diff
@@ -347,6 +348,24 @@
...
+/** Check if the SDK was distributed as a prebuilt XCFramework.
+ *
+ * The SENTRY_PREBUILT_FRAMEWORK macro is set during the XCFramework build
+ * (see build-xcframework-slice.sh). Source builds (SPM, CocoaPods) do not
+ * define it.
+ *
+ * @return YES if the SDK is a prebuilt XCFramework.
+ */
+static bool
+isPrebuiltFramework(void)
```
noahsmartin: I don't think we should have this, it adds another branch that is prone to edge cases, ideally we could unify on one method to detect debug

Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m
itaybre: This is what I was talking here: https://github.com/getsentry/sentry-cocoa/pull/7702#issuecomment-4119682529

I don't think there is a 100% way to indicate if a build is DEBUG or not (when using the prebuilt).
We should define that ourselves and document it because there will always be corner cases, no mater which solution we choose.
- Is a build built using DEBUG flag `debug` for sentry?
  - What if it is shipped to the appstore? it is still debug?
- Without the DEBUG flag but the debugger can be is `debug` enough for sentry?
  - This one might have optimizations
  
As stated before, as long as we document this I am fine with either solution we choose.
I already said my opinion

denrase: It's turning out harder to reach consensus here than I thought. 😅 Let's try to clarify some assumptions I made, please correct me if i'm wrong:

**isDebugBuild()**

If isDebugBuild() is true, we have a reliable signal that we are NOT in a release build, and can conclude early that we have build type `debug`.

If isDebugBuild() is false, I'm assuming we cannot make that determination, because of possible pre-build Sentry binary through XCFrameworks.

**get-task-allow**

Some documentation I looked up.

https://developer.apple.com/library/archive/technotes/tn2415/_index.html

> "The boolean value of get-task-allow determines whether Xcode's debugger can attach to the app"

https://developer.apple.com/forums/thread/118415

> Keep in mind that, like most entitlements, get-task-allow must be whitelisted by your provisioning profile, and only Development profiles have it so whitelisted.

This suggests that we definitely have a development profile, so it should be safe to mark build type as `develop` in this case. This should resolve the initial issue @noahsmartin has reported.

@noahsmartin Would you still remove the `isDebugBuild()` early return? If so, what's the reason behind this? We should never get a pre-build binary which has this true, right?

@itaybre Using only those two we'd remove the additional `isPrebuiltFramework` check, like Noah suggested. This is also what we had initially in this PR.

denrase: @noahsmartin @itaybre I have now reverted `isPrebuiltFramework` and we simply rely on `get-task-allow` to detect a debug build for a pre-build framework. If you want to add additional heuristics/checks, please feel free to take over the PR. 🙇

itaybre:
@denrase I think this is approved already, right?

Once the conversations are closed it will merge.

LGTM, we can evaluate later if we want to add more logic

PR #7704, opened 2026-03-17
fix: Move SessionTracker file I/O off the main thread

### LINKED ISSUE #5815: Move SessionTracker IO operations to a BG thread
### Description

The `SentrySessionTracker` does a couple of file IO operations on the main thread. Even though they are lightweight, we should consider moving these to a BG thread. We have to ensure that the session tracking logic doesn't end up with race conditions and we have to ensure the logic doesn't break if the app crashes.

It's worth noting that a big customer reported this issue, and we have a [customer case](https://www.notion.so/sentry/Critical-Differences-Sentry-vs-Bugsnag-Session-Tracking-2468b10e4b5d80a1b197ea0d2a9dfbaa) for this.

denrase:

Fixes #5815

- Existing tests run without issue.
- Tested with sample app (backgrounding + new session after 30s in bg threshold).

## :pencil: Checklist


Sources/Swift/Integrations/Session/SessionTracker.swift
```diff
@@ -106,13 +113,15 @@ final class SessionTracker {
     /// called due to a crash or unexpected behavior. Still, we don't want to lose such sessions and end
     /// them.
     func endCachedSession() {
-        let lastInForeground = SentryDependencyContainerSwiftHelper.readTimestampLastInForeground()
-        if lastInForeground != nil {
-            SentryDependencyContainerSwiftHelper.deleteTimestampLastInForeground()
-        }
+        dispatchQueue.dispatchAsync {
```
philprime, line 116: `h`: How is this behaving in case of crashes? Couldn't this lead to cases where the session is not properly ended? I don't have a particular case at hand right now, just a general concern

denrase: If the app crashes before the async block is executed, we pick it up on next launch when endCachedSession is re-tried.

If `willResignActive` fails to write the timestamps async due to a crash, the endCachedSession will be ended as abnormal, this is a slight behaviour change and is documented.

Issue #1562, opened 2021-12-03
C++ exception is not nicely displayed as stack lines in the UI


sindresorhus:
### Environment

SaaS (https://sentry.io/)

### Version

Sentry Cocoa 7.5.2

### Steps to Reproduce

- Get a crash report from a C++ exception when using Sentry Cocoa 7.5.2.

### Expected Result

I expected to see the stack trace lines with UI styling.

### Actual Result

Instead it just shows a stack line referencing `CPPExceptionTerminate`. The stack trace is just plain text in the error description field.

<img width="949" alt="Screen Shot 2021-12-03 at 12 57 55" src="https://user-images.githubusercontent.com/170270/144598819-0ad17f1a-5582-4e33-9f63-f929ccac741d.png">

---

Example report: https://sentry.io/organizations/sindresorhus/issues/2771169314/events/3a1f359eac1e494eae2b877e978368ad/

philipphofmann:
It could be that we need to fix this problem somewhere here
https://github.com/getsentry/sentry-cocoa/blob/229c663f6bdf225f8ef855f8355d48196a022cb8/Sources/Sentry/SentryCrashReportConverter.m#L347-L351

ainopara:
<img width="525" alt="ScreenShot-2022-03-25-3659" src="https://user-images.githubusercontent.com/1849450/160093967-59157ad6-512e-470b-b1ca-a83d94507862.png">
<img width="1318" alt="ScreenShot-2022-03-25-3660" src="https://user-images.githubusercontent.com/1849450/160094050-828c5977-4d23-438c-80ed-657259278ef9.png">

I have same issue.
When sentry sdk report a NSException caused crash, most useful information was lost. It seems an regression.
<img width="1050" alt="ScreenShot-2022-03-25-3663" src="https://user-images.githubusercontent.com/1849450/160095973-ccc12ff4-6e91-4c93-9a8c-4db471ac15bc.png">

philprime:
@philipphofmann is this issue still relevant? If not, can we close it?

sindresorhus:
Yes, it's still relevant.

denrase:
[...]

PR #7734, opened 2026-03-24
fix(macos): Don't overwrite exception value with notable addresses for AppKit _crashOnException crashes

denrase:
## :scroll: Description

When an NSException on macOS falls through to the mach exception monitor (via AppKit's _crashOnException:), enhanceValueFromNotableAddresses overwrites the exception value with "garbage" strings from registers/stack memory (e.g., 0x000000019de89d54 start + 7184). This happens for macOS apps that don't use SentryCrashExceptionApplication or have enableUncaughtNSExceptionReporting disabled.

The fix detects this specific crash pattern (EXC_BREAKPOINT with a crashed-thread frame inside AppKit) and skips the notable address enhancement, preserving the original mach exception description.

Relates to #1562

Unit test. Tested with sample app.

## :pencil: Checklist


Sources/Sentry/SentryCrashReportConverter.m
```diff
@@ -518,6 +521,36 @@ - (SentryException *)parseNSException
...
+        return NO;
+    }
+
+    // Only check top frames; AppKit is always deeper in the stack on macOS main-thread crashes.
+    NSArray *frames = [self rawStackTraceForThreadIndex:self.crashedThreadIndex];
+    NSUInteger limit = MIN(frames.count, 3u);
+    for (NSUInteger i = 0; i < limit; i++) {
+        uintptr_t addr = (uintptr_t)[frames[i][@"instruction_addr"] unsignedLongLongValue];
+        NSDictionary *image = [self binaryImageForAddress:addr];
+        if (image != nil && [image[@"name"] rangeOfString:@"AppKit"].location != NSNotFound) {
```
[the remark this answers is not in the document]
itaybre: I don't think there are images with no names, but we should cover ourselves just in case

itaybre, line 546: `m`: Are we 100% that both conditions are enough (exception type and appkit present)?
We already removed `_crashOnException` from the stacktrace, right? If not adding that could help

itaybre (commented):
Mostly looks good, I am just worried about this logic including other exceptions too

Sources/Sentry/SentryCrashReportConverter.m
denrase: That's a good point, will check if we have better heuristics for this so we only filter out unusual messages.

denrase: @itaybre After looking at this again, i'm not confident that these checks are enough to reliably dertmine if it would be better to opt-out here. I will close this for now.

denrase:
Closing because of discussion here https://github.com/getsentry/sentry-cocoa/pull/7734#discussion_r3110988664

PR #7736, opened 2026-03-24
fix(ios): Report ObjC exceptions that bypass NSSetUncaughtExceptionHandler

denrase:
## :scroll: Description

On iOS, ObjC exceptions inside dispatch callouts bypass NSSetUncaughtExceptionHandler and reach CPPExceptionTerminate via objc_terminate → std::terminate. Previously we skipped these, and the crash became SIGABRT with garbage notable addresses. Now we extract the NSException info and report it before forwarding to the original terminate handler.

The ObjC helper calls NSException properties which aren't async-signal-safe. This is more of a POC, needs review of what's safe in the terminate handler context and if this is even feasible at all.

Relates to #1562

## :pencil: Checklist

### DIFF
--- a/Samples/iOS-Swift/App/Sources/ErrorsViewController.swift
+++ b/Samples/iOS-Swift/App/Sources/ErrorsViewController.swift
@@ -88,11 +88,11 @@ class ErrorsViewController: UIViewController {
 
     @IBAction func captureNSException(_ sender: UIButton) {
         highlightButton(sender)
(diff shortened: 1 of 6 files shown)


denrase:
The assumption here was incorrect, the exception is already correctly reported, i did another repro and could not verify the bahaviour. closing.

```swift
DispatchQueue.main.async {
    let array = NSArray()
    array.object(at: 42)
}
```

Bildschirmfoto 2026-03-30 um 14 51 11

PR #7768, opened 2026-04-01
fix: add Package@swift-6.1.swift for Xcode 26 compatibility

uny:
## Problem

Xcode 26 (Swift 6.1) enforces [SE-0450 Package Traits](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md). When resolving SPM packages, it uses the version-specific manifest `Package@swift-6.1.swift` if available, and rejects packages that declare no traits with the error:

```
Disabled default traits on package 'sentry-cocoa' (Sentry) that declares no traits.
This is prohibited to allow packages to adopt traits initially without causing an API break.
```

This affects all users of sentry-cocoa 8.x on Xcode 26.

## Solution

Add `Package@swift-6.1.swift` — a version-specific manifest that SPM uses when the Swift 6.1 compiler is active (Xcode 26). This is the same approach used on the `main` branch for 9.x.

The manifest:
- Uses `swift-tools-version:6.1` (required for `traits` parameter)
- Declares `traits: [.default(enabledTraits: [])]` (satisfies SE-0450)
- Adds `swiftLanguageModes: [.v5]` (maintains Swift 5 compatibility)
- Points to the same 8.58.0 xcframework binaries (no SDK change)

The original `Package.swift` (swift-tools-version:5.3) is untouched and continues to work with older Xcode versions.

## Verification

Tested on macOS 26 + Xcode 26 (Swift 6.1):
- `swift package dump-package` ✅
- SPM resolution in Xcode 26 ✅ (no traits error)

Fixes: getsentry/sentry-kotlin-multiplatform#535

### DIFF
--- a/Package@swift-6.1.swift
+++ b/Package@swift-6.1.swift
@@ -1,0 +1,81 @@
+// swift-tools-version:6.1
+#if canImport(Darwin)
+import Darwin.C
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(MSVCRT)
+import MSVCRT
+#endif
+
+import PackageDescription
+
+let products: [Product] = [
+    .library(name: "Sentry", targets: ["Sentry", "SentryCppHelper"]),
+    .library(name: "Sentry-Dynamic", targets: ["Sentry-Dynamic"]),
+    .library(name: "Sentry-Dynamic-WithARM64e", targets: ["Sentry-Dynamic-WithARM64e"]),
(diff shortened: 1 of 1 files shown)


itaybre:
Looks like this got closed automatically

I am reopening it

itaybre:
Hey @uny,
I am continuing this PR in https://github.com/getsentry/sentry-cocoa/pull/7784 since there are some additional changes required or the release would fail.

PR #7777, opened 2026-04-06
feat: Track low power mode in device context

### LINKED ISSUE #5322: Detect low power mode
### Problem Statement

Particularly for performance related issues it would be helpful to see in the issue context if the device was in low power mode. Additionally low power mode state changes could be a useful breadcrumb

### Solution Brainstorm

Seems pretty straightforward to get the data: https://developer.apple.com/documentation/foundation/processinfo/islowpowermodeenabled

just not sure how it would be piped through

_No response_

itaybre:
## Summary
- Adds `low_power_mode` boolean to the device context, available on all event types
- Observes `NSProcessInfoPowerStateDidChangeNotification` in `SentryCrashIntegration` to keep the scope's device context in sync, ensuring the value is persisted for crashes and watchdog terminations
- Also reads the value at capture time in `SentryExtraContextProvider` for non-crash events
- Gated with `@available(macOS 12.0, *)` since the macOS deployment target is 10.14; available unconditionally on iOS, tvOS, watchOS, and visionOS

## Test plan
- [x] Added `testLowPowerMode` and `testLowPowerModeDisabled` tests for `SentryExtraContextProvider`
- [x] All existing `SentryExtraContextProviderTests` pass
- [x] All existing `SentryCrashIntegrationTests` pass (31 tests)
- [x] Builds successfully on iOS and macOS

Fixes: #5322


Sources/Swift/Helper/SentryExtraContextProvider.swift
```diff
@@ -52,6 +52,10 @@
         default:
             SentrySDKLog.warning("Unexpected thermal state enum value: \(thermalState)")
         }
+
+        if #available(macOS 12.0, *) {
+            extraDeviceContext["low_power_mode"] = NSNumber(value: processInfoWrapper.isLowPowerModeEnabled)
```
philprime, line 57: `m`: Do we need to document this field somewhere?

Sources/Swift/Integrations/SentryCrash/SentryCrashIntegration.swift
```diff
@@ -261,6 +279,29 @@ final class SentryCrashIntegration<Dependencies: CrashIntegrationProvider>: NSOb
         }
     }
 
+    @available(macOS 12.0, *)
+    @objc private func powerStateDidChange() {
+        updateLowPowerModeContext()
```
philprime: `m`: AFAICT the notification receives a ProcessInfo instance as the notification object, so I believe we should this one as best practice.

https://developer.apple.com/documentation/foundation/nsnotification/name-swift.struct/nsprocessinfopowerstatedidchange

philprime (approved):
LGTM with some additional comments to consider

Sources/Swift/Helper/SentryExtraContextProvider.swift
itaybre: Added to sentry-contentions [here](https://github.com/getsentry/sentry-conventions/pull/314)

PR #7784, opened 2026-04-09
fix: Add `Package@swift-6.1.swift` for Xcode 26 SPM compatibility

itaybre:
## Summary

This PR extends the release tooling introduced in #7768 to fully support `Package@swift-6.1.swift` alongside `Package.swift`.

- **`Package@swift-6.1.swift`**: Added inline marker comments (`//Sentry-Static`, `//Sentry-Dynamic`, etc.) to each binary target checksum line. `update-package-sha.sh` and `verify-package-sha.sh` already glob `Package@swift-*.swift` files, but rely on these comments for their `sed`/`grep` patterns.
- **`Utils/VersionBump/main.swift`**: Added `./Package@swift-6.1.swift` to the `files` array so version bump updates the download URLs in it during releases.
- **`.github/actions/prepare-package.swift/action.yml`**: Updated the action to loop over all `Package*.swift` files automatically (removed the now-unnecessary `package-file` input), consistent with `update-package-sha.sh` and `verify-package-sha.sh`.

Fixes: https://github.com/getsentry/sentry-kotlin-multiplatform/issues/535

## Test plan

- [ ] `version-bump-util` CI job covers `Package@swift-6.1.swift` via the `run_version_bump_util_for_prs` filter (`Package*.swift` glob)
- [ ] `release` CI job's `prepare-package.swift` action now processes both package files


romtsn (approved):
LGTM!

PR #7798, opened 2026-04-13
fix: Normalize profiling cpu usage to percent

### LINKED ISSUE #7456: fix: divide cpu_usage by TH_USAGE_SCALE in SentrySystemWrapper.cpuUsage()
## Problem

The `cpuUsage()` function in `SentrySystemWrapper` calculates total CPU usage by summing `threadInfo.cpu_usage` for each thread. However, it omits a necessary division by `TH_USAGE_SCALE` (1000), which is the scale factor for `thread_basic_info.cpu_usage` per Apple's Mach API.

This causes the reported CPU usage to be **1000× larger** than the actual value. For example, a true usage of 0.1% would be reported as 100%.

## Evidence

- **Apple docs**: `cpu_usage` is a scaled value; divide by `TH_USAGE_SCALE` (1000) to get percentage
- **SentryThreadHandle.cpp** (line 175): correctly uses `static_cast<float>(data.cpu_usage) / TH_USAGE_SCALE`
- **Current code**: `usage += Float(threadInfo.cpu_usage) / processorCount` — missing the scale division

## Proposed fix

```swift
usage += Float(threadInfo.cpu_usage) / Float(TH_USAGE_SCALE) / processorCount
```

`TH_USAGE_SCALE` is defined in `<mach/thread_info.h>` (1000) and is available when importing Darwin.

## Location

`Sources/Swift/Helper/SentrySystemWrapper.swift` — `cpuUsage()` method

## Note

The existing unit test `testCPUUsageReportsData` only asserts `(0.0...100.0).contains(cpuUsage.doubleValue)`. It can pass by chance when the process is nearly idle and the raw unscaled value stays under 100. Consider adding a test that verifies reasonable percentage values under load.

denrase:
## :scroll: Description

Fix SentrySystemWrapper CPU usage calculation by dividing Mach thread cpu_usage by TH_USAGE_SCALE, then normalizing by processor count and converting to 0..100 percent. Updates docs and adds test for the normalization logic.

Closes #7456

Unit Tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- fix: Normalize profiling cpu usage to percent (#7798)
+- Normalize profiling cpu usage to percent (#7798)
 - Detect development builds via provisioning profile and debugger attachment (#7702)
 
 ## 9.10.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- fix: Normalize profiling cpu usage to percent (#7798)
+
 ## 9.10.0
 
 ### Features
--- a/Sources/Sentry/SentrySystemWrapper.mm
+++ b/Sources/Sentry/SentrySystemWrapper.mm
@@ -6,6 +6,10 @@
 #    import <mach/mach.h>
 #    include <thread>
 
+@interface SentrySystemWrapper ()
++ (float)normalizeCPUUsage:(integer_t)threadCPUUsage processorCount:(long)processorCount;
+@end
+
 @implementation SentrySystemWrapper {
     float processorCount;
 }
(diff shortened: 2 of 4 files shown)


philprime (commented):
LGTM but I have one concern: Isn't this technically speaking a breaking change because the reported data and any related dashboards will report completely different numbers from now on?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- fix: Normalize profiling cpu usage to percent (#7798)
```
```suggestion
- Normalize profiling cpu usage to percent (#7798)
```

itaybre (commented):
@philprime concern seems valid, at this time we might not be able to update this untils a new major

denrase:
@philprime @itaybre Should we consider it a breaking change when it was incorrect before?

itaybre:
> @philprime @itaybre Should we consider it a breaking change when it was incorrect before?

If a user created a dashboard based on this, it would change behavior. So even if it was incorrect before, it is a breaking change (the dashboard wouldn't work with old and new data)

philprime:
@denrase after a discussion in the Team Apple SDK sync we believe that even though this is is a bug fix, it's a breaking change to any queries or dashboards built around this invalid value. Therefore we must postpone this to the next major release of v10, which is not going to happen until fall 2026

We have two options:
1. introduce a `V10` conditional compilation flag as we did back for v9, so that this change lives in the codebase, but is not included by customers
2. close the PR, leaving the issue open, and keeping the change around until we get to do it.

I believe option 2 is easier to do right now, as a conditional compilation flag would also require us to extend CI to make sure the SDK is still functional for v10 too.

denrase:
@philprime Sounds good to me, thank you! Closing for now and will re-visit once we have v10 out.

PR #7801, opened 2026-04-14
fix: Add foreground state to app context in app hang events

### LINKED ISSUE #5725: Ensure app hangs have is_foreground true
### Description

There might be a few edge cases for which the `SentryANRTrackingIntegration` captures app hang events for which `is_foreground` isn't `true`. The client sets this field here

https://github.com/getsentry/sentry-cocoa/blob/1b9991e062e58b3eca64ce9b3a4ced2a8ca0aae2/Sources/Sentry/SentryClient.m#L760-L777

The `SentryANRTrackingIntegration` could already set this field, because the SDK only reports app hangs when the app is in the foreground to ensure this field is always set. A customer reported that they get some fatal app hang events for which this field isn't set.

We fixed an issue with https://github.com/getsentry/sentry-cocoa/pull/5681 that led to wrongly reported app hangs when the app came back from the background, which should minimize the occurrence of the above-mentioned edge cases.

denrase:
## :scroll: Description

Set app.in_foreground and app.is_active when creating new app hang events in SentryHangTrackingIntegration, so persisted and fatal app hangs always carry the expected foreground context.

Closes #5725

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - Keep replayType as `buffer` for Session Replay triggered by an error (#7804)
 - Fix race condition in scope observer notifications causing EXC_BAD_ACCESS during cold launch (#7807)
 - Unsubscribe to system event during background to avoid reporting breadcrumbs with wrong timestamps on return to foreground (#7702)
-- Ensure app hangs have foreground app context (#7801)
+- Add foreground state to app context in app hang events (#7801)
 
 ## 9.10.0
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Detect development builds via provisioning profile and debugger attachment (#7702)
+- Ensure app hangs have foreground app context (#7801)
```
philprime: `l`: I think this could be rephrased to be more understandable for non-SDK-maintainers

```suggestion
- Add foreground state to app context in app hang events (#7801)
```

philprime (approved):

Issue #4580, opened 2024-11-27
Fix crash in sentrycrashjson_addFloatingPointElement


philipphofmann:
### Description

The SDK sometimes crashes with SIGPIPE or SIGABRT in `sentrycrashjson_addFloatingPointElement`

```
Exception Type: Unknown (SIGPIPE)

Thread 0 Crashed:
0   libsystem_c.dylib               0x1c6ac5f1c         __vfprintf
1   libsystem_c.dylib               0x1c6ac5ba4         _vsnprintf
2   libsystem_c.dylib               0x1c6aca818         snprintf
3   Sentry.framework                0x101810a5c         sentrycrashjson_addFloatingPointElement (Sentry.framework:324)
4   Sentry.framework                0x101812f90         encodeObject (Sentry.framework:342)
5   Sentry.framework                0x1018133dc         [inlined] encodeObject (Sentry.framework:384)
6   Sentry.framework                0x1018133dc         encodeObject (Sentry.framework:384)
7   Sentry.framework                0x101812d80         +[SentryCrashJSONCodec encode:options:error:] (Sentry.framework:427)
8   Sentry.framework                0x101824870         -[SentryCrashScopeObserver toJSONEncodedCString:] (Sentry.framework:149)
9   Sentry.framework                0x10182465c         -[SentryCrashScopeObserver addSerializedBreadcrumb:] (Sentry.framework:90)
10  Sentry.framework                0x10185fdf0         -[SentryScope addBreadcrumb:] (Sentry.framework:138)
11  Sentry.framework                0x10183db08         -[SentryHub configureScope:] (Sentry.framework:555)
12  Sentry.framework                0x101864e88         +[SentrySDK configureScope:] (Sentry.framework:406)
```

We see this also in our [internal SDK crash detection](https://sentry.sentry.io/issues/?project=4505469596663808&query=is%3Aunresolved%20addfloatingpoint&referrer=issue-list&sort=user&statsPeriod=30d&viewId=35146).

Originally reported here: https://github.com/getsentry/sentry-cocoa/issues/4579.

philipphofmann:
[...]

PR #7803, opened 2026-04-14
fix(breadcrumbs): Unsubscribe to system event during background to avoid reporting breadcrumbs with wrong timestamps on return to foreground

denrase:
## :scroll: Description

Per Apple's docs (https://developer.apple.com/documentation/uikit/processing-queued-notifications?language=objc), iOS queues notifications while an app is suspended and delivers them on resume. The time we report them, might not align witht he time they were triggered, so they should not provide diagnostic value, so we might as well not use them. This also aligns us with Android behaviour, which implemented the equivalent issue in https://github.com/getsentry/sentry-java/pull/4338.

This PR makes SentrySystemEventBreadcrumbs lifecycle-aware: unsubscribes from system event notifications on didEnterBackgroundNotification, re-subscribes on willEnterForegroundNotification. An isSubscribedToSystemEvents flag makes both operations idempotent.

Also fixes AND/OR matching bug in TestNSNotificationCenterWrapper.removeObserver(_:name:object:).

This is being done with #4580 in mind. The reported SIGPIPE crashes show snprintf in the stack trace, but snprintf writes to a stack buffer and cannot generate SIGPIPE. Our hypothesis is that the signal originates elsewhere (e.g. stale network connections breaking after resume) and is delivered asynchronously to the main thread while it happens to be encoding a breadcrumb.

SentryCrash treats SIGPIPE as fatal, so the crash report captures whatever the thread was doing at signal delivery time. So while this will in all likleyhood not fix the underlying problem, it might reduce noise so we can find the root cause.

Unit tests.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,9 @@
 ### Fixes
 
 - Detect development builds via provisioning profile and debugger attachment (#7702)
-- Unsubscribe system event breadcrumbs during background (#7702)
 - Keep replayType as `buffer` for Session Replay triggered by an error (#7804)
 - Fix race condition in scope observer notifications causing EXC_BAD_ACCESS during cold launch (#7807)
+- Unsubscribe to system event during background to avoid reporting breadcrumbs with wrong timestamps on return to foreground (#7702)
 
 ## 9.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Detect development builds via provisioning profile and debugger attachment (#7702)
+- Unsubscribe system event breadcrumbs during background (#7702)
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Detect development builds via provisioning profile and debugger attachment (#7702)
+- Unsubscribe system event breadcrumbs during background (#7702)
```
itaybre: `l`: a small change to reflect why this is a fix
```suggestion
- Unsubscribe to system event during background to avoid reporting breadcrumbs with wrong timestamps on return to foreground (#7702)
```

itaybre (approved):
LGTM, with one suggestion

PR #7804, opened 2026-04-14
fix(replay): Keep replayType as `buffer` for Session Replay triggered by an error

### LINKED ISSUE #5879: ReplayType incorrectly being identified as Session instead of Buffer
### What React Native libraries do you use?

React Native without Frameworks

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### Are you using any other error monitoring solution alongside Sentry?

Yes

### Other Error Monitoring Solution Name

_No response_

### @sentry/react-native SDK Version

7.4.0

### How does your development environment look like?

````
⬇  Place the `npx react-native@latest info` output here. ⬇




````

### Sentry.init()

````js
{
      dsn: env.SENTRY_DSN,
      release: env.APP_VERSION_NAME,
      dist: env.APP_VERSION_CODE,
      sendDefaultPii: true,
      tracesSampleRate: 0.1,
      profilesSampleRate: 0,
      replaysSessionSampleRate: 0,
      replaysOnErrorSampleRate: 0.05,
      integrations: [
        Sentry.mobileReplayIntegration({
          maskAllText: false,
          maskAllImages: false,
          maskAllVectors: false,
        }),
        Sentry.appStartIntegration(),
        Sentry.reactNativeTracingIntegration(),
        this.navigationIntegration,
        featureFlagsIntegration(),
      ],
      enableAutoSessionTracking: true,
      sessionTrackingIntervalMillis: 1000,
      enableCaptureFailedRequests: true,
      enableUserInteractionTracing: true,
      attachScreenshot: false,
      debug: false,
      enableAppStartTracking: true,
      enableNativeFramesTracking: true,
      enableStallTracking: true,
    }
````

### Steps to Reproduce

Capture session replay with init above. 

### Expected Result

Session replays without errors should not be coming in without errors. If they are they should be of replayType: buffer. 

### Actual Result

Session replays are coming in without any errors and are of replayType: Session.

romtsn:
Keep the replayType as `buffer` when Session Replay is triggered by an error.

When an on-error replay starts from a buffered session, the first replay segment is tagged as `buffer` but later segments were emitted as `session`. 

This preserves the initial replay type across later segments and keeps explicit manual promotion through the replay API as `session`. It also adds regression coverage for both the error-triggered buffer path and the explicit session path.

Example replay: https://sentry-sdks.sentry.io/explore/replays/4ea41582cd8747d98835d8c018c7182e

Closes getsentry/sentry-react-native#5879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Detect development builds via provisioning profile and debugger attachment (#7702)
+- Keep replayType as `buffer` for Session Replay triggered by an error
 
 ## 9.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixes
 
 - Detect development builds via provisioning profile and debugger attachment (#7702)
-- Keep replayType as `buffer` for Session Replay triggered by an error
+- Keep replayType as `buffer` for Session Replay triggered by an error (#7804)
 
 ## 9.10.0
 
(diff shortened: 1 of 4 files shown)


itaybre (approved):
LGTM

PR #7814, opened 2026-04-20
test: additional unit-tests for C-BIC

supervacuus:
Hi @noahsmartin, I am adding these here in isolation for our sync. These are the unit tests I added to align your implementation with mine last week. I only added a few minor debug helpers to the production code (could be gated by `SENTRY_TEST`, but that wasn't my primary concern last week). 

Only adding these here since you specifically raised unit tests as open topics. I already did quite a bit; let me know what you'd add in addition to that. Of course, for issues like these, the really interesting thing is parameterized E2E stress tests, but let's talk tomorrow.


supervacuus:
starting from a fresh branch anyway, according to our conversation, so the diffs here don't pay a lot of dividends.

PR #7817, opened 2026-04-21
fix: Support stack traces for standalone clients

### LINKED ISSUE #3409: Stack trace symbols missing after upgrading to 8.9.0 or later
### Platform

iOS

### Installed

Carthage

### Version

8.9.0

### Steps to Reproduce

We tried to update from 7.3.0 to the latest version (8.15.x) but our stacktraces don't show desymbolicated entries anymore.

After digging more, I've realized:
1. 8.9.0 is the first version that broke this
2. our setup is kind of weird

## Weird setup

I assume this is because we didn't have ways to properly disable certain features that we don't want (ie we don't want any swizzling), so our setup looks like this (using 8.8.0 which is the last version we could upgrade to):

```swift
let options = Options()
options.dsn = config.dsn
options.attachStacktrace = false
options.enableSwizzling = false
options.enableCrashHandler = false

// most of these should be covered by `enableSwizzling`, but let's be safe:
options.enableTracing = false
options.enableNetworkTracking = false
options.enableNetworkBreadcrumbs = false
options.enableFileIOTracing = false
options.enableCoreDataTracing = false
options.enableAutoPerformanceTracing = false
options.enableUserInteractionTracing = false
options.enableAutoBreadcrumbTracking = false
options.enablePreWarmedAppStartTracing = false
options.enableUIViewControllerTracing = false

sentryClient = SentryClient(options: options)
```

Note that `attachStacktrace = false`, but that's not the reason things are not working since we call the private method to attach stack trace only when we want (see https://github.com/getsentry/sentry-cocoa/issues/3386 for more context).

We also don't call `SentrySDK.start(options: options)` and instead use our own `SentryClient` (because of the requirement above). As far as I can tell, this is what breaks desymbolication.

In an ideal world, we would:
1. Have official support for configuring whether to attach or not a stacktrace to a given event (https://github.com/getsentry/sentry-cocoa/issues/2325)
2. `SentryClient` probably shouldn't be exposed publicly so people don't end up with a setup similar to ours

Given that the hints API is probably a few months away, I wonder if there's an interim solution. Here're the options I thought:

- expose the internal `SentryClient` from `SentrySDK` so that we can keep using our setup. We might be able to workaround this today by calling `[[SentrySDK currentHub] getClient]`. Obviously we would be relying even more on the internal structure of the SDK by doing this, but we might need to do this if there's no official support for what we want.
- expose a method in `SentrySDK` that basically calls `[client sendEvent:withScope: alwaysAttachStacktrace:]`
- don't do anything, keep locked at 8.8.0 and hope that the hints API comes soon

### Expected Result

Stack traces should be desymbolicated

### Actual Result

Stack traces are not desymbolicated

Note: I'm writing this issue as a more technical complement of a support email I've sent. In the email there are links to the actual events in Sentry.

_No response_

denrase:
## :scroll: Description

Standalone SentryClient / SentryHub usage is publicly documented for shared and library environments, but handled events with attached stacktraces regressed after local symbolication became debug-only. In that setup, stacktrace frame image metadata and debug_meta.images depended on the binary image cache being initialized through SentrySDK.start, which standalone clients never did.

This change starts the binary image cache during SentryClient initialization, so standalone client and hub usage once again captures the image metadata needed for handled-event stacktrace symbolication.

Closes https://github.com/getsentry/sentry-cocoa/issues/3409

Unit tests. Repro with sample app.

## :pencil: Checklist


denrase:
We probably want to wait until @supervacuus finishes up with their binary image cache PR based on https://github.com/getsentry/sentry-cocoa/pull/7269 from @noahsmartin, as it may affect how this needs to be dine here and I don't want to interfere with their implemention, I'd rather update here afterwards.

denrase:
This also moves the init away from the man callback in start, also need to check after @supervacuus PR.

itaybre:
@denrase looks like tests are still broken

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -34,6 +34,9 @@ import Foundation
     
     @objc public func start(_ isDebug: Bool) {
         lock.synchronized {
+            guard cache == nil else {
+                return
```
philprime, line 39: `l`: Please add an SDK debug log statement here, as it seems to silently fail to start otherwise.

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -58,6 +61,9 @@ import Foundation
     
     @objc public func stop() {
         lock.synchronized {
+            guard cache != nil else {
+                return
```
philprime, line 67: `l`: Same as above, please add a log statement

Sources/Sentry/SentryClient.m
```diff
@@ -133,6 +133,9 @@ - (instancetype)initWithOptions:(SentryOptions *)options
                                                    cacheDirectoryPath:options.cacheDirectoryPath
                                                        sendDefaultPii:options.sendDefaultPii];
 
+        [SentryDependencyContainer.sharedInstance.crashWrapper startBinaryImageCache];
```
philprime: `m`: Can we inject the crash wrapper and the binary image cache as constructor parameters, defaulting to the dependency container? That would be cleaner than using the singleton here

Tests/SentryTests/SentryBinaryImageCacheTests.swift
```diff
@@ -2,6 +2,12 @@
 @_spi(Private) import SentryTestUtils
 import XCTest
 
+private var externalAddedCallbackInvocations = 0
+
+private func recordExternalAddedCallback(_ image: UnsafePointer<SentryCrashBinaryImage>?) {
+    externalAddedCallbackInvocations += 1
+}
```
philprime: `l`: Please look into using our Invocations structure for this. It might be usable with a helper in the SentryBinaryImageCacheTests and would allow us to reuse testing structures

philprime (approved):
LGTM with one feedback point to address.

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -34,6 +34,9 @@ import Foundation
     
     @objc public func start(_ isDebug: Bool) {
         lock.synchronized {
+            guard cache == nil else {
+                return
+            }
```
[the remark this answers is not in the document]
denrase: options value is the same from both call sites

Sources/Swift/Core/Helper/SentryBinaryImageCache.swift
```diff
@@ -34,6 +34,10 @@ import Foundation
     
     @objc public func start(_ isDebug: Bool) {
         lock.synchronized {
+            guard cache == nil else {
+                SentrySDKLog.debug("SentryBinaryImageCache is already started. Skipping start.")
+                return
+            }
```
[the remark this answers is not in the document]
denrase: start/stop is not our concern here

PR #7821, opened 2026-04-22
ref: decouple VC swizzling from BinaryImageCache

supervacuus:
## :scroll: Description

As discussed yesterday, we first extract from the previous PR (https://github.com/getsentry/sentry-cocoa/pull/7269) the decoupling of the VC swizzling's image needs from the BIC.

I added a minor interface to inject images from the tests, so we can keep `testSwizzle_DebugDylib_GetsSwizzled`.

Two notes:

* It looks like neither `processInfoWrapper` nor `dispatchQueue` is used in `SentryUIViewControllerSwizzling`, and the initialization isn't following a particular abstraction either, so we could remove it?
* @noahsmartin, since you explicitly raised this yesterday: this approach still acquires a `dyld` reader lock on the `main` thread with every call to `_dyld_get_image_name()`, since `start()` runs as part of the performance integration setup. I assume this is acceptable here from your pov?

This is a follow-up to https://github.com/getsentry/sentry-cocoa/pull/7269, aimed at unblocking the main thread from dyld lock contention. This particular change only isolates VC swizzling from the BIC dependency, making it a sensible change in its own right, but it is still a precondition for the remaining changes.

reusing the tests already written.

## :pencil: Checklist


Sources/Swift/Core/Integrations/Performance/SentryUIViewControllerSwizzling.swift
```diff
@@ -11,6 +13,15 @@ import UIKit
 
 extension UIApplication: SentryUIApplication {}
 
+// This exists to allow tests to inject images.
```
itaybre: `l`: While technically true, I don't believe an alias is only for testing so IMO we could remove the comment and leave it as it is

Sources/Swift/Core/Integrations/Performance/SentryUIViewControllerSwizzling.swift
```diff
@@ -11,6 +13,15 @@ import UIKit
...
+private typealias SentryLoadedImageNamesProvider = () -> [String]
+
+private func defaultLoadedImageNamesProvider() -> [String] {
+    (0..<_dyld_image_count()).compactMap { i in
+        _dyld_get_image_name(i).map { String(cString: $0) }
+    }
```
itaybre:
```suggestion
    (0..<_dyld_image_count()).compactMap { index in
        _dyld_get_image_name(index).map { String(cString: $0) }
    }
```
`l`: wdyt?

itaybre (commented):
LGTM, with some comments + rerun CI due to some issues with codesigning (which should be fixed already)

itaybre (approved):
Forgot to stamp

Sources/Swift/Core/Integrations/Performance/SentryUIViewControllerSwizzling.swift
supervacuus: What I meant was that I added the `typealias` so we can inject images via arguments in tests. If we didn't do this, we could remove it. It has no meaning in the production code. Similarly, the new `init()` parameter would not exist, and the extracted `defaultLoadedImageNamesProvider()` would just sit with the rest of the `start()` code. I was mostly thinking of the next person who sees this abstraction and wonders why it has to exist.

itaybre: Ohh, I misunderstood then.

Maybe It should probably be closed to the function, not the typealias

supervacuus: fixed in: https://github.com/getsentry/sentry-cocoa/pull/7821/commits/f09f83984a633fde35ee5aac7c0f431a54b2f05e

Issue #4618, opened 2024-12-11
App Launch Time Slow：sentrycrashdl_getBinarylmageForHeader


kingnight:
### Platform

iOS

### Environment

Production

### Installed

Manually

### Version

8.37.0	

### Xcode Version

15.1

### Steps to Reproduce

![Image](https://github.com/user-attachments/assets/2e1b2ea8-cadd-4b33-bfaf-aa90fe00f2e6)

Info from Xcode Organizer：
* 100 ms spent in this code path during a 1156 ms launch
* iPhone 13
* 18.1.1 (22B91)

```
        SentrySDK.start { options in
            #if DEBUG
            options.enabled = false
            #else
            options.enabled = true
            #endif
            if AppBundle.isReleasePackage() {
                options.dsn = xxx
            }else{
                options.dsn = xxx
            }
            options.enableNetworkTracking = false
            options.enableCaptureFailedRequests = false
            options.sampleRate = 0.5
            options.tracesSampleRate = 0.3
            options.profilesSampleRate = 0.3
            options.appHangTimeoutInterval = 3
            options.beforeSend = { event in
                if let type = event.exceptions?.first?.type,type == "App Hanging" {
                    if let frames = event.exceptions?.first?.stacktrace?.frames,
                       !frames.isEmpty {
                        for item in frames.reversed() {
                            guard let function = item.function else { continue }
                            if AppDelegate.containsSentryWhiteListedStrings(function){
                                return nil
                            }
                        }
                    }
                }
                return event
            }

            SentrySDK.configureScope { scope in
                if let data:Data = ULog.shared.getActivityLog(){
                    scope.addAttachment(Attachment(data: data, filename: "crash.log"))
                    ULog.shared.removeAllActivityLog()
                }
            }
        }
```

### Expected Result

reduce app launch time

### Actual Result

23% slow launch time

kingnight:
@philipphofmann   how many binary images :  1072 

`_cache	__NSArrayM *	@"1072 elements"	0x00000003005ed260`

in debug mode

supervacuus:
> Speed up sentrycrashdl_getBinaryImageForHeader.
@supervacuus, as your native knowledge is way better than mine, do you see room for speed-ups in this method?

@philipphofmann, I can only see two immediately obvious optimization paths. 

1. The highest cost per segment iteration is the string comparisons of the segment names using `strcmp`. Using `memcmp` on the fixed 16-byte name field might be more performant. 
[...]

PR #7823, opened 2026-04-23
fix: initialize BinaryImageCache async to main thread 

supervacuus:
## :scroll: Description

This PR is a reworked version of https://github.com/getsentry/sentry-cocoa/pull/7269, following the discussed path during the sync on Tuesday.

In particular, I

* moved the entire cache state into a single struct that can now be atomically modified for tests. All functions now operate on that state.
* extracted test-only functions into their own test-side header and gated them for `SENTRY_TEST*`
* adapted and added test cases for the new implementation
* added comments to state and aligned the function naming with the remaining code
* added a one-time log, in case we actually reach the new limit of `4096` images
* removed the refactoring work in [Sources/Swift/Core/Helper/SentryBinaryImageCache.swift](https://github.com/getsentry/sentry-cocoa/pull/7269/changes#diff-cc933a3408f88e5d7b2c255cc602d8c81d9b0e2893da814e21da2e5e57c975a0), which is unrelated to this change, and so we can keep it small

It depends on the changes in https://github.com/getsentry/sentry-cocoa/pull/7821, which we discussed splitting out, but explicitly left the commit in here to allow the PR to build and run before the other PR is merged. The changes are equivalent to https://github.com/getsentry/sentry-cocoa/pull/7821 and can be ignored.

The goal is to resolve issues like https://github.com/getsentry/sentry-cocoa/issues/4618 by moving the costly callback for dyld image replay, which holds a dyld reader lock during the entire run, off the main thread.

Against the existing and added tests. 

I also ran the `test-sample-iOS-Swift-ui`, which caused `testNavigationTransaction()` to fail because `ASSERT_MESSAGE` never appears. I might need a bit of support from the maintainers to understand the path here. I could imagine this problem being caused by the new VC swizzle image initialization providing a stale snapshot. This did not fail once in CI, but on every local run (so it could be entirely a local issue).

Related to that, we currently have no real test case that doesn't run synchronously (`sentrycrashbic_startCacheImpl()` runs on the main thread for `SENTRY_TEST` and `SENTRY_TEST_CI`) that provides a bound on the async initialization and thus on the availability of the `BinaryImageCache` for any of its consumers.

## :pencil: Checklist


supervacuus:
The race mentioned in the three bot responses below is real and inherent in the current startup design.

* https://github.com/getsentry/sentry-cocoa/pull/7823#discussion_r3129366701
* https://github.com/getsentry/sentry-cocoa/pull/7823#discussion_r3129376163
* https://github.com/getsentry/sentry-cocoa/pull/7823#discussion_r3129516863

By design, the async C bootstrap publishes images, while `registerAddedCallback()` in Swift's init can concurrently replay already-cached entries. We currently have no mechanism to ensure callback delivery is "exactly once" per registration, so callback consumers may observe both drops and duplicates.

This affects callback consumers only; the canonical C cache used by crash reporting is not affected. Today, the only in-tree consumer is the Swift binary image cache, which means a Swift-side dedupe could mitigate duplicates, but not drops.

If we want to preserve the current "both paths may publish" architecture, the generic fix is generation-based delivery tracking on the entries: let both paths race, but let only the winner atomically claim delivery for the current registration. That solves the class of bugs raised above and also makes re-registration (`start()` -> `stop()` -> `start()`) robust. The tradeoff is implementation complexity.

If we only want to solve the current startup problem, a narrower option is to make the async bootstrap the single publisher for the initial image set: store the callback during Swift init, do not replay from `registerAddedCallback()` while bootstrap is in flight, and perform the one-shot replay at the end of the async bootstrap before switching to live delivery for future loads. That keeps the initial work off main and eliminates the startup race, but it does not, by itself, solve broader callback re-registration cases.

A pragmatic compromise would be to combine both: defer the initial Swift-side replay to the end of async bootstrap, eliminating the startup overlap, while introducing generation-based delivery tracking as a generic safeguard for later re-registrations.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -169,94 +265,120 @@ sentrycrashbic_shouldAddDyld(void)
...
+        return;
     }
 
-    SentryCrashBinaryImageNode *newNode = malloc(sizeof(SentryCrashBinaryImageNode));
-    newNode->available = true;
-    newNode->image = binaryImage;
-    newNode->next = NULL;
+    atomic_store_explicit(&entry->state, IMAGE_READY, memory_order_release);
 
-    return newNode;
```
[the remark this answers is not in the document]
supervacuus: repsonding here: https://github.com/getsentry/sentry-cocoa/pull/7823#issuecomment-4303916753

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -169,94 +265,120 @@ sentrycrashbic_shouldAddDyld(void)
...
+    sentrycrashbic_cacheChangeCallback callback
+        = atomic_load_explicit(&cache->addedCallback, memory_order_acquire);
+    if (callback != NULL) {
+        callback(&entry->image);
+    }
```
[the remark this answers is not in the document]
supervacuus: repsonding here: https://github.com/getsentry/sentry-cocoa/pull/7823#issuecomment-4303916753

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -13,141 +15,235 @@
...
-        imageAddedCallback(&newNode->image);
+
+    sentrycrashbic_cacheChangeCallback callback
+        = atomic_load_explicit(&cache->addedCallback, memory_order_acquire);
+
+    atomic_store_explicit(&entry->state, IMAGE_READY, memory_order_release);
+
+    if (callback != NULL) {
+        callback(&entry->image);
     }
```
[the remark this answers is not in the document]
supervacuus: repsonding here: https://github.com/getsentry/sentry-cocoa/pull/7823#issuecomment-4303916753

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -13,141 +15,235 @@
...
-        _sentry_register_func_for_add_image(CALLBACK);
-#    define sentry_dyld_register_func_for_remove_image(CALLBACK)                                   \
-        _sentry_register_func_for_remove_image(CALLBACK);
-#    define _will_add_image()                                                                      \
-        if (SentryWillAddImageCallback)                                                            \
-            SentryWillAddImageCallback();
+#    define registerDyldAddImageCallback(CALLBACK)                                                 \
+        do {                                                                                       \
+            currentCache()->registerFuncForAddImage(CALLBACK);                                     \
+        } while (0)
```
itaybre: I know this is a C++ helper to behave as a statement, but mind adding a comment regarding this (`do { } while (0)`) for future mantainers?

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -13,141 +15,235 @@
...
+    if (!atomic_exchange_explicit(&cache->didLogImageLimitReached, true, memory_order_acq_rel)) {
+        SENTRY_ASYNC_SAFE_LOG_ERROR(
+            "Binary image cache reached capacity of %d images. New images will not be tracked.",
+            MAX_DYLD_IMAGES);
     }
-    pthread_mutex_unlock(&binaryImagesMutex);
+}
+
+static void
+addImage(const struct mach_header *mh)
```
itaybre:
```suggestion
addImage(const struct mach_header *header)
```
`m`: Let's help future maintainers :)

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -13,141 +15,235 @@
...
         return;
     }
 
-    SentryCrashBinaryImage binaryImage = { 0 };
-    if (!sentrycrashdl_getBinaryImageForHeader(
-            (const void *)header, info.dli_fname, &binaryImage, false)) {
+    // Test hook: called just before adding the image
+    callWillAddImageCallback();
+
+    uint32_t idx = atomic_fetch_add_explicit(&cache->nextIndex, 1, memory_order_relaxed);
```
itaybre:
```suggestion
    uint32_t nextIndex = atomic_fetch_add_explicit(&cache->nextIndex, 1, memory_order_relaxed);
```

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -169,94 +265,120 @@ sentrycrashbic_shouldAddDyld(void)
...
+    // capture the crash.
+#if defined(SENTRY_TEST) || defined(SENTRY_TEST_CI)
+    startCacheImpl();
+#else
+    dispatch_async(dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^{ startCacheImpl(); });
+#endif
 }
 
 void
 sentrycrashbic_stopCache(void)
```
itaybre, line 253: `h`: Should we remove it them?

Sources/Swift/Core/Integrations/Performance/SentryUIViewControllerSwizzling.swift
```diff
@@ -11,6 +13,15 @@ import UIKit
...
 
+// This exists to allow tests to inject images.
+private typealias SentryLoadedImageNamesProvider = () -> [String]
+
+private func defaultLoadedImageNamesProvider() -> [String] {
+    (0..<_dyld_image_count()).compactMap { i in
+        _dyld_get_image_name(i).map { String(cString: $0) }
+    }
+}
+
```
itaybre: `h`: Leftover from the other PR?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- initialize BinaryImageCache async to main thread (#7823)
```
```suggestion
- Moved BinaryImageCache initialization to a background thread to reduce locks in  the main thread (#7823)
```
wdyt?

Sources/Swift/Core/Integrations/Performance/SentryUIViewControllerSwizzling.swift
supervacuus: No, this PR must be written on top of the other PR (see PR description), so I left the commit in there. I will rebase once #7821 is merged.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
supervacuus: We can, since this isn't an abstracted life-cycle function. However, depending on further state and sequencing (see https://github.com/getsentry/sentry-cocoa/pull/7823#issuecomment-4303916753), it could make sense to leave it in there, since we could life-cycle publish paths.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
supervacuus: Yeah, this can be much cleaner. We can get rid of these entirely without losing anything.

Sources/Swift/Core/Integrations/Performance/SentryUIViewControllerSwizzling.swift
```diff
@@ -38,19 +49,21 @@ class SentryUIViewControllerSwizzling {
         self.subClassFinder = subClassFinder
         self.imagesActedOnSubclassesOfUIViewControllers = NSMutableSet()
         self.processInfoWrapper = processInfoWrapper
-        self.binaryImageCache = binaryImageCache
         self.performanceTracker = performanceTracker
+        self.loadedImageNamesProvider = loadedImageNamesProvider
     }
 
     func start() {
```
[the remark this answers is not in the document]
supervacuus: That is assumed in this design.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
supervacuus: cleaned up in https://github.com/getsentry/sentry-cocoa/pull/7823/commits/9927bf34774bb5ed3d2359797d1a76de2877ea43

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -169,94 +294,117 @@ sentrycrashbic_shouldAddDyld(void)
...
-        SentryCrashBinaryImageNode *dyldNode = sentrycrashbic_getDyldNode();
-        tailNode = dyldNode;
-        rootNode.next = dyldNode;
-    } else {
-        tailNode = &rootNode;
-        rootNode.next = NULL;
-    }
-    pthread_mutex_unlock(&binaryImagesMutex);
-
     // During a call to _dyld_register_func_for_add_image() the callback func is called for every
```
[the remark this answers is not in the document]
supervacuus: Problem and related solutions discussed here: https://github.com/getsentry/sentry-cocoa/pull/7823#issuecomment-4303916753

Sources/Swift/Core/Integrations/Performance/SentryUIViewControllerSwizzling.swift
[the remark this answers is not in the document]
supervacuus: Yes, this is absolutely racy, but it is the implied risk introduced by the initial proposal in #7269. 

It also still takes a `dyld` lock on the main thread (via [_dyld_get_image_name()](https://github.com/apple-oss-distributions/dyld/blob/e9da5ae571b4191dcdbcfd827363f19847c84561/dyld/DyldAPIs.cpp#L295-L305)) that now could contend with our background thread that populates the `BIC` using `dyld_register_image_load_callback()` (which [holds the lock for the entirety of the replay](https://github.com/apple-oss-distributions/dyld/blob/e9da5ae571b4191dcdbcfd827363f19847c84561/dyld/DyldAPIs.cpp#L803-L820)).

It also provides a stale list of debug images for any framework loaded after startup, because the swizzler's image list is populated once and never updated.

**These risks are known and were discussed, with the conclusion that they are considered less critical than decoupling the VC swizzling setup from `BIC` initialization as an architectural measure and moving the longer-blocking `dyld_register_image_load_callback()` off the main thread.**

Notably, the bot's suggested fix is exactly the API we're moving off main in this PR, so adopting it here would reintroduce the cost the change was designed to avoid.

The reasons raised during the discussion as to why those risks are less critical were as follows:

* Incomplete/stale debug images for the VC swizzling path are not as correctness relevant as they are for `BIC` consumers
* The path is optional and can be turned off by users, whereas the `BIC` is core infrastructure to the SDK

Also, this feedback applies to #7821, not this PR. This PR only requires the changes in #7821 to work, which is why it was built on top of it.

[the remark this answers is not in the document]
philprime: Let's add this decision to the develop-docs/DECISIONS.md so we can find it again.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -1,153 +1,278 @@
...
-    if (newNode && imageAddedCallback) {
-        imageAddedCallback(&newNode->image);
+
+    sentrycrashbic_cacheChangeCallback callback
+        = atomic_load_explicit(&cache->addedCallback, memory_order_acquire);
+
+    atomic_store_explicit(&entry->state, IMAGE_READY, memory_order_release);
+
+    if (callback != NULL) {
+        callback(&entry->image);
```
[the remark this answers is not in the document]
supervacuus: Problem and related solutions discussed here: https://github.com/getsentry/sentry-cocoa/pull/7823#issuecomment-4303916753

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -169,94 +294,117 @@ sentrycrashbic_shouldAddDyld(void)
...
+
+        for (uint32_t i = 0; i < count; i++) {
+            PublishedBinaryImage *src = &cache->images[i];
+            if (atomic_load_explicit(&src->state, memory_order_acquire) != IMAGE_READY) {
+                continue;
+            }
+            callback(&src->image);
+        }
     }
 }
```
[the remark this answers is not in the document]
supervacuus: Problem and related solutions discussed here: https://github.com/getsentry/sentry-cocoa/pull/7823#issuecomment-4303916753

Sources/Swift/Core/Integrations/Performance/SentryUIViewControllerSwizzling.swift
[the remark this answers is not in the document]
supervacuus: @philprime, added here: https://github.com/getsentry/sentry-cocoa/pull/7823/commits/82adedac410a4a0c9fb20f8ba69088ab1bd0f6d6

develop-docs/DECISIONS.md
```diff
@@ -1,5 +1,25 @@
 # Decision Log
 
+## `SentryCrashBinaryImageCache` initializes off the main thread
+
+Date: April 21st 2026
+Contributors: @noahsmarting, @itaybre, @philprime, @supervacuus
```
philprime:
```suggestion
Contributors: @noahsmartin, @itaybre, @philprime, @supervacuus
```

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -169,94 +294,117 @@ sentrycrashbic_shouldAddDyld(void)
...
 void
 sentrycrashbic_registerAddedCallback(sentrycrashbic_cacheChangeCallback callback)
 {
-    imageAddedCallback = callback;
-    if (callback) {
-        pthread_mutex_lock(&binaryImagesMutex);
-        sentrycrashbic_iterateOverImages(&initialReportToCallback, callback);
-        pthread_mutex_unlock(&binaryImagesMutex);
+    SentryCrashBinaryImageCacheState *cache = currentCache();
+    atomic_store_explicit(&cache->addedCallback, callback, memory_order_release);
```
philprime, line 274: `m`: IIUC we add a potentially NULL callback to the cache. Is there a valid reason to keep NULL callbacks around? If not, shouldn't we use an early-return before this instead?


```suggestion
    if (callback == NULL) {
        return;
    }
    atomic_store_explicit(&cache->addedCallback, callback, memory_order_release);
```

philprime (approved):
LGTM with one comment

itaybre (approved):
LGTM

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
supervacuus: `NULL` is intentional here: this function unregisters the callback from the Swift subscriber:

https://github.com/supervacuus/sentry-cocoa/blob/076767fc3a828825982556055cada3e06e17b209/Sources/Swift/Core/Helper/SentryBinaryImageCache.swift#L59-L65

So writing it into the atomic is required because it means we don't call into the Swift consumer callbacks for future updates while the `SentryCrashBinaryImageCache` continues to collect.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -1,153 +1,278 @@
...
+} SentryCrashBinaryImageCacheState;
+
+static SentryCrashBinaryImageCacheState g_defaultCache = {
+#if SENTRY_HAS_DYLD_TEST_HOOKS
+    .addImageCallback = &_dyld_register_func_for_add_image,
+    .removeImageCallback = &_dyld_register_func_for_remove_image,
+#endif
+};
+#if defined(SENTRY_TEST) || defined(SENTRY_TEST_CI)
+static SentryCrashBinaryImageCacheState g_testCache = {
```
noahsmartin: It would be a bit easier to maintain and reason about the code if this was just defined in a separate test only target and tested independently from the rest of the app.

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -1,153 +1,278 @@
...
+    .addImageCallback = &_dyld_register_func_for_add_image,
+    .removeImageCallback = &_dyld_register_func_for_remove_image,
+};
+#endif
+static _Atomic(SentryCrashBinaryImageCacheState *) g_activeCache = &g_defaultCache;
```
noahsmartin, line 25: This should only be for testing purposes? I would recommend just deleting it entirely and unit testing it in an isolated way. Alternatively this could be moved into the test macro and you could use a type alias to switch between the test and non-test version at compile time (but never at runtime)

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
```diff
@@ -169,94 +294,117 @@ sentrycrashbic_shouldAddDyld(void)
...
-        SentryCrashBinaryImageNode *nextNode = node->next;
-        free(node);
-        node = nextNode;
-    }
+    // Intentionally left running. Once bootstrapped, the canonical C cache keeps tracking loaded
+    // images so crashes can still be symbolicated even if higher-level consumers temporarily stop.
+}
 
-    pthread_mutex_unlock(&binaryImagesMutex);
+// Resetting can create race conditions so should only be done in controlled test environments.
```
noahsmartin: I don't think we should risk race conditions even in tests. Why not create a new instance of `SentryCrashBinaryImageCacheState` in tests and not ever reset them or swap out the one used here?

Sources/SentryCrash/Recording/SentryCrashBinaryImageCache.c
supervacuus: I agree (also with the other comments), addressed in https://github.com/getsentry/sentry-cocoa/pull/7823/commits/870a2c5dcae48184b10f587cf7b87ce0801440bf

PR #7781, opened 2026-04-07
feat: enhance SentryUIRedactBuilder to support redaction of CALayer sublayers in iOS 26+

itaybre:
## :scroll: Description

Add's SwiftUI layers for text and images to automatically redact content on iOS 26

## :bulb: Motivation and Context

iOS 26 only uses layers for rendering SwiftUI content, thus our old logic broke and stopped masking text and images.

Fixes: #6390

## :green_heart: How did you test it?

Used our SwiftUI app with `.sentryReplayPreviewMask(opacity: 0.5)`

Example iOS 18 vs iOS 26
image

## :pencil: Checklist

You have to check all boxes before merging:

- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #7831, opened 2026-04-24
ref: Remove session replay environment checker and related options

itaybre:
## :scroll: Description

Removed all references to unreliable environment for Session Replay and checks related to this.

The unreliable environment was fixed in v9.11.0 so we don't need this checks anymore

## :pencil: Checklist


.github/workflows/test-cross-platform.yml
```diff
@@ -59,7 +59,7 @@ jobs:
         with:
           repository: getsentry/sentry-react-native
           path: sentry-react-native
-          ref: main
+          ref: chore/remove_enableSessionReplayInUnreliableEnvironment
```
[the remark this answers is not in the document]
itaybre: This is intentional, we can remove it afterwards (depending who merges first)

Sources/Swift/SentryExperimentalOptions.swift
```diff
@@ -13,23 +13,6 @@ public final class SentryExperimentalOptions: NSObject {
...
-     * Indicators for reliable environments include:
-     * - Running on an older version of iOS that doesn't have Liquid Glass (iOS 18 or earlier)
-     * - UIDesignRequiresCompatibility is explicitly set to YES in Info.plist
-     * - The app was built with Xcode < 26.0 (DTXcode < 2600)
-     *
-     * - Important: This flag allows to re-enable the session replay integration on iOS 26.0 and later, but please be aware that text and images may not be masked as expected.
-     *
-     * - Note: See [GitHub issues #6389](https://github.com/getsentry/sentry-cocoa/issues/6389) for more information.
-     */
-    public var enableSessionReplayInUnreliableEnvironment = false
```
philprime, line 31: `h`: I see two potential issues with this change:

1. We introduced this flag because of defensive programming approach because we could not reliably detect all sensitive data using Liquid Glass. With your changes in #7781 it should be better because it also captures layers used by SwiftUI and Liquid Glass, so we are back on track for over-masking instead of under-masking. But are you confident enough that masking on Liquid Glass is now reliable enough that we can remove the double opt-in with the session replay sample rate and this flag?
2. Removing this flag entirely makes the code base cleaner and it is as an experimental feature so it's fine to do the breaking change. What concerns me is that removing this will change the behavior of the SDK, re-enabling session replay for all apps who might rely on this behavior because they discovered edge-cases we did not consider/cover. I believe we should not remove this flag but as a first step we switch to setting it `true` by default, so if any user still relies on it they can set it to `false` in their SDK config.

itaybre: I believe this is a limitation on what we have right now, without a proper app to test all apple's features it is hard to guarantee everything was fixed.

At the end of the day is up to the user to confirm everything is masked as expected in their device / app.

Regarding 2. what's the difference between removing it and changing it to true?
If the users still need to disable session replay manually, they still need to modify their code (only the changed flag differs)

philprime (approved):
As discussed in today's Apple SDK sync, we are moving forward with this change and @itaybre will update the changelog with a warning, explaining that we are moving away from a double opt-in (setting `sessionReplaySampleRate` to a value higher than `0.0` and setting `enableSessionReplayInUnreliableEnvironment` to `true`) back to a single opt-in (setting `sessionReplaySampleRate` to a value higher than `0.0`), but also including a reference how SDK users can implement the check for liquid glass in their SDK start configuration block.

Sources/Swift/SentryExperimentalOptions.swift
itaybre: Resolving this as per our discussion

PR #7839, opened 2026-04-24
fix: race in SentryFramesTracker listeners

### LINKED ISSUE #7828: SentryFramesTracker: SIGABRT 'dangling reference' in recordTimestamp during Array resize (Swift)
### Platform

iOS

### Environment

Production

### Installed

Swift Package Manager

### Version

9.9.0

### Xcode Version

16.2

### iOS Version

iOS 26.2

### Device

iPhone 15 Pro (iPhone15,4)

Unknown — this is the first time we've observed this specific signature. Our app was on sentry-cocoa 9.9.0 when it occurred.

### Steps to Reproduce

Not consistently reproducible. Seen once in production (1 event / 1 user) from normal app usage on release 4.74.0.

### Expected Result

No crash during `SentryFramesTracker` display-link callback.

### Actual Result

`SIGABRT` with the Swift runtime "dangling reference" message:

> Object 0x126664780 of class _ContiguousArrayStorage deallocated with non-zero retain count 2. This object's deinit, or something called from it, may have created a strong reference to self which outlived deinit, resulting in a dangling reference.

This fires inside `SentryFramesTracker.recordTimestamp` while `Array._reserveCapacityAssumingUniqueBuffer` tries to consume-and-create-a-new backing buffer, suggesting a ref-count imbalance in the `_ContiguousArrayStorage` that backs one of the frames-tracker arrays — consistent with either a retain cycle or a race between the display-link thread and another thread mutating/reading the array.

Stack trace (top frames, from Sentry):

```
SentryFramesTracker.displayLinkCallback
SentryFramesTracker.displayLinkCallback  (inlined/wrapper)
SentryFramesTracker.recordTimestamp
Array._reserveCapacityAssumingUniqueBuffer
_ArrayBuffer._consumeAndCreateNew
swift::RefCounts<T>::doDecrementSlow<T>
_swift_release_dealloc
swift_deallocClassInstance
swift_deallocClassInstance.cold.1
swift::fatalError
swift::fatalErrorv
abort
__pthread_kill
```

CADisplayLink parent frames:

```
CA::Display::DisplayLinkItem::dispatch_
CA::Display::DisplayLink::dispatch_items
CA::Display::DisplayLink::dispatch_deferred_display_links
_UIUpdateSequenceRunNext
schedulerStepScheduledMainSectionContinue
UC::DriverCore::continueProcessing
```

### Context / prior related fixes

- [PR #7155](https://github.com/getsentry/sentry-cocoa/pull/7155) (merged 2026-01-20, in 9.3+) fixed a crash in the *listener* path (`removeListener`) by moving listener mutation onto the main queue. This crash is on a different path — `recordTimestamp` with the internal timestamp array.
- Older issues [#3511](https://github.com/getsentry/sentry-cocoa/issues/3511) and [#3607](https://github.com/getsentry/sentry-cocoa/issues/3607) reported `recordTimestamp:value:array:` crashes on the ObjC implementation in 8.x. Our stack is on the **Swift rewrite** shipped in 9.x, and the "dangling reference" wording points at a different root cause (lifetime issue on the backing `_ContiguousArrayStorage`, not a straight data race).
- Release notes for 9.10.0 and 9.11.0 contain no FramesTracker-related changes.

Happy to share additional event context, Sentry event ID, or test hypotheses if useful.

Not at this time — reporting to establish a paper trail. Will contribute if we can build a reliable repro.

itaybre:
## :scroll: Description

Replace `NSHashTable.weakObjects()` with a Swift array of weakly-boxed listeners in `SentryFramesTracker`.

Production crash reports show `EXC_BAD_ACCESS` / `SIGABRT` inside `-[NSConcreteHashTable removeItem:]` → `objc_msgSend`, called from `SentryFramesTracker.removeListener`'s dispatch block:

```
objc_msgSend
-[NSConcreteHashTable removeItem:]
<SentryFramesTracker.removeListener closure>
<NSHashTable.remove bridge>
__79-[_SentryDispatchQueueWrapperInternal dispatchAsyncOnMainQueueIfNotMainThread:]_block_invoke
```

Fixes: #7828

### Extra details:

PR #7155 moved every mutation of `listeners` onto the main thread to rule out concurrent mutators. That is necessary but not sufficient.

`NSHashTable.weakObjects()` stores each entry as a zeroing weak pointer inside its internal bucket storage. When a listener is deallocated, **the ObjC runtime zeroes that bucket slot on whatever thread the deallocation runs on**. Sentry never touches `listeners` off the main thread, but listener objects (`SentryTimeToDisplayTracker`, `SentryAppStartTracker`) can be released from any thread — e.g., a `SentryTracer` firing `finishCallback` from its idle/deadline `dispatch_after` queue, or the profiler releasing its last strong ref.

- \`make build-ios\` succeeds.
- Existing \`SentryFramesTrackerTests\` and \`SentryTimeToDisplayTrackerTest\` cover add / remove / stop / released-listener / listenersCount paths and keep passing.
- Race was not directly reproducible, matching the pattern of #7155.

## :pencil: Checklist

- [ ] No new PII added or SDK only sends newly added PII if \`sendDefaultPII\` is enabled.


Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -16,15 +16,28 @@ public protocol SentryFramesTrackerListener: NSObjectProtocol {
     func framesTrackerHasNewFrame(_ newFrameDate: Date)
 }
 
+// Wraps a listener with a zeroing weak reference. We manage weakness ourselves
+// instead of using NSHashTable.weakObjects() because NSHashTable's internal
+// buckets can be zeroed by the ObjC runtime on the deallocating thread while
+// the table is being iterated on another thread, causing
+// `NSConcreteHashTable removeItem:` to `objc_msgSend` a freed entry. Each box
+// is read via `objc_loadWeak`, which is safe against concurrent deallocation.
+private final class SentryFramesTrackerListenerBox {
```
philprime: `m`: Can we use [WeakReference](https://github.com/getsentry/sentry-cocoa/blob/ffb6adc5590113ef247904f6633f515bc005f56f/SentryTestUtils/Sources/WeakReference.swift#L7) here? We could move it from SentryTestUtils to Sentry. Alternative would be [SentryWeakMap](https://github.com/getsentry/sentry-cocoa/blob/ffb6adc5590113ef247904f6633f515bc005f56f/Sources/Sentry/include/SentryWeakMap.h#L34)

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -222,7 +243,7 @@ public class SentryFramesTracker: NSObject {
     }
     
     var listenersCount: Int {
-        listeners.count
+        listeners.reduce(0) { $0 + ($1.listener != nil ? 1 : 0) }
```
philprime: `l`: maybe easier to maintain:

```suggestion
        listeners.filter { $0 != nil }.count
```

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -40,7 +53,7 @@ public class SentryFramesTracker: NSObject {
     private var previousFrameTimestamp: CFTimeInterval = SentryFramesTracker.previousFrameInitialValue
     private var previousFrameSystemTimestamp: UInt64 = 0
     private var currentFrameRate: UInt64 = 60
-    private var listeners: NSHashTable<SentryFramesTrackerListener>
+    private var listeners: [SentryFramesTrackerListenerBox] = []
```
philprime: `m`: I think order is irrelevant, so we could use a `Set` instead

philprime (approved):
Approved but I think we have existing types we could use, so I'll leave it at your decision.

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
itaybre: No, but it makes it easier to prune empty elements

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -222,7 +236,7 @@ public class SentryFramesTracker: NSObject {
     }
     
     var listenersCount: Int {
-        listeners.count
+        listeners.filter { $0 != nil }.count
```
itaybre:
```suggestion
        listeners.filter { $0.value != nil }.count
```

PR #7840, opened 2026-04-24
ref: isolate SentryCrash behind SentryCrashReporter protocol

itaybre:
## :scroll: Description

Introduces a `SentryCrashReporter` protocol as the single abstraction point for SentryCrash access, and migrates all SDK consumers to depend on the protocol instead of concrete types.

(removed AI list here)

The rest of the SDK was reaching directly into SentryCrash concrete types, which made it hard to reason about the coupling between the event pipeline and the crash reporter, and hard to substitute in tests. Combined with the CI ratchet from the base branch (#`ci/sentrycrash-import-ratchet`), this establishes a single bridge point to SentryCrash and lets the ratchet hold the line going forward.

Base branch is `ci/sentrycrash-import-ratchet` so the protocol migration lands on top of the import-count guardrail.

- `make analyze`
- `make build-ios`
- `make test-ios` (unit tests covering Hub, Crash integration, ANR, Watchdog, Session Replay, Hang Tracking, App State, Extra Context)
- SentryCrash import ratchet stays at the baseline (86) — verified via `scripts/check-sentrycrash-imports.sh`.

## :pencil: Checklist

#skip-changelog


philprime (approved):
LGTM

PR #7843, opened 2026-04-28
feat(metrics)!: Promote experimental options to be generally available

philprime:
Promote `enableMetrics` and `beforeSendMetric` from `SentryExperimentalOptions` to top-level `Options`. Metrics has been stable for long enough that the experimental gate is no longer warranted — this matches the same move done previously for Structured Logs (#6359) and the swizzling options (#6592).

`enableMetrics` is also wired through `SentryOptionsInternal` dict-init so hybrid SDKs can toggle metrics on/off. `beforeSendMetric` stays Swift-only, since `SentryMetric` is a Swift struct and can't be bridged to ObjC; full ObjC support remains tracked in #6342.

Stale "experimental Metrics product" wording on `SentryAttributeValue` and `SentryAttributeContent` is removed — both are used by Logs (already GA) and Metrics. The `SentrySDK.metrics` headerdoc no longer claims "open beta" / "experimental".

#### Breaking change

`options.experimental.enableMetrics` and `options.experimental.beforeSendMetric` are removed with no deprecation shim. Use `options.enableMetrics` and `options.beforeSendMetric` instead. This matches how prior moves out of experimental were handled.

#### Notes

- `sdk_api.json` regenerated via `make generate-public-api`.
- Docs update for `docs.sentry.io/platforms/apple/metrics/` will land in a separate PR against `getsentry/sentry-docs`.
- Second commit (`docs(agents): ...`) updates `AGENTS.md` to require `make generate-public-api` on any PR touching the public API surface — independent of this change but small enough to ride along.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+> [!WARNING]
+> This release promotes Metrics out of experimental and **removes** `options.experimental.enableMetrics` and `options.experimental.beforeSendMetric`. If you set either of these, your app will fail to compile after upgrading.
+> Migrate to `options.enableMetrics` and `options.beforeSendMetric` (top-level on `Options`) — the defaults and behavior are unchanged.
+
+### Features
+
+- Metrics: Move options out of experimental (#7843)
+  - **Breaking**: `options.experimental.enableMetrics` and `options.experimental.beforeSendMetric` are removed. Use `options.enableMetrics` and `options.beforeSendMetric` instead.
+
 ### Fixes
 
 - Fix JSON encoding of infinite numeric values in crash reports (#7802)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,7 @@
 
 ### Features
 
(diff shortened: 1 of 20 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 ## Unreleased
 
+> [!WARNING]
+> This release promotes Metrics out of experimental and **removes** `options.experimental.enableMetrics` and `options.experimental.beforeSendMetric`. If you set either of these, your app will fail to compile after upgrading.
+> Migrate to `options.enableMetrics` and `options.beforeSendMetric` (top-level on `Options`) — the defaults and behavior are unchanged.
+
+### Features
+
+- Metrics: Move options out of experimental (#7843)
+  - **Breaking**: `options.experimental.enableMetrics` and `options.experimental.beforeSendMetric` are removed. Use `options.enableMetrics` and `options.beforeSendMetric` instead.
```
```suggestion
- Make feature Metrics generally available, moving experimental options to top-level options (#7843)
```

itaybre (approved):
LGTM

PR #7846, opened 2026-04-28
ref: Automatically select Device Name and OS based on Xcode and Platform for tests

itaybre:
Added scripts to automatically select proper device names and simulator versions based on the selected Xcode version.

I expected this to reduce the LOC for our github setup, but unfortunately I wasn't successful on that

#skip-changelog


.github/workflows/ui-tests-critical.yml
```diff
@@ -67,19 +63,15 @@ jobs:
           - name: iOS 18
             platform:
               runs-on: sequoia
-              xcode: "16.4"
-              test-destination-os: "18.4"
+              xcode: "16"
```
philprime, line 66: `m`: Picking Xcode 16 is not guarantee that it runs on iOS 18

philprime (approved):
LGTM with one comment

.github/workflows/ui-tests-critical.yml
itaybre: Didn't Xcode 16 come with iOS 18?

philprime: Yes, and it seems like Xcode 16 can't use iOS 26 simulators anyways. My point was just that Xcode and simulator versions are not strongly coupled.

PR #7847, opened 2026-04-28
ref(ci): consolidate ci-diagnostics into a composite action

itaybre:
## Summary

Replace the repeated `Run CI Diagnostics` block in 23 workflow files with a single `uses: ./.github/actions/ci-diagnostics`.

## Motivation

The diagnostics step was repeated ~26 times across workflows in 3-4-line blocks. A composite action gives us a single source of truth so any future change (e.g. uploading the captured environment as a job artifact) lands in one place rather than all callers.

## Checklist

- [x] No public API or behaviour changes
- [x] No new compiler warnings

#skip-changelog


itaybre:
Created without a fork in https://github.com/getsentry/sentry-cocoa/pull/7848

PR #7848, opened 2026-04-28
ref(ci): consolidate ci-diagnostics into a composite action

itaybre:
## Summary

Replace the repeated `Run CI Diagnostics` block in 23 workflow files with a single `uses: ./.github/actions/ci-diagnostics`.

## Motivation

The diagnostics step was repeated ~26 times across workflows in 3-4-line blocks. A composite action gives us a single source of truth so any future change (e.g. uploading the captured environment as a job artifact) lands in one place rather than all callers.

## Checklist

- [x] No public API or behaviour changes
- [x] No new compiler warnings

#skip-changelog


.github/workflows/analyze-language-trends.yml
```diff
@@ -54,9 +54,8 @@ jobs:
       - name: Upload Pages artifact
         uses: actions/upload-pages-artifact@fc324d3547104276b827a68afc52ff2a11cc49c9 # v5.0.0
 
-      - name: Run CI Diagnostics
+      - uses: ./.github/actions/ci-diagnostics
```
philprime: `h`: I think we should keep the name of the step, otherwise it looks like this: https://github.com/getsentry/sentry-cocoa/actions/runs/25078371480/job/73477025019

itaybre: Fair point, updating it

philprime (approved):

PR #7851, opened 2026-04-29
perf(session-replay): Reduce capture stutters

### LINKED ISSUE #6885: Low overhead session replay with runloop observers
Session reply uses a display link to decide when to capture a new frame, this leads to high overhead because even when the UI is not changing session replay is recording a frame. A lower overhead session replay would:

* Only capture a new frame on a new iteration of the run loop. UI is almost only updated from the main run loop, It is possible to have an animation rendering by Core Animation that doesn't use the main runloop.
* Disable recording when the run loop is in interactive mode, to prevent any overhead when scrolling that risks causing frame drops

romtsn:
Reduce Session Replay capture stutters by moving screenshot scheduling off display refresh callbacks and onto run loop activity.

The screenshot path is unchanged, so visual fidelity stays on the existing rendering, masking, scale, and video pipeline. The change only adjusts when we ask for screenshots. Session Replay now uses a run loop observer instead of a display link, captures after the run loop has processed UI work, treats tracking-mode run loop activity as interaction work, and keeps adaptive backoff for non-interaction captures that are themselves slow.

**Stacked on #8041**, which carries the video-assembly fixes (frame-slot rendering, retained previous-frame anchor, half-open segment windows, empty-segment dropping) that this scheduler relies on once capture intervals stretch past the configured frame interval. Land #8041 first; this PR then retargets `main` automatically.

Because the run loop observer — unlike a display link — keeps firing while the app is backgrounded, resume handling is reworked alongside the scheduler: session-mode (network) pause/resume is split from lifecycle pause/resume, and the integration tracks the application pause state so a connectivity reconnect while backgrounded clears the network-pause flag without restarting capture.

Suggested review order, by concern:

1. **Pacing policy** — `SentrySessionReplay.captureFrameIfNeeded` (stage order is documented in its headerdoc) plus the `CapturePacing` constants and adaptive-interval logic.
2. **Scheduler mechanics and segment bookkeeping** — run loop observer install/teardown, pending/pause segment handling. The two-domain threading model (lock-guarded vs main-thread-confined pacing state) is documented on the `lock` property.
3. **Resume paths** — `resume()` vs `resumeSessionMode(restartCaptureScheduler:)` and the reachability guard in `SentrySessionReplayIntegration`.
4. `SentrySessionReplayCaptureGuard` (new file) is a stateless view-hierarchy activity detector extracted for cohesion. The display-link era `newFrame(_:)` test entry point is gone; tests drive capture through the typed `captureFrameForTesting()` hook.

Latest physical-device comparison from the benchmark app on iPhone 15 running iOS 26.5, using Release builds:

| Metric | Main | This PR | Delta |
| --- | ---: | ---: | ---: |
| Average FPS | 55.1 | 58.4 | +3.3 |
| Sampled frames | 1,654 | 1,752 | +98 |
| Slow frames | 4.1% (67) | 1.4% (25) | -2.7pp (-42) |
| Frozen frames | 0 | 0 | 0 |
| p50 frame | 16.6ms | 16.6ms | 0ms |
| p90 frame | 16.6ms | 16.6ms | 0ms |
| p99 frame | 74.9ms | 25.7ms | -49.2ms |
| Max frame | 107.1ms | 82.8ms | -24.3ms |

This PR:

Session Replay physical-device benchmark results for this PR: 58.4 FPS, 1.4% slow frames, p99 25.7ms

Sample App Replay: https://sentry-sdks.sentry.io/explore/replays/8593836f355243c8863222a12493cb21
Benchmark App Replay: https://sentry-sdks.sentry.io/explore/replays/acc9cfbc692e4d2788b262da0e1fbd80

Main:

Session Replay physical-device benchmark results on main: 55.1 FPS, 4.1% slow frames, p99 74.9ms

Sample App Replay: https://sentry-sdks.sentry.io/explore/replays/2c737a44a1404cf0928a07c4e7d43ff8
Benchmark App Replay: https://sentry-sdks.sentry.io/explore/replays/e4891955c9bf4c539d111ebf268a11df

Closes #6885


Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -227,16 +258,149 @@ import UIKit
...
+            retain: nil,
+            release: nil,
+            copyDescription: nil
+        )
+        captureRunLoopObserver = CFRunLoopObserverCreate(
+            kCFAllocatorDefault,
+            activities,
+            true,
+            CFIndex.max,
+            { _, activity, context in
```
noahsmartin: The first parameter in this callback should be the observer, you can access the `mode` property and check if it's `.interactive`. That's how you can avoid firing the capture logic if the user is scrolling

philprime:
I quickly glanced over the changes and while I believe it to be viable (with the adoption of the existing `HangTracker.swift`) we need to do a full end-to-end verification with sample apps having a lot of tracked interactions, e.g. lots of scrolling, and compare it to sample apps without any activity in the UI.

My main concern is that we are not taking screenshots at a one second interval anymore, but instead the interval can be longer than that (less than a second is debounced).

In that case we need to look into two topics:

1. Extrapolating screenshots to reach the one second interval (for 1 FPS), i.e. if the time between screenshot A and screenshot B is more than one second we need to duplicate screenshot A.
2. Dynamic frame rates in MPEG containers, i.e. if the screenshots are taken in variable intervals larger than one second, we don't duplicate/extrapolate but instead the containers include information that the next frame will be in e.g. 2.4 seconds

romtsn:
Follow-up pushed in `1a2e9444b`.

This implements option 1 from Phil’s note: the video writer now holds the last captured bitmap across missing constant-FPS frame slots, including the tail of the requested replay window, so deferred/backed-off captures no longer compress replay time. I also fixed video duration calculation for frame rates above 1 FPS.

For Noah’s run-loop comment, the capture path skips tracking/interactive run-loop modes via the current run-loop mode check; this is covered by `testCaptureRunLoopObserver_whenRunLoopIsTracking_shouldNotCapture`.

Local verification: `make test-catalyst ONLY_TESTING=SentrySessionReplayTests,SentryOnDemandReplayTests,SentryVideoFrameProcessorTests` passes 67 tests, 0 failures.

romtsn:
Additional production-readiness evidence for Phil’s timing concern:

- Cross-checked Android: `sentry-java` already implements the same extrapolation one layer above `SimpleVideoEncoder`, in `ReplayCache.createVideoOf`. It walks the segment timeline at `1000 / frameRate` intervals and reuses the last known frame when no new screenshot exists for that slot. `SimpleMp4FrameMuxer` then writes constant-FPS presentation timestamps by output frame index.
- Cocoa now mirrors that model in `SentryVideoFrameProcessor`: we still write a constant-FPS MP4, but we hold the last bitmap across skipped output slots instead of letting sparse captures compress video time.
- Temp E2E smoke build: copied the benchmark app to `/tmp`, pointed its SPM dependency at `/Users/romtsn/Workspace/sentry-cocoa`, set the PR DSN, and built Release for iOS 26.4.1 Simulator. Xcode resolved `Sentry` from the local checkout and `xcodebuild` succeeded.
- Temp E2E smoke run: launched the benchmark on iPhone 17 / iOS 26.4.1 Simulator with replay on and replay off automation. Both 15s scrolling runs completed, persisted results, and reported 0 frozen frames.

I would not use the Simulator timing values as proof of the performance improvement; they are not comparable to the physical-device before/after benchmark in the PR description. This only verifies integration/build/run behavior for the current branch. Physical-device benchmarking remains the right validation for final performance confidence.

romtsn:
Pushed 6c8b177d4 to tighten skipped-frame extrapolation: sparse capture timestamps now use ceil when mapped into constant-FPS slots, so a fractional gap such as 2.4s is held through the next frame slot instead of rounding down and shortening the replay. Covered by testProcessFrames_WhenVideoEndHasFractionalGap_ShouldNotCompressDuration; replay-focused Catalyst suite passed locally (68 tests).

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -109,14 +120,24 @@ import UIKit
...
         }
         isSessionPaused = false
+        lock.unlock()
+
+        if shouldPreparePauseSegment {
+            prepareSegmentUntil(date: pauseDate)
+        }
     }
 
     public func resume() {
```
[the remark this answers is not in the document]
romtsn: Leaving this unchanged: `nextScreenshotAt` and `nextCaptureActivityCheckAt` are main-run-loop capture scheduler state. The run-loop observer reads them on the main thread, and the app lifecycle resume path is also delivered on the main thread. The lock in `resume()` protects adjacent replay state, but these pacing fields do not have an off-main reader/writer in the inspected call sites.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -109,37 +122,65 @@ import UIKit
...
+
+        if shouldStartCaptureScheduler {
+            resetCapturePacingAndStartScheduler()
+        }
+    }
+
+    private func prepareCaptureSchedulerResume() -> Bool {
+        if isSessionPaused {
+            return false
+        }
```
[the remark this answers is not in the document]
romtsn: Fixed in 07e00fc32. prepareCaptureSchedulerResume now only blocks scheduler restart when the replay is full-session and session mode is paused, so buffer replay restarts after app foreground while offline. Added testResume_whenBufferSessionModePaused_shouldRestartCaptureScheduler to cover screenshots and error replay capture.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -210,41 +251,427 @@ import UIKit
...
+
+    private func completeScreenshotCapture(
+        deferralDecision: ScreenshotDeferralDecision,
+        isInteractionCapture: Bool,
+        captureDuration: TimeInterval
+    ) {
+        runOnMainThread { [weak self] in
+            guard let self = self else { return }
+            defer { self.finishProcessingScreenshot() }
+
```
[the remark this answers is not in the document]
romtsn: Verified this path. The permanent-stuck scenario is not reachable because processingScreenshot is owned by the SentrySessionReplay instance. If self is nil in the async screenshot callback, that instance is already deallocated and there are no future captures on it to block. Added testDealloc_DoesNotRetainSessionReplayDuringAsyncScreenshot in 07e00fc32 to document that completing a pending screenshot after deallocation is harmless.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -41,6 +43,17 @@ import UIKit
...
+
+    deinit {
+        let retainedFrame = retainedFrameLock.synchronized {
+            let frame = retainedFrameBeforeCurrentFrames
+            retainedFrameBeforeCurrentFrames = nil
+            return frame
+        }
+        if let retainedFrame = retainedFrame {
+            removeFrameFile(retainedFrame)
+        }
```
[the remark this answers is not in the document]
romtsn: Verified this path. The described delete-while-read race is not reachable in the production call sites:

- Normal replay creation goes through `createVideoInBackgroundWith`, which dispatches `createVideoWith` onto the same serial `processingQueue` used by `releaseFramesUntil` and `addFrameAsync`. `createVideoWith` blocks that queue until `renderVideo` completes, so frame deletion cannot run concurrently with the asset worker reads.
- `deinit` cannot run while that queue block / instance method is still executing because `self` is strongly retained for the duration of video creation.
- Recovery uses a separate one-shot `SentryOnDemandReplay` loaded from disk and does not call `releaseFramesUntil` or add frames on that maker while rendering.

The direct `createVideoWith` test usage is synchronous and does not introduce concurrent deletion. Resolving this as a false positive rather than adding a broad render-time file lock.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -185,18 +233,18 @@ import UIKit
...
 
         self.replayType = replayType
-        startFullReplay()
+        startFullReplay(startedAt: lastScreenshotAt)
         let replayStart = dateProvider.date().addingTimeInterval(-replayOptions.errorReplayDuration - (Double(replayOptions.frameRate) / 2.0))
 
-        createAndCaptureInBackground(startedAt: replayStart, replayType: replayType)
+        createAndCaptureInBackground(startedAt: replayStart, endedAt: dateProvider.date(), replayType: replayType)
         return true
     }
```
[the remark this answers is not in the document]
romtsn: it's not a real bug — the flag is deliberately left set because the in-flight screenshot's completion handler clears it. Resetting it in stopCaptureScheduler would allow concurrent screenshots when the old completion fires and clears the flag for a new in-flight capture.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -210,41 +258,396 @@ import UIKit
...
+                return nil as CFRunLoopObserver?
+            }
+
+            captureRunLoopObserver = observer
+            return observer
+        }
+
+        if let observerToAdd = observerToAdd {
+            CFRunLoopAddObserver(CFRunLoopGetMain(), observerToAdd, .commonModes)
+        }
```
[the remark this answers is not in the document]
romtsn: The observer doesn't leak.

 CFRunLoopObserverCreateWithHandler follows the CF Create Rule — Swift takes +1 ownership of the returned CFRunLoopObserver? and manages it via ARC. When the guard at line 504 fails, observerToAdd is nil, the
  function returns, and the local observer variable goes out of scope. ARC releases it — no leak.

  The observer was never added to the run loop so there's nothing to remove, and its [weak self] closure is deallocated along with it. No retain cycle either.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -326,61 +747,76 @@ import UIKit
...
         lock.unlock()
-        
+
         SentrySDKLog.debug("[Session Replay] Getting screenshot from screenshot provider")
-        let timestamp = dateProvider.date()
         let screenName = delegate?.currentScreenNameForSessionReplay()
+        let captureStart = dateProvider.systemTime()
         screenshotProvider.image(view: rootView) { [weak self] screenshot in
-            self?.newImage(timestamp: timestamp, maskedViewImage: screenshot, forScreen: screenName)
+            guard let self = self else { return }
```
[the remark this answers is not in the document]
romtsn: I don’t think the exception scenario is valid as stated: an uncaught exception during UIKit rendering would crash/terminate rather than leave replay disabled for the rest of a live session

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayCaptureGuard.swift
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        return view.subviews.contains { containsActiveInteraction(in: $0) }
+    }
+
+    private func activeAnimationCount(in layer: CALayer, upTo limit: Int) -> Int {
+        var count = layer.animationKeys()?.count ?? 0
+        guard count < limit else { return count }
+
+        for sublayer in layer.sublayers ?? [] {
```
philprime: `h`: This can cause crashes, see `SentryUIRedactBuilder` for the workarounds and fixes around the sublayer recursion

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayCaptureGuard.swift
```diff
@@ -0,0 +1,58 @@
...
+
+        if activeAnimationCount(in: rootView.layer, upTo: Self.activeAnimationThreshold) >= Self.activeAnimationThreshold {
+            return .animation
+        }
+
+        return nil
+    }
+
+    private func containsActiveInteraction(in view: UIView) -> Bool {
+        if let scrollView = view as? UIScrollView, scrollView.isDragging || scrollView.isDecelerating || scrollView.isTracking {
```
philprime: `h`: Did you test this with a larger scroll-based app like BlueSky? I have the feeling that these kind of apps will have long gaps of static frames because the frame capturing is constantly debounced

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayCaptureGuard.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+#if (os(iOS) || os(tvOS)) && !SENTRY_NO_UI_FRAMEWORK
+import UIKit
+
+/// Inspects the view hierarchy to detect ongoing activity (user interactions or animations)
+/// that should influence when ``SentrySessionReplay`` captures the next screenshot.
+final class SentrySessionReplayCaptureGuard {
```
philprime: `m`: If possible let's adopt `struct` over `class`. Also this type contains no state, so it could be an enum with static methods instead

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -480,7 +492,11 @@ public class SentrySessionReplayIntegration: NSObject, SwiftIntegration, SentryS
     // MARK: - SentryReachabilityObserver
     public func connectivityChanged(_ connected: Bool, typeDescription: String) {
         SentrySDKLog.debug("[Session Replay] Connectivity changed to: \(connected ? "connected" : "disconnected"), type: \(typeDescription)")
-        if connected { sessionReplay?.resume() } else { sessionReplay?.pauseSessionMode() }
+        if connected {
+            sessionReplay?.resumeSessionMode(restartCaptureScheduler: !isApplicationStatePaused)
```
philprime, line 507: `h`: Please verify on which method this is called, we should not use locks on the main thread if possible

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -53,100 +71,138 @@ import UIKit
...
+            markPauseSegmentIfNeeded(at: pauseDate)
+        }
+
+        if shouldPreparePauseSegment {
+            prepareSegmentUntil(date: pauseDate)
         }
-        isSessionPaused = false
+    }
+
+    private func markPauseSegmentIfNeeded(at pauseDate: Date) -> Bool {
```
philprime: `m`: It would be great to have more context here in what a "pending pause segment" is, so future maintainers can more easily understand the code's purpose

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -29,22 +25,44 @@ import UIKit
...
+    /// `deferredScreenshotStart`) is not guarded by this lock; it is main-thread confined and only
+    /// mutated from the run-loop capture scheduler callbacks, blocks dispatched to the main
+    /// thread, and `start`.
     private let lock = NSLock()
+    private let captureGuard = SentrySessionReplayCaptureGuard()
+    /// Capture pacing state; main-thread confined and deliberately not guarded by `lock` (see its docs).
+    private var adaptiveScreenshotInterval: TimeInterval = 0
+    /// Capture pacing state; main-thread confined and deliberately not guarded by `lock` (see its docs).
+    private var deferredScreenshotStart: Date?
+    private var captureRunLoopObserver: CFRunLoopObserver?
```
philprime: `h`: Please consider abstracting the RunLoop Observer into an injectable type/struct using protocol-abstraction. This would allow us to inject a fake runloop in unit tests, rather than having a `SENTRY_TEST` gated test method

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -210,41 +266,396 @@ import UIKit
         event.tags = tags
     }
 
-    @objc 
-    private func newFrame(_ sender: CADisplayLink) {
-        guard let lastScreenShot = lastScreenShot, isRunning &&
-                !(isFullSession && isSessionPaused) //If replay is in session mode but it is paused we dont take screenshots
-        else { return }
+    #if SENTRY_TEST || SENTRY_TEST_CI || DEBUG
+    func captureFrameForTesting(isInteractiveRunLoopMode: Bool = false) {
```
philprime: `h`: See comment above w.r.t. getting rid of `SENTRY_TEST` gated method via dependency injection

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -210,41 +266,396 @@ import UIKit
...
+        captureFrameIfNeeded(isInteractiveRunLoopMode: isInteractiveRunLoopMode)
+    }
+    #endif
+
+    /// Decides on each run-loop pass whether to capture a screenshot. Stages, in order:
```
philprime, line 365: Thank you for this list of checks written in natural language

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -210,41 +266,396 @@ import UIKit
...
+            return observer
+        }
+
+        if let observerToAdd = observerToAdd {
+            CFRunLoopAddObserver(CFRunLoopGetMain(), observerToAdd, .commonModes)
+        }
+    }
+
+    /// Conversion factor between `SentryCurrentDateProvider.systemTime()` nanoseconds and seconds.
+    private static let nanosecondsPerSecond: TimeInterval = 1_000_000_000
```
philprime: `l`: Please check the code base, I believe this constant was defined somewhere else already and can be reused here instead. We might even have utilities to do conversions

philprime (commented):
Good progress, left a couple of comments to discuss

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayCaptureGuard.swift
romtsn: Nope, but I tested it on the benchmark app shared with us by a customer, which is scroll-heavy and was used for before/after verification on simulator and physical device. IIRC within a 30-second benchmark app run, the main branch managed to capture 9 interactions (scrolls), whereas this branch captured 6-7 interactions -- I believe it's still good enough compared to baseline.

We also now have a unit test covering the specific pacing concern: after adaptive backoff is triggered by a slow idle capture, a scroll interaction still captures at the configured replay frame rate instead of waiting for the backed-off interval.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
romtsn: Verified this callback is invoked from SentryReachability’s NWPathMonitor background queue in production, so this lock is not taken from the main thread on the real path.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/f3193b2e7dc94007faa8e02873a740ba72724170

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayCaptureGuard.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/ebe46c3467851c9ee7e36aef83b577c8d9f77532

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/4fd16d066eb216ced4d1db149a629b9550fbd9ec

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayCaptureGuard.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/b1ece5acb5e476d09a0eb83bfbb88c77f461c561 and https://github.com/getsentry/sentry-cocoa/pull/7851/commits/4bd13cc45851c34e9f9f0802d372bdffe390258d

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/f21c4ab018dfb01fe0dcc7a00c329df7f0a7d247

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/f21c4ab018dfb01fe0dcc7a00c329df7f0a7d247

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -184,19 +266,21 @@ import UIKit
             return false
         }
 
+        // `lastScreenshotAt` is main-thread confined with the capture pacing state.
+        let startedAt = runOnMainThreadSync { lastScreenshotAt }
```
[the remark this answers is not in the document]
romtsn: I don't think it's gonna deadlock though. But i'd accept this tradeoff because `captureReplay` is not being called often, we just block it for a short stint to grab the timestamp. Otherwise we'd have to do a larger lock-based timestamp fix which I'd avoid for this path.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -54,99 +97,142 @@ import UIKit
...
-        defer { lock.unlock() }
-        
-        if isSessionPaused {
-            isSessionPaused = false
-            return
+        let resumeDate = dateProvider.date()
+        let shouldStartCaptureScheduler = lock.synchronized { () -> Bool in
+            if isFullSession && isSessionPaused {
+                return false
+            }
```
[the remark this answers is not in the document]
romtsn: that's a deliberate decision not to run full session mode replay while offline as we can run out of storage space very quickly.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayRunLoopCaptureScheduler.swift
```diff
@@ -0,0 +1,131 @@
...
+            true,
+            CFIndex.max
+        ) { [weak self] observer, activity in
+            guard let observer = observer,
+                let self = self,
+                self.isValidObserver(observer),
+                self.shouldCapture(activity: activity)
+            else { return }
+
+            capture(self.currentRunLoopMode() == .tracking)
```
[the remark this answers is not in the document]
romtsn: this is likely a very rare case that this object is going to be deinited on a background thread, so probably not worth optimizing for (and we are also not really doing that across the codebase)

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -210,41 +309,328 @@ import UIKit
...
+            isCaptureSchedulerRunning = true
+            captureSchedulerToken = token
+            return true
+        }
+        guard shouldInstallObserver else { return false }
+
+        captureScheduler.start(token: token) { [weak self] isInteractiveRunLoopMode in
+            self?.captureFrameIfNeeded(isInteractiveRunLoopMode: isInteractiveRunLoopMode)
+        }
+        return true
```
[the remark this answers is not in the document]
romtsn: I think this is an extreme edge case that CFRunLoopObserverCreateWithHandler would return `nil`, not worth handling

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -54,99 +105,144 @@ import UIKit
...
-        displayLinkWrapper: SentryReplayDisplayLinkWrapper
+        captureScheduler: SentrySessionReplayRunLoopCaptureScheduler
     ) {
         self.replayOptions = replayOptions
         self.dateProvider = dateProvider
         self.delegate = delegate
         self.screenshotProvider = screenshotProvider
-        self.displayLink = displayLinkWrapper
         self.urlToCache = replayFolderPath
         self.replayMaker = replayMaker
```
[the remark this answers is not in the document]
romtsn: while that's right, I belive it'd be a better fix to enforce it on the API level potentially (do the hop to main thread in SentryReplayApi)

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
```diff
@@ -734,6 +692,7 @@ final class SentryUIRedactBuilder {
         // But UISwitch is in the list of ignored class identifiers by default, because it uses
         // non-sensitive images. Therefore we want to ignore the subtree of UISwitch, unless
         // it was removed from the list of ignored classes
+        let viewTypeId = type(of: view).description()
```
philprime, line 685: `l`: You removed quite some context from the comment block of why we use `type(of:)` here. Is that intentional?

Sources/Swift/Core/Tools/ViewCapture/SentryViewSubtreeTraversal.swift
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    static func isExcluded(
+        _ view: UIView,
+        excludedViewClassPatterns: Set<String>,
+        includedViewClassPatterns: Set<String>
+    ) -> Bool {
+        // Use string descriptions instead of NSClassFromString or stored class objects to avoid
+        // triggering Objective-C +initialize on UIKit classes.
+        let viewTypeId = type(of: view).description()
```
philprime, line 102: `h`: You removed all the context of why we use `type(of:)` here. The comment were elaborate on purpose, so please put it back in.

Sources/Swift/Core/Tools/ViewCapture/SentryViewSubtreeTraversal.swift
```diff
@@ -0,0 +1,79 @@
...
+                visit
+            )
+        }
+    }
+
+    static var defaultExcludedViewClassPatterns: Set<String> {
+        var result: Set<String> = []
+        #if os(iOS)
+        if #available(iOS 26.0, *) {
+            result.insert("CameraUI.ChromeSwiftUIView")
```
philprime: `m`: Should be a constant

Sources/Swift/Core/Tools/ViewCapture/SentryViewSubtreeTraversal.swift
```diff
@@ -0,0 +1,79 @@
+#if canImport(UIKit) && !SENTRY_NO_UI_FRAMEWORK
+#if os(iOS) || os(tvOS)
+import UIKit
+
+enum SentryViewSubtreeTraversal {
```
philprime, line 17: `h`: I see the benefit of extracting this, but to me it feels like a lot of additional context is missing which was added to SentryUIRedactBuilder in the many iterations of trying to handle all cases.

Sources/Swift/Core/Tools/ViewCapture/SentryViewSubtreeTraversal.swift
```diff
@@ -0,0 +1,79 @@
...
+        includedViewClassPatterns: Set<String>,
+        _ visit: (UIView) -> Bool
+    ) -> Bool {
+        guard !isExcluded(
+            view,
+            excludedViewClassPatterns: excludedViewClassPatterns,
+            includedViewClassPatterns: includedViewClassPatterns
+        ) else {
+            return false
+        }
```
philprime: `l`: nitpick, in some programming languages it's best practice to use `if-return` instead of `guard-not-return` because it resolves the double-negation

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -29,23 +28,75 @@ import UIKit
...
+    ) {
+        self.replayOptions = replayOptions
+        self.dateProvider = dateProvider
+        self.delegate = delegate
+        self.screenshotProvider = screenshotProvider
+        self.urlToCache = replayFolderPath
+        self.replayMaker = replayMaker
+        self.breadcrumbConverter = breadcrumbConverter
+        self.touchTracker = touchTracker
+        self.captureScheduler = DefaultSentrySessionReplayRunLoopCaptureScheduler()
```
philprime: `m`: Why is this not injected in all cases?

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -112,7 +120,7 @@ import UIKit
         lock.lock()
         defer { lock.unlock() }
         
-        displayLink.invalidate()
+        stopCaptureScheduler()
```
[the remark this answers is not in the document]
philprime: `m`: Please verify if this is valid feedback

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -54,99 +105,144 @@ import UIKit
...
-        lock.lock()
-        defer { lock.unlock() }
-        
-        displayLink.invalidate()
-        if isFullSession {
-            prepareSegmentUntil(date: dateProvider.date())
+        stopCaptureScheduler()
+
+        let pauseDate = dateProvider.date()
+        let shouldPreparePauseSegment = lock.synchronized { () -> Bool in
```
philprime, line 230: `h`: `pause()` can be called from the main thread, so this acquires a lock on the main thread. Make sure none of the logic in queuePendingPauseSegmentIfNeeded(at:) is dispatching to the main thread, otherwise we have a deadlock.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayRunLoopCaptureScheduler.swift
```diff
@@ -0,0 +1,131 @@
...
+        return true
+    }
+
+    private func runOnMainThreadSync(_ block: () -> Void) {
+        if Thread.isMainThread {
+            block()
+        } else {
+            DispatchQueue.main.sync(execute: block)
+        }
+    }
```
philprime, line 116: `l`: I think we have a helper for that in SentryDispatchQueueWrapper or similar.

philprime (commented):
Almost LGTM, left some feedback

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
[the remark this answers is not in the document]
romtsn: yep it had been addressed with `pendingPauseSegmentEnd` already, and there are tests covering it:
- testPause_whenSegmentCreationIsPending_shouldPreparePauseSegmentAfterPendingCompletes
- testPauseSessionMode_whenSegmentCreationIsPending_shouldPreparePauseSegmentAfterPendingCompletes

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
romtsn: Current code is fine. `queuePendingPauseSegmentIfNeeded(at:)` only does this:
```swift
  guard _isFullSession, pendingSegmentEnd != nil else { return false }
  pendingPauseSegmentEnd = pauseDate
  return true
```
No dispatch, no callback, no external call, no `prepareSegmentUntil`. The actual segment preparation happens after the lock is released.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayRunLoopCaptureScheduler.swift
romtsn: Good point! I'd prefer to keep it this way, if you don't mind. We only need this specific check and not the full queue abstraction, and going through a full dependency injection path + changing test fixtures plumbing seems like an overkill.

Sources/Swift/Core/Tools/ViewCapture/SentryViewSubtreeTraversal.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/52c61478f0be19aa7b2c08e0b8a2884ea3dbae92

Sources/Swift/Core/Tools/ViewCapture/SentryViewSubtreeTraversal.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/a00514dbb229bdd82c8684338fc7362315821faa

Sources/Swift/Core/Tools/ViewCapture/SentryUIRedactBuilder.swift
romtsn: nope, was just slop, addressed in https://github.com/getsentry/sentry-cocoa/pull/7851/commits/a00514dbb229bdd82c8684338fc7362315821faa

Sources/Swift/Core/Tools/ViewCapture/SentryViewSubtreeTraversal.swift
romtsn: I tried to restore all the context and put it into this class, let me know what you think! https://github.com/getsentry/sentry-cocoa/pull/7851/commits/2ac9a60003f4d4782e71ef204b31b2e50081d3bf and https://github.com/getsentry/sentry-cocoa/pull/7851/commits/a00514dbb229bdd82c8684338fc7362315821faa

Sources/Swift/Core/Tools/ViewCapture/SentryViewSubtreeTraversal.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/03785ebb8b1e68921204ffcc643604824c791b4e

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851/commits/69eb56c21cbe1faa485386dd62eab902a5e51fe7

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -54,99 +84,144 @@ import UIKit
...
+
+            if self.startCaptureScheduler(expectedGeneration: schedulerGeneration) {
+                self.resetCapturePacing(at: self.dateProvider.date())
+            }
         }
-        guard !isRunning else { 
-            SentrySDKLog.debug("[Session Replay] Session is already running, not resuming")
-            return 
+    }
+
```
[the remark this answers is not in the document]
romtsn: https://github.com/getsentry/sentry-cocoa/pull/7851#discussion_r3430926754

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -54,99 +84,144 @@ import UIKit
...
-        defer { lock.unlock() }
-        
-        if isSessionPaused {
-            isSessionPaused = false
-            return
+        let resumeDate = dateProvider.date()
+        let schedulerGeneration = lock.synchronized { () -> Int? in
+            if _isFullSession && isSessionPaused {
+                return nil
+            }
```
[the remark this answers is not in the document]
romtsn: not a real bug anymore, we covered this and there's a test for this: `testConnectivityReconnect_whenApplicationPaused_shouldWaitForForeground`

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -176,8 +176,9 @@ func removeReplayFile(at fileURL: URL) {
     }
 
     /// Used by replay recovery after `init(withContentFrom:)` loaded surviving frames from disk.
-    var oldestRecoveredFrameDate: Date? {
-        _frames.first?.time
+    var recoveredFrameDateRange: (oldest: Date, newest: Date)? {
```
philprime: `l`: You could use `DateInterval` here https://developer.apple.com/documentation/foundation/dateinterval/init(start:end:)

Sources/Swift/Integrations/SessionReplay/SessionReplayRecovery.swift
```diff
@@ -6,9 +6,9 @@ import Foundation
 /// Configuration for recovering a previous session replay after a crash or app restart.
 private struct PreviousReplayConfig {
     let type: SentryReplayType
-    let duration: TimeInterval
     let segmentId: Int
     let beginning: Date
+    let end: Date
```
philprime: `l`: Could also use [DateInterval](https://developer.apple.com/documentation/foundation/dateinterval/init(start:end:))

philprime (approved):
LGTM

Sources/Swift/Integrations/SessionReplay/SessionReplayRecovery.swift
romtsn: switchted to DateInterval in https://github.com/getsentry/sentry-cocoa/pull/7851/commits/ec37a8d6caa651aa9a3ffb681747421002a6735a

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
romtsn: switchted to DateInterval in https://github.com/getsentry/sentry-cocoa/commit/ec37a8d6caa651aa9a3ffb681747421002a6735a

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -326,59 +773,78 @@ import UIKit
...
+
         lock.lock()
         guard !processingScreenshot else {
             SentrySDKLog.debug("[Session Replay] Not taking screenshot, reason: processing screenshot")
             lock.unlock()
-            return
+            return false
         }
         processingScreenshot = true
         lock.unlock()
```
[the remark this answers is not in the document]
romtsn: no it wouldn't fail

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -54,99 +84,144 @@ import UIKit
...
+            isSessionPaused = false
+            if _isFullSession {
+                videoSegmentStart = resumeDate
+            }
         }
-        
-        videoSegmentStart = nil
-        displayLink.link(withTarget: self, selector: #selector(newFrame(_:)))
+        guard restartCaptureScheduler else { return }
+        resume()
```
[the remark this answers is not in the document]
romtsn: I do not think this is correct for the current branch. This no longer writes a raw `videoSegmentStart`; `resumeSessionMode` clears `isSessionPaused` and goes through `segmentState.resumeSession(at:)`.

If the scheduler is already running, `resume()` returning early is fine: the scheduler was intentionally skipping captures while `isSessionPaused` was true, and the segment state has already been moved to the reconnect time. Conditioning this on `captureSchedulerToken == nil` would keep the old paused interval as part of the next segment, which is the behavior we want to avoid.

For the background/foreground case, `testConnectivityReconnect_whenApplicationPaused_shouldWaitForForeground` covers that reconnect while paused does not restart capture until foreground.

Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift
```diff
@@ -54,99 +159,144 @@ import UIKit
...
+        let pauseDate = dateProvider.date()
+        let shouldPreparePauseSegment = lock.synchronized { () -> Bool in
+            guard _isFullSession else { return false }
+            guard !queuePendingPauseSegmentIfNeeded(at: pauseDate) else { return false }
+            return true
+        }
+
+        if shouldPreparePauseSegment {
+            prepareSegmentUntil(date: pauseDate)
         }
```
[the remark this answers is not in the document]
philprime: Not a real issue. The scenario assumes `isSessionPaused` stays `true` through foregrounding, but the connectivity observer handles this:

1. Network lost → `pauseSessionMode()` sets `isSessionPaused = true`
2. App backgrounds → `pause()` stops the scheduler
3. Network restored (even while backgrounded) → `connectivityChanged(true)` calls `resumeSessionMode()` which clears `isSessionPaused = false`
4. App foregrounds → `resume()` proceeds normally since `isSessionPaused` is already cleared

If the network is *still* down when the app foregrounds, replay correctly stays paused — there is no transport to send segments. When connectivity returns, `connectivityChanged(true)` clears the flag and restarts capture via `resumeSessionMode(restartCaptureScheduler: !isApplicationStatePaused)`.

PR #7852, opened 2026-04-29
fix: Fix format string argument order in SentryCrashReportStore

itaybre:
## Description

Fix mismatched format string arguments in the `SENTRY_ASYNC_SAFE_LOG_ERROR` call within `sentrycrashcrs_addUserReport` (partial write error path).

The format string `"Expected to write %d bytes to file %s, but only wrote %d"` expected `(int, char*, int)` but received `(char*, int, int)`. This caused `%s` to dereference `reportLength` (a small integer) as a memory address, resulting in a segfault when a partial write occurs (e.g., disk full).

## How to test

- The fix is a trivial argument reorder — verified by reading the format string and matching specifiers to arguments.

#skip-changelog


philprime (commented):
Nice catch, LGTM

philprime (approved):

PR #7853, opened 2026-04-29
fix: Handle EOF in sentrycrashfu_readBytesFromFD to prevent infinite loop

itaybre:
## Description

Fix an infinite loop in `sentrycrashfu_readBytesFromFD` when `read()` returns 0 (EOF). The `while (length > 0)` loop never terminates because `length -= 0` makes no progress.

This is reachable when reading a truncated or empty file — callers include `sentrycrashfu_readEntireFile` (crash report loading) and `SentrySessionReplaySyncC` (replay segment state).

Also fixes a copy-paste typo in the error message: "Could not write" → "Could not read".

## How to test

- Existing `SentryCrashFileUtils_Tests` pass (38/38, 0 failures)
- The EOF path is new — previously it was an infinite loop, so no existing test covers it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 
 ### Fixes
 
+- Fix infinite loop in `sentrycrashfu_readBytesFromFD` when `read()` returns EOF on a truncated file (#7853)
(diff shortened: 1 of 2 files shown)


philprime (approved):
Nice catch, LGTM

PR #7857, opened 2026-04-30
fix: Add depth limits to data sanitization and JSON encoder

philprime:
## Description

Add depth limits to the dictionary/array sanitization functions and the SentryCrash JSON encoder to prevent stack overflows from deeply nested structures.

**Sanitization:** Thread a shared `depth` parameter through the mutual recursion between `sentry_sanitize` and `sentry_sanitizeArray`. At depth 200, dictionaries return `nil` and arrays return `@[]`. The public API is unchanged. The `SentryArray` class was removed and its array sanitization logic consolidated into `SentryNSDictionarySanitize.m`.

**JSON encoder:** Add a bounds check on `containerLevel` against the `isObject[200]` array size in both `sentrycrashjson_beginArray` and `sentrycrashjson_beginObject`, returning `SentryCrashJSON_ERROR_DATA_TOO_LONG` when exceeded. The check is performed before incrementing `containerLevel` so it stays valid on error, since callers in `SentryCrashReport.c` may ignore the return value.

## Motivation

Deeply nested `NSError.userInfo` dictionaries or crafted crash report JSON could exhaust the stack. Found while auditing the serialization paths.

#skip-changelog


Sources/Sentry/include/SentryNSDictionarySanitize.h
```diff
@@ -4,5 +4,6 @@
 NS_ASSUME_NONNULL_BEGIN
 
 SENTRY_EXTERN NSDictionary *_Nullable sentry_sanitize(NSDictionary *_Nullable dictionary);
+SENTRY_EXTERN NSArray *sentry_sanitizeArray(NSArray *array);
```
NinjaLikesCheez: `l`: should this file be renamed to include that it handles `NSArray`?

philprime: Good catch, I'll rename it

Tests/SentryTests/Categories/SentryArrayTests.swift
```diff
@@ -158,10 +158,31 @@ class SentryArrayTests: XCTestCase {
...
+
+        // Assert
+        XCTAssertEqual(result.count, 1)
+        var current: [Any]? = result
+        var depth = 0
+        while let next = current?.first as? [Any] {
+            current = next
+            depth += 1
+        }
+        XCTAssertLessThan(depth, 250)
```
itaybre: `l`: While not wrong, can we make it exact?

Tests/SentryTests/Categories/SentryArrayTests.swift
```diff
@@ -158,10 +158,31 @@ class SentryArrayTests: XCTestCase {
...
+    }
 }
 
 // Helper class that inherits from NSObject but is not a real NSDictionary
-// This will cause sentry_sanitize to return nil when it checks isSubclassOfClass
+// This will cause sentry_sanitize_dictionary(to return nil when it checks isSubclassOfClass
```
itaybre:
```suggestion
// This will cause sentry_sanitize_dictionary( to return nil when it checks isSubclassOfClass
```

Tests/SentryTests/Categories/SentrySanitizerUtilsTests.swift
```diff
@@ -7,7 +7,7 @@ class SentryNSDictionarySanitizeTests: XCTestCase {
         // Arrange
         let dict: [String: Any]? = nil
         // Act
-        let sanitized = sentry_sanitize(dict)
+        let sanitized = sentry_sanitize_dictionary(dict)
```
itaybre, line 10: `l`: Should we rename this test file too? (and `SentryArrayTests.swift`)

Tests/SentryTests/Categories/SentryNSDictionarySanitizeTests.swift
```diff
@@ -142,11 +142,32 @@ class SentryNSDictionarySanitizeTests: XCTestCase {
...
+
+        // Assert
+        XCTAssertNotNil(sanitized)
+        var current: [String: Any]? = sanitized as? [String: Any]
+        var depth = 0
+        while let next = current?["nested"] as? [String: Any] {
+            current = next
+            depth += 1
+        }
+        XCTAssertLessThan(depth, 250)
```
itaybre: `l`: same here

itaybre (approved):
LGTM, with some comments

philprime:
I decided to clean it up and convert to Swift

Sources/Swift/Tools/SentrySanitizerUtils.swift
```diff
@@ -0,0 +1,63 @@
...
+        return string
+    case let number as NSNumber:
+        return number
+    case let dict as NSDictionary:
+        return sanitizeDictionary(dict, depth: depth)
+    case let array as NSArray:
+        return sanitizeArray(array, depth: depth)
+    case let date as NSDate:
+        return sentry_toIso8601String(date as Date)
+    default:
```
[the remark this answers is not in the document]
philprime: Not a real bug — the old Objective-C implementation handled `NSNull` identically.

In the ObjC code, `NSNull` didn't match any of the `isKindOfClass:` checks (`NSString`, `NSNumber`, `NSDictionary`, `NSArray`, `NSDate`), so it fell through to the `else` branch which called `[rawValue description]`, returning the string `@"<null>"`.

The new Swift code does exactly the same: `NSNull` doesn't match any `case` in the `switch`, falls into `default`, and `String(describing: rawValue)` also produces `"<null>"`.

**Neither implementation preserves `NSNull` as JSON null.** Both convert it to the string `"<null>"`. No behavioral change was introduced by the rewrite.

PR #7859, opened 2026-04-30
fix: Handle extra length call in SentryInvalidJSONString

itaybre:
## Description

Apple changed the implementation of `__CFStringEncodeByteStream` in recent platform versions to additionally call `length` on NSString subclasses. This caused `SentryInvalidJSONString`'s invocation counter to double-count, breaking the test utility's logic.

## Fix

Filter `length` calls by inspecting the call stack — only increment the counter when the call originates from the expected code path (not from `__CFStringEncodeByteStream`). This also fixes the previously-skipped watchOS test, so the `#if !os(watchOS)` / `XCTSkip` guard is removed.

## Test plan

- [x] Existing `SentryInvalidJSONStringTests` pass on all platforms including watchOS

#skip-changelog


itaybre:
@philprime I know this solution is kinda sketchy, but:
- What this model attempts is a bit unorthodox already
- This is only for testing

philprime (approved):

Issue #1650, opened 2022-01-26
Crash in monitorCachedData


stefanfuerst:
### Platform

macOS

### Installed

Manually

### Version

7.8

### Steps to Reproduce

Lately I'm often getting C++ Exception: _NSCoreDataException that look like this

```
libsystem_kernel    0x1bdbf4ebc  __semwait_signal
libsystem_c         0x1bdaffd84  nanosleep
libsystem_c         0x1bdb0a6e8  sleep
Sentry              0x1010a9384  monitorCachedData (Sources/SentryCrash/Recording/SentryCrashCachedData.c:139)
libsystem_pthread   0x1bdc2d23c  _pthread_start
```

I can't tell if this is not parsed properly or an issue with Sentry. Also I was not able to reproduce this locally.

### Expected Result

Display of details of the _NSCoreDataException

### Actual Result

Just C++ Exception: _NSCoreDataException

stefanfuerst:
I did some experiments by creating an exception on my own

This ends up in Sentry as **C++ Exception: _NSCoreDataException** without stack trace
```
- (void)applicationDidFinishLaunching:(NSNotification *)aNotification
{
    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(4 * NSEC_PER_SEC)), dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
        [[NSClassFromString(@"_NSCoreDataException") exceptionWithName:@"Oh Damn" reason:@"Crash to crash" userInfo:nil] raise];
    });
}
```
While generating a regular Exception is reported with the correct parameters and stack trace

```
- (void)applicationDidFinishLaunching:(NSNotification *)aNotification
{
    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(4 * NSEC_PER_SEC)), dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
        [NSException exceptionWithName:@"Oh Damn" reason:@"Crash to crash" userInfo:nil] raise];
    });
}

```
I did not yet figure out the point where reporting goes wrong

brustolin:
Hey @stefanfuerst, thanks for letting us know this.
I will take a look.

philipphofmann:
@brustolin was able to reproduce this. We can't give you an ETA for fixing this, though @stefanfuerst.

philipphofmann:
[...]

PR #7866, opened 2026-05-04
fix: Report crashed thread for null function calls

denrase:
## :scroll: Description

Calling a null function pointer on ARM64 can crash with PC=0. Our stack cursor stopped immediately in that case, producing an empty backtrace for the crashed thread. The converter then dropped the crashed/main thread, which could make Sentry show another non-crashed thread as the apparent source.

Continue unwinding from LR/FP when PC is zero, matching the KSCrash fix in https://github.com/kstenerud/KSCrash/pull/356. Add a macOS sample trigger and regression test for this crash shape.

Relates to https://github.com/getsentry/sentry-cocoa/issues/1650#issuecomment-4352450250

## :pencil: Checklist

### DIFF
--- a/Sources/SentryCrash/Recording/Tools/SentryCrashStackCursor_MachineContext.c
+++ b/Sources/SentryCrash/Recording/Tools/SentryCrashStackCursor_MachineContext.c
@@ -65,11 +65,8 @@ advanceCursor(SentryCrashStackCursor *cursor)
         return false;
     }
 
-    if (context->instructionAddress == 0) {
+    if (context->instructionAddress == 0 && cursor->state.currentDepth == 0) {
         context->instructionAddress = sentrycrashcpu_instructionAddress(context->machineContext);
-        if (context->instructionAddress == 0) {
-            goto tryAsyncChain;
-        }
         nextAddress = context->instructionAddress;
         goto successfulExit;
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,8 @@
 
 ### Fixes
 
-- Report crashed thread for null function calls ([#7866](https://github.com/getsentry/sentry-cocoa/pull/7866))
-
 - Prevent memory strings in stack overflow crash reports (#7841)
+- Report crashed thread for null function calls (#7866)
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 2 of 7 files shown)


denrase:
# Sample Repro

## Before

https://denrase.sentry.io/issues/7458499096/?project=4508007036551168&query=is%3Aunresolved&referrer=issue-stream
Bildschirmfoto 2026-05-04 um 14 40 57

## After
https://denrase.sentry.io/issues/7236191812/events/latest/?project=4508007036551168&query=is%3Aunresolved&referrer=latest-event

Bildschirmfoto 2026-05-04 um 14 36 17

itaybre:
@denrase did you forget to add "ready-to-merge" label or are you missing something?

Sources/SentryCrash/Recording/Tools/SentryCrashStackCursor_MachineContext.c
```diff
@@ -65,11 +65,8 @@ advanceCursor(SentryCrashStackCursor *cursor)
         return false;
     }
 
-    if (context->instructionAddress == 0) {
+    if (context->instructionAddress == 0 && cursor->state.currentDepth == 0) {
         context->instructionAddress = sentrycrashcpu_instructionAddress(context->machineContext);
-        if (context->instructionAddress == 0) {
-            goto tryAsyncChain;
-        }
```
itaybre, line 72: Don't we need to keep this?
The address is being modified 1 line above

itaybre: Our Path seems a bit different from the KSCrash one

denrase:
@itaybre I forgot, sorry.

Sources/SentryCrash/Recording/Tools/SentryCrashStackCursor_MachineContext.c
denrase: They have reverted this part, but i will re-check if I introduced an regression in our code.

denrase: Checked again, we now do the same as KSCrash after the fix. Only difference, they return false instead of `goto tryAsyncChain` (where we also only return false). Removing the return on `context->instructionAddress == 0` is part of the fix.

It lets the first frame being emitted as address 0. The currentDepth increment protects against re-entry. So instead of being done here, the next call can continue to LR / FP recovery path.

itaybre (approved):
LGTM

PR #7874, opened 2026-05-06
test: add a dummy NSPrincipalClass for the SentryTests bundle

supervacuus:
## :scroll: Description

This PR introduces an `NSPrincipalClass` for the `SentryTests` bundle without any behavior (beyond the inherited from `NSObject`), so that `NSBundle`/`XCTest` don't fall back on the first class in the ObjC class list.

The fallback this prevents was identified as a possible cause of [recurring but rare test failures](https://github.com/getsentry/sentry-cocoa/actions/runs/25377124290/job/74415236079?pr=7857), which could occur because the class list is not stable across builds and potentially leads to `+initialize` being called on `SentryInitializeForGettingSubclassesNotCalled`, which exists for a test explicitly showing that `+initialize` won't be called. While these are entirely different paths, the non-deterministic selection of an arbitrary class as the principal for `SentryTests` still leads to a failing, flaky CI run.

I ran the test suite.

## :pencil: Checklist


supervacuus:
Closing in favor of #7880.

Issue #7842, opened 2026-04-27
Change Array extension for Metrics attributes to extend Collections


philprime:
Right now we are extending `Array` but we should change this to `Collection` so also `Sets` can conform to Sentry's Attribute typing system

PR #7876, opened 2026-05-06
feat: Add Set conformance to SentryAttributeValue

NinjaLikesCheez:
## :scroll: Description

Extend Set to conform to `SentryAttributeValue` (see #7842). Fall back to the extension of Array when the Set contains mixed types so not duplicate logic. 

Allows consumers to use `Set` types in Metric Attributes

## :pencil: Checklist

### DIFF
--- a/Sources/Swift/Protocol/SentryAttributeValue.swift
+++ b/Sources/Swift/Protocol/SentryAttributeValue.swift
@@ -220,7 +220,7 @@ extension Array: SentryAttributeValue {
         guard !values.isEmpty, case .string = values[0].asSentryAttributeContent else {
             return nil
         }
-        
+
         let mappedStringValues = values.compactMap { element -> String? in
             guard case .string(let value) = element.asSentryAttributeContent else {
                 return nil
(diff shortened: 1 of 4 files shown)


Sources/Swift/Protocol/SentryAttributeValue.swift
```diff
@@ -220,7 +220,7 @@ extension Array: SentryAttributeValue {
         guard !values.isEmpty, case .string = values[0].asSentryAttributeContent else {
             return nil
         }
-        
+
         let mappedStringValues = values.compactMap { element -> String? in
             guard case .string(let value) = element.asSentryAttributeContent else {
                 return nil
```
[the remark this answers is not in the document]
NinjaLikesCheez: This keeps the same behaviour as the existing Array implementation so considering it as invalid (for now).

Sources/Swift/Protocol/SentryAttributeValue.swift
```diff
@@ -231,5 +231,35 @@ extension Array: SentryAttributeValue {
...
+        }
+        if Element.self == Float.self, let values = self as? Set<Float> {
+            return .doubleArray(Array(values).map(Double.init))
+        }
+        if Element.self == Int.self, let values = self as? Set<Int> {
+            return .integerArray(Array(values))
+        }
+        if Element.self == String.self, let values = self as? Set<String> {
+            return .stringArray(Array(values))
+        }
```
philprime: `m`: As all paths are using Set-to-Array conversion, which is conforming to `SentryAttributeValue`, couldn't we just go for the standard Array conversion in all cases? Is there a benefit of duplicating the casting logic?


```suggestion
```

philprime (commented):
Almost LGTM

Sources/Swift/Protocol/SentryAttributeValue.swift
NinjaLikesCheez: Nope, no benefit - I'll remove. Thanks!

philprime (approved):
LGTM 🎉

philprime:
@NinjaLikesCheez please create a follow-up PR in the sentry-docs repository to mention supported Swift types `Array` and `Set` here: https://docs.sentry.io/platforms/apple/metrics/#adding-attributes

PR #7879, opened 2026-05-07
chore(ci): Reduce CI log noise and standardize shell scripts

philprime:
## Summary

- Extend `ci-utils.sh` with full GitHub Actions workflow command wrappers (`log_debug`, `log_notice`, `log_warning`, `log_error`, `begin_group`/`end_group`, `mask_value`, `set_env`, `set_output`, `append_summary`, etc.)
- Refactor all CI shell scripts to use `ci-utils.sh` instead of raw `::error::`, `::warning::`, `::notice::` commands
- Downgrade `log_notice` calls to plain `echo` where the message is progress output, not an actionable PR annotation
- Standardize all scripts with `usage()` functions, `-h`/`--help` support, argument validation, and parameter summaries
- Add `begin_group`/`end_group` blocks for better log organization in GitHub Actions
- Delete unused `create-cherry-pick-branch.sh`
- Remove SwiftLint from xcframework build jobs to eliminate tooling warnings

## Motivation

CI logs were noisy with non-actionable annotations cluttering PR reviews. Scripts had inconsistent logging, missing usage documentation, and raw GitHub Actions workflow commands scattered throughout. This PR establishes a consistent pattern across all CI scripts for better debuggability and cleaner PR annotations.

image

#skip-changelog


Sentry.xcodeproj/project.pbxproj
```diff
@@ -2854,7 +2854,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if [[ \"$(uname -m)\" == arm64 ]]\nthen\n    export PATH=\"/opt/homebrew/bin:$PATH\"\nfi\n\nif command -v swiftlint >/dev/null 2>&1\nthen\n    swiftlint\nelse\n    echo \"warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions.\"\nfi\n";
+			shellScript = "if [[ \"$(uname -m)\" == arm64 ]]\nthen\n    export PATH=\"/opt/homebrew/bin:$PATH\"\nfi\n\nif command -v swiftlint >/dev/null 2>&1\nthen\n    swiftlint\nfi\n";
```
itaybre, line 2857: Can we skip the warning only in CI?

philprime: I could probably check if i can inject an environment flag and expand the script, but IMHO this just bloats the codebase and the risk of merging unlinted code doesn't really exist with our CI setup.

philprime:
@sentry review

scripts/ci-utils.sh
```diff
@@ -3,8 +3,23 @@
 # Utility functions for CI logging and grouping.
 # This file is intended to be sourced from other scripts.
 #
-# GitHub Actions workflow commands (::notice::, ::warning::, ::error::, ::group::, ::endgroup::)
-# follow the specification at: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands
+# Wraps all GitHub Actions workflow commands that use the ::command:: syntax.
+# Specification: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions
+#
+# Available functions:
+#   log_debug <message>          — ::debug::{message}
```
NinjaLikesCheez, line 11: `l`: would it make sense to add a `log_info` function that only echos the message back? That way scripts can have a 'unified' set of log functions instead of swapping between `log_*` and `echo` in the scripts?

NinjaLikesCheez (approved):
LGTM with one `l` comment

scripts/ci-utils.sh
philprime: I was considering this too, but then I thought we also sometimes just use the CLI output directly e.g. `cat` or `grep` so maybe this wouldn't even be used everywhere and kind of feels like over-engineering. Do you see it worth the effort?

NinjaLikesCheez: I would say it's more of a nice to have - it certainly reads better when looking through areas of logging in the scripts, so I think it 'makes sense'. If you'd like - I can make a commit with the change?

philprime: Not necessary, I'll update the PR. Thanks for the feedback.

PR #7875, opened 2026-05-06
ci: upload SentryTests binary metadata

supervacuus:
## :scroll: Description

Uploads a zip artifact for the `SentryTests` build to inspect metadata of the binary to allow class list and linker setup after the workflow is finished.

## :bulb: Motivation and Context

This is a debugging helper and a companion to https://github.com/getsentry/sentry-cocoa/pull/7874.

## :green_heart: How did you test it?

Tested locally against multiple platform test runs. But running it in CI is the ultimate test.

## :pencil: Checklist

You have to check all boxes before merging:

- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #7880, opened 2026-05-07
test: dynamically registered poison class for `SentrySubClassFinderTests`

supervacuus:
## :scroll: Description

This PR removes the statically linked `SentryInitializeForGettingSubclassesNotCalled` sentinel from the `SentryTests` Mach-O and changes `SentrySubClassFinderTests` to register that tracked class dynamically at test runtime instead.

It replaces: https://github.com/getsentry/sentry-cocoa/pull/7874 and https://github.com/getsentry/sentry-cocoa/pull/7875. However, I left the metadata uploads from #7875 in here. Leave this up to the reviewers whether we want them at all.

The investigation on https://github.com/getsentry/sentry-cocoa/pull/7874 and https://github.com/getsentry/sentry-cocoa/pull/7875 showed that the flaky CI failures were caused by an earlier _Foundation_ bundle-load path, not by the later explicit `XCTest` principal-class path.

During `-[NSBundle loadAndReturnError:]`, _Foundation_ calls into `__NSBundleGetPrincipalClassFromInfoDict`, but at that point, the runtime `infoDictionary` for `SentryTests.xctest` can still be empty. In that case, _Foundation_ falls back to `NSBundle._firstClassName` and sends a message that triggers normal ObjC `+initialize` on whatever class happens to be first in the image.

This means the real problem was not a missing plist entry, but that `SentryInitializeForGettingSubclassesNotCalled` intentionally failed in `+initialize` while being packaged into the test bundle Mach-O. Since any first class may be touched on that path, such a class is not safe to keep statically linked into the bundle.

By dynamically registering the tracked class only for the relevant test, we keep the original test intent: verifying that `SentrySubClassFinder` does not invoke `+initialize`, without depending on test-bundle load behavior.

**Open question for reviewers**: the old `SentryInitializeForGettingSubclassesNotCalled` file/class naming is now a bit stale because it only registers a dynamic sentinel at runtime. I am happy to rename in this PR if there is a preferred direction.

- `make test-ios ONLY_TESTING=SentrySubClassFinderTests`
- ran `./scripts/ci-collect-sentrytests-metadata.sh` locally to confirm:
  - `NSPrincipalClass` is no longer set for `SentryTests`
  - `SentryTestsBundlePrincipalClass` is gone
  - `SentryInitializeForGettingSubclassesNotCalled` is no longer present in the ObjC class list
- ran `./scripts/sentry-xcodebuild.sh` with `build-for-testing`/`test-without-building` against `--platform tvOS --os 17.5` as in CI

## :pencil: Checklist


Tests/SentryTests/Integrations/Performance/SentryInitializeForGettingSubclassesNotCalled.m
```diff
@@ -1,23 +1,46 @@
...
 
-@implementation SentryInitializeForGettingSubclassesCalled : NSObject
+    Class dynamicClass = objc_allocateClassPair([NSObject class], [className UTF8String], 0);
+    if (dynamicClass == Nil) {
+        return nil;
+    }
+
+    Class metaClass = object_getClass(dynamicClass);
+    BOOL addedInitialize
+        = class_addMethod(metaClass, @selector(initialize), (IMP)sentryDynamicInitialize, "v@:");
```
itaybre, line 30: `m`: Since we cannot add a test for this, are you sure +initialize is called when registered this way?

supervacuus: Yes, I am... but believe me, I tested it first, because I wanted to ensure the registration vs realization boundary. You can test it yourself with the stripped code in this PR:

```objc
#import <objc/runtime.h>
#include <stdio.h>

static BOOL called = NO;

static void dynInitialize(id self, SEL _cmd)
{
    called = YES;
    fprintf(stderr, "+initialize called for %s\n", class_getName((Class)self));
}

int main(void)
{
    @autoreleasepool {
        NSString *name = @"DynInitializeTestClass";
        Class cls = objc_allocateClassPair([NSObject class], name.UTF8String, 0);
        if (cls == Nil) {
            fprintf(stderr, "failed to allocate class\n");
            return 1;
        }

        Class meta = object_getClass(cls);
        BOOL added = class_addMethod(meta, @selector(initialize), (IMP)dynInitialize, "v@:");
        if (!added) {
            fprintf(stderr, "failed to add +initialize\n");
            return 2;
        }

        objc_registerClassPair(cls);

        fprintf(stderr, "before: called=%d\n", called);
        [(id)cls self];
        fprintf(stderr, "after: called=%d\n", called);
    }
    return 0;
}
```

`self` being the exact message the `NSBundle` sent our way to trigger `+initialize`.

```
$ clang -fobjc-arc -framework Foundation dyn_init_test.m -o dyn_init_test
$ ./dyn_init_test
before: called=0
+initialize called for DynInitializeTestClass
after: called=1
```

`+initialize` is a class method and as such lives in the meta-class. `objc_registerClassPair()` makes the class known to the runtime but causes no lazy `+initialize`. Only the `self` message causes realization.

supervacuus: And btw, if you reset `called` after the "after" `printf` and message the class again, `called` will not be true! So that is really a one-time thing, and with this setup, we observe it.

.github/workflows/unit-test-common.yml
```diff
@@ -235,6 +244,15 @@ jobs:
             results.xcresult
             flaky-results.xcresult
 
+      - name: Archiving SentryTests Metadata
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        if: ${{ (failure() || cancelled()) && inputs.scheme == 'Sentry' && steps.collect_sentrytests_metadata.outcome == 'success' }}
+        with:
+          name: sentrytests-metadata-${{inputs.platform}}-xcode-${{inputs.xcode}}-os-${{inputs.test-destination-os}}
```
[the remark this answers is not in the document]
supervacuus: Yeah, this is bad. If we want to keep the metadata upload, I will fix this.

Tests/SentryTests/Integrations/Performance/SentryInitializeForGettingSubclassesNotCalled.m
supervacuus: > Since we cannot add a test for this

While we can't test it for the finder, we could add another test that registers another class created the same way, intentionally triggers realization, and asserts that it was called. This way, we can be sure that during the same test run, the method sensitivity works, but the finder didn't trigger it. Does that make sense?

itaybre: Nono, I trust you, I just wanted to double check it.

I don't want to add tests for test utils either. This is fine

itaybre (approved):
LGTM

.github/workflows/unit-test-common.yml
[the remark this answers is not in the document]
supervacuus: resolved in https://github.com/getsentry/sentry-cocoa/pull/7880/commits/46254d73fb3d66a8883d32cc7f34f2987cb8bf81

supervacuus:
I guess the failing iOS tests are flaky and not introduced by the proposed changes: https://github.com/getsentry/sentry-cocoa/actions/runs/25683588651/job/75401336199?pr=7880#step:13:1104

PR #7891, opened 2026-05-08
fix: Crash with AVAssetDownloadTask in setState

### LINKED ISSUE #7886: Regression in 9.12.0: crash with AVAssetDownloadTask in urlSessionTask:setState: — "does not support originalRequest property"
## Summary

Sentry Cocoa 9.12.0 introduced a deterministic crash in apps that use `AVAssetDownloadURLSession` for offline asset downloads. Every state transition on an `AVAssetDownloadTask` raises an uncaught `NSGenericException`, terminating the app.

The crash is 100% reproducible on every asset download after upgrading from 9.11.0 to 9.12.x. Downgrading to 9.11.0 fully resolves it.

## Crash

```
*** Terminating app due to uncaught exception 'NSGenericException',
    reason: 'AVAssetDownloadTask does not support originalRequest property'
```

Top frames of the stack:

```
0  CoreFoundation              __exceptionPreprocess
1  libobjc.A.dylib             objc_exception_throw
2  CFNetwork                   ...
3  Sentry                      -[SentryNetworkTracker urlSessionTask:setState:] + 120
4  Sentry                      +[SentrySwizzleWrapperHelper swizzleURLSessionTask:]_block_invoke_2
5  CFNetwork                   <state machine>
6  libdispatch.dylib           <gcd>
7  CFNetwork                   ...
8  <App>                       AVAssetDownloadTask.resume()  via AVAssetDownloadURLSession
```

## Environment

* Sentry Cocoa: regression introduced in **9.12.0**, still present in **9.12.1**. Last good version: **9.11.0**.
* iOS 18.x and iOS 26.x both affected.
* Xcode 26.x.

The crash reproduces with default Sentry network options (`enableNetworkTracking`, `enableNetworkBreadcrumbs`, `enableCaptureFailedRequests` all enabled). The experimental flag `options.experimental.enableReplayNetworkDetailsCapturing` is **not** set (default `false`), so the new feature is not even opted into — the crash still occurs.

## Reproduction

Any code path using `AVAssetDownloadURLSession.makeAssetDownloadTask(...)` followed by `task.resume()` will crash. Minimal example:

```swift
let session = AVAssetDownloadURLSession(
    configuration: .background(withIdentifier: "com.example.dl"),
    assetDownloadDelegate: self,
    delegateQueue: .main
)
let asset = AVURLAsset(url: someHLSURL)
let task  = session.makeAssetDownloadTask(asset: asset, assetTitle: "video", assetArtworkData: nil, options: nil)
task?.resume()  // Triggers urlSessionTask:setState: on the background task -> crash
```

## Root cause

PR getsentry/sentry-cocoa#7588 (commit `ce1ec47d5a41fef293f369106071be7d6838e8a9`, "feat(network-details): Hook up request/response capture in SentryNetworkTracker", merged 2026-04-29) added the following to `-[SentryNetworkTracker urlSessionTask:setState:]` in `Sources/Sentry/SentryNetworkTracker.m`:

```objc
- (void)urlSessionTask:(NSURLSessionTask *)sessionTask setState:(NSURLSessionTaskState)newState
{
    if (!self.isNetworkTrackingEnabled && !self.isNetworkBreadcrumbEnabled
        && !self.isCaptureFailedRequestsEnabled) {
        return;
    }

#if SENTRY_TARGET_REPLAY_SUPPORTED
    SentryOptions *options = SentrySDK.startOption;
    NSString *urlString = sessionTask.originalRequest.URL.absoluteString;   // <-- always evaluated
    if ([self isNetworkDetailCaptureEnabledFor:urlString options:options]) {
        [self captureRequestDetails:sessionTask
               networkCaptureBodies:options.sessionReplay.networkCaptureBodies
              networkRequestHeaders:options.sessionReplay.networkRequestHeaders];
    }
#endif // SENTRY_TARGET_REPLAY_SUPPORTED

    if (![self isTaskSupported:sessionTask]) {   // <-- task-type filter runs LATER
        return;
    }
    ...
}
```

The new block evaluates `sessionTask.originalRequest` **unconditionally**, on every `NSURLSessionTask` subclass, and **before** `isTaskSupported:` filters out unsupported task types.

`isTaskSupported:` (`SentryNetworkTracker.m:529-536`) returns `YES` only for:

* `NSURLSessionDataTask`
* `NSURLSessionDownloadTask`
* `NSURLSessionUploadTask`

`AVAssetDownloadTask` is a **direct subclass of** `NSURLSessionTask` (not `NSURLSessionDownloadTask`), so it was previously filtered out and never touched. Apple explicitly throws on `[AVAssetDownloadTask originalRequest]`:

> `AVAssetDownloadTask does not support originalRequest property`

The runtime guard `isNetworkDetailCaptureEnabledFor:` does not help because the offending property access happens on the line above it, regardless of whether the feature is opted into.

In 9.11.0, `isTaskSupported:` was the first non-guard check in the method, so `AVAssetDownloadTask` exited safely.

## Suggested fix

Move the new `#if SENTRY_TARGET_REPLAY_SUPPORTED` block **after** the `isTaskSupported:` filter:

```objc
- (void)urlSessionTask:(NSURLSessionTask *)sessionTask setState:(NSURLSessionTaskState)newState
{
    if (!self.isNetworkTrackingEnabled && !self.isNetworkBreadcrumbEnabled
        && !self.isCaptureFailedRequestsEnabled) {
        return;
    }

    if (![self isTaskSupported:sessionTask]) {
        return;
    }

#if SENTRY_TARGET_REPLAY_SUPPORTED
    SentryOptions *options = SentrySDK.startOption;
    NSString *urlString = sessionTask.originalRequest.URL.absoluteString;
    if ([self isNetworkDetailCaptureEnabledFor:urlString options:options]) {
        [self captureRequestDetails:sessionTask
               networkCaptureBodies:options.sessionReplay.networkCaptureBodies
              networkRequestHeaders:options.sessionReplay.networkRequestHeaders];
    }
#endif
    ...
}
```

Alternatively (or additionally), defensively skip task classes Apple is known to throw on:

```objc
if ([sessionTask isKindOfClass:NSClassFromString(@"AVAssetDownloadTask")]) {
    return;  // Apple explicitly throws on -originalRequest for this subclass
}
```

There is also a similar `originalRequest.HTTPBody` access inside `captureRequestDetails:` that should likewise be guarded if `AVAssetDownloadTask` could ever reach it.

## Workaround

For users who cannot downgrade, disabling all three of the following options causes `urlSessionTask:setState:` to early-return at the very top, before the unsafe access:

```swift
options.enableNetworkTracking      = false
options.enableNetworkBreadcrumbs   = false
options.enableCaptureFailedRequests = false
```

Trade-off: full loss of HTTP breadcrumbs and network tracking.

## Impact

* Any iOS app using `AVAssetDownloadURLSession` (HLS offline playback, Apple's recommended approach for offline video) crashes 100% on every download with Sentry 9.12.0+.
* Crash is `EXC_CRASH (SIGABRT)` from an uncaught `NSGenericException`, surfaced to users immediately.

## Related

* Regression PR: getsentry/sentry-cocoa#7588 (`ce1ec47d5a41fef293f369106071be7d6838e8a9`)
* Part of the wider Session Replay network details capture work tracked in getsentry/sentry-cocoa#4944
* Sibling PRs in the same wave that may also need review for similar `originalRequest` access on unsupported task types: getsentry/sentry-cocoa#7580, getsentry/sentry-cocoa#7582, getsentry/sentry-cocoa#7584, getsentry/sentry-cocoa#7585, getsentry/sentry-cocoa#7590, getsentry/sentry-cocoa#7854

itaybre:
## Description

Fixes a regression introduced in 9.12.0 (#7588) where `SentryNetworkTracker.urlSessionTask:setState:` accesses `sessionTask.originalRequest` before the `isTaskSupported:` guard. `AVAssetDownloadTask` throws `NSGenericException` on that property, causing a 100% crash for any app using `AVAssetDownloadURLSession`.

The fix moves the `isTaskSupported:` check before the `#if SENTRY_TARGET_REPLAY_SUPPORTED` block that accesses `originalRequest`, restoring the guard order from 9.11.0.

Fixes #7886

## How to test

1. Create an `AVAssetDownloadURLSession` and call `makeAssetDownloadTask(...).resume()`
2. Before this fix: crash with `NSGenericException: AVAssetDownloadTask does not support originalRequest property`
3. After this fix: task resumes without crash

## Checklist

- [x] I reviewed the submitted code
- [x] I added applicable tests — verified in simulator that AVAssetDownloadTask no longer crashes
- [x] No new PII or credentials are captured/stored


NinjaLikesCheez (approved):
LGTM

PR #7892, opened 2026-05-08
fix: Defensive type checks for malformed recrash reports

### LINKED ISSUE #7888: SentryCrash.doctorReport crashes on launch with NSInvalidArgumentException (objectForKeyedSubscript: sent to __NSArrayM), causing a permanent crash-loop
### Platform
iOS

### Environment
Production

### Installed
CocoaPods (transitively via `@sentry/react-native` 8.3.0 → `RNSentry` pod → `Sentry` pod 9.6.0)

### Other Error Monitoring Solution
No

### Version
sentry-cocoa **9.6.0** (`Sentry` pod, locked by `@sentry/react-native@8.3.0`)

### Xcode Version
26.4.1 (build 17E202)

Not investigated for this report — the affected user is on a build that has shipped with sentry-cocoa 9.6.0 since release.

### Steps to Reproduce
I cannot reproduce this on demand — it has hit one user, but consistently. Observed externally:

1. User launches the app (production build, App Store distribution).
2. `SentrySDK.start` runs synchronously during app startup.
3. As part of `installIntegrations`, `SentryCrashIntegration` calls `SentryCrash.sendAllReportsWithCompletion:`, which iterates over previously persisted crash reports on disk.
4. While processing one of those reports, `-[SentryCrash doctorReport:]` throws `NSInvalidArgumentException: -[__NSArrayM objectForKeyedSubscript:]: unrecognized selector sent to instance 0x...`.
5. The exception is unhandled (`mechanism: nsexception`, `handled: no`, `level: fatal`) and crashes the app before any application code runs.
6. On the next launch the same on-disk report is replayed → crash repeats. The user is in a permanent crash-loop until they reinstall the app (which clears the on-disk Sentry crash store).

Frequency: 5 fatal events over ~17 hours from the same `identifierForVendor`, single user, no other users affected on the same build.
App context at the time of the crash: `in_foreground: true`, `is_active: false` (i.e. during launch, before the app has become active).

### Expected Result
`-[SentryCrash doctorReport:]` should defensively handle a stored crash report whose JSON shape doesn't match expectations — for example, a value the doctor expects to be an `NSDictionary` that is actually an `NSArray`. On a shape mismatch the SDK should either skip and delete the malformed report, or produce a best-effort report, but it must not throw an unhandled `NSException` from inside `SentrySDK.start`.

In particular: the SDK starting up should never be able to crash-loop the host app on subsequent launches based on data the SDK itself wrote to disk on a previous run.

### Actual Result

```
NSInvalidArgumentException
-[__NSArrayM objectForKeyedSubscript:]: unrecognized selector sent to instance 0x112b8c000
```

Stack (top frames trimmed for the throw, bottom frames trimmed at UIApplicationMain):

```
objc_exception_throw
-[NSObject(NSObject) doesNotRecognizeSelector:]
___forwarding___
__forwarding_prep_0___
-[SentryCrash doctorReport:]                               ← crash site
-[SentryCrash reportWithIntID:]
-[SentryCrash allReports]
-[SentryCrash sendAllReportsWithCompletion:]
SentryCrashSwift.sendAllReports
-[SentryCrashInstallation sendAllReportsWithCompletion:]
SentryCrashIntegration.init
SentrySwiftIntegrationInstaller.install
+[SentrySDKInternal installIntegrations]
__38+[SentrySDKInternal startWithOptions:]_block_invoke
-[_SentryDispatchQueueWrapperInternal dispatchAsyncOnMainQueueIfNotMainThread:]
SentryDispatchQueueWrapper.dispatchAsyncOnMainQueueIfNotMainThread
+[SentrySDKInternal startWithOptions:]
SentrySDK.start
+[RNSentryStart startWithOptions:]
+[RNSentryStart startWithOptions:error:]
-[RNSentry initNativeSdk:resolve:rejecter:]
... (UIApplicationMain)
```

### What I've ruled out

- **Application code**: every frame above `UIApplicationMain` is inside `Sentry.framework` / `RNSentry`. No first-party frames appear before the throw.
- **A doctor-report-specific fix in 9.7.0 → 9.12.1**: I scanned `CHANGELOG.md` and didn't find an entry that names `doctorReport`, `objectForKeyedSubscript:`, `__NSArrayM`, or "unrecognized selector". Adjacent fixes I noticed (mechanically plausible but not named matches): #7853 (infinite loop in `sentrycrashfu_readBytesFromFD` on truncated files, 9.12.0), #7857 (depth limits in JSON encoding, 9.12.1), #7859 (length handling in `SentryInvalidJSONString`, 9.12.1).

### Asks

1. Is this a known bug already fixed somewhere between 9.6.0 and HEAD that I missed?
2. If not, is there a safe `SentryOptions` knob — or, in a future release, a documented procedure — for users in this state to recover without reinstalling the app?
3. If neither, would a defensive `isKindOfClass:` check inside `doctorReport` (with "delete-and-skip" on shape mismatch) be acceptable as a PR?

I'm happy to share the full Sentry event JSON privately if it would help diagnose which key/field in the stored report is the array-vs-dict that the doctor is choking on.

Yes, with guidance on which file owns the doctor-report parsing and what the expected shape is supposed to be.

itaybre:
## Summary
- Add `isKindOfClass:` type guards in `doctorReport:` (`SentryCrash.m`) and `initWithReport:` (`SentryCrashReportConverter.m`) before subscripting the `recrash_report` field
- SentryCrash can produce structurally malformed recrash reports during signal handler execution (known issue — see existing defensive check in `initThreads:`). When `recrash_report` is a non-dictionary type (e.g. `NSArray`), the chained keyed subscript throws `NSInvalidArgumentException`, crashing the app on every launch in a permanent crash-loop
- The corrupted report is never deleted because the exception fires before `sendAllReportsWithCompletion:` reaches its cleanup handler

## Test plan
- [x] `make build-ios` passes
- [x] `SentryCrashTests` — 4 new tests covering `doctorReport:` with array, string, and nested-array malformed inputs
- [x] `SentryCrashReportConverterTests` — 2 new tests covering converter with array and string `recrash_report`
- [x] All existing tests pass (11 SentryCrashTests, 52 SentryCrashReportConverterTests)

Fixes #7888

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Prevent crash-loop from malformed recrash reports (#7888)
+
 ### Features
 
 - Add Set conformance to SentryAttributeValue (#7876)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Prevent crash-loop from malformed recrash reports (#7888)
+- Prevent crash-loop from malformed recrash reports (#7892)
 
 ### Features
 
--- a/Sources/Sentry/SentryCrashReportConverter.m
+++ b/Sources/Sentry/SentryCrashReportConverter.m
@@ -56,15 +56,18 @@
         self.userContext = userContextMerged;
 
         NSDictionary *crashContext;
-        // This is an incomplete crash report
-        if (nil != report[@"recrash_report"][@"crash"]) {
-            crashContext = report[@"recrash_report"][@"crash"];
+        id recrashReport = report[@"recrash_report"];
+        NSDictionary *recrashDict =
+            [recrashReport isKindOfClass:[NSDictionary class]] ? recrashReport : nil;
+
+        if (nil != recrashDict[@"crash"]) {
+            crashContext = recrashDict[@"crash"];
         } else {
             crashContext = report[@"crash"];
         }
 
-        if (nil != report[@"recrash_report"][@"binary_images"]) {
-            self.binaryImages = report[@"recrash_report"][@"binary_images"];
+        if (nil != recrashDict[@"binary_images"]) {
+            self.binaryImages = recrashDict[@"binary_images"];
         } else {
             self.binaryImages = report[@"binary_images"];
         }
--- a/Sources/SentryCrash/Recording/SentryCrash.m
+++ b/Sources/SentryCrash/Recording/SentryCrash.m
@@ -443,15 +443,18 @@ SYNTHESIZE_CRASH_STATE_PROPERTY(BOOL, crashedLastLaunch)
 
 - (void)doctorReport:(NSMutableDictionary *)report
 {
-    NSMutableDictionary *crashReport = report[@SentryCrashField_Crash];
-    if (crashReport != nil) {
-        crashReport[@SentryCrashField_Diagnosis] =
-            [[SentryCrashDoctor doctor] diagnoseCrash:report];
+    id crashField = report[@SentryCrashField_Crash];
+    if ([crashField isKindOfClass:[NSMutableDictionary class]]) {
+        crashField[@SentryCrashField_Diagnosis] = [[SentryCrashDoctor doctor] diagnoseCrash:report];
     }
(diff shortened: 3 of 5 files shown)


NinjaLikesCheez (approved):
LGTM

NinjaLikesCheez (approved):

philprime (approved):

PR #7899, opened 2026-05-15
feat: add app.vitals.start.reason to standalone app start

itaybre:
## Summary
- Add `app.vitals.start.reason` attribute set to `"launch"` on standalone app start transactions
- Add `SentrySpanDataKeyAppVitalsStartReason` constant to `SentrySpanDataKey`

On iOS the reason is always "launch" since the SDK only tracks app starts from the standard app launch path (no broadcast/service equivalents like Android).

## Test plan
- [x] `testReport_shouldSetStartReasonToLaunch`

#skip-changelog


NinjaLikesCheez (approved):

Sources/Swift/Integrations/AppStartTracking/AppStartReportingStrategy.swift
```diff
@@ -60,6 +60,9 @@ struct StandaloneTransactionStrategy: AppStartReportingStrategy {
         if let screen = SentryAppStartMeasurementProvider.consumeAppStartScreen() {
             tracer.setData(value: screen, key: "app.vitals.start.screen")
         }
+        // Only regular app launches are supported for now. Future work could add
```
philprime, line 67: m: Please create follow-up issues for this. There is a high certainty that we will be asked at some point if we support this.

philprime (approved):
LGTM

Sources/Swift/Integrations/AppStartTracking/AppStartReportingStrategy.swift
itaybre: Created: https://github.com/getsentry/sentry-cocoa/issues/7922

PR #7916, opened 2026-05-19
ref(ci): change update-package-sha.sh to use loop approach

philprime:
This pull request refactors the checksum update logic in the `scripts/update-package-sha.sh` script to make it more maintainable and robust. Instead of using hardcoded and repetitive commands for each framework variant, the script now uses a loop over a structured list of zip files and their corresponding markers, improving scalability and reducing the likelihood of errors.

Extracted from #7598 

#skip-changelog


NinjaLikesCheez (approved):
LGTM

PR #7918, opened 2026-05-19
feat(objc): SentryObjC wrapper SDK with compat architecture

### LINKED ISSUE #6342: Create an ObjC SDK

philprime:
SentryObjC adds an experimental pure Objective-C entry point for ObjC and ObjC++ consumers that cannot enable Clang modules or import generated Swift headers. It keeps the main SDK unchanged while exposing startup, options, scope/event types, session replay, feedback, metrics, logging, and related APIs through uniform `SentryObjC*` types.

Closes #6342. See also #4543.

**Architecture**

SentryObjC now follows the two-target design in `develop-docs/SENTRY-OBJC.md`:

```text
ObjC / ObjC++ consumer
  -> SentryObjC (hand-written Foundation-only ObjC headers)
  -> SentryObjCCompat (per-type Swift wrappers using internal import)
  -> SentryObjCInternal / Sentry
```

`Sources/SentryObjC/Public/` is the public ABI anchor: hand-written headers, uniform `SentryObjC*` names, no SDK imports, no `-Swift.h`, and matching platform guards through `SentryObjCDefines.h`. `Sources/SentryObjCCompat/` implements matching `@objc(SentryObjCFoo)` wrappers that delegate to the SDK and convert callback boundaries so Objective-C users stay on wrapper types.

**Distribution And Validation**

The SPM product exposes one `SentryObjC` library backed by `SentryObjC -> SentryObjCCompat -> SentryObjCInternal`, and the xcframework release pipeline builds `SentryObjC-Dynamic.xcframework.zip`. This branch also adds API extraction for `sdk_api_objc.json`, compat drift detection through `sdk_api_objc.diff.json`, and the `iOS-ObjectiveCpp-NoModules` sample as an integration check.

**Testing**

- `make build-framework-objc`
- `make build-sentryobjc-xcframework-local SDKS=iphonesimulator`
- `make build-sample-iOS-ObjectiveCpp-NoModules`
- `make test-ios ONLY_TESTING=SentryObjCCompatTests`
- Checked that consumer-facing SentryObjC headers do not include generated `*-Swift.h` headers.
- Full CI pass pending `ready-to-merge`.

**Known Gaps**

- ObjC header and Swift wrapper drift detection is partially automated, but it does not catch every mismatch category yet.
- A dedicated CI guard should still enforce the no-`-Swift.h` boundary for `SentryObjC.h` compiled with `-fmodules=NO`.

**Next Steps**

- [ ] Remove commented out `.binaryTarget(..)` target blocks in `Package.swift` files using a temporary patch approach.
- [ ] Expand verification tooling for linking without UI frameworks.
- [ ] Create public user-facing documentation.


Sources/Swift/Protocol/SentryAttributeContent.swift
```diff
@@ -220,4 +220,11 @@ extension SentryAttributeContent: ExpressibleByIntegerLiteral {
         self = .integer(value)
     }
 }
+
+extension SentryAttributeContent: SentryAttributeValue {
```
noahsmartin: I don't think this is necessary/used? Doesn't seem like we would need to add public API to make the objc wrapper

Package.swift
```diff
@@ -46,6 +47,11 @@ var targets: [Target] = [
         url: "https://github.com/getsentry/sentry-cocoa/releases/download/9.14.0/Sentry-WithoutUIKitOrAppKit-WithARM64e.xcframework.zip",
         checksum: "7725afd5092e9203939a30e750ef395338662a8ad0b57c5a9534d2d6a09cf52a" //Sentry-WithoutUIKitOrAppKit-WithARM64e
     ),
+    // .binaryTarget(
```
noahsmartin, line 50: I think I get what you're doing here. This is so that we can build a dynamic variant of it in CI right? I wonder if that is really necessary? What I would propose is building a static xcframework in CI, like I have in my PR: https://github.com/getsentry/sentry-cocoa/pull/7934 

I static xcframework is purely more useful than a dynamic one, because users could still use it in a dynamic framework if they wanted to. If there is demand for us to provide a dynamic version too then we can always bring this back in a second PR, but since you can make a dynamic version from a static version and you can't make a static version from a dynamic one, I'd start with a static-only distribution.

.github/workflows/api-stability.yml
```diff
@@ -88,6 +91,39 @@ jobs:
             exit 1
           fi
 
+      - name: Diagnose breaking changes for SentryObjC
```
noahsmartin, line 93: I get why we have this step, I'm wondering what the tests for SentryObjCCompat and "ObjC/ObjCCompat" are for? The Package.swift changes only have two new targets. "SentryObjCCompat" and "SentryObjC" So I'm not sure why we would need 3 new API stability checks?

Sources/Swift/Protocol/SentryMetric.swift
```diff
@@ -65,7 +65,7 @@ public struct SentryMetric {
     ///   - value: The numeric value of the metric
     ///   - unit: The unit of measurement for the metric value (optional)
     ///   - attributes: A dictionary of structured attributes to add to the metric
-    internal init(
+    @_spi(Private) public init(
```
noahsmartin: I also think this change is not needed

.github/workflows/api-stability.yml
philprime: We want these API stability checks to know if the API changes unexpectedly (as we did for a while since you introduced this great addition):

- Public API of Sentry in `sdk_api.json`
- Public API of SentrySwiftUI in `sdk_api_sentryswiftui.json`
- Public API of SentryObjC in `sdk_api_objc.json`

I introduced the `sdk_api_objccompat.json` for SentryObjCCompat to also track if the public API of the Swift implementation changes. This is useful to detect drift between the hand-written Objective-C headers in SentryObjC and the implementation it must match in SentryObjCCompat, otherwise we have runtime crashes with unknown selectors. 

That drift/diff in public API is in `sdk_api_objc.diff.json`, and while most of the code paths are tested in SentryObjCTests, this is just an additional safeguard to reduce risk of drift.

As the `sdk_api_objccompat.json` is mostly means to create the diff JSON, I'll remove it from git & CI and treat it as a temporary file instead.

Package.swift
philprime: I extended the already existing build setup and removed a lot of custom scripting from this PR.
I see the value of having a static xcframework too, but to not further increase the scope of this PR, I am not going add it for now.
We can do this in a follow-up PR.

I'll keep this commented out, because the clone would currently fail and we need a temporary patch to get CI properly working.
I'll do this also in a follow-up PR.

Sources/Swift/Protocol/SentryMetric.swift
philprime: `SentryObjCMetric` had an initializer which was used from tests, and to create the wrapped `SentryMetric` I had to change it from `internal` to `@_spi(Private) public`. 

I was able to change `SentryObjCCompat` to work without it, but I had to remove the Objective-C tests which had to create an instance of `SentryObjCMetric`.

Sources/Swift/Protocol/SentryAttributeContent.swift
philprime: It was used in the SentryObjCMetric to map the attributes from SentryObjCAttribute to SentryAttribute. I was able to do it all in SentryObjCCompat layer, so I could revert this

Sources/SentryObjCCompat/SentryObjCTransactionContext.swift
```diff
@@ -0,0 +1,73 @@
...
+    @objc public init(name: String, operation: String, traceId: SentryObjCId, spanId: SentryObjCSpanId, parentSpanId: SentryObjCSpanId?, parentSampled: SentryObjCSampleDecision, parentSampleRate: NSNumber?, parentSampleRand: NSNumber?) {
+        super.init(TransactionContext(name: name, operation: operation, trace: traceId.wrapped, spanId: spanId.wrapped, parentSpanId: parentSpanId?.wrapped, parentSampled: parentSampled.underlying, parentSampleRate: parentSampleRate, parentSampleRand: parentSampleRand))
+    }
+
+    @objc public var name: String {
+        wrappedTransaction.name
+    }
+
+    @objc public var nameSource: SentryObjCTransactionNameSource {
+        // The ObjC property returns a Swift-defined @objc enum that SPM can't
```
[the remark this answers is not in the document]
philprime: We have locked down the behavior using tests. This is a workaround as I could not get it to work otherwise.

Package@swift-6.1.swift
```diff
@@ -110,6 +128,26 @@ targets += [
         ])
 ]
 
+// BEGIN:OBJC_WRAPPER
+products.append(.library(name: "SentryObjC", targets: ["SentryObjC"]))
+targets += [
+    .target(
+        name: "SentryObjCCompat",
+        dependencies: ["SentryObjCInternal"],
+        path: "Sources/SentryObjCCompat"),
```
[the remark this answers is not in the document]
philprime: Will fix this in a follow-up PR to improve the validation for linkage to UI frameworks

Sources/SentryObjC/Public/SentryObjCAttributeContent.h
```diff
@@ -0,0 +1,56 @@
...
+/// Creates a double attribute content with the specified value.
++ (instancetype)double:(double)value;
+/// Creates a string array attribute content with the specified values.
++ (instancetype)stringArray:(NSArray<NSString *> *)values;
+/// Creates a boolean array attribute content with the specified values.
++ (instancetype)booleanArray:(NSArray<NSNumber *> *)values;
+/// Creates an integer array attribute content with the specified values.
++ (instancetype)integerArray:(NSArray<NSNumber *> *)values;
+/// Creates a double array attribute content with the specified values.
++ (instancetype)doubleArray:(NSArray<NSNumber *> *)values;
```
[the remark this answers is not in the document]
philprime: I checked this one. It does not appear to be a real issue.

`SentryObjCAttributeContent.type` is declared as `readonly, copy` in the ObjC header, and the Swift wrapper exposes it as a get-only computed property derived from the wrapped `SentryAttributeContent` enum case. There is no Swift setter like `set { attributeContent.type = newValue }`, so ObjC callers cannot pass an `NSMutableString` into `type` and mutate it later.

The cited header lines also point at factory methods, not a mutable `type` setter. No code change needed here.

Package.swift
```diff
@@ -107,6 +125,25 @@ targets += [
...
+        name: "SentryObjC",
+        dependencies: ["SentryObjCCompat"],
+        path: "Sources/SentryObjC",
+        publicHeadersPath: "Public",
+        cSettings: [
+            .headerSearchPath("Public")
+        ]
+    )
+]
+// END:OBJC_WRAPPER
```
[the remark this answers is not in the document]
philprime: I checked this locally and this does not appear to be a real issue.

`prepare-package.sh --remove-binary-targets true` rewrites the initial `var targets` array, but it does not remove the later source-backed `targets += [...]` block that defines `SentryObjCInternal`. The `// BEGIN:OBJC_WRAPPER` block is appended after that, so `SentryObjCCompat` still references an existing target.

I verified this by running the script against temp copies of all three manifests (`Package.swift`, `Package@swift-6.1.swift`, and `Package@swift-6.2.swift`) and then running `swift package dump-package`; manifest validation succeeds and includes `SentryObjCInternal`.

The current CI also does not show the package breakage this predicts: the relevant SPM/sample jobs are passing, including `Build with SPM`, `Sample SPM Debug`, and `Sample DistributionSample Debug`.

NinjaLikesCheez (approved):

PR #7920, opened 2026-05-19
fix: Subsequent VC transactions lost when launch profiling is active

### LINKED ISSUE #7919: UIViewController transactions lost during launch profiling with trace lifecycle
## Problem

When launch profiling with trace lifecycle is active alongside `enableTimeToFullDisplayTracing`, UIViewController transactions **after the first one** are silently lost. The first VC transaction is intentionally discarded as part of the launch tracer (see #7602), but subsequent VC navigations should produce their own independent transactions.

## Root cause

The first VC becomes a child of the launch tracer via `SentryPerformanceTracker.startSpanWithName:`. This is expected — the launch tracer drives profiling and is discarded. However, two issues prevent subsequent VCs from getting their own transactions:

1. **TTD tracker not created for launch tracer children.** The TTD tracker was only created for root `SentryTracer` spans. Since the first VC was a child span (not a root tracer), no TTD tracker was created. Without a TTD tracker, `sentry_stopAndDiscardLaunchProfileTracer` was never called, so `sentry_launchTracer` was never cleared. All subsequent VCs kept becoming children of the (never-cleared) launch tracer and were discarded along with it.

2. **Stale span on the active span stack.** The first VC's child span was pushed onto the active span stack (because it was empty) but never popped (because it had a `parentSpanId` from the launch tracer). Even if the launch tracer was eventually cleared, subsequent VCs found this stale span and became its children instead of getting their own root transactions.

## Why it's intermittent

The launch profiling config is persisted from the **previous** launch. Whether `sentry_launchTracer` exists depends on:

- **First launch after install**: no persisted config → no launch tracer → works
- **Subsequent launches, sampled out**: sampling decision is NO → no launch tracer → works
- **Subsequent launches, sampled in**: launch tracer created → bug triggers → broken

## Fix

See #7920.

itaybre:
## Summary

When launch profiling with trace lifecycle is active, UIViewController transactions after the first one are silently lost. The first VC becomes a child of the launch tracer (intentionally discarded per #7602 for profiling), but subsequent VCs should get their own independent root transactions.

Two issues caused subsequent VCs to be lost:

1. **No TTD tracker for launch tracer children.** The TTD tracker was only created for root `SentryTracer` spans. When the first VC was a child of the launch tracer, no TTD tracker was created, so `sentry_stopAndDiscardLaunchProfileTracer` was never called and `sentry_launchTracer` was never cleared. All subsequent VCs kept becoming children of the launch tracer and were discarded along with it.

2. **Stale span on the active span stack.** The first VC's child span was pushed onto `SentryPerformanceTracker`'s active span stack (because the stack was empty) but never popped (because it had a `parentSpanId`). After the launch tracer was eventually cleared, subsequent VCs found this stale span and became its children instead of getting their own root transactions.

## Changes

**`SentryDefaultUIViewControllerPerformanceTracker.m`**
- Allow TTD tracker creation when the VC span's root tracer has `app.lifecycle` operation (the launch tracer), so TTID fires and stops the launch tracer
- Only push root transactions (no `parentSpanId`) onto the active span stack, preventing stale launch tracer children from capturing subsequent VCs

Fixes #7919

## Test plan

- [x] 4 new tests in `SentryVCTrackerLaunchProfilingTests`
- [x] All 25 existing `SentryUIViewControllerPerformanceTrackerTests` pass
- [x] All 18 `SentryTimeToDisplayTrackerTest` pass
- [x] All 8 `SentryPerformanceTrackingIntegrationTests` pass
- [x] All 16 profiler tests (`SentryAppLaunchProfilingTests` + `SentryAppStartProfilingConfigurationChangeTests`) pass


Tests/SentryTests/Integrations/Performance/UIViewController/SentryVCTrackerLaunchProfilingTests.swift
```diff
@@ -0,0 +1,205 @@
+// Profiling is only supported on iOS
+#if os(iOS)
+
+import ObjectiveC
+@_spi(Private) @testable import Sentry
+@_spi(Private) import SentryTestUtils
+import XCTest
+
+class SentryVCTrackerLaunchProfilingTests: XCTestCase {
```
philprime, line 9: `l`: As this is heavily coupled to the app lifecycle, would it make sense to build an additional integration test in a real app sample? I'll leave the decision to you if you want to do this or not worth the effort.

philprime (approved):
LGTM

Tests/SentryTests/Integrations/Performance/UIViewController/SentryVCTrackerLaunchProfilingTests.swift
itaybre: Done!

PR #7927, opened 2026-05-20
fix(session): Prevent SessionTracker crash with profiling

### LINKED ISSUE #7921: Bad access crash in SessionTracker
### Platform

iOS

### Environment

Production

### Installed

Manually

### Other Error Monitoring Solution

No

### Other Error Monitoring Solution Name

_No response_

### Version

9.9.0

### Xcode Version

26.4 / 26.5

Definitely 8.35.1, we suspect it worked in 9.8.0 as well.

### Steps to Reproduce

We do not have clean repro, however:

It seems the bug requires having profiling options set.

We suspect the issue is caused by: https://github.com/getsentry/sentry-cocoa/commit/20c62b758

### Expected Result

No crash

### Actual Result

Nondeterministic crashing

We saw 30 users crash in 1d of 1% rollout.  We would expect 3000 users/day at 100%.
https://dropbox.sentry.io/issues/7490337990/events/1711751ea6b7448d95ab04bf77be7c4e/?project=4505546600087552

Probably not

denrase:
## :scroll: Description

Fixes a crash caused by racing access to the global profiling configuration while reevaluating the profile session sample decision. This was introduced by #7704 because the reevaluation call was done off the main queue.

Initially i just wanted to move reevaluation outside of the async block, but I encountered an additional sampling correctness issue.

Profile session sampling is now reevaluated only when a new Sentry session is started, instead of every time the app returns to the foreground. This matches Android behavior and keeps the sample decision stable for the lifetime of a session.

Closes #7921

Unit tests.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fix race conditions in scope observer iteration and propagation context locking (#7897)
 - Support stack traces for standalone clients (#7817)
+- Prevent SessionTracker crash with profiling (#7927)
 
 ## 9.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix race conditions in scope observer iteration and propagation context locking (#7897)
 - Support stack traces for standalone clients (#7817)
 - Prevent SessionTracker crash with profiling (#7927)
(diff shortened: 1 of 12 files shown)


philprime (approved):
LGTM

PR #7928, opened 2026-05-20
fix(replay): Defer startup until window exists

philprime:
Session Replay startup now stays pending until the SDK can resolve a foreground window. This handles apps that initialize Sentry before UIApplication or scene windows are available, while keeping lifecycle notifications from starting replay unless sampling or manual start requested it.

**Pending Startup Gate**

Replay startup is retried from app and scene activation notifications only while isPendingStart is true, and the flag is cleared before creating the replay instance.

**Lazy Window Lookup**

The integration resolves the application at retry time and uses getWindows() to share foreground scene and app delegate fallback logic.

### DIFF
--- a/Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
+++ b/Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
@@ -23,7 +23,6 @@ public class SentrySessionReplayIntegration: NSObject, SwiftIntegration, SentryS
     private let replayOptions: SentryReplayOptions
     private let rateLimits: RateLimits
     private let random: SentryRandomProtocol
-    private let application: SentryApplication?
     private var startedAsFullSession = false
     private let experimentalOptions: SentryExperimentalOptions
     private let notificationCenter: SentryNSNotificationCenterWrapper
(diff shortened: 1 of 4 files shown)


Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -175,44 +196,93 @@ public class SentrySessionReplayIntegration: NSObject, SwiftIntegration, SentryS
...
         runReplayForAvailableWindow()
     }
 
     private func runReplayForAvailableWindow() {
-        if let windowCount = application?.getWindows()?.count, windowCount > 0 {
-            SentrySDKLog.debug("[Session Replay] Running replay for available window")
-            // If a window its already available start replay right away
-            startWithOptions(replayOptions, experimentalOptions: experimentalOptions, fullSession: startedAsFullSession)
-        } else {
-            SentrySDKLog.debug("[Session Replay] Waiting for a scene to be available to started the replay")
```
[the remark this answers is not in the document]
itaybre: Worth reviewing this

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
```diff
@@ -95,10 +110,16 @@ public class SentrySessionReplayIntegration: NSObject, SwiftIntegration, SentryS
         
         SentrySDKInternal.currentHub().registerSessionListener(self)
         dependencies.reachability.add(self)
+
+        // Create observers of the application lifecycle for UIKit and SwiftUI
+        notificationCenter.addObserver(self, selector: #selector(applicationDidBecomeActiveHandler), name: UIApplication.didBecomeActiveNotification, object: nil)
```
itaybre, line 118: `m`: We have 2 observers for `UIApplication.didBecomeActiveNotification`, this one and [resume](https://github.com/getsentry/sentry-cocoa/blob/b9544d3eb3b3d0ac3492625ba19b029f62dc4a40/Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift#L323)
We should double check there are no conflicts with both since when you call `notificationCenter.removeObserver(self, name: UIApplication.didBecomeActiveNotification, object: nil)` it will remove all observers

Tests/SentryTests/Integrations/SessionReplay/SentrySessionReplayIntegrationTests.swift
```diff
@@ -115,6 +115,18 @@ class SentrySessionReplayIntegrationTests: XCTestCase {
         let sut = try getSut()
         XCTAssertNil(sut.sessionReplay)
     }
+
+    func testLifecycleNotificationDoesNotStartReplayWhenStartIsNotPending() throws {
```
itaybre: `m`: Would you mind adding another test case for a delayed start?

itaybre (approved):
Almost LGTM, some comments and also verify broken tests

Sources/Swift/Integrations/SessionReplay/SentrySessionReplayIntegration.swift
philprime: I created `SentrySessionReplayBackgroundForegroundObserver` so we can have two observers and manage them individually

PR #7931, opened 2026-05-20
feat: add appStartSampleRate to experimental options

### LINKED ISSUE #7235: API for setting app launch span sample rate
### Problem Statement

Users can set a global sample rate, or use a custom sampler, but there isn't a clear way to set a sample rate for app launches. Many apps will have hundreds of spans throughout the lifetime of a process but only one app launch. A reasonable global sample rate might greatly reduce the number of app launch spans that get sent to Sentry even though they are arguably the most important spans. We should provide an easy way to customize the sample rate just for app launch spans.

### Solution Brainstorm

One options would be to have an API to identify the span/transaction as being for app launch in the custom sampler, so you wouldn't set a global sample rate but you'd check if the span was app launch and return a different sample rate than a default value.

_No response_

itaybre:
## :scroll: Description

Add `experimental.appStartSampleRate` to override `tracesSampleRate` for standalone app start transactions. When set, the sampling logic uses this rate instead of `tracesSampleRate` for transactions with the `app.start` operation. A custom `tracesSampler` callback still takes precedence.

Fixes https://github.com/getsentry/sentry-cocoa/issues/7235

Users can set a global sample rate or use a custom sampler, but there isn't a clear way to set a sample rate specifically for app launches. A reasonable global sample rate might greatly reduce the number of app launch spans even though they are arguably the most important spans. This provides a simple dedicated API for that.

- Added unit tests for the new option property (default, nil, bounds validation)
- Added sampling behavior tests in `SentryHubTests` covering:
  - `appStartSampleRate` is used for `app.start` operation
  - `appStartSampleRate` is NOT used for other operations
  - `tracesSampler` callback takes precedence
  - `nil` falls back to `tracesSampleRate`
  - Correct sample rate is propagated to the transaction context
- Added sample app test override (`SentrySDKOverrides.Performance.appStartSampleRate`)
- Regenerated `sdk_api.json`

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Add `appStartSampleRate` to experimental options for overriding `tracesSampleRate` on standalone app start transactions (#7235)
+- Add `appStartSampleRate` to experimental options for overriding `tracesSampleRate` on standalone app start transactions (#7931)
 - Add standalone app start tracing as an experimental option (#7660), enable it via `options.experimental.enableStandaloneAppStartTracing = true`
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features
 
+- Add `appStartSampleRate` to experimental options for overriding `tracesSampleRate` on standalone app start transactions (#7235)
 - Add standalone app start tracing as an experimental option (#7660), enable it via `options.experimental.enableStandaloneAppStartTracing = true`
 
 ### Fixes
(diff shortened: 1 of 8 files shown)


buenaflor (commented):
@itaybre we shouldn't add this separate sample rate, we decided to go with making use of the `tracesSampler`

see @romtsn comment: https://github.com/getsentry/sentry-cocoa/issues/7235#issuecomment-4238032805

itaybre:
> @itaybre we shouldn't add this separate sample rate, we decided to go with making use of the `tracesSampler`
> 
> see @romtsn comment: [#7235 (comment)](https://github.com/getsentry/sentry-cocoa/issues/7235#issuecomment-4238032805)

Ohh, I didn't see that comment, I will close this then

PR #7933, opened 2026-05-21
Adopt .editorconfig for style

NinjaLikesCheez:
## :scroll: Description

This change adds an [`.editorconfig`](https://editorconfig.org/) file which codifies some of the in-use or declared code/file styles. 

This is particularly useful if, like me, you prefer things like tabs instead of spaces and are constantly having to fix up files to align with the project. With this config, your IDE (sometimes automatically, sometimes with an extension or setting) will pick up the projects preferred styling and override your global IDE settings to use the values in this file.

This helps align style across the project and alleviates the need for the developer to manually change their settings to achieve it. 

Not a code change - so no tests required :) 

#skip-changelog

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,63 @@
+# EditorConfig is awesome: https://editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+
+# Swift — matches existing source files
+[*.swift]
+indent_size = 4
+
+# Objective-C, C, C++ — matches .clang-format (IndentWidth: 4)
+[*.{m,h,c,mm,cpp,cc,hpp}]
+indent_size = 4
+
+# Ruby — standard Ruby convention
+[*.{rb,podspec}]
+indent_size = 2
+
+[{Gemfile,Fastfile,Matchfile,Appfile,Deliverfile}]
+indent_size = 2
+
+# Shell — matches existing scripts
+[*.sh]
+indent_size = 4
+
+# Python — PEP 8
+[*.py]
+indent_size = 4
+
+# YAML — matches dprint.json (indentWidth: 2)
+[*.{yml,yaml}]
+indent_size = 2
+
+# JSON — matches dprint.json (indentWidth: 2)
+[*.json]
+indent_size = 2
+
+# TOML
+[*.toml]
+indent_size = 2
+
+# Markdown — matches existing files
+[*.md]
+indent_size = 2
+trim_trailing_whitespace = false
+
+# Makefiles require tabs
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,63 @@
+# EditorConfig is awesome: https://editorconfig.org
```
philprime, line 1: `h`: We already have adopted `dprint` for formatting. Is this one aligned with dprint's rules? Do we need to adapt dprint's config to match what you have decided here? How are these rules enforced with `pre-commit`

NinjaLikesCheez: `editorconfig` isn't a formatter but rather a way to tell the IDE what the preferences are - so it will never change/format existing code. What it will do is tell your IDE that the `tab` key now means `x amount of spaces` in a given file-type. That means there is no enforcement - existing formatters will handle that, this is purely affecting how the IDE behaves w.r.t styling.

I used what was in `dprint.json` if a preference was stated (or in clang-format, etc) - but I didn't check the defaults that may be implicit. I will do that to ensure it's congruent.

philprime: I know `.editorconfig` and I use it in other projects too, so it's fine if we adopt it. All I am asking for is that we have it aligned with what is normal, e.g. Xcode also has changed from 4 spaces to 2 spaces and it's necessary to change the settings so that `CTRL` + `I` works. 

In the end the pre-commit hooks should win and not allow unformatted commits, but if an IDE is instructed to do one thing and formats on save (VSCode does that), but the pre-commit hooks then format it again, it renders hard to use

NinjaLikesCheez: > ... if an IDE is instructed to do one thing and formats on save (VSCode does that), but the pre-commit hooks then format it again, it renders hard to use

A good point I hadn't considered - I've put a table below to show the comparison & alignment. I've aligned with either: rules we currently have in formatters, 'standards' and default style guidelines (i.e. PEP8), or the existing default we have in the repo.

| Type | Formatter | Formatter setting | `.editorconfig` setting | 
|--|--|--|--|
ObjC/C/C++/ObjC++ | `clang-format` | indent: 4 (+ spaces implicitly inherited from 'based on WebKit style') | indent: 4, spaces | 
| JSON | `dprint` | indent: 4, tabs: false | indent 4, spaces |
| YAML | `dprint` | indent: 2, spaces inferred | indent 2, spaces | 
| Swift | `swiftlint` | No enforcement | indent 4, spaces based on existing source |
| Python | None | None | Align with PEP8 standard |
| Makefile | None | None | Tabs to align with Makefiles standard |
| Shell | None | None | Align with existing scripts |
| Ruby | None | None | Align with Ruby Style Guide |

I removed some entries in the `da8fff4f1625` that either didn't have a formatter or doesn't have set rules. I hope this answers your questions, but if not - let me know!

philprime (approved):
LGTM, let's just try it and iterate if we encounter issues

PR #7934, opened 2026-05-21
feat: add SentryObjC wrapper SDK

### LINKED ISSUE #4543: Replay options not visible from Objective-C without @import

### LINKED ISSUE #6342: Create an ObjC SDK

noahsmartin:
## :scroll: Description

Adds **SentryObjC**, a pure Objective-C wrapper around the Sentry Cocoa SDK. It provides the full Sentry public API surface through hand-written ObjC headers backed by Swift `@objc` wrapper classes, so that ObjC and ObjC++ projects — including those that cannot enable Clang modules — can use Sentry without importing Swift headers or configuring `-fmodules`.

### Architecture

A **two-target design** where every exposed SDK type gets a dedicated `SentryObjC*`-prefixed wrapper:

| Target | Language | Role |
|---|---|---|
| `SentryObjC` | ObjC | Hand-written `@interface` declarations + umbrella header. Pure ObjC, no behavior. |
| `SentryObjCCompat` | Swift | Per-type `@objc` wrapper classes using `internal import Sentry`. Delegates to the real SDK. |

At link time, the ObjC `@interface SentryObjCFoo` declared in the header resolves to the Swift `@objc(SentryObjCFoo)` class in `SentryObjCCompat`. Consumer translation units only ever see ObjC headers — no `-Swift.h`, no `@import`.

**SPM dependency chain:** `SentryObjC → SentryObjCCompat → SentryObjCInternal` (compiles the full SDK from source).

### What's included

- **~60 public ObjC headers** (`Sources/SentryObjC/Public/`) covering: SDK entry point, Options, Event, Scope, Span, Breadcrumb, User, Hub, Client, Feedback, Replay, Metrics, Logging, and more
- **~60 Swift wrapper implementations** (`Sources/SentryObjCCompat/`) with type-safe delegation to the underlying SDK
- **Platform-conditional compilation** for iOS/tvOS-only APIs (session replay, feedback) using shared `SentryObjCDefines.h` macros
- **XCFramework build pipeline** integrated into the standard variant system (`SentryObjC-Dynamic` — dynamic framework, same flow as Sentry-Dynamic and SentrySwiftUI)
- **API stability tracking** via clang AST extraction (`sdk_api_objc.json`, `sdk_api_objccompat.json`) and drift detection (`sdk_api_objc.diff.json`)
- **Full test coverage** — ObjC integration tests (`Tests/SentryObjCTests/`) and Swift wrapper unit tests
- **Architecture documentation** at `develop-docs/SENTRY-OBJC.md`
- **Sample app** updated at `Samples/iOS-ObjectiveCpp-NoModules/`

### Distribution

- **SPM:** `SentryObjC` product in `Package.swift` (compile-from-source); binary target placeholder for `SentryObjC-Dynamic` xcframework (commented out, enabled at release time)
- **XCFramework:** `SentryObjC-Dynamic.xcframework` built via the standard `build-xcframework-variant.sh` → `assemble-xcframework.sh` pipeline. Dynamic-only (no static variant). Build locally with `make build-sentryobjc-xcframework-local`

Many projects cannot use `@import Sentry` or `#import <Sentry/Sentry-Swift.h>`:
- **React Native** (≤0.76): AppDelegate is `.mm` (ObjC++), modules disabled by default
- **Haxe**: Build toolchain conflicts with `-fmodules` / `-fcxx-modules`
- **Custom build systems**: May not support module imports

With modules disabled, Swift-bridged APIs (`SentrySDK`, `SentryOptions`, `options.sessionReplay`, etc.) are inaccessible from ObjC++. SentryObjC solves this by providing a pure ObjC interface that works in any compilation mode.

Fixes #6342, fixes #4543

- Full ObjC integration test suite (`Tests/SentryObjCTests/`) covering all wrapped types
- Swift unit tests for wrapper logic
- Local xcframework build verified: `make build-sentryobjc-xcframework-local SDKS=iphonesimulator`
- XCFramework validation scripts pass (format + architecture checks)
- Sample app (`iOS-ObjectiveCpp-NoModules`) builds and runs with the wrapper
- API stability JSON files generated and committed

## :pencil: Checklist

#skip-changelog


philprime:
Superseded by the work in #7918

PR #7941, opened 2026-05-22
ref: End standalone app start at didFinishLaunching

### LINKED ISSUE #6883: Capture app start by default

itaybre:
## Summary

- Standalone app start transactions now end when `UIApplication.didFinishLaunchingNotification` is posted instead of waiting for the first frame render
- The "Initial Frame Render" child span is no longer emitted for standalone transactions since it does not apply
- Non-standalone (attach-to-transaction) behavior is unchanged

Decided to do this change on standalone app starts since it is still experimental and thus we can make "breaking changes"

Fixes #6883

## Test plan

- [x] All existing `AppStartReportingStrategyTests`, `SentryAppStartTrackerTests`, and `SentryBuildAppStartSpansTests` updated and passing
- [ ] Test with sample app: enable `enable-standalone-app-start-tracing`, verify app start transaction ends at didFinishLaunching and has no "Initial Frame Render" span

#skip-changelog

### DIFF
--- a/Sources/Sentry/SentryBuildAppStartSpans.m
+++ b/Sources/Sentry/SentryBuildAppStartSpans.m
@@ -130,14 +130,15 @@ sentryBuildAppStartSpansInternal(SentryTracer *tracer,
     }
     [appStartSpans addObject:didFinishLaunching];
 
-    id<SentrySpan> frameRenderSpan
-        = sentryBuildAppStartSpan(tracer, appStartSpanParentId, operation, @"Initial Frame Render");
-    [frameRenderSpan setStartTimestamp:appStartMeasurement.didFinishLaunchingTimestamp];
-    [frameRenderSpan setTimestamp:appStartEndTimestamp];
-    if (isStandalone && startType != nil) {
-        [frameRenderSpan setDataValue:startType forKey:SentrySpanDataKeyAppVitalsStartType];
+    // Standalone app starts end at didFinishLaunching, so the frame render span
+    // does not apply. Non-standalone starts still end at the first frame render.
+    if (!isStandalone) {
+        id<SentrySpan> frameRenderSpan = sentryBuildAppStartSpan(
+            tracer, appStartSpanParentId, operation, @"Initial Frame Render");
+        [frameRenderSpan setStartTimestamp:appStartMeasurement.didFinishLaunchingTimestamp];
+        [frameRenderSpan setTimestamp:appStartEndTimestamp];
+        [appStartSpans addObject:frameRenderSpan];
     }
-    [appStartSpans addObject:frameRenderSpan];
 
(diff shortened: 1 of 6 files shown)


Sources/Swift/Integrations/AppStartTracking/SentryAppStartTracker.swift
```diff
@@ -304,6 +314,9 @@ public final class SentryAppStartTracker: NSObject, SentryFramesTrackerListener
     @objc
     private func didFinishLaunching() {
         didFinishLaunchingTimestamp = dateProvider.date()
+        if reportingStrategy is StandaloneTransactionStrategy {
+            buildAppStartMeasurement(didFinishLaunchingTimestamp)
+        }
```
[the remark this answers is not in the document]
itaybre: This is by design, previously if the SDK was initialized after the first VC was displayed, sent transaction could contain invalid data

Sources/Sentry/SentryBuildAppStartSpans.m
```diff
@@ -130,14 +130,15 @@
...
 
-    id<SentrySpan> frameRenderSpan
-        = sentryBuildAppStartSpan(tracer, appStartSpanParentId, operation, @"Initial Frame Render");
-    [frameRenderSpan setStartTimestamp:appStartMeasurement.didFinishLaunchingTimestamp];
-    [frameRenderSpan setTimestamp:appStartEndTimestamp];
-    if (isStandalone && startType != nil) {
-        [frameRenderSpan setDataValue:startType forKey:SentrySpanDataKeyAppVitalsStartType];
+    // Standalone app starts end at didFinishLaunching, so the frame render span
+    // does not apply. Non-standalone starts still end at the first frame render.
+    if (!isStandalone) {
```
philprime, line 135: `h`: Isn't this a change to the logic? Should this be considered a refactor? If yes, why so?

philprime (commented):
Almost LGTM

Sources/Sentry/SentryBuildAppStartSpans.m
itaybre: Technically this isn't neither a bug fix nor a new feature. That's the definition for `ref` in the [engineering practices](https://develop.sentry.dev/engineering-practices/commit-messages/#type)

philprime: From internal chat: this change is related to an experimental feature and fine to be applied.

philprime (approved):
LGTM

PR #7943, opened 2026-05-23
fix: FramesTracker listener-capture deinit race

rickydaricky:
## Description

`SentryFramesTracker.addListener` and `removeListener` strongly captured `listener` in the closure passed to `dispatchAsyncOnMainQueueIfNotMainThread`. When a caller invoked `removeListener(self)` from inside its own `deinit` on a background thread, the dispatched main-queue block tried to retain the listener mid-dealloc. Under iOS 26's hardened ARC this is fatal — `swift_unknownObjectRetain` → `EXC_BREAKPOINT`. Older iOS versions silently survived the retain-against-freed-memory.

The fix captures `ObjectIdentifier(listener)` (a value type) and, for `addListener`, a pre-built `WeakReference(value: listener)` instance — neither of which holds the listener directly. The main-queue hop no longer extends the listener's lifetime, and identity matching against still-alive entries (read via `WeakReference.value` / `objc_loadWeak`) keeps working.

This is the third crash class in the same method area. #7155 fixed the original Obj-C `removeListener` crash; #7839 fixed the `NSConcreteHashTable` race that the comment on the `listeners` property records. The remaining race lived in the Swift closure-capture path, exposed when iOS 26's hardened ARC began trapping retains of mid-dealloc objects.

## Motivating crash (lldb-confirmed)

iPhone 17 Pro Max, iOS 26.5, sentry-cocoa 9.14.0 (current latest), bridged via `@sentry/capacitor@3.2.1`. Every WKWebView reload tore down the host's root `UIViewController`s off-main; one held a reference that called `tracker.removeListener(self)` from its `deinit` as cleanup.

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread',
            stop reason = EXC_BREAKPOINT (code=1, subcode=0x1992b51ac)
    frame #0: libswiftCore.dylib`swift_unknownObjectRetain + 44
    frame #1: App.debug.dylib`closure #1 in SentryFramesTracker.removeListener(_:) + 56
    frame #3: App.debug.dylib`__79-[_SentryDispatchQueueWrapperInternal dispatchAsyncOnMainQueueIfNotMainThread:]_block_invoke + 36
    frame #4: libdispatch.dylib`_dispatch_call_block_and_release + 32
    frame #5: libdispatch.dylib`_dispatch_client_callout + 16
    frame #6: libdispatch.dylib`_dispatch_main_queue_drain.cold.6 + 916
    frame #7: libdispatch.dylib`_dispatch_main_queue_drain + 176
    frame #8: libdispatch.dylib`_dispatch_main_queue_callback_4CF + 44
    frame #9: CoreFoundation`__CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 16
    ...
```

I shipped a downstream workaround (`enableAutoPerformanceTracing: false`) that disables FramesTracker entirely. I'd like to revert that as soon as a fix ships — happy to test a beta from this PR if useful.

## How tested

`testAddAndRemoveListener_DoNotRetainListenerAcrossMainQueueHop` exercises the structural invariant of the fix ("dispatched closures must not strongly capture `listener`") via a `weak` reference assertion, using `TestSentryDispatchQueueWrapper` with `blockBeforeMainBlock = { false }` to defer block execution and observe the listener's lifetime *before* the closures run.

- **With the fix**: passes (`weakListener` is `nil` after the autorelease pool drains; the drained blocks converge `listenersCount` to 0).
- **Without the fix** (verified by reverting `SentryFramesTracker.swift` and re-running): fails with
  > `XCTAssertNil failed: "<FrameTrackerListener: 0x...>" - Closures dispatched by addListener/removeListener must not retain the listener across the main-queue hop`

This is strictly stronger than a no-crash assertion (which would silently pass the unfixed code on pre-iOS-26 simulators / CI runners that don't enforce hardened-ARC retain-on-mid-dealloc trapping).

Verification loop run locally on iOS 26.5 simulator (iPhone 17 Pro):

```
make format
make analyze
make build-ios IOS_SIMULATOR_OS=26.5 IOS_DEVICE_NAME="iPhone 17 Pro"
make test-ios IOS_SIMULATOR_OS=26.5 IOS_DEVICE_NAME="iPhone 17 Pro" ONLY_TESTING=SentryFramesTrackerTests
```

All green: 52 tests, 0 failures.

## Checklist

- [x] Wrote tests
- [x] Added an entry in the `CHANGELOG.md` under the `## Unreleased` section
- [x] No public API change → no `sdk_api.json` regeneration needed
- [x] No platform-specific `#if` → covered by `make build-ios` / `make test-ios`

## Are you willing to submit a PR?

Yes — this is the PR.


Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
```diff
@@ -197,23 +197,70 @@ public class SentryFramesTracker: NSObject {
...
+        // closure body's `WeakReference(value: listener)` call would force
+        // the closure to capture `listener` strongly.
+        let ref = WeakReference(value: listener)
+        dispatchQueueWrapper.dispatchAsyncOnMainQueueIfNotMainThread { [weak self] in
+            guard let self else { return }
+            // Drop zeroed refs and any existing entry for this listener to
+            // avoid duplicate registrations, then append the pre-built ref.
+            self.listeners.removeAll { existing in
+                guard let live = existing.value else { return true }
+                return ObjectIdentifier(live) == targetIdentity
```
itaybre, line 249: `l`: I assume the performance impact for `ObjectIdentifier` is negligible, but would it be worth it saving it in `WeakReference` to avoid instantiating this on each loop?

itaybre (commented):
Hey @rickydaricky, thanks for this fix
I didn't know of that change in iOS 26

LGTM

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Fix crash in `SentryFramesTracker.add/removeListener` when called from a listener's own `init` / `deinit` on a background thread, observed on iOS 26
```
```suggestion
- Fix crash in `SentryFramesTracker.add/removeListener` when called from a listener's own `init` / `deinit` on a background thread, observed on iOS 26 (#7943)
```
This will help users find the relevant PR when reading the changelog and fix the error with the Danger workflow

christophaigner:
@rickydaricky thanks again for your contribution. We would do a `sentry-capacitor` release in the ~next week and would ideally include this fix and the version bump so you can update to the latest version.
Do you have time to incorporate the changes above or should we take over?

rickydaricky:
> @rickydaricky thanks again for your contribution. We would do a `sentry-capacitor` release in the ~next week and would ideally include this fix and the version bump so you can update to the latest version. Do you have time to incorporate the changes above or should we take over?

Hey sorry, I've been a bit swamped this week and out for much of it. I would really appreciate if y'all could take this over. Thanks so much!

Sources/Swift/Core/Integrations/FramesTracking/SentryFramesTracker.swift
itaybre: Double checked and looks like `ObjectIdentifier` is just a wrapper around the pointer address, so there shouldn't be performance impact here, no need to keep it

itaybre (approved):
LGTM

PR #7912, opened 2026-05-18
fix: Move SessionTracker I/O to Background

denrase:
## :scroll: Description

Move session tracker I/O to background.

## :bulb: Motivation and Context

Backport of https://github.com/getsentry/sentry-cocoa/pull/7704

## :green_heart: How did you test it?

- Existing tests run without issue.
- Tested with sample app (backgrounding + new session after 30s in bg threshold).

Bildschirmfoto 2026-05-19 um 13 22 37

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #7944, opened 2026-05-23
fix(session): Prevent SessionTracker crash with profiling

denrase:
## :scroll: Description

Fixes a crash caused by racing access to the global profiling configuration while reevaluating the profile session sample decision. This would be introduced by only backporting #7912, because the reevaluation call was done off the main queue.

Initially i just wanted to move reevaluation outside of the async block, but I encountered an additional sampling correctness issue.

Profile session sampling is now reevaluated only when a new Sentry session is started, instead of every time the app returns to the foreground. This matches Android behavior and keeps the sample decision stable for the lifetime of a session.

Backport of https://github.com/getsentry/sentry-cocoa/pull/7927

Added unit tests.

## :pencil: Checklist


philprime (approved):
LGTM

denrase:
Applied an additional fix for a flaky test: https://github.com/getsentry/sentry-cocoa/commit/78147195e63a98c1e6ca1503df9f36329a688fc2

PR #7949, opened 2026-05-26
feat: skip max app start duration limit for standalone tracing

### LINKED ISSUE #7427: App launches over 30s
### Description

There is some logic in the SDK that [discards app launches over 30s](https://github.com/getsentry/sentry-cocoa/blob/a8cd8757a272d7eb0bb03cbce71865d749ca5d29/Sources/Swift/Integrations/AppStartTracking/SentryAppStartTracker.swift#L24). This is only there because we know there are bugs causing app launches to be reported longer than they should be and instead of fixing the bugs we just drop events that are likely to be bugs. Of course there can still be buggy cases that get reported with < 30s app launch.

Rather than dropping all launches at some threshold, we just fix the bugs and remove this 30s threshold. This should be able to be done after https://github.com/getsentry/sentry-cocoa/issues/6886 which proposes making the default app launch end time applicationDidFinishLaunching

itaybre:
## Description

The 180s `maxAppStartDuration` safety guard that drops suspiciously long app starts now only applies to the attach-to-transaction strategy. When standalone app start tracing is enabled (`options.experimental.enableStandaloneAppStartTracing`), the limit is bypassed.

Closes #7427

## Changes

- **`SentryAppStartTracker.swift`**: The `maxAppStartDuration` check now skips when using `StandaloneTransactionStrategy`
- **`SentryAppStartTrackerTests.swift`**:
  - Fixed `testAppLaunches_MaximumAppStartDuration_NoAppStart` to fire the display link so it actually exercises the duration check
  - Replaced `testMaxDurationExceeded_whenStandalone_shouldClearAppStartTraceId` with `testLongDuration_whenStandalone_shouldNotDropAppStart` which verifies the "App Start" transaction is captured

## How to test

- Run `make test-ios ONLY_TESTING=SentryAppStartTrackerTests`

### DIFF
--- a/Sources/Swift/Integrations/AppStartTracking/SentryAppStartTracker.swift
+++ b/Sources/Swift/Integrations/AppStartTracking/SentryAppStartTracker.swift
@@ -209,7 +209,7 @@ public final class SentryAppStartTracker: NSObject, SentryFramesTrackerListener 
                 appStartTimestamp = sysctl.processStartTimestamp
             }
 
-            if !(self.reportingStrategy is StandaloneTransactionStrategy) && appStartDuration >= Self.maxAppStartDuration {
+            if !self.reportingStrategy.shouldSkipMaxAppStartDurationLimit() && appStartDuration >= Self.maxAppStartDuration {
                 SentrySDKLog.info("The app start exceeded the max duration of \(Self.maxAppStartDuration) seconds. Not measuring app start.")
                 return
             }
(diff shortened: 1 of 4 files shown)


Sources/Swift/Integrations/AppStartTracking/SentryAppStartTracker.swift
```diff
@@ -208,8 +209,7 @@ public final class SentryAppStartTracker: NSObject, SentryFramesTrackerListener
                 appStartTimestamp = sysctl.processStartTimestamp
             }
 
-            // Safety check to not report app starts that are completely off.
-            if appStartDuration >= Self.maxAppStartDuration {
+            if !(self.reportingStrategy is StandaloneTransactionStrategy) && appStartDuration >= Self.maxAppStartDuration {
```
philprime: `l`: WDYT about not using a type assertion here, but instead a boolean value of the reporting strategy which returns `true`?

Something like:
```
if !self.reportingStrategy.shouldSkipMaxAppStartDurationLimit && appStartDuration >= Self.maxAppStartDuration {
```

This could make it more readable on why this is an exception

philprime (approved):
LGTM with one comment

Sources/Swift/Integrations/AppStartTracking/SentryAppStartTracker.swift
itaybre: Sounds good

PR #7951, opened 2026-05-27
fix: handle Apple & Homebrew LLVM vended clang-format versions in tooling scripts

NinjaLikesCheez:
## :scroll: Description

Adds support for Apple & Homebrew LLVM versions of clang-format in the tooling version scripts.

Homebrew LLVM and Apple vended versions of clang-format have an additional prefix:

```sh
➜ /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang-format --version
Apple clang-format version 21.0.0 (clang-2100.1.1.101)

➜ /opt/homebrew/opt/llvm/bin/clang-format --version
Homebrew clang-format version 22.1.6
```

This differs to the often installed Homebrew version of `clang-format`:

```sh
➜ /opt/homebrew/bin/clang-format --version
clang-format version 22.1.6
```

This caused the tooling scripts (which selected the 3rd item in the output) to select the string 'version' instead of the actual version.

I did consider iterating through each item in the output and looking for something that came after the word "version" or looked like a version number, but decided instead to be more specific on the variants of tooling that I know exist to keep the code as small and as simple as possible.

Ran with all three variants of clang-format and ensured it chose the right output.

#skip-changelog

### DIFF
--- a/scripts/check-tooling-versions.sh
+++ b/scripts/check-tooling-versions.sh
@@ -10,7 +10,14 @@ cd "${0%/*}"
 # -- Begin Script --
 
 REMOTE_CLANG_FORMAT_VERSION=$(cat .clang-format-version)
-LOCAL_CLANG_FORMAT_VERSION=$(clang-format --version | awk '{print $3}')
+CLANG_FORMAT_VERSION_STR=$(clang-format --version)
+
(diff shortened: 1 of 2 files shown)


philprime (approved):
LGTM

PR #7954, opened 2026-05-27
fix(objc): use conditional import in headers

philprime:
This pull request updates the import logic across all public header files in the `SentryObjC` target to improve compatibility with our different integration setups.

At this point we are building and testing the library both using SPM and Xcode Targets, which are both using either quotation marks or angle bracket imports, hence we need to use the approach of conditional imports.

Note: this approach is not ideal in terms of maintainability and can be refactored in the future, for now it's the quickest way of getting to a working solution.

#skip-changelog

### DIFF
--- a/Sources/SentryObjC/Public/SentryObjC.h
+++ b/Sources/SentryObjC/Public/SentryObjC.h
@@ -1,83 +1,177 @@
 #import <Foundation/Foundation.h>
 
 // Platform detection and shared macros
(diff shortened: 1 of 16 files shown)


itaybre:
> #skip-changelog

`m`: An explanation of the change would be appreciated for future maintainers

philprime:
@itaybre this stacked PR is just for me to see if CI passes, no need for review.

But I see your point and I'll add docs.

PR #7956, opened 2026-05-27
WIP fix SentryObjCTests

noahsmartin:
The biggest changes here are:
- Switch all angle brackets to regular quoted imports
- Add SentryObjCTest target to Package.swift
- Remove the logic that deleted SentryObjC before running swift package tests

With this change the SwiftObjC tests run at the same time as the Distribution tests in CI. No extra changes are needed to CI pipelines to get them to run.


philprime:
Closing this as the ObjC PR is merged and it's easier if we restart the Xcode -> SPM migration from main

PR #7961, opened 2026-05-28
ci: fix RN Sentry podspec rewrite

philprime:
Fix the React Native cross-platform workflow RNSentry.podspec rewrite to strip both the old literal Sentry version constraint and the current sentry_cocoa_version variable form.

This keeps the local sentry-cocoa path injection from conflicting with RNSentry pinned CocoaPods dependency when sentry-react-native trails this branch Sentry version.

Stacked on #7918.

#skip-changelog


philprime:
Duplicate of https://github.com/getsentry/sentry-cocoa/pull/7959

PR #7967, opened 2026-05-29
chore: Track SentryObjC adoption

philprime:
Mark Sentry events and envelope headers produced through `SentryObjCSDK`
with the SDK name `sentry.cocoa.objc`.

This lets Sentry measure adoption of the ObjC wrapper while reusing the
existing SDK Interface `sdk.name` field. The marker only applies when the
current SDK name is the base Cocoa name and restores `sentry.cocoa` when the
ObjC wrapper closes after setting it.

#skip-changelog


itaybre (approved):
LGTM

PR #7977, opened 2026-06-01
docs: Document jq and yq usage

philprime:
Document the repo-wide expectation that JSON tasks use jq and YAML tasks use yq instead of shelling out to node or python for parsing.

Add jq and yq to the Brewfile so the documented tools are part of the local development toolchain.

#skip-changelog


NinjaLikesCheez (approved):

itaybre (approved):
LGTM

PR #7996, opened 2026-06-02
feat(objc): Add format string logging with message template attributes

### LINKED ISSUE #7557: Logs: support message template attributes from Objective-C
### Problem Statement

Message template attributes are currently only supported from Swift. We’d like to request them to be also supported from Objective-C.

(They are even supported with the native SDK, so why not with Objective-C?)

### Solution Brainstorm

I imagine this could be offered with an API like this:
```objc
[logger debugWithFormat:@"A %@ log message", formatted];
```

With extra attributes it would probably have to look like this to keep the variadic parameter last:
```objc
[logger debugWithAttributes:attributes format:@"A %@ log message", formatted];
```

_No response_

philprime:
Adds format-string logging methods to `SentryObjCLogger` for Objective-C consumers, with automatic extraction of structured message template attributes for Sentry's log grouping and search.

```objc
[logger infoWithFormat:@"User %@ processed %d items", userName, count];
// body:     "User Alice processed 42 items"
// template: "User %@ processed %d items"
// parameter.0: "Alice"  (string)
// parameter.1: 42       (integer)
```

### Decisions

- **Raw format string as template** — the original format string is sent as `sentry.message.template` rather than converting to `{N}` placeholders. Both formats are accepted by the server; this avoids reconstructing the template and reduces parser complexity.
- **va_list parsing is unavoidable** — C variadic functions provide no runtime type information. The format string is the only source of type metadata, so the parser must walk each specifier to call `va_arg` with the correct type. There is no Apple API to decompose `NSString` format strings.
- **Positional specifiers (`%1$@`) are not supported** — `va_list` is forward-only; arbitrary-order access would require a two-pass type map. This is documented in the code and header.
- **ObjC category on a Swift class** — `va_list` processing and `NS_FORMAT_FUNCTION` cannot be implemented in Swift, so this uses an ObjC category (`FormatString`) that forwards through the existing `attributes:` dict.

### Coverage

All specifiers from Apple's [String Format Specifiers](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Strings/Articles/formatSpecifiers.html) are handled: `%@`, `%d`/`%i`, `%u`/`%x`/`%X`/`%o`, `%f`/`%F`/`%e`/`%E`/`%g`/`%G`/`%a`/`%A`, `%c`/`%C`, `%s`/`%S`, `%p`, `%%`, legacy `%D`/`%O`/`%U`, and all length modifiers (`h`, `hh`, `l`, `ll`, `q`, `z`, `t`, `j`, `L`). 58 tests cover every specifier, length modifier, edge case (nil, invalid UTF-8), and dynamic width/precision.

Fixes #7557


philprime:
@sentry review

philprime:
@sentry review

Sources/SentryObjC/SentryObjCLogger.m
```diff
@@ -0,0 +1,520 @@
...
+
+// ──────────────────────────────────────────────────────────────────────────────
+// NSString format string parser
+//
+// Reference: Apple "String Format Specifiers"
+// https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Strings/Articles/formatSpecifiers.html
+//
+// NSString follows the IEEE printf specification plus %@ for Objective-C
+// objects. Positional specifiers (n$, e.g. %1$@ %2$s) are supported by
+// NSString but are NOT handled by this parserm because C's va_list is a
```
NinjaLikesCheez:
```suggestion
// NSString but are NOT handled by this parser because C's va_list is a
```
`l`: typo

NinjaLikesCheez (approved):

philprime:
Thanks @NinjaLikesCheez for the approval. I'll leave this open until @denrase had a chance to review it as he built the underlying implementation of Sentry Logs, so maybe there's a catch we need to keep in mind.

CHANGELOG.md
```diff
@@ -16,6 +16,11 @@
   - Change `#import <Sentry/Sentry.h>` to `#import <SentryObjC/SentryObjC.h>`
   - Rename `Sentry`-prefixed types to `SentryObjC` (e.g., `SentrySDK` → `SentryObjCSDK`, `SentryOptions` → `SentryObjCOptions`).
 - `SentrySDK.extendAppLaunch()` now returns the extended app launch span, allowing users to add child spans for granular breakdown of the app start period (#7985)
+- Add format-string logging to `SentryObjCLogger` with automatic message template extraction (#7996)
+
+  ```objc
+  [SentryObjCSDK.logger infoWithFormat:@"User %@ processed %d items", userName, count];
+  ```
```
itaybre: `l`: I believe this will be a merge conflict and get into v9.16.1

Sources/SentryObjC/SentryObjCLogger.m
```diff
@@ -0,0 +1,526 @@
...
+- (void)traceWithAttributes:(NSDictionary<NSString *, id> *)attributes
+                     format:(NSString *)format, ...
+{
+    va_list args;
+    va_start(args, format);
+    NSString *body;
+    NSMutableDictionary *attrs;
+    SentryObjCParseFormatString(format, args, &body, &attrs);
+    va_end(args);
+    NSMutableDictionary *merged = [attributes mutableCopy];
```
itaybre, line 393: `m`: Is it expected that user attributes will be overwritten by the SDK?

itaybre (approved):
LGTM, small comments

Sources/SentryObjC/SentryObjCLogger.m
philprime: It matches the internal logic in SentryLogger where the `logAttributes` are also overriding the user-provided ones:

https://github.com/getsentry/sentry-cocoa/blob/2509eff018d61b23b2fa94a463aee5b96dacc915/Sources/Swift/Tools/SentryLogger.swift#L183-L208

Sources/SentryObjC/SentryObjCLogger.m
```diff
@@ -0,0 +1,526 @@
...
+    SentryObjCLenT, // t  — ptrdiff_t
+    SentryObjCLenJ, // j  — intmax_t / uintmax_t
+    SentryObjCLenBigL // L  — long double
+};
+
+// Parses an NSString format string, extracts typed parameter values from args,
+// and produces the formatted body + an attributes dictionary containing
+// sentry.message.template and sentry.message.parameter.N entries.
+static void
+SentryObjCParseFormatString(NSString *format, va_list args, NSString *__autoreleasing *outBody,
```
denrase, line 77: Just something my coding gent uncovered, might be worth checking if this is indeed an issue:

The parser can desynchronize the va_list when it encounters an unsupported specifier that still consumes an argument.

Example: %n is accepted by NSString formatting and consumes an int *, but the parser falls into default and does not consume anything. The next supported specifier then reads the wrong argument:

```objc
int charsWritten = 0;
[SentryObjCSDK.logger debugWithFormat:@"Count%n then %d", &charsWritten, 42];

XCTAssertEqualObjects(self.capturedLog.body, @"Count then 42");
XCTAssertEqualObjects(self.capturedLog.attributes[@"sentry.message.parameter.0"].value, @42);
```    

```text
failed: ("1801092828") is not equal to ("42")
```

The body formats correctly as Count then 42, but sentry.message.parameter.0 becomes the pointer value from &written interpreted as an integer, instead of 42.

Tests/SentryObjCTests/SentryObjCLoggerTests.m
```diff
@@ -33,104 +40,690 @@ - (void)testLogger_shouldBeAccessibleFromSDK
...
+        self.capturedLog.attributes[@"sentry.message.parameter.0"].value, @"world");
+}
+
+- (void)testDebugWithFormat_withChar_shouldExtractStringParameter
+{
+    // -- Act --
+    [SentryObjCSDK.logger debugWithFormat:@"Char: %c", 'A'];
+
+    // -- Assert --
+    XCTAssertEqualObjects(self.capturedLog.body, @"Char: A");
```
denrase, line 360: m: We could also add the template assertion in all cases, just for completnes sake. Ignore if this is intenionally left out.

Tests/SentryObjCTests/SentryObjCLoggerTests.m
```diff
@@ -2,22 +2,29 @@
 @import XCTest;
 
 @interface SentryObjCLoggerTests : XCTestCase
```
denrase, line 4: l: Would it make sense to also add a test which test format precision behaviour/rounding? We capture the real value, but body has the formatted one.

```objc
[SentryObjCSDK.logger debugWithFormat:@"Score: %.2f", 95.555];

XCTAssertEqualObjects(self.capturedLog.body, @"Score: 95.56");
XCTAssertEqualObjects(
   self.capturedLog.attributes[@"sentry.message.parameter.0"].value,
   @95.555,
);
```

denrase (approved):
Looks good, just one edge case with unsupported specifierd. :bow

PR #8003, opened 2026-06-05
feat: use distinct span operations for app start phases

itaybre:
## Description

Standalone app start child spans now use specific operation names instead of the generic `app.start` operation. This improves observability and allows filtering by individual app start phases.

### Operation mapping (standalone spans only)

| Span description | Before | After |
|---|---|---|
| Pre Runtime Init | `app.start` | `app.start.pre_runtime_init` |
| Runtime Init to Pre Main Initializers | `app.start` | `app.start.runtime_init` |
| UIKit Init | `app.start` | `app.start.uikit_init` |
| Application Init | `app.start` | `app.start.application_init` |
| Extended App Start | `app.start` | `app.start.extended_app_start` |

Spotlight confirmation of the rename
image

Non-standalone spans retain their existing `app.start` operation unchanged.


Sources/Swift/Integrations/AppStartTracking/SentryExtendedAppLaunchManager.swift
```diff
@@ -3,6 +3,10 @@
 #if (os(iOS) || os(tvOS) || os(visionOS)) && !SENTRY_NO_UI_FRAMEWORK
 
 final class SentryExtendedAppLaunchManager {
+    
+    enum Constants {
+        static let extendedOperation = "\(SentrySpanOperationAppStart).extended_app_start"
```
philprime, line 8: `l`: It might be worth the effort to have an explicit tests verifying this constant so it's not changed by accident. The PR already has transitive tests, but they could change eventually

philprime (approved):
LGTM

PR #8010, opened 2026-06-07
test: Fix SentryDistributionTests when running with 'swift test'

noahsmartin:
Updated this test to work with `swift test` in preparation for some questions about swift testing at WWDC

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -135,7 +135,7 @@ jobs:
           PLATFORM: "iOS"
         run: ./scripts/ci-ensure-runtime-loaded.sh --os-version "$OS_VERSION" --platform "$PLATFORM"
       - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace
-      - run: set -o pipefail && NSUnbufferedIO=YES SKIP_BINARIES=1 xcodebuild test -scheme Sentry-Package -sdk iphonesimulator -destination 'platform=iOS Simulator,OS=18.5,name=iPhone 16 Pro' | tee raw-test-output-distribution.log | xcbeautify --preserve-unbeautified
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -135,7 +135,7 @@ jobs:
           PLATFORM: "iOS"
         run: ./scripts/ci-ensure-runtime-loaded.sh --os-version "$OS_VERSION" --platform "$PLATFORM"
       - run: rm -r Sentry.xcodeproj && rm -r Sentry.xcworkspace
-      - run: set -o pipefail && NSUnbufferedIO=YES SKIP_BINARIES=1 xcodebuild test -scheme Sentry-Package -sdk iphonesimulator -destination 'platform=iOS Simulator,OS=18.5,name=iPhone 16 Pro' | tee raw-test-output-distribution.log | xcbeautify --preserve-unbeautified
```
noahsmartin, line 138: This env var was no longer used

Sources/SentryDistributionTests/BinaryParserTests.swift
```diff
@@ -1,7 +1,14 @@
...
 
-@Test func testBinaryParserGetsCorrectUUID() throws {
+@Test("iOS Feature Test", .enabled(if: {
+    #if os(iOS)
+    return true
+    #else
+    return false
+    #endif
+}()))
+func testBinaryParserGetsCorrectUUID() throws {
```
NinjaLikesCheez, line 13: `l`: It seems this test is breaking with `swift test` because `BinaryParser.getMainBinaryUUID(_:)` relies on the Bundle API - can you add a `.bug` or something so that it's obvious why this is restricted to iOS. For example, it 'works' on macOS in Xcode because it grabs the `xctest` binary bundle.

NinjaLikesCheez (commented):
LGTM, one comment

NinjaLikesCheez (approved):

PR #7873, opened 2026-05-05
feat(feedback): Add managed user feedback form presentation APIs

denrase:
## :scroll: Description

Adds new iOS managed user feedback presentation APIs so apps can present the feedback form through standard UIKit or SwiftUI presentation flows instead of relying only on the floating widget.

### New APIs

```swift
// UIKit
let form = SentrySDK.FeedbackForm()
present(form, animated: true)
```

```swift
// UIKit with image attachment
let form = SentrySDK.FeedbackForm(image: image)
present(form, animated: true)
```

```swift
// Convenience API: SDK chooses a suitable presenter
SentrySDK.feedback.show()
SentrySDK.feedback.show(image: image)
```

```swift
// SwiftUI
.sheet(isPresented: $showFeedback) {
    SentrySDK.FeedbackFormView()
}
```

```swift
// SwiftUI with image attachment
.sheet(isPresented: $showFeedback) {
    SentrySDK.FeedbackFormView(image: image)
}
```

`SentrySDK.FeedbackForm` is an alias for `SentryUserFeedbackFormController`, and `SentrySDK.FeedbackFormView` is an alias for `SentryUserFeedbackFormView`.

The public form controller handles submission, validation errors, cancellation, keyboard layout, lifecycle hooks, and dismissal internally, so customers can instantiate and present it wherever their app has the right presentation context. This solves the main multi-window issue by letting apps choose the presenting view controller directly.

The convenience API reuses the SDK's active-window/relevant-view-controller lookup to find a presenter in a foreground-active scene. It avoids presenting when no presenter is available, when the presenter is already presenting, or when a feedback form is already displayed.

The integration driver now tracks the active form, hides the feedback widget while the form is presented, and restores the widget after the form closes.

### Configuration

The new APIs use the global configuration from `SentryOptions.configureUserFeedback`. A follow-up PR stacked on this one will introduce optional config that can be passed to individual presntations, overriding the global one.

## :bulb: Motivation and Context

Closes #7826

## :green_heart: How did you test it?

- Added/updated unit tests for form lifecycle callbacks, submission callbacks, global configuration reuse, duplicate presentation prevention, widget hide/restore behavior, and integration initialization.
- Added UIKit and SwiftUI sample app screens for direct form presentation, convenience presentation, and widget show/hide.
- Regenerated `sdk_api.json` for the new public APIs.

## :pencil: Checklist

You have to check all boxes before merging:

- [x] I added tests to verify the changes.
[...]

PR #8018, opened 2026-06-08
feat: Add per-form feedback configuration

denrase:
## :scroll: Description

This adds per-presentation configuration for user feedback forms across the UIKit controller, convenience API, SwiftUI form view, and SwiftUI modifier. The presentation-specific configuration is copied from the global feedback configuration while ignoring widget-only settings, so individual forms can customize labels, themes, tags, callbacks, screenshots, and animations without mutating global state.

```swift
SentrySDK.feedback.show { config in
    config.configureForm = { form in
        form.formTitle = "Report a Bug"
        form.submitButtonLabel = "Send Report"
    }
    config.tags = ["screen": "settings"]
}
```

Relates to #7826
Relates to #7873

Unit tests

## :pencil: Checklist

- [ ] If I added a new public API, I also added it to the [SentryObjC wrapper](../develop-docs/SENTRY-OBJC.md).


Sources/Swift/Integrations/UserFeedback/SentryFeedbackAPI.swift
```diff
@@ -22,16 +22,23 @@ import UIKit
...
+    /// - Parameters:
+    ///   - screenshot: An optional screenshot to attach to the feedback form.
+    ///   - configure: A closure to customize this feedback form presentation.
     /// - Important: Call this method from the main thread.
     /// - warning: This is an experimental feature and may still have bugs.
     @available(iOSApplicationExtension, unavailable)
-    public func show(screenshot: UIImage? = nil) {
+    public func show(
+        screenshot: UIImage? = nil,
+        configure: SentryUserFeedbackConfigurationCallback? = nil
```
philprime, line 40: `m`: Passing in `nil` for the configuration callback might not make sense, because by default it should just be `{ config in return config }` or just `{ $0 }`. It might be necessary for ObjC compatibility though

philprime (approved):
LGTM

Sources/Swift/Integrations/UserFeedback/SentryFeedbackAPI.swift
denrase: In the default case (when nil) we fall back to the global config, in this case we do not need to make a copy of the options.

PR #8025, opened 2026-06-09
docs: add V3 app hang tracking design spec

philprime:
## Summary

- Adds `develop-docs/APP-HANG-TRACKING.md` — comprehensive design spec for V3 app hang tracking based on CFRunLoopObserver
- Covers: core mechanism, tiered detection with stack trace sampling, signal categories, client-side sample aggregation, fatal hang persistence with duration recovery, event sending mechanics, option structure (`options.experimental.appHangs.*`), mutual exclusivity with V1/V2, platform behavior (all platforms supported, watchOS timing-only), and ANR Tracker V3 wrapper architecture
- Includes comparison matrix (V1 vs V2 vs V3), annotated open GitHub issues, and documented V1/V2 for reference

## Motivation

V1 and V2 have fundamental limitations: polling overhead when idle, coarse detection granularity, single-snapshot stack traces that may miss the root cause, and no fatal hang duration recovery. V3 replaces both with an event-driven CFRunLoopObserver approach that provides precise timing, sampled stack traces with confidence metrics, and incremental persistence for fatal hang recovery.

This is a living design document for discussion and iteration before implementation begins.

#skip-changelog

### DIFF
--- a/develop-docs/APP-HANG-TRACKING.md
+++ b/develop-docs/APP-HANG-TRACKING.md
@@ -39,6 +39,12 @@ Last updated: 2026-06-09
 
 ---
 
+## TL;DR — V3 at a Glance
+
+V3 replaces V1/V2 with a single **event-driven** hang detector built on `CFRunLoopObserver` + `DispatchSemaphore`. It has **zero overhead when the app is healthy** (no polling threads). A background queue waits on the semaphore with escalating timeouts: hitch noted at ~25ms, stack trace sampling starts at ~250ms, app hang reported at ~2s. Only **fully-blocking hangs** (one run loop iteration > threshold) are reported — V2's "non-fully-blocking" category is dropped. Stack traces are **sampled every ~250ms** by suspending only the main thread (not all threads), then aggregated client-side into a representative trace with a confidence metric. Fatal hangs are **incrementally persisted** to disk so duration is recoverable on next launch (±250ms). Works on **all Apple platforms**; watchOS gets timing-only detection (no stack traces). Opt-in via `options.experimental.appHangs.enableV3`.
+
+---
+
 ## V3 Design Spec
 
 V3 replaces both V1 and V2 with a unified, event-driven detection mechanism based on CFRunLoop observers. The prototype lives in `Sources/Swift/HangTracker.swift`.
--- a/develop-docs/APP-HANG-TRACKING.md
+++ b/develop-docs/APP-HANG-TRACKING.md
@@ -1,0 +1,899 @@
+# App Hang Tracking — Design Spec
+
+Status: **Living document**
+Last updated: 2026-06-09
+
+---
+
+## Table of Contents
+
+- [Comparison Matrix](#comparison-matrix)
+- [V3 Design Spec](#v3-design-spec)
+- [Open Issues and Feature Requests](#open-issues-and-feature-requests)
+- [Previous Solutions](#previous-solutions)
+  - [Architecture Overview](#architecture-overview)
+  - [V1: Main-Queue Dispatch Probe](#v1-main-queue-dispatch-probe)
+  - [V2: Frame-Delay Signal](#v2-frame-delay-signal)
+
+---
+
+## Comparison Matrix
+
+| Aspect                 | V1                              | V2                                    | V3                                                                            |
+| ---------------------- | ------------------------------- | ------------------------------------- | ----------------------------------------------------------------------------- |
+| Signal                 | Main-queue dispatch probe       | Frame delay query                     | RunLoop afterWaiting → beforeWaiting                                          |
+| Architecture           | Polling thread + sleep          | Polling thread + sleep                | CFRunLoopObserver + semaphore                                                 |
+| Granularity            | ~seconds (configurable timeout) | ~seconds (configurable timeout)       | ~milliseconds (hitchThreshold)                                                |
+| Hang types             | Unknown only                    | Fully-blocking / non-fully-blocking   | Fully-blocking only (non-fully-blocking dropped)                              |
+| Duration measurement   | No                              | Yes (min/max range)                   | Yes (precise, continuous)                                                     |
+| Stack trace capture    | At detection (~2s in)           | At detection (~2s in)                 | Sampled every ~250ms from samplingThreshold onward                            |
+| Stack trace quality    | Single snapshot, may miss cause | Single snapshot, may miss cause       | Multiple samples with confidence metric                                       |
+| Overhead when healthy  | Dispatch to main queue per tick | FramesTracker query per tick          | None (event-driven)                                                           |
+| Dependencies           | DispatchQueue, Thread           | DispatchQueue, Thread, FramesTracker  | CFRunLoop, DispatchQueue                                                      |
+| Platform support       | All                             | iOS, tvOS, visionOS (UIKit)           | All (stack traces on all except watchOS)                                      |
+| Suspension handling    | Wall-clock delta check          | Wall-clock delta check                | Wall-clock delta check per semaphore timeout                                  |
+| Background filtering   | Late (after threshold)          | Per-tick foreground check             | Per-sample check via SentryCrash state (safe during hang)                     |
+| Cooldown               | No                              | Yes (timeoutInterval after last hang) | Yes (configurable cooldownDuration, default = appHangThreshold)               |
+| Fatal hang persistence | Write once at detection         | Write once at detection               | Incremental writes — updated on each sample, duration recoverable on relaunch |
+| Fatal hang duration    | Unknown                         | Unknown                               | Estimated from last_sample_time − hang_start_time                             |
+
+---
+
+## V3 Design Spec
+
+V3 replaces both V1 and V2 with a unified, event-driven detection mechanism based on CFRunLoop observers. The prototype lives in `Sources/Swift/HangTracker.swift`.
+
+### Core Mechanism: Run Loop Observer + Semaphore
+
+Uses `CFRunLoopObserver` to directly measure main run loop busy time:
+
+1. Register an observer for `afterWaiting | beforeWaiting` on the main run loop
+2. On `afterWaiting`: record start time, create a `DispatchSemaphore`, dispatch `waitForHang` to a serial background queue
+3. Background queue waits on the semaphore with a configurable timeout
+4. On timeout: re-wait (the main thread is still busy)
+5. On `beforeWaiting`: signal the semaphore → the run loop iteration is over
+
+This approach is event-driven and has zero overhead during healthy run loop iterations. No polling threads, no dispatched probe blocks, no frame tracker queries.
+
+The `waitForHang` loop already provides periodic wake-ups on the background queue during a hang. Each timeout is an opportunity to capture a stack trace. No additional timer or sampling infrastructure is needed.
+
+### Tiered Detection and Stack Trace Sampling
+
+The hang tracker uses a tiered approach with configurable thresholds. Each semaphore timeout checks the elapsed duration and escalates through tiers:
+
+```mermaid
+sequenceDiagram
+    participant RL as Main RunLoop
+    participant BG as Background Queue
+
+    RL->>BG: afterWaiting (start timing)
+
+    Note over BG: semaphore wait = hitchThreshold (~25ms)
+    BG--xBG: timeout → note hitch started (no sampling)
+
+    loop wait hitchThreshold until elapsed ≥ samplingThreshold
+        BG--xBG: timeout → still below sampling threshold
+    end
+
+    Note over BG: elapsed ≥ samplingThreshold, switch wait to samplingInterval (~250ms)
+    BG->>RL: suspend main thread
+    BG-->>BG: capture stack trace sample #1
+    BG->>RL: resume main thread
+
+    loop every samplingInterval (~250ms)
+        Note over BG: semaphore wait = samplingInterval
+        BG->>RL: suspend main thread
+        BG-->>BG: capture stack trace sample
+        BG->>RL: resume main thread
+    end
+
+    Note over BG: elapsed ≥ appHangThreshold (~2s)
+    BG-->>BG: classify as app hang, build event with all samples
+
+    RL->>BG: beforeWaiting (signal semaphore)
+    Note over BG: hang is over, report final duration
+```
+
+**Tier 1, Hitch detection** (`hitchThreshold`, default ~25ms):
+
+First timeout fires. We note that the run loop iteration is slow. No stack trace is captured at this point, only a timestamp. This is the point where we know something is taking longer than a single frame. Cost: near zero.
+
+**Tier 2, Sampling begins** (`samplingThreshold`, default ~250ms):
+
+If still hanging, capture the first main thread stack trace. At 250ms into a fully-blocking hang, the main thread is very likely still in the blocking call. This early sample is the most valuable for root-cause attribution.
+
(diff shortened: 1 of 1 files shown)


develop-docs/APP-HANG-TRACKING.md
```diff
@@ -0,0 +1,905 @@
...
+| Fatal hang duration    | Unknown                         | Unknown                               | Estimated from last_sample_time − hang_start_time                             |
+
+---
+
+## TL;DR — V3 at a Glance
+
+V3 replaces V1/V2 with a single **event-driven** hang detector built on `CFRunLoopObserver` + `DispatchSemaphore`. It has **zero overhead when the app is healthy** (no polling threads). A background queue waits on the semaphore with escalating timeouts: hitch noted at ~25ms, stack trace sampling starts at ~250ms, app hang reported at ~2s. Only **fully-blocking hangs** (one run loop iteration > threshold) are reported — V2's "non-fully-blocking" category is dropped. Stack traces are **sampled every ~250ms** by suspending only the main thread (not all threads), then aggregated client-side into a representative trace with a confidence metric. Fatal hangs are **incrementally persisted** to disk so duration is recoverable on next launch (±250ms). Works on **all Apple platforms**; watchOS gets timing-only detection (no stack traces). Opt-in via `options.experimental.appHangs.enableV3`.
```
itaybre: `l`: nitpicking, but I wouldn't say 0 overhead: `beforeWaiting` executes in the main thread, every loop creates a new semaphore + dispatch to the background queue

itaybre: I think it would be good to verify this on a thread heavy app, to confirm there are no issues

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -0,0 +1,905 @@
...
+
+**Filtering logic in the `waitForHang` loop:**
+
+1. **Hang starts while backgrounded:** On the first semaphore timeout that would cross `samplingThreshold`, check `isApplicationInForeground`. If false, skip sampling and continue waiting. Re-check on each subsequent timeout. If the app returns to foreground mid-hang, begin sampling from that point.
+
+2. **App backgrounds mid-hang:** On each sampling timeout, check `isApplicationInForeground`. If the app has gone to background since the last sample, stop sampling. If the app returns to foreground, resume sampling. The hang duration reported in the event reflects wall-clock time of the full run loop iteration (afterWaiting to beforeWaiting), regardless of background transitions. Background time is **not** subtracted from the reported duration.
```
itaybre, line 153: `l`: could this be an option?
I am also wondering if the app can be suspended mid RunLoop, thus "causing" app hang events when coming from background

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -0,0 +1,905 @@
...
+| `experimental.appHangs.maxSamples`        | `40`                                          | Maximum stack trace samples retained per hang (ring buffer).                    |
+| `experimental.appHangs.cooldownDuration`  | `appHangThreshold` (e.g., 2s)                 | Minimum time after reporting a hang before a new hang event.                    |
+
+**V2 migration:**
+
+- `enableAppHangTracking`: **keep as-is.** Controls whether any hang tracking is enabled (V1/V2/V3). V3 only activates when both `enableAppHangTracking` and `experimental.appHangs.enableV3` are true.
+- `appHangTimeoutInterval`: **keep as-is.** When V3 is enabled, its value is used as the default for `experimental.appHangs.appHangThreshold` if the latter is not explicitly set. Explicit V3 setting takes precedence.
+- `enableReportNonFullyBlockingAppHangs`: **deprecate.** V3 does not have a non-fully-blocking category. The option becomes a no-op when V3 is active. Mark `@available(*, deprecated, message: "V3 hang tracking only reports fully-blocking hangs.")`. Remove in the next major version.
```
itaybre: `l`: next major: v10?

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -0,0 +1,905 @@
...
+} else {
+    → use SentryANRTrackerV1 (existing)
+}
+```
+
+When V3 is enabled, it wins unconditionally on all platforms. The V1/V2 trackers are not instantiated.
+
+### Stack Trace Capture Strategy
+
+Stack traces are captured on the background queue during semaphore timeouts. V3 suspends only the main thread (not all threads), unlike V2 which suspends all threads (up to 70) via `sentrycrashmc_suspendEnvironment`.
```
itaybre, line 273: `m`: Since we are moving out of `SentryCrash`, we might need to fully implement this in Sentry or plan how to use KSCrash for this

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -0,0 +1,905 @@
...
+- `confidence` is the dominance ratio (representative_count / total retained samples in the ring buffer, not samples_total). When no samples have been evicted, `samples_total` equals the retained count and the two denominators are identical.
+- `sample_groups` contains top-K frame groups with counts (from retained samples only), allowing the backend to show a breakdown (e.g., "87% in `syncWrite`, 13% in `decode`")
+
+The event contains a **single** stack trace: the representative one selected by the aggregation algorithm. This stack trace is attached as the event's exception stack trace, identical in format to V2. Raw samples are not sent. The `sample_groups` metadata in `mechanism.data` provides attribution context (which top-frame addresses were seen and how often) without transmitting full stack traces for each sample.
+
+**Backend dependency:** The `sample_groups`, `confidence`, `samples_total`, and `representative_count` fields in `mechanism.data` are new payload additions. The Sentry backend (relay, event processing, issue grouping, UI) needs to support these fields for the full V3 experience. Until the backend is updated:
```
itaybre: `h`: Also frontend changes since this data would need to be exposed.
Also, what's the plan to present this to the user?

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -0,0 +1,905 @@
...
+        Next->>Disk: readAppHangEvent()
+        Next->>Disk: deleteAppHangEvent()
+        Note over Next: upgrade to fatal, compute duration, send
+    end
+```
+
+**When to write:**
+
+1. **First write.** When elapsed time crosses `appHangThreshold` (e.g., 2s). The event includes the scope, context, all stack trace samples collected since `samplingThreshold`, and aggregation metadata.
+2. **Subsequent writes.** After each new stack trace sample (every `samplingInterval`, ~250ms). Each write overwrites the previous file with an updated event containing all samples and a refreshed `lastSampleTime`.
```
itaybre: `l`: Atomic writes, right? We don't want partial files here

itaybre (approved):
LGTM, some comments here and there

develop-docs/APP-HANG-TRACKING.md
philprime: I agree, the wording is too strong and we need to eventually benchmark the impact

develop-docs/APP-HANG-TRACKING.md
philprime: Updated to say "Remove in v10." explicitly.

develop-docs/APP-HANG-TRACKING.md
philprime: Added a "SentryCrash dependency" section under Stack Trace Capture Strategy. V3's single-thread suspend path uses standard Mach kernel APIs (`thread_suspend`/`thread_get_state`/`vm_read_overwrite`/`thread_resume`) and should be implemented directly in the Sentry layer without depending on SentryCrash.

develop-docs/APP-HANG-TRACKING.md
philprime: Investigated relay and the Sentry frontend: `mechanism.data` is a free-form object in relay (max depth 5, max bytes 2048) — arbitrary keys pass through without validation. The frontend already renders scalar `mechanism.data` fields as pills in the exception mechanism display. No backend or frontend changes are needed for V3. Updated the spec to reflect this.

develop-docs/APP-HANG-TRACKING.md
philprime: Good call — added explicit requirement for atomic writes (write to temp file, then `rename`) in the Disk I/O Considerations section.

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -0,0 +1,899 @@
...
+3. **On hang end.** Delete the file and send the event through the normal capture path with the final precise duration.
```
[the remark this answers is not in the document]
philprime: Fixed: on watchOS (no stack trace capture), the `waitForHang` loop still fires semaphore timeouts at `samplingInterval` and updates `last_sample_time` on each timeout. The persisted event is overwritten even without stack traces, so fatal hang duration recovery works on all platforms.

develop-docs/APP-HANG-TRACKING.md
philprime: Added a note clarifying that mid-RunLoop suspension is handled by the Suspension Detection mechanism (discards the hang if semaphore wait took >3x expected). Duration subtraction could be made configurable in the future but is not planned for V3 initial release.

naftaly:
Just butting in here because the spec read was amazing. If you're interested in saving yourselves some cycles, you might want to check out the way KSCrash has it setup, and either participate in improving it or just using it outright. I think this is going out in 2.6 which is in beta now. But regardless, the spec looks great to me :)

https://github.com/kstenerud/KSCrash/blob/develop/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -907,3 +910,56 @@ The main thread is **guaranteed to still be hanging** during capture because all
...
+
+The timer is created and destroyed on every run loop iteration (created on `afterWaiting`, invalidated on `beforeWaiting`).
+
+### Key Differences
+
+| Aspect               | KSCrash Watchdog                                            | V3                                                            |
+| -------------------- | ----------------------------------------------------------- | ------------------------------------------------------------- |
+| Escalation mechanism | `CFRunLoopTimer` on a watchdog `CFRunLoop`                  | `DispatchSemaphore` timeout loop on a serial background queue |
+| Threshold            | Single (250ms)                                              | Tiered (25ms hitch → 250ms sampling → 2s hang)                |
+| Stack trace capture  | Once at detection, suspends all threads                     | Sampled every 250ms, main thread only                         |
```
naftaly, line 936: fwiw, the TimeProfiler does the sampling, the watchdog only fires that the issue exist and reports it. We chose this method as they are distinct from each other and can easily be joined later on remotely as needed.

See [KSCrash+Hang.h](https://github.com/kstenerud/KSCrash/blob/develop/Sources/KSCrashRecording/include/KSCrash%2BHang.h)

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -907,3 +910,56 @@ The main thread is **guaranteed to still be hanging** during capture because all
...
+| Persistence          | Full JSON report once + 24-byte mmap'd binary sidecar       | Full JSON overwrite per sample                                |
+| Thread safety        | `os_unfair_lock` + `_Atomic uint64_t` for `enterTime`       | Observers lock + serial background queue                      |
+| Suspension detection | Not implemented                                             | Wall-clock delta check (`actualWait > expectedWait * 3.0`)    |
+| Timer lifecycle      | Created per `afterWaiting`, invalidated per `beforeWaiting` | Semaphore created per run loop iteration                      |
+
+### Timer vs Semaphore Decision
```
naftaly, line 942: Using a semaphore was considered for KSCrash. We chose the timer instead as it allows the OS to take a much larger stand as to how things should work vs. forcing it into our way with a semaphore. Also, there's the whole semaphores don't play well with a lot of the concurrency systems (priority inversion, etc, ...).

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -907,3 +910,56 @@ The main thread is **guaranteed to still be hanging** during capture because all
...
+| -------------------- | ----------------------------------------------------------- | ------------------------------------------------------------- |
+| Escalation mechanism | `CFRunLoopTimer` on a watchdog `CFRunLoop`                  | `DispatchSemaphore` timeout loop on a serial background queue |
+| Threshold            | Single (250ms)                                              | Tiered (25ms hitch → 250ms sampling → 2s hang)                |
+| Stack trace capture  | Once at detection, suspends all threads                     | Sampled every 250ms, main thread only                         |
+| Persistence          | Full JSON report once + 24-byte mmap'd binary sidecar       | Full JSON overwrite per sample                                |
+| Thread safety        | `os_unfair_lock` + `_Atomic uint64_t` for `enterTime`       | Observers lock + serial background queue                      |
+| Suspension detection | Not implemented                                             | Wall-clock delta check (`actualWait > expectedWait * 3.0`)    |
```
naftaly, line 939: KSCrash prefers to leave this up to the user of the framework as they may want to deal with this kind of suspension in a different way.

develop-docs/APP-HANG-TRACKING.md
```diff
@@ -907,3 +910,56 @@ The main thread is **guaranteed to still be hanging** during capture because all
...
+- **`DispatchSemaphore`** wait is a direct kernel wait (`semaphore_timedwait` Mach trap). Lighter per-wakeup than a run loop iteration, but the OS cannot coalesce it with other timers — each timeout is an independent kernel wakeup.
+
+In practice, the difference is negligible. Both wake a thread every ~250ms during a hang, do a small amount of work, then go back to sleep. The dominant cost is the wakeup itself, not whether it came from a timer or semaphore timeout.
+
+The semaphore has a structural advantage for V3: tiered escalation (25ms → 250ms) maps naturally to changing the wait timeout. With a timer, changing the interval requires invalidating and recreating it. V3 retains the semaphore approach.
+
+### Stack Trace Capture: All Threads vs Main Thread Only
+
+KSCrash suspends **all threads** via `ksmc_suspendEnvironment()` (`task_threads()` + `thread_suspend()` on each) and captures stack traces for every thread. V3 suspends **only the main thread** via a single `thread_suspend()` call and captures only the main thread's stack trace.
```
naftaly, line 955: We could definitely document this better. The goal was to have a regular "termination" report for when a hang starts. This gives us a few things:
- A report ready if this turns into an actual watchdog timeout termination.
- All threads captured make it much easier to understand a hang from a diagnostics POV.
- If chosen, a TimeProfile can also be run to get only the main thread at intervals in order to view the hang as some sort of graph. See this [profile](https://exitcodeclub.com/report/309124522fa14c7c8931) for example.

naftaly (commented):
Amazing, thanks so much for doing the comparison. Please reach out of we can be of any help understanding some choices we may have not documented well or just want to talk about it.

develop-docs/APP-HANG-TRACKING.md
philprime: @noahsmartin I believe you advocated for using a semaphore in our case. Did we consider challenges in concurrency systems?

philprime:
Thank you @naftaly for your feedback! We'll be in touch to see how we can collaborate on this.

philprime:
Closing this in favor of our [internal documentation](https://linear.app/getsentry/project/app-hang-detection-using-run-loop-observers-cocoa-6301bab5fc09/activity#project-update-289280d6)

PR #8029, opened 2026-06-09
docs: expand README install section per #7924

tsushanth:
### Description

Closes #7924.

The README's existing Installation section was a single sentence assuming the reader already knows how to wire SPM or drag in an XCFramework. This PR adds:

- An anchor link to the canonical install docs at \`docs.sentry.io/platforms/apple/install/\`
- Step-by-step **SPM (Xcode UI)** instructions
- A **\`Package.swift\`** snippet for SPM-managed projects
- A short **XCFrameworks** section pointing at the Releases page
- A link to the CocoaPods migration guide inside the existing deprecation warning

### Motivation

Newcomers integrating Sentry shouldn't have to leave the README to figure out the first 30 seconds of setup.

### How tested

Rendered the README locally + verified the install doc link resolves.

\`#skip-changelog\` — docs-only change.

PR #8033, opened 2026-06-10
docs: add link to installation docs to README

philprime:
Adds a more direct link to the installation options of the SDK

Supersedes #8029 

#skip-changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,10 @@ _Bad software is everywhere, and we're tired of it. Sentry is on a mission to he
 
 # Installation
 
-SPM is the recommended way to include Sentry into your project. We also provide pre-built XCFrameworks on [our GitHub Releases page](https://github.com/getsentry/sentry-cocoa/releases).
+SPM is the recommended way to include Sentry into your project.
+We also provide pre-built XCFrameworks on [our GitHub Releases page](https://github.com/getsentry/sentry-cocoa/releases).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,7 +22,10 @@ _Bad software is everywhere, and we're tired of it. Sentry is on a mission to he
 
 # Installation
 
-SPM is the recommended way to include Sentry into your project. We also provide pre-built XCFrameworks on [our GitHub Releases page](https://github.com/getsentry/sentry-cocoa/releases).
+SPM is the recommended way to include Sentry into your project.
```
NinjaLikesCheez, line 25: `l`: on the linked doc page we say we recommend the wizard - should this be adjusted to:

```suggestion
We recommend using the Sentry Wizard to integrate Sentry into your project via SPM, for more information see our [documentation](https://docs.sentry.io/platforms/apple/install/)
```

This way the documentation link is right at the top too.

NinjaLikesCheez (approved):
LGTM, one comment for thought

README.md
philprime: I would not do that, because if we change the installation page, it would be outdated in the README again, hence why I would like to keep it simple here.
But open for better wording!

NinjaLikesCheez: Fair enough!

PR #8036, opened 2026-06-10
fix: ship dSYMs in SentryObjC-Dynamic xcframework

### LINKED ISSUE #8035: SentryObjC artifacts ship no dSYMs
SentryObjC dynamic framework artifacts ship no dSYMs
- `Sentry-Dynamic.xcframework`: ships `dSYMs/` per slice
- `SentryObjC-Dynamic.xcframework`: no dSYM bundles at all, and the binaries are unstripped

Verified in `9.16.1` and `9.17.0` releases.

**Proposed solution:** Run the same dSYM step as Sentry-Dynamic.

I'm integrating SentryObjC in the Sentry's Godot SDK where this issue surfaced:
- https://github.com/getsentry/sentry-godot/pull/755

itaybre:
## Description

`SentryObjC-Dynamic.xcframework` ships without dSYM bundles because the custom `swiftc` relink step didn't extract them, and the assembly script didn't include them.

### Root cause

The dynamic framework is built by re-linking a static library (`libSentryObjC.a`) as a dynamic library via `swiftc -emit-library`. Unlike `Sentry-Dynamic.xcframework` (which uses `xcodebuild archive` and gets dSYMs for free), this manual relink step never ran `dsymutil` to extract debug symbols.

### Changes

**Scripts:**
- `build-dynamic-framework-sentryobjc.sh` — run `dsymutil` to extract dSYMs after linking, then `strip -x` to remove debug symbols from the binary
- `assemble-xcframework-sentryobjc.sh` — detect dSYM bundles alongside frameworks and pass them via `-debug-symbols` (absolute path) to `xcodebuild -create-xcframework`

**CI workflows:**
- `build-sentryobjc-static-slices.yml` — also build the dynamic framework on the same runner so `dsymutil` can access the `.o` files from the `xcodebuild archive` step. Tar the framework + dSYM to preserve macOS framework symlinks through artifact upload
- `assemble-sentryobjc-dynamic-xcframework.yml` — download and extract pre-built dynamic framework slices instead of re-linking from static libraries

**Housekeeping:**
- `file-filters.yml` — replace stale workflow and script references with current SentryObjC build pipeline files

Fixes #8035


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Ship dSYMs in SentryObjC-Dynamic.xcframework artifacts (#8035)
```
```suggestion
- Ship dSYMs in SentryObjC-Dynamic.xcframework artifacts (#8036)
```

.github/workflows/build-sentryobjc-slices.yml
```diff
@@ -61,13 +61,50 @@ jobs:
             --sdk "$SDK" \
             --output-dir XCFrameworkBuildPath
 
-      - name: Upload slice
+      # Build the dynamic framework on the same runner so dsymutil can
+      # access the .o files from the archive step above. Running this in a
+      # separate job (the previous approach) meant the object files were gone
+      # and dsymutil produced an empty dSYM.
+      - name: Build SentryObjC dynamic framework for ${{ matrix.sdk }}
```
philprime, line 68: `m`: We should rename this workflow because now it's not only building static slices but both

philprime (approved):
LGTM, thank you for fixing this

PR #8037, opened 2026-06-11
fix(objc): box resilient value types to fix missing x86_64 symbols

### LINKED ISSUE #8034: SentryObjC dynamic package is missing several symbols in x86_64 arch
Hi, Cocoa team! 👋 
I maintain Sentry's Godot Engine SDK, and just started working on SentryObjC integration in here: 
- https://github.com/getsentry/sentry-godot/pull/755

I found an issue where not all symbols are available in `x86_64` slice in the shipped dynamic framework bundle in `9.16.1` and `9.17.0` releases.

The error surfaces upon linking:
```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_SentryObjCAttributeContent", referenced from:
       in cocoa_metric.os
  "_OBJC_CLASS_$_SentryObjCMetricValue", referenced from:
       in cocoa_metric.os
  "_OBJC_CLASS_$_SentryObjCUnit", referenced from:
       in cocoa_metric.os
```
These types are affected (maybe more): `SentryObjCAttributeContent`, `SentryObjCUnit`, `SentryObjCMetric`, `SentryObjCMetricValue`.

I didn't have a chance to look why this happens yet - documenting the issue here for posterity.

**Important detail**: Our floor is macOS `10.14` on x86_64 which happens to be the Cocoa's floor as well.

### Reproduction project 
Super-minimal repro project: [mrp-sentryobjc-x86_64-link.zip](https://github.com/user-attachments/files/28805619/mrp-sentryobjc-x86_64-link.zip)
- Xcode 26.0.1 (I don't think it will matter).
- Download, then run `repro.sh` (after inspecting its contents ofc 😉).

itaybre:
## Description

Fixes #8034 — SentryObjC dynamic framework is missing `_OBJC_CLASS_$_` symbols for 4 classes in the x86_64 architecture slice:

- `SentryObjCAttributeContent`
- `SentryObjCUnit`
- `SentryObjCMetric`
- `SentryObjCMetricValue`

## Root cause

These 4 wrapper classes in `SentryObjCCompat` store properties of pure Swift value types (`SentryAttributeContent`, `SentryUnit`, `SentryMetricValue`, `SentryMetric`) from the `SentrySwift` module.

When building with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES`, the Swift compiler treats `SentrySwift` types as **resilient** — their in-memory size is unknown at compile time. On **x86_64**, this causes the compiler to generate a "class metadata template" (`CMt`) for runtime class realization instead of emitting a static `_OBJC_CLASS_$_` linker symbol. Consumers then get undefined symbol errors at link time.

On **arm64**, the compiler emits a statically-laid-out class, so the issue only manifests on x86_64.

The other 41 working wrapper classes all wrap ObjC classes (pointer-sized, always known layout), which is why they aren't affected.

## Fix

Wrap the resilient stored properties in a `Box<T>` indirection class (new file: `Box.swift`). This makes the ivar pointer-sized (always known at compile time), so the compiler emits a static class layout on all architectures. The box adds one heap allocation per wrapper instance, which is negligible since these are already `NSObject` subclasses.

## Verification

Built `SentryObjC` for macOS Release with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` and confirmed via `nm` that all 4 `_OBJC_CLASS_$_` symbols are now present in both arm64 and x86_64 object files, with no `CMt` (class metadata template) symbols.


philprime (approved):
LGTM, but we will have to expand this to other types as well when we start to refactor the underlying SDK to use Swift-only classes or even Swift data structures. I'll create a script to detect a drift between architectures in a follow-up PR

PR #8038, opened 2026-06-11
ci: add cross-architecture symbol drift detection

philprime:
Adds `validate-xcframework-symbols.sh` that compares ObjC class symbols across architectures within fat binaries in each XCFramework. Detects when resilient value types cause `_OBJC_CLASS_$_` symbols to be emitted on arm64 but missing on x86_64 — the class of bug fixed in #8037.

Wired into all existing XCFramework validation points:
- Local builds (`build-xcframework-local.sh`)
- CI assembly workflows (Sentry variants, SentryObjC static/dynamic)
- Release validation job

Verified locally: running the script against `SentryObjC-Dynamic.xcframework` built from `main` (before the boxing fix in #8037) correctly fails with the 4 missing symbols. After the fix, it passes.

#skip-changelog


itaybre:
I guess it shouldn't fail now after updating with main

itaybre (approved):
LGTM

PR #8041, opened 2026-06-11
fix(session-replay): Stabilize video assembly

romtsn:
Extracts the replay video-assembly fixes from #7851 into their own PR so they can be reviewed and land first; #7851 (the capture-scheduler/backoff rework) builds on top of them and will be retargeted onto this branch. Only `SentryOnDemandReplay` and `SentryVideoFrameProcessor` (plus their tests) change; the public surface of both is untouched, and nothing here references the new capture pacing.

These fixes were motivated by the backoff work, which makes irregular capture cadence the norm — but the conditions they handle already occur on `main` when captures are skipped under load, and two of them are flat-out pre-existing issues. Behavior changes, all in how segment videos are assembled from captured frames:

- The last frame captured before a segment window is retained, so recovered and resumed segments start with the correct screen state instead of dropping it. A recovered crash replay's start timestamp now reflects that retained frame (see the updated `testSessionReplayForCrash` expectation).
- Fractional gaps between captured frames are filled (and bounded) so video timing stays stable when captures are skipped.
- Video windows are half-open, so consecutive segments no longer duplicate the boundary frame.
- Segments that end up with no frames are dropped instead of being emitted as empty videos.
- Video timing stays stable when a cached frame image cannot be read back from disk.
- `oldestFrameDate` reads are serialized on the processing queue to avoid racing frame mutations.

Review the `SentryVideoFrameProcessor` diff first (frame-index based appending is the core change), then `SentryOnDemandReplay` (window selection and retained-frame bookkeeping). Verified with the four session-replay test suites (116 tests) against main's display-link capture path.


Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
```diff
@@ -58,43 +77,121 @@ class SentryVideoFrameProcessor {
                 return onCompletion(.failure(videoWriter.error ?? SentryOnDemandReplayError.errorRenderingVideo))
             }
             guard frameIndex < videoFrames.count else {
+                if let videoEnd = videoEnd, let videoStart = videoStart {
```
philprime, line 85: `m`: Please add a comment block here explaining the reasoning for this handling, especially what it means if `videoEnd` and `videoStart` are given vs. when they are absent. I don't see any documentation explaining this to future maintainers

Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
```diff
@@ -58,43 +77,121 @@ class SentryVideoFrameProcessor {
...
+            ) else { return }
+
+            frameIndex += 1
+        }
+    }
+    // swiftlint:enable function_body_length cyclomatic_complexity
+
+    private func handleAppendResult(
+        _ result: AppendFrameResult,
+        onCompletion: @escaping (Result<SentryRenderVideoResult, Error>) -> Void
```
philprime: `m`: Even though this is a refactoring, I don't see a reason why this method needs an async completion handler. Couldn't this be a throwing method with a tuple of two values instead?

Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
```diff
@@ -58,43 +77,121 @@ class SentryVideoFrameProcessor {
...
+        case .failure:
+            SentrySDKLog.error("[Session Replay] Failed to append image to pixel buffer, cancelling the writing session, reason: \(String(describing: videoWriter.error))")
+            videoWriter.inputs.forEach { $0.markAsFinished() }
+            videoWriter.cancelWriting()
+            onCompletion(.failure(videoWriter.error ?? SentryOnDemandReplayError.errorRenderingVideo))
+            return false
+        }
+    }
+
+    private func appendLastFrame(untilFrameIndex targetFrameIndex: Int, videoWriterInput: AVAssetWriterInput) -> AppendFrameResult {
```
philprime: `h`: I am a bit confused here. The method indicates that it just appends a single frame, but the logic is a loop so it can append multiple ones. Is this method repeating the last frame until the target index is reached?

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -221,16 +264,43 @@ import UIKit
...
+            retainedFrameBeforeCurrentFrames
+        }.flatMap { $0.time < date ? $0 : nil }
+        let currentFrame = _frames.last(where: { $0.time < date })
+
+        guard let retained = retainedFrame else { return currentFrame }
+        guard let current = currentFrame else { return retained }
+        return retained.time > current.time ? retained : current
+    }
+
+    private func removeFrameFile(_ frame: SentryReplayFrame) {
```
philprime: `m`: This seems to be a duplicate of the SentryVideoFrameProcessor extension introduced in this PR, can this be deduplicated?

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -41,6 +45,17 @@ import UIKit
         self.processingQueue = processingQueue
         self.assetWorkerQueue = assetWorkerQueue
     }
+
+    deinit {
+        let retainedFrame = retainedFrameLock.synchronized {
```
philprime, line 63: `m`: did not verify it, but we can not rely on cleanup in deinit for crash cases. What happens if the app restarts after a crash and the retained frame was not cleaned up? Do we need different handling between crash recovery and clean shutdown?

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
```diff
@@ -130,20 +145,32 @@ import UIKit
...
                 }
             }
             SentrySDKLog.debug("[Session Replay] Frames released, remaining frames count: \(self._frames.count)")
         }
     }
 
     public var oldestFrameDate: Date? {
-        return _frames.first?.time
+        var oldestFrameDate: Date?
+        processingQueue.dispatchSync {
```
philprime: `h`: Are certain that this is not called on the main thread? Because a synchronized dispatch from main can cause performance issues

philprime (commented):
Good progress, left a couple of comments to discuss

Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
romtsn: added in https://github.com/getsentry/sentry-cocoa/pull/8041/commits/e10075d9ab44e12cd6539960c5681858f51d9c3c

Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
romtsn: switched to a throwing method + do-catch in https://github.com/getsentry/sentry-cocoa/pull/8041/commits/7212439c9a50910c2d5064658fe689b8c3ecec66

Sources/Swift/Integrations/SessionReplay/SentryVideoFrameProcessor.swift
romtsn: renamed to `repeatLastFrame` + added a doc comment in https://github.com/getsentry/sentry-cocoa/pull/8041/commits/f28011fd366f701a5fac16ad791e833b7cb2269a

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
romtsn: created a shared helper in https://github.com/getsentry/sentry-cocoa/pull/8041/commits/b36cc551e1c2d8b2ae75f9a7ea0a2ff9592b16b2

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
romtsn: added a comment explaining it in https://github.com/getsentry/sentry-cocoa/pull/8041/commits/48952c86d7e4acbaca0c06a28132a2ba04cc07ac. For the crash case it'd be handled by the crash recovery codepath (we don't need the retainedFrame there), and after sending the last remaining segment, it would be deleted as part of the regular cleanup after recovery.

Sources/Swift/Integrations/SessionReplay/SentryOnDemandReplay.swift
romtsn: Reverted this to how it is on `main` in https://github.com/getsentry/sentry-cocoa/pull/8041/commits/20ff61cb41f1b9c0fafbd9a433fcc2d7d8f6ef85. Also changed from `public` to internal as it's only used in one codepath -- upon crash recovery. We don't need to synchronize there because we call this before any of the frames are being processed and never again. Also, no need for setting the retained frame -- If the retained PNG survived the crash, `loadFrames` already puts it into `_frames`, sorted by timestamp. Setting a separate retained field would be duplicate state for no gain.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@
 ### Fixes
 
 - Session replay video assembly: drop empty video segments, avoid duplicating frames at segment boundaries, and keep video timing stable when captured frames are skipped or unreadable (#8041)
+- Show feedback form from shake or screenshot without widget (#8050)
```
philprime, line 40: `m`: This needs to be fixed before merge

philprime (approved):
LGTM

PR #8048, opened 2026-06-11
fix(feedback): Keep the User Feedback screenshot trigger active after form dismissal

### LINKED ISSUE #7641: showFormForScreenshots only works once
### Platform

iOS

### Environment

Develop

### Installed

Swift Package Manager

### Version

8.58.0

### Xcode Version

26.3

_No response_

### Steps to Reproduce

When using the Sentry User Feedback widget with showFormForScreenshots = true, the screenshot trigger only works once. After the first screenshot opens the feedback form and the form is dismissed, subsequent screenshots no longer trigger the feedback form.

1. Enable the feedback widget:

`SentrySDK.feedback.showWidget()`

2. Configure feedback to allow screenshot-triggered forms:

`showFormForScreenshots = true`

3. Take a screenshot on the device.
4. The feedback form opens.
5. Dismiss the feedback form.
6. Take another screenshot.

### Expected Result

Every screenshot should trigger the feedback form.

### Actual Result

The feedback form only opens once.

_No response_

tsushanth:
## :scroll: Description

Fixes `showFormForScreenshots = true` only working once per app launch.

The screenshot observer now stays registered, and screenshot events are ignored while a feedback form is already displayed. The widget early-return path also registers the screenshot observer.

Fixes #7641.

Previously, `userCapturedScreenshot()` removed the screenshot observer before presenting the form, but it was never re-registered after dismissal.

Added coverage for screenshot behavior during/after form presentation and observer cleanup on driver deinit.

- [x] `make test-ios IOS_SIMULATOR_OS=18.5 IOS_DEVICE_NAME='iPhone 16e' ONLY_TESTING='SentryTests/UserFeedbackIntegrationTests'`

## :pencil: Checklist

- [x] If I added a new public API, I also added it to the [SentryObjC wrapper](../develop-docs/SENTRY-OBJC.md).


Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
denrase: I think we can get away with makig this even simpler, we can just guard here on the active form and not de-register. This also covers the case where there is no presenter for the form or one is already presented. In this case we's not re-register the observer as well.

 ```swift
 @objc func userCapturedScreenshot() {
     guard !displayingForm else {
         SentrySDKLog.debug("Screenshot ignored — feedback form is already displayed")
         return
     }

     showForm(screenshot: screenshotSource.appScreenshots().first)
 }
 ```

Tests/SentryTests/Integrations/Feedback/UserFeedbackIntegrationTests.swift
```diff
@@ -351,6 +351,97 @@ final class UserFeedbackIntegrationTests: XCTestCase {
         withExtendedLifetime(window) { }
     }
 
+    // Regression test for #7641 — `showFormForScreenshots = true` only worked
+    // on the first screenshot. After the form was dismissed, subsequent
+    // screenshots no longer triggered the feedback flow because
+    // userCapturedScreenshot() removed the notification observer and nothing
+    // re-registered it on form close.
+    func testScreenshotTrigger_afterFormDismissal_stillPresentsForm() throws {
```
denrase: Both tests somewhat overlap, think with the simple guard we can collapse this into one:

 1. Post screenshot → presents once.
 2. Post screenshot again while form is still active → still only one presentation.
 3. Dismiss form.
 4. Post screenshot again → presents second time.

denrase (changes requested):
Thank you for contributing and finding this! 🥳Left some feedback.

tsushanth:
Thanks @denrase, good catches on both. Pushed bfa4978:

1. `userCapturedScreenshot` now guards on `displayingForm` and leaves the observer attached. `userFeedbackFormDidClose` no longer calls `observeScreenshots()` and the `stopObservingScreenshots()` call inside `userCapturedScreenshot` is gone. That covers the no-presenter / already-presented cases for free and mirrors the existing `handleShakeGesture` pattern.

2. The two regression tests are collapsed into one four-step sequence: present, post-during-active (still 1), dismiss, post-after-close (now 2).

The merge conflict on the branch is on `.github/workflows/*` files I can't update through this PR (my PAT doesn't carry the `workflow` scope), so I left them as-is rather than force-push a rebased branch that would fail server-side anyway. Happy to resolve via the "update branch" button or however you'd like to handle it once you're ready to merge.

Sources/Swift/Integrations/UserFeedback/SentryUserFeedbackIntegrationDriver.swift
tsushanth: Done in bfa4978 — `userCapturedScreenshot` now does the `guard !displayingForm` you suggested and leaves the observer attached for the integration's lifetime. `userFeedbackFormDidClose` no longer touches `observeScreenshots()`, and the `stopObservingScreenshots()` call inside the screenshot handler is gone.

Tests/SentryTests/Integrations/Feedback/UserFeedbackIntegrationTests.swift
tsushanth: Done in bfa4978 — the two cases are collapsed into `testScreenshotTrigger_presentsExactlyOncePerScreenshotAcrossDismissal` with the four-step sequence: post (1), post-while-active (still 1), dismiss, post-again (now 2).

tsushanth:
Pushed [61dc439](https://github.com/getsentry/sentry-cocoa/pull/8048/commits/61dc43956) — dropped the `stopObservingScreenshots()` + re-register from `observeScreenshots()` per your suggestion. The function is now byte-identical to `main`; the only behavior change vs `main` is the `displayingForm` guard inside `userCapturedScreenshot`. That single guard covers all three subcases you flagged (active form, missing presenter, already-presented).

For the test side, the file only has one new test in the PR — `testScreenshotTrigger_presentsExactlyOncePerScreenshotAcrossDismissal` — and it's already shaped exactly as you described: post → present (1), post again with form still up → still 1, dismiss form, post → present (2). So nothing to collapse on the test side, but let me know if you'd like the comment trimmed.

denrase:
@tsushanth Thank you, I will take a look and push it over the finish line if we need anything else. 🙇

denrase (commented):
@tsushanth Thank you for contibuting! :bow I have added an additional test and cleaned up a bit. Doing general commit now to give other maintainers a change to review (/cc @philprime ) to have a second pair of eyes.

NinjaLikesCheez (approved):

denrase (approved):

denrase:
@tsushanth Thank you for contributing!

